PR #2, opened 2016-10-01
Better time ticking and sync

dktapps:
There's no need to spam clients with SetTimePacket. It's just unnecessary extra overhead. Time progression is handled automatically by the client. All that needs to be done is to ensure that client/server time calculation is the same.
### Changes
- Removed SetTimePacket spam every 5 seconds
- Time ticking now uses the current level tick rate to ensure that it remains in sync even when the server is lagging.
### Tests

Tested in various circumstances, no issues found. Please review.

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -156,8 +156,6 @@ class Level implements ChunkManager, Metadatable{
 
 	private $cacheChunks = false;
 
-	private $sendTimeTicker = 0;
-
 	/** @var Server */
 	private $server;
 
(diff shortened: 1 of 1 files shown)


hiilikewiki (approved):
Works for me.

catgirlinspace (approved):
Now clients don't receive as many packets!

GiantQuartz:
Alright, works for me.

inxomnyaa (approved):
Great. Will implement into CS.

dktapps:
@thebigsmileXD This project is not here for you to copy-paste from. If you're going to do that, at least don't advertise it.

archie426:
@dktapps Yeah, the idea is, PocketMines' future is not in shoddy spoons as dktapps rightly said. What is more, you are not exactly helping PocketMine by doing that, are you.

archie426:
ps, @dktapps can you merge plz, it all works fine

dktapps:
@archie426 Please don't turn this into an argument site. Try and stay neutral.

archie426:
Oops, sorry

thebigsmileXD:
:fu:

PR #3, opened 2016-10-01
Use of Utils::getRandomBytes(16)

SOF3:
Whichever guy did that, somehow this one just got missed out?


dktapps:
I cherry-picked that commit directly from Genisys, and BOOTUP_RANDOM was for some reason removed from Genisys by @ishitatsuyuki .

PR #4, opened 2016-10-01
Fixed RawChicken::getAdditionalEffects sometimes returns void

SOF3:


src/pocketmine/item/RawChicken.php
```diff
@@ -40,6 +40,7 @@ public function getAdditionalEffects() : array{
 		if(mt_rand(0, 9) < 3){
 			return Effect::getEffect(Effect::HUNGER)->setDuration(600);
 		}
```
dktapps: Well, this is never going to return an array anyway...

iPocket: Then, empty string or bool?

iPocket: https://github.com/pmmp/PocketMine-MP/pull/4/files#diff-561903db00d738e771f0fb58714e2084L39

dktapps: It should be an array to conform with the Food declaration. https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/item/Food.php#L50

thebigsmileXD: Damn, he is back to CS code xD

SOF3:
Fixed, please squash.

src/pocketmine/item/RawChicken.php
archie426: CS code?!

archie426: Isnt that VS code, I have never heard of CS before, which do you recommend?

dktapps: This isn't CS code. It's @PEMapModder 's code.

thebigsmileXD: The mess of no-array-food was CS

iPocket: @archie426 dump, he was talking about ClearSky code...

PR #5, opened 2016-10-01
Typo fix in fetching timezone from web

iPocket:


dktapps (approved):

SOF3:
This actually doesn't really matter. It should be considered as formatting rather than a bug or a typo.

PR #6, opened 2016-10-01
Use \pocketmine\Thread::getThreadName() directly instead of doing it manually

iPocket:


dktapps (approved):
Spotted this myself the other day, but forgot about it. Thanks.

PR #7, opened 2016-10-01
Fixed kicked for flying when walking on lily pads

dktapps:
Standing on top of a lily pad for roughly 5 seconds currently causes players to be kicked for flying.

### Changes
- Changed Flowable to return null when _recalculating_ bounding boxes.
- Fixed lily pad bounding box (1/64 tall, #blamemojang for making it different in PE)
- Removed some redundant properties (already in parent class or not needed/incorrect)

### Tests
Walked on top of lily pads, the bug no longer occurs. Tested with various other flowables and found no issues.

### DIFF
--- a/src/pocketmine/block/Flowable.php
+++ b/src/pocketmine/block/Flowable.php
@@ -21,9 +21,6 @@
 
 namespace pocketmine\block;
 
-
-
-
 abstract class Flowable extends Transparent{
 
 	public function canBeFlowedInto(){
@@ -42,7 +39,7 @@ abstract class Flowable extends Transparent{
 		return false;
 	}
 
-	public function getBoundingBox(){
+	protected function recalculateBoundingBox(){
 		return null;
 	}
 }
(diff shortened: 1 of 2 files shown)


dktapps:
@XFuryMCPE this is why I said #blameshoghicp :P took me ages to find this
https://github.com/pmmp/PocketMine-MP/blame/master/src/pocketmine/block/Flowable.php

XFuryMCPE:
wow

XFuryMCPE:
Add this to Genisys too

archie426:
@dktapps how are we supposed to add blazes as well, he might of left us a bit of help #blameshoghi

SOF3:
@XFuryMCPE I think we have entered the age of automation where when one spoon pushes a commit, other repos would automatically cherry pick it.

archie426:
Please can I ask you NOT to add it to Genisys, the owner is here working on this beautiful project so it is pointless.

dktapps:
This is not a spoon discussion forum. Enough.

archie426:
@dktapps sorry

PR #8, opened 2016-10-02
Update IntArrayTag.php

SOF3:
Thanks PhpStorm


dktapps:
Lol, I found something like this at Genisys ages ago. Thanks

PR #11, opened 2016-10-03
PhpStorm automated formatting

SOF3:
And further reviewed by human (@SOF3)

To satisfy needs of automatic formatting, block and item ID constants of the same value are placed together using `const A = 1,B = 1;` rather than `const A = 1; const B = 1;`.

Needs second review by another human. (Please use GitHub PR reviews to leave comments)

This is the last in the OCD trilogy.


dktapps (approved):
Looks fine to me. (Finally, consistency!)

src/pocketmine/utils/Random.php
```diff
@@ -100,7 +100,7 @@ public function nextSignedInt(){
 		$this->y = $this->z;
 		$this->z = $this->w;
 		$this->w = ($this->w ^ (($this->w >> 19) & 0x7fffffff)
-							 ^ ($t ^ (($t >> 8) & 0x7fffffff))) & 0xffffffff;
```
SOF3: This is necessary because tabs may have different size on different editors, and using spaces for inline indentation is more appropriate.

dktapps:
Looks fine to me. Any further changes?

PR #14, opened 2016-10-07
Added options to disable saving <player>.dat

dktapps:
Some server owners do not want to have player data saved due to the fact that players' data should be/will be reset next time they join. Saving such unnecessary data wastes processing and disk time and can also be a nuisance for large servers saving useless/unwanted data for thousands of players. Examples of this are minigame servers, where the player will always start from zero when they join.

References: https://github.com/iTXTech/Genisys/issues/2035

This pull requests adds an option to pocketmine.yml to disable the saving of `<player>.dat`.

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -668,6 +668,10 @@ class Server{
 		$this->craftingManager->registerRecipe($recipe);
 	}
 
+	public function shouldSavePlayerData() : bool{
+		return (bool) $this->getProperty("player.save-player-data", true);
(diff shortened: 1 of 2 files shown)


SOF3 (changes requested):
Delete the notice that is shown when a player joins a server and his dat file is not found, if !shouldSavePlayerData()..

SOF3 (approved):

dktapps:
Tested this, working flawlessly. Ready to merge?

PR #27, opened 2016-10-16
Removed some not needed code. Moved the default mcr/mca worlds' spawn in the center of the r.0.0.mc* region file

svilex:
$diffX,Y,Z variables were not needed since they were the same as $dx,y,z with an opposite sign and were used as a power of 2 ( $diffX,Y,Z ** 2 ) that's always positive. $diffY was also used in a condition, to check if the value was inside the interval ( -$yS < $diffY < $yS ) , so, the sign is not needed to check if it's inside this interval.

The default mcr/mca spawn was not centered in the 0,0 region causing the generation of not needed regions walking just a bit

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -1356,16 +1356,12 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 
 			$this->move($dx, $dy, $dz);
 
-			$diffX = $this->x - $newPos->x;
-			$diffY = $this->y - $newPos->y;
-			$diffZ = $this->z - $newPos->z;
-
 			$yS = 0.5 + $this->ySize;
-			if($diffY >= -$yS or $diffY <= $yS){
-				$diffY = 0;
+			if(-$yS <= $dy or $dy <= $yS){
+				$dy = 0;
 			}
 
-			$diff = ($diffX ** 2 + $diffY ** 2 + $diffZ ** 2) / ($tickDiff ** 2);
+			$diff = ($dx ** 2 + $dy ** 2 + $dz ** 2) / ($tickDiff ** 2);
 
 			if($this->isSurvival()){
 				if(!$revert and !$this->isSleeping()){
--- a/src/pocketmine/level/format/mcregion/McRegion.php
+++ b/src/pocketmine/level/format/mcregion/McRegion.php
@@ -85,9 +85,9 @@ class McRegion extends BaseLevelProvider{
 			"initialized" => new ByteTag("initialized", 1),
 			"GameType" => new IntTag("GameType", 0),
 			"generatorVersion" => new IntTag("generatorVersion", 1), //2 in MCPE
-			"SpawnX" => new IntTag("SpawnX", 128),
+			"SpawnX" => new IntTag("SpawnX", 255),
 			"SpawnY" => new IntTag("SpawnY", 70),
-			"SpawnZ" => new IntTag("SpawnZ", 128),
+			"SpawnZ" => new IntTag("SpawnZ", 255),
 			"version" => new IntTag("version", 19133),
 			"DayTime" => new IntTag("DayTime", 0),
 			"LastPlayed" => new LongTag("LastPlayed", microtime(true) * 1000),


dktapps:
Please try to keep your changes to separate commits/PRs.

svilex:
@dktapps Ok, sorry. Should i make a new PR ?

dktapps:
Leave it this time, the changes are small enough to be reviewed easily. But for future reference please keep changes separate.

src/pocketmine/level/format/mcregion/McRegion.php
```diff
@@ -85,9 +85,9 @@ public static function generate($path, $name, $seed, $generator, array $options
 			"initialized" => new ByteTag("initialized", 1),
 			"GameType" => new IntTag("GameType", 0),
 			"generatorVersion" => new IntTag("generatorVersion", 1), //2 in MCPE
-			"SpawnX" => new IntTag("SpawnX", 128),
+			"SpawnX" => new IntTag("SpawnX", 255),
```
jojoe77777: Shouldn't it be 256, if the old was 128 (not 127)?

src/pocketmine/Player.php
```diff
@@ -1356,16 +1356,12 @@ protected function processMovement($tickDiff){
 
 			$this->move($dx, $dy, $dz);
 
-			$diffX = $this->x - $newPos->x;
-			$diffY = $this->y - $newPos->y;
-			$diffZ = $this->z - $newPos->z;
```
SOF3: Seems `$diffZ` is `-$dz` instead of `$dz`?

src/pocketmine/Player.php
```diff
@@ -1356,16 +1356,12 @@ protected function processMovement($tickDiff){
 
 			$this->move($dx, $dy, $dz);
 
-			$diffX = $this->x - $newPos->x;
-			$diffY = $this->y - $newPos->y;
```
SOF3: `$diffY` is equivalent to `-$dy` not `$dy`

src/pocketmine/Player.php
```diff
@@ -1356,16 +1356,12 @@ protected function processMovement($tickDiff){
...
-			$diffZ = $this->z - $newPos->z;
-
 			$yS = 0.5 + $this->ySize;
-			if($diffY >= -$yS or $diffY <= $yS){
```
SOF3: The original code seems to have logic problems. I think it should be an `and` operator such that it can be expressed as `-$yS <= $diffY <= $yS` mathematically

src/pocketmine/Player.php
```diff
@@ -1356,16 +1356,12 @@ protected function processMovement($tickDiff){
...
-				$diffY = 0;
+			if(-$yS <= $dy or $dy <= $yS){
```
SOF3: So `-$yS <= $dy and $dy <= $yS` can be expressed as `-yS <= dy <= yS` mathematically, which is much more reasonable (`$yS > 0` assumed). Otherwise, this statement is always true.

src/pocketmine/level/format/mcregion/McRegion.php
svilex: A region contains 32 * 32 chunks.
A chunk is 16 * 16 * 128.
16 * 32 / 2 - 1 = 255

SOF3: Then why was it 128 before?

There are 512 blocks horizontally, so you want the coordinate between the 256th and 257th, i.e. index 255 and 256. The middle of the index-255 block is 255.5, so this should be 256.

svilex:
@SOF3 I can't add commits to this pull request, i deleted my fork . Is there another way?

dktapps:
@svilex That's very bad practice. You should always retain it until the PR is merged. Deleting it simply makes it unchangeable.

SOF3:
@svilex close this pull request.

Your new fork still shares the same Git tree, so you can still create a new branch in your new fork and merge the new commit 38d328a31053399f0968a2dec63dc943ba62167c into your branch:

```bash
git clone https://github.com/svilex/PocketMine-MP.git --recursive && cd PocketMine-MP
git checkout 38d328a31053399f0968a2dec63dc943ba62167c
git checkout -b my-new-branch
git push -u origin my-new-branch
```

SOF3:
Edit: Better not merge this commit directly. Make two new pull requests instead.

PR #29, opened 2016-10-16
Removed unneeded vars - fixed logical condition

svilex:


svilex:
I think there are floating point inaccuracy problems here:

sometimes `$diffY` isn't the same as `-$dy` .

dktapps:
It's been reverted, the commit caused some movement issues.

SOF3:
@svilex I think `$a - $b` and `-($b - $a)` always give the same floating point result.

svilex:
@SOF3 That's what i tought but , try using a `var_dump()`

PR #35, opened 2016-10-18
Fix --disable-readline command-line option does not work, close #34

### LINKED ISSUE #34: --disable-readline command-line option does not work
### Issue description
`readline` cannot be disabled with `--disable-readline` if the extension is present.

1. Run a server on a Linux box with `readline` PHP extension, with `--disable-readline` option.

### OS and versions

* PocketMine-MP: https://github.com/pmmp/PocketMine-MP/commit/d53810c6bc5676233cf85cfe3abc15dea3c8482a
* PHP: 7.0.10 with readline extension
* Server OS: Any Linux distro
* Game version: irrelevant

dktapps:
Global variables are thread-local. Calling getopt() in a thread will attempt to retrieve the arguments of the thread, _not_ the parent process. This PR fixes that by checking command-line arguments in the thread constructor instead of in the thread's run() method.

### DIFF
--- a/src/pocketmine/command/CommandReader.php
+++ b/src/pocketmine/command/CommandReader.php
@@ -31,6 +31,8 @@ class CommandReader extends Thread{
 
 	public function __construct(){
 		$this->buffer = new \Threaded;
+		$opts = getopt("", ["disable-readline"]);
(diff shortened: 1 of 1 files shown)


dktapps:
😞 it still doesn't work on Travis

PR #40, opened 2016-10-21
Extra action on use of cURL in main thread

SOF3:
# Introduction

It is a very bad idea to access the Internet on the main thread, especially HTTP requests. They slow down the execution of the main thread and is highly discouraged in a production server.
# Description

This pull request adds the option to carry out extra action when the `Utils::getURL()` or `Utils::postURL()` methods are used (there should be more, but I can only think of these two functions right now; I am doubtful if it is possible to do it with others like mysqli).

Actions are listed in pocketmine.yml:

``` yaml
#If set to deny, an exception will be thrown (you may see an error)
#If set to warn, a warning will be shown on console and Internet access continues
#If set to empty, a warning will be shown on console, and empty content is returned (will not download anything from the Internet).
#If set to allow, nothing will happen and Internet access continues
```
# Changes
## For server admins

A new option in `pocketmine.yml` is added: map `settings.sync-internet-access`, including:
- enum string `settings.sync-internet-access.curl`
- enum string `settings.sync-internet-access.suppressed`

If pocketmine.yml is left unchanged, a warning is logged every time a plugin uses `Utils::getURL()` or `Utils::postURL()`.
## For developers

An `Utils::syncInternetAccess` method can be used for if you **really** want to access the Internet in the main thread.

Public static boolean method `Utils::isMainThread()` is added for checking whether the current context is in the main thread.
# To-do
- Language values `pocketmine.thread.useInternet.curl` and `pocketmine.thread.useInternet.suppressed` should be added.
# Possible enhancements

Add timings specific for syncInternetAccess calls.
# Tests

This script plugin has been used for testing this pull request:

``` php
<?php

/**
 * @name TestSyncInternet
 * @main TestSyncInternet\Main
 * @version 1.0.0
 * @api 2.0.0
 */

use pocketmine\utils\Utils;
use pocketmine\event\server\QueryRegenerateEvent as QRE;

class Main extends \pocketmine\plugin\PluginBase implements \pocketmine\event\Listener{
        public function onLoad(){
                echo "Load: ", strlen(Utils::getURL("http://example.com")), PHP_EOL;
        }

        public function onEnable(){
                echo "Enable: ", strlen(Utils::getURL("http://example.com")), PHP_EOL;
                $this->getServer()->getScheduler()->scheduleDelayedTask(new class($this) extends \pocketmine\scheduler\PluginTask{
                        public function onRun($ticks){
                                $server = \pocketmine\Server::getInstance();
                                $server->getPluginManager()->callEvent(new QRE($server, 5));
                                $server->getScheduler()->scheduleAsyncTask(new class extends \pocketmine\scheduler\AsyncTask{
                                        public function onRun(){
                                                echo "Async: ", strlen(Utils::getURL("http://example.com")), PHP_EOL;
                                        }
                                });
                        }
                }, 10);
        }

        public function onDisable(){
                echo "Disable: ", strlen(Utils::getURL("http://example.com")), PHP_EOL;
        }

        public function e_regenEvent(QRE $e){
                echo "Regen.sync: ", strlen(Utils::syncInternetAccess($this, [Utils::class, "getURL"], "http://example.com")), PHP_EOL;
                echo "Regen: ", strlen(Utils::getURL("http://example.com")), PHP_EOL;
        }
}
```

Works as expected under different values for the new options in pocketmine.yml

### DIFF
--- a/src/pocketmine/resources/pocketmine.yml
+++ b/src/pocketmine/resources/pocketmine.yml
@@ -32,6 +32,9 @@ settings:
  sync-internet-access:
   #For HTTP requests
   curl: warn
+  #If the plugin wants to suppress the actions specified above
+  #"empty" is not allowed in this option
+  suppressed: allow
 
 
 memory:
--- a/src/pocketmine/resources/pocketmine.yml
+++ b/src/pocketmine/resources/pocketmine.yml
@@ -23,6 +23,16 @@ settings:
  #Set this approximately to your number of cores.
  #If set to auto, it'll try to detect the number of cores (or use 2)
  async-workers: auto
+ #Allow plugins to access the Internet on the main thread (except during a plugin is loaded/unloaded).
+ #Plugins can still bypass this check if they specify that intentionally want to access the Internet on the main thread
+ #If set to deny, an exception will be thrown (you may see an error)
+ #If set to warn, a warning will be shown on console and Internet access continues
+ #If set to empty, a warning will be shown on console, and empty content is returned (will not download anything from the Internet).
+ #If set to allow, nothing will happen and Internet access continues
+ sync-internet-access:
+  #For HTTP requests
+  curl: warn
+
 
 memory:
  #Global soft memory limit in megabytes. Set to 0 to disable
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -1404,6 +1404,7 @@ class Server{
 	public function __construct(\ClassLoader $autoloader, \ThreadedLogger $logger, $filePath, $dataPath, $pluginPath){
 		self::$instance = $this;
 		self::$sleeper = new \Threaded;
+		Utils::$mainThreadId = \Thread::getCurrentThreadId();
 		$this->autoloader = $autoloader;
 		$this->logger = $logger;
 
@@ -2040,7 +2041,9 @@ class Server{
 
 		$this->logger->info($this->getLanguage()->translateString("pocketmine.server.startFinished", [round(microtime(true) - \pocketmine\START_TIME, 3)]));
 
+		Utils::$serverRunning = true;
 		$this->tickProcessor();
+		Utils::$serverRunning = false;
 		$this->forceShutdown();
 	}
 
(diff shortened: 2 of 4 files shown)


ishitatsuyuki:
It's simply a warning though. That's why we should use reactor pattern...

iPocket (approved):

SOF3:
@ishitatsuyuki actually isn't that what the AsyncTask::onCompletion() is for?

ishitatsuyuki:
Yeah, that's partially true, however a nonblocking pattern will be... Well, off topic.

src/pocketmine/resources/pocketmine.yml
```diff
@@ -23,6 +23,19 @@ settings:
...
+ #If set to deny, an exception will be thrown (you may see an error)
+ #If set to warn, a warning will be shown on console and Internet access continues
+ #If set to empty, a warning will be shown on console, and empty content is returned (will not download anything from the Internet).
+ #If set to allow, nothing will happen and Internet access continues
+ sync-internet-access:
+  #For HTTP requests
+  curl: warn
+  #If the plugin wants to suppress the actions specified above
+  #"empty" is not allowed in this option
+  suppressed: allow
```
robske110: i don't know if this should be an option

dktapps (approved):
Tighter restrictions on internet access definitely ought to be imposed.

dktapps:
I think this is somewhat redundant to be honest. You can apply restrictions to these things, but plugins can still use curl directly, not to mention the fact that a method exists to allow use on the main thread anyway, which would also be easily abused to get around this restriction. I think this is not foolproof and leaves most bases uncovered.

robske110:
Adding to that, I have not found a single plugin that does this during runtime. On startup/shutdown some do that. But never during runtime.

madcode15:
**Bump.**

**Anyways, I am sort of with of what @dktapps said. There is just too many problems with this.**

- Developers can easily bypass this by making there own methods.
- The extra config options just make user experience more confusing.
- Owners can simply remove the plugin causing the issue. If the server lags due to a plugin abusing these methods, timings will call out the plugin as a consequence.
- What this is really doing is limiting a API function that was put here in the first place to be used by developers. If you are so concerned about it, why not just remove it?

PR #50, opened 2016-10-23
Remove useless statement

iPocket:


SOF3 (approved):
Really, this almost doesn't affect performance at all, but I'm fine with this change.

legoboy0215:
I wanted to do that too, but doesn't this seem too minuscule? Anyways, that extension seems to be for Unit testing anyways.

SOF3 (dismissed):
The extension being totally unused in the core, the only difference would be to completely delete these lines or not. Whether commenting out one or two more lines does not affect anything at all.

legoboy0215:
What about the Pocketmine native extension? I never heard of anyone using it lol

SOF3:
@legoboy0215 it is probably an extension written for pre-1.4.1 versions.

dktapps (approved):

SOF3 (approved):

PR #55, opened 2016-10-27
Improved action check for InteractPacket

jojoe77777:
As of 0.16, when a player highlights (right-click/tap) an entity, the client starts spamming InteractPackets with an action of 4. The server checks that it's not a right-click, and assumes it's left-click. Now the server only processes it as an attack when the action is 1 (left-click)

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -2324,7 +2324,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 
 				$cancelled = false;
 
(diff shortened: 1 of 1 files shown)


dktapps:
This needs to be checked against vanilla behaviour to find out if the spam issue also happens in vanilla. Good work though.

jojoe77777:
@dktapps Still, only a left-click should be an attack.

dktapps:
@jojoe77777 that's what I mean, and yet this doesn't happen in vanilla MCPE afaik. Needs further investigation.

dktapps:
TODO: Rewrite this handler and also use action 4 properly. Good work.

PR #57, opened 2016-10-27
0.16 particle protocols.

Muqsit:
12 and 18 were causing client-sided crashes btw..

### DIFF
--- a/src/pocketmine/level/particle/Particle.php
+++ b/src/pocketmine/level/particle/Particle.php
@@ -24,41 +24,43 @@ namespace pocketmine\level\particle;
 use pocketmine\math\Vector3;
 use pocketmine\network\protocol\DataPacket;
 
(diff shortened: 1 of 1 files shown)


src/pocketmine/level/particle/Particle.php
```diff
@@ -24,41 +24,43 @@
 use pocketmine\math\Vector3;
 use pocketmine\network\protocol\DataPacket;
 
-abstract class Particle extends Vector3{
-
+abstract class Particle extends Vector3{	
+
```
SOF3: Why?

Muqsit: same thing right?

dktapps:
We're already testing particles internally and a few of these IDs are incorrect. Thanks anyway.

Muqsit:
no probs mate.

PR #58, opened 2016-10-28
add sea lantern

FunnyCrafteur:


dktapps:
Thanks for contribution, but this is already present on the dev-0.16/blocks branch.

PR #67, opened 2016-10-30
Fixed detection of system PHP binaries

tnytown:


SOF3 (approved):
Tested on Git Bash on Windows, working well.

dktapps (approved):
Other users have informed me that it is working well.

Issue #24, opened 2016-10-14
Level object memory leak on world unload


dktapps:
## Issue description

When a world is unloaded, the Level object is not destroyed due to Position objects holding references to it. This causes myriad bugs, the most famous ones being examples like getName() on null (https://github.com/PocketMine/PocketMine-MP/pull/4205), Level::getProvider returning null (https://github.com/iTXTech/Genisys/issues/1908).

The memory leak is very minor, barely worthy of the name "leak" at all. However, the effects are widespread.
## Why this happens

Players with spawn points in worlds that are later unloaded will, for example, hold Position objects pointing to their spawn point in the unloaded world. These Position objects hold strong references to the Level object, which is subsequently not destroyed even after everything related to it is closed, most notably the LevelProvider.

The getName() on null bug occurs when a player with a spawn point in an unloaded world either attempts to teleport back to their spawn point, or when they quit from the server and their data is saved.
## Steps to reproduce the issue

1. Set a player's spawn point in world x
2. Unload world x
3. Kill the player, or otherwise attempt to teleport them to their spawn point.
### Use Level ID

Instead of holding a direct Level reference, Position objects could instead keep the level _ID_ of the level that it points to. Then, when getLevel() is called on a Position object, obtain a Level reference from the server directly. If the world is an unloaded world, this will return null.
#### Problems
- Every Position object would either need to hold a Server reference, or obtain one on-demand using Server::getInstance().
- Position currently has a _public_ Level property. This is used directly by many things in the core, and removal of this property would require either a magic __get() method or all access to the level property to be done with Position::getLevel(), which would have to get a Level reference from the Server ondemand. Both of these solutions have potential to impact on performance due to how widespread the use of Position is.
### Use WeakRef

Use WeakRef to hold a Level weak reference instead of a strong one. This would eliminate the problem, but would bring back some performance issues which are the reason Position weak references were removed before now (they previously did use weak references).
#### Problems
- WeakRef doesn't play nice on PHP7, especially when used with pthreads.
- Performance/memory impacts.
## OS and versions

- PocketMine-MP: https://github.com/pmmp/PocketMine-MP/commit/2fba1073ddf4965eb5d2bbbbbfb4a5c5f9377db3
- PHP: doesn't matter
- OS: doesn't matter

legoboy0215:
You finally found the reason 👍 

I think the first solution is more viable.

dktapps:
[...]

PR #70, opened 2016-10-31
Fixed bugs related to #24 such as getName() on null

dktapps:
This pull request fixes the leak described in #24 and a very large number of bugs that resulted from the issue.

Description of the actual issue can be seen in #24.

This pull request adds a new type of Position, WeakPosition, which instead of holding a Level _reference_, holds a level _ID_ instead. When WeakPosition::getLevel() is called, a Level reference is returned from Server ondemand, or null if the level has been unloaded.

WeakPosition is now used for player spawn positions, which was the primary cause of this bug. Now, instead of the famous crashes, the player will simply return to the server's default spawn point if the world in which the player's spawn point is set is not loaded.

### Possible improvements
- Use the level name instead of ID. This will allow spawnpoints to return to the world if it is reloaded. Using a Level ID is faster, but once the subject world is unloaded, the spawnpoint will never be re-validated until the server is restarted.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -86,6 +86,7 @@ use pocketmine\level\Level;
 use pocketmine\level\Location;
 use pocketmine\level\Position;
 use pocketmine\level\sound\LaunchSound;
+use pocketmine\level\WeakPosition;
 use pocketmine\math\AxisAlignedBB;
 use pocketmine\math\Vector2;
 use pocketmine\math\Vector3;
@@ -228,7 +229,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	protected $viewDistance;
 	protected $chunksPerTick;
 	protected $spawnThreshold;
-	/** @var null|Position */
+	/** @var null|WeakPosition */
 	private $spawnPosition = null;
 
 	protected $inAirTicks = 0;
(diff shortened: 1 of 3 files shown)


SOF3:
Note:
* This may require a major API bump since this is a backwards-incompatible change.
* If you are making such a backwards-incompatible change, it is a good idea to extend it to other places where available.
* You may want to specify that `Position::$level` should no longer be accessed directly, but instead with `Position::getLevel()`.

robske110:
Wait, i don't see any plugin break here...

SOF3:
Oh right, it is implemented better than I expected.

robske110:
:P

SOF3:
What about changing `Position::$level` to private and add a `__get` method? This will probably a bit slower, but this can discourage people from using it instead of `Position::getLevel()`.

robske110:
I don't think that would be a good idea, how about just deprecting accessing $level directly? (In the original Position.
A third way would be to add __get methods in Position and WeakPosition.

\

And how about creating WeakLocation? (Simply location but extending WeakPosition)

dktapps:
@SOF3 performance issue. This is _only_ used for player spawn points, nothing else.

Deprecating $level is a very complex task, which is why I opted for this approach instead. This approach cleanly fixes the bug without performance issues and requiring a bare minimum of changes.

robske110:
> And how about creating WeakLocation? (Simply location but extending WeakPosition)

dktapps:
@robske110 unnecessary, this pull request is primarily fixing a bug. Further features can be implemented later.

dktapps:
Does anyone have any objections to merging this?

SOF3 (approved):

PR #73, opened 2016-11-03
Added whisper

Kripth:
It displays the message at commands.message.display.incoming (%s whispers to you: %s)

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -2981,10 +2981,14 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 		$this->dataPacket($pk);
 	}
 	
-	public function whisper($from, $message){
+	/**
+	 * @param string|Player $sender
+	 * @param string        $message
+	 */
+	public function sendWhisper($sender, $message){
 		$pk = new TextPacket();
 		$pk->type = TextPacket::TYPE_WHISPER;
-		$pk->source = $from instanceof Player ? $from->getDisplayName() : $from;
+		$pk->source = $sender instanceof Player ? $sender->getDisplayName() : $sender;
 		$pk->message = $message;
 		$this->dataPacket($pk);
 	}
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -2982,13 +2982,13 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	}
 	
 	/**
-	 * @param string|Player $sender
-	 * @param string        $message
+	 * @param string $sender
+	 * @param string $message
 	 */
 	public function sendWhisper($sender, $message){
 		$pk = new TextPacket();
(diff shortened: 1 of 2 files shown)


Sandertv:
Is this really needed? Plugins like EssentialsPE provide this feature as well.

src/pocketmine/Player.php
```diff
@@ -2980,6 +2980,14 @@ public function sendTip($message){
 		$pk->message = $message;
 		$this->dataPacket($pk);
 	}
+	
+	public function whisper($from, $message){
+		$pk = new TextPacket();
+		$pk->type = TextPacket::TYPE_WHISPER;
+		$pk->source = $from instanceof Player ? $from->getDisplayName() : $from;
```
dktapps: Are you sure this is a good idea? What if the parameter is some other non-player object? I think it would be a better idea to restrict this directly to strings only. Please use type-hinting where possible.

src/pocketmine/Player.php
```diff
@@ -2980,6 +2980,14 @@ public function sendTip($message){
 		$pk->message = $message;
 		$this->dataPacket($pk);
 	}
+	
+	public function whisper($from, $message){
```
SOF3: Weak typing. Please specify the types through either type hints or doc-comment.

SOF3 (dismissed):

src/pocketmine/Player.php
SOF3: Actually, it would be better to call it `sendWhisper` in consistency with other similar methods.

robske110 (approved):
Looks good now!

dktapps:
Please see my earlier comment on the diff.

src/pocketmine/Player.php
Kripth: I took sendMessage as example, where the parameter could be a string or a TextContainer. If this causes misunderstandings I'll limit it to a string.

dktapps: I was actually referring to the possibility that the sender parameter could be something else _apart_ from a Player or string. Supplying such an invalid object would potentially cause a fatal crash

robske110: Fatal crashes should occur if someone called a function in a totally wrong way right?

SOF3:
@dktapps why not use type hints directly?

dktapps:
@SOF3 consistency because I'm OCD about stuff like this... PROBLEM?

dktapps:
We can finish up on the typehinting mess in php7.1 and in a future API update. For now, I think this is good to merge.

dktapps (approved):

SOF3 (approved):

legoboy0215:
I have a question... What does this really do?

robske110:
Add a command similar to /tell?

dktapps:
@legoboy0215 nothing yet, just adds an API method and functionality

PR #77, opened 2016-11-04
Critical error upon missing submodules

SOF3:


dktapps (approved):

hiilikewiki (approved):

src/pocketmine/PocketMine.php
```diff
@@ -120,7 +120,11 @@ function dummy(){
 	$autoloader->addPath(\pocketmine\PATH . "src" . DIRECTORY_SEPARATOR . "spl");
 	$autoloader->register(true);
 
-	if(!class_exists(RakLib::class)){
+	try{
+		if(!class_exists(RakLib::class)){
+			throw new \Exception;
```
dktapps: why ...?

SOF3: This line actually should never run. Just in case PHP somehow no longer throws that exception. To imitate a class not found exception.

I don't want velociraptors under my bed, so I didn't use goto statements.

PR #80, opened 2016-11-05
Add new ResourcePack Packets

EmreTr1:
Added ResourcePackChunkDataPacket and ResourcePackChunkRequestPacket


dktapps:
This is currently unnecessary, it will be added when it's needed.

EmreTr1:
@dktapps okey 👍

Issue #83, opened 2016-11-05
Gamemode text output issue on client


L3ice:
### Issue description

When you change the gamemode of a player via command, it does not show to which gamemode it was changed.

### Steps to reproduce the issue

1. Change gamemode from the server or the client
2. Look at the output on the client

### OS and versions

* PocketMine-MP: https://github.com/pmmp/PocketMine-MP/commit/557257baed69b8d77e8188ca133f611c8453ab60
* PHP: 7
* Server OS: Windows 10
* Game version: Win 10 & PE

### Crashdump, backtrace or other files

*![upload](https://cloud.githubusercontent.com/assets/5413827/20038271/bcca0944-a430-11e6-9a24-7a99ac54bb05.png)

Sandertv:
I can confirm this.

iPocket:
"Non-Image content-type returned"
Reupload?

Awzaw:
this can be fixed by setting force-language to true in pocketmine.yml

L3ice:
But using the language files from pocketmine is not a good solution if u have clients with different languages

PR #86, opened 2016-11-05
Partial fix for #83

Intyre:
Shows the correct message ingame but not in the console.
PocketMine doesn't handle the translation for gamemode changes, MCPE
does. No idea why but it requires the second and third parameter....

```
commands.gamemode.success.other=Set %2$s's game mode to %3$s
commands.gamemode.success.self=Set own game mode to %3$s
```

To solve the problem we have to change the language files and let
PocketMine-MP translate it.

### DIFF
--- a/src/pocketmine/command/defaults/GamemodeCommand.php
+++ b/src/pocketmine/command/defaults/GamemodeCommand.php
@@ -77,10 +77,10 @@ class GamemodeCommand extends VanillaCommand{
 			$sender->sendMessage("Game mode change for " . $target->getName() . " failed!");
 		}else{
 			if($target === $sender){
-				Command::broadcastCommandMessage($sender, new TranslationContainer("commands.gamemode.success.self", [Server::getGamemodeString($gameMode)]));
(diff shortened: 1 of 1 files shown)


SOF3:
Through common-sense guessing, I suspect they actually share the same set of parameters, namely `?`, `target player` and `target mode`. `?` is likely to be the doer. ~~It might support MCPE in the future better if you set them like this.~~ Never mind about the guesses. They will just change them again the next version.

dktapps:
This [patch](https://github.com/pmmp/PocketMine-MP/files/624723/fix-gamemode.patch.txt) fixes the console-side issue and also force-language.

Muqsit:
What the heck xD

robske110:
bump. @dktapps suggested a patch to be applied to this PR...

dktapps:
This will have to wait until #207 is resolved. It doesn't make sense to merge this without translations server-side.

dktapps:
Rebased, tested this, works great. Will merge this after merging language changes.

PR #95, opened 2016-11-09
Make /timings command more verbose about errors

dktapps:
It's highly irritating running /timings paste and getting that generic "An error happened..." message. This PR makes the command throw and log an exception with the error details if curl_exec fails.


SOF3 (approved):
So you are using this mini-exception style too :D

PR #96, opened 2016-11-10
Safer anvil section requests, fix "top-of-the-map" spawn position crash

dktapps:
Fix the nasty crash described by @awzaw internally.

This required a a little maths on my part to work out how to reproduce, but the way to do it is to create a sandwich of blocks.

Create a flat **ANVIL** world with the below settings:
```
worlds:
 #These settings will override the generator set in server.properties and allows loading multiple levels
 #Example:
 world:
  seed: 404
  generator: FLAT:2;7,120x1,3x3,2;1;decoration(treecount=80 grasscount=45)
```
You will spawn at y = 125.
- Leave an empty 1-block gap, and build a solid layer.
- Stand on top of this layer and set the world spawn on top of it.
- Use /kill on yourself or quit/rejoin. You will see this error:
```
Notice: Undefined offset: 8 in C:\Users\Admin\Downloads\server\src\pocketmine\level\format\generic\BaseChunk.php on line 81
[11:35:57] [Server thread/CRITICAL]: Error: "Call to a member function getFullBlock() on null" (EXCEPTION) in "/src/pocketmine/level/format/generic/BaseChunk" at line 81
```

## Why this happens
Level->getSafeSpawn() misuses Chunk->getFullBlock() to skip retrieving the target chunk every time a block in the column is to be checked. While this is better for performance, there is a [point in the method ](https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/level/Level.php#L2609) where it will pass 127 to the next part. While this is fine, since players can spawn higher than this, the +1 in the next part checking headroom causes the method to request a block from a section which does not exist. 128 >> 4 = 8, but there are only indexes 0-7 of sections.

This could happen to any misused BaseChunk methods due to the lack of checks on the supplied Y coordinate. I expect @shoghicp didn't consider this because those methods are not API methods.

## Change details
Instead of accessing BaseChunk->sections directly, all methods which could encounter this issue now use BaseChunk->getSection(). getSection() has been altered to return a dummy EmptyChunkSection.

## Tests
This has been tested and confirmed that the described crash does not occur anymore.

### DIFF
--- a/src/pocketmine/level/format/anvil/Chunk.php
+++ b/src/pocketmine/level/format/anvil/Chunk.php
@@ -82,12 +82,12 @@ class Chunk extends BaseChunk{
 		foreach($this->nbt->Sections as $section){
 			if($section instanceof CompoundTag){
 				$y = (int) $section["Y"];
-				if($y < 8){
+				if($y < self::SECTION_COUNT){
 					$sections[$y] = new ChunkSection($section);
 				}
 			}
 		}
-		for($y = 0; $y < 8; ++$y){
+		for($y = 0; $y < self::SECTION_COUNT; ++$y){
 			if(!isset($sections[$y])){
 				$sections[$y] = new EmptyChunkSection($y);
 			}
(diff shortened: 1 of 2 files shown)


dktapps:
This has been rendered obsolete by internal changes made to the chunk implementation for 0.17.

PR #99, opened 2016-11-11
Disallow scheduling the same AsyncTask instance more than once

SOF3:
Normally an AsyncTask should not be scheduled more than once. Otherwise, some states in the task may be messed up, causing potentially unexpected issues.

This pull request enforces such check.


iPocket (approved):
Good check.

SOF3 (commented):
Need to implement the same check to schedule to worker.

SOF3:
Rebased in da23568

PR #103, opened 2016-11-14
Add client translations for Bed

L3ice:
Added the use of the built-in translation from mcpe.

### DIFF
--- a/src/pocketmine/block/Bed.php
+++ b/src/pocketmine/block/Bed.php
@@ -87,7 +87,7 @@ class Bed extends Transparent{
 				$b = $blockWest;
 			}else{
 				if($player instanceof Player){
-					$player->sendMessage(new TranslationContainer(TextFormat::GRAY . "%tile.bed.notValid"));
+					$player->sendMessage(TextFormat::GRAY . "This bed is incomplete");
 				}
 
 				return true;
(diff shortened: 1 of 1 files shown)


robske110 (changes requested):
You also need to include this in the PM translations, in order to not break force-language

SOF3 (changes requested):
As @robske110 mentioned.

src/pocketmine/block/Bed.php
```diff
@@ -86,15 +87,15 @@ public function onActivate(Item $item, Player $player = null){
 				$b = $blockWest;
 			}else{
 				if($player instanceof Player){
-					$player->sendMessage(TextFormat::GRAY . "This bed is incomplete");
+					$player->sendMessage(new TranslationContainer(TextFormat::GRAY . "%tile.bed.notValid"));
```
dktapps: In the game files, this is actually "Your home bed was missing or obstructed". I don't think this is correct.

dktapps (dismissed):

robske110:
I qoute:
@robske110 suggested changes ? days ago
You also need to include this in the PM translations, in order to not break force-language
 @SOF3	SOF3 requested changes 17 days ago
As @robske110 mentioned.

L3ice:
I think they know that...

dktapps:
Merged changes in the referenced commit.

PR #121, opened 2016-11-23
Remove circular dependency between Item and NBT modules

dktapps:
This pull request removes a circular dependency between the Item and NBT modules. This will allow the NBT module to be removed from the core and isolated as a standalone library which can be used for other purposes outside of PocketMine.

## API changes
The API changes in this pull request may be bundled with the 1.6.1 release for API 2.1.0.
- Removed `NBT::putItemHelper()` and `NBT::getItemHelper()`. These have been replaced by `Item->nbtSerialize()` and `Item::nbtDeserialize()` respectively.


SOF3 (approved):

robske110 (approved):
Nice!

PR #125, opened 2016-11-25
Removed outdated GUI submodule

dktapps:


SOF3 (approved):
Seems that it doesn't need to be a submodule of PocketMine at all. PocketMine-MP-GUI doesn't seem to directly interact with classes from PocketMine.

PR #126, opened 2016-11-26
Improved varint write performance

dktapps:
Significantly faster especially for single values.

This has only been tested internally with 0.17 and has not been tested with 0.16 due to my current lack of a 0.16 test device. This can be merged as soon as it has been tested.


robske110 (approved):
Tested on 0.16, works.

PR #134, opened 2016-12-01
Fix #133

### LINKED ISSUE #133: Flow calculations need a total rewrite
### Issue description
We recently fixed a CPU leak relating to still water bodies. Flow calculations are still TERRIBLE for performance and need rewriting.

### OS and versions

* PocketMine-MP: 756760a
* PHP: 7.0.13 x64
* Server OS: Linux
* Game version: irrelevant

* ...

dktapps:
Do not perform scheduled updates on still lava/water. This is in line with vanilla behaviour, where still water only starts to spread out when a neighbouring block is updated.

This has been tested and confirmed to fix the CPU leak described in #133 , however I am not sure this is the best way to fix the issue. Please review.

(Flow calculations really need a total rewrite, but that's a job for another time.)

### DIFF
--- a/src/pocketmine/block/StillLava.php
+++ b/src/pocketmine/block/StillLava.php
@@ -21,9 +21,18 @@
 
 namespace pocketmine\block;
 
+use pocketmine\level\Level;
+
 class StillLava extends Lava{
 
 	protected $id = self::STILL_LAVA;
+
+	public function onUpdate($type){
+		if($type !== Level::BLOCK_UPDATE_SCHEDULED){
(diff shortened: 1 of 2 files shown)


dktapps:
This may be somewhat hacky, but I consider it a critical bugfix and as I said flow calculations need a total rewrite regardless. Does anyone have any objections to merging this?

thebigsmileXD:
Well. Whats about.. When you break a block next to it. Shouldn't it start flowing? MCPE rivers generate still water too. it flows on update.

PR #141, opened 2016-12-04
Entity Effect Events

JackNoordhuis:
Add EntityEffectEvent class that the EntityEffectAddEvent and
EntityEffectRemoveEvent classes extend. Add event calls to Effect class.

This is purely an API addition that allows developers to modify the effect behaviour through the use of plugins.

I have tested this and it works as expected. Some more testing may need to be done to make sure the event's work as expect in all cases.

### DIFF
--- a/src/pocketmine/entity/Effect.php
+++ b/src/pocketmine/entity/Effect.php
@@ -22,6 +22,8 @@
 namespace pocketmine\entity;
 
 use pocketmine\event\entity\EntityDamageEvent;
+use pocketmine\event\entity\EntityEffectAddEvent;
+use pocketmine\event\entity\EntityEffectRemoveEvent;
 use pocketmine\event\entity\EntityRegainHealthEvent;
 use pocketmine\event\player\PlayerExhaustEvent;
 use pocketmine\network\protocol\MobEffectPacket;
@@ -251,6 +253,10 @@ class Effect{
 	}
 
 	public function add(Entity $entity, $modify = false, Effect $oldEffect = null){
+		$entity->getLevel()->getServer()->getPluginManager()->callEvent($ev = new EntityEffectAddEvent($entity, $this, $modify, $oldEffect));
+		if($ev->isCancelled()){
+			return;
+		}
 		if($entity instanceof Player){
 			$pk = new MobEffectPacket();
 			$pk->eid = 0;
@@ -258,7 +264,7 @@ class Effect{
 			$pk->amplifier = $this->getAmplifier();
 			$pk->particles = $this->isVisible();
 			$pk->duration = $this->getDuration();
-			if($modify){
+			if($ev->willModify()){
 				$pk->eventId = MobEffectPacket::EVENT_MODIFY;
 			}else{
 				$pk->eventId = MobEffectPacket::EVENT_ADD;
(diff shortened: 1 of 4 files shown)


JackNoordhuis:
This might be useful to anyone wanting to test out these changes: https://github.com/JackNoordhuis/PocketMine-Plugins/tree/master/EffectEventTesting

src/pocketmine/entity/Effect.php
```diff
@@ -251,14 +253,19 @@ public function setColor($r, $g, $b){
 	}
 
 	public function add(Entity $entity, $modify = false, Effect $oldEffect = null){
+		$entity->getLevel()->getServer()->getPluginManager()->callEvent($ev = new EntityEffectAddEvent
+		($entity, $this, $modify, $oldEffect));
```
dktapps: please fix your formatting here.

JackNoordhuis: Accidently changed some settings in PHPStorm when I was messing around the other day, fixed.

dktapps (approved):
Looks good to me, but will let someone else take a look also.

JackNoordhuis:
Squashed commits just because a derpy formatting fix doesn't deserve its own commit and for convenience.

src/pocketmine/event/entity/EntityEffectAddEvent.php
```diff
@@ -0,0 +1,61 @@
...
+	/** @var Effect */
+	private $oldEffect;
+
+	public function __construct(Entity $entity, Effect $effect, $modify, $oldEffect){
+		parent::__construct($entity, $effect);
+		$this->modify = $modify;
+		$this->oldEffect = $oldEffect;
+	}
+
+	public function willModify(): bool{
```
SOF3: I thought our code style is `willModify() : bool{`

SOF3 (changes requested):

src/pocketmine/event/entity/EntityEffectAddEvent.php
dktapps: OCD ftw...

src/pocketmine/event/entity/EntityEffectAddEvent.php
```diff
@@ -0,0 +1,61 @@
...
+		return $this->modify;
+	}
+
+	public function setWillModify(bool $modify){
+		$this->modify = $modify;
+	}
+
+	public function hasOldEffect(): bool{
```
dktapps: @SOF3 missed one :P

src/pocketmine/event/entity/EntityEffectEvent.php
```diff
@@ -0,0 +1,41 @@
...
+
+	/** @var Effect */
+	private $effect;
+
+	public function __construct(Entity $entity, Effect $effect){
+		$this->entity = $entity;
+		$this->effect = $effect;
+	}
+
+	public function getEffect(): Effect{
```
dktapps: and here ;)

SOF3: I don't need to triple comment these, do I ;)

dktapps: maybe? he won't miss any of them now :O

src/pocketmine/event/entity/EntityEffectAddEvent.php
JackNoordhuis: Had a feeling it was supposed to be like that, will change.

JackNoordhuis:
Fixed the return type hint formatting.

SOF3 (approved):

PR #148, opened 2016-12-05
Deprecated and renamed Level::getName() to Level::getSavedName()

SOF3:
As this method is often misunderstood as the getter for the unique name of the level, it has been renamed to `Level::getSavedName()`. If a more reliable level identifier is desired, consider using `Level::getId()` or `Level::getFolderName()` instead.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -1642,7 +1642,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 
 		if(($level = $this->server->getLevelByName($nbt["Level"])) === null){
 			$this->setLevel($this->server->getDefaultLevel());
-			$nbt["Level"] = $this->level->getName();
+			$nbt["Level"] = $this->level->getSavedName();
 			$nbt["Pos"][0] = $this->level->getSpawnLocation()->x;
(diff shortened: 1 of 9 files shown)


dktapps:
I don't think `getSavedName()` is much better to be truthful, and definitely don't think it warrants such changes. Maybe a doc comment to state that the method `Returns the display name of the level.`

SOF3:
@dktapps I don't think a doc comment would be enough, as many people read the method names assuming that's what they want, without reading the documentation in detail unless they have confusion about it (and usually people won't expect there to be a "folder name" and a "display/saved name"). Directly `getName()` often refers to a unique string identifier, and this is especially true when there is another method that does so. In my opinion, `getSavedName()` implies that there is another name getter, which is more self-explanatory.

dktapps:
"Saved name" also leaves room for confusion on that score in my opinion. "Display name" I think would be better.

SOF3:
@robske110 exactly. This PR aims to prevent such confusion as well as to remind developers about this difference, rather than asserting they are the same or that it is the value used in getLevelByName.

dktapps:
I agree with @robske110 on this one. Quotes from team chat:

> renaming it to getSavedName() is barely any clearer
> I think doc comments to indicate that the method does not return a unique name, and to point out that to get a unique identifier use the folder name instead

PR #151, opened 2016-12-06
Added BatchBlockUpdateEvent and made updateAround() use it

dktapps:
This pull request adds batch block update events.

### Why?
This is cleaner than the original method, and also better for performance since only one event is fired instead of six. The performance is the main reason to do this.

### Backwards incompatible changes
Plugins which use BlockUpdateEvent fired from updateAround() will no longer detect the event since a new one is used.

### Possible improvements
Adapt BlockEvent to allow it to handle block lists, instead of a separate event tree.

### DIFF
--- a/src/pocketmine/event/block/BatchBlockEvent.php
+++ b/src/pocketmine/event/block/BatchBlockEvent.php
@@ -1,0 +1,63 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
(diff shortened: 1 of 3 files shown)


src/pocketmine/event/block/BatchBlockEvent.php
```diff
@@ -0,0 +1,63 @@
...
+	 */
+	public function setBlockList(array $blocks){
+		$this->blocks = $blocks;
+	}
+
+	/**
+	 * @return null
+	 */
+	public function getBlock(){
+		return null;
```
SOF3: I think this is bad. It shouldn't extend BlockEvent.

SOF3 (changes requested):
Also, would even be better if the whole method is refactored such that the event knows the reason why it eas updated.

PR #154, opened 2016-12-07
Fix trailing slashes in MOTD breaks server list, close #153

### LINKED ISSUE #153: Trailing backslash in MOTD breaks server list display
### Issue description
PocketMine/PocketMine-MP#4083

It appears that adding a trailing slash to the server name escapes the `;` which is used as the separator in queries. This produces some amusing effects as seen in the issue above.

1. Set your server MOTD to something with a \ at the end
2. Open MCPE and wait for it to ping the server.

### OS and versions

* PocketMine-MP: 5d16ecc0034ed2262ec2f95ccc0e7bbca17f24d5 
* PHP: 7.0.13
* Server OS: Windows 10 Pro
* Game version: PE/Win10

dktapps:
Cherry-picked from @PrimusLV 's PR to the old repository (PocketMine/PocketMine-MP#4131)

I have tested this and confirmed it works.

### Limitations and possible alternative solution
This removes all trailing slashes from the end of the server name, which might break some formatting. An alternative, slightly hackier way to do this was proposed [here](https://github.com/PocketMine/PocketMine-MP/issues/4083#issuecomment-200920496). This does in fact work well.

### DIFF
--- a/src/pocketmine/network/RakLibInterface.php
+++ b/src/pocketmine/network/RakLibInterface.php
@@ -170,7 +170,7 @@ class RakLibInterface implements ServerInstance, AdvancedSourceInterface{
 		$info = $this->server->getQueryInformation();
 
 		$this->interface->sendOption("name",
-			"MCPE;" . addcslashes($name, ";") . ";" .
+			"MCPE;" . rtrim(addcslashes($name, ";"), '\\') . ";" .
 			Info::CURRENT_PROTOCOL . ";" .
 			\pocketmine\MINECRAFT_VERSION_NETWORK . ";" .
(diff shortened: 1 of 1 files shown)


dktapps:
@robske110 because let's say you have an MOTD that looks like `\\\PocketMineServer\\\`, using rtrim will remove those trailing slashes and break the appearance. Putting a space before the semicolon is a neater way to do it, but also kinda hacky. What do you think? ¯\_(ツ)_/¯

dktapps:
screenshot_2016-12-20-22-47-19

If you're OCD, you'll see straight away why a space isn't suitable. ¯\_(ツ)_/¯

PR #155, opened 2016-12-08
Added ability to force literal gamemode checks for Player->isSurvival(), Player->isCreative() and Player->isAdventure()

dktapps:
I don't know about anyone else, but not realizing that these methods don't return the literal fact of the gamemode of a player can be a giant pain in the butt for developers, who then end up having to implement their own methods to perform literal checks or perform manual `if($player->getGamemode() === blah)`.

This PR adds the ability to request a literal result from `Player->isSurvival()`, `Player->isCreative()` and `Player->isAdventure()`.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -1186,28 +1186,58 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


SOF3 (approved):
Great resolution :+1:. The parameter should work much better than directly replacing the function, which breaks backwards compatibility.

dktapps:
I remember having a use case for this that should be implemented, but I don't recall what it was... ah well.

PR #161, opened 2016-12-11
Deprecate Minecraft game version constants in the `pocketmine` root namespace

dktapps:
Moved to ProtocolInfo in the network module instead.

### Why?
Because I usually auto-generate basic protocol data for updates, and this makes it annoying because I have to manually edit constants in PocketMine.php. Also, if we ever do manage to modularise the network component, the version supported by the server will be decided by the networking component.

### DIFF
--- a/src/pocketmine/PocketMine.php
+++ b/src/pocketmine/PocketMine.php
@@ -65,6 +65,7 @@ namespace {
 }
 
 namespace pocketmine {
+	use pocketmine\network\protocol\Info as ProtocolInfo;
 	use pocketmine\utils\Binary;
 	use pocketmine\utils\MainLogger;
 	use pocketmine\utils\ServerKiller;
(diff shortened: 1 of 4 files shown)


src/pocketmine/PocketMine.php
```diff
@@ -146,6 +145,14 @@ function dummy(){
 	define('pocketmine\DATA', isset($opts["data"]) ? $opts["data"] . DIRECTORY_SEPARATOR : \getcwd() . DIRECTORY_SEPARATOR);
 	define('pocketmine\PLUGIN_PATH', isset($opts["plugins"]) ? $opts["plugins"] . DIRECTORY_SEPARATOR : \getcwd() . DIRECTORY_SEPARATOR . "plugins" . DIRECTORY_SEPARATOR);
 
+	/**
+	 * @deprecated
+	 * These have moved to {@link \pocketmine\network\protocol\Info}. Use those constants instead.
+	 */
+	const MINECRAFT_VERSION = ProtocolInfo::MINECRAFT_VERSION;
+	const MINECRAFT_VERSION_NETWORK = ProtocolInfo::MINECRAFT_VERSION_NETWORK;
```
ishitatsuyuki: You'd better write two comments, if you expect a correct doc.

ishitatsuyuki (changes requested):

SOF3 (approved):

PR #164, opened 2016-12-14
Update IDs for compatibility + 0.16

inxomnyaa:
Adds compatibility with Genisys/ClearSky ids (more plugins supported)
Adds the new ID's for 0.16
Pre-comment the ID's for 0.17
Rename BlockIds::HOPPER to HOPPER_BLOCK for compatibility with ItemIds::HOPPER


src/pocketmine/block/BlockIds.php
```diff
@@ -41,7 +41,7 @@
 	const IRON_ORE = 15;
 	const COAL_ORE = 16;
 	const LOG = 17, WOOD = 17, TRUNK = 17;
-	const LEAVES = 18;
+	const LEAVES = 18, LEAVE = 18;
```
jojoe77777: :facepalm:

JackNoordhuis: :leaf:

src/pocketmine/block/BlockIds.php
```diff
@@ -166,43 +166,44 @@
...
-
-	const LEAVES2 = 161;
+	//const STAINED_GLASS_PANE = 160; //Confirmed
+	const LEAVES2 = 161, LEAVE2 = 161;
 	const WOOD2 = 162, TRUNK2 = 162, LOG2 = 162;
 	const ACACIA_WOOD_STAIRS = 163, ACACIA_WOODEN_STAIRS = 163;
 	const DARK_OAK_WOOD_STAIRS = 164, DARK_OAK_WOODEN_STAIRS = 164;
-	const SLIME_BLOCK = 165;
-
+	const SLIME_BLOCK = 165, const SLIMEBLOCK = 165;
```
JackNoordhuis: Don't need the second constant declaration here

JackNoordhuis (changes requested):

src/pocketmine/block/BlockIds.php
SOF3: 1. This is not the valid singular form of leaves.
2. I doubt anyone would want to use `LEAF` as the block name... I really doubt whether we should put so many different forms, just to make the Item::fromString() method more powerful, while it isn't needed.

src/pocketmine/block/BlockIds.php
```diff
@@ -87,7 +87,7 @@
 	const FURNACE = 61;
 	const BURNING_FURNACE = 62, LIT_FURNACE = 62;
 	const SIGN_POST = 63;
-	const DOOR_BLOCK = 64, WOODEN_DOOR_BLOCK = 64, WOOD_DOOR_BLOCK = 64;
+	const DOOR_BLOCK = 64, WOODEN_DOOR_BLOCK = 64, WOOD_DOOR_BLOCK = 64, OAK_DOOR_BLOCK = 64;
```
SOF3: This implies that this ID only applies for oak doors, or that using this ID always results in an oak door.

src/pocketmine/block/BlockIds.php
```diff
@@ -142,18 +142,18 @@
 	const ENCHANTING_TABLE = 116, ENCHANT_TABLE = 116, ENCHANTMENT_TABLE = 116;
 	const BREWING_STAND_BLOCK = 117;
 	const CAULDRON_BLOCK = 118;
-
-	const END_PORTAL_FRAME = 120, END_PORTAL = 120;
+	//const END_PORTAL = 119; //Confirmed
```
SOF3: Why don't we leave it to be added when it is ready? And what is the purpose of `//Confirmed`?

src/pocketmine/block/BlockIds.php
```diff
@@ -142,18 +142,18 @@
...
+	const END_PORTAL_FRAME = 120;
 	const END_STONE = 121;
-
+	//const DRAGON_EGG = 122; //Confirmed
 	const REDSTONE_LAMP = 123, INACTIVE_REDSTONE_LAMP = 123;
 	const LIT_REDSTONE_LAMP = 124, ACTIVE_REDSTONE_LAMP = 124;
 	const DROPPER = 125;
 	const ACTIVATOR_RAIL = 126;
-	const COCOA_BLOCK = 127, COCOA_PODS = 127;
+	const COCOA_BLOCK = 127, COCOA_PODS = 127, COCOA_BEANS = 127, COCOA_POD = 127;
```
SOF3: Cocoa beans is the item...
And same with LEAVES.

src/pocketmine/block/BlockIds.php
```diff
@@ -166,43 +166,44 @@
 	const CARROT_BLOCK = 141;
 	const POTATO_BLOCK = 142;
 	const WOODEN_BUTTON = 143;
-	const MOB_HEAD_BLOCK = 144, SKULL_BLOCK = 144;
-	const ANVIL = 145;
+	const HEAD_BLOCK = 144, MOB_HEAD_BLOCK = 144, SKULL_BLOCK = 144;
+	const ANVIL_BLOCK = 145, ANVIL = 145;
```
SOF3: No need to specify `_BLOCK` if there doesn't exist a respective item for it.

src/pocketmine/block/BlockIds.php
```diff
@@ -166,43 +166,44 @@
...
 	const TRAPPED_CHEST = 146;
 	const WEIGHTED_PRESSURE_PLATE_LIGHT = 147, LIGHT_WEIGHTED_PRESSURE_PLATE = 147, GOLD_PRESSURE_PLATE = 147;
 	const WEIGHTED_PRESSURE_PLATE_HEAVY = 148, HEAVY_WEIGHTED_PRESSURE_PLATE = 148, IRON_PRESSURE_PLATE = 148;
-	const COMPARATOR_BLOCK = 149, UNPOWERED_COMPARATOR_BLOCK = 149;
-	const POWERED_COMPARATOR_BLOCK = 150;
-	const DAYLIGHT_SENSOR = 151;
+	const COMPARATOR_BLOCK = 149, UNPOWERED_COMPARATOR_BLOCK = 149, UNPOWERED_COMPARATOR = 149;
+
```
SOF3: why this empty line?

src/pocketmine/block/BlockIds.php
```diff
@@ -166,43 +166,44 @@
...
 	const HOPPER_BLOCK = 154;
 	const QUARTZ_BLOCK = 155;
 	const QUARTZ_STAIRS = 156;
 	const DOUBLE_WOOD_SLAB = 157, DOUBLE_WOODEN_SLAB = 157, DOUBLE_WOOD_SLABS = 157, DOUBLE_WOODEN_SLABS = 157;
 	const WOOD_SLAB = 158, WOODEN_SLAB = 158, WOOD_SLABS = 158, WOODEN_SLABS = 158;
 	const STAINED_CLAY = 159, STAINED_HARDENED_CLAY = 159;
-
-	const LEAVES2 = 161;
+	//const STAINED_GLASS_PANE = 160; //Confirmed
+	const LEAVES2 = 161, LEAVE2 = 161;
```
SOF3: Same with LEAVE.

SOF3: Same problem with `LEAVE`.

src/pocketmine/block/BlockIds.php
```diff
@@ -166,43 +166,44 @@
...
 	const PACKED_ICE = 174;
 	const DOUBLE_PLANT = 175;
 
-	const INVERTED_DAYLIGHT_SENSOR = 178, DAYLIGHT_SENSOR_INVERTED = 178;
+	const INVERTED_DAYLIGHT_SENSOR = 178, DAYLIGHT_SENSOR_INVERTED = 178, DAYLIGHT_DETECTOR_INVERTED = 178;
 	const RED_SANDSTONE = 179;
 	const RED_SANDSTONE_STAIRS = 180;
-	const DOUBLE_RED_SANDSTONE_SLAB = 181;
-	const RED_SANDSTONE_SLAB = 182;
+	const DOUBLE_RED_SANDSTONE_SLAB = 181, DOUBLE_STONE_SLAB2 = 181;
```
SOF3: Why add a `STONE_SLAB_2` to replace `RED_SANDSTONE_SLAB`?? This is inaccurate and bad when we have a better way of naming it...

src/pocketmine/block/BlockIds.php
```diff
@@ -216,15 +217,23 @@
...
 	const STONECUTTER = 245;
 	const GLOWING_OBSIDIAN = 246;
 	const NETHER_REACTOR = 247;
 	const UPDATE_BLOCK = 248;
 	const ATEUPD_BLOCK = 249;
+	const PISTON_EXTENSION = 250;
 	const BLOCK_MOVED_BY_PISTON = 250;
 	const OBSERVER = 251;
-	const INFO_RESERVED6 = 255;
+	const INFO_RESERVED6 = 255, RESERVED = 255;
```
SOF3: What is the purpose of this?

src/pocketmine/item/ItemIds.php
```diff
@@ -111,7 +112,7 @@
 	const PAPER = 339;
 	const BOOK = 340;
 	const SLIMEBALL = 341;
-	const MINECART_WITH_CHEST = 342;
+	const MINECART_WITH_CHEST = 342, CHEST_MINECART = 342;
```
SOF3: Is it really meaningful to add this alias?

src/pocketmine/item/ItemIds.php
```diff
@@ -150,10 +151,10 @@
 	const MAGMA_CREAM = 378;
 	const BREWING_STAND = 379;
 	const CAULDRON = 380;
-
+	//const EYE_OF_ENDER = 380; //Confirmed
 	const GLISTERING_MELON = 382;
 	const SPAWN_EGG = 383;
-	const BOTTLE_O_ENCHANTING = 384, ENCHANTING_BOTTLE = 384;
+	const BOTTLE_O_ENCHANTING = 384, ENCHANTING_BOTTLE = 384, EXP_BOTTLE = 384;
```
SOF3: More accurately it should be `EXPERIENCE_BOTTLE` not `EXP_BOTTLE`.

SOF3 (changes requested):
Thanks for your contribution, but we're going to close this pull request because of the various issues mentioned in the inline commit comments, as well as possible conflict with our mcpe-1.0 branch.

You might be thinking that this is going to support /give etc. commands better, but the fact is, with upcoming command API updates, this method no longer needs to support different names due to limitations in the client.

dktapps:
> Rename BlockIds::HOPPER to HOPPER_BLOCK for compatibility with ItemIds::HOPPER

We discussed things like this internally, and I don't believe this is necessary. Since nothing ever accesses these constants directly via the interfaces, but instead by way of `Block::WHATEVER` and `Item::WHATEVER`, it is perfectly valid to have `Block::HOPPER` refer to the block and `Item::HOPPER` refer to a different ID for the Item, since in vanilla gameplay you will never acquire a Hopper Block item. If you specifically want the Hopper Block, you can explicitly ask for `Block::HOPPER` instead of `Item::HOPPER_BLOCK`.

thebigsmileXD:
Overall question: Did you even read the title? 'Compatibility'

I won't create any PR once again

SOF3:
I don't think compatibility is needed. If we want compatibility, we have to add all the terribly unorganized methods present in those spoons, which is totally unreasonable -- this is the official PocketMine repository, and we have no reason to conform to spoons' standards.

src/pocketmine/block/BlockIds.php
robske110: Leave!

Monkeyboy2805: lol

src/pocketmine/block/BlockIds.php
Monkeyboy2805: 👊

Issue #22, opened 2016-10-13
Vertical movement more than 10 blocks triggers movement anti-cheat


DarkWav:
### Issue description

When a player climbs a high tower with ladders, its being dragged back. This also happens when anti flight is disabled, and in creative mode.

### Steps to reproduce the issue

1. Build a really high tower with ladders (at least 30 blocks (for safety that it happens))
2. Climb the ladder. you´ll be dragged back on the ground (note that it will not instantly happen, you´ll have to climb some blocks.

### OS and versions

- PocketMine-MP: github master (commit 2fba1073ddf4965eb5d2bbbbbfb4a5c5f9377db3)
- PHP: [7.0.11](https://github.com/DarkWav/PHP7-Windows-PocketMine/raw/master/PHP_7.0.11-PRO_x86_Windows.zip)
- OS: Windows 10 Pro
### Crashdump, backtrace or other files

none

RocoWayne:
im having the same problem!

SOF3:
> I am pretty sure its a bug in pocketmine because it will not happen on genisys. 

What's wrong even if this happens in genisys?

dktapps:
This is not a Genisys discussion forum. Please refrain from turning it into one.

dktapps:
This would appear to be a movement-check bug, but since I can't reproduce it, it's hard to say.

DarkWav:
@SOF3, im sorry for that "unrelatedness" but i thought it may help @dktapps .

DarkWav:
@dktapps maybe you can replace pocketmine´ s movement ckeck with genisys´ s movement check to fix the problem.

dktapps:
\>_> This project does not exist for the purposes of copy-pasting. How many times do I have to say it!

legoboy0215:
@DarkWav This is not some plugin project you own. You can't just copy paste code from other _**unrelated**_ projects...

DarkWav:
no just a suggestion you dont have to do it :D
@legoboy0215 WHen did i say i own this ?????????

TheUSBFlash:
Reason: Player's bounding box doesn't collide with Ladder's bounding box server side. Due to being in a free space/air part of the ladder, the server thinks you are flying.
If you make the AxisBB of the ladder bigger, it all works fine.
https://github.com/ClearSkyTeam/ClearSky/commit/c103fcb1ecf0a15492612d4e684e67a7912cc190

dktapps:
[...]

PR #168, opened 2016-12-15
Increase ladder bounding box to 3/16, fix kicked for flying when sneaking on ladders

dktapps:
Possible fix for #22 - I cannot reproduce the issue since making this change, although since I don't know what exactly caused the issue to begin with I can't say for sure.

This is in accordance with the (Ladder wiki page)[http://minecraft.gamepedia.com/Ladder#History], which states that (for PC at least) the ladder bounding box is 3/16 of a block wide. It appears to be the same in PE also.


dktapps:
This has been checked against vanilla Minecraft data and is confirmed to be correct.

PR #171, opened 2016-12-17
Re-implemented CommandReader, fixed CommandReader hanging on shutdown, close #25

### LINKED ISSUE #25: Shutdown hangs at "Stopping CommandReader thread" on some systems
### Issue description

On Windows, shutdown will hang at "Stopping CommandReader thread" until another input is received, such as hitting the Enter key.
### Why this happens

`stream_set_blocking()` does not work on Windows. This results in the `fgets()` call to stdin to block the thread until another input is received. This issue does not occur on Linux as far as I can tell.
### Possible solutions

This is a rather difficult problem to resolve cleanly. Since `Thread::kill()` was removed from pthreads in v3, the thread cannot be forcefully killed.
I have made a previous attempt to resolve this with an implementation using `stream_select` which _also_ fails on Windows (returns 1 when no streams have changed >:-[)

1. Shut down a PocketMine server on Windows using /stop
### OS and versions

- PocketMine-MP: https://github.com/pmmp/PocketMine-MP/commit/2fba1073ddf4965eb5d2bbbbbfb4a5c5f9377db3
- PHP: 7.0+
- OS: Any Windows version
### References

https://github.com/PocketMine/PocketMine-MP/issues/3990
https://bugs.php.net/bug.php?id=34972

dktapps:
Re-implemented the Command Reader to fix the issue described in #25.

This pull changes the behaviour of the Command Reader stdin stream to be blocking by default, which is better for performance than the original method of querying the stream every 0.01 seconds. It now uses `stream_select` to poll the status of the stream to check if there have been any changes before entering a blocking stream read. This works flawlessly on most regular systems, including in Powershell and CMD on Windows.

This PR also adds auto-detection for fifo piped input. This was necessary because most people use MinTTY as a console for Windows servers, which is essentially a fifo pipe, not an actual console. `stream_select` always returns 1 for fifo pipes, which causes Windows servers under MinTTY to end up in a blocked state anyway, necessitating a force-kill on shutdown _only_ under the circumstances that the fifo pipe is blocking and does not terminate.

Under MinTTY, the thread will throw an exception to indicate that it is blocked. This will cause the server to skip trying to stop that thread and stop other threads safely, before performing process suicide to take the CR thread with it. Without debug, this will merely show an `Exit 1` in the MinTTY console; if debug is enabled the user will see a debug message to say that the CR thread was blocked, and that the server had to be force-killed to get rid of it. The Exit 1 can also be removed to have the console close automatically by removing the `-h error` option from the MinTTY command line in start.cmd.

### DIFF
--- a/src/pocketmine/PocketMine.php
+++ b/src/pocketmine/PocketMine.php
@@ -2,11 +2,11 @@
 
 /*
  *
- *  ____            _        _   __  __ _                  __  __ ____  
- * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \ 
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
  * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
- * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/ 
- * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_| 
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
@@ -15,7 +15,7 @@
  *
  * @author PocketMine Team
  * @link http://www.pocketmine.net/
- * 
+ *
  *
 */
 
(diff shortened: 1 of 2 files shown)


src/pocketmine/PocketMine.php
```diff
@@ -487,16 +487,31 @@ function cleanPath($path){
...
 	$killer->start();
 
+	$erroredThreads = 0;
+
 	foreach(ThreadManager::getInstance()->getAll() as $id => $thread){
 		$logger->debug("Stopping " . $thread->getThreadName() . " thread");
-		$thread->quit();
+		try{
+			$thread->quit();
+			$logger->debug($thread->getThreadName() . " thread stopped successfully.");
```
robske110: Is $thread->quit() blocking? I don't think you can assume it has stopped successfully.

src/pocketmine/command/CommandReader.php
```diff
@@ -19,58 +19,138 @@
...
+	private $type = self::TYPE_STREAM;
+	private $lastLine = -1;
 
 	public function __construct(){
 		$this->buffer = new \Threaded;
 		$opts = getopt("", ["disable-readline"]);
-		$this->readline = (extension_loaded("readline") and !isset($opts["disable-readline"]));
+
+		if((extension_loaded("readline") and !isset($opts["disable-readline"]) and !$this->isPipe(STDIN))){
+			$this->type = self::TYPE_READLINE;
```
robske110: it would be nice to output a debug message about what type is being used. that would help in many cases.

src/pocketmine/command/CommandReader.php
```diff
@@ -19,58 +19,138 @@
...
+					}
+
+					if(($raw = fgets($stdin)) !== false){
+						$line = trim($raw);
+					}else{
+						return false; //user pressed ctrl+c?
+					}
+
+					break;
+				case self::TYPE_PIPED:
```
robske110: finally :)

src/pocketmine/PocketMine.php
dktapps: :facepalm:
this is _why_ the CommandReader thread causes a hang

src/pocketmine/command/CommandReader.php
robske110: @dktapps

robske110: No, getting the ThreadedLogger in here is too complicated and unnecessary

robske110:
again a bump. read above. @SOF3 \@everyone

dktapps:
I'm going to merge this, since the Enter on shutdown is getting irritating and nobody has reported any issues with it. I'm sure if there are problems they will be reported.

PR #177, opened 2016-12-19
Added support for API alpha/beta/whatever suffixes

dktapps:
For major API changes, we need a better way to handle things than to just bump API to blah.x.x and then let users wait for however many months before a release rolls around with a finalised version of that version of the API.

This pull request adds support for API version suffixes, allowing plugin developers to design plugins for alpha/beta/rc API versions while they are still in development, while removing the necessity to bump the major API every time a major change happens during development.

This works pretty much the same as the old system does, except now the server does an _extra_ check for an API version suffix. Due to the potentially backwards-incompatible changes during development, zero tolerance is given for wrong versions. 

The format of the API versions will now look like this: `3.0.0` for release, and `3.0.0-alpha1` for indev. A mismatch, for example `3.0.0-alpha1 and 3.0.0-alpha2` will cause the server not to load the plugin.


dktapps:
Found a bug, don't merge this yet

PR #183, opened 2016-12-20
mcpe 1.0 chunkskylight

robske110:
This re-adds chunk skylight population support into mcpe-1.0

### DIFF
--- a/src/pocketmine/level/format/generic/GenericChunk.php
+++ b/src/pocketmine/level/format/generic/GenericChunk.php
@@ -119,7 +119,7 @@ class GenericChunk implements Chunk{
 			$this->heightMap = $heightMap;
 		}else{
 			assert(count($heightMap) === 0, "Wrong HeightMap value count, expected 256, got " . count($heightMap));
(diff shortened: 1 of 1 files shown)


robske110:
It would be nice if multiple people would test this. I am not sure if this works.

dktapps:
This is mostly working fine now, apart from some lighting issues under trees that I can't make sense of.

src/pocketmine/level/format/generic/GenericChunk.php
```diff
@@ -283,14 +284,14 @@ public function populateSkyLight(){
 				}
 
 				for(; $y > 0; --$y){
-					if(Block::$solid[$this->getBlockId($x, $y, $z)]){
+					if($this->getBlockId($x, $y, $z) !== Block::AIR){
```
jasonwynn10: what about leaves? they are transparent, so light should pass through them

robske110: No, they diffuse skylight. This was and will be a very basic population.

PR #186, opened 2016-12-21
Some PHP7 updates

xpyctum:


dktapps:
PHP 7.1 isn't supported yet due to pthreads problems, I think you might want to remove the `void` return types.

xpyctum:
``` TypeError: "Return value of pocketmine\Server::registerEntities() must be an instance of pocketmine\void, none returned" (EXCEPTION) in "/src/pocketmine/Server" at line 2442```
Okay.

dktapps:
Hmm, I only just noticed you've made these changes on the dimensions branch... this needs to be on a separate branch so it can be tested separately from other changes, but changing the base will mess up the merge and I don't think it's worth it for a small pull request.

I'm going to close this PR, but thank you anyway for your contribution.

xpyctum:
You are welcome ;)

PR #189, opened 2016-12-21
PHP 7 updates. 

xpyctum:
- [x] Block part
- [x] Items part
- [x] Tile part
- [ ] Math part

### DIFF
--- a/src/pocketmine/item/Item.php
+++ b/src/pocketmine/item/Item.php
@@ -756,7 +756,7 @@ class Item implements ItemIds, \JsonSerializable{
 	public final function deepEquals(Item $item, bool $checkDamage = true, bool $checkCompound = true) : bool{
 		if($this->equals($item, $checkDamage, $checkCompound)){
 			return true;
-		}elseif($item->hasCompoundTag() or $this->hasCompoundTag()){
+		}elseif($item->hasCompoundTag() and $this->hasCompoundTag()){
 			return NBT::matchTree($this->getNamedTag(), $item->getNamedTag());
 		}
 


src/pocketmine/block/Block.php
```diff
@@ -641,7 +641,7 @@ public function getBoundingBox(){
 	/**
 	 * @return AxisAlignedBB
 	 */
-	protected function recalculateBoundingBox(){
+	protected function recalculateBoundingBox() : AxisAlignedBB{
```
dktapps: This is a bad idea, some blocks such as Flowable return null on this.

dktapps (dismissed):

src/pocketmine/block/Block.php
xpyctum: If it was be php 7.1 we can do that
```php
protected function recalculateBoundingBox() : ?AxisAlignedBB{
```
for fix this issue.
What do you think?

xpyctum: Okay. I will remove this, but don't forget that after update to 7.1 :)

badp1g:
How to walk? Oleg please help me

src/pocketmine/block/Block.php
SOF3: In my opinion, it is better to start the PHP 7 updates after 7.1 so that we can add everything together, reducing the risk of inconsistency (even if we don't really care about the consistency between different non-master commits).

dktapps:
In response to @SOF3 's comment, I'm going to leave this open for now.

xpyctum:
So, we can add more files for php7?

robske110:
?!

src/pocketmine/tile/Container.php
```diff
@@ -41,5 +41,5 @@ public function setItem($index, Item $item);
 	/**
 	 * @return int
 	 */
-	public function getSize();
+	public function getSize() : Int;
```
dktapps: I think you meant `int` not `Int` ?

dktapps:
Please also enable strict types on modified files where possible provided it is possible to do so.

src/pocketmine/tile/Container.php
xpyctum: @dktapps oops

xpyctum:
if it will give a performance that will do :)

xpyctum:
Too lazy to do it now for the block :<

dktapps:
I rebased the branch you're PRing to, you may need to rebase yours on top of it to keep the merge clean. Sorry about that.

PR #191, opened 2016-12-21
Fixed effect amplifier overflow crash, close #147

### LINKED ISSUE #147: Attributes exceeds range error
### Issue description
After applying a speed effect to a player with an amplifier higher than 127 the server acts fine. When the player whom has the effect leaves and then joins the server again (while retaining the effect) an error will be thrown when the server tries to apply the saved effect.

[View Error](https://gist.github.com/JackNoordhuis/0ecc6333789fc09c2bd7b4b9942d9b16)

1. Apply a speed effect to yourself with a high value (/effect {player} speed 1000 128)
2. Leave the server
3. Join the server
4. Profit!

### OS and versions
* PocketMine-MP: 1.6.1dev 「Unleashed」 implementing API version 2.1.0 for Minecraft: PE v0.16.0.5 alpha (protocol version 91) 
* Commit: https://github.com/pmmp/PocketMine-MP/commit/4674d34469584c60816fedeb292cbbd7376fb019
* PHP: 7
* Server OS: Doesn't matter (MacOS if you really wanna know)

Error: https://gist.github.com/JackNoordhuis/0ecc6333789fc09c2bd7b4b9942d9b16

dktapps:
Because amplifiers are stored as ByteTags and read back from storage as signed, anything over 128 would become a negative amplifier and cause a crash next time the player joined.

## Changes
- /effect command is now capped to adding effects in the range 0-255
- Fixed amplifiers being read as negative.

This will require translation updates to ensure that force-language and the console work correctly.

### DIFF
--- a/src/pocketmine/entity/Effect.php
+++ b/src/pocketmine/entity/Effect.php
@@ -169,8 +169,8 @@ class Effect{
 	 *
 	 * @return $this
 	 */
-	public function setAmplifier($amplifier){
-		$this->amplifier = (int) $amplifier;
+	public function setAmplifier(int $amplifier){
+		$this->amplifier = $amplifier & 0xff;
 		return $this;
 	}
 
--- a/src/pocketmine/command/defaults/EffectCommand.php
+++ b/src/pocketmine/command/defaults/EffectCommand.php
@@ -90,6 +90,13 @@ class EffectCommand extends VanillaCommand{
 
 		if(count($args) >= 4){
 			$amplification = (int) $args[3];
+			if($amplification > 255){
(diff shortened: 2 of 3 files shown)


src/pocketmine/entity/Effect.php
```diff
@@ -169,8 +169,8 @@ public function getAmplifier(){
 	 *
 	 * @return $this
```
SOF3: Not related to this PR, but I wonder if it is really right to put `$this` here...

Muqsit:
Nice!

dktapps:
This isn't yet complete, see #207 . However it's a critical bugfix so I considered it wise to merge it regardless.

PR #202, opened 2016-12-23
PHP7 updates

dktapps:
Blame me for rebasing the branch. My bad.


SOF3:
Would actually be better to squash all :D

NycuRO:
@dktapps You can use: https://github.com/michaelbutler/php5to7

for changes.

PR #206, opened 2016-12-26
Fix duplicated chat messages

robske110:
This fixes duplicated chat messages if a line break (\n) was included.

Although I do not get the reason to send lines separated in the first
place, it somehow messes up the client, and it duplicates them in a
strange pattern. Funny enough, after 2hrs of desperately trying to find the place where the server duplicates the packets, i discovered that locking/suspending my device/minecraft, and unlocking/resuming it will magically make the duplicated messages disappear.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -2996,6 +2996,14 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 			$message = $message->getText();
 		}
 
+		$message = $this->server->getLanguage()->translateString($message);
+
+		$pk = new TextPacket();
+		$pk->type = TextPacket::TYPE_RAW;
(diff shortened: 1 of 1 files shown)


Muqsit:
Tested, NICE!

Leo3418:
If Mojang doesn't manage to fix this bug in future releases, I think adding a workaround in PocketMine should be done.

dktapps:
@Leo3418 They aren't going to leave it unfixed, this bug affects vanilla too. It's been reported in their JIRA issue tracker.

(It was supposedly fixed in 1.0.0.0, but ... lol)

robske110:
Due to this PR mitigating a client bug in some cases, this is not going to be merged. Feel free to use this fix though if you have a lot of newlines everywhere.

PR #217, opened 2016-12-30
Kick on ban.

Dog2puppy:
If you ban a user using a plugin, the user can continue to play even though they're banned. This fixes that by kicking the player. This Pull Request also changes the /ban command so that it no longer handles the kick because it's now handled by the ban function.


jojoe77777:
$target is/(can be?) the player's name.

Dog2puppy:
@jojoe77777 Opps. Should fix that.

jojoe77777:
@Dog2puppy but the player could be offline

Dog2puppy:
@jojoe77777 Fixed.

PR #225, opened 2017-01-02
MobHeads are stackable to 64

Infernus102:


robske110:
lol, you can just submit a PR to merge master into mcpe-1.0 but that's also done internally

Issue #119, opened 2016-11-22
Movement-check bugs


VlaD3n:
### Issue description

Hello issue with PM 0.16.X
I found a mistake when you're trying to climb the rungs writes "* player * moved wrongly !!" the console and starts to shake Player

### Steps to reproduce the issue

1. Trying to get up the stairs
2. If there is a shake, but now look at the console

### OS and versions

* PocketMine-MP: 0.16.0.5 1.6.1dev-#82
* PHP: 7
* Server OS: Debian 7
* Game version: PE

### Crashdump, backtrace or other files

* ...

iTCMega:
its **_work fine_** at the console! and no **shake** ! :100:  :+1:

VlaD3n:
What? My friend tested it on my server is the core and he said that when the stairs climbed, it started to shake! While it may be under the stairs, he meant a ladder ...

legoboy0215:
The "{player} moved wrongly" is back again?

dktapps:
It was never gone, only in Genisys the anti cheat code responsible was simply commented out.

legoboy0215:
I noticed that.

daterrell:
I have also reproduced this, but you _must_ be in gamemode 0 (survival).  The default game mode can be either, but the player climbing the stairs must be in survival.  It also helps to run up the stairs instead of walking.

dktapps:
Also reproduced when walking through doors.

dktapps:
Door bounding boxes are off, by 0.005 of a block (3/15 in vanilla, 3/16 in PocketMine). Mega facedesk Mojang.

dktapps:
https://bugs.mojang.com/browse/MCPE-19214

dktapps:
Relates to #169

dktapps:
Most everything here should been have been fixed in #259, with the exception of stairs (and possibly doors, but that is a Mojang issue). Since stairs are due to wrong bounding boxes server-side, I'm going to close this with reference to #169.

PR #233, opened 2017-01-07
Remove movement checks

dktapps:
Anti-cheat should be done by plugins for those who want it. It is otherwise usually just an irritation.

Removing these checks fixes a bunch of issues, including:

- being TPed up/down ladders (#22)
- flying up/down pull-backs in blocks in spectator mode
- [player] moved wrongly! when moving more than 0.25 blocks in a tick or when server is lagging (#119)

NOTE: This is the REMOVAL of some highly irritating and buggy anti-cheat. 

**If you depend on this anti-cheat, I recommend you wait until there is a plugin substitute available to provide this.**

### Possible improvements
Add CheatEvents which are cancelled by default, and let plugins handle these events.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -1349,20 +1349,17 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 
 		$revert = false;
 
-		if(($distanceSquared / ($tickDiff ** 2)) > 100){
-			$revert = true;
-		}else{
-			if($this->chunk === null or !$this->chunk->isGenerated()){
-				$chunk = $this->level->getChunk($newPos->x >> 4, $newPos->z >> 4, false);
-				if($chunk === null or !$chunk->isGenerated()){
-					$revert = true;
-					$this->nextChunkOrderRun = 0;
-				}else{
-					if($this->chunk !== null){
-						$this->chunk->removeEntity($this);
-					}
-					$this->chunk = $chunk;
+
+		if($this->chunk === null or !$this->chunk->isGenerated()){
+			$chunk = $this->level->getChunk($newPos->x >> 4, $newPos->z >> 4, false);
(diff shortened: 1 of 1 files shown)


Muqsit:
Maybe add functions that will assist in making an anti cheat. For example, the isFlying() and isMoving() you did for the gspoon

xxFlare:
@dktapps 

By removing:
`if(($distanceSquared / ($tickDiff ** 2)) > 100){
$revert = true;	
}`

You would be allowing players with teleportation hacks to teleport around the world, without PocketMine blocking it. Also, it's mainly the anti no-clip (invalid movement) check that causes irritations.

dktapps:
> Anti-cheat should be done by plugins for those who want it.

xxFlare:
I disagree. Advance anti-cheats should be done by plugins.

Here is what shoghicp said a while back:
https://github.com/TruDan/NoCheatPE/issues/3

dktapps:
and if I _want_ to be able to perform these kind of cheats on my server? if I want to be able to perform normal gameplay without shitty anti-cheat bugging out everything I do? MCPE doesn't perform any such checks, so there is no reason for the core to do it, it simply reduces gameplay experience. Let plugins do the bad implementations.

legoboy0215:
You might want to merge this after an official/unofficial anti-cheat plugin comes out. I wouldn't want to use PM if basic cheating cannot be prevented.

dktapps:
Going to hold this for the time being, since the issue with anti-noclip indicates a more deep rooted issue with bounding boxes. A pull-back issue occurs when flying or running very close to walls.

dktapps:
I'm going to close this, since it just removes the symptoms of the bug rather than the root cause. I think the underlying issues will manifest itself in other, more subtle forms if not found and fixed. If there is one good thing about anti-cheat like this, it is that the annoyances it causes are very obvious and easily reproduced, so this can be used as a test case against future bug fixes.

PR #234, opened 2017-01-07
Fixes sendHeldItem to stop crashes when executing.

Sandertv:
This pull request fixes issues with plugins such as Slapper. Human entities would crash when getting near entities, due to $entity->getServer() not being a defined method. Instead I've replaced it with getHolder and getLevel in order to be able to get the server.

### DIFF
--- a/src/pocketmine/inventory/PlayerInventory.php
+++ b/src/pocketmine/inventory/PlayerInventory.php
@@ -137,7 +137,7 @@ class PlayerInventory extends BaseInventory{
 				$this->sendSlot($this->getHeldItemSlot(), $target);
 			}
 		}else{
-			$this->holder->getServer()->broadcastPacket($target, $pk);
+			$this->getHolder()->getLevel()->getServer()->broadcastPacket($target, $pk);
 			foreach($target as $player){
 				if($player === $this->getHolder()){
 					$this->sendSlot($this->getHeldItemSlot(), $player);
--- a/src/pocketmine/item/Food.php
+++ b/src/pocketmine/item/Food.php
@@ -57,7 +57,7 @@ abstract class Food extends Item implements FoodSource{
 		if($human instanceof Player){
 			$human->dataPacket($pk);
 		}
-		$human->getServer()->broadcastPacket($human->getViewers(), $pk);
+		$human->getLevel()->getServer()->broadcastPacket($human->getViewers(), $pk);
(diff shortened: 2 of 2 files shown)


dktapps:
looks defined to me... https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/inventory/BaseInventory.php#L50

Sandertv:
Hmmm i must have been blind. I don't think getting the server directly from the holder always works however, in this case for example with a Human slapper.

jojoe77777:
And since no-one has reported this "crash"...

dktapps:
well, the real question in that case is why not, isn't it?

dktapps:
If you have a crash issue, I would like to see an error log, and a crashdump if possible.

Sandertv:
Here's the crashdump:
```
PocketMine-MP Crash Dump Sat Jan 7 06:58:05 EST 2017

Error: Call to undefined method slapper\entities\SlapperHuman::getServer()
File: /src/pocketmine/inventory/PlayerInventory
Line: 140
Type: notice

Code:
[131] 		$pk->slot = $this->getHeldItemSlot();
[132] 		$pk->selectedSlot = $this->getHeldItemIndex();
[133] 
[134] 		if(!is_array($target)){
[135] 			$target->dataPacket($pk);
[136] 			if($target === $this->getHolder()){
[137] 				$this->sendSlot($this->getHeldItemSlot(), $target);
[138] 			}
[139] 		}else{
[140] 			$this->holder->getServer()->broadcastPacket($target, $pk);
[141] 			foreach($target as $player){
[142] 				if($player === $this->getHolder()){
[143] 					$this->sendSlot($this->getHeldItemSlot(), $player);
[144] 					break;
[145] 				}
[146] 			}
[147] 		}
[148] 	}
[149] 
[150] 	public function onSlotChange($index, $before){
```

dktapps:
you might also want to change https://github.com/pmmp/PocketMine-MP/commit/0bd7ea211d7b3b23d437ad985cf0f29a28bfb333#diff-08f73f87f3a8a8d72fe4081b908a68bbR61 to get the reference via the level also.

Sandertv:
I'll do that.

dktapps:
Looks good to me, OK to merge?

Sandertv:
I think so!

Issue #156, opened 2016-12-08
Crash issues [Chunk Error]


Misteboss:
### Issue description

Very often after a reset of a map, the server crash with this error, i checked and the maps is loaded, i don't know why :|. Some friends have the same problem.
### Steps to reproduce the issue

1. ... Reset a map (unload and load it)
2. ... join the world, the server will crash 1/6 with this error

### OS and versions

* PocketMine-MP: 5d16ecc
* PHP: 7.0.2
* Server OS: Debian
* Game version: PE/Win10 

### Crashdump, backtrace or other files

* ... Console : 016-12-08 Unloading level "BuildUHC1vs1(2)"
 Preparing level "BuildUHC1vs1(2)"
 Unloading level "BuildUHC1vs1(2)"
reparing level "BuildUHC1vs1(2)"
 Error: "Call to a member function getChunk() on null" (EXCEPTION) in "/src/pocketmine/level/Level" at line 2462
An unrecoverable error has occurred and the server has crashed. Creating a crash dump
An unrecoverable error has occurred and the server has crashed. Creating a crash dump

Misteboss:
It concerns that line 
-> 		$chunk = $this->provider->getChunk($x, $z, $generate);

Misteboss:
Oh and it's caused by .mcr worlds i think cause when the server crash it's caused by a map in .mcr world format and never by .mca worlds, is it possible to convert the worlds or to fix that ?

dktapps:
Please provide a proper crashdump.

Misteboss:
Ok i  going to send you one now ;)

Misteboss:
This is the crashdumps :
Crash Dump Thu Dec 8 09:05:05 EST 2016

Error: Call to a member function startTiming() on null
/src/pocketmine/level/format/mcregion/McRegion
Line: 209
Type: notice

Code:
[200] 
[201] 	public function loadChunk($chunkX, $chunkZ, $create = false){
[202] 		$index = Level::chunkHash($chunkX, $chunkZ);
[203] 		if(isset($this->chunks[$index])){
[204] 			return true;
[205] 		}
[206] 		$regionX = $regionZ = null;
[207] 		self::getRegionIndex($chunkX, $chunkZ, $regionX, $regionZ);
[208] 		$this->loadRegion($regionX, $regionZ);
[209] 		$this->level->timings->syncChunkLoadDataTimer->startTiming();
[...]

PR #241, opened 2017-01-08
Better guards against Positions leaking unloaded Level objects

dktapps:
Should highlight the real cause of #156 .

### Changes
- Added `Level->isClosed()` method and related properties to allow detecting if a level has been unloaded
- `Position->getLevel()` now checks if its target Level has been unloaded, and if so destroys the reference before returning anything.
- `Position->setLevel()` will now check if the supplied Level object is usable before setting it, and throw an exception if the Level is closed.

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -260,6 +260,8 @@ class Level implements ChunkManager, Metadatable{
 	/** @var Generator */
 	private $generatorInstance;
 
+	private $closed = false;
+
 	public static function chunkHash(int $x, int $z){
 		return PHP_INT_SIZE === 8 ? (($x & 0xFFFFFFFF) << 32) | ($z & 0xFFFFFFFF) : $x . ":" . $z;
 	}
@@ -404,7 +406,12 @@ class Level implements ChunkManager, Metadatable{
 		return $this->levelId;
 	}
 
+	public function isClosed() : bool{
+		return $this->closed;
+	}
+
 	public function close(){
+		assert(!$this->closed, "Tried to close a level which is already closed");
 
 		if($this->getAutoSave()){
 			$this->save();
@@ -421,6 +428,8 @@ class Level implements ChunkManager, Metadatable{
 		$this->blockMetadata = null;
 		$this->blockCache = [];
 		$this->temporalPosition = null;
+
+		$this->closed = true;
 	}
 
 	public function addSound(Sound $sound, array $players = null){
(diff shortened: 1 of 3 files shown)


src/pocketmine/level/Level.php
```diff
@@ -404,7 +406,12 @@ public function getServer() : Server{
 		return $this->levelId;
 	}
 
+	public function isClosed() : bool{
```
robske110: finally :+1:

src/pocketmine/level/Position.php
```diff
@@ -47,19 +48,40 @@ public static function fromObject(Vector3 $pos, Level $level = null){
...
+	 *
+	 * @throws \InvalidArgumentException if the specified Level has been closed
+	 */
+	public function setLevel(Level $level = null){
+		if($level !== null and $level->isClosed()){
+			throw new \InvalidArgumentException("Specified level has been unloaded and cannot be used");
+		}
+
 		$this->level = $level;
 		return $this;
```
PrimusLV: Why fluent?

src/pocketmine/level/WeakPosition.php
```diff
@@ -53,7 +53,18 @@ public function getLevel(){
...
+	 *
+	 * @throws \InvalidArgumentException if the specified Level has been closed
+	 */
+	public function setLevel(Level $level = null){
+		if($level !== null and $level->isClosed()){
+			throw new \InvalidArgumentException("Specified level has been unloaded and cannot be used");
+		}
+
 		$this->levelId = ($level !== null ? $level->getId() : -1);
 		return $this;
```
PrimusLV: And fluent here?

src/pocketmine/level/Position.php
PrimusLV: Why fluent?

src/pocketmine/level/WeakPosition.php
dktapps: Not my doing. I left it alone since Position is very widely used and I don't fancy digging out the usages of this return value.

src/pocketmine/level/Position.php
robske110: why not?

src/pocketmine/level/WeakPosition.php
SOF3: Actually I don't see a reason not to be fluent as long as you remember it. It is really annoying when you have a lot to modify at a time.

PR #249, opened 2017-01-11
Fixed duplicate SetTimePacket(), and fixed useless overwrite of $pos.

xxFlare:
Some code clean up.

Removes the overwrite of $pos and fixes SetTimePacket sending twice (It is already sent when the packet is first received).


xxFlare:
@jojoe77777 Forgot about that, alright.

Issue #262, opened 2017-01-15
Chunk sending improvements


dktapps:
### Issue description
- Chunks are sent in a square spiral. They should be sent in a circle, or diamond. At full render this will cause about 20% more chunks to be sent than needed, due to the fact that you can't see the corners anyway.
- The radius calculation is off. Default `max-chunks` around a player is 192. This is actually just under 7 chunks render-distance. Considering a lot of devices these days have them cranked up to 14, and some like W10 can go right up to 56, this is a little off. It results in clearly-visible edges to the world, while the terrain should fade into the horizon. For 14 render-distance the server should actually be sending ~615 chunks in a _circle_ around the player.
- RequestChunkRadius/ChunkRadiusUpdated are not used correctly: `max-chunks` is totally ignored when a player changes their render distance. A W10 player for example, changing radius to 56, could cause the server to load nearly 13,000 chunks and cause a crash.

dktapps:
View distance is also completely wrong: in PocketMine it's actually treated as the total number of chunks. It should be the chunk radius.

xxFlare:
@dktapps 

Also some problems with orderChunks():
- Chunks are ordered on schedule rather then logic (Every 200 ticks, or 20 ticks when a player moves. A better way to do it is when a player moves a significant distance)
- Chunks order twice when you join due to the teleport function being called and chunkOrderRun set to 5 by default.

dktapps:
I actually observe chunk ordering is called 3 times on join 🤔

dktapps:
### TODO
- [ ] Send chunks in circles as it is done in vanilla MCPE
- [ ] Fix an issue where the edges of the world becomes visible when a player moves, but not into another chunk.
- [ ] Order chunks ondemand instead of on schedule

xxFlare:
@dktapps Interesting... chunks used to actually send in circles. https://github.com/pmmp/PocketMine-MP/commit/f133154919e8fe340972e4beb12b2ba7b6d478fe

> Fix an issue where the edges of the world becomes visible when a player moves, but not into another chunk. (client-sided rendering issue with adjacent unloaded chunks)

Is this the bug where you load a chunk, moving away until its unloaded, and when you come back it just won't load client side?

dktapps:
no. It's a client-sided rendering issue where chunks are invisible if there's an unloaded chunk next to it. See MiNET wiki.

How it used to be done two years ago is not relevant. What matters is how it works _now_.

[...]

PR #263, opened 2017-01-15
Added a hack for multiworld chunk bug

dktapps:
Sends an empty chunk column when unloading old-world chunks to clear the column client-side. Note this isn't always foolproof, since PocketMine's tracking of chunks loaded by players is bad and inaccurate. This will be improved in the future (#262). It does however work in most cases.


archie426:
Is there not a way to do it without hacks?

dktapps:
@archie426 Multiworld in its entirety is a hack. Best we can do is make the hacks better.

archie426:
@dktapps Yep.

kabluinc:
After this update, moving worlds takes much longer now. (about a 5 second client freeze and the message "moved too fast, reverting movement" in console")

dktapps:
That's a debugging message, we are working on debugging false-positives in the anti-cheat. Just ignore it.

PR #269, opened 2017-01-17
Added CraftingDataPacket cache, reduced lag spikes on player join

### LINKED ISSUE #248: Performance issue with sendRecipeList.
### Issue description
This function is taking too long to execute when players join the game (by default, not from plugin).
https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/Server.php#L2169

You can confirm this with timings.

It seems steadfast fixed the problem with this solution:
https://github.com/Hydreon/Steadfast2/blob/master/src/pocketmine/Server.php#L2337

1. Turn on server.
2. Run /timings on
3. Join and leave multiple times.
4. paste timings

You can then see it takes a significant amount of time to execute.

### OS and versions
* PocketMine-MP: https://github.com/pmmp/PocketMine-MP/commit/57f3f9b4edd117fac180c8c468a73d1b1eb7fc9d
* PHP: 7.0.3
* Server OS: Linux
* Game version: PE

dktapps:
Closes #248 

This will cause the server to build a CraftingDataPacket on startup, which will then be used for all players, instead of creating a new packet for every player joining. This reduces timings of CraftingDataPacket by more than 90% on my machine with one player. With more players the difference will be more significant.

I have not tested the plugin API side of things, but whenever the CraftingManager has a new recipe added, the crafting cache will be destroyed, and rebuilt next time a player joins. This should work fine for plugins, but as I have said I haven't tested.

### Tests
Tested with one player, but no API tests have been performed. If anyone would like to trial-run their plugin which modifies crafting stuff against this, then please do.

### NOTE
For maximum performance, the packet is also immediately _encoded_ when the cache is created. This is because actually _encoding_ the packet takes up nearly half of the time in the timings reports. (13/29ms on my machine). This may affect plugins' capability to modify the packet contents before sending.

### Possible improvements
- Add an option to disable the cache
- Add an option for whether to encode the packet straightaway when creating the cache (better for performance, possibly not so good for plugins)

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -45,8 +45,6 @@ use pocketmine\event\TranslationContainer;
 use pocketmine\inventory\CraftingManager;
 use pocketmine\inventory\InventoryType;
 use pocketmine\inventory\Recipe;
-use pocketmine\inventory\ShapedRecipe;
-use pocketmine\inventory\ShapelessRecipe;
 use pocketmine\item\enchantment\Enchantment;
 use pocketmine\item\Item;
 use pocketmine\lang\BaseLang;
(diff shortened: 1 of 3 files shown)


robske110 (approved):
Let's just hope those three functions are the only ones where the cache needs to be cleared. This should be tested fully through.

markkrueg:
Currently testing in server with approximately 20 players. Will update if any issues are reported.

dktapps:
Going to merge this, I'm sure issue reports will be made if this causes bugs.

PR #274, opened 2017-01-19
Workaround for chat duplication. TODO: Revert this

dktapps:
Obsoletes #206 

This solution is _somewhat_ more effective than the referenced one above. Since multiple TextPackets sent in the space of 1 tick cause dupes of later messages, this _mostly_ alleviates the issue by queuing the messages up and sending them all in one packet on the next available opportunity.

This does not completely resolve the issue though. This only squashes `raw` messages into a single packet, so if, for example, a `translation` and a `raw` are sent in the same tick, like they are in the /help command (header uses translation, entries use raw), then the top messages will still be duplicated.

This does however completely eliminate the /status output issue and mostly eliminates /help issues apart from the one mentioned above.

In the meantime, we wait for Mojang to fix this https://bugs.mojang.com/browse/MCPE-17631. When they do, this change will be reverted.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -2991,15 +2991,10 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 			$message = $message->getText();
 		}
 
-		$mes = explode("\n", $this->server->getLanguage()->translateString($message));
-		foreach($mes as $m){
-			if($m !== ""){
-				$pk = new TextPacket();
-				$pk->type = TextPacket::TYPE_RAW;
-				$pk->message = $m;
-				$this->dataPacket($pk);
-			}
-		}
+		$pk = new TextPacket();
+		$pk->type = TextPacket::TYPE_RAW;
+		$pk->message = $this->server->getLanguage()->translateString($message);
+		$this->dataPacket($pk);
 	}
 
 	public function sendTranslation($message, array $parameters = []){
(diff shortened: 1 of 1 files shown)


dktapps:
@robske110 could you review this?

robske110:
To sum up: I just don't think any hack related to this should be in PM, while i do understand that this is a very annoying issue.

dktapps:
There are precedents for this for similar irritating issues. As long as it's made clear that it is a hack and should be removed later, I think it's OK to proceed, provided it actually does help the scenario. I rejected the original PR due to the fact it only covers part of the problem.

src/pocketmine/Player.php
```diff
@@ -2998,12 +3010,14 @@ public function sendMessage($message){
 
 		$mes = explode("\n", $this->server->getLanguage()->translateString($message));
 		foreach($mes as $m){
-			if($m !== ""){
+			//TODO: Remove this workaround (broken client)
+			$this->messageQueue[] = $m;
```
robske110: Why actually still dividing the newlines?

markkrueg:
This definitely works better than the previous one. Not 100%, but better. 

One strange side effect though. Here is output from a plugin:

before:
before_chat_tweak
after:
after_chat_tweak

```
$sender->sendMessage(TextFormat::GREEN . "FactionsPro Help Page 1 of 3" . TextFormat::RED . "\n/f accept\n/f create <name>\n/f del\n/f demote <player>\n/f deny\n/f home");
```

robske110:
it actually fixes a bug of TextFormats not going through \n's

dktapps:
Damn it Jenkins...

PR #259, opened 2017-01-14
Fix most movement-check false-positives and many position bugs

dktapps:
Fixed myriad movement-checking bugs:

- fixed ladders, close #22 
- fixed auto-jump pull-backs
- fixed pull-backs flying up/down walls
- fixed pull-backs walking _along_ walls
- fixed all known spectator mode pull-back issues (flying up/down pull-backs)
- fixed placing blocks in front of self sometimes vanish
- fixed most known cases of "player moved wrongly!" (does not include stairs, due to stairs bounding boxes actually being wrong (TODO))

A floating-point accuracy has been added to Binary float reading methods. Network Vec3s are now decoded and rounded to an accuracy of 4 decimal places.

Closes #258 , see that for details.

Close #264

PR #290, opened 2017-01-24
Litle optimization

FaigerSYS:
It's better to create cache and compare it with packet buffer (faster)

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -223,6 +223,9 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	/** @var Player[] */
 	protected $hiddenPlayers = [];
 
+	/** @var string */
+	protected $lastPositionCache;
+
 	/** @var Vector3 */
 	protected $newPosition;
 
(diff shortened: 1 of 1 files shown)


robske110:
why do you need an additional cache? the position is already saved?

dktapps:
Only a microscopic amount faster in reality. I doubt such an optimization makes any noticeable difference.

robske110:
Why ` < 0.01` actually? Isn't it just 0 if the player didn't move? or is there some value trimming/rounding somewhere.

dktapps:
#259 floating-point errors when encoded in 4 bytes

robske110:
It just looks so messy/hacky...

ghost (changes requested):
I doubt how well this might improve performance. Any reliable statistics from a production server?

Also, would you confirm that the values will really be identical?

dktapps:
As @poggit-orphanage said.

Muqsit:
...andd Donald Knuth said.

Dog2puppy:
Trump*

PR #297, opened 2017-01-27
Fixed wrong commit parsing on non-default branches

Leo3418:
This pull request closes #295.  

After this change, PocketMine-MP will first scan `.git/refs/heads` folder and store all contents of the folder to `$git_heads` array using `scandir` function.  
If the user is using `master` branch, there should be only 3 items in that array:
- `.`
- `..`
- `master`

However, if the user is not using the default branch, an extra file with the other branch's name will appear. So if the length of the array is more than 3, we can say that the user is not using `master` branch. Then, PocketMine-MP reads that extra file and get correct Git commit hash tag.

#### Testing
To trigger a crash and see Git commit hash, I deleted `Player.php` during development.  
Here is part of the crash dump when `mcpe-1.0` file exists in `.git/refs/heads` folder:
```
PocketMine-MP Crash Dump Fri Jan 27 03:37:25 UTC 2017

Error: Class pocketmine\Player not found
/src/spl/BaseClassLoader
Line: 144
Type: notice

THIS CRASH WAS CAUSED BY A PLUGIN

Code:
[135] 
[136] 			if(method_exists($name, "onClassLoaded") and (new ReflectionClass($name))->getMethod("onClassLoaded")->isStatic()){
[137] 				$name::onClassLoaded();
[138] 			}
[139] 			
[140] 			$this->classes[] = $name;
[141] 
[142] 			return true;
[143] 		}elseif($this->getParent() === null){
[144] 			throw new ClassNotFoundException("Class $name not found");
[145] 		}
[146] 
[147] 		return false;
[148] 	}
[149] 
[150] 	/**
[151] 	 * Returns the path for the class, if any
[152] 	 *
[153] 	 * @param string $name
[154] 	 *

Backtrace:
#0 /src/pocketmine/Server(405): spl_autoload_call(string pocketmine\Player)
#1 /src/pocketmine/Server(1962): pocketmine\Server::getGamemodeString(integer 0)
#2 /src/pocketmine/Server(1587): pocketmine\Server->start(boolean)
#3 /src/pocketmine/PocketMine(500): pocketmine\Server->__construct(pocketmine\CompatibleClassLoader object, pocketmine\utils\MainLogger object, string F:\pmmp\PocketMine-MP\, string F:\pmmp\PocketMine-MP\, string F:\pmmp\PocketMine-MP\plugins\)

PocketMine-MP version: 1.6.2dev #0 [Protocol 100; API 3.0.0-ALPHA3]
[...]

PR #299, opened 2017-01-27
Fix Git commit hash resolving

### LINKED ISSUE #295: Git commit parsing is wrong
### Issue description
https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/PocketMine.php#L473-L477 will produce incorrect hashes if the user is not using the master branch.

SOF3:
This pull request tries to detect the HEAD from `.git/HEAD`, and then attempts to resolve the ref SHA if the HEAD is a ref, resolving #295 and closes #297.

### DIFF
--- a/src/pocketmine/PocketMine.php
+++ b/src/pocketmine/PocketMine.php
@@ -470,9 +470,18 @@ namespace pocketmine {
 		exit(1); //Exit with error
 	}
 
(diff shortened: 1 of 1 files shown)


dktapps:
yes, this is how I intended to do it l. Will test shortly. +1

dktapps (approved):
TODO: make the git hash more useful, but that's a job for later. As a bug fix this is fine. Merging.

Issue #292, opened 2017-01-26
Bad/annoying things happen when the server is lagging heavily due to client/server ticks out of sync


dktapps:
### Issue description
We've all seen it happen where you fly a distance when the server is lagging a lot, and then boom you're dragged back to point A. It's really irritating.

This is because when the server is under heavy load, ticks may _not_ accurately represent 50 milliseconds, while client-side they still do. Let's say you move from 0,60,0 to 50,60,50 while the server is frozen. Because of the overload, those position changes will not be processed until the current heavy-load issue is out of the way. But the server still thinks that took _1 tick_. Result: the server thinks you moved ~70 blocks in a single tick - nope, get your ass back here.

The long and short of it is that server main thread times and ticks cannot be trusted to provide an accurate representation of time difference, and this causes issues when handling clients who aren't lagging at all.

### Steps to reproduce the issue

1. lag your server
2. fly a long distance
3. get yanked back

### OS and versions

* PocketMine-MP: https://github.com/pmmp/PocketMine-MP/commit/740a8ad436272327fded022d6558713ad0566a75
* PHP: 7.0.14
* Server OS: Windows 10
* Game version: not relevant

Muqsit:
"nope, get your ass back here" 😁

dktapps:
Contrary to this issue's original thoughts, this is more often the result of network lag as opposed to the server lagging behind. That's an issue that can't be completely fixed, only worked around (see "blink" hacks on PC servers).

PR #300, opened 2017-01-27
Skip ticks when server is under heavy load

dktapps:
Bump the tick counter when a tick took longer than 50ms to execute. This resolves issues such as #292 where a single tick lags enough for _several_ ticks to pass client-side, resulting in issues such as the server thinking the client has moved too far in a single tick.

### Possible improvements
- Add a server watchdog as described [here](http://minecraft.gamepedia.com/Server.properties#Minecraft_server_properties) to forcefully kill the server if a tick takes longer than a certain number of milliseconds. (is pthreads consistent enough for this to work reliably?)

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -2374,9 +2374,15 @@ class Server{
 
 		Timings::$serverTickTimer->stopTiming();
 
-		$now = microtime(true);
-		$this->currentTPS = min(20, 1 / max(0.001, $now - $tickTime));
-		$this->currentUse = min(1, ($now - $tickTime) / 0.05);
+		$totalTime = round(microtime(true) - $tickTime, 4);
+		$this->currentTPS = min(20, 1 / max(0.001, $totalTime));
+		$this->currentUse = min(1, $totalTime / 0.05);
+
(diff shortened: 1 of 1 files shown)


dktapps:
Probably should be considering something more like Spigot's TPS catch-up.

robske110:
I don't think that makes sense here.

madecode15:
This sort of doges a underlying problem. If a tick takes longer than 50ms to execute, the consequence of that is the server should lag. Users should run timings to see what is eating up there TPS to fix it, not depend on the server skipping ticks.

Though on the flip side, this could help those with under powered devices running tests servers to test things or as a extra precaution for production servers. Most of the anti hack issues I get come from running the server locally, as for production I never have issues with them.

thebigsmileXD:
@madecode15 i have a pretty powerful server, and actually only that tick skipping causes issues because entities are trying to catch up in weird ways, sometimes positions of pets for example get weirdly incorrect though respecting $currentTick. I think 
> Probably should be considering something more like Spigot's TPS catch-up.


Is actually a really good idea, but too resource heavy for pocketmine servers

dktapps:
I haven't found time to make any progress with this, so I'm going to close this for now.

PR #301, opened 2017-01-28
Stop the server sending blank messages to players

JackNoordhuis:
### Description
Stops the server queuing blank messages to send to players.

### Reason to modify
Some plugins allow users to edit messages and they can be set to a blank message. Currently a blank message will result in a blank message line being rendered on the client, this didn't happen before the hack to fix duplicated messages was added.

### Tests & Reviews
I've tested this myself with this script plugin:
```
<?php

/**
 * @name MessageSendTest
 * @main jacknoordhuis\MessageSendTest
 * @api 3.0.0-ALPHA3
 * @description Sends a blank message to players every 4 seconds
 * @author JackNoordhuis
 * @version 1.0.0
 */

namespace jacknoordhuis{
	class MessageSendTest extends \pocketmine\plugin\PluginBase{
		public function onEnable(){
			$this->getServer()->getScheduler()->scheduleRepeatingTask(new class($this) extends \pocketmine\scheduler\PluginTask{
				public function onRun($tick){
					foreach($this->getOwner()->getServer()->getOnlinePlayers() as $p) $p->sendMessage("");
				}
			}, 80);
		}
	}
}
```

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -3025,14 +3025,16 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 			$message = $message->getText();
 		}
 
-		//TODO: Remove this workaround (broken client MCPE 1.0.0)
-		$this->messageQueue[] = $this->server->getLanguage()->translateString($message);
-		/*
-		$pk = new TextPacket();
-		$pk->type = TextPacket::TYPE_RAW;
-		$pk->message = $this->server->getLanguage()->translateString($message);
-		$this->dataPacket($pk);
-		*/
+		if(($translated = $this->server->getLanguage()->translateString($message)) !== ""){
+			//TODO: Remove this workaround (broken client MCPE 1.0.0)
+			$this->messageQueue[] = $translated;
+			/*
+			$pk = new TextPacket();
+			$pk->type = TextPacket::TYPE_RAW;
+			$pk->message = $this->server->getLanguage()->translateString($message);
+			$this->dataPacket($pk);
+			*/
+		}
 	}
 
 	public function sendTranslation($message, array $parameters = []){


madcode15:
I'm sorry, but this is very dumb and obsolete.

What if you want to send a blank message? Many PC/Bukkit servers send blank messages for "fancy" game start menus, or in earlier versions of minecraft, clear chat.

This is just extra unneeded overhead.

Don't want to send a blank message? Easy solution. **Don't send it.** PocketMine should not have to baby sit plugins just because developers they don't know how to remove a line of code, in this case,  `$player->sendMessage("");`

KnownUnown:
> dumb
Why the hostility? Unnecessary, maybe. While it does restore the behavior to pre-1.0, was anyone really depending on that?
> obsolete
On what precedent?

src/pocketmine/Player.php
```diff
@@ -3025,14 +3025,16 @@ public function sendMessage($message){
...
-		$this->messageQueue[] = $this->server->getLanguage()->translateString($message);
-		/*
-		$pk = new TextPacket();
-		$pk->type = TextPacket::TYPE_RAW;
-		$pk->message = $this->server->getLanguage()->translateString($message);
-		$this->dataPacket($pk);
-		*/
+		if(($translated = $this->server->getLanguage()->translateString($message)) !== ""){
+			//TODO: Remove this workaround (broken client MCPE 1.0.0)
+			$this->messageQueue[] = $this->server->getLanguage()->translateString($translated);
```
SOF3: Why are you translating the translated message again?

JackNoordhuis:
@madcode15 I'm sorry that you failed to realise this has never been possible before the hack to fix message duplication was implemented, but you could still send blank messages. How? It's simple! We use our brains and do the obvious – \n.

src/pocketmine/Player.php
JackNoordhuis: This is what happens when I copy and paste code I've already written – had to reset the branch because git was being git. Will fix.

JackNoordhuis:
@dktapps This restores part of the behaviour from before the hack was implemented, this check was removed for some reason.

dktapps:
`Player->sendMessage("")` should indeed have the effect of sending a blank line, no? Message sending works just like `echo` does in bash. `echo ""` should print an empty line. I think this behaviour should not be removed.

dktapps:
This has been controversial and seems somewhat unpopular. I'm going to close this, but feel free to throw any further arguments you have.

Issue #302, opened 2017-01-28
Chunk tracking system needs a complete re-write


madecode15:
### Issue description
I haven't used PocketMine much since late 2014, so when I came back to it I got shocked a the amount of RAM it uses.

My PocketMine 1.4 server would never exceed 200mb.

In a memory leak, memory.. gets leaked. PocketMine is different.

With some players on my server (about 100), I noticed 750mb of RAM was being used, and with SteadFast, 1.4 based, the RAM being used was around 130MB.

I figured to let the server continue until it reaches my memory limit to get a back trace of what was causing the memory leak... And, it did not crash. The memory stops exceeding at 750MB and everything continues.

I doubt this is a memory leak, I think it is more of chunk loading taking too much memory with 1.5s ChunkLoader system.

For example:
https://github.com/pmmp/PocketMine-MP/blob/478a62e17bf5ff3876d692dc54d6a1ccd80659ea/src/pocketmine/level/Level.php#L569

Uses more memory then:
https://github.com/pmmp/PocketMine-MP/blob/754e0dbb4935f83ca4115efac15e23e639631d46/src/pocketmine/level/Level.php#L406

Anyways, chunk tracking needs a complete re-write and is a total mess.

**Here is a link to the original PM issue for some legacy: https://github.com/PocketMine/PocketMine-MP/issues/3153**

### Steps to reproduce the issue
1. Start server.
2. Get 50-100 players joined, let them load chunks, etc

* PocketMine-MP: https://github.com/pmmp/PocketMine-MP/commit/0535acf211f9f306a610307f71b29349b5e156d2
* PHP: 7.0.3
* Server OS: Linux
* Game version: PE

madcode15:
Added a update because my root cause seems incorrect (my fault). The issue is still persistent, however. PM is no stranger to memory leaks, I'm really looking forward to the 1.0.0 branch because of how much cleaner @dktapps re-wrote and re-fractured level provider code which will make working off level related leaks easier.

robske110:
Wait, doChunkGarbageCollection is not called on schedule currently. The PR adds that.

madcode15:
@robske110 That clears up the confusion. I look forward to testing your PR.

dktapps:
I've played around with various configuration options and haven't been able to reproduce this leak. Can you provide any more detail on the environment and configuration you are using?

madcode15:
@dktapps I'll try to list more steps and upload my config when I get enough time to test again.

[...]

PR #303, opened 2017-01-28
Address #302

robske110:
I think this is the right way to do it. I don’t even know why we don’t collect that in the first place. I'm still digging in the history, trying to find the old way this has been done. Weird thing: The /gc command does this, and I always thought doing the /gc command was the same as the scheduled garbage collector.

I also optimised performance of low memory trigger, probably unnecessary.

EDIT:
It was actually done in 1.4 and it looked like this:
```php
if($this->getProperty("chunk-gc.period-in-ticks", 600) > 0){
	$this->scheduler->scheduleDelayedRepeatingTask(new CallbackTask([$this, "doLevelGC"]), $this->getProperty("chunk-gc.period-in-ticks", 600), $this->getProperty("chunk-gc.period-in-ticks", 600));
}
```
(https://github.com/pmmp/PocketMine-MP/blob/3de14d8ba6d275c870d922cc46888cd9282a5bb3/src/pocketmine/Server.php#L1671)
doLevelGC in Server.php was simply iterating through the levels and calling the doChunkGarbageCollection.

### DIFF
--- a/src/pocketmine/MemoryManager.php
+++ b/src/pocketmine/MemoryManager.php
@@ -129,7 +129,7 @@ class MemoryManager{
 	public function trigger($memory, $limit, $global = false, $triggerCount = 0){
 		$this->server->getLogger()->debug(sprintf("[Memory Manager] %sLow memory triggered, limit %gMB, using %gMB",
 			$global ? "Global " : "", round(($limit / 1024) / 1024, 2), round(($memory / 1024) / 1024, 2)));
-		
+
 		foreach($this->server->getLevels() as $level){
 			if($this->cacheTrigger){
 				$level->clearCache(true);
--- a/src/pocketmine/MemoryManager.php
+++ b/src/pocketmine/MemoryManager.php
@@ -182,6 +182,11 @@ class MemoryManager{
 		if($this->garbageCollectionPeriod > 0 and ++$this->garbageCollectionTicker >= $this->garbageCollectionPeriod){
 			$this->garbageCollectionTicker = 0;
 			$this->triggerGarbageCollector();
+			foreach($this->server->getLevels() as $level){
+				if($this->chunkCollect){
+					$level->doChunkGarbageCollection();
+				}
+			}
 		}
 
 		Timings::$memoryManagerTimer->stopTiming();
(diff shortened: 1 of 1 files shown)


madcode15:
Hi @robske110,

While I appreciate the attempt to resolve I think there is a misunderstanding.

https://github.com/Hydreon/Steadfast2/blob/master/src/pocketmine/Server.php#L1735

Would this link help you understand perhaps what I am trying to say?

robske110:
I don't think we need another interval for this.
This function is called by
https://github.com/robske110/pmmp-PocketMine-MP/blob/chunk-garbage-collection/src/pocketmine/MemoryManager.php#L184
at the interval
https://github.com/robske110/pmmp-PocketMine-MP/blob/chunk-garbage-collection/src/pocketmine/MemoryManager.php#L103
set in
https://github.com/robske110/pmmp-PocketMine-MP/blob/chunk-garbage-collection/src/pocketmine/resources/pocketmine.yml#L52

madcode15:
@robske101

You are correct. Good work, I did not realize it was being called from there. This pull request seems good. +1

src/pocketmine/MemoryManager.php
```diff
@@ -200,10 +200,7 @@ public function triggerGarbageCollector(){
 		$cycles = gc_collect_cycles();
 		
 		foreach($this->server->getLevels() as $level){
-			if($this->cacheTrigger){
-				$level->clearCache(true);
-			}
```
dktapps: I think removing this is a bad idea. After some time, the cached material is not going to be useful anymore, and should be fully cleared.

xxFlare: Chunk cache dose not really use a lot of MB, and can help if tons of players join a "lobby" world. Regardless the current cache implementation is crap and chunk cache should be removed by time with a check when the cache was last used instead of clearing the array.

robske110:
Obsolete?

dktapps:
I don't believe this serves any purpose any more, since chunks are now automatically unloaded when there are no loaders attached them for a certain period of time. This would not even provide any relief from #419, since a chunk is still considered "in use" when it has loaders attached to it, even if those loaders are not valid.

jasonwynn10:
Would it hurt to have it anyway?

PR #304, opened 2017-01-28
Re-implemented chunk sending

### LINKED ISSUE #262: Chunk sending improvements

dktapps:
Re-implement chunk sending, send chunks inside a radius instead of below a count.

This is **SIGNIFICANTLY** better than the current method, which uses some stupid square-spiral crap under a certain count. Due to the way chunk orders are mirrored, this also appears to clients to be loading chunks a hell of a lot faster, since it starts in the centre and moves out from the axis, so the furthest chunks are always directly in front of the player.

This sends chunks in concentric squares around players. When the radius is hit, it will pad out the radius until a full circle of chunks is loaded around the player.

To set your server chunk radius, change `view-distance` in server.properties. Values are intended to be the same as MCPE render distance values. With matching client and server render distances the chunks should reach the horizon.

### TODO
- [ ] Send chunks in concentric circles as it is done in vanilla MCPE
- [x] Use `spawn-radius` for player spawning.
- [ ] Fix an issue where the edges of the world becomes visible when a player moves, but not into another chunk.
- [ ] Order chunks ondemand instead of on schedule

### Changes
- The setting `max-chunks` has been removed from pocketmine.yml in favour of a more vanilla `view-distance` in server.properties.
- The API method `Server->getViewDistance()` has been altered to now give a chunk radius value instead of the max-chunks crap we had before.
- API methods `Player->getViewDistance()` and `Player->setViewDistance()` have been added. The `setViewDistance()` performs safety checks and is capped at the value set in `view-distance` set in server.properties, so plugins can't mess things up.
- API method `Server->getAllowedViewDistance()` has been added, and takes an existing value. The value will be clamped between 2 (hardcoded minimum radius) and the maximum set in server.properties.

### Fixes
- Fixes chunk sending around players for a much more enjoyable experience.
- **This also fixes a dangerous memory leak related to clients changing their render distance** (blame @Intyre), where a player could increase their render distance and cause loading of a lot more chunks than are supposed to be allowed by the server. A W10 player changing their radius to 56 chunks for example, would cause the server to load thousands of chunks, completely disregarding the `max-chunks` setting in pocketmine.yml.

### NOTES
- You may notice increased memory usage per player **when increasing these values to something respectable**. This is normal and expected. The current default max-chunks is 192, which is slightly under the equivalent of radius 8 chunks (if the chunks had been being sent correctly...). Increasing it to, let's say, 14 radius will cause loading of 600+ chunks per player, which will obviously use more memory.

- The default value for `view-distance` in server.properties is 6 chunks, which is 130~ chunks. This is the expected value for most lower-end clients to use. Render distance 8 is 200~ chunks, etc.

- A player cannot exceed the render distance limit set in server.properties - the server will simply not send any more chunks. This may cause the horizon to not be quite perfect if the client's minimum render distance is higher than the server's max (W10 is a good example of this).

Closes #262

### DIFF
--- a/src/pocketmine/MemoryManager.php
+++ b/src/pocketmine/MemoryManager.php
@@ -47,7 +47,7 @@ class MemoryManager{
 	private $garbageCollectionTrigger;
 	private $garbageCollectionAsync;
 
-	private $chunkLimit;
+	private $chunkRadiusOverride;
 	private $chunkCollect;
 	private $chunkTrigger;
 
@@ -104,7 +104,7 @@ class MemoryManager{
 		$this->garbageCollectionTrigger = (bool) $this->server->getProperty("memory.garbage-collection.low-memory-trigger", true);
 		$this->garbageCollectionAsync = (bool) $this->server->getProperty("memory.garbage-collection.collect-async-worker", true);
 
-		$this->chunkLimit = (int) $this->server->getProperty("memory.max-chunks.trigger-limit", 96);
+		$this->chunkRadiusOverride = (int) $this->server->getProperty("memory.max-chunks.chunk-radius", 4);
 		$this->chunkCollect = (bool) $this->server->getProperty("memory.max-chunks.trigger-chunk-collect", true);
 		$this->chunkTrigger = (bool) $this->server->getProperty("memory.max-chunks.low-memory-trigger", true);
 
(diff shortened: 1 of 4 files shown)


madcode15:
@dktapps And.. You are correct.

And this commit has really improved my server performance. OrderChunks() does not even show up in my timings reports anymore.

The square crap was annoying .. it also made it nearly impossible to have that chunk shadow effect in the distance. +1

dktapps:
Yea, it's not perfect yet though... WIP

dktapps:
Probably ought to pad with empty chunks instead of actually requesting an extra ring. This is after all only a rendering issue, no need to send full extra chunks to get full distance.

dktapps:
Will merge this now, because I need the changes here for other development. The remaining TODOs will be moved to an issue.

PR #316, opened 2017-02-02
Sea Lantern Lighting Bug

Haxley:
Changed Sea Lantern from extending Solid to extending Transparent, like
GlowingObsidian and Glowstone, so that lighting works properly.

### DIFF
--- a/src/pocketmine/block/SeaLantern.php
+++ b/src/pocketmine/block/SeaLantern.php
@@ -23,7 +23,7 @@ namespace pocketmine\block;
 
 use pocketmine\item\Item;
 
-class SeaLantern extends Solid{
+class SeaLantern extends Transparent{
 
 	protected $id = self::SEA_LANTERN;
 


src/pocketmine/block/SeaLantern.php
```diff
@@ -49,4 +49,4 @@ public function getDrops(Item $item){
 		];
 	}
 
-}
```
robske110: please remove this :)

Haxley: @robske110 why? It fixes a bug. 

Sea Lanterns Transparent:

Sea Lanterns Solid:

Sea Lanterns Transparent:

Sea Lanterns Solid:


The top half of the Rainbow/DSX building is dark because Sea Lanterns are Solid with lightFilter 1 instead of 15

`				      	if($block->isSolid()){

						if($block->isTransparent()){

							if($block instanceof Liquid or $block instanceof Ice){

								self::$lightFilter[$id] = 2;

							}else{

								self::$lightFilter[$id] = 1;

							}

						}else{

							self::$lightFilter[$id] = 15;

						}

					}else{

						self::$lightFilter[$id] = 1;

					}`

dktapps:
yes, but changing it to Transparent is hacky, just like it was for glowing obsidian...

Since there are precedents, I will merge this, but it must be changed in the new block API. (Note to self)

src/pocketmine/block/SeaLantern.php
robske110: :facepalm: i meant the newline at the end of the file. It didn't contribute to the fix did it?

Haxley: @robske110 the line of whitespace?

Haxley:
@dktapps I tried to figure out a better way to do it, like maybe extend from another "Lantern" class like Liquid or Ice and adding an elif in the aforementioned Block class code-block. But changing to Transparent seemed much more simple, elegant, and the convention with glowstone and glowing obsidian. Sea Lanterns have been broken since they were released and this noob (me) came on to try and fix lol. Let me know how you'd approach it and I'll see if I can fix it. Anyway, thanks.

PR #323, opened 2017-02-04
Added support for Composer

dktapps:
This pull request adds Composer support to PocketMine-MP.

### What is Composer?
Composer is a dependency manager for PHP projects. Please see the [Composer documentation](https://getcomposer.org/doc/) for a more detailed overview of what Composer is and does.

### Why?
Adding Composer to the project brings much more ease to managing extension and library dependencies. This will allow PocketMine-MP to make use of pre-written libraries for various things instead of writing in-house implementations or copy-pasting code into the project.

### Changes to CI
I have altered the test process of Travis-CI to:
- Install dependencies via Composer
- Build a server phar including the `vendor` directory.

### Changes to the installation process
- If you wish to run PocketMine-MP from source-code or compile a phar from source-code, Composer is now REQUIRED to install necessary dependencies before compiling. Currently, there are no external dependencies - Composer is currently only in use for extension management, but this will change in the near future.

#### Steps to setup PocketMine-MP in source form with Composer
- Install Composer from [here](https://getcomposer.org/download/)
- cd into the directory where you cloned PocketMine-MP
- Run `composer install`
- Composer will tell you if any of the required dependencies are missing, including extensions.

*NOTE: You need the OpenSSL for PHP extension to use Composer.*

#### Compiling a phar
After installing dependencies, you may compile a phar using DevTools. You will need to use a new version of DevTools, as you will need to include multiple paths in the phar.
```
php DevTools.phar --make src,vendor --relative ./ --entry src/pocketmine/PocketMine.php --out PocketMine-MP.phar
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,5 @@ Desktop.ini
 # Sphinx-doc
 /docs/build/
 !/docs/requirements.txt
+
+vendor/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ before_script:
  - make install
  - cd ..
  - echo "extension=pthreads.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
+ - composer install
 
 script:
  - ./tests/travis.sh
--- a/composer.json
+++ b/composer.json
@@ -16,5 +16,13 @@
       "ext-weakref": ">=0.3.3",
       "ext-yaml": ">=2.0.0",
       "ext-zlib": ">=1.2.11"
+   },
+   "autoload": {
+      "exclude-from-classmap": [
+        "src/spl/stubs"
+      ],
+      "psr-0": {
+         "": ["src", "src/spl"]
+      }
    }
 }
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,20 @@
+{
+   "name": "pmmp/pocketmine-mp",
+   "description": "A server software for Minecraft: Pocket Edition written in PHP",
+   "type": "project",
+   "homepage": "https://pmmp.io",
+   "license": "LGPL-3.0",
+   "require": {
+      "php": ">=7.0",
+      "ext-curl": "*",
+      "ext-json": "*",
+      "ext-mbstring": "*",
+      "ext-openssl": "*",
+      "ext-phar": "*",
+      "ext-pthreads": ">=3.1.6",
+      "ext-sockets": "*",
+      "ext-weakref": ">=0.3.3",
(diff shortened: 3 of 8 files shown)


markkrueg:
Please do continue writing the instructions so we have a full process for compiling our own phar. I compile my own server quite often, and would like to use the correct new procedure. I don't mind the extra steps if necessary.

Also may need to update https://github.com/pmmp/php-build-scripts/blob/master/compile.sh for ssl support.

dktapps:
Do note that this PR is work in progress, including its description. I was working on DevTools at the time to have it build phars with Composer deps included.

dschwartz783:
I was thinking that automagically offering to install missing dependencies when PM starts might be ideal.

dktapps:
Bump. Note to self: need to get this finished so we can add auth and encryption.

timmillwood:
+1 to this!

TheDiamondYT1:
If the extensions are checked by composer, can we remove the checks from PocketMine.php?

TheDiamondYT1:
?

eternalharvest:
I want to test our plugin with static code analysis tool to check the code is compatible with specific PMMP. Can i use this feature for this purpose?

I want to check whether if types such as class, interface, and so on... in my code base is compatible with PMMP. But if I use static analysis tool, we need to checkout specified version of PMMP and setup include paths for class loader. I wish I can automate these step with this feature.

dktapps:
We're now including OpenSSL and Composer with our PHP packages.

xZeroMCPE:
The current PHP bin PMMP provides does not provide the GD library  which splits out:

`Call to undefined function imagecreatefrompng()`

So you should include `--with-gd` by default. Right now I'm forced to use a custom compiled bin. The gd library is very useful.

dktapps:
very useful _to you_, it is not required by PM, and the extra time taken to download libpng is often unnecessary. The compile script offers a `-g` flag to include gd.

dktapps:
(although this has no relevance to the PR... please take it elsewhere)

HimbeersaftLP:
Someone should make a GUI for the script lel

dschwartz783:
@HimbeersaftLP Server admins should not need GUIs. Windows admins are just lazy.

HimbeersaftLP:
I know, but it would help from a lot of complaining noobs....

Would you rather be spammed with annoying questions or surrounded by happy people?

dktapps:
annoying questions will be spammed anyway, and I have no time or motivation to write a compile script GUI.

PR #324, opened 2017-02-04
Fixed "Too many open files"

madcode15:
If I found this earlier I would of combined it with my last PR. :/

I can't believe this was not closed.


dktapps:
huh... how many times did you have to dump before that happened? :D

PR #325, opened 2017-02-04
Fixed the awful debug > 1 confusion

ghost:
This may be controversial.

I always see users thinking there are more then one debug levels. This aims to fix confusions, and looks much nicer to me in my opinion.


dktapps (changes requested):
I'm pretty sure \pocketmine\DEBUG is used elsewhere, it's scattered throughout the code.

I plan to implement proper debug levels in the future, so this serves no other purpose than reducing user confusion in the meantime.

madcode15:
Alright.

madcode15:
Ugh, Github client sucks, command line is much better especially when colaborating.

dktapps:
yeah, don't use that... console ftw

PR #327, opened 2017-02-05
Update the generator for better lands generator !

FRISCOWZMCPE:
update it or add a RemoveGenerator function soo we can edit the Normal generator with a plugin


dktapps:
your copy-pasting skills are incredible...

PR #328, opened 2017-02-05
Fixed half-trees on chunk boundaries, possibly fixed #265

### LINKED ISSUE #265: Cobble turns to Dirt when broken (mcpe-1.0)
### Issue description
When breaking cobblestone blocks, they will occasionally change to unbroken dirt blocks at the point of breaking.

1. Use the included world file (it is freshly generated with only this stairway created).
2. Break blocks as shown in the video with diamond pickaxe.

### OS and versions
* PocketMine-MP: PocketMine-MP 1.6.2dev 「Unleashed」 implementing API version 3.0.0-ALPHA2 for Minecraft: PE v1.0.0.16 (protocol version 100) commit bcf049a
* PHP: 7.0.10
* Server OS: Ubuntu 16.04.1 LTS
* Game version: Android 1.0.0

https://youtu.be/Ttjv8LZi43U
[world.zip](https://github.com/pmmp/PocketMine-MP/files/706891/world.zip)

dktapps:
When `Level->setChunk()` is called by anything, `ChunkLoader->onChunkChanged()` should be called for any attached loaders. This is _supposed_ to trigger a chunk resend for players, but currently PocketMine-MP sticks the load request into a load queue that for the most part is overwritten every second.

This pull request fixes the issue by removing the specified chunk from the player's index of usedChunks, which will cause the server to automatically resend that chunk next time a chunk order is made.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -3602,19 +3602,23 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	}
 
 	public function onChunkChanged(Chunk $chunk){
-		$this->loadQueue[Level::chunkHash($chunk->getX(), $chunk->getZ())] = abs(($this->x >> 4) - $chunk->getX()) + abs(($this->z >> 4) - $chunk->getZ());
+		unset($this->usedChunks[Level::chunkHash($chunk->getX(), $chunk->getZ())]);
 	}
 
 	public function onChunkLoaded(Chunk $chunk){
+
 	}
 
 	public function onChunkPopulated(Chunk $chunk){
+
 	}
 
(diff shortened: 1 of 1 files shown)


dschwartz783:
Shouldn't the surrounding chunks just be calculated, before a requested chunk actually gets sent? Would generate less traffic than this, and could easily be used to fill in trees, etc. Not sure if PM is designed well enough for that to be feasible though.

dktapps:
@dschwartz783 yeah... but this is just a bugfix.

dktapps:
Since this is a more general bug as opposed to specifically generator-related, I'm going to merge this. Will however put an issue in the tracker to note the population issue with resending entire chunks. This fixes the generator bug, but it's obviously not the best way to do it.

PR #331, opened 2017-02-07
Added extra data sent by client in 1.0.3

rj0495:


legoboy0215:
Why not camelCase?

robske110:
I think we should rename skinToken now, it's no skinToken anymore. Also, as @legiboy0215 camelCase is to be used.

dktapps:
Thanks for your contribution, but it doesn't make sense to add them like this. It's messy and manual labour every time new fields are added, which defeats the point of using json.

PR #334, opened 2017-02-08
Added ability to use config as array

FaigerSYS:


SOF3:
Related: PocketMine/PocketMine-MP#1386

SOF3 (commented):
As @sekjun9878 has mentioned in PocketMine/PocketMine-MP#1386, this change probably reduces code legibility.

Currently we are already seeing so many people messing up $this->config in their main class, trying to make another $config property in their subclass, while it is extreme confusion that it is sometimes Config and sometimes `Config->getAll()`. I'm not voting affirmatively or negatively in this comment. Anyone wanting to approve this pull request, keep in mind that, although PMMP isn't a nanny government, it is still undesirable to see plugin developers sometimes using array syntax and sometimes using the object-oriented notation. It will probably get very confusing to readers -- because of the wide use of getAll(), I would hesitate to let the legibility drop by removing the apparent assertion of the value type through object-oriented notation.

src/pocketmine/utils/Config.php
```diff
@@ -253,6 +253,46 @@ public function __unset($k){
...
+		return $this->get($k);
+	}
+
+	/**
+	 * @param $k
+	 * @param $v
+	 */
+	public function offsetSet($k, $v){
+		if(is_array($this->config)){
+			if(is_null($k)){
```
SOF3: Could better use `$k === null` as per PocketMine's convention.

dktapps:
please _stop_ using the web editor -_-

FaigerSYS:
Sorry, but at the moment I have no way to use `git`

dktapps:
1. Is there a use-case for this?
2. Given the readability + confusion issues, is it worth it?
3. I think [this](https://github.com/pmmp/PocketMine-MP/pull/334/files#diff-e3ea669ec6e43583528f59a8c7035e76R268) is a bad idea. The only time I can think of where the array key would be null is if you used syntax like `$this->config[] = $value`, which is highly discouraged. I don't think allowing that would provide a valid output in, for example, YAML (or any other format for that matter).

dktapps:
There is one precedent for this - the NBT library. However, the reason for that is because the NBT library stores values in objects, and the array access provides direct access to the values without needing `NamedTag->getValue()`. 

However, that's _also_ something I disagree with, because for new developers it is confusing to see someone else's code while learning, and think - hey, this is an array! Cool! but it's actually an array-access _object_. Better to be explicit so there is no confusion over the nature of the thing. Configs are one of the first things new developers will play with, and unless there is very clear documentation on their use (which there isn't) I think it's best if the code explains itself.

FaigerSYS:
Hmm, you're right. I thought that it could be more convenient, but it can be really confusing.

dktapps:
If you need this behaviour, you could also implement it into a plugin without too much difficulty.

PR #347, opened 2017-02-11
Fixed wrong temperature and temperature -> rainfall

Sandertv:
This Pull Request fixes two very minor things. One of which being the temperature of the birch biome. The correct value can be found here: http://minecraftpc.wikia.com/wiki/Birch_Forest. Furthermore, this pull request fixes temperature being set twice, while one of these should be rainfall, not temperature.


dktapps:
lmao, another one

PR #348, opened 2017-02-11
Added some very useless functions for easier scale changing of entities.

Sandertv:
I added some very simple functions for changing and getting the size of entities. This pull request is very useless. If there's more of these things I should add please tell me.


MrPowerGamerBR:
So useless that I won't even bother using it, I mean, who would want to spawn a huge creeper, or small players, or a huge chicken for an awesome boss fight? That's why it is useless.

10/10 would merge

dktapps:
Very useless huh? Fine.

Sandertv:
Wasn't the actual functions that are useless, but the pull request that kinda was, as it really wasn't any effort doing it.

dktapps:
Not at all. I don't intend to mess with crap, only rewrite it. If you want to make changes to it, feel free, but I don't see the point of spending my own effort adding to something I plan to rewrite. If someone else wants to, I'll review it. 👍

PR #355, opened 2017-02-14
Added API for transferring players to other servers

dktapps:
Added `Player->transfer()` and `PlayerTransferEvent`. See diff for details.

#### Why?
Because components of the network module, such as packet classes, are not covered by the PocketMine-MP API and are subject to change at any time. It is much easier and cleaner to create a core transfer API that can later be deprecated if some client features happen to be removed or broken.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -64,6 +64,7 @@ use pocketmine\event\player\PlayerRespawnEvent;
 use pocketmine\event\player\PlayerToggleFlightEvent;
 use pocketmine\event\player\PlayerToggleSneakEvent;
 use pocketmine\event\player\PlayerToggleSprintEvent;
+use pocketmine\event\player\PlayerTransferEvent;
 use pocketmine\event\server\DataPacketReceiveEvent;
 use pocketmine\event\server\DataPacketSendEvent;
 use pocketmine\event\TextContainer;
@@ -127,6 +128,7 @@ use pocketmine\network\protocol\SetTimePacket;
 use pocketmine\network\protocol\StartGamePacket;
 use pocketmine\network\protocol\TakeItemEntityPacket;
 use pocketmine\network\protocol\TextPacket;
+use pocketmine\network\protocol\TransferPacket;
 use pocketmine\network\protocol\UpdateAttributesPacket;
 use pocketmine\network\protocol\UpdateBlockPacket;
 use pocketmine\network\SourceInterface;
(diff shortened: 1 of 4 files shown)


src/pocketmine/Player.php
```diff
@@ -2976,6 +2978,27 @@ public function handleDataPacket(DataPacket $packet){
...
+		$this->server->getPluginManager()->callEvent($ev = new PlayerTransferEvent($this, $address, $port));
+		if(!$ev->isCancelled()){
+			$pk = new TransferPacket();
+			$pk->address = $ev->getAddress();
+			$pk->port = $ev->getPort();
+			$this->dataPacket($pk);
+			$this->close($this->getDisplayName() . " transferred to another server", "transferred to another server", false); //TODO: add language strings
+			return true;
+		}
+		return false;
```
SOF3: Note: Consider returning `$ev` instead, in case the caller wants to know the real final destination.

SOF3 (approved):

src/pocketmine/event/player/PlayerTransferEvent.php
```diff
@@ -0,0 +1,56 @@
...
+
+use pocketmine\event\Cancellable;
+use pocketmine\Player;
+
+class PlayerTransferEvent extends PlayerEvent implements Cancellable{
+	public static $handlerList = null;
+
+	protected $address;
+	protected $port = 19132;
+
```
SOF3: A $cause field would be appreciated. Though, not sure if it should be an int like those in EntityDamageEvent, or a string reason for more flexibility.

dktapps: well, the cause of a transfer is never going to be anything _but_ a plugin, so I consider that a little illogical.

SOF3: A string that identifies the plugin + an identifier under the plugin's space?

E.g. `shoghicp.fasttransfer.command`

dktapps: and a plugin would use that _how_? I don't see the sense in that.

SOF3: For example, another plugin would want to specifically cancel transfers from a particular plugin's particular action, but not others. This is an efficient way of interacting and sharing data with other plugins without direct dependency.

I have also considered a model like Android's Intent API, which allows actions to be taken even without knowing the specific existence of the other application (but for example, any applications that work with that particular type of data). This is however too big a change for the whole event system, so not gonna consider it in this pull request.

dktapps:
Will add a /transferserver command since MCPE now has one.

SOF3 (approved):
Code looks OK, except the event part which can be further extended

MCPEPIG:
Could you add this?
https://github.com/TesseractTeam/Tesseract/blob/master/src/pocketmine/network/protocol/LoginPacket.php#L116
https://github.com/TesseractTeam/Tesseract/blob/master/src/pocketmine/network/protocol/LoginPacket.php#L119
As well as a function to get them?

dktapps:
This data is already accessible via DataPacketReceive event, and with less mess. I would also thank you to not advertise spoons on our repository.

MCPEPIG:
Ok.

MCPEPIG:
But could you still add it?

dktapps:
I believe I said this data is already accessible...

SOF3:
Wait, we forgot to register permissions...

PR #363, opened 2017-02-16
Add TransferCommand.php

ghost:
Transfer Command!

### DIFF
--- a/src/pocketmine/command/defaults/TransferCommand.php
+++ b/src/pocketmine/command/defaults/TransferCommand.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace pocketmine\command\defaults;
+
+use pocketmine\network\protocol\TransferPacket;
+use pocketmine\command\CommandSender;
+use pocketmine\command\Command;
+use pocketmine\Server;
+use pocketmine\Player;
+
+class TransferCommand extends VanillaCommand{
+	
+	public function __construct($name){
+		parent::__construct(
+		$name,
+		"%pocketmine.command.transfer.description",
+		"%pocketmine.command.transfer.usage",
+		["transfer","connect"]
+		);
+		$this->setPermission("pocketmine.command.transfer");
+	}
+
+	public function execute(CommandSender $sender, $currentAlias, array $args){
+		if($sender instanceof Player){
+
+			if(isset($args[1])){
+				$port = $args[1];
+			}else{
+				$port = 19132;
+			}
+
+        		$pk = new TransferPacket();
+        		$pk->address = $args[0];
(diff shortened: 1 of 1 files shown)


SOF3 (changes requested):
---

src/pocketmine/command/defaults/TransferCommand.php
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	public function execute(CommandSender $sender, $currentAlias, array $args){
+		if($sender instanceof Player){
+
+			if(isset($args[1])){
+				$port = $args[1];
+			}else{
+				$port = 19132;
+			}
```
SOF3: Ever heard of something called null coalescence?

src/pocketmine/command/defaults/TransferCommand.php
```diff
@@ -0,0 +1,42 @@
...
+
+        		$pk = new TransferPacket();
+        		$pk->address = $args[0];
+        		$pk->port = $port;
```
SOF3: And what is the type of this value? Int or string?

src/pocketmine/command/defaults/TransferCommand.php
```diff
@@ -0,0 +1,42 @@
...
+		if($sender instanceof Player){
+
+			if(isset($args[1])){
+				$port = $args[1];
+			}else{
+				$port = 19132;
+			}
+
+        		$pk = new TransferPacket();
+        		$pk->address = $args[0];
```
SOF3: You should never assume user input is valid...

src/pocketmine/command/defaults/TransferCommand.php
```diff
@@ -0,0 +1,42 @@
...
+        		$pk->port = $port;
+        		$sender->dataPacket($pk);
+
+        		Command::broadcastCommandMessage($sender, "Connected to " . $args[0] . ":" . $port);
+		}else{
+			$sender->sendMessage("please run this command in game.");
```
SOF3: Translations?

SOF3 (changes requested):
Not that this will be merged once it's done, but please at least fix your indentation.

dktapps:
This should be done by plugins. The core will provide API only.

AppleDevelops:
Oh my

AryToNeX:
My opinion: you should provide /transferserver command 'cause it's available in single player.
Plugins will use the API to provide advanced functions such as ping before transfer, portals and so on.

dktapps:
Interesting, didn't spot that. 👍

PR #366, opened 2017-02-16
Use parse_ini_file instead of home-grown solution

dktapps:
I don't have any idea why this mess was here.


jasonwynn10:
in php 7 Hash marks (#) are no longer recognized as comments.

dktapps:
@jasonwynn10 yeah, I realized that last night... oops

PR #367, opened 2017-02-17
Throw exception when attempting to save a non-generated chunk

dktapps:
Should (hopefully) highlight what's going on with the big dirty holes in some generated maps.


TheClimbing:
@dktapps Here is the [crashdump](http://pastebin.com/eLWSM33s)

dktapps:
Awesome, that confirms it. 👍 thanks

TheClimbing:
Should I test it now?

dktapps:
Sure, feel free

PR #377, opened 2017-02-22
Simplified Vector3::getOppositeSide()

dktapps:
Vector3s are like dice. 1 is always opposite 0, 2 is always opposite 3, etc. Flipping the last bit is all that is necessary to find the opposite side.


SOF3 (approved):
Great! Never thought of this.

SOF3 (approved):
Not tested, but looks OK. Be aware that this is an API change, although should not affect anything more than an E_STRICT.

PR #400, opened 2017-03-05
Add fireball item and partial functionality

TheDiamondYT1:
This pull request adds the fireball *(fire charge)* item to PocketMine. When you tap on the ground using this item it will place fire, like flint and steel. This is exactly what should happen in vanilla.

I named the item "fireball" as per http://minecraft.gamepedia.com/Fire_Charge, so you can now `/give (player) fireball`, however i have kept the original implementation of `/give (player) fire_charge` to avoid breaking any plugins.

### DIFF
--- a/src/pocketmine/item/Fireball.php
+++ b/src/pocketmine/item/Fireball.php
@@ -1,0 +1,46 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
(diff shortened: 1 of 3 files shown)


corytortoise:
Looks great!

Muqsit:
Please don't take this negatively but...
Whenever you are doing any sort of additions to pmmp, make sure the pull request adds everything mentioned on the Minecraft Wiki.
http://minecraft.gamepedia.com/Fire_charge

You forgot to add the Fire Charge projectile entity.

mal0ne-23:
They will not accept "Partial" functionality

PR #401, opened 2017-03-06
Correction of continuation of color.

kametan0730MC:
Correction of continuation of color when sending multiple message.


src/pocketmine/Player.php
```diff
@@ -1616,7 +1616,7 @@ public function onUpdate($currentTick){
 		if(count($this->messageQueue) > 0){
 			$pk = new TextPacket();
 			$pk->type = TextPacket::TYPE_RAW;
-			$pk->message = implode("\n", $this->messageQueue);
+			$pk->message = implode(TextFormat::WHITE.TextFormat::RESET."\n", $this->messageQueue);
```
dktapps: Please correct your formatting. And I don't think the WHITE is necessary.

dktapps:
This also only covers one specific case of the issue. I don't think it is worth messing with, especially since that code is a client bug-hack anyway.

PR #422, opened 2017-03-13
Added support for commands to have names/aliases containing spaces

dktapps:
This is a pretty straightforward pull request allowing commands to have names and/or aliases containing spaces.

### Why?
MCPE supports this feature. It will be needed in the future. I also think this may be beneficial to developers who feel restricted by the lack of spaces support (like usernames).

Examples of this behaviour being utilised in vanilla are in the `time set` and `time add` commands. I'm not 100% sure why they were implemented like this instead of as overloads, but I assume there is a reason.

This is the first phase of new command implementation. Please test and review.

### DIFF
--- a/src/pocketmine/command/SimpleCommandMap.php
+++ b/src/pocketmine/command/SimpleCommandMap.php
@@ -177,15 +177,35 @@ class SimpleCommandMap implements CommandMap{
 		return true;
 	}
 
-	public function dispatch(CommandSender $sender, $commandLine){
-		$args = explode(" ", $commandLine);
+	/**
+	 * Returns a command to match the specified command line, or null if no matching command was found.
+	 * This method is intended to provide capability for handling commands with spaces in their name.
+	 * The referenced parameters will be modified accordingly depending on the resulting matched command.
+	 *
+	 * @param string   &$commandName
+	 * @param string[] &$args
+	 *
+	 * @return Command|null
+	 */
+	public function matchCommand(string &$commandName, array &$args){
+		$count = max(count($args), 255);
 
-		if(count($args) === 0){
-			return false;
+		for($i = 0; $i < $count; ++$i){
+			$commandName .= array_shift($args);
+			if(($command = $this->getCommand($commandName)) instanceof Command){
+				return $command;
+			}
+
+			$commandName .= " ";
 		}
 
-		$sentCommandLabel = strtolower(array_shift($args));
-		$target = $this->getCommand($sentCommandLabel);
+		return null;
+	}
+
+	public function dispatch(CommandSender $sender, $commandLine){
+		$args = explode(" ", $commandLine);
+		$sentCommandLabel = "";
+		$target = $this->matchCommand($sentCommandLabel, $args);
 
 		if($target === null){
 			return false;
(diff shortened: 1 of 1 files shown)


robske110 (approved):
Works flawlessly. 👍 
Test plugin:
```php
<?php

/**
 * @name SpacyCommands
 * @main robske_110\DHP\SpacyCommands
 * @version devhelperplgns/null
 * @api 3.0.0-ALPHA4
 * @description Testing commands with spaces. Gø fôr īt!
 * @author robske_110
 */


namespace robske_110\DHP{

	use robske_110\SPC\ScriptPluginCommands;

	class SpacyCommands extends PluginBase{
		
		public function onLoad(){
			$id = ScriptPluginCommands::initFor($this);
			ScriptPluginCommands::addCommand($id, [
				'name' => 'test mkay',
				'description' => 'Outputs your arg1',
				'permission' => 'op',
				'usage' => '/test mkay <sth>'
			]);
			ScriptPluginCommands::addCommand($id, [
				'name' => 'test wow',
				'description' => 'Outputs wow.',
				'permission' => 'op',
				'usage' => '/test wow'
			]);
			ScriptPluginCommands::addCommand($id, [
				'name' => 'server please the stop',
				'description' => 'lol',
				'aliases' => ['the stop please server', 'server shut up', 'you crash'],
				'permission' => 'true',
			]);
			$this->getServer()->getCommandMap()->registerAll($this->getDescription()->getName(), ScriptPluginCommands::getCommands($id));
		}
		
		public function onCommand(CommandSender $sender, Command $command, $label, array $args){
			switch($command->getName()){
				case "server please the stop":
					if($label == "you crash"){
						$this->getServer()->forceShutdown();
					}
					$this->getServer()->shutdown();
					return true;
				case "test mkay":
					if(isset($args[0])){
						$sender->sendMessage($args[0]);
						return true;
					}
					return false;
				break;
				case "test wow":
					echo("wow");
				break;
			}
		}
	}
}
/** LIBARIES */
namespace robske_110\SPC{
	use pocketmine\command\PluginCommand;
	use pocketmine\Plugin\Plugin;
	
	/**
	  * @author robske_110
	  * @version 0.1.2-php7-noutils
	  */
	abstract class ScriptPluginCommands{
		private static $plugins;
		
		public static function initFor(Plugin $plugin) : int {
			self::$plugins[] = [$plugin];
			return count(self::$plugins) - 1;
		}
		
		public static function addCommand(int $id, array $data){
			if(!isset(self::$plugins[$id])){
				Utils::critical("addCommand() has been called with an unkown ID!".PHP_EOL);
			}
			$cmd = new PluginCommand($data["name"], self::$plugins[$id][0]);
			if(isset($data["description"])){
				$cmd->setDescription($data["description"]);
			}
			if(isset($data["usage"])){
				$cmd->setUsage($data["usage"]);
			}
			if(isset($data["permission"])){
				$cmd->setPermission($data["permission"]);
			}
			if(isset($data["aliases"]) && is_array($data["aliases"])){
				$aliases = [];
				foreach($data["aliases"] as $alias){
					$aliases[] = $alias;
				}
				$cmd->setAliases($aliases);
			}
			self::$plugins[$id][1][] = $cmd;
		}
		
		public static function getCommands(int $id) : array {
			if(!isset(self::$plugins[$id])){
				echo("getCommands() has been called with an unkown ID!".PHP_EOL);
			}
			return self::$plugins[$id][1];
		}
	}
}
```

Issue #317, opened 2017-02-03
Error backtrace in crashdump is useless


dktapps:
### Issue description
If the server crashes due to something like a fatal error, the resulting backtrace in the crashdump will look like this:

```
#0 (): pocketmine\Server->crashDump(boolean)
```

Pretty useless, huh?

The issue is that when the server stops due to a fatal error, the registered shutdown function is called _internally_ from within PHP. Then [this](https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/CrashDump.php#L146) is called, and skips to the third stack frame. The problem with this is that it obtains a backtrace using [this](https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/PocketMine.php#L367-L368), which will _only_ point to frame 0 as being the crashDump() function, since that function was called internally by PHP.

### Steps to reproduce the issue

1. Crash the server with a fatal error.
2. Read your crashdump backtrace.

### OS and versions

* PocketMine-MP: 613226c6618eab0488d866e4d72ca7a73f69fa00
* PHP: 7.0.14
* Server OS: Windows 10 Home x64

dktapps:
Seems to be some issues with exception handlers + pthreads. Needs investigation before this can be fixed.

dktapps:
This issue cannot be resolved without xdebug, since fatal errors cannot be backtraced without a debugging tool.

PR #425, opened 2017-03-14
Stop skipping stack frames

dktapps:
- Fixes highest stack frame missing in exception logs
- fixed issues with line breaks and \r in error messages
- Crash dumps with an error will no longer include the `Server->crashDump()` stack frame. (This also means that currently, crashes with errors without xdebug will have no stack trace at all. See #317)


NycuRO:
yey 👍

PR #438, opened 2017-03-18
Fix client-side death bug

IrishPacks:
So here was the bug:

Whenever the player was damaged down to below 3 - 4 hearts and then if they healed quickly in some way(whether it would be plugin-based, effect-based, etc.) it would drop their health below 0 and kill them client-side, while they would be unaffected server-side.

After extensive testing both with and without this commit using the same techniques and repeating them many times, I've determined this was the bug causing it. As far as I'm aware SetHealthPacket is no longer used, because it's handled by attributes. So this will fix annoying issues with random client-side deaths when at a low enough health, making it for all around better gameplay.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -3340,15 +3340,6 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 		$this->dataPacket($pk);
 	}
 
-	public function setHealth($amount){
-		parent::setHealth($amount);
-		if($this->spawned === true){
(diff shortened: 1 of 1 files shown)


IrishPacks:
Check PMMP freenode chat.

dktapps:
This appears to be a client-sided issue. The existence of SetHealthPacket is now also debatable, since attributes do exist... hmm.

Anyway, please attend to my review comments. You might as well just remove that overridden method completely.

IrishPacks:
Fixed and rebased 👍

PR #439, opened 2017-03-18
Fixed hunger not saving

IrishPacks:
Fixed hunger not saving whatsoever. Tested extensively 👍


dktapps:
You may want to configure your IDE to use tabs instead of spaces for PocketMine-MP.

PR #446, opened 2017-03-22
Add ender pearl

TheDiamondYT1:
This pull request adds the Enderpearl item, Enderpearl entity and the ability to shoot an enderpearl then teleport to that spot.

The amount of hearts you will lose is 3 and a half, like it vanilla. 
*(Yes, i know i put '3' in the code but it makes me lose 3.5 so yeah)*

### DIFF
--- a/src/pocketmine/entity/Enderpearl.php
+++ b/src/pocketmine/entity/Enderpearl.php
@@ -1,0 +1,85 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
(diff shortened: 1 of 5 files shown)


TheDiamondYT1:
Hmm im guessing i will have to remove the age check (i copied the class from snowball entity) so i dont get teleported if the enderpearl is in the air too long?

dktapps:
You are attempting to two of the worst parts of PocketMine-MP. One part has a rewrite already underway, the other is pending rewrite. This PR comes at a bad time.

TheDiamondYT1:
@dktapps Oh, sorry. Should i close it?

dktapps:
Thanks for your contribution, but yes, I think this should be closed.

Other readers take note: a PR being rejected doesn't mean we have anything against you, always pay attention to _why_ the PR is closed. It's not something to be afraid of.

TheDiamondYT1:
@dktapps Just so i dont make another PR to it, whats the **other** worst part of pmmp?

dktapps:
Entities are the worst part. Really awful stuff.

Blocks/items are also terrible, rewrite is currently under way on the api3/blocks branch. Entities are pending.

madecode15:
@dktapps Yeah... entity ticking is a mess between classes... Also it is terribly optimized with performance.

Thunder33345:
You should try to put notes so other contributors dont do the same thing

PR #447, opened 2017-03-22
Added pocketmine.yml option to disable the title ticker

dktapps:
This gets really spammy on some consoles when you stick it in the background, but I don't want to lose colour for the sake of that.


TheDiamondYT1 (approved):

TheDiamondYT1:
Why would you lose colour?

dktapps:
You cannot currently disable the title ticker without disabling console colours.

TheDiamondYT1:
Oh...

PR #451, opened 2017-03-22
Added Permissions to bypass spawn protection (#440)

jasonw4331:
This PR adds in a `pocketmine.spawnprotect.bypass` permission that is default for ops and will allow a player to edit the area within the protected spawn area. This also removes the hard coding for ops only to edit the spawn protected area.

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -1508,7 +1508,7 @@ class Level implements ChunkManager, Metadatable{
 
 			if(($player->isSurvival() and $item instanceof Item and !$target->isBreakable($item)) or $player->isSpectator()){
 				$ev->setCancelled();
-			}elseif(!$player->hasPermission("pocketmine.bypassspawnprotect") and ($distance = $this->server->getSpawnRadius()) > -1){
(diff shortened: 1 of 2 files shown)


TheDiamondYT1 (approved):

dktapps:
Looks OK at first glance, will test this later.

dktapps:
I think the triple s may get confusing, it's not the easiest to type. Maybe `pocketmine.spawnprotect.bypass` ?

jasonwynn10:
Okay, but what would `pocketmine.spawnprotect` be used for?

dktapps:
It doesn't have to be used for anything, it's just a node.

madecode15:
Permission checking is pretty slow, you might want to consider performance before merging.

jasonwynn10:
@madecode15 is there a better method to check the player's permissions?

dktapps:
I'd like evidence that such a thing is `pretty slow` before messing with things. Until then it's simply premature optimization based on a comment. _Recalculation_ of permissions is slow, true, but I have no reason to believe that a simple permission check is "slow".

dktapps (approved):
Tested this, works fine.

PR #459, opened 2017-03-24
Throw an exception when something attempts to serialize Server

dktapps:
As per @Sandertv 's issue, this will provide clarification as to what is actually happening when something unintentionally serializes a referenced Server object instead of a confuzzling `Serialization of Closure is not allowed` exception.

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -2434,4 +2434,13 @@ class Server{
 
 		return true;
 	}
+
+	/**
+	 * Called when something attempts to serialize the server instance.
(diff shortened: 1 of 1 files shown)


Muqsit:
Quality support.

TheDiamondYT1:
Does adding an __ like override default php functions?

Sandertv:
Awesome, @dktapps. Thank you so much.

legoboy0215:
@TheDiamondYT1 http://php.net/manual/en/language.oop5.magic.php#object.sleep

PR #461, opened 2017-03-25
Fixed 2 errors that bugged NetBeans IDE

RTGDaCoder:
Hi! I'm not sure whatever i changed makes stuff worse or better. Hopefully, it did somethin good!

Thanks


dktapps:
IDEs require a human brain to make sense of the things they don't like. Don't blindly trust what it tells you.

Issue #251, opened 2017-01-12
Block update recursion causes a crash when xdebug is enabled


thebigsmileXD:
### Issue description

Death loops when breaking huge cacti

### Steps to reproduce the issue

1. Build a high cactus
2. Break the lowest block

### OS and versions

* PocketMine-MP: https://github.com/pmmp/PocketMine-MP/commit/67940ad6ab4b9632823a749a9e7fa973d4500620
* PHP:
* Server OS:
* Game version: PE/Win10 (delete as appropriate)

### Crashdump, backtrace or other files

Gc too many cycles

dktapps:
Cannot reproduce, but related issues:
- Immense lag
- Far too many dropped cacti
- Drops freeze in midair

thebigsmileXD:
Should i provide a video as proof?

SOF3:
We don't need **proof**. It is to everyone's least concern whether the bug exists or not if it cannot be properly identified and fixed.

thebigsmileXD:
Nice, so because you can't find the bug you say it doesn't exist?
![image](https://cloud.githubusercontent.com/assets/8733998/21892781/be70173e-d8d8-11e6-812c-2ae336ad76e0.png)

dktapps:
Do read his comment again. We don't need proof the bug exists; we need a way we can reproduce the bug. As I have already stated, I am unable to reproduce this bug. Until it can be reproduced, it cannot be fixed unless there is an obvious cause, in which case reproducing is still preferable anyway, to ensure that any changes actually _fix_ the issue.

thebigsmileXD:
By the way, same happens if you have a giant row of wall signs and break one.

SOF3:
* [Cactus.php:76](https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/block/Cactus.php#L76) Cactus::onUpdate()
* [Level.php:1594](https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/level/Level.php#L1594) Level::useBreakOn()
* [Block.php:359](https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/block/Block.php#L359) Block::onBreak()
* [Level.php:1445](https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/level/Level.php#L1445) Level::setBlock()
[...]

PR #464, opened 2017-03-25
Fixed block update recursion issues

dktapps:
Fixes: #251, PocketMine/PocketMine-MP#3752

Obsoletes: PocketMine/PocketMine-MP#4200, PocketMine/PocketMine-MP#4057, PocketMine/PocketMine-MP#3806, PocketMine/PocketMine-MP#3763, PocketMine/PocketMine-MP#3762.

## Changes
This PR adds an extra queue for neighbour block updates. Instead of directly updating blocks when a block is changed, neighbour block updates are now queued to be run at the end of the tick.

@PEMapModder 's original suggested solution [here](https://github.com/PocketMine/PocketMine-MP/issues/3752#issuecomment-164487640) would have worked, except for the issue of plugin events, and also the issue that affected blocks would have needed to handle the scheduled update, which is wrong. This  pull request caters for all possibilities of such recursion without breaking plugin events.

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -193,11 +193,11 @@ class Level implements ChunkManager, Metadatable{
 	private $changedBlocks = [];
 
 	/** @var ReversePriorityQueue */
-	private $updateQueue;
-	private $updateQueueIndex = [];
+	private $scheduledBlockUpdateQueue;
+	private $scheduledBlockUpdateQueueIndex = [];
 
 	/** @var \SplQueue */
-	private $neighbourBlockUpdates = [];
+	private $neighbourBlockUpdateQueue = [];
 
 	/** @var Player[][] */
 	private $chunkSendQueue = [];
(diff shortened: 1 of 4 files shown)


src/pocketmine/level/Level.php
```diff
@@ -1044,15 +1063,45 @@ public function updateAround(Vector3 $pos){
...
+
+	/**
+	 * Schedules a block update to be executed after the specified number of ticks.
+	 * Blocks will be updated with the scheduled update type.
+	 *
+	 * @param Vector3 $pos
+	 * @param int     $delay
+	 */
+	public function scheduleDelayedBlockUpdate(Vector3 $pos, int $delay){
+		if(isset($this->scheduledBlockUpdateQueueIndex[$index = Level::blockHash($pos->x, $pos->y, $pos->z)]) and $this->scheduledBlockUpdateQueueIndex[$index] <= $delay){
```
dktapps: I don't know. Ask Shoghi.

Issue #466, opened 2017-03-26
Excess use of NBT slows everything down


dktapps:
### Issue description
The NBT library is extremely slow. For example, reading player data on login contributes ~25ms on player login. Player logins as a whole are extremely slow, and this contributes ~50% of the lag on login processing (~50ms on my machine).

This is nothing to do with disk reading or decompression either. I did some timing of player data reading and gained the below results:
- Disk read: 0.1ms
- zlib decompress: 0.2ms
- Read NBT and create NBT object tree: ~22ms

### OS and versions

* PocketMine-MP: dda8c6cc8f04d917688d003ebe6c4b2b62cd94af
* PHP: 7.0.16
* Server OS: w10
* Game version: irrelevant

phoshp:
If you solve this problem, you are supe.

ghost:
I was tracing back the where the 22ms was coming from, and I got most of it from this single chunk of code.

https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/entity/Human.php#L276

dktapps:
@madecode15 that's the other half of the 50ms I noted above. Raw deserialization by itself takes 25ms.

dktapps:
NBT really does need a total rewrite. It's slow and badly designed.

kniffo80:
Is the NBT implementation coming from a library or is it self implemented?

dktapps:
It's homegrown by shoghicp. Blame him, as always.

kniffo80:
I took a look if there's a ready-made lib that can be used. The only one i found for PHP which MAY suit (although the last commit was in december 2015) ist this one: https://github.com/rickselby/NBT

But that's just an idea (to use external library) - when i learned one thing while being a software developer: reuse existing code and don't invent the wheel twice. But as I said: only an idea. Maybe it doesn't suit for the PMMP cases ...

ishitatsuyuki:
Does a C-based extension help here?

EDIT: or, desirable?

dktapps:
yes, I intend to write an extension for this.

PR #468, opened 2017-03-27
Don't use stdClass for nbt, add temporary magic methods.

tarik02:
It's a little optimization of CompoundTag class, that has performance leaks with stdClass's.

### DIFF
--- a/src/pocketmine/nbt/tag/CompoundTag.php
+++ b/src/pocketmine/nbt/tag/CompoundTag.php
@@ -25,16 +25,17 @@ use pocketmine\nbt\NBT;
 
 #include <rules/NBT.h>
 
-class CompoundTag extends NamedTag implements \ArrayAccess{
+class CompoundTag extends NamedTag implements \ArrayAccess, \IteratorAggregate{
 
 	/**
 	 * @param string     $name
 	 * @param NamedTag[] $value
 	 */
 	public function __construct($name = "", $value = []){
+		$this->value = [];
 		$this->__name = $name;
 		foreach($value as $tag){
-			$this->{$tag->getName()} = $tag;
+			$this->value[$tag->getName()] = $tag;
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


Tarik02:
This is not full solution of the issue, but it can speed up CompoundTags.
https://github.com/pmmp/PocketMine-MP/issues/466

dktapps:
I don't think micro optimization is going to help the situation, the NBT library needs a full rewrite. Regardless, I'd like to see some tests that show some evidence that this provides an enhancement.

Tarik02:
Yes, this don't helps fully, but this can be temponary solution of problem.

Muqsit:
Nice, but whats the need of a temporary solution when it's eventually going to be removed soon?

kniffo80:
@Muqsit removed soon? Where did you get that from?

Tarik02:
@kniffo80, nbt in pocketmine will be rewritten from scratch soon, so this improvement will be removed.

madecode15:
I tested this. There are no noticeable performance improvements, seems to be a micro-optimization. For issue #466 dktapps is talking about fixing the slow 22ms it takes to process NBT data. This is in no way a temporary solution...

SOF3:
I believe @shoghicp had evidence that his approach is better than using __get when he initially wrote this library.

Tarik02:
@SOF3, yes, but stdClass is not so fast thing.

dktapps:
Everything in PHP is slow. Playing on the micro optimizations isn't going to help anybody.

Muqsit:
Make another C extension for handling NBTs 😂

dktapps:
Note that C-ifying everything won't make it magically faster. The ChunkUtils extension produced a huge performance improvement to Anvil worlds because it cut out thousands of PHP method calls (ord(), chr()) by dealing with bytes on a C level (method calls are very expensive). The same won't apply so much to NBT.

Tarik02:
I think, that nbt is so slow because of many instanceof operators.

madecode15:
@Tarik02 

No, instanceof call are not "slow". They are not the fastest operation either but that has nothing to do with anything.

After reviewing the code I think it is best to leave this untouched. According to php.net:
> Caution
PHP reserves all function names starting with __ as magical. It is recommended that you do not use function names with __ in PHP unless you want some documented magic functionality.

You can micro-optimize something hundreds of times and you won't save a single ms.

Best to wait for a actual solution by dktapps.

madecode15:
@dktapps Is the ChunkUtils extension public now?

dktapps:
https://github.com/dktapps/PocketMine-C-ChunkUtils

madecode15:
I don't have much time to benchmark it right now, but is it faster then the MCRegion world provider? If I remember correctly PocketMine process Anvil the slowest (which chunk utils speed up).

dktapps:
Any results you may remember from before 1.0 are now invalidated. The chunk format changed in 1.0 to accommodate the 256 build height.

McRegion used to be the fastest of the bunch due to its save format matching the MCPE network format. That's no longer true, which results in McRegion having a significant slowdown due to conversion. We also don't recommend using McRegion anymore since it doesn't support to 256-block build height.

Anvil without the C extension is extremely slow though, yes. The new chunk format is similar to Anvil, but with data order XZY instead of YZX. The PHP conversions of those chunks take a large amount of time, hence the performance issues. The C extension alleviates those.

PR #472, opened 2017-03-28
Set forceMovement to null when player is closed.

ghost:
forceMovement can hold a reference to a position sometimes, just like $this->spawnPosition. Except if a player leaves and the reference is on, it does not get nulled.


dktapps:
This isn't really an issue, since force movement will only be a vector if the player is currently teleporting. The level ref would be removed when next accessed, if accessed. The chances of this actually occurring are very slim.

The only time level refs become a problem are when something attempts to use them after thet have been, like the famous getName() on null (#24).

PR #475, opened 2017-03-29
Fix Fishing Rod

KennFatt:
- Set Fishing Rod as Tool
- Damageable (65 Durability)
- Set max count to 1 (as default tool)
- Fix "Unknown" to "Fishing Rod"

### DIFF
--- a/src/pocketmine/item/FishingRod.php
+++ b/src/pocketmine/item/FishingRod.php
@@ -21,8 +21,8 @@
 
 namespace pocketmine\item;
 
-class FishingRod extends Item{
+class FishingRod extends Tools {
 	public function __construct($meta = 0, $count = 1){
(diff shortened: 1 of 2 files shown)


corytortoise:
You should have done this from the api3/blocks branch, because a lot of item classes are being almost entirely rewritten there.

KennFatt:
@corytortoise oh ok sorry, i'll move it. thanks for notice me! :gem:

corytortoise:
You may want to look at the commits there for reference as to how dktapps wants tools added, assuming he isn't against anything being added to that branch right now.

KennFatt:
@corytortoise ikr, if he ignore it, at least he read fishing rod is buggy and he will fix it with his way.

PR #485, opened 2017-03-29
Added a getName() function for Trees, added a little more documentation.

Sandertv:
This pull request is nothing special. It makes getting the name of Tree objects a bit easier, which is useful for plugins at times. I also added a bit more documentation. It seemed a bit unclear to me that growing trees didn't work if you grew them from a dirt block, but it should be the block above it.

### DIFF
--- a/src/pocketmine/level/generator/object/BirchTree.php
+++ b/src/pocketmine/level/generator/object/BirchTree.php
@@ -45,7 +45,7 @@ class BirchTree extends Tree{
 		parent::placeObject($level, $x, $y, $z, $random);
 	}
 	
-	public function getName(): string {
+	public function getName() : string{
 		return "Birch Tree";
 	}
 }
--- a/src/pocketmine/level/generator/object/BirchTree.php
+++ b/src/pocketmine/level/generator/object/BirchTree.php
@@ -44,4 +44,8 @@ class BirchTree extends Tree{
 		}
 		parent::placeObject($level, $x, $y, $z, $random);
 	}
+	
+	public function getName(): string {
+		return "Birch Tree";
+	}
 }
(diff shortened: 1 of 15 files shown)


Sandertv:
I'm trying to fix the spaces that should be before the : , but git won't let me push those...

dktapps:
May I ask what editor you are using?

Sandertv:
PHPStorm. Just found out that it had the auto reformat code option on when making a commit. Sorry for that.

dktapps:
There's still trailing whitespace everywhere, would you mind fixing it please?

dktapps:
Also, you could use [this](https://github.com/pmmp/PocketMine-MP/pull/485/files#diff-d35e4fcc44efbe13605fefcbdf2a5373R40) and check it against wood block names instead of hardcoding it. I don't like hardcoding stuff.

Sandertv:
Fixing trailing whitespace right now... Problem for the wood checking though, is that it will get messy when large trees get implemented. Any ideas on that?

Sandertv:
Actually, I just thought of something that should do the job.

src/pocketmine/level/generator/object/BirchTree.php
```diff
@@ -35,6 +35,7 @@ public function __construct($superBirch = false){
 		$this->leafBlock = Block::LEAVES;
 		$this->type = Wood::BIRCH;
 		$this->superBirch = (bool) $superBirch;
+		$this->isBig = (bool)$superBirch;
```
Sandertv: *facepalm* 😫

Sandertv:
Okay this can be merged now, if there's no more things that have to be changed.

Sandertv:
I've just accidentally pushed something that should be put in a separated pull request to this one. It's not an actual change, but I'll find a way to fix it if you don't want to merge it like this.

dktapps:
Easily fixed. Just cherry-pick the commit onto a different branch, and then:

```
git reset HEAD~1 --hard
git push -f
```

That will reset the branch back by 1 commit and force-update your fork. Make sure you cherry-pick the offending commit somewhere else or you will lose it.

Sandertv:
Obsoleted by https://github.com/pmmp/PocketMine-MP/pull/815

PR #512, opened 2017-04-04
Add and fix some enchantment functions.

KN19294:
**Purpose of this commit** is just mere API enhancement. These functions narrow down enchantment checks and removal.
PocketMine's code formatting has been followed.

### DIFF
--- a/src/pocketmine/item/Item.php
+++ b/src/pocketmine/item/Item.php
@@ -596,7 +596,13 @@ class Item implements ItemIds, \JsonSerializable{
 		}
 
 		if(!$found){
-			$tag->ench->{count($tag->ench) + 1} = new CompoundTag("", [
+			$count = 0;
+			foreach($tag->ench as $key => $value){
+				if(is_numeric($key)){
+					++$count;
+				}
+			}
+			$tag->ench->{$count + 1} = new CompoundTag("", [
 				"id" => new ShortTag("id", $ench->getId()),
 				"lvl" => new ShortTag("lvl", $ench->getLevel())
 			]);
--- a/src/pocketmine/item/Item.php
+++ b/src/pocketmine/item/Item.php
@@ -494,14 +494,7 @@ class Item implements ItemIds, \JsonSerializable{
 		}
 
 		$tag = $this->getNamedTag();
-		if(isset($tag->ench)){
-			$tag = $tag->ench;
-			if($tag instanceof ListTag){
-				return true;
-			}
-		}
-
-		return false;
+		return isset($tag->ench) and $tag->ench instanceof ListTag;
 	}
 
 	/**
@@ -566,6 +559,14 @@ class Item implements ItemIds, \JsonSerializable{
 		}
 		$this->setNamedTag($tag);
 	}
+	
+	public function removeEnchantments(){
+		if($this->hasEnchantments()){
+			$tag = $this->getNamedTag();
+			unset($tag->ench);
+			$this->setNamedTag($tag);
+		}
+	}
 
 	/**
 	 * @param Enchantment $ench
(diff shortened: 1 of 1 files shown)


src/pocketmine/item/Item.php
```diff
@@ -524,6 +542,27 @@ public function getEnchantment(int $id){
 
 		return null;
 	}
+	
+	/**
+	* @param int $id
+	* @param int $level
+	*/
+	public function removeEnchantment(int $id, int $level = null){
```
jasonwynn10: Why specify the level? If you are just removing the enchantment, then only the id should be needed.

Muqsit: Or what if you want to remove an enchantment of a specific id and level?

madecode15:
PocketMines enchantment system is a utter mess. Better to wait until shoghicp's inventory transaction update so dktapps can rewrite inventory.

src/pocketmine/item/Item.php
```diff
@@ -505,6 +505,24 @@ public function hasEnchantments() : bool{
 	}
 
 	/**
+	* @param int $id
+	*
+	* @return bool
+	*/
+	public function hasEnchantment(int $id) : bool{
```
dktapps: This would be more useful if a level was able to be specified.

PMMPFTW:
@madecode15 That doesn't make sense. Even if the transaction system were to have a change, it wouldn't change anything related to NBT tags. "ench" nbt tag is used in vanilla MCPE and other Minecraft variants too. And these enchantment functions are just manipulation of item's nbt tag, just like other manipulation functions such as getCustomName() and setCustomName().
If that were true, then why did @dktapps implement setLore() and getLore()?

src/pocketmine/item/Item.php
PMMPFTW: @dktapps I'll do it once I reach home.

src/pocketmine/item/Item.php
dktapps: I would prefer an unused int value such as -1 here as opposed to `null` I don't like using `null` apart from for objects due to it encouraging unexpected behaviour.

PMMPFTW:
Hey, @dktapps.
I have updated the code to your liking. Hope it is acceptable.

dktapps:
Looks good at first glance, I've needed these.

However I don't have time to properly test and review this at the moment, and I don't currently have a computer to test. This may have to wait for a few days until I get a new computer. Thanks for
for your patience.

PMMPFTW:
@dktapps Sounds good. Till then, I'll try finding errors by deep testing the code.

src/pocketmine/item/Item.php
ImagicalGamer: You cannot add Sharpness 2 and Sharpness 1 to a weapon, only 1 of each enchantment is supported making the level irrelevant to remove enchantments.

mal0ne-23: Agree, level in the remove function is pointless. Just check for the level in hasEnchantment, then remove it by ID.

PMMPFTW: Not gonna change. It's possible to multi enchant an item. If you know how nbt tags work, you can easily add enchants of the same ID, but different levels.

PMMPFTW:
The latest commit is a critical fix. Only one enchant per item could be added per item before. The moment you try to add another enchant, it would just overwrite the ench NBT's key.

Muqsit:
Oh, dam*.
I was really confused as to why enchants were not working *facepalm*

src/pocketmine/item/Item.php
```diff
@@ -596,7 +596,13 @@ public function addEnchantment(Enchantment $ench){
...
 
 		if(!$found){
-			$tag->ench->{count($tag->ench) + 1} = new CompoundTag("", [
+			$count = 0;
+			foreach($tag->ench as $key => $value){
+				if(is_numeric($key)){
+					++$count;
+				}
+			}
+			$tag->ench->{$count + 1} = new CompoundTag("", [
```
dktapps: How does this change anything other than making the code more ugly? You've simply increased the complexity of the code for the exact same results. A ListTag will only have numeric indices anyway.

dktapps (dismissed):

src/pocketmine/item/Item.php
IrishPacks: only one enchant could be added before, tested on latest commit without this commit.

dktapps: The code is exactly the same. Read my comment.

Muqsit: It is the same thing, yet also fixes the bug. The loop isn't needed though.

Muqsit: Isn't it supposed to be

$tag->ench{count($tag->ench)} = ...

rather than

$tag->ench{count($tag->ench) + 1} = ...


Count, unlike keys, doesn't take the first value of the array to be the 0th value.

dktapps: @Muqsit well, it doesn't care about keys... if 0 1 2 3 4 are set, count is 5, that's basic common sense. I believe you are right though.

TheDiamondYT1:
You'd be better off moving enchantments to JSON. 

Take at look at the effect class (i think in the network branch?) to see how the effects are loaded from json.

dktapps:
@TheDiamondYT1 That isn't the topic of this PR. This PR is adding some API, not changing base functionality.

TheDiamondYT1:
@dktapps Ah, sorry. I got bored of reading it half way through :P

Muqsit:
Looks good

dktapps:
This seems good to me apart from minor formatting issues, but I can resolve those before I merge. Does anyone have any reviews to make based on tests?

Muqsit:
@dktapps Tested, there were no bugs. I haven't tested the removeEnchantment() function though. (Yeah that unnecessary tabular space is annoying).

dktapps:
Reformatted locally and merged this to api3/blocks, it's more suitable and more needed there at the moment. Thanks.

src/pocketmine/item/Item.php
```diff
@@ -494,16 +494,30 @@ public function hasEnchantments() : bool{
...
+				if($level === -1 or $entry["lvl"] === $level){
+					return true;
+				}
 			}
 		}
-
 		return false;
 	}
-
+
```
Muqsit: well, that was unnecessary :|

PR #517, opened 2017-04-05
Update Config.php for supporting Chinese character encode??

BlockForWhale:
Old Config.php can't save Chinese character well. It was saved as unicode. Adding this option "JSON_UNESCAPED_UNICODE" can fix this bug.

### DIFF
--- a/src/pocketmine/utils/Config.php
+++ b/src/pocketmine/utils/Config.php
@@ -187,7 +187,7 @@ class Config{
 						$content = $this->writeProperties();
 						break;
 					case Config::JSON:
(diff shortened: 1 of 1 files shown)


robske110:
Well, it isn't considered a bug, it is a feature of json, because it aims to provide a simple text, which can be read easily. So to stick to the json "standard" we should leave this as is.

SOF3:
To ensure backwards compatibility where some users may read JSON configs by readers that don't support non-ASCII JSON code, I recommend making it optional.

SOF3:
Obsoleted by #524.

PR #524, opened 2017-04-06
Allow custom JSON pretty print options

### LINKED ISSUE #516: Fix JSON_encode for Chinese characters
For Chinese characters, JSON config file cannot save it well. All the characters will be changed to unicode characters.Like This:
{"\u4efb\u52a1\u7ebf\u7a0b": ""}
The right text is:
{"任务线程": ""}
I think it is a mistake of Config.php.
I found it's easy to solve, just need to add one option at Config.php line 190:
Origin:
`$content = json_encode($this->config, JSON_PRETTY_PRINT | JSON_BIGINT_AS_STRING);`
Modified:
`$content = json_encode($this->config, JSON_PRETTY_PRINT | JSON_BIGINT_AS_STRING | JSON_UNESCAPED_UNICODE);`
Thank you for your consideration. (Never mind why I don't use pull...)

SOF3:
Resolves #516
Closes #517 

Example:

```php
$config = new Config("path/to.json", Config::JSON);
$config->enableJsonOption(JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE)->save();
```

### DIFF
--- a/src/pocketmine/utils/Config.php
+++ b/src/pocketmine/utils/Config.php
@@ -50,6 +50,8 @@ class Config{
 	private $correct = false;
 	/** @var int */
 	private $type = Config::DETECT;
+	/** @var int */
+	private $jsonOptions = JSON_PRETTY_PRINT | JSON_BIGINT_AS_STRING;
 
 	public static $formats = [
 		"properties" => Config::PROPERTIES,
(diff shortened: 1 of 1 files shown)


dktapps:
To be honest I think the config types need some abstraction, since this is just one extra useless property for other config types. However, as solution for this issue this looks fine.

BlockForWhale:
Yeah. Better solution than mine:-). @SOF3

PR #811, opened 2017-04-10
Actually use iusername instead of repeatedly lowercasing player names

TheDiamondYT1:
I messed up the commit name, lol.


jasonwynn10:
Isn't that changing the code stye?

TheDiamondYT1:
@jasonwynn10 I have no idea. @dktapps made a commit to set iusername to the lowercase player name, so why not utilize that variable instead of using strtolower every time

dktapps (approved):

dktapps:
In future please prefer creating a separate branch on your fork for creating your pull request.

PR #814, opened 2017-04-11
Visibility keyword after final or abstract keyword.

Sandertv:
This pull request changes visibility keywords to be before the abstract and final keyword. This follows the PSR-2 standards, and should be discussed if somebody has objections. See https://github.com/Sandertv/PocketMine-MP/commit/b2eb0dfaa2f44fe012e80f094511b15b12651cfb

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -3069,7 +3069,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	 * @param string $reason  Reason showed in console
 	 * @param bool   $notify
 	 */
-	public final function close($message = "", $reason = "generic reason", $notify = true){
+	final public function close($message = "", $reason = "generic reason", $notify = true){
 
 		if($this->connected and !$this->closed){
 			if($notify and strlen((string) $reason) > 0){
--- a/src/pocketmine/block/Block.php
+++ b/src/pocketmine/block/Block.php
@@ -482,7 +482,7 @@ class Block extends Position implements BlockIds, Metadatable{
 	/**
 	 * @return int
 	 */
-	public final function getId(){
+	final public function getId(){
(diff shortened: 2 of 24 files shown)


dktapps:
As per the PSR-2 style guidelines [here](http://www.php-fig.org/psr/psr-2/):

> Visibility MUST be declared on all properties and methods; abstract and final MUST be declared before the visibility; static MUST be declared after the visibility.

This is therefore incorrect.

TheDiamondYT1:
This just looks ***ing terrible imho

Sandertv:
@TheDiamondYT1 
I prefer visibility before static/abstract/final too, but either of them is at least better than it is now. Now it's just a combination of both.

SOF3 (approved):
This correctly resolves all violations of PSR-2 regarding order of visibility and inheritance modifiers in this repo.

Note that in PocketMine-SPL, there is one violation of `final public` in ThreadedLoggerAttachment.php and two violations of `public static` in leveldb.php.

kniffo80:
Coming from the JAVA world - this looks a little weird :expressionless:

dktapps:
Nobody cares about the Java world. This is PHP.

TheDiamondYT1:
Java 4 life

SOF3:
This is not the place to show off what languages you know or what code styles you support.

Before someone starts a _jihad_ to DDoS servers not using Java, I'm locking this conversation. If you don't like this choice of code style, blame @shoghicp who chose PSR-2, or blame whoever drafted PSR-2 who decided this. We never require other developers to use the same code style, so if you did not even contribute any abstract or final functions to the PocketMine source code, this is entirely none of your business.

Please also note that neither the government of Java Island nor its supervising government of Indonesia has enacted the law that fines whoever places `abstract` before a `public` method.

PR #815, opened 2017-04-11
Added getName() for trees, obsoletes #485.

Sandertv:
This pull request obsoletes #485.

Adds getName() for trees and a little more documentation for tree functions.


dktapps:
Note that this will stop working when translations are added for block name translation. This is also hardcoded to English.

Sandertv:
Ouh... That's correct...

dktapps:
Can't you use the class short name for this purpose? I'm not keen on this. I'm sure a little reflection and regex would do this job just fine from a plugin point of view.

Sandertv:
This isn't required anyways. It's easy enough to mirror this from a plugin point of view.

Issue #816, opened 2017-04-11
Cosmetic issue


dktapps:
https://forums.pmmp.io/profile-posts/2097/

PR #817, opened 2017-04-11
Removed redundant ZLIB_ENCODING_DEFLATE argument for compressed NBT decoding

JackNoordhuis:
### Description
Removes redundant compression argument from NBT::readCompressed() and
NBT:: readNetworkCompressed()

### Reason to modify
Addresses #816 

### Tests & Reviews
I haven't tested this but I don't see how it could break anything lol


Muqsit:
https://github.com/pmmp/PocketMine-MP/blob/api3/network/src/pocketmine/nbt/NBT.php#L426-#L432 ;)

SOF3 (approved):
Should not have any problems. Already corrects all usages.

PR #822, opened 2017-04-13
Improved cURL functions

SOF3:
Adds the `&$headers` and `&$httpCode` parameters to getURL() and postURL() and merged their code into `Utils::simpleCurl()`. `Utils::simpleCurl()` is also exposed as a public API method.

Regarding the order of parameters/return values, `httpCode, headers, body` is the order sent in HTTP, while `body, headers, httpCode` is the reverse of it (mnemonic reason), and in descending order of general usage frequency.
I would agree if the return value of `simpleCurl` should be reversed.

This change shall have 100% backwards compatibility as getURL() and postURL() are modified into backward-compatible wrappers for simpleCurl().


SOF3:
This branch will change in target and be replaced by another pull request.

PR #842, opened 2017-04-17
Remove movement hack

madecode15:
Mojang patched the issue where clients sent non stop position updates making this obsolete.


dktapps:
Incorrect. They did not fix anything; it was the change to a non-zero entity ID that fixed the spamming issue. Regardless, this conflicts with changes on api3/network, so I'm going to close this.

PR #844, opened 2017-04-17
Added title command and API for resetting title duration

TheDiamondYT1:
## What this adds
* Title Command
* Player->resetTitles()
* Player->addSubTitle(string)

If you need me to change anything please just ask.

Apparently any translations with the same key as in MCPE displays as `%ommands.(cmd).usage`. Probably a client bug, or because im using 1.0.5 and theres already been an update.

~~There's currently an issue where settings the subtitle using the command *doesnt* set it. Maybe we need a seperate method in the Player class to set the subtitle?~~

~~EDIT: We definitely need a seperate method to set subtitle, i'll add it now~~

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -3416,9 +3416,18 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	public function addTitle(string $title, string $subtitle = "", int $fadeIn = -1, int $stay = -1, int $fadeOut = -1){
 		$this->setTitleDuration($fadeIn, $stay, $fadeOut);
 		if($subtitle !== ""){
-			$this->sendTitleText($subtitle, SetTitlePacket::TYPE_SET_SUBTITLE);
+			$this->setSubTitle($subtitle);
 		}
 		$this->sendTitleText($title, SetTitlePacket::TYPE_SET_TITLE);
+	}
+	
+	/**
+	 * Sets the subtitle message, without a title.
+	 *
+	 * @param strint $subtitle
+	 */
+	public function setSubTitle(string $subtitle){
+	    $this->sendTitleText($subtitle, SetTitlePacket::TYPE_SET_SUBTITLE);
 	}
 
 	/**
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -3422,9 +3422,9 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	}
 	
 	/**
-	 * Sets the subtitle message, without a title.
+	 * Sets the subtitle message, without sending a title.
 	 *
-	 * @param strint $subtitle
+	 * @param string $subtitle
 	 */
 	public function addSubTitle(string $subtitle){
 	    $this->sendTitleText($subtitle, SetTitlePacket::TYPE_SET_SUBTITLE);
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -3438,6 +3438,15 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 		$pk->type = SetTitlePacket::TYPE_CLEAR_TITLE;
 		$this->dataPacket($pk);
 	}
+	
+	/**
+	 * Resets the title duration settings.
+	 */
+	public function resetTitles(){
+	    $pk = new SetTitlePacket();
+	    $pk->type = SetTitlePacket::TYPE_RESET_TITLE;
+	    $this->dataPacket($pk);
+	}
 
 	/**
 	 * Sets the title duration.
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -3416,7 +3416,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	public function addTitle(string $title, string $subtitle = "", int $fadeIn = -1, int $stay = -1, int $fadeOut = -1){
 		$this->setTitleDuration($fadeIn, $stay, $fadeOut);
 		if($subtitle !== ""){
-			$this->setSubTitle($subtitle);
+			$this->addSubTitle($subtitle);
 		}
 		$this->sendTitleText($title, SetTitlePacket::TYPE_SET_TITLE);
 	}
@@ -3426,7 +3426,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	 *
 	 * @param strint $subtitle
 	 */
-	public function setSubTitle(string $subtitle){
+	public function addSubTitle(string $subtitle){
 	    $this->sendTitleText($subtitle, SetTitlePacket::TYPE_SET_SUBTITLE);
 	}
 
--- a/src/pocketmine/command/defaults/TitleCommand.php
+++ b/src/pocketmine/command/defaults/TitleCommand.php
@@ -43,59 +43,54 @@ class TitleCommand extends VanillaCommand{
 		}
 
 		if(count($args) < 2){
-		    $sender->sendMessage(new TranslationContainer(TextFormat::RED . "%commands.generic.usage", [$this->usageMessage]));
+		    $sender->sendMessage(new TranslationContainer("commands.generic.usage", [$this->usageMessage]));
 		    return true;
 		}
 		
 		$player = $sender->getServer()->getPlayer($args[0]);
 		
 		if($player === null){
-		    $sender->sendMessage(new TranslationContainer(TextFormat::RED . "%commands.generic.player.notFound"));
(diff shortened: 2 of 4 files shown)


TheDiamondYT1:
If you wait i will push the title command aswell.

ImagicalGamer:
Might I suggest something other than title as the command name? It has ActionBar and Subtitle, the name may be misleading to the more stupid members of the community.

0x15f (approved):

jojoe77777:
> Might I suggest something other than title as the command name? It has ActionBar and Subtitle, the name may be misleading to the more stupid members of the community.

@ImagicalGamer it's the same as vanilla MCPE, blame shoghi

AppleDevelops:
Don't blame it on shoghi, it's like this on MCPC too

jojoe77777:
@AppleDevelops always blame shoghi

TheDiamondYT1:
@jojoe77777 @AppleDevelops @ImagicalGamer I just went onto MCPE, typed /title and looked at the autocompletions.

Matthww:
Since there already is a "AddTitle" is it not better to call the subtitle: "AddSubTitle" instead of "SendSubTitle" ?

dktapps:
@Matthww Subtitles won't work by themselves. You can't "add" one without a corresponding title. That function should only be used for _changing_ the subtitle by plugins.

Matthww:
@dktapps I know. Even though it doesn't make a difference. If you already called the title feature "AddTitle" why would you call the subtitle "SendSubTitle" instead of "AddSubTitle"?

dktapps:
My question is why he has made a separate method for that at all. addTitle() with parameters should be sufficient. It doesn't make sense to add a sub by itself because it won't work.

Matthww:
@dktapps Oh I get it. Why didn't you tag him instead. I was just giving him a suggestion.

dktapps:
@Matthww I misread the PR discussion slightly 😄

src/pocketmine/Player.php
```diff
@@ -3416,10 +3416,19 @@ public function kick($reason = "", $isAdmin = true){
...
-			$this->sendTitleText($subtitle, SetTitlePacket::TYPE_SET_SUBTITLE);
+			$this->addSubTitle($subtitle);
 		}
 		$this->sendTitleText($title, SetTitlePacket::TYPE_SET_TITLE);
 	}
+	
+	/**
+	 * Sets the subtitle message, without a title.
+	 *
+	 * @param strint $subtitle
```
dktapps: Typo

src/pocketmine/Player.php
```diff
@@ -3416,10 +3416,19 @@ public function kick($reason = "", $isAdmin = true){
...
+	 */
+	public function addSubTitle(string $subtitle){
```
dktapps: See my comments in the discussion about this method.

dktapps (dismissed):

src/pocketmine/Player.php
TheDiamondYT1: It has to be a seperate method because i called addTitle( "" , "subtitle"); and it showed a blank title with the subtitle, which shouldnt happen

dktapps: Ah, I see. My bad.

src/pocketmine/Player.php
Sandertv: That's not a typo, @dktapps. It's a combination of a string and an integer :D

dktapps:
Looks good, will re-review after I test. My local working directory is full of light-population code mess at the moment, so I can't test this immediately.

TheDiamondYT1:
@dktapps Maybe I could just add a `$immediate = false` to `addTitle()` then if its true, sender the title packet and if false it doesn't.

TheDiamondYT1:
Bump

dktapps:
I derped when playing with /title and attempted to use /title d duration instead of /title d times. A case statement would be more suitable.

src/pocketmine/command/defaults/TitleCommand.php
```diff
@@ -0,0 +1,97 @@
...
+		        return false;
+		    }
+		    $player->addActionBarMessage(implode(" ", array_slice($args, 2)));    
+		}elseif($args[1] === "times"){
+		    if(count($args) !== 5){
+		        $sender->sendMessage(new TranslationContainer(TextFormat::RED . "%commands.generic.usage", [$this->usageMessage]));
+		    
+		        return false;
+		    }
+		    $player->setTitleDuration($this->getInteger($sender, $args[3]), $this->getInteger($sender, $args[4]), $this->getInteger($sender, $args[5]));
```
dktapps: These argument indices are incorrect. Player is 0, target is 1, fade-in should be 2, stay-time 3, fade-out 4. This causes unexpected behaviour when trying to add titles to a player.

src/pocketmine/command/defaults/TitleCommand.php
```diff
@@ -0,0 +1,97 @@
...
+		}
+		
+		$player = $sender->getServer()->getPlayer($args[0]);
+		
+		if($player === null){
+		    $sender->sendMessage(new TranslationContainer(TextFormat::RED . "%commands.generic.player.notFound"));
+		    return true;
+		}
+		
+		if($args[1] === "clear"){
```
Muqsit: Maybe switch rather than multiple elseifs?

dktapps (changes requested):

src/pocketmine/command/defaults/TitleCommand.php
TheDiamondYT1: Yes, doing it.

src/pocketmine/command/defaults/TitleCommand.php
TheDiamondYT1: What do you mean? Player, target? Same thing?

dktapps: I meant subcommand, my bad. Don't know why I said that.

dktapps: To clarify, this occurs, and the duration number is used as fade-in, and the fade-out number is used as duration.
```
Notice: Undefined offset: 5 in src\pocketmine\command\defaults\TitleCommand.php on line 91
```

Matthww:
you're using spaces all over the place.

TheDiamondYT1:
@Matthww Just following what the **other** command classes structure is like.

TheDiamondYT1:
I will fix translations for all commands including this one in another PR.

dktapps:
Please use tabs instead of spaces, and strip trailing whitespace once you've finished making your changes.

TheDiamondYT1:
I thought i was using tabs, it looks fine to me. I will remove whitespace but for some reason github viewer sometimes shows extra whitespaces.

dktapps:
There are a mixture of tabs and spaces, if you use your cursor to highlight parts of the diff you'll see what I mean.

Matthww:
aee7bfdee8f8d8db9e338e7651b69d1d

TheDiamondYT1:
Yeah, well, - here's the thing. Im on an Android device because my computer has been acting up lately. I coded this whole PR on my device.

jasonwynn10:
🤦‍♂️ Doing PRs without an IDE is SO unprofessional. Let alone only using github's editor!

TheDiamondYT1:
@jasonwynn10 Im using an IDE, just not a PHP one :P

GitHub's editor doesnt work on my kindle fire.

dktapps:
@jasonwynn10 I wrote the new chunk implementation for the 1.0 update on my phone, is that unprofessional too? ;P

As long as the end result is OK, it doesn't matter too much how you got there. Just as long as everyone understands that it is _wiser_ to use a computer and a smart IDE. I discourage use of mobile devices though.

TheDiamondYT1:
@dktapps I use a Terminal Emulator called Termux. It has the full git in its package manager.

And i use a Java IDE (😂) called AIDE for php.

dktapps:
Termux is a good solution, yea. I use it a lot.

TheDiamondYT1:
@dktapps What IDE do you use?

*Apologizes for being off topic*

dktapps:
@TheDiamondYT1 I use QuickEdit for my editing needs, I don't tend to write actual code on mobile these days. I don't like AIDE and haven't found any other suitable editors or IDEs.

TheDiamondYT1:
Bump

dktapps:
Will review tomorrow morning.

dktapps:
I had to reformat this locally, merged it to master as https://github.com/pmmp/PocketMine-MP/commit/653b6b55a9ab44d24e9402d0a253b3d394c62908

TheDiamondYT1:
Yeah sorry for the formatting. I edited AIDE to use spaces instead of tabs so i dont get a mixture in future PRs.

SalmonDE:
And you didn't add `Server::broadcastTitle()`? Is there a reason why this wasn't added?

PR #846, opened 2017-04-18
Optimized spawn code, also gets rid of a misleading variable name.

ghost:
$chunkLoadCount was just being used for spawning, and it did not accurately represent the amount of chunks loaded.

This PR will count the $usedChunks array to check how much chunks are being used, also moved $this->spawned === false as the first check as a microoptimization.


madecode15:
I'm going to close this for now, I see multiple improvements that can be made to sendChunk() and would rather put them all in one PR. Waiting for the network branch to be merged first to avoid more conflicts.

PR #853, opened 2017-04-19
Remove redundant null check in PluginCommand

SalmonDE:
# Introduction

This pr removes the null check in PluginCommand::setExecutor() because $executor won't be null in any case.

## Relevant issues

No relevant issues (atleast I think so)

# Changes
## API changes

No API changes -> same result as before.
## Behavioural changes

I doubt there will be any performance improvements, no.
# Backwards compatibility

Yes, it is fully compatible.

# Follow-up

No actions suggested from my side.

# Tests

You will need a plugin which manually registers a PluginCommand and sets a CommandExecutor for it. Sorry for not including this here.

The result should be, that $this->executor returns the CommandExecutor mentioned before.

### DIFF
--- a/src/pocketmine/command/PluginCommand.php
+++ b/src/pocketmine/command/PluginCommand.php
@@ -70,7 +70,7 @@ class PluginCommand extends Command implements PluginIdentifiableCommand{
 	 * @param CommandExecutor $executor
 	 */
 	public function setExecutor(CommandExecutor $executor){
-		$this->executor = ($executor != null) ? $executor : $this->owningPlugin;
+		$this->executor = $executor;
 	}
 
 	/**


iPocket:
I don't think this is a bug fix.

dktapps:
@iPocket well, what else am I supposed to label it?

SalmonDE:
@iPocket Actually in this case we should blame 64FF00 instead 839f6621eb39a9337aab3a4ac4a2e803a81a6816

Also, I agree, this isn't a bug fix. :)

iPocket:
@dktapps PR: Optimization.

dktapps:
My question here is whether this is behaviour is intended and the author forgot to make the CommandExecutor argument nullable, or whether this PR is indeed be the correct approach. As seen [here](https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/plugin/Plugin.php#L34), plugins themselves are CommandExecutors. Comments?

iPocket:
But why?

dktapps:
@iPocket so you could pass null to reset the executor to its default owning plugin?

iPocket:
I meant why does it extend it.

dktapps:
I assume so that plugin authors can implement the onCommand() method into their plugins without having to add the `implements` manually.

SalmonDE:
@dktapps Can't you just pass the plugin instance then?

dktapps:
You have a point.

PR #861, opened 2017-04-22
Fixed semicolon typo

matcracker:
## Introduction

Just fixed a semicolon error in the code.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -1474,7 +1474,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 				$this->checkNearEntities($tickDiff);
 			}
 
-			$this->speed = ($to->subtract($from))->divide($tickDiff);
+			$this->speed = ($to->subtract($from)->divide($tickDiff));
 		}elseif($distanceSquared == 0){
 			$this->speed = new Vector3(0, 0, 0);
 		}


Sandertv:
Actually, this is intended

matcracker:
Why? o_O

TheDiamondYT1:
@matcracker Because Vector3::subtract() returns a Vector3

dktapps:
Those extra parentheses aren't actually needed at all, but I put them in to clarify that the subtraction should always occur before the division. With method calls this obviously doesn't make any difference, but with operator overloading this would be required. It's merely a cosmetic clarification.

TheDiamondYT1:
@dktapps I thought they were reqiired, because without them you would be calling `divide()` on a function? Or is that just Java.

dktapps:
No, you'd be calling divide() on the _return_ value of the previous method.

TheDiamondYT1:
@dktapps 
In java, doing
```java
foo().bar();
```
would return an error saying saying you cant call bar() on void.

However 
```java
(foo()).bar));
```
Works.

Sandertv:
Again, this is not Java!

matcracker:
_**Error**_
$this->speed = ($to->subtract($from))->divide($tickDiff); 

_**Fix 1**_
$this->speed = ($to->subtract($from)->divide($tickDiff)); 

_**Fix 2**_
$subtract = $to->subtract($from);
$this->speed = $subtract->divide($tickDiff);

So, is it a PHP issue or PHPStorm stupid? 📦

jojoe77777:
> So, is it a PHP issue or PHPStorm stupid? 📦

Or you?

There's nothing wrong with `$this->speed = ($to->subtract($from))->divide($tickDiff);`

matcracker:
It's PHPStorm that it give me this error, I know that is correct ...

PR #864, opened 2017-04-24
Return class specification

aquaminer:
Result:

### DIFF
--- a/src/pocketmine/event/server/DataPacketReceiveEvent.php
+++ b/src/pocketmine/event/server/DataPacketReceiveEvent.php
@@ -39,7 +39,9 @@ class DataPacketReceiveEvent extends ServerEvent implements Cancellable{
 	public function getPacket(){
 		return $this->packet;
 	}
-
+	 /**
+      *  @return Player
+      */
 	public function getPlayer(){
 		return $this->player;
 	}


src/pocketmine/event/server/DataPacketReceiveEvent.php
```diff
@@ -39,8 +39,10 @@ public function __construct(Player $player, DataPacket $packet){
 	public function getPacket(){
 		return $this->packet;
 	}
-
+	 /**
```
TheDiamondYT1: Terrible formatting for such a simple fix...

Sandertv (changes requested):
I don't really see the point in only adding a PHP doc. Might be better if you'd add Typehints.

Muqsit:
Why a comment and not strict mode?

opkiler22789:
@Muqsit, Sorry, but I do not know these little things

madecode15:
@Muqsit @opkiler22789 How are coding a plugin if you don't know how to type hint?

TheDiamondYT1:
@madecode15 its simple lol. But typehinting is good

TheDiamondYT1:
Im about to make a pr to typehint all events

SOF3 (changes requested):
Considering that network branch is being rewritten, it's not meaningful to add this here right now. Moreover, PhpStorm can detect the field type from the constructor assignment of the field.

PR #865, opened 2017-04-24
Typehint and document more events, obseletes #864

TheDiamondYT1:
## Introduction

Typehint and document plugin and server events.

### Relevant issues
* Obseletes #864

If you need any more events typehinting i'll do it


dktapps:
This will be a crippling API change for many plugins which have their own events.

TheDiamondYT1:
@dktapps Oh. I didn't even consider that.

PR #887, opened 2017-05-03
Call PlayerInteractEvent (ItemFrameDropItemPacket)

Muqsit:
## Introduction

Item frames are one of the many essentials that contribute to decoration. In PocketMine-MP, there's no way to stop players from triggering ItemFrameDropItemPacket (other than using DataPacketReceiveEvent, of course; which is like the worst-but-the-only-possible decision).
To make it a possibility to cancel ItemFrameDropItemPacket from being triggered, calling PlayerInteractEvent would be a suitable fix.

### Relevant issues

None.
### API changes

I don't think so, as this can be considered as a "fix".
### Behavioural changes

ItemFrameDropItemPacket now triggers PlayerInteractEvent.

## Backwards compatibility

Compatible.
## Follow-up

None.
## Tests

The following code will stop ItemFrames from dropping the item.
```
    if(
        $event->getBlock()->getId() === \pocketmine\block\BlockIds::ITEM_FRAME_BLOCK and
        $event->getAction() === PlayerInteractEvent::LEFT_CLICK_BLOCK
    ){//PlayerInteractEvent::RIGHT_CLICK_BLOCK => When player puts an item in the item frame.
        $event->setCancelled();
    }
}
```

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -3241,7 +3241,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 
 		$tile = $this->level->getTile($this->temporalVector->setComponents($packet->x, $packet->y, $packet->z));
 		if($tile instanceof ItemFrame){
-			$ev = new PlayerInteractEvent($this, $this->inventory->getItemInHand(), $tile->getBlock(), $tile->getBlock()->getDamage(), PlayerInteractEvent::LEFT_CLICK_BLOCK);
+			$ev = new PlayerInteractEvent($this, $this->inventory->getItemInHand(), $tile->getBlock(), 5 - $tile->getBlock()->getDamage(), PlayerInteractEvent::LEFT_CLICK_BLOCK);
 			$this->server->getPluginManager()->callEvent($ev);
 
 			if($this->isSpectator()){
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -3241,7 +3241,14 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 
 		$tile = $this->level->getTile($this->temporalVector->setComponents($packet->x, $packet->y, $packet->z));
 		if($tile instanceof ItemFrame){
+			$ev = new PlayerInteractEvent($this, $this->inventory->getItemInHand(), $tile->getBlock(), $tile->getBlock()->getDamage(), PlayerInteractEvent::LEFT_CLICK_BLOCK);
+			$this->server->getPluginManager()->callEvent($ev);
+
 			if($this->isSpectator()){
+				$ev->setCancelled();
+			}
+
+			if($ev->isCancelled()){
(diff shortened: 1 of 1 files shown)


thebigsmileXD:
I'd actually like to move it into another event. Placing item into and taking out from the frame *cough see https://github.com/ClearSkyTeam/PocketMine-MP/blob/cs-1.0/src/pocketmine/event/block/ItemFrameDropItemEvent.php cough*

Muqsit:
@thebigsmileXD That logic can basically apply to every tile then. There should be a whole new BlockEntityInteractEvent in that case.

dktapps:
I disagree. This was already discussed elsewhere. Adding a specific event for this doesn't make any sense and just complicates things for no reason. The basic interaction is fine. Plugin developers shouldn't need to give any consideration towards the fact that tiles are a thing (tiles are just blocks with some extra data, after all).

With regards to this pull request, I'm happy with this, minus testing on my own part. Will test later.

thebigsmileXD:
Well even if, right and leftclick need to be handled, i experienced issues that itemframes still rotate client side

Muqsit:
@thebigsmileXD I believe that can be handled too. It triggers PlayerInteractEvent::RIGHT_CLICK_BLOCK.

thebigsmileXD:
Isn't ItemFrameDropItemPacket (drop) left click?

Muqsit:
Yeah but weren't you talking about item frames rotating client-side.

src/pocketmine/Player.php
```diff
@@ -3241,7 +3241,14 @@ public function handleItemFrameDropItem(ItemFrameDropItemPacket $packet) : bool{
 
 		$tile = $this->level->getTile($this->temporalVector->setComponents($packet->x, $packet->y, $packet->z));
 		if($tile instanceof ItemFrame){
+			$ev = new PlayerInteractEvent($this, $this->inventory->getItemInHand(), $tile->getBlock(), $tile->getBlock()->getDamage(), PlayerInteractEvent::LEFT_CLICK_BLOCK);
```
SOF3: I'm not sure if LEFT_CLICK_BLOCK is appropriate. Usually, LEFT_CLICK_BLOCK implies start breaking a block, which may lead to confusion.

dktapps: @SOF3 not really... it _is_ a left-click, exactly as it says. People who misinterpret that, well, it's their own problem.

dktapps: I don't think the the damage value as clicked-face here is correct. See https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/block/ItemFrame.php#L111-L116 for placement values.

Other than that, it works fine as a resolution for this problem.

Muqsit: Would
```
$face = 5 - $damage;
```
be a better fix or should I just copy $faces array and then search the array?

dktapps: Sure, neat solution.

PR #890, opened 2017-05-05
Added addEnchant() function

DaPigGuy:
## Introduction

This lets plugins add custom enchants.

### Relevant issues

None
### API changes

Added addEnchant() function & construct function is now public

### Behavioural changes

No.
## Backwards compatibility

No.

## Follow-up

 None

## Tests

Try to add an enchant.
 `        Enchantment::addEnchantment(new Enchantment(100, "Lifesteal", self::RARITY_RARE, self::ACTIVATION_HELD, self::SLOT_SWORD), 100);
        var_dump(Enchantment::getEnchantment(100));:`

### DIFF
--- a/src/pocketmine/item/enchantment/Enchantment.php
+++ b/src/pocketmine/item/enchantment/Enchantment.php
@@ -123,8 +123,8 @@ class Enchantment{
 	 * @return Enchantment|null
 	 */
 	public static function getEnchantmentByName(string $name){
-		if(defined(Enchantment::class . "::TYPE_" . strtoupper($name))){
-			return self::getEnchantment(constant(Enchantment::class . "::TYPE_" . strtoupper($name)));
+		if(defined(Enchantment::class . "::" . strtoupper($name))){
+			return self::getEnchantment(constant(Enchantment::class . "::" . strtoupper($name)));
 		}
 		return null;
 	}
--- a/src/pocketmine/item/enchantment/Enchantment.php
+++ b/src/pocketmine/item/enchantment/Enchantment.php
@@ -143,7 +143,7 @@ class Enchantment{
 	 * @param int $activationType
 	 * @param int $slot
 	 */
-	private function __construct(int $id, string $name, int $rarity, int $activationType, int $slot){
+	public function __construct(int $id, string $name, int $rarity, int $activationType, int $slot){
 		$this->id = $id;
 		$this->name = $name;
 		$this->rarity = $rarity;
--- a/src/pocketmine/item/enchantment/Enchantment.php
+++ b/src/pocketmine/item/enchantment/Enchantment.php
@@ -104,6 +104,19 @@ class Enchantment{
 		return null;
 	}
 
+    /**
+     * @param Enchantment $enchantment
+     * @param int $id
+     * @return bool
+     */
+    public static function addEnchantment(Enchantment $enchantment, int $id){
+		if(self::getEnchantmentByName($enchantment->getName())){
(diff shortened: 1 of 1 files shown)


buchwasa (approved):

Muqsit:
This is unneeded. If plugins wanted to, they can already do that, just that they wouldn't have to use the PocketMine functions.
```
public function hasEnchantment(Item $item, int $id){
    if(isset($item->getNamedTag()->ench)){
        foreach($item->getNamedTag()["ench"] as $ench){
            if($ench["id"] === $id){
                return true;
            }
        }
    }
}

public function addEnchantment(Item $item, int $id, int $level){
    $nbt = $item->getNamedTag() ?? new CompoundTag("", []);
    $nbt->ench = $nbt->ench ?? new ListTag("ench", []);
    foreach($nbt["ench"] as $k => $en){
        if($en["id"] === $id){
            unset($nbt->ench->{$k});
            break;
        }
    }
    $nbt->ench->{count($nbt->ench)} = new CompoundTag("", [
        new ShortTag("id", $id),
        new ShortTag("lvl", $level)
    ]);
    $item->setNamedTag($nbt);
}
```

thebigsmileXD:
@Muqsit Just because you maybe did it this way, it doesn't use actual api methods. 
@MCPEPIG The move is understandable, i did that locally before. Almost every class with id's have a way to register new Classes (Network, Entity). This way they are directly handled my the core instead of plugins. +approve for the idea, needs testing

Muqsit:
@thebigsmileXD 
1. It isn't mandatory to use API methods in PHP. Since when was this a rule?
2. What inventory bugs? I don't think PMMP is bothered about Win10  inventory management as much as you think. (Besides, there are no enchantment related in bugs plus shoghi is gonna make inventories on servers and realms be managed by the game).

src/pocketmine/item/enchantment/Enchantment.php
```diff
@@ -104,14 +104,27 @@ public static function getEnchantment(int $id){
 		return null;
 	}
 
+    /**
+     * @param Enchantment $enchantment
+     * @param int $id
+     * @return bool
+     */
+    public static function addEnchantment(Enchantment $enchantment, int $id){
```
Muqsit: R.I.P formatting, use tabs.

thebigsmileXD:
@Muqsit you wouldn't believe how much it actually causes issues on classic GUI. Also shoghi actually is busy tweeting bananas.

Muqsit:
@thebigsmileXD What are you talking about? Looks like you haven't ever used pmmp. There are no enchantment-related inventory issues in pmmp.

PMMPFTW:
This is useless. The cause of this PR is mainly focused on making it possible to add custom enchantments. Stop trying to act like other "SPOONS". Just because a crappy tess***** spoon did this doesn't mean pmmp has to do it this way too.
Speaking about custom, you should focus more in publicizing the private InventoryType constructor as it would allow us to register custom inventories which would allow creating plugins that could add fully working enchantment tables, brewing stands etc.
This is garbage. Stop copying forks.
And @thebigsmileXD, we already know your crappy fork's fate. Just stop trying to garabagize pmmp.

dktapps:
This conflicts with local changes to enchantments.

dktapps:
@pmmpftw It would be appreciated if you didn't use pull request discussions to talk about spoons or slag off others' work just for the sake of being right. At least even make sure you had your facts correct. Locking.

PR #895, opened 2017-05-07
Kick the player who's tryna login with the username!

Assblocker:
## Introduction

This is necessary because if not, then the other logged in player gets kicked an he gets mad and starts complaining!
### API changes
No

### Behavioural changes
No

## Backwards compatibility
No

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -1764,13 +1764,13 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 
 		foreach($this->server->getOnlinePlayers() as $p){
 			if($p !== $this and $p->iusername === $this->iusername){
-				if($p->kick("logged in from another location") === false){
+				if($this->kick("logged in from another location") === false){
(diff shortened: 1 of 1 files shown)


KnownUnown:
This can be argued the other way as well. The currently logged in player may not be the real one, and the actual player who owns the username may be trying to join. The whole matter should be handled by authentication plugins anyway with PlayerPreLoginEvent.

SOF3:
A better mechanism to tackle the issue can be implemented in a plugin, e.g. [checking IP](https://github.com/PEMapModder/HereAuth/blob/master/src/HereAuth/EventRouter.php#L89-L125)

thebigsmileXD:
There are tons of plugins that do that, and this was a thing in pm once.. maybe it got removed.

dktapps:
This matches to vanilla behaviour. If you want to alter that, use a plugin.

dktapps:
also, who the hell is going to merge a PR from someone with a name like that... 🤦‍♂️

Assblocker:
Tesseract did XD

PR #898, opened 2017-05-07
Use json_decode for effects data

TheDiamondYT1:
## Introduction

Matches [this commit](https://github.com/pmmp/PocketMine-MP/commit/fe4322bea957b422ea8596712ccaf022a388c9b7#diff-c0f9e4f47d9a1ae7abba9ce1d681fd89R86) (OCD)

### DIFF
--- a/src/pocketmine/entity/Effect.php
+++ b/src/pocketmine/entity/Effect.php
@@ -64,20 +64,20 @@ class Effect{
 			throw new \RuntimeException("Effects data could not be read");
 		}
 
-		foreach($data as $name => $data){
-			$color = hexdec(substr($data["color"], 3));
+		foreach($data as $effectName => $effectData){
+			$color = hexdec(substr($effectData["color"], 3));
 			$r = ($color >> 16) & 0xff;
 			$g = ($color >> 8) & 0xff;
 			$b = $color & 0xff;
-			self::registerEffect($name, new Effect(
-				$data["id"],
-				$data["name"],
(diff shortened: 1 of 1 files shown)


src/pocketmine/entity/Effect.php
```diff
@@ -60,9 +59,12 @@ class Effect{
...
 
 	public static function init(){
-		$config = new Config(\pocketmine\PATH . "src/pocketmine/resources/effects.json", Config::JSON, []);
+		$data = json_decode(file_get_contents(\pocketmine\PATH . "src/pocketmine/resources/effects.json"), true);
+		if(!is_array($data)){
+			throw new \RuntimeException("Effects data could not be read");
+		}
 
-		foreach($config->getAll() as $name => $data){
+		foreach($data as $name => $data){
```
dktapps: Bad move.

dktapps (changes requested):

src/pocketmine/entity/Effect.php
dktapps: (lucky PhpStorm spotted this, I almost merged this without looking >_>)

TheDiamondYT1: What is wrong?

TheDiamondYT1: Oh..im so dumb

TheDiamondYT1:
Sorry about that @dktapps. Im actually using IntelliJ... :rofl:

dktapps (approved):

PR #903, opened 2017-05-09
Added "extensions" attribute in plugin.yml with version checking

SOF3:
## Introduction

With the prevalence of third-party modifications of PocketMine-MP, the set of available PHP extensions in a PocketMine runtime is no longer standardized; the original set of available PHP extensions is messed up as well.

This pull request introduces the `extensions` attribute, which will make sure the current PHP runtime has the required extensions with the expected versions.

### API changes
The `PluginDescription::getRequiredExtensions()` method is added to return an array whose keys are extension names and values are extension version constraints (a string array).

See the [Tests](#tests) section for possible ways of declaring extensions.

### Behavioural changes

Plugins will fail to load if the extensions are missing or have an incorrect version, or if the extension versions are declared syntactically wrong.

## Backwards compatibility

Plugins that do not have the `extensions` attribute in their plugin.yml will be considered as requiring no extensions (to be checked), identical to this:

```yaml
extensions: []
```

## Follow-up

Nil

## Tests

The following fragments of `plugin.yml` have been tested and have the expected behaviour of crashing or passing, tested on a server with pthreads 3.1.6 and without the ncurses extension.

The version constraint is the concatenation of two strings, _comparator_ and *compared_version*. The comparator should be one of `<`, `lt`, `<=`, `le`, `>`, `gt`, `>=`, `ge`, `==`, `=`, `eq`, `!=`, `<>`, `ne`. Then this condition is checked whether to be true:

```
extension_version {comparator} compared_version
```

### No extensions attribute
```yaml
```
Loads correctly.

### As a simple string, extension present
```yaml
extensions: pthreads
```
Loads correctly.

### As a simple string, extension missing
```yaml
extensions: ncurses
```

Fails to load:
```
[19:06:24] [Server thread/ERROR]: Required extension ncurses not loaded
[19:06:24] [Server thread/CRITICAL]: Could not load plugin 'TestExtensions'
```

### As a linear array, one extension present and one extension missing
```yaml
extensions: [pthreads, ncurses]
```

Fails to load:
```
[19:06:58] [Server thread/ERROR]: Required extension ncurses not loaded
[19:06:58] [Server thread/CRITICAL]: Could not load plugin 'TestExtensions'
```

### As a map, using asterisk, extension present
**Note: `*` in YAML does not get parsed as a string. Use `"*"` to escape.
```yaml
extensions:
  pthreads: "*"
```

Loads correctly.

### As a map, using asterisk, extension absent
```yaml
extensions:
  ncurses: "*"
```

Fails to load:
```
[19:10:56] [Server thread/ERROR]: Required extension ncurses not loaded
[19:10:56] [Server thread/CRITICAL]: Could not load plugin 'TestExtensions'
```

### As a map, using a string as a true version constraint
**Note: `>` is a special symbol in YAML. _Always_ Quote the version constraint.**
```yaml
extensions:
  pthreads: ">=3"
```

Loads correctly.

### As a map, using a string as a false version constraint
```yaml
extensions:
  pthreads: "<>3.1.6"
```

Fails to load:
```
[19:21:39] [Server thread/ERROR]: Required extension pthreads has an incompatible version (3.1.6 not <>3.1.6)
[19:21:39] [Server thread/CRITICAL]: Could not load plugin 'TestExtensions'
```

### As a map, using an empty string as version constraint
This test shows what happens if `>` is used and the version gets YAML-parsed incorrectly into an empty string.
```yaml
extensions:
  pthreads: >2
```

Fails to load:
```
[19:14:33] [Server thread/ERROR]: One of the extension version constraints of pthreads is empty. Consider quoting the version string in plugin.yml
[19:14:33] [Server thread/CRITICAL]: Could not load plugin 'TestExtensions'
```

### As a map, using two true version constraints
```yaml
extensions:
  pthreads:
    - ">=3"
    - "<4"
```

Loads correctly.

### As a map, using one true version constraint and one false version constraint
```yaml
extensions:
  pthreads:
    - ">=3"
    - "<>3.1.6"
```

Fails to load:
```
[19:16:20] [Server thread/ERROR]: Required extension pthreads has an incompatible version (3.1.6 not <>3.1.6)
[19:16:20] [Server thread/CRITICAL]: Could not load plugin 'TestExtensions'
```

### DIFF
--- a/src/pocketmine/plugin/PluginDescription.php
+++ b/src/pocketmine/plugin/PluginDescription.php
@@ -164,7 +164,7 @@ class PluginDescription{
 	/**
 	 * @return array
 	 */
-	public function getExtensions(){
+	public function getRequiredExtensions(){
 		return $this->extensions;
 	}
 
@@ -173,7 +173,7 @@ class PluginDescription{
 	 *
 	 * @throws PluginException if there are required extensions missing or have incompatible version, or if the version constraint cannot be parsed
 	 */
-	public function checkExtensions(){
+	public function checkRequiredExtensions(){
 		foreach($this->extensions as $name => $versionConstrs){
 			if(!extension_loaded($name)){
 				throw new PluginException("Required extension $name not loaded");
--- a/src/pocketmine/plugin/PluginDescription.php
+++ b/src/pocketmine/plugin/PluginDescription.php
@@ -27,6 +27,7 @@ class PluginDescription{
 	private $name;
 	private $main;
 	private $api;
+	private $extensions = [];
 	private $depend = [];
 	private $softDepend = [];
 	private $loadBefore = [];
@@ -74,6 +75,17 @@ class PluginDescription{
 
 		if(isset($plugin["depend"])){
 			$this->depend = (array) $plugin["depend"];
+		}
+		if(isset($plugin["extensions"])){
+			$extensions = (array) $plugin["extensions"];
+			$isLinear = $extensions === array_values($extensions);
+			foreach($extensions as $k => $v){
+				if($isLinear){
+					$k = $v;
+					$v = "*";
+				}
+				$this->extensions[$k] = is_array($v) ? $v : [$v];
+			}
 		}
 		if(isset($plugin["softdepend"])){
 			$this->softDepend = (array) $plugin["softdepend"];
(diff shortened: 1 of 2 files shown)


thebigsmileXD:
As i see, this would for example let a graphical converting plugin check for GD? Neat.

SOF3:
If this gets merged and everyone likes the version comparison mechanism, I might make the same for plugins.

thebigsmileXD:
Could you maybe also add an update url and make updating plugins buildt in?

src/pocketmine/plugin/PluginDescription.php
```diff
@@ -152,6 +164,50 @@ public function getCommands(){
 	/**
 	 * @return array
 	 */
+	public function getExtensions(){
```
dktapps: I feel this may be ambiguous... maybe `getRequiredExtensions()`?

DaPigGuy (approved):

SOF3:
I'm not sure if this should be called `extensions` or `extension`, since it is `depend` not `depends` for dependency plugins.

dktapps:
just call it `extensions` and be done with it, forget your OCD.

TheDiamondYT1:
@dktapps Have you ever had OCD? You cant just forget it.

SOF3:
This is not a problem of OCD. This is a problem of consistency.
However, it would be a better idea to change/add `depend` to `depends` (same with `author`/`authors`), changing them all to [prefer plurals, both string and string[] accepted] rather than creating even more inconsistency.

dktapps (approved):
Looks great, I meant to do this months ago myself.

src/pocketmine/plugin/PluginDescription.php
```diff
@@ -75,6 +76,17 @@ private function loadMap(array $plugin){
...
 		}
+		if(isset($plugin["extensions"])){
+			$extensions = (array) $plugin["extensions"];
+			$isLinear = $extensions === array_values($extensions);
+			foreach($extensions as $k => $v){
+				if($isLinear){
+					$k = $v;
+					$v = "*";
+				}
+				$this->extensions[$k] = is_array($v) ? $v : [$v];
```
dktapps: hand stuck in toaster...

can we assume the type of `extensions` to be `string[][]`?

src/pocketmine/plugin/PluginDescription.php
```diff
@@ -149,6 +161,50 @@ public function getCommands(){
...
+	 *
+	 * @throws PluginException if there are required extensions missing or have incompatible version, or if the version constraint cannot be parsed
+	 */
+	public function checkRequiredExtensions(){
+		foreach($this->extensions as $name => $versionConstrs){
+			if(!extension_loaded($name)){
+				throw new PluginException("Required extension $name not loaded");
+			}
+
+			if(!is_array($versionConstrs)){
```
dktapps: from what I can understand from the code that loads this information, this condition will always be false, but because of the lack of proper type annotation, it's non-obvious.

PR #905, opened 2017-05-09
Doors

geraldebberink:
## Introduction
I (actually my children) found that they could not use doors. (They spawned in an birch area). 

## Changes
Added the doors to the best of my knowledge. Everything is based upon the WoodDoor block class and the WoodenDoor item class.

The problem for me personally is that I do not know how to creat the phar file, so I cannot confirm everything working.

### DIFF
--- a/src/pocketmine/block/AcaciaDoor.php
+++ b/src/pocketmine/block/AcaciaDoor.php
@@ -50,7 +50,7 @@ class AcaciaDoor extends Door{
 
 	public function getDrops(Item $item){
 		return [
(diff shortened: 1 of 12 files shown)


dktapps:
This is already done on a development branch.

geraldebberink:
I checked all the branches before. Is there a hidden one?
Personally I do not like to duplicate work.

dktapps:
A rewrite has taken place on the `api3/blocks` for exactly that reason. The changes will be merged to master at a later date (no specific date can be given).

PR #910, opened 2017-05-12
Fix issue with colored message on some languages

Johnmacrocraft:
## Introduction

On some languages, there is a small issue on colored message.

Example:
In this example, italic is color.
If there's a message Testing _{%0}_ (English), it'll be shown as Bla _{%0} Hi_ (Some languages).

So this pull request fixes it.

### API changes

No

### Behavioural changes

No

## Backwards compatibility

Yes

### DIFF
--- a/src/pocketmine/Achievement.php
+++ b/src/pocketmine/Achievement.php
@@ -107,7 +107,7 @@ abstract class Achievement{
 
 	public static function broadcast(Player $player, $achievementId){
 		if(isset(Achievement::$list[$achievementId])){
-			$translation = new TranslationContainer("chat.type.achievement", [$player->getDisplayName(), TextFormat::GREEN . Achievement::$list[$achievementId]["name"]]);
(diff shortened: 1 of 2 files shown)


TheDiamondYT1:
What languages would this be an issue with? It can be reproduced easier if you tell us.

Sandertv:
Yeah I get what he means, for example if you give a player an item with the /give command with a custom name, it would turn the rest of the message to that colour as well.

Johnmacrocraft:
@TheDiamondYT1
Achievement: Korean, Arabic (Not sure), Dutch, German, Hungarian, Japanese, Turkish
Server startup: Korean, Chinese Traditional, Japanese, Turkish
@Sandertv Yeah, that's what I mean.

Muqsit:
Why not TextFormat::RESET?

Johnmacrocraft:
@Muqsit Changed it according to your suggestion.

AppleDevelops:
@TheDiamondYT1 We like to change things to make other English speaking countries annoyed

Issue #152, opened 2016-12-06
Unsafe excess use of Level->getBlock() flooding block caches


dktapps:
### Issue description
If something spams Level->getBlock(), the block cache will fill up very rapidly. Since it is only cleared every 5 seconds or so, if something nukes it inside those 5 seconds the server memory will fill up very rapidly and crash of an Out Of Memory Error.

Issues such as iTXTech/Genisys#1788 result from the lack of proper cache size limiting. This is not the root cause of such issues (no doubt plugins misusing getBlock()) but a limitation should be imposed on the allowed size of the caches to prevent this happening.

### Steps to reproduce the issue

1. Spam getBlock(), maybe with a world edit plugin or something
2. Watch the server crash

Easiest to do if you reduce the memory limit.

### Proposed actions
Make the chunk- and block-caches have a fixed size and throw exceptions should the size attempt to exceed these limits.

### OS and versions

* PocketMine-MP: https://github.com/pmmp/PocketMine-MP/commit/5d16ecc0034ed2262ec2f95ccc0e7bbca17f24d5 , but irrespective of version this happens
* PHP: 7.0.13
* Server OS: irrelevant
* Game version: PE/Win10 (delete as appropriate)

### Crashdump, backtrace or other files

iTXTech/Genisys#1788, iTXTech/Genisys#1302

Sandertv:
I'm working on a worldediting plugin, and I'm trying to implement an undo command. What amount of using Level->getBlock() do you think will be maximum? If it is quite small, are there any alternatives for this?

dktapps:
Shit, it seems this can happen with a vanilla installation... :O

dktapps:
Related to #231

dktapps:
```
Fatal error: Allowed memory size of 1073741824 bytes exhausted (tried to allocate 67108864 bytes) in C:\Users\User\Downloads\server\src\pocketmine\level\Level.php on line 1266

Call Stack:
    0.0039     360136   1. {main}() C:\Users\User\Downloads\server\src\pocketmine\PocketMine.php:0
    0.3766     549968   2. pocketmine\Server->__construct() C:\Users\User\Downloads\server\src\pocketmine\PocketMine.php:492
   10.1355   17456544   3. pocketmine\Server->start() C:\Users\User\Downloads\server\src\pocketmine\Server.php:1599
[...]

PR #915, opened 2017-05-13
Removed masks for Y coordinates, throw exception for block hash Y out of bounds, fixed #914

### LINKED ISSUE #914: Water flow wraps around to the top of the world when falling into the void
### Issue description

 Hey guys! I just found some crazy bug that needs to be fixed.. so if water fall to void it will spawn back to sky.. i got some photos if you do not understand what i mean..

1. ... 
2. ...

### OS and versions

* PocketMine-MP: PocketMine-MP 1.6.2dev-76
* PHP: 3.0.0
* Server OS: Minecraft: PE v1.0.7.0 (protocol version 107)
* Game version: PE/Win10 (delete as appropriate)

http://imgur.com/a/jEpyE
http://imgur.com/a/0u4eC
* ...

dktapps:
## Introduction
These masks produced many unexpected behaviours relating to Y coordinates out of bounds, like #914.

This PR carries changes required to fix this issue, along with several others. Adding an exception throw for out-of-bounds Y coordinates for block hashes uncovered several wraparound problems, such as:

- Flying below y=0 would cause block collision checks for blocks up at y=~250
- Breaking blocks at y=0 would trigger block updates at the top of the map.

This may also cause other crashy issues which I haven't identified yet, so this needs careful testing before merging.

### Relevant issues
- fixes #914

### API changes
- Added `addToCache` parameter to `Level->getBlock()` to allow plugin developers to choose to not cache the returned block object if they are making frequent requests to the method. This may mitigate issues with block-cache issues like #152 .

### Behavioural changes
- An exception will now be thrown when attempting to create a block coordinate hash with an invalid Y coordinate.
- bugs fixed above

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -269,7 +269,7 @@ class Level implements ChunkManager, Metadatable{
 	}
 
 	public static function blockHash(int $x, int $y, int $z){
-		if(($y & ~Level::Y_MASK) !== 0){
+		if($y < 0 or $y >= Level::Y_MAX){
 			throw new \InvalidArgumentException("Y coordinate $y is out of range!");
 		}
 		return (($x & 0xFFFFFFF) << 36) | (($y & Level::Y_MASK) << 28) | ($z & 0xFFFFFFF);
@@ -1098,8 +1098,8 @@ class Level implements ChunkManager, Metadatable{
 		$pos = $pos->floor();
 
 		for($i = 0; $i <= 5; ++$i){
-			if($this->isInWorld($s = $pos->getSide($i))){
-				$this->neighbourBlockUpdateQueue->enqueue(Level::blockHash($pos->x, $pos->y, $pos->z));
+			if($this->isInWorld($side = $pos->getSide($i))){
+				$this->neighbourBlockUpdateQueue->enqueue(Level::blockHash($side->x, $side->y, $side->z));
 			}
 		}
 	}
@@ -1328,6 +1328,7 @@ class Level implements ChunkManager, Metadatable{
 		$pos = $pos->floor();
 
 		$fullState = 0;
+		$index = null;
 
 		if($this->isInWorld($pos)){
 			$index = Level::blockHash($pos->x, $pos->y, $pos->z);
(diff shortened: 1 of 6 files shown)


src/pocketmine/level/Level.php
```diff
@@ -1274,22 +1276,43 @@ public function getFullBlock(int $x, int $y, int $z) : int{
...
+	}
+
+	/**
+	 * Gets the Block object at the Vector3 location
 	 *
 	 * @param Vector3 $pos
-	 * @param bool    $cached
+	 * @param bool    $cached Whether to use the block cache for getting the block (faster, but may be inaccurate)
+	 * @param bool    $addToCache Whether to cache the block object created by this method call.
+	 *                             Note for plugin developers: If you are using this method a lot (thousands of times for many positions), you may want to set this to false to avoid using excessive amounts of memory.
```
Sandertv: That's awesome. Thanks for that @dktapps.

src/pocketmine/level/Level.php
```diff
@@ -1069,12 +1072,11 @@ public function scheduleDelayedBlockUpdate(Vector3 $pos, int $delay){
...
 
-		$this->neighbourBlockUpdateQueue->enqueue(Level::blockHash($pos->x + 1, $pos->y, $pos->z));
-		$this->neighbourBlockUpdateQueue->enqueue(Level::blockHash($pos->x - 1, $pos->y, $pos->z));
-		$this->neighbourBlockUpdateQueue->enqueue(Level::blockHash($pos->x, $pos->y + 1, $pos->z));
-		$this->neighbourBlockUpdateQueue->enqueue(Level::blockHash($pos->x, $pos->y - 1, $pos->z));
-		$this->neighbourBlockUpdateQueue->enqueue(Level::blockHash($pos->x, $pos->y, $pos->z + 1));
-		$this->neighbourBlockUpdateQueue->enqueue(Level::blockHash($pos->x, $pos->y, $pos->z - 1));
+		for($i = 0; $i <= 5; ++$i){
+			if($this->isInWorld($s = $pos->getSide($i))){
+				$this->neighbourBlockUpdateQueue->enqueue(Level::blockHash($pos->x, $pos->y, $pos->z));
```
jasonwynn10: Why is this not using `$s` instead of `$pos`?

dktapps: Thanks.

dktapps:
preprocessor is going to be a problem...

dktapps:
I guess for optimized builds the exception throw is not strictly necessary.

PR #916, opened 2017-05-14
Added short-hand Event->call() for easier event calling

dktapps:
## Introduction
`$this->server->getPluginManager()->callEvent($ev)` is long and annoying. `$ev->call()` is a much nicer and quicker short-form that can now be used as an alternative.

This PR is still a work in progress since there are a very large number of these event calls scattered throughout the code. A sample has been supplied as an example. If everyone is happy with it I will convert the remaining usages.

### API changes
- Added `Event->call()` which can be substituted for long `blah->getBlah()->getBlah()->getServer()->getPluginManager()->...` calls.

## Backwards compatibility
No BC breaks, only API additions.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -890,7 +890,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 
 		$pos = $this->level->getSafeSpawn($this);
 
-		$this->server->getPluginManager()->callEvent($ev = new PlayerRespawnEvent($this, $pos));
+		($ev = new PlayerRespawnEvent($this, $pos))->call();
 
 		$pos = $ev->getRespawnPosition();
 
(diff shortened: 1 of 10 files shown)


TheDiamondYT1:
Maybe deprecate `PluginManager->callEvent()`?

dktapps:
No, this is simply syntactic sugar for developers.

TheDiamondYT1:
@dktapps Okay

src/pocketmine/Player.php
```diff
@@ -1086,7 +1086,7 @@ public function dataPacket(DataPacket $packet, $needACK = false){
 		$timings = Timings::getSendDataPacketTimings($packet);
 		$timings->startTiming();
 
-		$this->server->getPluginManager()->callEvent($ev = new DataPacketSendEvent($this, $packet));
+		($ev = new DataPacketSendEvent($this, $packet))->call();
 		if($ev->isCancelled()){
```
SOF3: What about also returning isCancelled() in the return value, or a constant `false` if the event is not a Cancellable?

SOF3 (changes requested):

SOF3 (approved):

TheDiamondYT1:
When will this be merged?

src/pocketmine/Player.php
SOF3: Then what about returning `$this`?

dktapps:
This syntax, while shorter, is more ugly than the current approach. This branch also has conflicts with master that I have no interest in resolving, so I'm going to close it.

jasonwynn10:
:frowning:

PR #919, opened 2017-05-15
Fixed undefined offset when retrieving something from the thread store that doesn't exist

dktapps:
Not sure whether this ought to be returning null or throwing an exception.


src/pocketmine/scheduler/AsyncTask.php
```diff
@@ -139,7 +139,7 @@ public function getTaskId(){
 	 */
 	public function getFromThreadStore($identifier){
 		global $store;
-		return $this->isGarbage() ? null : $store[$identifier];
+		return $this->isGarbage() or !isset($store[$identifier]) ? null : $store[$identifier];
```
SOF3: You fell into the precedence trap.

PR #920, opened 2017-05-15
Added PlayerDataSaveEvent, resolved #487

### LINKED ISSUE #487: api3/network: Remove data saving from PlayerQuitEvent, add separate events for saving player data
### Issue description
Lots of plugins like to do weird things during PlayerQuitEvent, such as unloading their level (#284).

Irrelevant of the crash noted in referenced issue, if said crash was fixed, unloading a player's level during PlayerQuitEvent will have some unintended side-effects due to the way player data saving is tied up with PlayerQuitEvent. These side-effects include the player's current level and spawn position not being saved due to the level being inaccessible.

### Proposed solution
Separate the two events - have a PlayerDataSaveEvent independent from PlayerQuitEvent. On quit, the player's data will then be saved _before_ the PlayerQuitEvent is called. PlayerDataSaveEvent can then be used anytime a player is saved. This will also allow weird stuff like #284 to be done without affecting saved data.

dktapps:
## Introduction
#487 describes why having auto-save controlled by PlayerQuitEvent is a bad idea. This pull request adds a PlayerDataSaveEvent, which resolves this problem and gives plugins more control over saving of player data.

### Relevant issues
#487

### API changes
- Added PlayerDataSaveEvent. This event will be called when the server is about to save offline player data. Note that there is no `getPlayer()` for this event, only `getPlayerName()`. You cannot assume a player is online when their data is being saved.
- Removed the following from PlayerQuitEvent:
  - `getAutoSave()`
  - `setAutoSave(bool)`
  - auto-save parameter from PlayerQuitEvent constructor
- Server option to disable saving player data saving can now be overridden by plugins (this is reflected by the cancelled state of the event)

### Behavioural changes
- Server->getOfflinePlayerData() will no longer save data to disk if it doesn't already exist. This is because new Players already save their own data on creation, causing the data to be saved twice when a new Player is created.

## Backwards compatibility
Minor BC break for plugins using PlayerQuitEvent to control auto-saving. This will be covered in the next ALPHA API bump.

## Tests
Loosely tested. The core has no trouble with this, and I've tested cancelling and un-cancelling this new event, works fine. I haven't tested using a plugin to alter the saved data (this may cause some issues due to NBT library bad design).

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -3608,7 +3608,6 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 			$this->server->getPluginManager()->unsubscribeFromPermission(Server::BROADCAST_CHANNEL_ADMINISTRATIVE, $this);
 
 			if($this->joined){
-				//TODO: add events for player data saving
 				$this->save();
 
 				$this->server->getPluginManager()->callEvent($ev = new PlayerQuitEvent($this, $message));
@@ -3709,7 +3708,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 		}
 
 		$this->namedtag["playerGameType"] = $this->gamemode;
-		$this->namedtag["lastPlayed"] = new LongTag("lastPlayed", floor(microtime(true) * 1000));
+		$this->namedtag["lastPlayed"] = floor(microtime(true) * 1000);
 
 		if($this->username != "" and $this->namedtag instanceof CompoundTag){
 			$this->server->saveOfflinePlayerData($this->username, $this->namedtag, $async);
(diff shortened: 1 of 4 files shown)


Muqsit:
Nice! I think this justifies why the PlayerQuitEvent eats up a good amount of TPS on my server.

matcracker:
Can I propose to replace getPlayerName() with getOfflinePlayer() and return an OfflinePlayer instance?

jasonwynn10:
The player won't always be offline though.

thebigsmileXD:
@matcracker 
> You cannot assume a player is online when their data is being saved.

Thx for reading next time.

matcracker:
Apparently, you can get the offline player ;) @thebigsmileXD @jasonwynn10

PR #922, opened 2017-05-17
Fixed AsyncTask memory leaks

dktapps:
## Introduction
AsyncTask objects were not being correctly collected after completion, causing the server to leak memory if async tasks were used frequently. This was a major show-stopper for stuff like asynchronous chunk serialization, and a big problem for plugins such as @SOF3's Hormones.

### Behavioural changes
- AsyncTasks don't leak anymore

## Tests
A new submodule, [PocketMine-TesterPlugin](https://github.com/pmmp/PocketMine-TesterPlugin), has been added. Currently this plugin only contains a regression test for this particular bug.
However, the plugin can be easily extended to add more tests (TODO).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -14,3 +14,6 @@
 [submodule "tests/plugins/PocketMine-DevTools"]
 	path = tests/plugins/PocketMine-DevTools
 	url = https://github.com/pmmp/PocketMine-DevTools.git
+[submodule "tests/plugins/PocketMine-TesterPlugin"]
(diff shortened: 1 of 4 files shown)


madecode15:
With this patch, I am no longer leaking memory and my average usage has gone from 700mb to less then 200mb. Though, I had aysnc chunk sending enabled & a plugin that created AysncTasks often when players joined and quit.

This should also help timings as Memory Manager won't try to free memory when a soft limit, wasting time on GC. Though it depends on how much your server relies on these tasks.

PR #929, opened 2017-05-20
Improve checks for Position::equals() and Location::equals()

Muqsit:
## Introduction

Since Location extends Position extends Vector3, and since equals() is not being overwritten in Location and Position, Location::equals(Location) would be equal to Position::equals(Position) which would be equal to Vector3::equals(Vector3) if Location's $x, $y and $z are equal to Position's $x, $y, and $z are equal to Vector3's $x, $y, and $z.

_This was an issue reported in the development section of [forums.pmmp.io](https://forums.pmmp.io)._
### Relevant issues
* https://forums.pmmp.io/threads/rotating-head-seems-to-change-position.2349/

### API changes

* _Position::equals(PositionToCheck)_ now checks whether
  * _Position::$level === PositionToCheck::$level_

* _Location::equals(LocationToCheck)_ now checks whether
  * _Location::$level === LocationToCheck::$level_
  * _Location::$yaw == LocationToCheck::$yaw_
  * _Location::$pitch == LocationToCheck::$pitch_

### Behavioural changes

None.

## Backwards compatibility

Checking whether Position::equals(Position2) now checks whether Position's level is the same as Position2's level. If in your plugins, Position::equals(Position2) returns true though the levels are different, you'll need to convert Position2 to Vector3.

Same for Location. Location::equals(Location2) returns true only if Location::$level is the same as Location2::$level and Location::$yaw is equal to (==, not === because of floats and ints) Location2::$yaw and Location::$pitch is equal to Location2::$pitch.

## Follow-up

None, since PocketMine has never used Vector3::equals() (nor Position::equals() or Location::equals()).

## Tests

Returns true:
```php
/** @var Location $loc */
$loc2 = clone $loc;
return $loc->equals($loc2);

/** @var Position $pos */
$pos2 = clone $pos;
return $pos->equals($pos2);

$p1 = new Position(100, 220, 130, $level = Level::class);
$p2 = new Position(100, 220, 130, $level);
return $p1->equals($p2);
return $p1->equals(new Vector3($p2->x, $p2->y, $p2->z));

$l1 = new Location(100, 200, 300, 100.24, 100.34, $level = Level::class);
$l2 = new Location($l1->x, $l1->y, $l1->z, 100, 100, $level);
return !$l1->equals($l2);//100.24 != 100 and 100.34 != 100
return $l1->equals(new Vector3($l2->x, $l2->y, $l2->z));
```

### DIFF
--- a/src/pocketmine/level/Position.php
+++ b/src/pocketmine/level/Position.php
@@ -124,7 +124,7 @@ class Position extends Vector3{
 
 	public function equals(Vector3 $v){
 		if($v instanceof Position){
-			return parent::equals($v) and $v->getLevel()->getId() === $this->level->getId();
+			return parent::equals($v) and $v->getLevel() === $this->level;
 		}
 		return parent::equals($v);
 	}
--- a/src/pocketmine/level/Position.php
+++ b/src/pocketmine/level/Position.php
@@ -124,7 +124,7 @@ class Position extends Vector3{
 
 	public function equals(Vector3 $v){
 		if($v instanceof Position){
-			return parent::equals($v) and $v->getLevel() === $this->level;
+			return parent::equals($v) and $v->getLevel() === $this->getLevel();
 		}
 		return parent::equals($v);
 	}
--- a/src/pocketmine/level/Position.php
+++ b/src/pocketmine/level/Position.php
@@ -122,4 +122,10 @@ class Position extends Vector3{
 		return $this;
 	}
 
+	public function equals(Vector3 $v){
+		if($v instanceof Position){
+			return parent::equals($v) and $v->getLevel()->getId() === $this->level->getId();
+		}
+		return parent::equals($v);
+	}
 }
--- a/src/pocketmine/level/Location.php
+++ b/src/pocketmine/level/Location.php
@@ -68,4 +68,11 @@ class Location extends Position{
 	public function __toString(){
 		return "Location (level=" . ($this->isValid() ? $this->getLevel()->getName() : "null") . ", x=$this->x, y=$this->y, z=$this->z, yaw=$this->yaw, pitch=$this->pitch)";
 	}
+
+	public function equals(Vector3 $v){
+		if($v instanceof Location){
+			return parent::equals($v) and $v->yaw == $this->yaw and $v->pitch == $this->pitch;
+		}
+		return parent::equals($v);
+	}
 }


SOF3 (dismissed):
This is a backward incompatible change. Recommend merging in next API bump.

~~There is also an issue with the symmetry of the equals method. A good equals method should be symmetric, i.e. if $a->equals($b) returns true, $b->equals($a) should also return true.~~

~~According to the above code logic, an equals comparison between a Location and a Position would not be symmetric.~~

~~Consider using a canEqual method that checks inheritance in both classes: https://projectlombok.org/features/EqualsAndHashCode.html~~

~~Also note that the class name should not be checked to be equal directly, for that subclasses extending either of these two classes may not declare new fields or may not expect the equals() method to check new fields.~~

Muqsit:
@SOF3 But this is symmertical.
* Position::equals(Location) returns true if:
  * Position::$x == Location::$x
  * Position::$y == Location::$y
  * Position::$z == Location::$z
  * Position::$level === Location::$level

* Location::equals(Position) returns true if:
  * Location::$x == Position::$x
  * Location::$y == Position::$y
  * Location::$z == Location::$z
  * Location::$level == Position::$level

This is because Location::equals(Position) checks for yaw and pitch only if Position is instance of Location (which it is not).

It's fully backward compatible in all 9 possible cases (Vector3, Position and Location).

dktapps:
Yes, but now you'll start getting false in cases where Position->equals(Position) for example where previously you'd get true. Therefore this is a behavioural change and a possible BC break.

Muqsit:
@dktapps Yeah but isn't it more sensible for Position::equals(Position) to return true if the levels of both positions are same. It will return true for all the previous cases where the levels of both the positions were same.

A solution for the argument would be adding a second parameter to not break previous cases, though such a case would be rare.

SOF3:
@Muqsit the main problem is in that a developer may hold a Position instance thinking that it is a Vector3.

src/pocketmine/level/Position.php
```diff
@@ -122,4 +122,10 @@ public function setComponents($x, $y, $z){
 		return $this;
 	}
 
+	public function equals(Vector3 $v){
+		if($v instanceof Position){
+			return parent::equals($v) and $v->getLevel() === $this->level;
```
dktapps: Prefer using getLevel() here, some descendents of Position such as WeakPosition don't use the level field.

Muqsit: Oh, alright

Muqsit:
@SOF3 That would be a rare case. For the sake of the argument, should I add a second parameter to Position::equals() and Location::equals() such as...
```php
public function equals(Vector3 $v, bool $deepCheck = false){
    if($deepCheck and $v instanceof Location|Position){
        return parent::equals($v) and new change;
    }
    return parent::equals($v);
}
```

SOF3:
Rare case?... citation needed

Muqsit:
Yes, think of a case where someone would try to check whether Position::equals(Position).
Even better, from what I've seen, people use Position::round()->equals(Position::round()) or Position::floor()->equals(Position::floor()) which is basically Vector3::equals(Vector3) since round() and floor() return Vector3.

matcracker (approved):

SOF3:
A float epsilon comparison parameter might be useful.

dktapps:
Bump. Did we decide on what to do with this yet?

Muqsit:
@dktapps SOF3's first response.

jasonwynn10:
Merge with alpha6?

PR #932, opened 2017-05-20
Add Server::broadcastTitle()

Muqsit:
## Introduction

Since there is a Server::broadcastMessage(), Server::broadcastTip() and Server::broadcastPopup(); why not a Server::broadcastTitle()?
### API changes

* Added Server::broadcastTitle() function to broadcast titles to a group of recipients.

### Behavioural changes

* Added Server::broadcastTitle() function to broadcast titles to a group of recipients.

## Backwards compatibility

Compatible.

## Tests

```php
/** @var Server $server */
$server->broadcastTitle("This is a title", "and THIS, is a subtitle!");//broadcasts title "This is a title", and subtitle "and THIS, is a subtitle!"

//$server::$serverList = [Player("alistair"), Player("billy")];
$players = [
    $server->getPlayer("a"),
    $server->getPlayer("b")
];

//The following broadcasts the given title and subtitle to players: "Aliastair" and "Billy"
$server->broadcastTitle("If you can read this", "then you are an honored player!", -1, -1, -1, $players);
```

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -78,6 +78,8 @@ use pocketmine\network\mcpe\protocol\BatchPacket;
 use pocketmine\network\mcpe\protocol\DataPacket;
 use pocketmine\network\mcpe\protocol\ProtocolInfo;
 use pocketmine\network\mcpe\protocol\PlayerListPacket;
+use pocketmine\network\mcpe\protocol\SetTitlePacket;
+use pocketmine\network\mcpe\protocol\TextPacket;
 use pocketmine\network\mcpe\RakLibInterface;
 use pocketmine\network\Network;
 use pocketmine\network\query\QueryHandler;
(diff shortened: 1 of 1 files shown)


dktapps (dismissed):
This is a bad idea with multi-version support planned for the future. Prefer using API methods. Micro optimizations like this just complicate things.

dktapps:
and if you're concerned about saving a few microseconds, just remember disk I/O is synchronous... that's probably the biggest single performance issue in the whole of PocketMine right now.

Muqsit:
Makes sense, updated.

src/pocketmine/Server.php
```diff
@@ -1717,6 +1717,35 @@ public function broadcastPopup($popup, $recipients = null){
...
 	/**
+	 * @param string $title
+	 * @param string $subtitle
+	 * @param int    $fadeIn Duration in ticks for fade-in. If -1 is given, client-sided defaults will be used.
+	 * @param int    $stay Duration in ticks to stay on screen for
+	 * @param int    $fadeOut Duration in ticks for fade-out.
+	 *
+	 * @return int
+	 */
+	public function broadcastTitle(string $title, string $subtitle = "", int $fadeIn = -1, int $stay = -1, int $fadeOut = -1, $recipients = null){
```
dktapps: Missing doc comment for $recipients

Muqsit: My bad

dktapps (approved):

PR #939, opened 2017-05-25
Fix Button placement, enhance breaking logic

### LINKED ISSUE #937: Levers and buttons rotate to odd angles when placed
### Issue description

Expected result: when you place a lever or button on a wall, it should remain fixed to the wall
Actual result: after a tick or so, the button or level changes orientation, often to the same compass direction.

https://imgur.com/CmNhjkV

1. Place a lever or button in creative, or open a map from the Java version of Minecraft with buttons/levers on walls
2. The lever or button quickly rotates to an odd orientation

### OS and versions
* PocketMine-MP: 1.6.2dev "Unleashed" (API 3.0.0-ALPHA5)
* PHP: 7.0.18
* Server OS: MacOS Sierra 10.12.5
* Game version: PE

thebigsmileXD:
## Introduction

Several issues according to the placement and breaking of buttons and levers were made.
This PR is for fixing those placements for buttons and the breaking/placing logic of torches
It also fixes hardness and resistance
EDIT:
I forgot to mention that this also deletes Woodenbutton.php and makes buttons one class, like fencegates or glazed terracotta.

### Relevant issues

* Fixes #937 and more

## Backwards compatibility

Compatible

## Follow-up

After: update lever code, and probably add a function to check if a blockface is "solid" (stairs top, slab top (and if mojang also pushes the changes of 1.12 to PE: stairs sides))

## Tests

Fully tested in the beta. Simply looks like some weird glich forces buttons on top of slabs to _NOT_ work, torches use the same code and they work fine..


src/pocketmine/block/Button.php
```diff
@@ -0,0 +1,81 @@
...
+
+		return false;
+	}
+
+	/**
+	 * Returns the best tool type to use for breaking this type of block.
+	 * @return int
+	 */
+	public function getToolType(){
+		return $this->getId() === Block::STONE_BUTTON ? Tool::TYPE_PICKAXE : Tool::TYPE_AXE;
```
dktapps: It only makes sense to squash blocks into the same class type when they share all the same properties and no difference in code. This is hacky and reduces scalability.

PR #940, opened 2017-05-25
Add EndRod

inxomnyaa:
## Introduction

It adds end rods as decoration and light source
Supports vanilla-like rotation and exact same placement

### Behavioural changes

new block  ¯\_(ツ)_/¯

## Tests

Tested and used since month in a fork. No issues known

### DIFF
--- a/src/pocketmine/block/EndRod.php
+++ b/src/pocketmine/block/EndRod.php
@@ -1,0 +1,106 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * @author PocketMine Team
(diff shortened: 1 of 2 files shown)


src/pocketmine/block/EndRod.php
```diff
@@ -0,0 +1,70 @@
...
+			5 => 4,
+		];
+		$this->meta = ($target->getId() === self::END_ROD && $faces[$face] == $target->getDamage()) ? Vector3::getOppositeSide($faces[$face]) : $faces[$face];
+		$this->getLevel()->setBlock($block, $this, true, true);
+		return true;
+	}
+
+	public function getDrops(Item $item): array{
+		return [
+			[$this->id, 0, 1],
```
jasonwynn10: this needs changed according to 00afc5c363c8e54fd8f82dfdb3a3bff76672de9d

jasonw4331 (changes requested):

dktapps (changes requested):
This causes players to be kicked for flying when walking on top of end-rods. Also see @jasonwynn10 's comment about the drops.

src/pocketmine/block/EndRod.php
thebigsmileXD: i see

thebigsmileXD:
@dktapps "This", what exactly, i can't see it for a strange reason

TheDiamondYT1:
If i were you @dktapps i wouldnt accept block pull requests untill the new json system is in place

src/pocketmine/block/EndRod.php
```diff
@@ -0,0 +1,70 @@
...
+ *
+*/
+
+namespace pocketmine\block;
+
+use pocketmine\item\Item;
+use pocketmine\math\Vector3;
+use pocketmine\Player;
+
+class EndRod extends Flowable{
```
jasonwynn10: Shouldn't extend flowable

jasonw4331 (changes requested):

jasonwynn10:
@TheDiamondYT1 this is a PR on the api3/blocks branch, so it will be added with everything else in the rewrite.

thebigsmileXD:
Bounding box fixed.

thebigsmileXD:
Opening a new PR due to git/github issues

PR #941, opened 2017-05-26
Remove unneeded foreach loop in Human::getDrops()

Muqsit:
## Introduction

The foreach loop was useless as it was doing nothing other than copying every value of Human::getInventory()->getContents() into another array.

In other words:
```
//METHOD 1
$drops = [];
foreach(Human::getInventory()->getContents() as $item){
    $drops[] = $item;
}
//METHOD 2
$drops = Human::getInventory()->getContents();
```
METHOD 1 = METHOD 2

The only difference would be the array keys.
In METHOD 2, the array keys are basically the slot numbers. so it may look something like:
```
$drops = [
    0 => Item::class,
    1 => Item::class,
    100 => Armor::class, Air::class,
    101 => Armor::class | Air::class
];
```
while in METHOD 1:
```
$drops = [
    0 => Item::class,
    1 => Item::class,
    2 => Armor::class, Air::class,
    3 => Armor::class | Air::class
];
```

That can be solved using array_values(), but:
* Unneeded
* ~~Makes much more sense treating keys as slot numbers.~~
### API changes

None
### Behavioural changes

A micro-optimization, and a sensible change. Living::getDrops() returns Living::getInventory()->getContents() which is a much better approach ~~as BaseInventory::getContents() returns an array in which the keys are the slots of the values (items).~~

### DIFF
--- a/src/pocketmine/entity/Human.php
+++ b/src/pocketmine/entity/Human.php
@@ -425,14 +425,7 @@ class Human extends Creature implements ProjectileSource, InventoryHolder{
 	}
 
 	public function getDrops(){
-		$drops = [];
-		if($this->inventory !== null){
-			foreach($this->inventory->getContents() as $item){
-				$drops[] = $item;
-			}
-		}
-
-		return $drops;
+		return $this->inventory !== null ? $this->inventory->getContents() : [];
 	}
 
 	public function saveNBT(){
--- a/src/pocketmine/entity/Human.php
+++ b/src/pocketmine/entity/Human.php
@@ -425,7 +425,7 @@ class Human extends Creature implements ProjectileSource, InventoryHolder{
 	}
 
 	public function getDrops(){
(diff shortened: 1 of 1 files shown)


thebigsmileXD:
The slot id's in drops are indeed useless.

Muqsit:
@thebigsmileXD That wasn't the point of this change. Read the title again.

thebigsmileXD:
But its a side effect.

Muqsit:
It's not a side effect at all. How would using array keys (that aren't being used anyway) a side effect?

dktapps:
@Muqsit remember that it won't just be the core code that uses this method. It is, after all, an API method.

dktapps:
I'd prefer use of array_values to ensure backwards-compatibility.

dktapps (approved):

PR #947, opened 2017-05-27
Added an easier function to quickly get an OfflinePlayer object from a Player object

Sandertv:
## Introduction

This pull request adds a very simple function: $player->getOfflinePlayer(). I felt like it would be easier to do this instead of having to construct a new OfflinePlayer object without API. 

### Relevant issues

No issues for this kind of addition.

### API changes

A new API function in the Player class:
`$player->getOfflinePlayer()`

### Behavioural changes

This does change the way anything behaves, nor performance.

## Backwards compatibility
 No backwards incompatible changes are made.

## Follow-up

-
## Tests

-

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -321,6 +321,16 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	/** @var PermissibleBase */
 	private $perm = null;
 
+
+	/**
+	 * Returns a new OfflinePlayer object with the name of the current player.
+	 *
+	 * @return OfflinePlayer
+	 */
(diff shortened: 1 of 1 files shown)


dktapps:
why? I can't think of any circumstances where this would be needed.

Sandertv:
This could make things easier with for example an API where an OfflinePlayer should be passed to the function instead of a Player object. If you have a Player object, it would be easier to use this function instead of having to construct a new object yourself.

I agree this isn't something which would be used very widely, but could come in handy in some occasions.

dktapps:
again: why? why wouldn't you simply use the IPlayer interface?

Sandertv:
Good point 😄

PR #948, opened 2017-05-27
Update core for PHP 7.1

dktapps:
## Introduction
read the title

### API changes
Alterations to `pocketmine\Thread->start()` and `pocketmine\Worker->start()` in accordance with pthreads 3.1.7-dev

### Behavioural changes
less performance because PHP 7.1 is somewhat slower than 7.0.

## Backwards compatibility
This is _not_ backwards-compatible with PHP 7.0, however most plugins won't be suffering breaking changes yet, at least until we start upgrading the core to use new features.

## Follow-up
- typehint all the things
- nullable typehints and void return types

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,20 @@
 language: php
 
 php:
- - 7.0
+ - 7.1
 
 before_script:
- - pecl install channel://pecl.php.net/pthreads-3.1.6
+# - pecl install channel://pecl.php.net/pthreads-3.1.6
  - pecl install channel://pecl.php.net/weakref-0.3.3
  - echo | pecl install channel://pecl.php.net/yaml-2.0.0
+ - git clone https://github.com/SirSnyder/pthreads.git -b php/71 --depth=1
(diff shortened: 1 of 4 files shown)


TheDiamondYT1:
If its slower, why use it?

dktapps:
To demonstrate that it _is_ doable, we will need to upgrade at some point. This won't be merged for a while yet regardless, this version requires a dev build of pthreads (and pthreads itself has major issues at the moment).

Muqsit:
Um, isn't pthreads an issue too?

dktapps:
I'm going to have a play about with earlier versions... the big loss of performance started with php 7.1.2.

dktapps:
This has been rebased on top of master and will now also work on php 7.2.0alpha1 with the correct binaries.

AppleDevelops:
Is PHP 7.2 any faster?

dktapps:
I don't know yet, all I know is that it works.

PR #952, opened 2017-05-29
Remove damage variable in Entity->knockBack

mal0ne-23:
## Introduction

the damage number isn't used whatsoever in the knockback function.

### API changes

knockBack second variable $damage removed

### Behavioural changes

Nope.

## Backwards compatibility

Remove the damage variable in plugins.

## Tests

no tests done, but kinda common sense

### DIFF
--- a/src/pocketmine/entity/Living.php
+++ b/src/pocketmine/entity/Living.php
@@ -161,7 +161,7 @@ abstract class Living extends Entity implements Damageable{
 
 				$deltaX = $this->x - $e->x;
 				$deltaZ = $this->z - $e->z;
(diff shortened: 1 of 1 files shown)


dktapps:
I also remember a pull request to the old repo to remove the attacker parameter that got rejected, with reason being that plugins extending entities may want to use those parameters. Also, just because it doesn't get used currently, doesn't mean it won't in the future. This is a backwards-compatibility break.

dktapps:
https://github.com/PocketMine/PocketMine-MP/pull/3427 This is shoghicp 's take on it... please read.

PR #959, opened 2017-05-31
Added addLore() method

ghost:
## Introduction

This PR is necessary for adding lores in items rather than using a hack on setLore()

### Relevant issues

None

### API changes

` Item::addLore() `

### Behavioural changes

None

## Backwards compatibility

None

### DIFF
--- a/src/pocketmine/item/Item.php
+++ b/src/pocketmine/item/Item.php
@@ -695,7 +695,7 @@ class Item implements ItemIds, \JsonSerializable{
 		}
 		
 		$tag->display->Lore->setTagType(NBT::TAG_String);
-		$tag->display->Lore[$tag->display->Lore->getCount()] = new StringTag("", $lore);
+		$tag->display->Lore->{$tag->display->Lore->getCount()} = new StringTag("", $lore);
 		$this->setNamedTag($tag);
 	}
 
--- a/src/pocketmine/item/Item.php
+++ b/src/pocketmine/item/Item.php
@@ -684,6 +684,21 @@ class Item implements ItemIds, \JsonSerializable{
 		return [];
 	}
 
+	public function addLore(string $lore) {
+		$tag = $this->getNamedTag() ?? new CompoundTag("", []);
+		if(!isset($tag->display)){
+			$tag->display = new CompoundTag("display", []);
+		}
+		
+		if(!isset($tag->display->Lore)) {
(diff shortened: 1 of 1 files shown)


Muqsit:
It would be better if the function utilizes setLore() method.

KairusDarkSeeker:
@Musit Good idea, how do I change it?

src/pocketmine/item/Item.php
```diff
@@ -684,6 +684,21 @@ public function getLore() : array{
...
+		if(!isset($tag->display)){
+			$tag->display = new CompoundTag("display", []);
+		}
+		
+		if(!isset($tag->display->Lore)) {
+			$tag->display->Lore = new ListTag("Lore");
+		}
+		
+		$tag->display->Lore->setTagType(NBT::TAG_String);
+		$tag->display->Lore[$tag->display->Lore->getCount()] = new StringTag("", $lore);
```
Muqsit: I guess this should be
`$tag->display->Lore->{$tag->display->Lore->getCount()}`

KairusDarkSeeker: Both $tag->display->Lore->{$tag->display->Lore->getCount()} and $tag->display->Lore->[$tag->display->Lore->getCount()] works

Sandertv: Even though both work you're using two different methods in one line now. I'd say change it to what @Muqsit recommends.

KairusDarkSeeker: @Sandertv okay

KairusDarkSeeker:
Wow, using git on a vps is much more efficient than using the web editor

KairusDarkSeeker:
@Muqsit Can't, I already made the PR so yeah

thebigsmileXD:
@Muqsit https://github.com/pmmp/PocketMine-MP/commit/3e76c3a6ddd8ec9cb25a20c96bbbaadb72e4f665#diff-be0be6071c2c2f0abecaa7dd4e16e2f1R686 there it isn't utilized. is addlore actually needeed?

dktapps (changes requested):
I don't see the need for all the extra (copy-pasted) code. Simply:
```
$lore = $item->getLore();
$lore[] = "new line";
$item->setLore($lore);
```

KairusDarkSeeker:
@dktapps okay

PR #962, opened 2017-06-01
TAG_Byte should always be signed

dktapps:
This is correct, but this change is practically guaranteed to break something.

This fixes an issue with inventory where null slot link (-1) is saved as a TAG_Byte but then read back as unsigned (255).


SalmonDE:
Fixed my issue 👍

dktapps:
Checked over the core code, there doesn't seem to be anything vanilla that breaks with respect to this.

PR #969, opened 2017-06-02
Binary strict types

dktapps:
## Introduction
It's critical to the functionality of the server that network sent data is always valid. This PR cleans up some Binary methods, adds documentation, and most importantly adds strict types for packets and Binary methods.

This will help highlight issues when a required field is missing from a packet (as it will default to NULL and raise a TypeError). This is now particularly important for plugins which use packets when packet field names change (such as AddEntityPacket's `eid` into `entityRuntimeId`) as this will now crash loudly to indicate when the developer is doing something stupid, instead of silently interpreting NULL as 0, an empty string, or whatever else.

### Behavioural changes
Strict types have been enabled on Binary and DataPacket methods. This prevents accidental use of uninitialized/undefined/null packet fields.

## Backwards compatibility
See above.

## Tests
This has been subjected to basic testing; the server is joinable and various basic gameplay things have been tested. This however hasn't been tested extensively and may still raise errors in places due to previously hidden undefined behaviour.

### DIFF
--- a/src/pocketmine/network/mcpe/protocol/AddEntityPacket.php
+++ b/src/pocketmine/network/mcpe/protocol/AddEntityPacket.php
@@ -38,8 +38,8 @@ class AddEntityPacket extends DataPacket{
 	public $speedX;
 	public $speedY;
 	public $speedZ;
-	public $yaw;
-	public $pitch;
+	public $yaw = 0.0;
+	public $pitch = 0.0;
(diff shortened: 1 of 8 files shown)


dktapps:
I'm going to merge this directly; the changes here will help highlight issues which have cropped up since 1.1. Any crashes can be resolved as they are reported.

PR #973, opened 2017-06-02
Added asVector3(), asPosition() and asLocation()

matcracker:
## Introduction

This pull request allow developers to get faster access to the block position without work around the level or its coordinates.
### API changes
Added `Position::asPosition()` function.
Added `Location::asLocation()` function.
Added `Vector3::asVector3()` function.
Deprecated `Entity::getLocation()` function. _(Note by @SOF3: Undone while squash-merging)_
Deprecated `Entity::getPosition()` function. _(Note by @SOF3: Undone while squash-merging)_

### DIFF
--- a/src/pocketmine/entity/Entity.php
+++ b/src/pocketmine/entity/Entity.php
@@ -1383,10 +1383,20 @@ abstract class Entity extends Location implements Metadatable{
 		return true;
 	}
 
+	/**
+	 * @deprecated Use asPosition() instead getPosition()
+	 * 
+	 * @return Position
+	 */
 	public function getPosition(){
 		return new Position($this->x, $this->y, $this->z, $this->level);
 	}
 
+	/**
+	 * @deprecated Use asLocation() instead getLocation()
+	 * 
+	 * @return Location
+	 */
 	public function getLocation(){
 		return new Location($this->x, $this->y, $this->z, $this->yaw, $this->pitch, $this->level);
 	}
--- a/src/pocketmine/level/Location.php
+++ b/src/pocketmine/level/Location.php
@@ -62,7 +62,7 @@ class Location extends Position{
 	 * 
 	 * @return Location
 	 */
-	public function asLocation() : Location {
+	public function asLocation() : Location{
 		return new Location($this->x, $this->y, $this->z, $this->yaw, $this->pitch, $this->level);
 	}
 
--- a/src/pocketmine/level/Location.php
+++ b/src/pocketmine/level/Location.php
@@ -57,6 +57,15 @@ class Location extends Position{
 		return new Location($pos->x, $pos->y, $pos->z, $yaw, $pitch, $level ?? (($pos instanceof Position) ? $pos->level : null));
 	}
 
+	/**
+	 * Return a Location instance
+	 * 
+	 * @return Location
+	 */
+	public function asLocation() : Location {
+		return new Location($this->x, $this->y, $this->z, $this->yaw, $this->pitch, $this->level);
+	}
+
 	public function getYaw(){
 		return $this->yaw;
 	}
--- a/src/pocketmine/level/Position.php
+++ b/src/pocketmine/level/Position.php
@@ -47,6 +47,15 @@ class Position extends Vector3{
 	}
 
 	/**
+	 * Return a Position instance
+	 * 
+	 * @return Position
+	 */
+	public function asPosition() : Position {
+		return new Position($this->x, $this->y, $this->z, $this->level);
+	}
+	
+	/**
 	 * Returns the target Level, or null if the target is not valid.
 	 * If a reference exists to a Level which is closed, the reference will be destroyed and null will be returned.
 	 *
--- a/src/pocketmine/level/Position.php
+++ b/src/pocketmine/level/Position.php
@@ -51,7 +51,7 @@ class Position extends Vector3{
 	 * 
 	 * @return Position
 	 */
-	public function asPosition() : Position {
+	public function asPosition() : Position{
 		return new Position($this->x, $this->y, $this->z, $this->level);
 	}
 	
--- a/src/pocketmine/math/Vector3.php
+++ b/src/pocketmine/math/Vector3.php
@@ -158,6 +158,15 @@ class Vector3{
 	}
 
 	/**
+	 * Return a Vector3 instance
+	 * 
+	 * @return Vector3
+	 */
+	public function asVector3() : Vector3 {
+		return new Vector3($this->x, $this->y, $this->z);
+	}
+
+	/**
 	 * Returns the Vector3 side number opposite the specified one
 	 *
 	 * @param int $side 0-5 one of the Vector3::SIDE_* constants
--- a/src/pocketmine/math/Vector3.php
+++ b/src/pocketmine/math/Vector3.php
@@ -162,7 +162,7 @@ class Vector3{
 	 * 
 	 * @return Vector3
 	 */
-	public function asVector3() : Vector3 {
+	public function asVector3() : Vector3{
 		return new Vector3($this->x, $this->y, $this->z);
 	}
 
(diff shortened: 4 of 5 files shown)


Muqsit:
class Block extends Position
Isn't that enough?

dktapps:
Why? A Block _is_ a position.

You also cannot assume that any of these fields are defined when this method is called.

Besides, if you _really_ want a position, just use `Position::fromObject($block, $block->getLevel())`.

matcracker:
Uh huh, I didn't know there was a `Position::fromObject($block, $block->getLevel())`, good to know. Anyway, in my option, it's an hidden method. See my next commit, would be better.

dktapps:
> You also cannot assume that any of these fields are defined when this method is called.

you still didn't cater for this, and I object to adding extra API methods unless they actually contribute something useful to the API.

SOF3:
I recommend adding the `getLocation()`, `getPosition()` and `getVector3()` methods to those three classes directly, since similar methods also exist in Player class etc.

matcracker:
According to `getLocation()` constructor, it requires yaw and pitch that those blocks haven't got.
`Position` extends `Vector3`, so it's useless to get the `gerVector3()` method.

Muqsit:
@SOF3 has a better idea.
Add getPosition() in Location,
Add getVector3() in Position.
Done.

matcracker:
@Muqsit and where do you add `getLocation()`?
Summary:
`Position::getVector3()`
`Position::getLocation()`
`Location::getPosition()`

Muqsit:
@matcracker you dont need to add it.

IrishPacks:
A couple of months ago I was working on a plugin that dealt with storing tons of data in an array(Up to 12000 blocks at least). I was just storing the block, and I noticed how bad the performance was, so what I did was instead of storing the block, I decided to create a new Vector3 and store that instead, which lead to MASSIVE performance increases.

Long story short, Block::getPosition() should be made a function for situations like those. As stated before, blocks ARE positions, but they carry massive amounts of unneeded data when storing in bulk.

dktapps:
> Besides, if you really want a position, just use Position::fromObject($block, $block->getLevel()).

I don't see the point in adding another method for this.

Muqsit:
@IrishPacks Why only Block::getPosition()? Why not Entity::getPosition() and every single thing that'll extend Position in the future. I don't know why SOF3's idea doesn't make sense, it's much more flexible!

IrishPacks:
@Muqsit I did skim those, but that would make more sense to apply it to anything that is a position, vector3, etc. :+1: from me

SOF3:
Something that extends Position doesn't necessarily extend Location. It is more useful to create these:

```php
public function Vector3::getVector3() : Vector3;
public function Position::getPosition() : Position;
public function Location::getLocation() : Location;
```

However, it may be more descriptive to call them `asVector3()`, `asPosition()` and `asLocation()` since the function actually returns an equal value, not a value that it owns.

matcracker:
@SOF3 What do they return? A self instance or a new object instance?

SOF3:
@matcracker you're referring to a this instance, not self instance. 
And my answer is, it should be a new instance. A getter sounds like it's fetching an existing value, while it's actually something similar to type conversion, which returns ne

src/pocketmine/level/Location.php
```diff
@@ -57,6 +57,15 @@ public static function fromObject(Vector3 $pos, Level $level = null, $yaw = 0.0,
 		return new Location($pos->x, $pos->y, $pos->z, $yaw, $pitch, $level ?? (($pos instanceof Position) ? $pos->level : null));
 	}
 
+	/**
+	 * Return a Location instance
+	 * 
+	 * @return Location
+	 */
+	public function asLocation() : Location {
+		return new Location($this->x, $this->y, $this->z, $this->yaw, $this->pitch, $this->level);
```
Muqsit: return self;

matcracker: @SOF3 said:
> ... it should be a new instance. A getter sounds like it's fetching an existing value, while it's actually something similar to type conversion, which returns new objects.

dktapps: That would return descendents of the class, which defeats the object of doing this. (No pun intended).

src/pocketmine/level/Location.php
```diff
@@ -57,6 +57,15 @@ public static function fromObject(Vector3 $pos, Level $level = null, $yaw = 0.0,
 		return new Location($pos->x, $pos->y, $pos->z, $yaw, $pitch, $level ?? (($pos instanceof Position) ? $pos->level : null));
 	}
 
+	/**
+	 * Return a Location instance
+	 * 
+	 * @return Location
+	 */
+	public function asLocation() : Location {
```
Muqsit: code formatting

src/pocketmine/level/Position.php
```diff
@@ -47,6 +47,15 @@ public static function fromObject(Vector3 $pos, Level $level = null){
 	}
 
 	/**
+	 * Return a Position instance
+	 * 
+	 * @return Position
+	 */
+	public function asPosition() : Position {
```
Muqsit: code formatting

src/pocketmine/math/Vector3.php
```diff
@@ -158,6 +158,15 @@ public function getSide($side, $step = 1){
 	}
 
 	/**
+	 * Return a Vector3 instance
+	 * 
+	 * @return Vector3
+	 */
+	public function asVector3() : Vector3 {
```
Muqsit: code formatting

Muqsit: code formatting

Muqsit:
Um, what about [these](https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/entity/Entity.php#L1386-#L1392)? I don't think we'll be needing them anymore. But if we remove it, we'll need to change a lot of `Entity::getPosition()` and `Entity::getLocation()` functions to the new `as*()` functions.... unless we name the new functions `get*()`.

matcracker:
This is why I initially wanted to implement just getPosition() on Block class.

SOF3:
@Muqsit \x42lameShighi, the answer to life, the universe and everything.
Deprecating the Entity getLocation() methods might be reasonable, but there is no strong reason to remove them soon.

src/pocketmine/math/Vector3.php
```diff
@@ -158,6 +158,15 @@ public function getSide($side, $step = 1){
...
+		return new Vector3($this->x, $this->y, $this->z);
```
thebigsmileXD: Doesn't this return itself? why not just clone self;?

dktapps: as I said above...

> That would return descendents of the class, which defeats the object of doing this. (No pun intended).

thebigsmileXD: But xxxclass as xxxclass is just useless.

SOF3: @thebigsmileXD this whole pull request adds functions that serve the exact same purpose as the `Player::getPosition()` method.

src/pocketmine/entity/Entity.php
```diff
@@ -1383,10 +1383,20 @@ protected function switchLevel(Level $targetLevel){
...
+	 * @deprecated Use asPosition() instead getPosition()
+	 * 
+	 * @return Position
+	 */
 	public function getPosition(){
 		return new Position($this->x, $this->y, $this->z, $this->level);
 	}
 
+	/**
+	 * @deprecated Use asLocation() instead getLocation()
```
Muqsit: That being said, you'll need to change all deprecated Entity::get*() fields to Entity::as*() fields... pain! :/

SOF3: @Muqsit there are actually many places in the PocketMine source that should use this method but did not do so.

dktapps: I don't think there's any need for deprecating these. While entities are indeed positions, it also does make sense to be able to say Entity->getPosition() or Entity->getLocation(). The new methods are intended to make life easier for developers, yes?

dktapps:
This has been merged in the referenced commit with some minor alterations. Thanks for contributing.

PR #993, opened 2017-06-03
Strict type all the things

dktapps:
## Introduction
I decided to bite the bullet on this one and just do it. Typehinting and improved documentation will follow.

This pull request makes strict types the standard throughout the PocketMine-MP source code.

### Why?
One of the biggest weaknesses of previous versions of PHP is the tendecy to produce unexpected behaviour when something bad happens. Strict types help prevent this, as they disable type coercion for function arguments and return types, which prevents undefined/unexpected behaviour.

You can see already looking at the commits that enabling strict types already highlighted dozens of issues which were previously hidden and unknown, causing weird behaviour. This has helped fix a large number of hidden bugs.

### Behavioural changes
The PocketMine-MP core code is now ALL subjected to strict-type checking rules.

### API changes
There are no major API changes to note with this pull request, since strict type rules only apply per-file. However, do be aware that:
- Function argument strictness is decided by the _caller_, not the callee. This means that in most cases you'll still be able to call core API methods without type-casts if you don't have strict types enabled in your plugin code, because the arguments will be coerced to match the signature if the caller does not have strict types enabled. I dislike this behaviour, and recommend you enable strict types in your plugins to help highlight issues when you do something stupid, however this is not a requirement.
- Return type strictness is decided by the callee. This means that if you define an `int` return type and have strict types enabled, you may not return a float or string or anything else.
- We will be type-hinting most function arguments and return types in the core code (where applicable) in the near future. This may break code that inherits from the core code which does not declare compatible types on overridden methods. Currently this should not be a problem, but here is a warning for the future.
### Tests
The server is joinable, and basic functionality of the server has been verified; however this has not been tested in detail, and beyond a shadow of a doubt there will be many more crashes to see yet. 

However, I've always been a firm supporter of the idea that crash > unexpected behaviour. Crashes tell you when the developer has done something stupid, and the sooner and more loudly it crashes, the quicker and easier it is to fix.

### DIFF
--- a/src/pocketmine/Achievement.php
+++ b/src/pocketmine/Achievement.php
@@ -110,7 +110,7 @@ abstract class Achievement{
 	public static function broadcast(Player $player, $achievementId){
 		if(isset(Achievement::$list[$achievementId])){
 			$translation = new TranslationContainer("chat.type.achievement", [$player->getDisplayName(), TextFormat::GREEN . Achievement::$list[$achievementId]["name"] . TextFormat::RESET]);
-			if(Server::getInstance()->getConfigString("announce-player-achievements", true) === true){
+			if(Server::getInstance()->getConfigBoolean("announce-player-achievements", true) === true){
 				Server::getInstance()->broadcastMessage($translation);
 			}else{
 				$player->sendMessage($translation);
--- a/src/pocketmine/Achievement.php
+++ b/src/pocketmine/Achievement.php
@@ -19,6 +19,8 @@
  *
 */
 
+declare(strict_types=1);
+
 namespace pocketmine;
 
 use pocketmine\event\TranslationContainer;
--- a/src/pocketmine/Collectable.php
+++ b/src/pocketmine/Collectable.php
@@ -19,6 +19,8 @@
  *
 */
 
+declare(strict_types=1);
+
 namespace pocketmine;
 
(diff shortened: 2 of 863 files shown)


dktapps:
Unbelievable... this is exposing so many bugs it's unreal

Qimie:
ikr

dktapps:
@PEMapModder this is exposing all the bugs in your attributes mess :(

Muqsit:
Location::$yaw and Location::$pitch should be 0.00 by default if I'm correct?

dktapps:
@Muqsit they are already set in the constructor https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/level/Location.php#L39

Muqsit:
@dktapps Oh well, my bad. I had run into continuous errors due to the strict types :P

dktapps:
I would like to get this merged ASAP. As far as I know there are no strict type checking violations in the core in its current state, but regardless if people could test & review this that would be great.

0x15f (commented):
Bugs galore 🐜

jasonw4331 (approved):

Qimie:
OCD Commit. Thats funny. Hahahha. Good job Dylan!

SOF3 (approved):
Whether this PR causes new crashes or not, they are already-existent issues and should not stop this PR from being merged.

PR #1010, opened 2017-06-05
Minor cosmetic update/optimization

xxFlare:
## Introduction
I was playing around with some code until I stumbled onto this. The return value is not used (https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/network/Network.php#L171), and it looked really ugly, promoting me to make this PR request.

### DIFF
--- a/src/pocketmine/network/mcpe/RakLibInterface.php
+++ b/src/pocketmine/network/mcpe/RakLibInterface.php
@@ -73,20 +73,14 @@ class RakLibInterface implements ServerInstance, AdvancedSourceInterface{
 	}
 
 	public function process(){
-		$work = false;
-		if($this->interface->handlePacket()){
-			$work = true;
-			while($this->interface->handlePacket()){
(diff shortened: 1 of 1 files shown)


dktapps:
I don't suppose any plugins might have their own network interfaces which use this return value... hmm.

xxFlare:
Wow... this issue is more widespread, PocketMine has a lot of internal functions (not for plugins) sending unused return values.

It would be better to fix it all in one commit instead of opening new PRs for each one.

matcracker:
There isn't problems, you can continue with this pull. Now the pull requests can be merged in a single commit.

TheDiamondYT1:
@matcracker Multiple PRs cannot be merged into one commit. Multiple commits in one PR can be merged into a single commit.

SOF3:
@TheDiamondYT1 he's referring to multiple commits of the same pull request.

PR #1013, opened 2017-06-05
Add Entity::isPlayer() and CommandSender::isPlayer() function

Muqsit:
## Introduction

I believe this was an incomplete piece of PocketMine. Entity::$isPlayer existed and so did ConsoleCommandSender::isPlayer(). I assume these were added for the very reason to remove instanceof checks.

### API changes

Added `Entity::isPlayer()` function which returns true if the entity is an instance of Player.
Any class extending CommandSender will now be required to have an `isPlayer()` function which should return a boolean value.

### Behavioural changes

None

## Backwards compatibility

This commit(s) is/are backwards compatible and don't break anything.

## Tests

```php
/** @var \pocketmine\entity\Item $itemEntity */
var_dump($itemEntity->isPlayer()); //returns false

/** @var Human $human */
var_dump($human->isPlayer()); //returns false

/** @var Player $player */
var_dump($player->isPlayer()); //returns true
```

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -1139,7 +1139,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 		}
 
 		foreach($this->level->getNearbyEntities($this->boundingBox->grow(2, 1, 2), $this) as $p){
-			if($p instanceof Player){
+			if($p->isPlayer()){
 				if($p->sleeping !== null and $pos->distance($p->sleeping) <= 0.1){
 					return false;
 				}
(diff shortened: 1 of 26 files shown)


Muqsit:
This hasn't been fully tested. The function might return true even if the player quits. _Basically: once a player, always a player_ is what is happening here.

dktapps:
Given that you're fussy about performance... didn't you know that method calls are actually 3x slower than an instanceof check?

instanceof also leaves no room for misinterpretation when it comes to IDEs. In some places you've changed, IDEs will no longer be able to autocomplete Player-specific methods.

dktapps:
Some more issues to note:

- `isPlayer` is only set in the Entity parent constructor. This means a crash will occur when this method is called during login before the completion of the resource packs sequence.
- `isPlayer` is not a final field, and doesn't provide a guarantee that the object is actually a Player. This can lead to crashes if the value is spoofed (undefined behaviour) because the core may attempt to call Player methods which do not exist in the target object.

PR #1025, opened 2017-06-07
Fixed DevTools getting disabled when a plugin throws an exception dur…

dktapps:
…ing load

fixes pmmp/PocketMine-DevTools#11


dktapps:
As a happy side-effect this also fixes plugins throwing exceptions during onLoad() crashing the server.

Awzaw:
works for me 👍

PR #1034, opened 2017-06-08
Added Plugins section to the issue template

dktapps:
## Introduction
more detail on how to handle plugins when a crash occurs.

It's very tiring when people keep creating issues on the repository about their plugins suddenly crashing due to core changes which haven't been dealt with by the plugin author. I've altered the issue template to ask for the following details:
- Is the issue reproducible without any plugins loaded?
- If not, has the issue creator asked for help on the forums before creating the issue?
- Can the creator provide reproducing code if this is an API issue?

### Relevant issues
more than I can be bothered listing

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -25,6 +25,12 @@ ALSO NOTE: NO support whatsoever will be provided for forks or spoons of PocketM
 * Server OS:
 * Game version: PE/Win10 (delete as appropriate)
 
+### Plugins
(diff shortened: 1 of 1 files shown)


robske110 (approved):
Really good, small informing and needed text. Maybe this text is so neatly written that people will read it. It may be better to list do-this-things under 
>If the issue is **not** reproducible without plugins:

instead of questions.

Telling to do `rm -rf ./plugins` still would be more effective though!

SOF3:
No need to put it in comments?

dktapps:
I don't think so - people will be required to answer these questions anyway, and I don't want the answer getting hidden.

PR #1047, opened 2017-06-09
Fix #1046

### LINKED ISSUE #1046: Sleeping in a bed causes error in console
### Issue description

Sleeping in a bed caused an error in console.
Tested in Creative mode.

[11:21:26] [Server thread/CRITICAL]: TypeError: "Argument 1 passed to pocketmine\utils\BinaryStream::putBool() must be of the type boolean, null given, called in phar://C:/Users/Mark/Desktop/1.1/PocketMine-MP.phar/src/pocketmine/network/mcpe/protocol/SetSpawnPositionPacket.php on line 63" (EXCEPTION) in "/src/pocketmine/utils/BinaryStream__64bit" at line 92

1. Wait for night
2. Sleep in bed

### OS and versions

* PocketMine-MP:  PocketMine-MP 1.6.2dev-168
* PHP: 7
* Server OS: Win 10
* Game Version: PE/Win10 (delete as appropriate)

### Plugins
Yes

* N/A

Misteboss:
Fix #1046 issue, it was happening because spawnForced field wasn't filled

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -1184,7 +1184,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 		$pk->y = (int) $this->spawnPosition->y;
 		$pk->z = (int) $this->spawnPosition->z;
 		$pk->spawnType = SetSpawnPositionPacket::TYPE_PLAYER_SPAWN;
-		$pk->spawnForced = true;
+		$pk->spawnForced = false;
 		$this->dataPacket($pk);
 	}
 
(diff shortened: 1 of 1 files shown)


src/pocketmine/Player.php
```diff
@@ -1184,6 +1184,7 @@ public function setSpawn(Vector3 $pos){
 		$pk->y = (int) $this->spawnPosition->y;
 		$pk->z = (int) $this->spawnPosition->z;
 		$pk->spawnType = SetSpawnPositionPacket::TYPE_PLAYER_SPAWN;
+		$pk->spawnForced = true;
```
dktapps: I'm not sure this should be true... spawnForced means players always spawn at their spawn point when they join the server.

Other than that, looks good.

dktapps:
Don't worry about Travis failing... there's something wrong with PECL so all the builds are crashing.

src/pocketmine/Player.php
Misteboss: Idk cause it's a PlayerSpawn and not a World Spawn so.. ?

dktapps: I'd leave it as false anyway, vanilla always sends false for this field.

dktapps: This pull request is primarily fixing a core bug, feature additions should be done separately.

Misteboss: Ok no problem ^^

PR #1059, opened 2017-06-10
Fix missing typecast

glemsom:
## Introduction
Fix missing typecasting during spawn and creation of new worlds

### API changes
None

### Behavioural changes
N/A
## Backwards compatibility
N/A

## Follow-up
Typecasting might be missing in more places 

## Tests
N/A


dktapps:
We don't support 32-bit systems anymore. These typecasts are not required if not using 32-bit.

Issue #1087, opened 2017-06-14
Possibility to drop items in Spectator mode


SuperMaXAleX:
### Issue description
The title is saying everything - You can drop items/blocks in Spectator mode.

### Steps to reproduce the issue
1. Be in Spectator mode
2. Try to drop an item/block

### OS and versions
* PocketMine-MP 1.6.2dev 「Unleashed」 implementing API version 3.0.0-ALPHA5 for Minecraft: PE v1.1.0.55 (protocol version 113)
* PHP: 7.0.3
* Server OS: OS X El Capitan
* Game version: PE/W10

### Plugins
- Tested without plugins

### Crashdump, backtrace or other files
* None

SuperMaXAleX:
RIP this issue

bettylovedertderrdy:
genisyspro better !

dktapps:
Fixed in MCPE 1.2 inventory changes.

PR #1090, opened 2017-06-14
Fix Spectator Orygen Decreasing + Spectator Item Dropping

0x15f:
## Introduction
Previously oxygen would decrease while underwater as a spectator, when the gamemode was switched to survival and the player is still underwater they would start drowning, this pr also fixes issue #1087 where spectators can (possibly) drop items.

### DIFF
--- a/src/pocketmine/entity/Living.php
+++ b/src/pocketmine/entity/Living.php
@@ -236,6 +236,7 @@ abstract class Living extends Entity implements Damageable{
 					$this->setDataProperty(self::DATA_AIR, self::DATA_TYPE_SHORT, 400);
 				}else{
 					if($this instanceof Player && $this->isSpectator()){
+						Timings::$timerLivingEntityBaseTick->stopTiming();
 						return $hasUpdate;
 					}
 					$hasUpdate = true;


src/pocketmine/Player.php
```diff
@@ -2864,6 +2864,10 @@ public function handleDropItem(DropItemPacket $packet) : bool{
 			// Windows 10 Edition drops the contents of the crafting grid on container close - including air.
 			return true;
 		}
+		
+		if($this->gamemode === self::SPECTATOR) {
```
TheDiamondYT1:
```php
$this->isSpectator()
```
To match the rest of the code

src/pocketmine/entity/Living.php
```diff
@@ -234,6 +235,9 @@ public function entityBaseTick($tickDiff = 1){
 				if($this instanceof WaterAnimal){
 					$this->setDataProperty(self::DATA_AIR, self::DATA_TYPE_SHORT, 400);
 				}else{
+					if($this instanceof Player && $this->getGamemode() === Player::SPECTATOR){
```
TheDiamondYT1: Read my comment above

src/pocketmine/Player.php
ImagicalGamer: I tested the execute times for $this->isSpectator() and $this->gamemode === self::SPECTATOR, $this->isSpectator() takes longer to execute than $this->gamemode === self::SPECTATOR. We're all about performance right? ;)

TheDiamondYT1: But surely that wou!d only be a very micro optimization?

ImagicalGamer: It's a micro optimization yes, but what if a server has 9000 players (on 1 PM server not realistic but theoretical) and the function is run for each of them? Every micro optimization counts ;)

dktapps: Code readability and ease of development takes priority over saving a couple of microseconds here and there. If you think quibbling over things like this will improve performance in any noticeable manner, you're wasting your time.

ImagicalGamer: I never said it was noticeable, I prefer $this->gamemode === self::SPECTATOR over $this->isSpectator(), if you prefer it the other way you're welcome to close the PR and someone else can make a new one.

ImagicalGamer: The code can be read just fine, if you have issues reading it I would suggest visiting the eye doctor. Nor does the code make developing harder, if 14 extra chars on a line make things considerably  difficult for you thats not my issue.

xxFlare: It's not so much that it's difficult to read... It's the advantages of using an API that save trouble later.
1) If in the future there is a change in the spectator system, the isSpectator() function can be modified and the pmmp team won't have to update every non-api spectator check.
2) APIs make code look cleaner, and reduces boilerplate code. (Readability)
3) Reduces the chance of having bugs.

> I never said it was noticeable

> but what if a server has 9000 players (on 1 PM server not realistic but theoretical) and the function is run for each of them? Every micro optimization counts ;)

You implied your intension pretty well. Performance issues don't come from calling a function.. they are mostly caused by file I/O, and big loops running expensive operations every tick. No micro optimization counts.

”We should forget about small efficiencies, say about 97% of the time: premature optimization is the root of all evil.” ~  DonaldKnuth

mal0ne-23: Jake stop being petty and just update what you did for fucks sake

ImagicalGamer:
Would you like me to squash the commits @dktapps? Or will you do it if/when it's merged?

jasonwynn10:
How does this PR deal with timings? Following the changes in the Living class, Timings is never stopped.

ImagicalGamer:
@jasonwynn10 I started and stopped timings just to test performance.

jasonwynn10:
@ImagicalGamer on line 268 of Living, timings is stopped before the function returns its value. Your change makes the function return without stopping Timings.

ImagicalGamer:
@jasonwynn10 I didn't notice that, I'll fix that now.

Tarik02:
Isn't better to use something like 'water proof' like https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/Player.php#L1373

dktapps:
The bubbles issue has already been resolved on another branch.

dktapps:
This is a hacky solution and contains multiple changes. This does not comply with the guidelines.

Issue #843, opened 2017-04-17
api3/network-mcpe-1.1 errors with texture shadings


wsngamerz:
### Issue description
As you can see from the image below, i have gotton alot of texture issues with a fresh install of pmmp. Just wanted to let you guys know.
![screenshot_20170417-143228](https://cloud.githubusercontent.com/assets/15077095/25090469/421fae7c-237b-11e7-94d5-7a867891b3ce.png)

Sandertv:
PocketMine isn't to blame for texture issues. Blame Mojang I'd say.

dktapps:
This is a client-sided issue with chunk replacement. It has been reported to Mojang.

wsngamerz:
but this issue only occurs on the pmmp server

dktapps:
MiNET also has these issues with rendering.

As for the black grass, that is due to a very basic sky light implementation. Previously to 1.1 sky light was not rendered like this by the game. I'm already aware of this problem and am working on a fix.

wsngamerz:
Ah okay, thanks

dktapps:
Previously-generated terrain will still be affected by the lighting problems, but as of commit dab73d8 the lighting problems should be resolved. The rendering glitches however will not.

PR #1100, opened 2017-06-16
Chunk of changes involving chunks

### LINKED ISSUE #530: Sometimes when you join chunks aren't loading
### Issue description
Sometimes chunks aren't loading when you spawn and you are moving very fast and when i try to /kill me i don't receive the RespawnPacket and i need to restart my game

1. ... Join the server 
2. ... See

### OS and versions

* PocketMine-MP: last api3\network commit
* PHP: 7.0.2
* Server OS: Debian
* Game version: PE/Win10 

* ...

xxFlare:
# About
This pull request fixes a good chunk of bugs involving chunks.

### What’s fixed/changed:
- Chunks no longer order twice during login.
- Fixed chunks ordering twice when Player::teleportImmediate() is called.
- Fixed (or hack-fixed) getting stuck during teleports. (That nasty bug where you are stuck in a teleport limbo and all you can see is void)
- This also fixes the respawn bug issue, where again once you click re-spawn all you see is void.
- PocketMine now corrects the value of the chunk population ~and generation~ queue size if its below one. This is because the server won’t work properly if ~these~ this value~s~ ~are~ is below one, and players won’t be able to spawn.
- Removed delayed teleports.

###  Performance improvements:
- Chunks are no longer ordered on a schedule. They will be ordered when a player moves around in relationship to your servers view distance.
- Chunks are no longer ordered twice on logins.

### Why change to a new non-schedule ordering system?
It’s a waste of CPU to order chunks on schedule. If you have a lot of players it could impact performance as well. The old system ordered chunks every 200 ticks if a player is idle, and if the player starts moving it would order chunks every 20 ticks. This caused orderChunks() to unnecessarily waste server CPU. Also idle players don't need there chunks ordered every 200 ticks.

### The new system:
The players view distance is taken (example: 8). It is then multiplied by 20 and that will be the players nextChunkOrderRun reset value. 8*20= 160, meaning each valid player movement will subtract this value by one or two if the player is sprinting, and when it reaches 0 chunks will be ordered. This system makes it so chunks order only as often as needed too, if the view distance is high, chunks won’t have to order as much (vice versa).

### Teleporting fixes:
This is more of a hack fix, so changes may need to be made here before merging. During delayed teleports, the $forceMovement Vector3 object interferes with the players teleport process. The client will start to spazz out as PocketMine is telling the client to go to different directions. This PR removes forceMovement from the teleport process. This may cause some problems, and unintended side-effects.

**Small cosmetic change: If check teleport position evaluates to true, spawnToAll does not need to be called again.**

### Immediate teleport:
Fixed a silly typo that ordered chunks twice.

### API changes:
Backwards compatible changes; plugins that are coded properly will not break. The two API changes are as follows:
- Added $orderChunks bool to teleport() function.
- Added $orderChunks bool to setViewDistance() function.
- Deprecated teleportImmediate().

 This PR may fix #530 joining's issue. Though I can confirm it fixes the respawning bug reported there. It also knocks off a item off #262 (non-scheduled ordering system). Current concerns are the effects of removing $forceMovement from the teleport process. WIP.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -273,7 +273,6 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	/** @var Vector3 */
 	protected $forceMovement = null;
 	/** @var Vector3 */
-	protected $teleportPosition = null;
 	protected $connected = true;
 	protected $ip;
 	protected $removeFormat = true;
@@ -853,7 +852,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 			$this->level->registerChunkLoader($this, $X, $Z, false);
 
 			if(!$this->level->populateChunk($X, $Z)){
-				if($this->spawned and $this->teleportPosition === null){
+				if($this->spawned){
 					continue;
 				}else{
 					break;
@@ -864,7 +863,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 			$this->level->requestChunk($X, $Z, $this);
 		}
 
-		if($this->spawned === false and $this->teleportPosition === null and count($this->usedChunks) >= $this->spawnThreshold){
+		if($this->spawned === false and count($this->usedChunks) >= $this->spawnThreshold){
 			$this->doFirstSpawn();
 		}
 
@@ -1497,7 +1496,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	}
 
 	protected function processMovement($tickDiff){
-		if(!$this->isAlive() or !$this->spawned or $this->newPosition === null or $this->teleportPosition !== null or $this->isSleeping()){
+		if(!$this->isAlive() or !$this->spawned or $this->newPosition === null or $this->isSleeping()){
 			return;
 		}
 
@@ -1712,8 +1711,6 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 				}
 			}
 		}
-
-		$this->checkTeleportPosition();
 
 		$this->timings->stopTiming();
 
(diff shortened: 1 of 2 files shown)


AppleDevelops:
Only one issue spotted so far, and that's when you log in, trees are partly invisible, the player breaks part of the tree, or reloads the chunk and the tree is then visible again.

ImagicalGamer:
@AppleDevelops isn't that a known issue? Trees generating in multiple chunks, It occasionally happens on the master branch.

AppleDevelops:
@ImagicalGamer It happens more often with this PR, I’d see it with 2 or 3 trees on the mater branch, and on this PR, it affected all trees.

Muqsit:
Bukkit much :p

legoboy0215:
I like the pun.

xxFlare:
@dktapps Would you prefer the delayed  teleport system be removed? Ironically it causes more problems then when PM just used teleport immediately. The logic of teleporting after all chunks are loaded sounds good on paper, but works horribly with the way PM locks players. Only time it may help is if chunks fail to load, but even then the player will still be locked into there old position leaking more CPU a tick checking chunks..

KennFatt:
its run really fast and clean, and i trying to teleport another world, then it happen to me:

screenshot 23
screenshot 19
screenshot 21
screenshot 22

AppleDevelops:
*facepalm*

MattWAnderson:
Check out the other issues ;) lol

xxFlare:
@KennFatt What's your servers view distance? Seems like it did not order chunks in time.

Muqsit:
For God's sake... It's not gonna send chunks *faster*. It's an efficient way of sending chunks.
It's like... saving money doesn't mean you are earning more money.

dktapps:
Since nobody figured out how to use the search box yet, here's an issue ticket number: #843

KennFatt:
haha lol, i thinks its related with chunk issue, but there is much issue with chunk so i posted it here. thanks guys for notice me

xxFlare:
Removed the delayed teleport system since most people seem to hate it, causes more bugs then it actually does anything (#1106).

Removing the delayed teleport system makes clients spawn into the game a bit faster too.

MCPE 0.10.5 and MCPE 0.11.0 hated being teleported into unloaded chunks. MCPE used to freeze and crash when teleported into a unloaded chunk. That is when shoghicp implemented this system during the 1.5 transition.

Newer MCPE versions don't crash and do not completely freeze the game when you are teleported in a unloaded chunk. I currently have this running on a production server with 300 players and have not got any complaints so far. Regardless, this needs more more testing and I still need to find a way to disallow dead people (modded clients) to move that does not interfere with respawning.

SuperMaXAleX:
Awesome work ! (But format correctly your code, there aren’t good on the 2 last commits)

xxFlare:
Updated description. This is ready for review.

Prisane:
Chunks are not loaded when after respawn you die. And sometimes chunks are not loaded on world. But its run really fast.
dasda

Misteboss:
@Prisane Confirmed, I have the same issue that you

xinghao2003:
So When can it be merge 😂

PR #1114, opened 2017-06-19
Added missing fields to server name for ping responses

dktapps:
## Introduction
This data is used to send back responses when a client pings the server, to show the server in the server list. However, there were 3 fields missing that vanilla MCPE sends that PocketMine-MP was not:

- world seed - Specified as `-1`, since with multiple worlds, which seed do you pick?
- host name - In vanilla this would be the name of the player hosting the multiplayer game.
- gamemode - Server global gamemode.

This pull request will also provide a quick way to identify a PocketMine-MP server based on its ping response.

There are some unconfirmed bugs with this such as servers appearing in the wrong place in the server list if the list contains servers which do not include these extra fields. Please test and review.

### API changes
- Added API method `Server::getGamemodeName(int $gamemode)`

### Behavioural changes
- Server ping response now includes previously missing fields.

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -440,6 +440,21 @@ class Server{
 		return "UNKNOWN";
 	}
 
+	public static function getGamemodeName(int $mode) : string{
+		switch($mode){
+			case Player::SURVIVAL:
+				return "Survival";
+			case Player::CREATIVE:
+				return "Creative";
+			case Player::ADVENTURE:
+				return "Adventure";
+			case Player::SPECTATOR:
+				return "Spectator";
+			default:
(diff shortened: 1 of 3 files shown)


robske110:
You could send the seed of the dedault level, but -1 is probably the best choice. And i think the gamemode function should be in Player because it is player-specific and also using Player constants. I sometimes am confused by the misc functions that have just been thrown into Server.

jojoe77777:
@robske110 the client shouldn't be able to see any world seeds, otherwise they can use tools to find certain generated structures

robske110:
Actually, you're right. The seed will be generated completely different if given to mcpe anyways.

PR #1115, opened 2017-06-19
Fixes MCPE syntax checked command parameters being ordered wrongly

Sandertv:
## Introduction

An issue exists with the ordering of command parameters when making use of the MCPE command parameter syntax checking, having to do with the ordering of parameters, because these are ordered alphabetically client sided. (Issue can be found mentioned here: https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/Player.php#L3335)

This issue aims to place the arguments back on their original position to fix the issue of commands being ordered alphabetically.

For clarification:
MCPE syntax checking devides the parameters in two groups, the optional ones and the non-optional ones. Both these groups get sorted alphabetically seperately, and get merged together after this has been done. 

### Relevant issues

Attempts to fix the issue mentioned here: https://github.com/BlockHorizons/BlockPets/issues/5, due to the parameters being ordered incorrectly and messing up things.

### API changes

No API changes.

### Behavioural changes

Now orders parameters correctly before building the command line.

## Backwards compatibility

There is no backwards incompatibility in this pull request.

## Follow-up

**Requires Testing**. I currently have no way to test this pull request, as my only laptop is a 32 bit one. I'll try to test this as soon as possible.

## Tests

Hopefully soon to come.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -3361,8 +3361,9 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 			$mergedOrderedParameters = array_merge($optionalParameters, $nonOptionalParameters);
 
 			$inputJson = $packet->inputJson;
+			$i = 0;
 			foreach($packet->inputJson as $key => $arg){
-				$inputJson[$mergedOrderedParameters[$key]["key"]] = $arg;
+				$inputJson[$mergedOrderedParameters[$i++][$key]["key"]] = $arg;
 			}
 			foreach($inputJson as $key => $arg){
 				$commandText .= " " . $arg;
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -3345,7 +3345,18 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 				$nonOptionalParameters[$key] = $overload;
 			}
 		}
-		$mergedOrderedParameters = array_merge(sort($optionalParameters), sort($nonOptionalParameters));
+		$optionalSortingArray = [];
+		foreach($optionalParameters as $key => $parameter){
+			$optionalSortingArray[$key] = $parameter["name"];
+		}
+		$nonOptionalSortingArray = [];
+		foreach($nonOptionalParameters as $key => $parameter){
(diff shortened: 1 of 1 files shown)


TheDiamondYT1:
I can test...use the latest pocketmine beforr 32-bit was dropped 😂

Misteboss:
Tested and doesn't work for me :/, good syntax doesn't appear and when i execute the command i have this error.
Tested with pocketmine commands.
`[21:53:38] [Server thread/CRITICAL]: ErrorException: "Undefined index: args" (EXCEPTION) in "/src/pocketmine/Player" at line 3071`

Sandertv:
Found an issue with it yeah, I put the check if inputJson is null at the wrong position. Will fix in a second.

Sandertv:
@Misteboss try now. For further testing, this should be done with a plugin using the method described here: https://forums.pmmp.io/threads/temporary-mcpe-command-syntax-checking.2320/ with a lot of arguments that are not in alphabetic order. (The name key)

Misteboss:
No change i have the same error now

Sandertv:
@Misteboss there's nothing in that code that uses an index called args. Only some spoons have $packet->args instead of $packet->inputJson.

Misteboss:
I am using pocketmine so..

Sandertv:
As I said, there's NO part of the code that could produce an error with args on that position.

Misteboss:
So I did some var_dump() to see what it gives me, this is the result :
capture d ecran 2017-06-22 a 12 47 28
capture d ecran 2017-06-22 a 12 48 35

Misteboss:
I don't know why @jasonwynn10 approved this PR if she doesn't work..

dktapps:
Just an FYI: This will be rendered obsolete when MCPE 1.2 rolls out due to a complete rewrite in the command handling system by Mojang.

Sandertv:
@Misteboss would you mind trying this out once more? I suspect it should work now.

Sandertv:
Please try the latest commit if you can @Misteboss. It should work.

Sandertv:
Oh. This is obsolete anyway once 1.1 comes out, which won't be too long. I can wait until then.

PR #1119, opened 2017-06-20
Changed CompoundTag access

### LINKED ISSUE #1118: Item->clearCustomBlockData() fails
### Issue description
https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/item/Item.php#L447

dktapps:
## Introduction
There is a lot of undefinedness around anything that uses NBT in the PocketMine-MP core. This pull request aims to reduce this problem by tightening the restrictions on the way CompoundTag children are accessed, and add methods for accessing them safely instead.

## Relevant issues
- fixed #1118 
- fixed villager profession doesn't make any difference client-side

### API changes
The method of assigning a child tag to a CompoundTag has changed. It is no longer legal to do the following:
- `$nbt->SomeTagName = new IntTag("SomeTagName", 5);`
- `unset($nbt->SomeTagName);`
- `$nbt->SomeTagName->getValue();`

You are instead encouraged to use the new CompoundTag API methods:
- `getTag(string $name)`
- `getCompoundTag(string $name)`
- `setTag(NamedTag $tag)`
- `exists(string $name)`
- `remove(string $name)`

This aims to reduce confusion between CompoundTag offsets and the tags' own names. For example, the following code:
`$nbt->SomeTagName = new IntTag("SomeOtherNameThatDoesntMatch", 5);`
will cause the IntTag to be saved with the name passed in the IntTag constructor, _not_ its field name in the CompoundTag. This duplication/inconsistency caused many bugs throughout the core code.

## Backwards compatibility
This is a backwards-compatibility break due to the API changes noted above.

## Tests
This has only been lightly tested (world loading, enchantments, spawning entities) but there are a lot of changes in this PR and I have no doubt I have introduced problems somewhere. Please test and review carefully.

Areas affected by these changes include:
- Level loading and saving (all formats)
- Entity loading and saving (all)
- Player data loading and saving
- Tile creation, spawning loading and saving.

### DIFF
--- a/src/pocketmine/entity/Entity.php
+++ b/src/pocketmine/entity/Entity.php
@@ -348,43 +348,45 @@ abstract class Entity extends Location implements Metadatable{
 		$this->server = $level->getServer();
 
 		$this->boundingBox = new AxisAlignedBB(0, 0, 0, 0, 0, 0);
+
+		/** @var ListTag $pos */
+		$pos = $this->namedtag->getTag("Pos");
+		/** @var ListTag $rotation */
+		$rotation = $this->namedtag->getTag("Rotation");
 		$this->setPositionAndRotation(
-			$this->temporalVector->setComponents(
-				$this->namedtag["Pos"][0],
-				$this->namedtag["Pos"][1],
-				$this->namedtag["Pos"][2]
-			),
-			$this->namedtag->Rotation[0],
-			$this->namedtag->Rotation[1]
+			$this->temporalVector->setComponents($pos[0], $pos[1], $pos[2]),
+			$rotation[0],
+			$rotation[1]
 		);
 
-
-		$this->setMotion($this->temporalVector->setComponents($this->namedtag["Motion"][0], $this->namedtag["Motion"][1], $this->namedtag["Motion"][2]));
+		/** @var ListTag $motion */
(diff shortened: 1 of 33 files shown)


jasonwynn10:
I believe this will need another API bump due to the amount of plugin use

Sandertv:
Agreed with @jasonwynn10, this breaks a LOT of plugins, and another API bump should definitely be considered...

dktapps:
well obviously... but you're supposed to use tags, not the master branch. We're due a release very soon anyway.

matcracker (approved):

src/pocketmine/entity/Entity.php
```diff
@@ -782,34 +783,37 @@ public function saveNBT(){
...
 				$this->setNameTagVisible($this->namedtag["CustomNameVisible"] > 0);
 			}
 		}
 
 		$this->scheduleUpdate();
 
-		if(isset($this->namedtag->ActiveEffects)){
-			foreach($this->namedtag->ActiveEffects->getValue() as $e){
-				$amplifier = Binary::unsignByte($e->Amplifier->getValue()); //0-255 only
+		$effectsTag = $this->namedtag->exists("ActiveEffects");
```
dktapps: oops

PR #1127, opened 2017-06-22
Ability to customize ID when registering Items

mal0ne-23:
## Introduction

This allows you to register custom items that aren't in PocketMine with ease, via plugins.

Let's say, you have a custom item class called "Orange Stick", and it has the same ID as BlazeRod. Registering it currently would just overwrite it.

But, if you were to do Item::registerItem(new OrangeStick(), 500);, you'd be able to do /give <player> 500, to give them an Orange Stick item. 

Am I making sense? This is necessary for plugins that want to register custom items that are using an item that is already registered.

### API changes

registerItem function in item\Item class now has one extra parameter, $extraId

### Behavioural changes

None.

## Backwards compatibility

Completely reverse compatible

## Tests

Yes this has been tested on my server. I modified it on my own, because I'm registering custom items, and the current API doesn't work for what I was doing before.
Code:
`
public function registerWeapons(){

	Item::registerItem(new BookOfSpells(), self::BOOK_OF_SPELLS);

	Item::registerItem(new ConcussionGrenade(), self::CONCUSSION_GRENADE);

	Item::registerItem(new BrassKnuckles(), self::BRASS_KNUCKLES);

	Item::registerItem(new Gun(), self::GUN);

	Item::registerItem(new ReflexHammer(), self::REFLEX_HAMMER);

	Item::registerItem(new Defibrillator(), self::DEFIBRILLATOR);

	Item::registerItem(new Syringe(), self::SYRINGE);

	Item::registerItem(new ThrowingKnife(), self::THROWING_KNIFE);

	Item::registerItem(new Shuriken(), self::SHURIKEN);

	Item::registerItem(new EnderPearl(), self::ENDER_PEARL);

	Item::registerItem(new Decoy(), self::DECOY);

	Item::registerItem(new FireCharge(), self::FIRE_CHARGE);

	Item::registerItem(new Flamethrower(), self::FLAMETHROWER);

	Item::registerItem(new MaloneSword(), self::MALONE_SWORD);

}`

### DIFF
--- a/src/pocketmine/item/Item.php
+++ b/src/pocketmine/item/Item.php
@@ -210,10 +210,11 @@ class Item implements ItemIds, \JsonSerializable{
 	 *
 	 * @param Item $item
 	 */
-	public static function registerItem(Item $item){
-		$existing = self::$list[$item->id] ?? [];
+	public static function registerItem(Item $item, $extraId = -1){
+		$id = ($extraId == -1 ? $item->id : $extraId);
+		$existing = self::$list[$id] ?? [];
 		$existing[$item->meta & 0xffff] = $item;
-		self::$list[$item->id] = $existing;
+		self::$list[$id] = $existing;
 	}
 
 	private static $creative = [];
--- a/src/pocketmine/item/Item.php
+++ b/src/pocketmine/item/Item.php
@@ -210,7 +210,7 @@ class Item implements ItemIds, \JsonSerializable{
 	 *
 	 * @param Item $item
 	 */
-	public static function registerItem(Item $item, $extraId = -1){
+	public static function registerItem(Item $item, int $extraId = -1){
 		$id = ($extraId == -1 ? $item->id : $extraId);
 		$existing = self::$list[$id] ?? [];
 		$existing[$item->meta & 0xffff] = $item;


TheDiamondYT1:
You're registering a custom item class...so why not just override the id in the class?

dktapps:
This does not make logical sense, you should set the ID when you construct the item.

mal0ne-23:
When you set the ID when constructing, it changes what item it is when it shows. What if someone wants a new Book class, but without completely overwriting the already registered book?

dktapps:
Then you give it a different ID in the constructor, as I already said. If you want a custom book with ID 500, you pass it in the constructor. How you would make that work without client modifications is not my concern

dktapps:
Unless you plan to somehow spoof the item ID to make this work, I don't see what purpose it serves. There are better ways to achieve this. Having an item with one ID registered under a different ID will cause lots of bugs.

mal0ne-23:
Ah. okay.

mal0ne-23:
Sorry

dktapps:
No need to be sorry, it's an interesting idea. However this is a bad way to achieve this goal. I'll see what I can come up with.

PR #1129, opened 2017-06-23
requestChunkTask exception handling

xxFlare:
### Explanation
requestChunkTask throws an exception, since the exception is not caught and handled, chunks will never leave the sendChunkTasks and chunkSendQueue array, forever looping around each tick / not releasing memory.

Found this when I was using a corrupt world.

I doubt try/catch will effect performance in any way

### API changes
None.

### Changes
- Properly handle ChunkException with try/catch

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -68,6 +68,7 @@ use pocketmine\item\Item;
 use pocketmine\level\format\Chunk;
 use pocketmine\level\format\io\BaseLevelProvider;
 use pocketmine\level\format\io\LevelProvider;
+use pocketmine\level\format\io\ChunkException;
 use pocketmine\level\generator\GenerationTask;
 use pocketmine\level\generator\Generator;
 use pocketmine\level\generator\GeneratorRegisterTask;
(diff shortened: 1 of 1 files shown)


dktapps:
This isn't really a "fix", it's just exception handling.

src/pocketmine/level/Level.php
```diff
@@ -2385,9 +2386,13 @@ private function processChunkRequest(){
 					continue;
 				}
 				$this->timings->syncChunkSendPrepareTimer->startTiming();
-				$task = $this->provider->requestChunkTask($x, $z);
-				if($task !== null){
-					$this->server->getScheduler()->scheduleAsyncTask($task);
+				try{
+					$this->server->getScheduler()->scheduleAsyncTask($this->provider->requestChunkTask($x, $z));
+				}
+				catch(ChunkException $e){
```
dktapps: ...

dktapps:
I've never seen this exceptiom thrown before, would you mind telling how this occurred in more detail? Corrupted chunks would be regenerated under normal circumstances.

dktapps:
...

PR #1140, opened 2017-06-26
Possible solution for #1134

### LINKED ISSUE #1134: No way to set custom username rules
### Issue
There is currently no way to set custom username rules. Player::isValidUserName uses hard coded rules no one can edit. It would be great if there was a way server owners can use a API to implement there own username policies without having to override the function by extending the player class. Or, possibly changing the $this->close to kick, so owners can cancel PlayerKickEvent.

xxFlare:
## Introduction
Allow custom username policies, you can now choose to cancel PlayerKickEvent to implement your own username policies.

### Relevant issues
Before it was impossible for plugins to allow certain usernames to be joinable via API.

Closes #1134 

### API changes
Only additions, properly coded plugins will not break.

## Follow-up
Perhaps change more Player::close calls to Player::kick?

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -1953,8 +1953,9 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 		$this->rawUUID = $this->uuid->toBinary();
 
 		if(!Player::isValidUserName($packet->username)){
-			$this->close("", "disconnectionScreen.invalidName");
-			return true;
+			if($this->kick("disconnectionScreen.invalidName", false)){
+				return true;
+			}
 		}
 
 		if(!Player::isValidSkin($packet->skin)){


Muqsit:
This can be harmful though, pretty much like SQL injections.

xxFlare:
@Muqsit That is very true. Though, plugin creators should use prepared statements.

Muqsit:
@xxFlare Agree, but it might make some things impossible. For example, if a player has a space in their name, and you'd like to /give <player> something, I don't think you can use the player's fullname in this case. But then again, this promotes vanilla behavior..

TheDiamondYT1:
@Muqsit Like in bash, add support for quotation marks in commands?

**Bash**
```bash
cp "A Folder With Spaces" dest
```

Muqsit:
@TheDiamondYT1 its gonna be too much for a little change. It's better to let it be `ctype_alnum()`, and this is just one of the issues that would arise with this change.

PR #1144, opened 2017-06-27
Added API method to allow changing /help page sizes for CommandSenders

dktapps:
## Introduction
Allows plugins to alter the page size for clients shown when using /help.

Using /help on a client can get irritating, especially if you have a larger screen. This change allows plugins to set a custom /help page size which suits everyone. I currently just set it to `PHP_INT_MAX` since the scrollable chat window makes it unnecessary to have multiple pages.

### API changes
- added API method `CommandSender->getScreenLineHeight() : int`
- added API method `CommandSender->setScreenLineHeight(int)`

## Tests
tested very quickly, probably ought to check for negative page sizes.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -328,6 +328,9 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	/** @var PermissibleBase */
 	private $perm = null;
 
+	/** @var int|null */
+	protected $lineHeight = null;
+
 	public function getLeaveMessage(){
 		return new TranslationContainer(TextFormat::YELLOW . "%multiplayer.player.left", [
 			$this->getDisplayName()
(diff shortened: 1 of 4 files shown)


SOF3:
This does not conveniently allow setting for different command senders. What about allowing a callable of the prototype `function(CommandSender)`?

SOF3 (changes requested):
This does not conveniently allow setting for different command senders. What about allowing a callable of the prototype `function(CommandSender) : int`?

Also, making command output pagination an API function that plugins can use is also useful.

dktapps:
@SOF3 by all means, knock yourself out, but this is a simple change to fix a simple irritation. Console has always used PHP_INT_MAX anyway.

SOF3 (approved):
Warning: this is a backwards-incompatible API change. If you are sure you don't want to maintain BC by making an extra interface that declares the new method, this PR is OK :+1:

PR #1147, opened 2017-06-28
Fix drop item behaviour make compatible with pc version

eternalharvest:
## Introduction

Support the behaviour dropping item one by one, for the compatibility to the pc version.

I'm currently working on the BigBrother project. And I want to implement the dropping item behavior. Probably, Minecraft PE version do not support dropping item one by one, so that ppmp interpret `DropItemPacket` as dropping whole item. But this behaviour is not compatible with pc version of minecraft. So, I wrote this patch to solve this problem.

### Behavioural changes
decrement the counter of item in hand by `DropItemPacket#item->count`. `PlayerInventory#setItem()` clear inventory (holding item) if item count is 0 or negative value. this patch do not affect any PE version of minecraft client.

## Tests

I was tested this patch on the iOS version, and PC version on Mac with BigBrother (my code base).

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -2877,7 +2877,17 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 			return true;
 		}
 
-		$this->inventory->setItemInHand(Item::get(Item::AIR, 0, 1));
+		if($packet->item->getCount() < $item->getCount()){
+			// PC edition users may drop any number of item they have
+			$item->setCount($item->getCount() - $packet->item->getCount());
+			$this->inventory->setItemInHand($item);
+
+			// do not trust item id originates from incomming packet
+			$item->setCount($packet->item->getCount());
+		}else{
+			$this->inventory->setItemInHand(Item::get(Item::AIR, 0, 1));
+		}
+
 		$motion = $this->getDirectionVector()->multiply(0.4);
 
 		$this->level->dropItem($this->add(0, 1.3, 0), $item, $motion, 40);


eternalharvest:
force updated the patch to make it more simple.

jojoe77777:
You can't trust the client to provide the correct item

eternalharvest:
Thank you for the reviewing my code.

I has rewritten the code to avoid exploit vulnerability. Even if any mulcious user send DropItemPacket with wrong item id, anyone can't drop item not holding in hand. And also, maximum number of dropped item is limited to the number holding in hand.

I think, problem was fixed. But If anyone found another problem please tell me.

jojoe77777:
It still doesn't check the id.

eternalharvest:
u are right, but the variable `$item` is originally retrieved from the function `PlayerInventory#getItemInHand()`, so I think there is no problem.

eternalharvest:
I rewrote it again because the code was difficult to understand.

Muqsit:
Use Item::getCount(), I don't think Item::$count is supposed to be public as it isn't directly being used anywhere.
Btw nice work.

eternalharvest:
I know what u mean, and I also had worried about that.
But I had found the code that decrements `Item#count`, so i wrote the code decrement `Item#count` directory. Please, refer to the following link.

https://github.com/pmmp/PocketMine-MP/blob/70bd9afd3789673ec41bf257c4194f0a5b343a35/src/pocketmine/Player.php#L2703-L2704

Anyway, I'll try to rewrite it.
I wonder that why `Item#count` is public...

eternalharvest:
I rewrote it again.
Please tell me if any problem is there.

eternalharvest:
Perhaps, this function should be implemented with a BigBrother plug-in...

But I can't implement this function because I can't access `DropItemPacket#count` through `PlayerDropItemEvent` event api. Should I suggest api changes about `PlayerDropItemEvent` event to implement this function?

eternalharvest:
@hmy2001 who is the maintainer of BigBrother plugin solve this issue within the plugin.
So, this pull-request is no longer needed.
Thank you for discussing on my pull-request.
I close this pull-request.

Misteboss:
@TheDiamondYT1 He said maintainer, it's time to buy glasses.

PR #1152, opened 2017-06-30
Allow spaces in player's username

sn3akrr:
## Introduction

XBOX Live gamertags support spaces. Without this, it limits gameplay to some players logged into xbox live.

### Behavioral Changes
Allows players with gamertags with spaces to join

## Follow-up
Finish custom command data, so players with spaced names can be called correctly in commands

## Tests

Try to join the server with a username that contains spaces, such as "Name With Spaces" xD

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -225,7 +225,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	public static function isValidUserName(string $name) : bool{
 		$lname = strtolower($name);
 		$len = strlen($name);
-		return $lname !== "rcon" and $lname !== "console" and $len >= 1 and $len <= 15 and stristr($name, "\n") == false and stristr($name, "\t") == false and preg_match("/[^A-Za-z0-9_\s]/", $name) === 0;
+		return $lname !== "rcon" and $lname !== "console" and $len >= 1 and $len <= 15 and stristr($name, "\n") === false and stristr($name, "\t") === false and preg_match("/[^A-Za-z0-9_\s]/", $name) === 0;
 	}
 
 	/**
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -225,7 +225,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	public static function isValidUserName(string $name) : bool{
 		$lname = strtolower($name);
 		$len = strlen($name);
-		return $lname !== "rcon" and $lname !== "console" and $len >= 1 and $len <= 16 and preg_match("/[^A-Za-z0-9_]/", $name) === 0;
+		return $lname !== "rcon" and $lname !== "console" and $len >= 1 and $len <= 16 and preg_match("/[^A-Za-z0-9_\s]/", $name) === 0;
 	}
 
 	/**
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -225,7 +225,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	public static function isValidUserName(string $name) : bool{
 		$lname = strtolower($name);
 		$len = strlen($name);
-		return $lname !== "rcon" and $lname !== "console" and $len >= 1 and $len <= 16 and preg_match("/[^A-Za-z0-9_\s]/", $name) === 0;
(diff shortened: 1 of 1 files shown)


jojoe77777:
Good luck with commands.

mal0ne-23:
Oh yeah.. That should be done before this is pulled. The command data support..

mal0ne-23:
Eventually this will need to be supported, like this isn't a pointless update. Once XBOX Live is enforced, players with gamertags like "monkey man" (god i wish i had that gamertag) wouldn't be able to player on PocketMine servers, due to their username containing spaces.

TheDiamondYT1:
Maybe commands with spaces in player names could be done like this:a

In bash, you can can copy multiple directories to a single destination directory in one command, but without quotation marks.
```bash
cp -rf dir1 dir2 dir3 destination
``` 
The above executes correctly. 

Since command arguments are an array, we can use [array_slice()](http://php.net/manual/en/function.array-slice.php) to get the last few arguments of an array. For example:
```php
# Command
/tp player1 player2 x y z

# In the above example, we can shift the last 3 arguments off the array, and use a foreach loop on the players.
```
This could impact on performance, probably, maybe. I dont know.

dktapps:
XBOX won't be enforced, I don't know why people seem to think it will be. Enforcing XBOX login is the server's decision.

Regardless, MCPE does indeed support names with spaces in them, however PocketMine-MP's command parser does not provide the capability to handle this yet. I'm holding off on command rewrites until 1.2 because a rewrite of commands is coming from Mojang that will be less of a pain in the ass to deal with than the system we have now.

@TheDiamondYT1 just shut up about performance already. It's a damned command. So what? you might lose a couple of nanoseconds? well, that really sucks for those poor nanoseconds, but nobody else is going to care.

mal0ne-23:
I don't mean XBOX being enforced on the server. I mean MCPE forcing it. You aren't allowed to logout of XBOX Live on Windows 10. And Xbox One and 360 players are going to be stuck with their gamertag on servers.

Of course the server owner can choose whether or not to check for XBOX live login. That doesn't have to be forced. But, idk.. Not sure if you were talking about the PocketMine server or MCPE enforcing Live in your comment @dktapps, correct me if I'm wrong.

dktapps:
@mal0ne-23 remind me what version of MCW10 you're using... it stopped forcing XBOX login back at 0.16. But I take your point.

Regardless, this PR isn't about whether to support XBOX players or not, it's whether to allow spaces in usernames. That's a completely independent problem.

mal0ne-23:
Ah, nvm it doesn't force XBOX login. But yeah, consoles will most likely force it. Another issue though for another time.

VCraftMCPE:
@Dktapps in 1.2 update u will be forced to use XBox login to join servers

Driesboy:
rlly? why??

Op 2-aug.-2017 22:50 schreef "VCraftMCPE" <notifications@github.com>:

dktapps:
Yes, we are aware. That's irrelevant to this pull request.

dktapps (changes requested):
This will also allow players with new-lines and tab characters in their names to join the server. Possibly not expected behaviour.

Muqsit:
Change the `<= 16` to `<= 15` while you're at it.

sx

mal0ne-23:
done

jojoe77777:
Now players can't connect with a 16 char name to local servers over LAN

mal0ne-23:
Yeah.. Because you can totally do that when xbox live is enforced.

jojoe77777:
@mal0ne-23 Xbox live login isn't enforced over LAN

src/pocketmine/Player.php
```diff
@@ -225,7 +225,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	public static function isValidUserName(string $name) : bool{
 		$lname = strtolower($name);
 		$len = strlen($name);
-		return $lname !== "rcon" and $lname !== "console" and $len >= 1 and $len <= 16 and preg_match("/[^A-Za-z0-9_\s]/", $name) === 0;
+		return $lname !== "rcon" and $lname !== "console" and $len >= 1 and $len <= 15 and stristr($name, "\n") == false and stristr($name, "\t") == false and preg_match("/[^A-Za-z0-9_\s]/", $name) === 0;
```
TheDiamondYT1: `===` or just `!`

TheDiamondYT1 (changes requested):

Kenner-G:
@dktapps 

So as what I understand, Xbox Live will not be enforced in the 1.2 update correct?

Sandertv:
In 1.2 xbox live accounts will already be enforced.

SuperMaXAleX:
_facepalm_

PR #1160, opened 2017-07-01
Stop timing out players before they can join

madecode15:
### Introduction
This code was causing players to timeout before they can actually join the server. This helps players with slower networks, and sometimes when the server is unable to keep up.

Also, this code isn't really needed because players will automatically timeout when PocketMine stops receiving packets from them.

Lastly, this code never considers the servers TPS. If your TPS plummets, this code won't do any good because it measures a players login time difference with `microtome(true)`, not compensating the fact that the server could be behind a lot of ticks. Older versions of PM never had this code, and never had a big issue with login timeouts.

Tested & result:
Players who were complaining about getting login timeouts are now able to our servers after I removed this.

This is only one way to solve the issue. I'm sure there are better ways to do this then to completely nuke it (maybe use actual server tick time?).

### API changes:
None

### Performance improvements:
None

### Fixes
Allow bad connections to join & help players join when TPS is taking a nose dive.

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -2271,9 +2271,7 @@ class Server{
 
 	private function checkTickUpdates($currentTick, $tickTime){
 		foreach($this->players as $p){
-			if(!$p->loggedIn and ($tickTime - $p->creationTime) >= 10){
-				$p->close("", "Login timeout");
-			}elseif($this->alwaysTickPlayers and $p->joined){
+			if($this->alwaysTickPlayers and $p->joined){
 				$p->onUpdate($currentTick);
 			}
 		}


dktapps:
This will also cause problems with incompatible players when significant protocol changes are made which make the new client unrecognizable. The RakNet layer will still send and receive everything as normal but the MCPE layer will just sit... and wait... and wait... and wait...

madecode15:
@dktapps Maybe using server tick time would be a better solution.. hmm..

dktapps:
Not really. The client will still send the Login within a reasonable space of real time, it won't respect server ticks.

dktapps:
You should also be aware that players with bad connections will put significant extra load on your server.

mal0ne-23:
I've never got timed out while connecting to a pocketmine server.. Might be time to get better wifi.

madecode15:
@mal0ne-23 🤦‍♂️ 

1) Not all server players don't have good wifi
2) This only becomes an issue when you have more than 70 players online

madecode15:
@SuperMaXAleX PocketMine can handle 50-60 players without losing performance, like I said it is an issue when you have more than 70-90 players on.

madecode15:
This solution is not valid due because what dktapps said earlier is true

PR #1184, opened 2017-07-09
Update Block.php

DavidBrockway:
Implemented the SHULKER BOX blocks tested and works.


Sandertv:
Interesting how you can mess up the formatting by adding only one line...

DavidBrockway:
How does it mess up the formating i have checked it against mcpe normal blocks and its in the right order?

PR #1188, opened 2017-07-11
Ignore Hotbar Slots with Air

junkdude:
## Introduction

This fixes drag and dropping on w10 when taking items to/from the hotbar. 

### API changes

None
### Behavioural changes

None
## Backwards compatibility

Not sure

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -2282,6 +2282,13 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 			$item = $this->inventory->getItem($packet->inventorySlot);
 
 			if(!$item->equals($packet->item)){
+				if($item->getId() === 0) {
+					return false;
+				}
+
+				if($packet->item->getId() === 0) {
+					return false;
+				}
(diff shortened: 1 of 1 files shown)


dktapps:
I'd expect some kind of explanation to demonstrate you actually have a clue what you're doing and not just hacking it.

junkdude:
Okay, so if you try to shift-click an item from the hotbar on W10 it wont work because the air id and the $item id are two different well id's. This fixes this by ignoring empty slots or slots that have "air" in them.

dktapps:
I'm afraid I don't see what you are driving at. Shift-clicking things in and out of the hotbar works perfectly fine.

junkdude:
Not for me, it ends up saying tried to equip air but have diamond sword in target slot

junkdude:
Keep in mind im on classic ui

dktapps:
I remain unconvinced. Shift-clicking is treated just like any other inventory transaction. I cannot reproduce the bug you're talking about.

junkdude:
My bad i meant drag and dropping

dktapps:
That particular issue has nothing to do with this handler, it's an issue with transaction processing. This will be resolved in the 1.2 inventory rewrite.

junkdude:
Mkay

Issue #1021, opened 2017-06-06
The BossEventPacket is missing 4 breaks


intagaming:
### Issue description

src\pocketmine\network\mcpe\protocol\BossEventPacket.php
As title, in the decode() and encode(). This causes TYPE_SHOW to run as 3 types including itself.

### Steps to reproduce the issue

1. Open file
2. Navigate to the function decode() and encode(), 

### OS and versions

* PocketMine-MP: 29e88d8
* PHP: 7.0
* Server OS: Any
* Game version: PE/Win10

### Crashdump, backtrace or other files

* ...

intagaming:
Works as intended only if you have provided the unknownShort for it. So if I only want to show it why do I need to do the UNKNOWN_6 or the custom "TEXTURE" ?

dktapps:
It is written like that by design. You understand how a case statement works, yes?

intagaming:
That's confusing without an documentation. The default value is not there to be created with the boss bar. I can't seem to find anywhere the API instruction for an appropriate way to initialize a BossEventPacket, without posting a forum topic.

TheDiamondYT1:
@intagaming Post a forum topic...thats what the forum is for.

https://forums.pmmp.io

PR #1191, opened 2017-07-11
Fix BossEventPacket

IrishPacks:
## Introduction

I was trying to send this packet to a player, but kept getting errors about parameters being null and what not, so I looked into it and I saw that two breaks were missing in the switch statements for encode and decode.

### Relevant issues

Not any that I could find, but it does fix an issue with sending BossEventPacket

### API changes
N/A

### Behavioural changes
N/A

## Backwards compatibility
Compatible with the api3/blocks branch

## Follow-up
Whether this was intended by Dylan or not is beyond me, but I believe it was an accident.


SOF3:
See #1021 &mdash; Don't blindly believe everything your PHP tutorial and/or your PHP IDE tells you.

IrishPacks:
For one, I didn't follow everything my PHP tutorial and my IDE told me, I just didn't believe that you needed to run 3 types when showing it to a player, my fault.

PR #1195, opened 2017-07-12
fixing getTile

opkiler22789:
## Introduction
` Argument 2 passed to pocketmine\level\format\Chunk::getTile() must be of the type integer, string given, called in /root/aquamine/src/pocketmine/level/Level.php on line 1971`

## Tests
no errors

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -1968,7 +1968,7 @@ class Level implements ChunkManager, Metadatable{
 		$chunk = $this->getChunk($pos->x >> 4, $pos->z >> 4, false);
 
 		if($chunk !== null){
-			return $chunk->getTile($pos->x & 0x0f, $pos->y, $pos->z & 0x0f);
+			return $chunk->getTile((int) $pos->x & 0x0f, (int) $pos->y, (int) $pos->z & 0x0f); 
 		}
 
 		return null;


dktapps:
I don't see what there is to fix... this is caused by plugins not juggling their types correctly.

opkiler22789:
@dktapps 
```
 public $frames = [
        1 => [
            "vector" => "493:73:291",
            "activated" => false,
        ],
        2 => [
             "vector" => "493:72:291",
            "activated" => false,
        ],
        3 => [
             "vector" => "493:71:291",
            "activated" => false,
        ],
        4 => [
             "vector" => "493:73:290",
            "activated" => false,
        ],
        5 => [
             "vector" => "493:72:290",
            "activated" => false,
        ],
        6 => [
             "vector" => "493:71:290",
            "activated" => false,
        ],
        7 => [
             "vector" => "493:73:289",
            "activated" => false,
        ],
        8 => [
             "vector" => "493:72:289",
            "activated" => false,
        ],
        9 => [
             "vector" => "493:71:289",
            "activated" => false,
        ],
        10 => [
             "vector" => "493:72:288",
            "activated" => false,
        ]
    ];
 public function getTileByFrameID(int $id) : ItemFrame{
        $vector = explode(':', $this->frames[$id]['vector']);
        return $this->getServer()->getDefaultLevel()->getTile(new Vector3($vector[0], $vector[1], $vector[2]));
    }
```
Did I do something wrong? Correction that I propose, corrects

dktapps:
Did you even read what I said?

opkiler22789:
@dktapps Yes, did you mean that the plugin does not correctly manipulate their types? And how much I understood that that not so in the plug-in

KnownUnown:
@opkiler22789 he means you should cast the types in the function, as you have knowledge of what the correct type should be, like so:
```php
public function getTileByFrameID(int $id) : ItemFrame{
        $vector = explode(':', $this->frames[$id]['vector']);
        return $this->getServer()->getDefaultLevel()->getTile(new Vector3((int) $vector[0], (int) $vector[1], (int) $vector[2]));
}
```
This is not a fault of the API, since developers should always pass correct types on their end. You were passing a string to `Vector3` in that function (which itself does not have typehints due to problems of distinguishing ints and floats).

opkiler22789:
@KnownUnown Thanks, absolutely forgot about the types of values

PR #1196, opened 2017-07-12
Changes username validation regex

Noobidoo:
## Introduction
Currently not all valid usernames are able to join. Noticed as i have a space in mine.
According to xbox.com your gamer tag has to abide to these rules
"You can use up to 15 characters: Aa-Zz, 0-9, and single spaces. It can’t start with a number and can’t start or end with a space."

### Behavioural changes
Changed regex in username validation to follow xbox.com guidlines

## Follow-up
Test that regex works as i only did test in a [regex tester](https://regex101.com/r/9nOPbK/1).


KnownUnown:
Doesn't work for me - only matches consecutive spaces.

SuperMaXAleX:
It's almost the same as #1152 but with restriction

dktapps:
Duplicate of #1152

PR #1202, opened 2017-07-13
Improve usability of Sign API

robske110:
## Introduction

This makes all methods available on every Sign related class. It was quite weird to have setLines in Sign but not in SignChangeEvent and setLine&getLine in SignChangeEvent and not in Sign.

### Relevant issues
none

### API changes

Added setLines() to SignChangeEvent.
Added setLine() and getLine to tile/Sign.

### Behavioural changes

none

## Backwards compatibility
doesn’t break anything

## Tests

You can test this using [this](https://github.com/pmmp/PocketMine-MP/files/1156309/SignAPIfunctions.php.zip) Script Plugin.

### DIFF
--- a/src/pocketmine/event/block/SignChangeEvent.php
+++ b/src/pocketmine/event/block/SignChangeEvent.php
@@ -69,6 +69,9 @@ class SignChangeEvent extends BlockEvent implements Cancellable{
 	 * @return string
 	 */
 	public function getLine($index){
+		if($index < 0 or $index > 3){
+			throw new \InvalidArgumentException("Index must be in the range 0-3!");
+		}
 		return $this->lines[$index];
 	}
 
@@ -76,6 +79,9 @@ class SignChangeEvent extends BlockEvent implements Cancellable{
 	 * @param string[] $lines
 	 */
 	public function setLines(array $lines){
+		if(count($lines) > 4){
+			throw new \InvalidArgumentException("Array size cannot be greater than 4!");
+		}
 		$this->lines = $lines;
 	}
 
@@ -84,6 +90,9 @@ class SignChangeEvent extends BlockEvent implements Cancellable{
 	 * @param string $line
 	 */
 	public function setLine($index, $line){
+		if($index < 0 or $index > 3){
+			throw new \InvalidArgumentException("Index must be in the range 0-3!");
+		}
 		$this->lines[$index] = $line;
 	}
 }
--- a/src/pocketmine/event/block/SignChangeEvent.php
+++ b/src/pocketmine/event/block/SignChangeEvent.php
@@ -80,7 +80,7 @@ class SignChangeEvent extends BlockEvent implements Cancellable{
 	 */
 	public function setLines(array $lines){
 		if(count($lines) !== 4){
-			throw new \InvalidArgumentException("Array size cannot be greater than 4!");
+			throw new \InvalidArgumentException("Array size must be 4!");
 		}
 		$this->lines = $lines;
 	}
--- a/src/pocketmine/event/block/SignChangeEvent.php
+++ b/src/pocketmine/event/block/SignChangeEvent.php
@@ -73,6 +73,13 @@ class SignChangeEvent extends BlockEvent implements Cancellable{
 	}
 
 	/**
+	 * @param array string[]
+	 */
+	public function setLines(array $lines){
+		$this->lines = $lines;
+	}
+
+	/**
 	 * @param int    $index 0-3
 	 * @param string $line
 	 */
--- a/src/pocketmine/event/block/SignChangeEvent.php
+++ b/src/pocketmine/event/block/SignChangeEvent.php
@@ -79,7 +79,7 @@ class SignChangeEvent extends BlockEvent implements Cancellable{
 	 * @param string[] $lines
 	 */
 	public function setLines(array $lines){
-		if(count($lines) > 4){
+		if(count($lines) !== 4){
 			throw new \InvalidArgumentException("Array size cannot be greater than 4!");
 		}
 		$this->lines = $lines;
--- a/src/pocketmine/event/block/SignChangeEvent.php
+++ b/src/pocketmine/event/block/SignChangeEvent.php
@@ -73,7 +73,7 @@ class SignChangeEvent extends BlockEvent implements Cancellable{
 	}
 
 	/**
-	 * @param array string[]
+	 * @param string[] $lines
 	 */
 	public function setLines(array $lines){
 		$this->lines = $lines;
--- a/src/pocketmine/tile/Sign.php
+++ b/src/pocketmine/tile/Sign.php
@@ -71,6 +71,9 @@ class Sign extends Spawnable{
 	 * @param bool $update
 	 */
 	public function setLine(int $index, string $line, bool $update = true){
+		if($index < 0 or $index > 3){
+			throw new \InvalidArgumentException("Index must be in the range 0-3!");
+		}
 		$this->namedtag["Text".($index + 1)] = $line;
 		if($update){
 			$this->onChanged();
@@ -83,6 +86,9 @@ class Sign extends Spawnable{
 	 * @return string
 	 */
 	public function getLine(int $index): string{
+		if($index < 0 or $index > 3){
+			throw new \InvalidArgumentException("Index must be in the range 0-3!");
+		}
 		return $this->namedtag["Text".($index + 1)];
 	}
 
--- a/src/pocketmine/tile/Sign.php
+++ b/src/pocketmine/tile/Sign.php
@@ -71,7 +71,7 @@ class Sign extends Spawnable{
 	 * @param bool $update
 	 */
 	public function setLine(int $index, string $line, bool $update = true){
-		$this->namedtag["Text".$index + 1] = $line;
+		$this->namedtag["Text".($index + 1)] = $line;
 		if($update){
 			$this->onChanged();
 		}
@@ -83,7 +83,7 @@ class Sign extends Spawnable{
 	 * @return string
 	 */
 	public function getLine(int $index): string{
-		return $this->namedtag["Text".$index + 1];
+		return $this->namedtag["Text".($index + 1)];
 	}
 
 	public function getText(){
--- a/src/pocketmine/tile/Sign.php
+++ b/src/pocketmine/tile/Sign.php
@@ -74,7 +74,7 @@ class Sign extends Spawnable{
 		if($index < 0 or $index > 3){
 			throw new \InvalidArgumentException("Index must be in the range 0-3!");
 		}
-		$this->namedtag["Text".($index + 1)] = $line;
+		$this->namedtag["Text" . ($index + 1)] = $line;
 		if($update){
 			$this->onChanged();
 		}
@@ -85,11 +85,11 @@ class Sign extends Spawnable{
 	 *
 	 * @return string
 	 */
-	public function getLine(int $index): string{
+	public function getLine(int $index) : string{
 		if($index < 0 or $index > 3){
 			throw new \InvalidArgumentException("Index must be in the range 0-3!");
 		}
-		return $this->namedtag["Text".($index + 1)];
+		return $this->namedtag["Text" . ($index + 1)];
 	}
 
 	public function getText(){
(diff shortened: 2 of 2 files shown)


TheDiamondYT1:
> You can test this using this Script Plugin.
The mysterious Script Plugin that does not exist! :0

robske110:
It does now

src/pocketmine/event/block/SignChangeEvent.php
```diff
@@ -73,6 +73,13 @@ public function getLine($index){
 	}
 
 	/**
+	 * @param array string[]
+	 */
+	public function setLines(array $lines){
+		$this->lines = $lines;
```
SOF3: Should this check the array size?

src/pocketmine/event/block/SignChangeEvent.php
```diff
@@ -73,6 +73,13 @@ public function getLine($index){
 	}
 
 	/**
+	 * @param array string[]
```
SOF3: Shouldn't this be `@param string[] $lines`?

src/pocketmine/tile/Sign.php
```diff
@@ -64,6 +64,27 @@ public function setText($line1 = "", $line2 = "", $line3 = "", $line4 = ""){
 
 		return true;
 	}
+	
+	/**
+	 * @param int    $index 0-3
+	 * @param string $line
+	 * @param bool $update
```
SOF3: Inconsistent PHPdoc formatting

src/pocketmine/tile/Sign.php
```diff
@@ -64,6 +64,27 @@ public function setText($line1 = "", $line2 = "", $line3 = "", $line4 = ""){
```
SOF3: I would propose an API change, where `setText()` accepts `null` by default and only changes the lines if it is not null.

SOF3 (changes requested):

src/pocketmine/tile/Sign.php
```diff
@@ -71,6 +71,9 @@ public function setText($line1 = "", $line2 = "", $line3 = "", $line4 = ""){
 	 * @param bool $update
 	 */
 	public function setLine(int $index, string $line, bool $update = true){
+		if($index < 0 && $index > 3){
```
robske110: Kill me... && facepalm

src/pocketmine/tile/Sign.php
TheDiamondYT1: @SOF3 It's funny how hard it is to actually get such a simple commit right... i do it too

src/pocketmine/event/block/SignChangeEvent.php
robske110: probably

src/pocketmine/tile/Sign.php
robske110: I could include that here, as this is a general "improve sign api" thing

src/pocketmine/event/block/SignChangeEvent.php
TheDiamondYT1: Ummm.... This should be  `@param string[] $lines`

TheDiamondYT1: Ummm.... This should be  `@param string[] $lines`

src/pocketmine/event/block/SignChangeEvent.php
```diff
@@ -73,6 +73,13 @@ public function getLine($index){
 	}
 
 	/**
+	 * @param string[] array
+	 */
+	public function setLines(array $lines){
```
TheDiamondYT1: `@param string[] $lines`

TheDiamondYT1 (changes requested):

robske110:
okay, now everything should be resolved

src/pocketmine/event/block/SignChangeEvent.php
robske110: wtf did i do...

src/pocketmine/event/block/SignChangeEvent.php
```diff
@@ -73,6 +73,13 @@ public function getLine($index){
 	}
 
 	/**
+	 * @param string[] $lines
+	 */
+	public function setLines(array $lines){
+		$this->lines = $lines;
```
SOF3: What about `assert(count($lines) === 4);` here?

src/pocketmine/tile/Sign.php
```diff
@@ -55,15 +55,42 @@ public function saveNBT(){
 		unset($this->namedtag->Creator);
 	}
 
-	public function setText($line1 = "", $line2 = "", $line3 = "", $line4 = ""){
-		$this->namedtag->Text1 = new StringTag("Text1", $line1);
-		$this->namedtag->Text2 = new StringTag("Text2", $line2);
-		$this->namedtag->Text3 = new StringTag("Text3", $line3);
-		$this->namedtag->Text4 = new StringTag("Text4", $line4);
+	public function setText($line1 = null, $line2 = null, $line3 = null, $line4 = null){
+		if($line1 !== NULL) $this->namedtag->Text1 = new StringTag("Text1", $line1);
```
SOF3: PocketMine uses `null` by convention. It might be confusing here.

SOF3 (approved):

src/pocketmine/event/block/SignChangeEvent.php
robske110: it wouldn't help much. You could still be supplying 1=>"lol" 402=>"aah" 94=>"l" 93=>"la"

src/pocketmine/tile/Sign.php
robske110: Didn't mean to put this on here at all, will place this in a second PR

src/pocketmine/event/block/SignChangeEvent.php
thebigsmileXD: When assertions are turned off, it would fail @SOF3

thebigsmileXD: What you could do instead is call setLine of the tile. This would automatically provide the out of range information

src/pocketmine/tile/Sign.php
robske110: Done in #1204

src/pocketmine/event/block/SignChangeEvent.php
SOF3: If assertions are turned off, the user doesn't mind about faulty code, so no need to check. This is not a critical check that must be executed all the time.

src/pocketmine/tile/Sign.php
```diff
@@ -64,6 +64,33 @@ public function setText($line1 = "", $line2 = "", $line3 = "", $line4 = ""){
 
 		return true;
 	}
+	
+	/**
+	 * @param int    $index 0-3
+	 * @param string $line
+	 * @param bool   $update
+	 */
+	public function setLine(int $index, string $line, bool $update = true){
```
dktapps: is there any need for the `update` parameter?

robske110: well, if you want to change multiple lines you don't need to update every time. It is not necessary.

thebigsmileXD: setLine sounds like 1 line to me.

robske110: wtf?

dktapps (approved):

src/pocketmine/event/block/SignChangeEvent.php
```diff
@@ -69,14 +69,30 @@ public function getLines(){
...
+			throw new \InvalidArgumentException("Index must be in the range 0-3!");
+		}
 		return $this->lines[$index];
 	}
 
 	/**
+	 * @param string[] $lines
+	 */
+	public function setLines(array $lines){
+		if(count($lines) > 4){
```
dktapps: This should probably always be a fixed 4 lines, nothing more, nothing less. Otherwise it may cause unexpected behaviour.

SOF3: @robske110 That would depend on #1204. It is not guaranteed that SignChangeEvent is always for an empty sign; plugins may fire it for things like editing signs.

SOF3 (changes requested):

thebigsmileXD:
It turns out that this PR simply overcomplicates stuff..

robske110:
No, making it is being overcomplicated. It will actually help plugin devs. I actually have use cases already...

robske110:
@SOF3 fixed

src/pocketmine/event/block/SignChangeEvent.php
```diff
@@ -79,7 +79,7 @@ public function getLine($index){
 	 * @param string[] $lines
 	 */
 	public function setLines(array $lines){
-		if(count($lines) > 4){
+		if(count($lines) === 4){
```
dktapps: This condition is inverted. It should be `!== 4`. You're throwing an exception when the line count is correct.

matcracker: Why !== 4? If I want to set just only, for example, 2 lines; I can't.

dktapps: Because you must always have 4 lines in the array or it could lead to undefined behaviour. If you want to set 2 lines you can use setLine().

matcracker: I disagree. I think it's better to push empty values in base of array size. 
For example, if I set an array of 2 lines, the method automatically detect how many empty string need to be pushed inside the lines.

matcracker: In this way, the developer doesn't need to create an array of 4 strings.

matcracker: This is what I mean:

```
public function setLines(array $lines){
	if(($count = count($lines)) > 4){
		throw new \InvalidArgumentException("Array size cannot be greater than 4!");
	}
        for($i = 0; $i <= 4 - $count; $i++)
             array_push($lines, "");
	$this->lines = $lines;
}

```

dktapps: and that will cause potentially unexpected behaviour by overwriting the remaining lines. That is why I said use setLine() instead.

You could have also used `array_pad()` instead of that awful looking micro optimized code.

matcracker: It's obvious, in any way the setLines() will overwrite all the lines. If I need to edit just "x" lines I will use setLine(), but if need to create a sign with 2 lines, with my way the developer just need to do "setLines(array("line1", "line1"));", with the actually way he does "setLines(array("line1", "line2", "", "");"

True, array_pad() is better xD

src/pocketmine/event/block/SignChangeEvent.php
```diff
@@ -69,14 +69,30 @@ public function getLines(){
...
+			throw new \InvalidArgumentException("Index must be in the range 0-3!");
+		}
 		return $this->lines[$index];
 	}
 
 	/**
+	 * @param string[] $lines
+	 */
+	public function setLines(array $lines){
+		if(count($lines) === 4){
```
jasonwynn10: Allows arrays 6 lines and more

jasonw4331 (changes requested):

src/pocketmine/event/block/SignChangeEvent.php
SOF3: Use the array union operator:

```php
$lines += ["", "", "", ""];
```

Might this clarify the behaviour?

src/pocketmine/event/block/SignChangeEvent.php
robske110: What?

robske110:
So, there are 3 possibilities here:
setLines(array $lines), if it gets an array smaller than 4:
-it pads the rest with empty strings 
-it pads the rest with the current data
Or simply let it accept only 4 lines so we won't have to chose.
You can use the Reactions to vote.

SOF3 (approved):

dktapps (approved):

robske110:
I have now throughly tested this, I can say with a certainty that this works as intended.
There is still the controversy about setLines().
I also updated the script test plugin (To test the new features)

robske110:
Sorry to be impatient, but I'd really like to see this merged :)

PR #1204, opened 2017-07-13
setText now nullable instead of overwriting with empty lines

robske110:
## Introduction

This PR is created upon an idea by @SOF3. As it was not really fitting into the PR, because it is BC breaking here it is to be easily discussed apart from other changes.

### Relevant issues
none

## Changes
Instead of needing to reset or provide lines when using setText() now having null on any line will preserve it.

### Behavioural changes
none

## Backwards compatibility
If a plugin relied on setText() to reset with its default params it will not work anymore. This is a rare use case though. Any other use case will still work, because either empty strings are provided or actual text is being written.

## Follow-up
TODO: Check if core uses setText as explained above in Breaking changes.

## Tests
TODO: write ScriptPlugin…

### DIFF
--- a/src/pocketmine/tile/Sign.php
+++ b/src/pocketmine/tile/Sign.php
@@ -54,7 +54,18 @@ class Sign extends Spawnable{
 		unset($this->namedtag->Creator);
 	}
 
-	public function setText($line1 = null, $line2 = null, $line3 = null, $line4 = null){
+	/**
+	 * Changes contents of the specific lines to the string provided. 
+	 * Leaves contents of the specifc lines as is if null is provided.
+	 *
+	 * @param null|string $line1
+	 * @param null|string $line2
+	 * @param null|string $line3
+	 * @param null|string $line4
+	 *
+	 * @return bool
+	 */
+	public function setText($line1 = "", $line2 = "", $line3 = "", $line4 = ""){
 		if($line1 !== null){
 			$this->namedtag->Text1->setValue("Text1", $line1);
 		}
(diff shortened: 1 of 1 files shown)


Muqsit:
Formatting...

robske110:
Formatting?

TheDiamondYT1:
@robske110 Maybe he means to `if` statment and the line setting code is on the same line, instead of using brackets and multiple lines

Muqsit:
@robske110 PocketMine's doesn't have one liner if statements.
```
//Code
if(true) a = b;

//PocketMine'netstat code formatting
if(true){
    a = b;
}
```

SOF3 (approved):

TheDiamondYT1:
Is this good to be merged, since #1202 was?

TheDiamondYT1:
Wait i've just thought — why is the default parameter changed from `""` to `null`?

dktapps:
I don't believe there is any need for a method signature change: there's no typehints on this method yet, which means it is still legal to pass null.

If this change is merged, it will however mean that this method cannot be typehinted until we upgrade to PHP 7.2 (it will require a nullable typehint `?string $line1 = ""` to allow this alteration without any behavioural changes for existing code).

src/pocketmine/tile/Sign.php
```diff
@@ -55,11 +55,19 @@ public function saveNBT(){
 		unset($this->namedtag->Creator);
 	}
 
-	public function setText($line1 = "", $line2 = "", $line3 = "", $line4 = ""){
```
dktapps: Please also add PhpDoc to this function to make sure it's clear, since there are no typehints.

robske110:
oh god, how can one make two spelling mistakes in a commit message

dktapps:
Could you rebase this and resolve the conflicts?

robske110:
Rebased onto master.

robske110:
Rebased onto php/7.0.

PR #1205, opened 2017-07-13
Fix floating text disappearing bug

mal0ne-23:
## Introduction

I've noticed when using floating text with the Item entity, it seems like it despawns after awhile to the client. Since players never despawn unless they disconnect client side, this shouldn't happen anymore

### API changes

None

### Behavioural changes

Floating text boxes around title is a little darker, but positioned the same.

## Backwards compatibility
Fully backwards compatible

## Tests

Tested on my own server.

### DIFF
--- a/src/pocketmine/level/particle/FloatingTextParticle.php
+++ b/src/pocketmine/level/particle/FloatingTextParticle.php
@@ -24,9 +24,10 @@ declare(strict_types=1);
 namespace pocketmine\level\particle;
 
 use pocketmine\entity\Entity;
-use pocketmine\entity\Item as ItemEntity;
+use pocketmine\item\Item;
 use pocketmine\math\Vector3;
-use pocketmine\network\mcpe\protocol\AddEntityPacket;
+use pocketmine\utils\UUID;
+use pocketmine\network\mcpe\protocol\AddPlayerPacket;
 use pocketmine\network\mcpe\protocol\RemoveEntityPacket;
 
 class FloatingTextParticle extends Particle{
@@ -78,17 +79,14 @@ class FloatingTextParticle extends Particle{
 
 		if(!$this->invisible){
 
-			$pk = new AddEntityPacket();
+			$pk = new AddPlayerPacket();
+			$pk->uuid = UUID::fromRandom();
+			$pk->username = $this->title;
 			$pk->entityRuntimeId = $this->entityId;
-			$pk->type = ItemEntity::NETWORK_ID;
 			$pk->x = $this->x;
-			$pk->y = $this->y - 0.75;
+			$pk->y = $this->y - 0.50;
 			$pk->z = $this->z;
-			$pk->speedX = 0;
-			$pk->speedY = 0;
-			$pk->speedZ = 0;
-			$pk->yaw = 0;
-			$pk->pitch = 0;
+			$pk->item = Item::get(0);
 			$flags = (
 				(1 << Entity::DATA_FLAG_CAN_SHOW_NAMETAG) |
 				(1 << Entity::DATA_FLAG_ALWAYS_SHOW_NAMETAG) |
@@ -97,6 +95,7 @@ class FloatingTextParticle extends Particle{
 			$pk->metadata = [
 				Entity::DATA_FLAGS => [Entity::DATA_TYPE_LONG, $flags],
 				Entity::DATA_NAMETAG => [Entity::DATA_TYPE_STRING, $this->title . ($this->text !== "" ? "\n" . $this->text : "")],
+				Entity::DATA_SCALE => [Entity::DATA_TYPE_FLOAT, 0],
 			];
 
 			$p[] = $pk;


SOF3:
Have you made strict fair tests to confirm that human entities and item entities really have such distinction?

mal0ne-23:
Yes, I have been having this issue with floating text disappearing for a LONG time. It doesn't occur anymore.

dktapps:
[insert CITATION_NEEDED.png]

Is this problem reproducible without plugins? (with the exception of the minimal code needed to spawn the entity)

mal0ne-23:
Yes. I spawn the floating text, and minutes after it's just gone.

src/pocketmine/level/particle/FloatingTextParticle.php
```diff
@@ -97,6 +95,7 @@ public function encode(){
 			$pk->metadata = [
 				Entity::DATA_FLAGS => [Entity::DATA_TYPE_LONG, $flags],
 				Entity::DATA_NAMETAG => [Entity::DATA_TYPE_STRING, $this->title . ($this->text !== "" ? "\n" . $this->text : "")],
+				Entity::DATA_SCALE => [Entity::DATA_TYPE_FLOAT, 0],
```
Sandertv: A float data type should perhaps be a float, not an integer.

dktapps: It doesn't matter. Integers are accepted as floats in PHP.

Sandertv: They are? Oh my bad. I thought strict types would be strict enough to not.

matcracker:
I have the same issue, but I tried to force the loading some chunks and the issue is fixed. I didn't try your solution but I will do.

mal0ne-23:
Force loading chunks doesn't have anything to do with it. I think it may be that Item entities have a preset despawn time on the client, so they disappear for players on the server when it reaches that time. That's my guess, but it has not happened since I updated my code to use this.

keithkfng:
Confirmed... it just disappears right in front of you - needs to be merged ASAP

dktapps:
If you have to use a different entity, I'd prefer something that's not a player so that it doesn't add to the InteractPacket spam problem. However the primary goal is to find out why this issue is actually occurring and whether this is fixable without changing the implementation.

mal0ne-23:
The scale is 0, which makes it impossible to interact with it actually. I've tested that too, you can't hover over it.

Muqsit:
Using AddItemEntityPacket() would be a better alternative. No UUIDs, usernames and other useless properties.
Now, I know some might quote me on this saying that you can't make an item entity float in the air (it falls on the ground when you spawn it in air). That can be avoided using DATA_FLAG_IMMOBILE.

mal0ne-23:
Item entities also move when you place a block over them. This doesn't occur with player entities

mal0ne-23:
I DID NOT MEAN TO DO THAT

src/pocketmine/level/particle/FloatingTextParticle.php
```diff
@@ -78,25 +79,23 @@ public function encode(){
...
 			$pk->x = $this->x;
-			$pk->y = $this->y - 0.75;
+			$pk->y = $this->y - 0.50;
 			$pk->z = $this->z;
-			$pk->speedX = 0;
-			$pk->speedY = 0;
-			$pk->speedZ = 0;
-			$pk->yaw = 0;
-			$pk->pitch = 0;
+			$pk->item = Item::get(0);
```
IrishPacks: Change this to the AIR constant for OCD sake?

mal0ne-23: Does it matter?

IrishPacks: to me it does, and it's good practice.

TheDiamondYT1: I agree

AppleDevelops: @TheDiamondYT1 The day someone listens to you is the day when the world ends

mal0ne-23: k

TheDiamondYT1: @AppleDevelops lol wut

mal0ne-23:
bUmP?

samalero:
In EconomyShop|Sell use this code for create "flying" blocks
```
public function spawnTo(Player $player){
		$pk = new AddItemEntityPacket;
		$pk->entityRuntimeId = $this->eid;
		$pk->item = $this->item;
		$pk->x = $this->pos->x + 0.5;
		$pk->y = $this->pos->y;
		$pk->z = $this->pos->z + 0.5;
		$pk->speedX = $pk->speedY = $pk->speedZ = 0;

		$player->dataPacket($pk);
	}
```
but blocks also dissapear...

TheDiamondYT1:
Unknown command 'bUmP', maybe check the name case?

mal0ne-23:
The InteractPacket spam issue was fixed in 1.2, making this not a problem, with the pull correct?

dktapps:
That may be true, but this pull request targets a branch which is not 1.2.

PR #1222, opened 2017-07-20
Return the dropped item entity when using Level->dropItem to be able to further manipulate it

Sandertv:
## Introduction

This pull request is very simple and returns the ItemEntity dropped using Level->dropItem.

### API changes

Level->dropItem now returns either null, or the item entity dropped if successful.

### Behavioural changes

No considerable changes.

## Backwards compatibility
 This pull request is 100% backwards compatible. It only adds something.

## Follow-up

 -


Tarik02:
Lol https://github.com/iTXTech/Genisys/pull/1686/files :-)

Sandertv:
I've actually never seen that lol.

PR #1231, opened 2017-07-22
Update PluginDescription.php

hoyinm14mc:
## Introduction

A user on the forum reported this problem.
https://forums.pmmp.io/threads/help.3272


dktapps:
This bug was already fixed, update your version.

PR #1247, opened 2017-07-26
Added capability to specify compatible protocol versions in plugin.yml

dktapps:
## Introduction
Very often these days I see plugins which use packets to do things that the API doesn't provide the capability to do. This is problematic due to the way the protocol can, and will, change with every version of the game, because MCPE protocol changes are not covered by the PocketMine-MP API.

This pull request adds the capability for a plugin developer to specify a list of MCPE protocol versions that the plugin is compatible with. This provides a solution for the possibility of plugins crashing when things are changed/added/removed with each MCPE update without requiring an API bump.

Plugins can now declare the `mcpe-protocol` attribute in plugin.yml, containing a list of protocol versions that the plugin is compatible with.
### API changes
- added `PluginDescription->getCompatibleMcpeProtocols()`

### Behavioural changes
- Plugins which declare the `mcpe-protocol` attribute will now not be loaded if the server protocol version does not match the plugin's protocol version(s).

## Backwards compatibility
No BC breaks are introduced with this PR. Plugins which do not declare the new attribute will be loaded as normal.

## Tests
TBD

### DIFF
--- a/src/pocketmine/plugin/PluginDescription.php
+++ b/src/pocketmine/plugin/PluginDescription.php
@@ -23,12 +23,15 @@ declare(strict_types=1);
 
 namespace pocketmine\plugin;
 
+use pocketmine\network\mcpe\protocol\ProtocolInfo;
 use pocketmine\permission\Permission;
 
 class PluginDescription{
 	private $name;
 	private $main;
 	private $api;
+	/** @var int[] */
+	private $compatibleMcpeProtocols = [];
 	private $extensions = [];
 	private $depend = [];
 	private $softDepend = [];
@@ -75,6 +78,8 @@ class PluginDescription{
 		if(stripos($this->main, "pocketmine\\") === 0){
 			throw new PluginException("Invalid PluginDescription main, cannot start within the PocketMine namespace");
 		}
+
+		$this->compatibleMcpeProtocols = array_map("intval", (array) ($plugin["mcpe-protocol"] ?? []));
 
 		if(isset($plugin["commands"]) and is_array($plugin["commands"])){
 			$this->commands = $plugin["commands"];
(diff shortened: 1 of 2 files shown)


buchwasa (approved):

tarik02 (approved):

dktapps:
Probably ought to allow non-array values.

```
mcpe-protocol: 113
```
```
mcpe-protocol: [113, 120]
```

TheDiamondYT1:
@dktapps In that case you should do the same with extensions and depend

dktapps:
That's beyond the scope of this pull request.

src/pocketmine/plugin/PluginManager.php
```diff
@@ -258,6 +259,14 @@ public function loadPlugins(string $directory, array $newLoaders = null){
 								continue;
 							}
 
+							if(count($pluginMcpeProtocols = $description->getCompatibleMcpeProtocols()) > 0){
+								$serverMcpeProtocols = [ProtocolInfo::CURRENT_PROTOCOL];
+								if(count(array_intersect($pluginMcpeProtocols, $serverMcpeProtocols)) === 0){
```
SOF3: Why not in_array()? Are you expecting the server to support multiple protocol versions in the future?

src/pocketmine/plugin/PluginDescription.php
```diff
@@ -76,6 +79,8 @@ private function loadMap(array $plugin){
 			throw new PluginException("Invalid PluginDescription main, cannot start within the PocketMine namespace");
 		}
 
+		$this->compatibleMcpeProtocols = array_map(function($v){ return (int) $v; }, $plugin["mcpe-protocol"] ?? $plugin["mcpe-protocols"] ?? []);
```
SOF3: Will PhpStorm try to insert two line feeds when you auto reformat 🙈 
Can actually be replaced with `array_map("intval",`.

SOF3 (approved):

SOF3:
@TheDiamondYT1 `extensions` already allows strings, as explained in #903. It is worth considering adding the plural form though.

jasonw4331 (approved):

SOF3 (approved):

PR #1250, opened 2017-07-27
Added new events for network interface handling, close #1232

### LINKED ISSUE #1232: Disabling default raknet interface
Sometimes servers do not use default raknet interface(synapse etc.). There are a way to remove default interface but only after it started and binded.

### Possible solution
Create cancelable event before registering default raknet interface(also needed to register it after plugins started).

dktapps:
## Introduction
Added events to allow plugins to do their own network interface handling.

### Relevant issues
- close #1232

### API changes
- added `NetworkInterfaceRegisterEvent` (cancellable), `NetworkInterfaceUnregisterEvent`, and `NetworkInterfaceCrashEvent`.

### Behavioural changes
- Possible behavioural changes possible for plugins which used onLoad() and load-order Startup, however I think this is null and void anyway because of the AutoUpdater now being asynchronous. The change allows plugins which are loaded with STARTUP order to cancel registration of the RakLibInterface.

## Backwards compatibility
see above

## Follow-up
It's quite annoying the way many things within PocketMine-MP _do_ things during their constructors, such as starting a thread, registering themselves to whatever, etc. It makes it quite inconvenient to simply unregister things. In this PR, the RakLib thread will still be started when the interface is registered, only to be stopped moments later when the interface is closed due to the event being cancelled. A follow-up to this should be to find a suitable way to move this behaviour from the constructor to a dedicated method so that this does not occur.

## Tests
https://gist.github.com/dktapps/103824e007ccb114bd701de4a2ecd39c - A plugin to remove the default RakLibInterface.

### DIFF
--- a/src/pocketmine/event/server/NetworkInterfaceCrashEvent.php
+++ b/src/pocketmine/event/server/NetworkInterfaceCrashEvent.php
@@ -1,0 +1,50 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * @author PocketMine Team
+ * @link http://www.pocketmine.net/
+ *
+ *
+*/
+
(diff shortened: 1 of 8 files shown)


Tarik02:
Thanks

tarik02 (approved):

dktapps (commented):
This may cause problems with plugins which access the updater during onEnable().

Tarik02:
@dktapps which plugins will access update? So you can move update before enabling raknet interface.

dktapps:
I'm afraid I don't understand what you mean.

dktapps:
I think this is good to merge now, would anyone like to review?

tarik02 (approved):

mal0ne-23:
Plz merge

src/pocketmine/event/server/NetworkInterfaceCrashEvent.php
```diff
@@ -0,0 +1,50 @@
...
+declare(strict_types=1);
+
+namespace pocketmine\event\server;
+
+use pocketmine\network\SourceInterface;
+
+/**
+ * Called when a network interface crashes, with relevant crash information.
+ */
+class NetworkInterfaceCrashEvent extends NetworkInterfaceEvent{
```
dktapps: I really don't know what kind of idiot would do something like this, instead of actually fixing the error handling that causes the problem... 🤦‍♂️

PR #1253, opened 2017-07-28
Fix enchantment typo

tarik02:
## Introduction

Enchantments file was copied from effects file, so somebody forgot to change one `Effect` to `Enchantment`

## Backwards compatibility

This is backwards incompatible(plugins was using Enchantment::getEffectByName needs to be updated)


robske110:
good catch, looks like it has been copied from the effect class...

Tarik02:
@robske110 yes, it is

Tarik02:
I think will be better to split Enchantment item and Enchantment type like with items(ItemStack, Item)

dktapps:
This issue has already been fixed on the api3/blocks branch.

PR #1256, opened 2017-07-28
Added precision and mode arguments to Vector3::round()

SOF3:
## Introduction

Vectors may be rounded with a given precision for many reasons, such as for cleaner presentation to users, etc. Some people may also prefer different modes of rounding.

### Relevant issues

nil

### API changes

Added optional arguments `(int $precision, int $mode)` to `Vector3::round()`.

### Behavioural changes

nil

## Backwards compatibility

This is backward-incompatible for subclasses of Vector3 overriding the `round()` method. However, due to the low possibility of the existence of such classes, this should not be taken into consideration when adding such optional arguments.

## Follow-up

nil

## Tests

(I'm lazy)

### DIFF
--- a/src/pocketmine/math/Vector3.php
+++ b/src/pocketmine/math/Vector3.php
@@ -132,8 +132,10 @@ class Vector3{
 		return new Vector3((int) floor($this->x), (int) floor($this->y), (int) floor($this->z));
 	}
 
-	public function round(){
-		return new Vector3((int) round($this->x), (int) round($this->y), (int) round($this->z));
+	public function round(int $precision = 0, int $mode = PHP_ROUND_HALF_UP){
+		return $precision > 0 ?
+			new Vector3(round($this->x, $precision, $mode), round($this->y, $precision, $mode), round($this->z, $precision, $mode)) :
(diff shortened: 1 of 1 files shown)


dktapps (approved):

Muqsit:
I SWEAR I ALMOST CREATED A PR FOR THIS 2 DAYS AGO!

src/pocketmine/math/Vector3.php
```diff
@@ -132,8 +132,10 @@ public function floor(){
 		return new Vector3((int) floor($this->x), (int) floor($this->y), (int) floor($this->z));
 	}
 
-	public function round(){
-		return new Vector3((int) round($this->x), (int) round($this->y), (int) round($this->z));
+	public function round(int $precision = 0, int $mode = PHP_ROUND_HALF_UP){
+		return $precision > 0 ?
+			new Vector3(round($this->x, $precision, $mode), round($this->y, $precision, $mode), round($this->z, $precision, $mode)) :
+			new Vector3((int) round($this->x, $precision, $mode), (int) round($this->y, $precision, $mode), (int) round($this->z, $precision, $mode));
```
dktapps: I don't know if the int type casts are needed here. Vector3 is implicitly float anyway. If people really want an int they should cast it themselves or use getFloor*()

PR #1259, opened 2017-07-30
Tile NBT usage enhancements

Muqsit:
### Relevant issues

It looks like for every time PocketMine needs to change the value of a tile's NBT tag, it creates a totally new NBT tag and assigns it a value.
```php
/** @var \pocketmine\tile\Furnace $furnace */

//onUpdate()...
if($furnace->namedtag["BurnTime"] > 0){
    $furnace->namedtag["BurnTime"] = new ShortTag("BurnTime", ((int) $furnace->namedtag["BurnTime"]) - 1);
}
```

Its pointless. You can change the tag's value using `Tag::setValue($value)`.
It also seems pointless to `(mixed)` cast every nbt tag. This can be avoided by using `Tag::getValue()`.
```php
//Before:
if($furnace->namedtag["BurnTime"] > 0){
    $furnace->namedtag["BurnTime"] = new ShortTag("BurnTime", ((int) $furnace->namedtag["BurnTime"]) - 1);
}

//After:
if($furnace->namedtag->BurnTime->getValue() > 0){
    $furnace->namedtag->BurnTime->setValue($furnace->namedtag->BurnTime->getValue() - 1);
}
```

### API changes

This pull request goes hand in hand with #1116. You can still assign an index to a CompoundTag array, but it's going to be useless.
If you are modifying a tile, please use the new way which again, is what's explained in #1116.

```php
//Do NOT:
$tileNBT = new CompoundTag("", [
    "x" => new IntTag("x", 100)
]);
var_dump((int) $tileNBT["x"]);

//Do:
$tileNBT = new CompoundTag("", [
    new IntTag("x", 100)
]);
var_dump($tileNBT->x->getValue());

"------------------------------"

//Do NOT:
$tileNBT["x"] = new IntTag("x", ((int) $tileNBT["x"]) + 10);

//Do:
$tileNBT->x->setValue($tileNBT->x->getValue() + 10);
```

Spawnable::getSpawnCompound() is no longer an abstract function and returns generic tile NBT.

Added abstract function `addAdditionalSpawnData(CompoundTag $nbt)` which is where all the tile-specific nbt tags are applied.
### Behavioural changes

This can be considered a 'micro-optimization' since we aren't creating multiple new NBT tags every tick as seen in case of Furnace.

## Backwards compatibility

This PR is backwards compatible.

## Tests

Tests done:
**Chest**
Place a chest. Put some items in it. Close the chest. Restart the server. Open the chest. Works just like it should.

**Enchanting Table:**
Place an enchanting table. Open it. Close it. Restart the server. Open it. Close it. Works.

**Flower Pot:**
Place a flower pot. Add a flower (tried with cactus). Restart the server. Works.

**Item Frame:**
Place an item frame. Rotate item 45° (1 tap). Restart the server. Works.

**Sign:**
Place a sign. Fill up all four lines. Restart the server. Works.
Place a sign. Fill up the 2nd and 4th line. Restart the server. Works.

**Skull:**
Place a skull with `rotation % 90 !== 0`. Restart the server. Works.
Place a skull with `rotation % 90 === 0`. Restart the server. Works.

## API Changes made after initial changes
Spawnable::getSpawnCompound() is now a final function. It will return the common NBT tag for tiles...
```php
final public function getSpawnCompound() : CompoundTag{
    return new CompoundTag("", [
        $this->namedtag->id,
        $this->namedtag->x,
        $this->namedtag->y,
        $this->namedtag->z
    ]);
}
```

### DIFF
--- a/src/pocketmine/tile/Chest.php
+++ b/src/pocketmine/tile/Chest.php
@@ -277,28 +277,17 @@ class Chest extends Spawnable implements InventoryHolder, Container, Nameable{
 	}
 
 	public function getSpawnCompound() : CompoundTag{
+		$nbt = parent::getSpawnCompound();
+
 		if($this->isPaired()){
-			$c = new CompoundTag("", [
-				new StringTag("id", Tile::CHEST),
-				new IntTag("x", (int) $this->x),
-				new IntTag("y", (int) $this->y),
-				new IntTag("z", (int) $this->z),
-				new IntTag("pairx", (int) $this->namedtag["pairx"]),
-				new IntTag("pairz", (int) $this->namedtag["pairz"])
-			]);
-		}else{
-			$c = new CompoundTag("", [
-				new StringTag("id", Tile::CHEST),
-				new IntTag("x", (int) $this->x),
-				new IntTag("y", (int) $this->y),
-				new IntTag("z", (int) $this->z)
-			]);
+			$nbt->pairx = $this->namedtag->pairx;
+			$nbt->pairz = $this->namedtag->pairz;
 		}
 
 		if($this->hasName()){
-			$c->CustomName = $this->namedtag->CustomName;
+			$nbt->CustomName = $this->namedtag->CustomName;
 		}
 
-		return $c;
+		return $nbt;
 	}
 }
--- a/src/pocketmine/tile/Chest.php
+++ b/src/pocketmine/tile/Chest.php
@@ -276,9 +276,7 @@ class Chest extends Spawnable implements InventoryHolder, Container, Nameable{
 		return true;
 	}
 
-	public function getSpawnCompound() : CompoundTag{
-		$nbt = parent::getSpawnCompound();
-
+	public function addAdditionalSpawnData(CompoundTag $nbt){
 		if($this->isPaired()){
 			$nbt->pairx = $this->namedtag->pairx;
 			$nbt->pairz = $this->namedtag->pairz;
@@ -287,7 +285,5 @@ class Chest extends Spawnable implements InventoryHolder, Container, Nameable{
 		if($this->hasName()){
 			$nbt->CustomName = $this->namedtag->CustomName;
 		}
-
-		return $nbt;
 	}
 }
--- a/src/pocketmine/tile/Chest.php
+++ b/src/pocketmine/tile/Chest.php
@@ -95,7 +95,7 @@ class Chest extends Spawnable implements InventoryHolder, Container, Nameable{
 	 */
 	protected function getSlotIndex(int $index){
 		foreach($this->namedtag->Items as $i => $slot){
-			if((int) $slot->Slot->getValue() === $index){
+			if($slot->Slot->getValue() === $index){
 				return (int) $i;
 			}
 		}
--- a/src/pocketmine/tile/Chest.php
+++ b/src/pocketmine/tile/Chest.php
@@ -74,7 +74,7 @@ class Chest extends Spawnable implements InventoryHolder, Container, Nameable{
 	}
 
 	public function saveNBT(){
-		$this->namedtag->Items = new ListTag("Items", []);
+		$this->namedtag->Items->setValue([]);
 		$this->namedtag->Items->setTagType(NBT::TAG_Compound);
 		for($index = 0; $index < $this->getSize(); ++$index){
 			$this->setItem($index, $this->inventory->getItem($index));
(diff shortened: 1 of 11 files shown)


src/pocketmine/tile/FlowerPot.php
```diff
@@ -66,12 +65,12 @@ public function canAddItem(Item $item) : bool{
 	}
 
 	public function getItem() : Item{
-		return Item::get((int) ($this->namedtag["item"] ?? 0), (int) ($this->namedtag["mData"] ?? 0), 1);
+		return Item::get($this->namedtag->item->getValue(), $this->namedtag->mData->getValue(), 1);
```
Muqsit: In case you were thinking why I did not check whether `$this->namedtag->item` or `$this->namedtag->mData` is set before using `getValue()`, see https://github.com/Muqsit/PocketMine-MP/blob/b7100070e2b2071a1b38579d337362e579dfbef9/src/pocketmine/tile/FlowerPot.php#L35-#L40

src/pocketmine/tile/Chest.php
```diff
@@ -74,7 +74,7 @@ public function close(){
 	}
 
 	public function saveNBT(){
-		$this->namedtag->Items = new ListTag("Items", []);
+		$this->namedtag->Items->setValue([]);
 		$this->namedtag->Items->setTagType(NBT::TAG_Compound);
```
Muqsit: I'm guessing this line is unnecessary because https://github.com/Muqsit/PocketMine-MP/blob/b7100070e2b2071a1b38579d337362e579dfbef9/src/pocketmine/tile/Chest.php#L51

dktapps:
Looks good at first glance, I'll review this properly later when I have the brainpower to do so.

src/pocketmine/tile/Furnace.php
```diff
@@ -170,7 +169,7 @@ public function setItem(int $index, Item $item){
 			}
 			$this->namedtag->Items[$i] = $d;
 		}else{
-			$this->namedtag->Items[$i] = $d;
+			$this->namedtag->Items[$i]= $d;
```
dktapps: formatting

solo5star:
Wow, It looks really cool. I hope this PR merge.

src/pocketmine/tile/FlowerPot.php
jasonwynn10: It's possible for a plugin to unset them though. Unlikely: yes, but still possible

dktapps:
Since there is a lot of undefined behaviour around the NBT implementation (type checking and such) it is sometimes preferable to create a new NBT tag to ensure that the tag is of the expected type (plugins can cause problems, as seen in earlier comments).

Your alterations to the spawn compound handling are reasonable. A possible improvement might be to make `getSpawnCompound()` final and have a separate abstract method something like `addAdditionalSpawnData()` which would be called by `getSpawnCompound()`. That would prevent accidents with forgetting the parent call and would also ensure that future additions must follow the new rules - this PR, for example, does not clean up the Bed tile implementation I added since this PR was opened. If I hadn't remembered beds myself it would have gone unnoticed.

Muqsit:
Nice, this also reduces some boilerplate code.

dktapps:
Looks good to me, have you tested this fully?

Muqsit:
@dktapps Yes

src/pocketmine/tile/FlowerPot.php
Muqsit: @jasonwynn10 It's technically possible to unset or redefine almost any variable using reflection classes, I think the plugin should be blamed for their incorrect usages.

src/pocketmine/tile/Sign.php
```diff
@@ -89,29 +86,24 @@ public function getLine(int $index) : string{
 		if($index < 0 or $index > 3){
 			throw new \InvalidArgumentException("Index must be in the range 0-3!");
 		}
-		return (string) $this->namedtag["Text" . ($index + 1)];
+		return $this->namedtag{"Text" . ($index + 1)}->getValue();
```
dktapps: Missed something?

Muqsit: The casting was unneeded

dktapps: I'm not talking about the cast. You're missing a `->`.

 The cast wouldn't be needed regardless (it's there to shut PhpStorm up because it can't determine the type of the value gotten from the tag).

Muqsit:
:D

PR #1260, opened 2017-07-30
Add Vector2::setComponents, Vector2::getComponents, Vector3::getComponents

Tarik02:
## Introduction

Vector3 contains method setComponents, so lets add the same to Vector2.
Also like they have setComponents, so lets add getComponents

### API changes

- Added Vector2::setComponents
- Added Vector2::getComponents
- Added Vector3::getComponents

## Backwards compatibility

Backwards compatible

### DIFF
--- a/src/pocketmine/math/Vector2.php
+++ b/src/pocketmine/math/Vector2.php
@@ -126,6 +126,15 @@ class Vector2{
 	}
 
 	/**
+	 * Returns array of vector's components(x, y)
+	 *
+	 * @return array
+	 */
+	public function getComponents() : array {
+		return [$this->x, $this->y];
+	}
+
+	/**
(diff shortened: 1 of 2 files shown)


dktapps:
what is the use case for getComponents()?

Tarik02:
@dktapps
```php
list($x, $y, $z) = $sign->getComponents();
```
is really looks better than
```php
$x = $sign->getX();
$y = $sign->getY();
$z = $sign->getZ();
```

Tarik02:
@SOF3 so what do you want to change in this PR?

SOF3:
@Tarik02 I don't think addition of `getComponents()` is reasonable.

jasonwynn10:
whatabout setComponents() ?

PR #1276, opened 2017-08-06
Try fix permission command

Andrew1481432:
## Introduction
 Checking the rights when entering commands

### Behavioural changes
If the player has no right all needed command are hidden


dktapps:
This does nothing other than what `Command->generateCustomCommandData()` already did before issues started cropping up with it.

PR #1277, opened 2017-08-06
Remove InventoryType

dktapps:
## Introduction
This class is an unnecessary extra restriction on plugins wishing to implement new things. Due to InventoryType being sealed in (private constructor) it is difficult to create new things that rely on inventory.

It also doesn't make sense to store default properties like this when they can be assigned in the actual classes themselves. This PR makes it cleaner to implement new inventory types without needing hacks.

### API changes
- Removed `InventoryType`
- Added `BaseInventory->getDefaultSize()`, `BaseInventory->getNetworkType()`, `BaseInventory->getName()`
- Remove unused `CraftingInventory` and `SlotType`
- The signature of `BaseInventory->__construct()` has been altered to remove InventoryType. The signature is now `InventryHolder, Item[], ?int, ?string`.
- Removed useless `CraftingInventory` class and `SlotType` interface.

## Backwards compatibility
TBD

## Tests
TBD


dktapps:
This will instead be merged into the mcpe-1.2 branch since that branch is getting an inventory rewrite anyway.

TheDiamondYT1:
@dktapps Make that branch public 😃

PR #1296, opened 2017-08-17
Fix some type errors

Encritary:


dktapps:
don't pass strings to vectors and you won't have problems...

PR #1297, opened 2017-08-17
Fix some type errors on custom ExplodePacket use

Encritary:
## Introduction

Original error was:
`[12:25:56] [Server thread/CRITICAL]: TypeError: "Argument 1 passed to pocketmine\network\mcpe\protocol\DataPacket::putSignedBlockPosition() must be of the type integer, float given, called in /home/mrgenga/mcpm/src/pocketmine/network/mcpe/protocol/ExplodePacket.php on line 65" (EXCEPTION) in "src/pocketmine/network/mcpe/protocol/DataPacket" at line 350`
This PR fixed it
## Tests

I relaunched same code and it worked without any errors.

### DIFF
--- a/src/pocketmine/network/mcpe/protocol/ExplodePacket.php
+++ b/src/pocketmine/network/mcpe/protocol/ExplodePacket.php
@@ -62,7 +62,7 @@ class ExplodePacket extends DataPacket{
 		$this->putUnsignedVarInt(count($this->records));
 		if(count($this->records) > 0){
 			foreach($this->records as $record){
-				$this->putSignedBlockPosition($record->x, $record->y, $record->z);
+				$this->putSignedBlockPosition((int) $record->x, (int) $record->y, (int) $record->z);
 			}
 		}
 	}


Muqsit:
don't pass floats to vectors and you won't have problems...

dktapps:
I would say this is the plugin's responsibility to manage since you should know that block positions are floats, not ints. You should resolve the bug on your own code instead of spamming PRs at us.

However, since there is no strict definition of whether vector3s should be floats or ints, this is (possibly) a reasonable change.

Muqsit:
@dktapps wouldn't it be better to make this change to putSignedBlockPosition() function itself? I can't find another instance where int casting would be required but just saying.

dktapps:
No, it would not. That would simply conceal issues caused by plugins, and that method has int typehints anyway. It's only a problem specifically here because Vector3s can be ints or floats. Everywhere else the types are documented.

dktapps:
To clarify, this is the only place in the protocol where PM currently uses Vector3s for block positions. Everywhere else it's documented ints.

Issue #1211, opened 2017-07-15
Stair kick you..


QuanFromChina:
![image_uploaded_from_ios_720](https://user-images.githubusercontent.com/29927587/28241318-61b80786-6960-11e7-9828-1390570627c1.jpg)
![image_uploaded_from_ios-1](https://user-images.githubusercontent.com/29927587/28241321-6b7933bc-6960-11e7-9875-fdd0b34f00b5.jpg)

PocketMine-MP: Build #300
OS: Window
PHP: 7

TheDiamondYT1:
Related to #169

TheDiamondYT1:
Why the downvote?

dktapps:
It's because stairs are considered to have the same bounding box area as slabs do (due to lack of implementation of complex AABBs).

QuanFromChina:
@dktapps hi can we hack stair BB to full block ?

Tarik02:
@QuanFromChina no because then you will get "moven wrongly" while moving on the stairs

QuanFromChina:
@Tarik02 But this is criticial issue it make gameplay impossible !

masterofmcpe:
same issue, this why i use genyisys pro... pmmp never fix bugs until api3/blocks merged

Tarik02:
@QuanFromChina you won't able to move on stairs if you will do it.

TheDiamondYT1:
@QuanFromChina We do not *'hack'* things unless it is absolutely essentials that we do. If you want to use hacky, buggy, and possibly slow server softwares then be my guest — but if you want to use a server software that is frequently updated, feature rich, and fast, then stay here and stop complaining.

PR #1298, opened 2017-08-17
Implement Stained glass and stained glass pane

daniktheboss:
## Introduction
Adds stained glass and stained glass pane

## Tests
Tested, has all stained glass

### DIFF
--- a/src/pocketmine/block/StainedGlass.php
+++ b/src/pocketmine/block/StainedGlass.php
@@ -26,7 +26,7 @@ namespace pocketmine\block;
 use pocketmine\block\utils\ColorBlockMetaHelper;
 use pocketmine\item\Item;
 
-class StainedGlass extends StainedGlass{
+class StainedGlass extends Glass{
 
 	protected $id = self::STAINED_GLASS;
 
--- a/src/pocketmine/block/StainedGlass.php
+++ b/src/pocketmine/block/StainedGlass.php
@@ -1,0 +1,49 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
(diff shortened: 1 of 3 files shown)


xZeroMCPE:
tf

daniktheboss:
@xZeroMCPE ??

TheDiamondYT1:
Why not just construct a GlassBlock/Pane and set the name? Would eliminate the need for a custom class

TheDiamondYT1:
Also this is 1.2 change, so commit to the 1.2 branch

mal0ne-23:
@TheDiamondYT1 Doesn't stained glass have a completely different ID than normal glass though? Correct me if I'm wrong

TheDiamondYT1:
@mal0ne-23 https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/block/Block.php#L418

The first parameter is the id

mal0ne-23:
@xZeroMCPE @AppleDevelops Also, I understand you guys don't like Danik, but keep that out of here. It wasn't copy pasted/stolen from anywhere and doesn't have any errors.

xZeroMCPE:
@daniktheboss Who says' I don't like him? You should definitely verify where you get your information; before stating it. 

I'd also recommended you to keep all this out of this issue.

TheDiamondYT1:
@xZeroMCPE You shouldnt have responsed in the first place

dktapps:
@TheDiamondYT1 no. If you look at the majority of them they have a constructor override which only accepts meta. Regardless, that wouldn't make sense in this case because stained glass is a coloured block and each of its _variants_ have different meta.

This pull request looks reasonable, however it would be preferable if `StainedGlass` extended `Glass` and `StainedGlassPane` extended `GlassPane`, since they share all the same properties except ID and names.

dktapps:
This can be accepted on master, stained glass already exists in the game since 1.0, it just wasn't properly implemented until 1.2.

src/pocketmine/block/StainedGlass.php
```diff
@@ -26,7 +26,7 @@
 use pocketmine\block\utils\ColorBlockMetaHelper;
 use pocketmine\item\Item;
 
-class StainedGlass extends Transparent{
+class StainedGlass extends StainedGlass{
```
TheDiamondYT1: Ummmmm

TheDiamondYT1:
Oh whoops

QuanFromChina:
@TheDiamondYT1 Stop spam your stupid " whoops comment " ! 130 kid watch this project ! Shut up !

TheDiamondYT1:
@QuanFromChina Says you https://github.com/pmmp/PocketMine-MP/issues/1211

PR #1302, opened 2017-08-19
Move block registration to its own class

dktapps:


dktapps:
I'm not sure if the fast-access property arrays should be in the block factory... not sure where else to put them though.

PR #1303, opened 2017-08-19
Added basic support for blocks with multiple AABBs, fixed stairs

### LINKED ISSUE #1211: Stair kick you..

dktapps:
## Introduction
Blocks such as stairs do not have a simple cuboid collision box. Because PocketMine-MP has (until now) considered stairs to have the same collision box as slabs, this has caused many issues with movement and anti-flight anti-cheat.

### Relevant issues
fixes #1211 and a ton of other "moved wrongly!" complaints

## Changes
- separation of concepts of "bounding box" and "collision boxes" (bounding box encapsulates the block, collision boxes are the things that entities collide with)

### API changes
- added API method `Block->getCollisionBoxes()`

### Behavioural changes
- fixed kicked for flying when walking on stairs
- fixed arrows being able to be shot through the top part of blocks

## Follow-up
This DOES NOT account for the following (literal) corner cases:
image
image
This should be done in the future.

## Tests
- tested extensively using zero-gravity arrows and as a player subjected to anti-cheat.

### DIFF
--- a/src/pocketmine/block/Block.php
+++ b/src/pocketmine/block/Block.php
@@ -70,8 +70,8 @@ class Block extends Position implements BlockIds, Metadatable{
 	public $boundingBox = null;
 
 
-	/** @var AxisAlignedBB[] */
-	protected $collisionBoxes = [];
+	/** @var AxisAlignedBB[]|null */
+	protected $collisionBoxes = null;
 
 	/**
 	 * @param int         $id     The block type's ID, 0-255
@@ -515,7 +515,7 @@ class Block extends Position implements BlockIds, Metadatable{
 	 * @return AxisAlignedBB[]
 	 */
 	public function getCollisionBoxes() : array{
-		if(empty($this->collisionBoxes) === 0){
+		if($this->collisionBoxes === null){
 			$this->collisionBoxes = $this->recalculateCollisionBoxes();
 		}
 
--- a/src/pocketmine/block/Block.php
+++ b/src/pocketmine/block/Block.php
@@ -71,7 +71,7 @@ class Block extends Position implements BlockIds, Metadatable{
 
 
 	/** @var AxisAlignedBB[] */
-	protected $boundingBoxes = [];
+	protected $collisionBoxes = [];
 
 	/**
 	 * @param int         $id     The block type's ID, 0-255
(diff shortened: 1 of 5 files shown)


thediamondyt-2:
when will this be merged?

matcracker:
Does this issue fix carpets too?

dktapps:
I doubt it, since I was unaware that there were any problems with carpet, and carpets don't have a complex AABB.

matcracker:
Because sometimes, when I walk from a normal block to a carpet, it stops me and I need to start walking again.

matcracker:
I can confirm for carpets and slabs, tested with last commit (https://github.com/pmmp/PocketMine-MP/commit/55e32424a09fa699379c53bc0c573ec93c5b8e2f) of master branch and without plugins.

And another question, I don't know if it's related. When I'm running and jumping at the same time (in any blocks), I'm "teleported" in the ground (as if the jump is cancelled). Is it normal?

dktapps:
> carpets don't have a complex AABB.

therefore carpets are irrelevant to this PR.

SOF3 (approved):
An interface for bounding boxes `IBoundingBox` (with another implementation `BatchBoundingBox`) could be better and preserve more compatibility, but considering that most methods in AxisAlignedBB are limited to simple bounding boxes anyway, such change is not going to preserve much compatibility anyway (in terms of strict object typing), so I would say that this PR is good enough.

dktapps:
There is an issue here with how "bounding box" is defined. I believe the bounding box should actually be a single simple cuboid which encapsulates the entire visible block, as seen in-game with the wireframe outline. What stairs have is several _collision_ boxes.

dktapps:
This also applies to other blocks like fences - their bounding box is a simple cuboid which encapsulates the block itself, whereas the _collision_ box is the part that goes up to +1.5.

src/pocketmine/block/Block.php
```diff
@@ -808,85 +842,84 @@ protected function recalculateBoundingBox(){
...
-			$f = 1;
-		}elseif($vector === $v5){
-			$f = 2;
-		}elseif($vector === $v6){
-			$f = 3;
+			return MovingObjectPosition::fromBlock($this->x, $this->y, $this->z, $f, $vector->add($this->x, $this->y, $this->z));
 		}
 
-		return MovingObjectPosition::fromBlock($this->x, $this->y, $this->z, $f, $vector->add($this->x, $this->y, $this->z));
+		return null;
```
dktapps: The changed logic in this function is not correct - it ought to return the ray-traced point closest to the starting position.

PR #1306, opened 2017-08-21
Fix server MOTD showing on the wrong server in server list

### LINKED ISSUE #1208: MOTD still doing weird whenever it's custom
### Issue description

When I try to set a custom MOTD with
```
$this->getServer()->getNetwork()->setName("Server MOTD");
```
The Server would appear "offline" and shows the MOTD somewhere else

1. Change the MOTD with a plugin

### OS and versions

* PocketMine-MP: 58214bb
* PHP: 7.0.20
* Server OS: Linux Debian
* Game version: PE/Win10 (delete as appropriate) Win10

### Plugins

* ...

mal0ne-23:
## Introduction

The server MOTD will usually jump to another server in the list, and then update to the correct one after a bit.

### Relevant issues
Fixes #1208 

### API changes

None
### Behavioural changes

nope

## Tests

Tested and works.

### DIFF
--- a/src/pocketmine/network/mcpe/RakLibInterface.php
+++ b/src/pocketmine/network/mcpe/RakLibInterface.php
@@ -179,7 +179,7 @@ class RakLibInterface implements ServerInstance, AdvancedSourceInterface{
 				ProtocolInfo::MINECRAFT_VERSION_NETWORK,
 				$info->getPlayerCount(),
 				$info->getMaxPlayerCount(),
-				"0",
+				(string) mt_rand(0, PHP_INT_MAX),
 				$this->server->getName(),
 				Server::getGamemodeName($this->server->getGamemode())
 			]) . ";"


jasonwynn10:
Why does this solve the issue?

mal0ne-23:
Because the previous "0" was supposed to be the server's ID. There are no methods currently in Raklib to get the server's ID though... Keeping it always 0 would send 0 to the list more than once (assuming most servers use pmmp), and then there would be duplicates and the client does dumb shit. Now it's always random, so each server has it's own ID on the list (unless there's a rare occasion of two of the IDs being duplicated)

robske110:
@mal0ne-23 
>There are no methods currently in Raklib to get the server's ID though...
???
Are you talking about the server id? That can be get really easily in RakLibInterface.
Just replace the `"0"` with ?`(string) $this->server->getServerUniqueId()`.
That would nearly eradicate the possibility of a random collision.

jasonwynn10:
@robske110 If it were supposed to be the Server Unique Id, then I think dk would have used that to begin with. I think that the `"0"` is actually the index of the server in the list of servers on the client

xxFlare:
@robske110  The 1 in 9223372036854775807 chance of a random collision? 😛

Anyways, this PR isn't the correct solution:
image1

mal0ne-23:
@xxFlare This has been tested and works on 1.2, idk about 1.1...

http://imgur.com/a/N7jhu

robske110:
@xxFlare that is probably because the server is using this function perodically to update player cnt/other stuff and then it constantly sends new ids. That's why the server unique id should be used here probably

jojoe77777:
@xxFlare mt_rand max value is 2147483647 (mt_getrandmax())

On-topic: as robske110 said, `$this->server->getServerUniqueId()`

dktapps:
Sadly none of you are correct... this is indeed the simplest solution short of changing RakLib. This field is supposed to be the RakNet server ID, not a PM UUID.

dktapps:
The correct server ID to use is [this one](https://github.com/pmmp/RakLib/blob/master/server/SessionManager.php#L84), however it's a little inaccessible. Since I was unable to reproduce this issue, I suggested this quick hack and it worked.

PR #1308, opened 2017-08-22
Ability to get a players quit reason

xxFlare:
### Introduction
There was no way to do this before. I don't believe plugins should be able to modify the reason though. That can be changed though if requested.

This allows the possibility for plugins to track how players leave the server.

### DIFF
--- a/src/pocketmine/event/player/PlayerQuitEvent.php
+++ b/src/pocketmine/event/player/PlayerQuitEvent.php
@@ -40,7 +40,7 @@ class PlayerQuitEvent extends PlayerEvent{
 	/**
 	 * @param Player                      $player
 	 * @param TranslationContainer|string $quitMessage
-     * @param string                      $quitReason
+	 * @param string                      $quitReason
 	 */
 	public function __construct(Player $player, $quitMessage, $quitReason){
 		$this->player = $player;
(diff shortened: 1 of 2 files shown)


src/pocketmine/event/player/PlayerQuitEvent.php
```diff
@@ -34,14 +34,18 @@ class PlayerQuitEvent extends PlayerEvent{
...
+	/** @var string */
+	protected $quitReason;
 
 	/**
 	 * @param Player                      $player
 	 * @param TranslationContainer|string $quitMessage
+	 * @param string                      $quitReason
 	 */
-	public function __construct(Player $player, $quitMessage){
+	public function __construct(Player $player, $quitMessage, $quitReason){
```
dktapps: don't forget typehint here...

src/pocketmine/event/player/PlayerQuitEvent.php
```diff
@@ -57,4 +61,11 @@ public function setQuitMessage($quitMessage){
 	public function getQuitMessage(){
 		return $this->quitMessage;
 	}
+
+	/**
+	 * @return string
+	 */
+	public function getQuitReason(){
```
dktapps: ... and here.

src/pocketmine/event/player/PlayerQuitEvent.php
xxFlare: done.

PR #1310, opened 2017-08-23
Remove Entity->length

Muqsit:
## Introduction

This commit removes an unused variable (Entity->length). I believe this was supposed to be an alias to Entity->width, would still be a bad idea.

### API changes

None
### Behavioural changes

None
## Backwards compatibility

This commit is backwards compatible.


dktapps:
An entity's AABB is always square when viewed from above, so this field is indeed redundant.

PR #1312, opened 2017-08-24
ServerScheduler::scheduleAsyncTask() returns the worker used

SOF3:
## Introduction
Plugins might want to keep on using the same AsyncWorker after the first AsyncTask.

A downside is that this may encourage abuse by throttling a certain AsyncWorker.

### API changes

`ServerScheduler::scheduleAsyncTask()` and `AsyncPool::submitTask` now return an `int`, which is the worker ID assigned for the task, unless it is rejected, which returns `-1`.

### Behavioural changes

nil

## Backwards compatibility

None, unless some classes extend AsyncWorker.

## Follow-up

nil

## Tests

None yet. I admit that I am lazy.

### DIFF
--- a/src/pocketmine/scheduler/AsyncPool.php
+++ b/src/pocketmine/scheduler/AsyncPool.php
@@ -47,7 +47,7 @@ class AsyncPool{
 		$this->server = $server;
 		$this->size = $size;
 
-		$memoryLimit =  (int) max(-1, (int) $this->server->getProperty("memory.async-worker-hard-limit", 1024));
+		$memoryLimit = (int) max(-1, (int) $this->server->getProperty("memory.async-worker-hard-limit", 1024));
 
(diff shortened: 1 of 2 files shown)


dktapps:
You mention changes to AsyncWorker API, but I only see changes to ServerScheduler and AsyncPool.

robske110:
What does keeping to use a specific AsyncWorker for a plugin improve?

SOF3:
@dktapps Thanks, that's a typo.
@robske110 That can help load management (not sure how exactly, but might be helpful), and more importantly, execute two consecutive AsyncTasks in the same worker to do things that require using the same non-thread-safe resource.

dktapps (approved):
This change is minor and does not offer any notable BC breaks. The benefits as I see them are minimal, but I can't see any reason to say no.

PR #1320, opened 2017-08-25
Stop clearing chunk cache after 768 batched packets are cached.

xxFlare:
### Introduction
This should not need to be cleared after 768 batched packets are cached since cached chunks are automatically removed from cache when the actual chunk unloads.

At worst, memory manager will clear the cache by force if the user correctly set his/hers memory limits, though chunk cache itself does not use much memory.

Changes:
- Stop clearing cache after the array meets a size threshold of 768 packets.
- Renamed the $full parameter to $force because it sounds more accurate.


dktapps:
This is probably acceptable, since, as you say, cached chunks are disposed of on unload.

PR #1322, opened 2017-08-26
Added explicit AsyncTask->storeLocal(), removed AsyncTask->__construct() object storage

dktapps:
## Introduction
Far too often I see people using IDEs which generate the constructors for them and then accidentally unintentionally store things in the object store. This parent constructor behaviour is unexpected. 

If a developer wants to store something, they should now do so explicitly by calling storeLocal(). The intention of this is more clear and there is no doubt about what such a call will do.

This pull request also clarifies the documentation on using the scheduler object storage.

### API changes
- AsyncTask no longer has its own constructor.
- Tasks wishing to store objects for later use must use `storeLocal()` instead.

## Backwards compatibility
This is a backwards-compatibility break for any plugins which used the constructor to store objects.

## Tests
TesterPlugin submodule has been updated to reflect the changes in this pull request.

### DIFF
--- a/src/pocketmine/scheduler/AsyncTask.php
+++ b/src/pocketmine/scheduler/AsyncTask.php
@@ -33,7 +33,15 @@ use pocketmine\Server;
  * with no AsyncTask running. Therefore, an AsyncTask SHOULD NOT execute for more than a few seconds. For tasks that
  * run for a long time or infinitely, start another {@link \pocketmine\Thread} instead.
  *
- * WARNING: Do not call PocketMine-MP API methods, or save objects (and arrays containing objects) from/on other Threads!!
+ * WARNING: Any non-Threaded objects WILL BE SERIALIZED when assigned to members of AsyncTasks or other Threaded object.
+ * If later accessed from said Threaded object, you will be operating on a COPY OF THE OBJECT, NOT THE ORIGINAL OBJECT.
+ * If you want to store non-serializable objects to access when the task completes, store them using
+ * {@link AsyncTask#storeLocal}.
+ *
+ * WARNING: As of pthreads v3.1.6, arrays are converted to Volatile objects when assigned as members of Threaded objects.
+ * Keep this in mind when using arrays stored as members of your AsyncTask.
+ *
+ * WARNING: Do not call PocketMine-MP API methods from other Threads!!
  */
 abstract class AsyncTask extends Collectable{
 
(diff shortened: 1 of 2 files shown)


src/pocketmine/scheduler/AsyncTask.php
```diff
@@ -33,7 +33,15 @@
...
- * WARNING: Do not call PocketMine-MP API methods, or save objects (and arrays containing objects) from/on other Threads!!
+ * WARNING: Any non-Threaded objects WILL BE SERIALIZED when assigned to members of AsyncTasks or other Threaded object.
+ * If later accessed from said Threaded object, you will be operating on a COPY OF THE OBJECT, NOT THE ORIGINAL OBJECT.
+ * If you want to store non-serializable objects to access when the task completes, store them using
+ * {@link AsyncTask#storeLocal}.
+ *
+ * WARNING: As of pthreads v3.1.6, arrays are converted to Volatile objects when assigned as members of Threaded objects.
+ * Keep this in mind when using arrays stored as members of your AsyncTask.
+ *
+ * WARNING: Do not call PocketMine-MP API methods from other Threads!!
```
Muqsit: Why? Does "other Threads" include the main thread? Because I call `NBT::readCompresed()` and `Item::nbtDeserialize()` in one of my plugins.

dktapps: A lot of things are not initialized on workers that are initialized on the main thread, which may cause unexpected behaviour. For example, because the ItemFactory is not initialized, calling nbtDeserialize() on a worker will give you a generic `Item` object no matter what item was stored.

Muqsit: Oh noes... I messed up in many places.

src/pocketmine/scheduler/AsyncTask.php
```diff
@@ -216,14 +201,44 @@ public function onProgressUpdate(Server $server, $progress){
 	}
 
 	/**
-	 * Call this method from {@link AsyncTask#onCompletion} to fetch the data stored in the constructor, if any, and
-	 * clears it from the storage.
+	 * Saves mixed data in ServerScheduler's object storage on the main thread. You may use this to retain references to
+	 * objects or arrays which you need to access in {@link AsyncTask#onCompletion} which cannot be stored as a
```
SOF3: onCompletion() or onProgressUpdate().

dktapps: I think it's best not to introduce more confusion over fetch vs peek. This documents behaviour and the most common use, it's up to developers to decide how to use it.

SOF3 (approved):

PR #1334, opened 2017-08-29
Disallow large-chests from opening when a non-transparent block is above

### LINKED ISSUE #1165: large-chest can be open even if one of upper block is not transparent
### Issue description

In general, chest or large-chest can not be opened if upper block is not transparent.
But, it seems that large-chest can be open even if one of upper block is not transparent.

**Expected Result:**
large-chest must not be opened if one of upper block is not transparent.

1. Place large-chest on the ground (place chest and place another chest by side)
2. Place a not transparent block on the one of chest pair.

unnamed

### OS and versions

* PocketMine-MP: master branch b4a149cce81bfce22bb57e07554946620ecb2f74
* PHP: 7.0.3
* Server OS: Ubuntu
* Game version: PE (iPhone SE)

### Plugins
yes, I can reproduce this issue without any plugins.

* no server log related to this issue

Muqsit:
## Introduction

In PocketMine, you can open double chests having a non-transparent block above them. This goes against the vanilla MCPE behaviour.

### Relevant issues

* Fixes #1165

### Behavioural changes

Large chests having a non-transparent block above either of the chests cannot be opened.

## Tests

**Can be opened:**
- Place slabs/glowstone above one side of a double chest.
-  Place slabs/glowstone above both sides of the double chest.

**Cannot be opened:**
- Place stone above one side of a double chest.
- Place stone above both sides of the double chest.

### DIFF
--- a/src/pocketmine/block/Chest.php
+++ b/src/pocketmine/block/Chest.php
@@ -136,18 +136,11 @@ class Chest extends Transparent{
 
 	public function onActivate(Item $item, Player $player = null) : bool{
 		if($player instanceof Player){
-			$top = $this->getSide(Vector3::SIDE_UP);
-			if(!$top->isTransparent()){
-				return true;
-			}
 
 			$t = $this->getLevel()->getTile($this);
 			$chest = null;
 			if($t instanceof TileChest){
 				$chest = $t;
-				if($chest->getPair() instanceof TileChest and !$chest->getPair()->getBlock()->getSide(Vector3::SIDE_UP)->isTransparent()){
-					return true;
-				}
 			}else{
 				$nbt = new CompoundTag("", [
 					new ListTag("Items", []),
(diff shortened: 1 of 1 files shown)


src/pocketmine/block/Chest.php
```diff
@@ -137,14 +137,17 @@ public function onBreak(Item $item, Player $player = null) : bool{
...
 				return true;
 			}
 
 			$t = $this->getLevel()->getTile($this);
 			$chest = null;
 			if($t instanceof TileChest){
 				$chest = $t;
+				if($chest->getPair() instanceof TileChest and !$chest->getPair()->getBlock()->getSide(Vector3::SIDE_UP)->isTransparent()){
+					return true;
+				}
```
dktapps: I think this would be better positioned with the `lock`ing code further down. A newly-created chest tile may pair with its neighbours.

Muqsit: Ooooooooh

dktapps:
In future please create a separate branch on your fork for pull requests instead of committing directly to master. You'll have less headaches if you decide to create multiple pull requests, and it'll make it easier to update your fork.

Muqsit:
@dktapps Oh alright.
I was testing out how the game handles "Locked" chests. Since MCPE vanilla currently doesn't have the feature, I tested it on PC. On PC, the client is sent a debug message when they attempt to open a locked chest using an invalid key.
"{CHESTNAME} is locked!"
xsxs

dktapps:
Merged in https://github.com/pmmp/PocketMine-MP/commit/138d85307beb4d74cb131406524bcd957faf2d1f, didn't merge directly due to whitespace fixes locally. Thanks.

PR #1344, opened 2017-09-02
Fix error on Chunk::getTile()

nnnlog:
## Introduction

If use method Chunk::getTile(), happen error.
So, I fix this error.

Error : 
```
[14:36:53] [Server thread/CRITICAL]: /set에서 명령어 '/set 0'을 실행하는 중 예외가 발생했습니다: Argument 2 passed to pocketmine\level\format\Chunk::getTile() must be of the type integer, string given, called in phar:///home/swiftnode/leo/PocketMine-MP.phar/src/pocketmine/level/Level.php on line 1920
[14:36:53] [Server thread/CRITICAL]: TypeError: "Argument 2 passed to pocketmine\level\format\Chunk::getTile() must be of the type integer, string given, called in phar:///home/swiftnode/leo/PocketMine-MP.phar/src/pocketmine/level/Level.php on line 1920" (EXCEPTION) in "src/pocketmine/level/format/Chunk" at line 686
```

### Behavioural changes

I change Chunk::getTile().

## Backwards compatibility
compatibility

## Tests
It works good. not occur error

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -1915,10 +1915,10 @@ class Level implements ChunkManager, Metadatable{
 	 * @return Tile|null
 	 */
 	public function getTile(Vector3 $pos){
-		$chunk = $this->getChunk($pos->x >> 4, $pos->z >> 4, false);
+		$chunk = $this->getChunk((int) $pos->x >> 4,(int) $pos->z >> 4, false);
 
 		if($chunk !== null){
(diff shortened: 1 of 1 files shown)


dktapps:
if I only knew how many PRs like this I'd seen... do you want to add typecasts on every single other method that uses Vector3s like this when plugins feed them the wrong types? Fix your code, don't hack ours.

PR #1353, opened 2017-09-03
players spawn in the middle of the block

armagadon159753:
this pull request has been tested even after executing the following command: /setworldspawn

### DIFF
--- a/src/pocketmine/level/format/io/BaseLevelProvider.php
+++ b/src/pocketmine/level/format/io/BaseLevelProvider.php
@@ -101,7 +101,7 @@ abstract class BaseLevelProvider implements LevelProvider{
 	}
 
 	public function getSpawn() : Vector3{
-		return new Vector3((float) $this->levelData["SpawnX"], (float) $this->levelData["SpawnY"], (float) $this->levelData["SpawnZ"]);
+		return new Vector3((float) $this->levelData["SpawnX"] + 0.5, (float) $this->levelData["SpawnY"], (float) $this->levelData["SpawnZ"] + 0.5);
 	}
 
 	public function setSpawn(Vector3 $pos){


jasonwynn10:
This isn't vanilla behavior

legoboy0215:
I like the idea though.

mal0ne-23:
Regardless of whether you like the idea or not, PMMP is made to simulate vanilla behavior. You can do anything you want with spawning using a plugin.

armagadon159753:
OK i understand but un vanilla players spawn on Biddle if thé block and not beside

TheDiamondYT1:
@armagadon159753 That comment contradicts your pull request...

Sandertv:
Players actually do spawn in the middle of a block in vanilla, as @armagadon159753 mentioned, not on the corner of a block. (Tested on 1.2 Beta)

TheDiamondYT1:
@Sandertv Oh, does this PR make players spawn in the middle of a block?

If so, this is a very confusing title. It sounds like the problem is spawning in the middle of the block and he is fixing it.

Sandertv:
Block coordinates are always the corner of a block. As this is the case, adding 0.5 to both X and Z will spawn the player in the middle.

armagadon159753:
@TheDiamondYT1 yes tested i working sorry for confuise

dktapps:
This behaviour is correct, but I'm not sure about the implementation... thoughts anyone?

For example, teleporting to integer coordinates will place you at an offset of +0.5 in MCPC.

armagadon159753:
@dktapps I have not yet tested the teleportation but I will keep

dktapps:
This ought to be handled by the teleport mechanism (teleporting to block positions offsets the player by 0.5 blocks).

armagadon159753:
thank you for paying attention ;) @dktapps

PR #1367, opened 2017-09-08
Overrides add function to return an object of the same class

GiantQuartz:
With the current PocketMine-MP Location/Position classes, if you call "add" or "substract" functions on a Position/Location instance, it'll return a Vector3 instance instead an object of the same class.

### DIFF
--- a/src/pocketmine/level/Location.php
+++ b/src/pocketmine/level/Location.php
@@ -88,4 +88,19 @@ class Location extends Position{
 		}
 		return parent::equals($v);
 	}
+    
+    /**
+     * @param Vector3|int $x
+     * @param int         $y
+     * @param int         $z
+     *
+     * @return Location
+     */
+    public function add($x, $y = 0, $z = 0){
+        if($x instanceof Vector3){
+            return new Location($this->x + $x->x, $this->y + $x->y, $this->z + $x->z, $this->yaw, $this->pitch, $this->level);
+        }else{
+            return new Location($this->x + $x, $this->y + $y, $this->z + $z, $this->yaw, $this->pitch, $this->level);
+        }
+    }
 }
--- a/src/pocketmine/level/Position.php
+++ b/src/pocketmine/level/Position.php
@@ -139,4 +139,19 @@ class Position extends Vector3{
 		}
 		return parent::equals($v);
 	}
+    
+    /**
+     * @param Vector3|int $x
+     * @param int         $y
+     * @param int         $z
+     *
+     * @return Position
+     */
+    public function add($x, $y = 0, $z = 0){
+        if($x instanceof Vector3){
+            return new Position($this->x + $x->x, $this->y + $x->y, $this->z + $x->z, $this->level);
+        }else{
+            return new Position($this->x + $x, $this->y + $y, $this->z + $z, $this->level);
+        }
+    }
 }


GiantQuartz:
I don't get why is it incomplete, and about the BC Break... Where is the problem? Location and Position inherit from Vector3.

SOF3:
About incomplete: You only added add(), but what about subtract() etc.? This will lead to a lot of confusion.

This change is BC Break because it will return Position objects when the caller expects a Vector3 object. Returning a Position/Location object includes an extra reference to a Level object, which, if handled inappropriately, might lead to memory leak. Some plugins <sup>[citation needed]</sup> even explicitly call `Vector3::asVector3()` to make sure the object contains only three scalar data.

I recommend using Position::fromObject() if you actually need to use Position.

Please note that add(), subtract(), multiply(), divide(), dot(), cross(), etc. are all _mathematical_ methods. A vector is a mathematical concept that can be used with addition, multiplication, cross product, etc., while Position/Location are only convenience subclasses used at the program level, and are not related to the mathematical concepts at all.

Let's look at this case. Suppose we do a similar thing to Vector3::subtract(), and we call $b->subtract($a) to find the vector from $a to $b. The result is a relative vector, not a position vector, so it is totally illogical to return a Position object. Even if you do, what should be the level in the result?
Similarly, when a Position object is passed to Position::add(), and the other Position object contains a different level, what do you do?

My point from the program philosophy's perspective is that Vector3::add() is a mathematical function and should not be mixed with PocketMine-specific things.

May I repeat, the BC impact is very l arge. This change is likely to result in a lot of memory leak that might be very difficult for plugin developers to track, and therefore I vote to reject this pull request.

SOF3:
Apart from memory leak, a Position object is also not serializable. Plugins might be passing the result of Vector3::add() to other threads (e.g. for AsyncTask), only to find out that this change is suddenly causing their originally-serializable Vector3 objects ending up with "Cannot serialize the Server instance" errors.

Muqsit:
A cleaner way, at least for add and subtract would be cloning the object and then use `setComponents()`.
```php
public function add($x, float $y = 0, float $z = 0) : self{
    $clone = clone $this;
    if($x instanceof Vector3){
        return $clone->setComponents($x->x, $x->y, $x->z);
    }
    return $clone->setComponents($x, $y, $z);
}
```

dktapps:
@Muqsit and if the object in question happens to be a heavy object, like a Player or Entity?

GiantQuartz:
About incomplete:
substract() uses add() to substract the specified amounts, so there is no need to override it.

About BC Break:
Creating a new method might be the solution, it is pretty annoying when you have to use similar positions/locations multiple times and you have to create new objects because add() does not return a Position/Location object. Creating this function might save us a lot of lines by [code reusing](https://en.wikipedia.org/wiki/Code_reuse).

pantichawanas:
i'm not sure about this - as @dktapps says , entity movement leak is pretty bad on 32 systems and it can make it worse ...

dktapps:
@pantichawanas do stop talking crap. We don't support 32-bit anymore, I never mentioned 32-bit. You're talking utter garbage.

pantichawanas:
@dktapps i not talk crap , me and genisys pro team have debugged this leak and will fix soon ... this pr makes leak worse as you said:

and if the object in question happens to be a heavy object, like a Player or Entity?

im not talk crap

IrishPacks:
My only question is why did you link a Wikipedia article about code reusing? Dylan, SOF3, and many others are already aware of what code reuse is, not to mention it's just common sense.

SOF3:
If you want convenience methods, create your own ones that clones the position and alter its components. This pull request causes too much backward incompatibility and it actually poses more inconvenience than convenience (because now you have to check if the instance is a Vector3 and not a Position to make sure it has no memory leak/thread safety). This pull request also creates a confusing situation when $a->add($b) gives a diff result from $b->add($a), while the add() method is supposed to create symmetric results. As a result, I am closing this pull request as rejected.

matcracker:
I think he should use the overloading. With that he can use two methods: `$this->add(Vector3 $vector3)` and `$this->add(int $x, int $y, int $z)`.

dktapps:
@matcracker PHP doesn't support overloading, please detach yourself from your Java knowledge.

dktapps:
did you actually even read the first sentence of that page?

matcracker:
Sure, but it's not a cool solution. It's a bit confusing

PR #1368, opened 2017-09-09
Update Grass.php

sero583:
Sorry for formatting, quickfix for Sugarcane killing grass block which should not happen in my opinon. Maybe a fix in the lightning code would also fix this instead of this quickfix

## Introduction
Grass simply turns to Dirt when a Sugarcane block is above it.

### Relevant issues

Didn't find any issue, also this is not a major issue or some kind of that.


dktapps:
We don't accept hacks.

sero583:
Update: @dktapps sorry for the trouble, I watched what other plugin did and they registered custom blocks, so this aint a bug and is not valid. Next time I better test it also without any plugins installed.

PR #1383, opened 2017-09-12
Quoted command args & allow playernames with spaces

dktapps:
## Introduction
blah

### API changes
- Plugins handling commands may now receive string arguments containing spaces.

### Behavioural changes
- players with spaces can now be handled on a server
- allow quoting any string in a command to instruct the parser to treat it as a single argument

I suppose I should mention this or someone is guaranteed to say something about it: If you, for whatever reason, previously happened to use quotes in command arguments for some reason, you will now need to escape those quotes using a backslash (`\`).
This currently also applies to things like `/say` where the argument should be rawtext and therefore quotes shouldn't be needed (this issue will be resolved when the new command API is written).

## Backwards compatibility
probably some issues, but I'm tired... TBD

## Tests
- tested joining with username `l o l`
- tested with `/op "l o l"`, works correctly

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -184,7 +184,7 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 	public static function isValidUserName(string $name) : bool{
 		$lname = strtolower($name);
 		$len = strlen($name);
-		return $lname !== "rcon" and $lname !== "console" and $len >= 1 and $len <= 16 and preg_match("/[^A-Za-z0-9_]/", $name) === 0;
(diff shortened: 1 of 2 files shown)


xZeroMCPE:
Great so far, nice job.

SOF3 (approved):

kenygamer:
Hmm the player names with spaces should be able to enable/disable from the `pocketmine.yml`

dktapps:
what would be the point of that? if you don't want to accept players with spaces after this is merged, I'm sure you're capable of writing a plugin to kick them.

PR #1393, opened 2017-09-21
Implemented FloatingTextParticle as an entity

dktapps:
This offers several major advantages over the original implementation:
- it will be automatically spawned to players when they join or get within viewing distance
- it will be automatically (de)spawned to players when changing levels
- it will be saved to disk with the world, so no need to recreate it

### API changes
- Added `FloatingText` entity
- Added `Level->addFloatingText(Vector3, string, string)`
- Deprecated `FloatingTextParticle`

## Backwards compatibility
- Currently backwards compatible, but I don't intend to maintain the old implementation. When it breaks next time, get ready to move to the new version.

## Tests
TBD

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -29,6 +29,7 @@ namespace pocketmine\level;
 use pocketmine\block\Block;
 use pocketmine\block\BlockFactory;
 use pocketmine\entity\Entity;
+use pocketmine\entity\FloatingText;
 use pocketmine\entity\Item as DroppedItem;
 use pocketmine\entity\object\ExperienceOrb;
 use pocketmine\entity\projectile\Arrow;
@@ -1665,6 +1666,26 @@ class Level implements ChunkManager, Metadatable{
 		}
 
 		return $orbs;
+	}
+
+	/**
+	 * @param Vector3 $pos
+	 * @param string  $text
+	 * @param string  $title
+	 *
+	 * @return null|Entity
+	 */
+	public function addFloatingText(Vector3 $pos, string $text, string $title = ""){
+		$entity = Entity::createEntity("FloatingText", $this, Entity::createBaseNBT($pos));
+		assert($entity !== null);
+
+		if($entity instanceof FloatingText){
+			$entity->setTitle($title);
+			$entity->setText($text);
+		}
+
+		$entity->spawnToAll();
+		return $entity;
 	}
 
 	/**
--- a/src/pocketmine/entity/Entity.php
+++ b/src/pocketmine/entity/Entity.php
@@ -238,6 +238,7 @@ abstract class Entity extends Location implements Metadatable, EntityIds{
 		Entity::registerEntity(Zombie::class, false, ['Zombie',	'minecraft:zombie']);
 
 		Entity::registerEntity(Human::class, true);
+		Entity::registerEntity(FloatingText::class, true);
 	}
 
 
--- a/src/pocketmine/entity/FloatingText.php
+++ b/src/pocketmine/entity/FloatingText.php
@@ -1,0 +1,101 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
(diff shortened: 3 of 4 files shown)


SOF3 (changes requested):
This change cannot fully replace the original FloatingTextParticle, as it does not retain the convenient ability to spawn different text to different players.

src/pocketmine/level/Level.php
```diff
@@ -1505,6 +1506,42 @@ public function dropItem(Vector3 $source, Item $item, Vector3 $motion = null, in
...
+			]),
+			new ListTag("Rotation", [
+				new FloatTag("", 0),
+				new FloatTag("", 0)
+			])
+		]));
+
+		assert($entity !== null);
+
+		if($entity instanceof FloatingText){
```
SOF3: Why is this not an assertion?

dktapps: plugins might override it with their own entities

dktapps:
@SOF3 if you have any proposals to easily resolve that problem, I'm all ears... personally I don't see why you couldn't just despawnFrom(Player) and then do whatever.

SOF3:
@dktapps FloatingTextParticle is already the best solution for this problem. Explicitly despawning for players is a lot of additional trouble compared with FloatingTextParticle.
And therefore my suggestion is simply not to deprecate FloatingTextParticle. It is a very convenient way of spawning individual text to players. Currently, the PocketMine API does not provide a convenient way of sending different entity data to the different players without additional trouble in handling spawn/despawns. Comparatively, FloatingTextParticle makes it much easier to manage. The only thing is the potential need to despawn the particle when the player switches level, which is a minor problem compared with all the complexity of entity spawning.

dktapps:
judging by the polls, everyone disagrees with you... including myself. You can specify a list of players to send specific entity data to in Entity->sendData().

zKoz210:
Time to accept the request! <3

SOF3 (changes requested):
Perhaps I should recap what's needed in this PR before more people start urging this to be merged. The main problem is in that FloatingTextParticle is deprecated but FloatingText (Entity) does not provide functionally similar alternatives to features provided by FloatingTextParticle, including:

- [x] It doesn't provide the ability to spawn FloatingText (Entity) that isn't saved to chunks. This might increase the difficulty for highly-specialized plugins (a.k.a. "server core" plugins) (they may hardcode what floating texts to spawn, so it would be spawned once every time without adding other code to check) to manage their floating texts.
  - This can be solved by adding a way to mark entities as "not saved", similar to how players are not saved, as in https://github.com/pmmp/PocketMine-MP/blob/8adcd64/src/pocketmine/level/format/io/leveldb/LevelDB.php#L514
  - But how do we handle unloaded chunks? Unlike players, FloatingText does not constitute a legitimate reason to keep a chunk loaded. Plugins might lose a reference or hold invalid reference to a FloatingText they spawned, assuming that it is still valid while it was actually despawned due to chunk unloading.
- [ ] It is inconvenient, although possible, to spawn dynamic FloatingText
  - Problems include:
    - [ ] Only spawning to certain players
      - This can be done through explicitly despawning/spawning, but the code flow looks redundant.
    - [ ] Spawning different text content to different players
      - This can be done through explicit sendData(), but this kills the "convenient" nature of FloatingText; why don't plugins just spawn a Human instead of going through all this trouble?
    - [ ] Updating the text of the entity actively (actively = not due to spawning to players/despawning from players/loading chunks)
      - There is currently no convenient way to get the instance of the FloatingText from a Level through any means of identification, without iterating all entities and possibly checking the custom NBT data injected by the plugin.

I agree that all these can be solved by plugins, but not through much more code compared to FloatingTextParticle. This kills the point of creating a FloatingText entity type in addition to the Human type.

dktapps:
A possible solution to the problem would be to make the FloatingText entity a wrapper around FloatingTextParticle. To some degree this would offer the best of both worlds - not removing the original functionalities, while making it easier for floating text to be conveniently automatically spawned and despawned.

dktapps:
> A possible solution to the problem would be to make the FloatingText entity a wrapper around FloatingTextParticle. 

this is more trouble than it's worth because of the hacky nature of FloatingTextParticle. It's impossible to preserve backwards compatibility here.

dktapps:
I've closed this due to lack of progress on discussed issues. While I would like to just merge this directly, I am unable to provide working solutions to the problems described by @SOF3 at this point in time.

PR #1437, opened 2017-10-01
Fix coloured message

Johnmacrocraft:
## Introduction

I thought I fixed it on #910, but I didn't think about if TextFormat::RESET is used on console, it'll change to gray, not white like in client. (Haven't tested it on second part (when I changed it from white to reset) because I thought it has no problem, sorry)

~~Blame @Muqsit!! feat. dktapps - BlameShoghicp Records~~

### Relevant issues

Nothing.

### API changes

Nothing.

### Behavioural changes

Now these two messages are shown as expected.

## Backwards compatibility

Nothing.

## Follow-up

Probably also fix stuffs like server reload command message.

## Tests

Works fine. Tested on local.

### DIFF
--- a/src/pocketmine/Achievement.php
+++ b/src/pocketmine/Achievement.php
@@ -115,7 +115,7 @@ abstract class Achievement{
 	 */
 	public static function broadcast(Player $player, string $achievementId) : bool{
 		if(isset(Achievement::$list[$achievementId])){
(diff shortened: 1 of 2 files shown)


dktapps:
I'm not sure I understand what the problem is... RESET simply resets the colour to default, no? I don't like hardcoding white, what if I change my console colours?

Johnmacrocraft:
@dktapps I guess gray is default color in Mintty, so reset will reset the color to gray on it.
If this pull request gets rejected, you'll probably need to change this part on Server.php:
> ($version->isDev() ? TextFormat::YELLOW : "") . $version->get(true) . TextFormat::WHITE

PR #1446, opened 2017-10-03
Fixed canUseChunkCache() logic

xxFlare:
This function did not work correctly before due to skewed logic. This should be the correct behavior.

### DIFF
--- a/src/pocketmine/MemoryManager.php
+++ b/src/pocketmine/MemoryManager.php
@@ -72,7 +72,7 @@ class MemoryManager{
 	private $chunkTrigger;
 
 	/** @var bool */
-	private $chunkCache;
+	private $disableChunkCacheOnLowMemory;
 	/** @var bool */
 	private $cacheTrigger;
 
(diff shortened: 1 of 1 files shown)


madecode15:
ever heard of dont fix whats not broke? -.-

src/pocketmine/MemoryManager.php
```diff
@@ -151,7 +151,7 @@ public function isLowMemory() : bool{
 	 * @return bool
 	 */
 	public function canUseChunkCache() : bool{
-		return !($this->lowMemory and $this->chunkTrigger);
+		return !$this->lowMemory or !$this->chunkCache;
```
dktapps: the field name is very confusing, I'm open to a better name. It actually indicates whether the chunk cache should be disabled on low memory.

Just to verify your logic:
the old condition: NOT low memory, AND clear cache on low memory. (bad, nonsense?)
the new condition: NOT low memory, OR NOT disable chunk cache on low memory. (good)

You can see how this would be confusing to future viewers.

dktapps:
@madecode15 it looks very much broken to me... if you were to disable the low memory cache clear for chunks, the cache would not operate at all with the current logic.

dktapps:
Looks good, thanks!

dschwartz783:
The old logic makes more sense to me. If the memory is low, AND the cache is disabled on low memory, don't use the cache. Any other condition would result in the cache being used. What am I missing?

dktapps:
@dschwartz783 the old cond doesn't make sense, it used the wrong field. It was instead checking whether the chunk cache should be cleared on a LowMemoryEvent, as opposed to whether the chunk cache should be disabled altogether on low memory. 

The new condition is functionally the same as the old one, apart from using the correct field and being less confusing to read.

PR #1452, opened 2017-10-08
Added API to allow flagging an entity not to be saved to disk when it…

dktapps:
…s chunk is saved

## Introduction
Sometimes plugins want to create an entity that only exists for the runtime of the server, and don't want the entity getting saved to disk.

Currently achieving this goal can be pretty challenging because plugins have to hook events to listen for chunk saving, and then remove any entities which they don't want saved. This can be particularly problematic during server shutdown, as plugins are disabled before levels are unloaded.

This PR adds the capability to allow flagging an entity as non-permanent so that it doesn't get saved to disk when the chunk is saved.

## Relevant issues
#1393 

### API changes
- Added methods `Entity->canSaveWithChunk()` and `Entity->setCanSaveWithChunk()`

## Backwards compatibility
- No BC breaks here, only API additions.

## Tests
TBD

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -3376,6 +3376,14 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 		return [];
 	}
 
+	public function canSaveWithChunk() : bool{
+		return false;
+	}
+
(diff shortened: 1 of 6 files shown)


mal0ne-23:
tested and works

SOF3 (approved):

PR #1453, opened 2017-10-08
Peaceful difficulty allows PvP

matcracker:
## Introduction

Worlds with peaceful difficulty not allows players to hit themself. 

### Behavioural changes

Now peaceful difficulty allows PvP.


dktapps:
Note to self for future reference: Damage dealt to the player in Peaceful is the same as in Easy mode. The difference is that the player can regenerate in Peaceful.

PR #1455, opened 2017-10-09
Modified misleading comment

jojoe77777:
## Introduction

> Plugins that create tasks MUST extend PluginTask

This implies that the plugin must extend PluginTask, which is incorrect.

Changed to:
> Tasks created by plugins MUST extend PluginTask


dktapps:
Merged, thanks!

PR #1461, opened 2017-10-11
added a hacky pocketmine.yml auto updater

### LINKED ISSUE #450: pocketmine.yml version control
We either need to rebuild configs or warn users their config is outdated. The problem with this is that libyaml cannot retain comments in yml files.

dktapps:
## Introduction
pocketmine.yml has been a pain for a long time because new configuration options don't appear in it automatically when the stock one gets changed. This is because parsing and emitting a yml file destroys all comments within it, and pocketmine.yml is heavily annotated with explaining comments.

This PR solves the issue by converting the comments into YAML keys before parsing, so that values from the old config can be set into the new one without destroying comments. After the copying is done, the fake keys are then converted back into normal comments.

This requires that indentation of comments in the stock pocketmine.yml be at the same level as the settings they describe. End-of-line comments will not be retained.

### Relevant issues
close #450

### Behavioural changes
pocketmine.yml is now automatically updated, retaining original configuration options where applicable. The resulting output format is slightly different due to yaml generation style.

## Tests
lightly tested, needs more work


dktapps:
This needs a proper way to identify sections which should retain all user configs instead of hardcoding them (aliases, worlds)

dktapps:
I think this might be finally ready for testing & review.

fuyutsuki:
Confirmed that `pocketmine.yml.bak` file was generated and updated properly.

dktapps:
This is a very bad and limited solution which is not fit for production in its current form. Additionally, it targets one specific use case and doesn't scale well to other uses.

I'm going to close this, but the branch will be retained for further work.

PR #1462, opened 2017-10-13
Implement ender chest

JackNoordhuis:
## Introduction
Adds a basic vanilla ender chest implementation.

### API changes
~~`Tile::addAdditionalSpawnData()` has been changed from an abstract function to an empty one due to ender chest tiles not needing to add additional spawn data to the nbt. This does not affect the existing tile implementation.~~

## Backwards compatibility
Everything should be working fine 💯 

## Tests
- /give @p ender_chest
- Place ender chest
- Profit!


jojoe77777:
> /give @p ender_chest

You sure about that?

JackNoordhuis:
> You sure about that?

Would've thought the command selector stuff would've been merged by now.

Although PocketMine doesn't have command selectors implemented, I would've thought @p is a pretty self-explanatory replacement for 'player_name_here'.

src/pocketmine/block/EnderChest.php
```diff
@@ -0,0 +1,110 @@
...
+			$player->getEnderChestInventory()->openAt($enderChest);
+			$player->addWindow($player->getEnderChestInventory());
+		}
+
+		return true;
+	}
+
+	public function getDrops(Item $item) : array{
+		return [
+			ItemFactory::get(Item::ENDER_CHEST, 0, 8)
```
Muqsit: Ender chests drop 8 obsidian. That drop is only applicable if the item has the `Enchantment::SILK_TOUCH` enchant.

Muqsit:
EnderChest::getBlastResistance() should return 3000 as per https://minecraft.gamepedia.com/Ender_Chest

src/pocketmine/block/EnderChest.php
JackNoordhuis: Yea, I was halfway through looking at other blocks to double check silk touch hadn't been implemented and forgot to change the item id back to obsidian.

Muqsit (approved):

src/pocketmine/entity/Human.php
```diff
@@ -546,6 +571,7 @@ public function close(){
 				}
 
 				$this->inventory = null;
+				$this->enderChestInventory = null;
```
dktapps: you aren't removing it from viewers, which might cause leaks if plugins open it to other players.

src/pocketmine/inventory/EnderChestInventory.php
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	private function broadcastBlockEventPacket(int $case1, int $case2){
+		$pk = new BlockEventPacket();
+		$pk->x = $this->getHolder()->getX();
+		$pk->y = $this->getHolder()->getY();
+		$pk->z = $this->getHolder()->getZ();
+		$pk->case1 = $case1;
+		$pk->case2 = $case2;
+	}
```
dktapps: the code in here looks similar to that of ChestInventory, can it be reused instead of duplicating?

src/pocketmine/inventory/EnderChestInventory.php
```diff
@@ -0,0 +1,110 @@
...
+
+	public function __construct(Human $owner){
+		$this->owner = $owner;
+		parent::__construct(new FakeBlockMenu($this, $owner->getPosition()));
+	}
+
+	/**
+	 * @return Human
+	 */
+	public function getOwner(){
```
dktapps: this can be typehinted Human

src/pocketmine/inventory/EnderChestInventory.php
```diff
@@ -0,0 +1,110 @@
...
+	public function getDefaultSize() : int{
+		return 27;
+	}
+
+	/**
+	 * Set the holders position to that of a tile
+	 *
+	 * @param EnderChest $enderChest
+	 */
+	public function openAt(EnderChest $enderChest) {
```
dktapps: formatting issue

src/pocketmine/tile/Spawnable.php
```diff
@@ -96,7 +96,9 @@ protected function onChanged(){
 	 *
 	 * @param CompoundTag $nbt
 	 */
-	abstract public function addAdditionalSpawnData(CompoundTag $nbt);
+	public function addAdditionalSpawnData(CompoundTag $nbt){
+
```
dktapps: I would prefer this be explicitly overridden in EnderChest to make the fact clear, otherwise this may be unintentionally missed for other things.

src/pocketmine/entity/Human.php
```diff
@@ -341,6 +350,12 @@ protected function initEntity(){
 			}
 		}
 
+		if(isset($this->namedtag->EnderChestInventory) and $this->namedtag->EnderChestInventory instanceof ListTag){
+			foreach($this->namedtag->EnderChestInventory as $i => $item){
+				$this->enderChestInventory->setItem($item["Slot"] - 9, ItemItem::nbtDeserialize($item));
```
dktapps: why slot - 9? there's no hotbar to deal with here

dktapps:
this looks interesting, thanks for the PR!

src/pocketmine/entity/Human.php
```diff
@@ -569,8 +569,11 @@ public function close(){
 				foreach($this->inventory->getViewers() as $viewer){
 					$viewer->removeWindow($this->inventory);
 				}
-
 				$this->inventory = null;
+
+				foreach($this->enderChestInventory->getViewers() as $viewer){
+					$viewer->removeWindow($this->inventory);
```
dktapps: wrong window

dktapps:
no need to rush by the way, take your time 😉 I prefer a well-done PR over a quickly-done one.

dktapps:
I feel bad for `Oleg Pudeyev`, must get pinged everywhere on MC-related projects.

src/pocketmine/block/EnderChest.php
```diff
@@ -0,0 +1,114 @@
...
+		return Tool::TYPE_PICKAXE;
+	}
+
+	protected function recalculateBoundingBox() : ?AxisAlignedBB{
+		return new AxisAlignedBB(
+			$this->x + 0.0625,
+			$this->y,
+			$this->z + 0.0625,
+			$this->x + 0.9375,
+			$this->y + 0.9475,
```
dktapps: is this a typo or intended?

src/pocketmine/inventory/EnderChestInventory.php
```diff
@@ -0,0 +1,81 @@
...
+	public function getDefaultSize() : int{
+		return 27;
+	}
+
+	/**
+	 * Set the holders position to that of a tile
+	 *
+	 * @param EnderChest $enderChest
+	 */
+	public function openAt(EnderChest $enderChest){
```
dktapps: this function name is a little misleading - it's not actually opening the chest, merely altering its target.

JackNoordhuis: Would you rather the`Player::addWindow()` call moved to here or just change the function name?

dktapps: Just change the function name - plugins might allow other players to view each others' ender chest inventories.

src/pocketmine/block/EnderChest.php
JackNoordhuis: typo, will fix.

dktapps: according to tests with MCPE, the bounding boxes are actually +0.05 and +0.95 offset, with +0.95 height, not 0.0625. You get kicked for flying when walking on top of chests. That will need to be fixed separately.

Regardless, this class should be able to extend Chest, no?

dktapps: actually that doesn't even make sense... I see 0.0625 in IDA, but my actual in-game collision shows 0.05. Needs more investigation.

JackNoordhuis: Making the class extend chest should fix it, one would think it safe to assume chest and ender chests would share the same geometric properties.

Muqsit:
Hopefully this gets merged soon, can't wait to create an /echest plugin 😀

src/pocketmine/block/EnderChest.php
Muqsit: yeah, their bbs look the same
cd

src/pocketmine/tile/EnderChest.php
```diff
@@ -0,0 +1,34 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\tile;
+
+use pocketmine\nbt\tag\CompoundTag;
+
+class EnderChest extends Spawnable{
+
+	public function addAdditionalSpawnData(CompoundTag $nbt){
```
Infernus101: : void

XDqrkneZ (approved):

jasonwynn10:
Isn't the bounding box on an ender chest smaller than a normal or trapped chest?

andris155:
Bug: Remove item Ender Chest -> Rejoin -> Removed item find again Ender Chest.

andris155:
https://www.youtube.com/watch?v=E0NQrsvc-oc

dktapps (changes requested):
This appears to be dysfunctional when used with player data which was not previously generated by Genisys (items are deleted).

src/pocketmine/entity/Human.php
```diff
@@ -501,6 +516,16 @@ public function saveNBT(){
 			$this->namedtag->SelectedInventorySlot = new IntTag("SelectedInventorySlot", $this->inventory->getHeldItemIndex());
 		}
 
+		if($this->enderChestInventory !== null){
+			$slotCount = $this->enderChestInventory->getSize();
+			for($slot = 0; $slot < $slotCount; ++$slot){
+				$item = $this->enderChestInventory->getItem($slot);
+				if($item->getId() !== ItemItem::AIR){
+					$this->namedtag->EnderChestInventory[$slot] = $item->nbtSerialize($slot);
```
dktapps: If the player's data was previously generated by PocketMine-MP (not spoons) this tag won't exist. This means that you're performing array-access on null, which results in the items silently being destroyed on quit.

dktapps (changes requested):

dktapps:
I've resolved merge conflicts, please remember to pull before making further changes.

src/pocketmine/block/EnderChest.php
```diff
@@ -0,0 +1,109 @@
...
+			return [ItemFactory::get(Item::OBSIDIAN, 0, 8)];
+		}
+
+		return [];
+	}
+
+	public function getFuelTime() : int{
+		return 0;
+	}
+
```
dktapps: light level emission is missing

dktapps:
thanks, I'll re-review when I get home later.

zKoz210:
Remove tag `Waiting on Author` 😄

dktapps:
> Bug: Remove item Ender Chest -> Rejoin -> Removed item find again Ender Chest.

I can also reproduce this bug.

JackNoordhuis:
I only managed to reproduce it once while testing. I'll see if I can find the cause of the item duplication when I have time.

zKoz210:
> Bug: Remove item Ender Chest -> Rejoin -> Removed item find again Ender Chest.

Fixed https://github.com/pmmp/PocketMine-MP/commit/bbbed22467d060a865fa3e03fc28c81185069d2e

@JackNoordhuis Please resolve conflicts for further testing

dktapps:
I reproduced the bug with no plugins, so that's not the problem.

zKoz210:
@dktapps I just checked it, I have no problems with saving things.

dktapps:
@zKoz210 that bug requires a plugin's interference to cause problems. I reproduced this issue without plugins.

What is actually happening here is a combination of [this](https://github.com/pmmp/PocketMine-MP/pull/1462/files#diff-0fd2f8cb04cdc4d4181aa065c6306f76R521) and [this](https://github.com/pmmp/PocketMine-MP/pull/1462/files#diff-0fd2f8cb04cdc4d4181aa065c6306f76R528). Since the `EnderChestInventory` is not replaced with an empty one on save, before writing the items, what happens is this:

1. Items are loaded from NBT
2. Player removes some items from their ender chest (so that the slot becomes air) BUT the NBT is not updated (as expected).
3. Player's data is saved, the `EnderChestInventory` tag already exists, so the old one is reused (containing the original items that were in there when the player joined)
4. Only non-air slots are written to the NBT (https://github.com/pmmp/PocketMine-MP/pull/1462/files#diff-0fd2f8cb04cdc4d4181aa065c6306f76R528), overwriting the originals. This means that if a slot that was previously not air is cleared, then that slot will still retain its original item, resulting in a dupe on rejoin.

Steps to reproduce:
1. Put some items in an ender chest
2. quit the server and rejoin
3. take a stack of items out of the ender chest (clear a slot)
4. quit and rejoin
5. open the ender chest. The item stack you just deleted from the ender chest is still there.

zKoz210:
@dktapps I realized my mistake! Thank you)

I found problem in line: https://github.com/pmmp/PocketMine-MP/pull/1462/files#diff-0fd2f8cb04cdc4d4181aa065c6306f76R528

src/pocketmine/entity/Human.php
```diff
@@ -502,6 +517,20 @@ public function saveNBT(){
 			$this->namedtag->SelectedInventorySlot = new IntTag("SelectedInventorySlot", $this->inventory->getHeldItemIndex());
 		}
 
+		if($this->enderChestInventory !== null){
+			if(!$this->namedtag->hasTag("EnderChestInventory", ListTag::class)){
```
dktapps: This should not be conditional (see PR comments)

dktapps (changes requested):

dktapps (dismissed):

dktapps:
There is an issue with this implementation resulting from a cyclic reference between FakeBlockMenu and its inventory which prevents either one being destroyed until GC kicks in. 
While this issue already exists with things like AnvilInventory, the problem is larger with EnderChestInventory because it references its `owner` Player directly, which results in the Player being leaked when they disconnect from the server.

dktapps (approved):

dktapps:
Looks good as far as I can see now, any further changes/reviews to make?

zKoz210 (approved):

zKoz210:
@dktapps I think there is nothing more to add to it. You will be able to reviews it now?

jasonwynn10:
Still need to add Light level emission

zKoz210:
@jasonwynn10 https://github.com/pmmp/PocketMine-MP/pull/1462/files#diff-7c6a35148cb6d2ae2b0e2fd18f73a3a4R48

PR #1463, opened 2017-10-13
Fix for explosions not removing tiles

### LINKED ISSUE #1450: explosions don't remove tiles

JackNoordhuis:
## Introduction
Prior to the commit below, blocks would be removed but any tiles attached to the block would remain (#1450).

### Relevant issues

* Fixes #1450

## Follow-up
Disclaimer: This does not fix tiles that have not been removed by previous explosions, you should use a plugin to achieve this.

## Tests
None – it's a pretty straightforward solution.

### DIFF
--- a/src/pocketmine/level/Explosion.php
+++ b/src/pocketmine/level/Explosion.php
@@ -40,6 +40,8 @@ use pocketmine\math\Math;
 use pocketmine\math\Vector3;
 use pocketmine\network\mcpe\protocol\ExplodePacket;
 use pocketmine\network\mcpe\protocol\LevelSoundEventPacket;
+use pocketmine\tile\Chest;
+use pocketmine\tile\Container;
 use pocketmine\tile\Tile;
 
 class Explosion{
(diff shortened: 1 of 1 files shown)


dktapps:
This will degrade explosion performance, which is why I didn't do it like this myself. Whether the degradation is acceptable will require tests to see.

Muqsit:
Maybe add a BlockEntity class and extend Chests, Furnaces, Skulls etc to the BlockEntity class?

dktapps:
or an interface... I'm planning to add an interface like that anyway for automatic tile creation for relevant blocks.

dktapps:
I'll test this later and see what the difference is like. Thanks for the PR.

dktapps:
There is another problem here: the bug I tested was chest tiles not getting destroyed. When destroyed, they ought to drop their contents. This will simply delete them outright.

src/pocketmine/level/Explosion.php
```diff
@@ -209,7 +213,20 @@ public function explodeB() : bool{
 
 			$t = $this->level->getTile($block);
 			if($t instanceof Tile){
```
Muqsit: Better not call `Level::getTile()` if `$yieldDrops` is not true
```php
if($yieldDrops){
    $t = $this->level->getTile($block);
    if($t instanceof Container){
        //...
    }
}
```

dktapps: @Muqsit and what about tiles that need to be removed which are not containers?

Muqsit: updated.

dktapps:
Looking good apart from some minor formatting issues.

PR #1468, opened 2017-10-14
Renamed swapPage to swapPages and cleaned up page pushing.

Sandertv:
## Introduction

Renamed WritableBook->swapPage() to WritableBook->swapPages() and cleaned up page pushing.
Doesn't have any behavioural changes.

### DIFF
--- a/src/pocketmine/item/WritableBook.php
+++ b/src/pocketmine/item/WritableBook.php
@@ -195,7 +195,11 @@ class WritableBook extends Item{
 			if($downwards){
 				unset($namedTag->pages->{$key});
 			}
-			$namedTag->pages->{$key + $type} = clone $page;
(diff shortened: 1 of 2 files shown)


src/pocketmine/item/WritableBook.php
```diff
@@ -195,10 +195,7 @@ private function pushPages(int $pageId, CompoundTag $namedTag, bool $downwards =
 			if($downwards){
 				unset($namedTag->pages->{$key});
 			}
-			$namedTag->pages->{$key + $type} = new CompoundTag("", [
-				new StringTag("text", $page->text->getValue()),
-				new StringTag("photoname", "")
-			]);
+			$namedTag->pages->{$key + $type} = clone $page;
```
dktapps: As discussed internally I'm not clear why this needs to be cloned at all.

PR #1474, opened 2017-10-17
Fix error related server wizard

nnnlog:
## Introduction
When you install server, escape string("\n") doesn't work.
```
[?] Language (eng): eng
[*] English has been correctly selected.
[*] Welcome to PocketMine-MP!\nBefore starting setting up your new server you have to accept the license.\nPocketMine-MP is licensed under the LGPL License,\nthat you can read opening the LICENSE file on this folder.
```
## Tests

Tested


mal0ne-23:
This might be something we'd want to update in the language file directly, as this looks a little hacky

SOF3 (changes requested):
This change should apply to all messages, not just this particular one.

dktapps:
The referenced commit applies a better solution, so this has been closed. Thanks for the PR anyway.

PR #1476, opened 2017-10-17
Forms API

dktapps:


src/pocketmine/form/MenuForm.php
```diff
@@ -0,0 +1,142 @@
...
+	public function getType() : string{
+		return Form::TYPE_MENU;
+	}
+
+	public function getOption(int $position) : ?MenuOption{
+		return $this->options[$position] ?? null;
+	}
+
+	/**
+	 * Returns the index of the option selected by the user. Pass this to {@link getButton} to get the button object
```
SOF3: Dead link to `getButton`. I believe this should be `Pass this to {@link MenuForm::getOption} to get the MenuOption object`.

src/pocketmine/form/MenuForm.php
```diff
@@ -0,0 +1,142 @@
...
+	/**
+	 * Returns the menu option selected by the user.
+	 *
+	 * @return MenuOption
+	 * @throws \InvalidStateException if no option is selected or if the selected option doesn't exist
+	 */
+	public function getSelectedOption() : MenuOption{
+		$index = $this->getSelectedOptionIndex();
+		if($index === null){
+			throw new \InvalidStateException("No option selected, maybe the form hasn't been submitted yet");
```
SOF3: Consider changing the message to `Form is closed or has not been submitted yet` to be more precise.

src/pocketmine/form/MenuForm.php
```diff
@@ -0,0 +1,142 @@
...
+	 */
+	public function getSelectedOptionIndex() : ?int{
+		return $this->selectedOption;
+	}
+
+	/**
+	 * Sets the selected option to the specified index or null. null = no selection.
+	 * @param int $option
+	 */
+	public function setSelectedOptionIndex(int $option) : void{
```
SOF3: Is there a practical reason to expose this method publicly?

src/pocketmine/form/ModalForm.php
```diff
@@ -0,0 +1,119 @@
...
+	public function getChoice() : ?bool{
+		return $this->choice;
+	}
+
+	/**
+	 * Sets the option selected by the player. true = Yes, false = No
+	 *
+	 * @param bool $choice
+	 */
+	public function setChoice(bool $choice) : void{
```
SOF3: Is there a reason to expose this method publicly?

src/pocketmine/form/Form.php
```diff
@@ -0,0 +1,108 @@
...
+	 * they wish.
+	 *
+	 * @param Player $player
+	 */
+	abstract public function onSubmit(Player $player) : void;
+
+	/**
+	 * Clears response data from a form, useful if you want to reuse the same form object several times.
+	 */
+	public function clearResponseData() : void{
```
SOF3: This method is very confusing. It is not called by the server, but relies on the developer to call it.

This leads to inconsistency:
1. Is a Form object expected to be reused for a single player?
2. Is a Form object expected to be reused for multiple players?
3. Can the developer use the form values outside onSubmit()?

The answer is:
* If the same Form instance is only sent once, you can use the form values outside onSubmit() and hold the Form instance forever.
* If the same Form instance is sent more than once (same player or not), it would be a very bad practice to depend on the value held in the Form instance outside onSubmit(), even in the same tick.

It is particularly tempting to hold a CustomForm instance and use the values there later, since the data cannot be easily extracted. I can foresee this is going to cause a lot of concurrency bugs that can be very difficult to identify. Therefore, it is suggested that, starting from the core, one of either approach ("resend same instance" or "new instance each resend") be adopted to avoid confusion.

* Adopting the "resend same instance" approach should automatically call the `clearResponseData()` method after each onSubmit() call.
* Adopting the "new instance each resend" approach should remove this method and add a flag to identify a Form as sent and throw an exception when attempting to use sendForm() with it.

Deciding which approach to adopt should require public consultation.

src/pocketmine/form/MenuOption.php
```diff
@@ -0,0 +1,69 @@
...
+		];
+
+		if($this->hasImage()){
+			$json["image"] = $this->image;
+		}
+
+		return $json;
+	}
+
+}
```
SOF3: What about also adding an empty `public function onSelected()`, which is called before/after `MenuForm::onSubmit()` is called? This can allow developers to write their handlers in the menu options instead of from the big MenuForm object, resulting in extra routing work. Without adding this in the core, developers have to add a few lines of routing code + assertion that the `MenuOption::onSelected()` method exists in order to put the handler code with the MenuOption.

src/pocketmine/form/ServerSettingsForm.php
```diff
@@ -0,0 +1,61 @@
...
+
+namespace pocketmine\form;
+
+use pocketmine\form\element\CustomFormElement;
+
+/**
+ * Represents a custom form which can be shown in the Settings menu on the client. This is exactly the same as a regular
+ * CustomForm, except that this type can also have an icon which can be shown on the settings section button.
+ */
+class ServerSettingsForm extends CustomForm{
```
SOF3: All other Form classes are abstract. Why is this suddenly non-abstract?

SOF3 (dismissed):
There seems to be an inconsistency, where calling getters of form input without form submission sometimes gets a `null` but sometimes gets an `InvalidStateException`.
Throwing an exception may also result in some dependence (the assumption that "if I don't get this exception, I am OK"), which does not work as expected when someone tries to get the value from a form they resent but is not yet resubmitted.

// I have not reviewed the code under `pocketmine\form\element` yet.

src/pocketmine/form/MenuOption.php
dktapps: before, or after? If a developer does it themselves, they are explicit about when they want it to be called, which can be any time - if the core does it, then developers have to get used to BEFORE, or AFTER, no in-betweens.

src/pocketmine/form/ModalForm.php
dktapps: Plugins may want to modify form result data when it's submitted (there will be events for this)

src/pocketmine/form/element/Label.php
```diff
@@ -0,0 +1,48 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\form\element;
+
+/**
+ * Element which displays some text on a form.
+ */
+class Label extends CustomFormElement{
+
```
thebigsmileXD: Suggestion: Construct with content -> new Label("content here");

src/pocketmine/form/element/Dropdown.php
```diff
@@ -0,0 +1,131 @@
...
+	protected $selectedOption;
+	/** @var string[] */
+	protected $options;
+
+	/**
+	 * @param string   $text
+	 * @param string[] $steps
+	 * @param int      $defaultStepIndex
+	 */
+	public function __construct(string $text, array $steps, int $defaultStepIndex = 0){
```
thebigsmileXD: Calling it Steps makes it confusing - If possible, rename to $options

inxomnyaa (changes requested):

src/pocketmine/Player.php
```diff
@@ -3279,6 +3279,11 @@ public function sendWhisper(string $sender, string $message){
 	}
 
 	public function sendForm(Form $form) : void{
+		if($form->hasBeenSent()){
+			throw new \InvalidArgumentException("Cannot send the same form more than once, create a new one instead");
```
thebigsmileXD: But: Why? So i can not cache a Form and send it to a player? So it has to be recreated (rebuildt in the core) every time? Isn't that just a waste of calculation if the form contents would never change?

dktapps: The core code has no foolproof way to tell if your form is immutable (so to speak). You can clone the form object if you're concerned about the overhead.

thebigsmileXD:
"Offtoppic" question: Could you merge the changes from master into the branch / update it?

src/pocketmine/Player.php
SOF3: I have already explained in my previous comment:

> It is particularly tempting to hold a CustomForm instance and use the values there later, since the data cannot be easily extracted. I can foresee this is going to cause a lot of concurrency bugs that can be very difficult to identify. Therefore, it is suggested that, starting from the core, one of either approach ("resend same instance" or "new instance each resend") be adopted to avoid confusion.
>
> * Adopting the "resend same instance" approach should automatically call the clearResponseData() method after each onSubmit() call.
> * Adopting the "new instance each resend" approach should remove this method and add a flag to identify a Form as sent and throw an exception when attempting to use sendForm() with it.

As it has been noticed that resetting contents each time makes it very inconvenient to retain data (especially inconvenient for custom form data), the "new instance each resend" approach is considered more appropriate.

SOF3: Regarding the overhead of form creation overhead, perhaps you should compare it against encoding the form, sending it to the client, making the client display it and sending it back. As one wouldn't reasonably send forms very frequently, the impact on performance is negligible, so trying to cache forms for the sake of performance would be premature optimization, and therefore your comment is 4√(all evil), and so you are 8√(all evil).

src/pocketmine/form/ModalForm.php
```diff
@@ -0,0 +1,119 @@
...
+
+use pocketmine\Player;
+
+/**
+ * This form type presents a simple "yes/no" dialog with two buttons.
+ */
+abstract class ModalForm extends Form{
+
+	/** @var string */
+	private $content;
```
thebigsmileXD: Please concider adding a function to get/set the content

inxomnyaa (changes requested):

AyyItzJ4y:
how do you make a player open a form when they hold click a specific block

thebigsmileXD:
Please, for questions head over to the Forums. This is a place to discuss the changes and issues, not some random guy's issues about not reading the docs / commits

YaBoiLewis:
Have images handling been added yet, if so, there should be a class for that.

dktapps:
Have eyes been added yet, if so, they should be used.

dktapps:
_wishes github had a facepalm reaction_

thebigsmileXD:
🤦 🌚 👌🏼  
image
https://userstyles.org/styles/37035/github-dark

VCraftMCPE:
Wiki for that or read it manu...

thebigsmileXD:
@SOF3 May you explain the queing feature? In general, is there a forums thread yet? This could keep offtoppic questions out. If so, edit my message and put the link here: <link>

SOF3 (approved):
Much more can be added (e.g. events, pre-send manipulation, form merging), but so far the API seems to be expandable enough for the long term.

dresnite (approved):

thebigsmileXD:
Found something whereas i am not sure if intended to behave like that or a bug:

When returning $this as ?Form in onSubmit, It sends the same formular again after having it submitted, when submitting a second time it does not though.

dktapps:
sounds like a bug to me, you shouldn't be able to send the same form twice at all.

SOF3:
Potential API expansion: Since the form may be queued for some time, data may need refreshing when sent. Although it cannot be actively updated while the player is opening the form anyway, less outdated is better than more outdated.
Therefore I'm suggesting that MenuOption[] etc. be changed to Traversable or similar stuff.

inxomnyaa (commented):
But: why?

SOF3:
Why what?
In case you're referring to f87b23b, varargs are not used for two reasons:

#### Forward compatibility adding new constructor args
WIth varargs used, it is impossible to add new optional parameters to the method without breaking backward compatibility (BC). This problem is especially important in constructors, because for other methods (e.g. `Inventory::addItem(Item ...$item)`), an overloading method can be added (e.g. adding `bool $send = true` can be done by simply adding another method `Inventory::addItemSend(bool $send, Item ...$item)` or just `Inventory::addItemNoSend(Item ...$item)`). But since a constructor in PHP cannot be overloaded (it's usually impossible to overload a pure varargs method anyway even in languages with method overloading like Java), varargs should be avoided.

#### Consistency in subclasses <sup>[citation needed]</sup>
By convention <sup>[citation needed]</sup>, the parameters in a subclass constructor should start with the superclass constructor's parameters first, then the subclass's own parameters (except for optional parameters <sup>[citation needed]</sup>). Using varargs in a superclass constructor makes the order of parameters in the subclass constructor very confusing.

thebigsmileXD:
But using parameters whose object types are not named in the function parameters is not confusing?

SOF3:
Then it's PHP's problem. We have already tried our best to improve this &mdash; assert type validity and PHP doc. Varargs affects the fundamental program design, and is not desirable in the long term just because PHP does not support array-of types.

HipsterAF:
switch to Python?

src/pocketmine/form/ServerSettingsForm.php
```diff
@@ -23,20 +23,21 @@
...
+ * Passing this form to {@link Player::sendForm()} will not show a form with an icon nor set this form as the server
+ * settings.
  */
 abstract class ServerSettingsForm extends CustomForm{
 	/**
 	 * @var FormIcon|null
 	 */
 	private $icon;
 
-	public function __construct(string $title, ?FormIcon $icon, CustomFormElement ...$elements){
```
dktapps: signature needs to be documented (PhpDoc)

dktapps:
This has been stagnating for a few days now. Since I've seen that there's dissatisfaction that this hasn't been merged yet, some points:

- we are still not happy with this API yet, and are still discussing possible changes to it. We prefer not to push one set of API and then break that shiny new API again a few commits later.
- there are still crippling bugs because not everything has been properly tested (such as chained forms)
- I've been busy lately and haven't had chance to keep up with what's going on.

While we could merge this as-is anytime between tags, and then do later changes and fixes later, this would cause confusion and unexpected breakages. Keeping it in the PR makes it easier to review.

I know this is frustrating, and there is a general trend of things taking some time to be done. Please have patience.

inxomnyaa (approved):

dschwartz783:
It's not like people can't shoehorn in forms as they are now, though. You only need to send a simple, very readable JSON. This'll be more convenient, but right now, it's quite easy to implement a GUI anyway. Not worth complaining about too much.

thebigsmileXD:
I got to reproduce the double-sending issue:
The form sends twice when clicking the [x] or the ESC/back key/button and when returning $this instead of null in onSubmit - but it feels like this happens only about 75% of the time, and only if it is `prepend`ed.
I found no way to properly 100% reproduce it.

dktapps:
I know why it's happening, just haven't had time to fix it yet.

dschwartz783:
@thebigsmileXD What exactly are you downvoting? Feel free to explain yourself. Perhaps you believe there is no way to do this without a nice cozy API? Perhaps you think it's worth complaining about? If so, please take your complaints elsewhere. This is not the place for them.

thebigsmileXD:
This is not a forum. Please stop commenting unrelated stuff @dschwartz783

xISRAPILx:
The form sometimes does not work, it does not show.
The function itself is called in PlayerJoinEvent. The message "a" shows in the console, but the form does not.
image

thebigsmileXD:
@xISRAPILx the player is not ready to receive forms at that moment, because he is still logging in. The real issue is: because the server thinks the client never closes that form, you can not use any other form after thatt

jasonwynn10:
Maybe a form timeout could be made of use to avoid that issue where all forms become locked

xISRAPILx:
Are there methods besides the timer for using forms in PlayerJoinEvent? @thebigsmileXD

jarne:
@xISRAPILx with a scheduled task, but that's not the right place for questions here, please ask in the forums next time.

AvgZing (approved):

QuiverlyRivalry:
Are we going to merge this for ALPHA-10 or ALPHA-11?

dktapps:
It will be merged when we're happy with it. Read the PR discussion instead of spamming useless comments.

thebigsmileXD:
This comment is a cleanup-combiner comment to reduce spam with issues

Current issues:
- Double-opening forms when returning $this on onSubmit
- Client can not open any form if an client sided json error occurs/a form is sent before the client is able to render it

dktapps:
I've had unconfirmed reports that the latter issue is fixed in 1.2.6. This will need a fresh round of testing.

dktapps:
I'm debating allowing resending forms provided that they aren't already in use, f.e. you can't resend a form while it's already in another player's queue, but you _can_ resend it after it's been submitted. The point of restricting that was to prevent possible concurrency issues when sending forms to multiple players simultaneously.

QuiverlyRivalry:
I can confirm that in 1.2.5 forms sometimes never sent the form after player has been on server for a few minutes. But in 1.2.6 it doesn’t have that glitch

dktapps:
I just discovered that I can break a form queue by spamming the escape button to get to the pause screen. This isn't good...

inxomnyaa (approved):

inxomnyaa (approved):

dktapps:
I crippled returning forms. Not so fast.

inxomnyaa (approved):
Looks a bit better and fixes them not sending at all 😂

dktapps:
Aside from the bugs, some things we're not happy with yet:

- We don't allow sending the same form instance twice. I don't think this is reasonable, a better restriction would be to only allow sending to one player at a time (makes more sense).
- Leading on from the above point, the disposability of forms leads to the problem that a form cannot be easily copied, i.e. duplicating the form and erasing the data for reuse. This requires, then, that a form object must always be recreated. This is a little inconvenient.
- Plugins wanting to access their owners inside a form currently have to implement their own methods for getting the plugin's PluginBase instance. For example, the following code:
```php
                $player->sendForm(new class($this, $username, $error) extends ModalForm{
			/** @var Main */
			private $plugin;

			public function __construct(Main $plugin, string $username, string $error){
				$this->plugin = $plugin;
				parent::__construct("Account transfer failed!", "Failed to transfer \"$username\" account to yours.\nError: $error\n\nWould you like to try again?");
			}

			public function onSubmit(Player $player) : ?Form{
				if($this->getChoice()){
					$this->plugin->requestAccountDetails($player);
				}else{
					$this->plugin->warnCancelTransfer($player);
				}
				return null;
			}
		});
```
requires that the form being sent must declare a `$plugin` property and assign it in the constructor, in order to access it in `onSubmit()`. This leads to boilerplate code.

thebigsmileXD:
> Plugins wanting to access their owners inside a form currently have to implement their own methods for getting the plugin's PluginBase instance. For example, the following code:

Actually that applies to any parameter. Passing a variable that a player input into a form and displaying it in another form requires constructor rewrites. I suggest an additional parameter `array $arguments = []`

SOF3:
@thebigsmileXD Explicitly requiring passing the Plugin also has the advantage that it's possible to know the scope of the form and manage it with plugin enable/disable stuff.

AryToNeX:
Why don't make some events like `PlayerFormSubmitEvent` and `PlayerFormCloseEvent` to handle form submitting and closing? It would be a lot simple, especially if forms' classes are not abstract so we can make Form objects right out of the box and we can send them easily, without managing them by extending them with our own classes in plugins. For example, this piece of code in a Plugin class:
```
public function onSubmit(PlayerFormSubmitEvent $e){
    // assuming we are handling a Toggle element in index 0 that enables something
    /** @var Toggle $toggle */
    $toggle = $e->getForm()->getElement(0);
    $this->doSomethingToPlayer($e->getPlayer, $toggle->getValue());
    $e->getPlayer()->sendForm(new CustomForm("Test Form", array(
        new Toggle("Test toggle 1", false),
        new Toggle("Test toggle 2", true)
    )));
}
```
will be MUCH simpler to implement for plugin developers as we don't have to pass the owner to our form to do things on `onSubmit()`.
Also, if you think of a way to make forms instantiable out of the box then when we are sending a form to a player it would be a good idea to have a custom internal form name or ID to keep track of what form is being sent to the player; for example this piece of code:
```
public function onSubmit(PlayerFormSubmitEvent $e){
    // first we want to make sure the form we are receiving is a "TestToggleForm1" form
    if($e->getForm()->getInternalName() !== "TestToggleForm1") return;
    // then we manage to do something...

    // assuming we are handling a Toggle element in index 0 that enables something
    /** @var Toggle $toggle */
    $toggle = $e->getForm()->getElement(0);
    $this->doSomethingToPlayer($e->getPlayer, $toggle->getValue());
    // maybe the constructor for a CustomForm object can be
    // "string $internalName, string $title, array $elements"
    $e->getPlayer()->sendForm(new CustomForm("TestToggleForm2", "Test Form 2", array(
        new Toggle("Test toggle 1", false),
        new Toggle("Test toggle 2", true)
    )));
}
```
Obviously this is more straight-forward but it will certainly have problems in more complicated plugins that use forms to do complicated things but I wrote this suggestion in a couple of minutes so it definitely needs some refining here and there but the main idea is there.

SOF3:
@AryToNeX This will certainly make it much harder to pass extra data through the form.

thebigsmileXD:
@AryToNeX I highly say NO THANKS to events. customui was discontinued just because events created plugin-crossover calls, and every plugin reacted to the event.

dktapps:
Events are a reasonable ask (since they are just, well, events) but definitely a bad idea to make them the primary way to handle a form.

```php
public function onSubmit(PlayerFormSubmitEvent $e){
    // first we want to make sure the form we are receiving is a "TestToggleForm1" form
    if($e->getForm()->getInternalName() !== "TestToggleForm1") return;
    ...
```
here's the problem already.

QuiverlyRivalry:
Not meant to be rude, but, can we get this merged by New Year?

WaldoFS:
Not meant to be rude, but this PR make teaspoon plugin error

Sandertv:
Plugins should adapt to the core, not the other way around. Furthermore, this is a development branch.

dktapps:
@WaldoFS that's entirely irrelevant to the discussion. TeaSpoon is its own project.

Thunder33345:
@WaldoFS Not meant to be rude, plugin are suppose to make themself work with the core, not the other way around

WaldoFS:
Ok, thanks all 😘

GooglePlugin:
Has it been done? I tested thebigsmileXD's WarpUI with this branch, but it didn't work properly.
Here are some bugs I found:
 - Forms doesn't appear after sending a form.
 - Form::onSubmit is not called after submitting a form.

robske110:
@GooglePlugin When somebody got time for it. Also, i think the problems you mentioned are probably WarpUI's problems.

kenygamer:
How do you change the ‘Submit’ button to other text?

Thunder33345:
This is not a plugin support forum
please use forums.pmmp.io

dktapps:
> When will it be done?

you must be new round here...

dktapps:
I'm going to close this since there hasn't been any movement on it recently. A new PR will be created when we have something worth shipping.

PR #1485, opened 2017-10-22
Fixed items not removed when put in the lava and refactored Block::get() on FallingSand

matcracker:
## Introduction

When you drop items in the lava they aren't removed.

Refactored Block::get() to BlockFactory::get() in FallingSand
### Relevant issues

* Fixed items not removed when put in the lava


dktapps:
Please confine your changes to separate PRs in future.

As for the lava bug - I can't reproduce it, please give more reproducing information.

PR #1492, opened 2017-10-24
Cancel PlayerInteractEvent automatically on left-click block when spa…

dktapps:
…wn protection is triggered

## Introduction
Currently spawn protection doesn't prevent deletion of fire, and broadcasts block-break animations when the player will not actually be able to break blocks in the zone anyway. This PR automatically sets PlayerInteractEvent to cancelled on left-click if spawn protection is triggered, before calling the event.

### Behavioural changes
- Block-break animations will no longer be seen in spawn protection radius if the player can't break the block.
- Fire can't be deleted anymore inside the spawn protection radius.

## Backwards compatibility
This might affect existing plugins which change things to do with spawn protection using events.

## Tests
TBD

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -2626,7 +2626,7 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 
 				$target = $this->level->getBlock($pos);
 				$ev = new PlayerInteractEvent($this, $this->inventory->getItemInHand(), $target, $packet->face, $target->getId() === 0 ? PlayerInteractEvent::LEFT_CLICK_AIR : PlayerInteractEvent::LEFT_CLICK_BLOCK);
-				if(!$this->level->checkSpawnProtection($this, $target)){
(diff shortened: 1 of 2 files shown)


zKoz210:
Why do need this?

If I have 0 admins, it means that the spawn does not need to be protected?

```count($this->server->getOps()->getAll()) > 0```

https://github.com/pmmp/PocketMine-MP/blob/6daa0135d53c62b0eb771bbee25297a4ff771b83/src/pocketmine/level/Level.php#L1592

mal0ne-23:
@zKoz210 shut up

zKoz210:
@mal0ne-23 You will come to gag to your parents, okay? If you can't give specific criticism in my post, then just keep silent, because your opinion here, nobody cares.

dktapps:
This discussion is irrelevant to the PR anyway...

dktapps:
@zKoz210 in case you missed it, I was referring to your first comment. How spawn protection behaves when there are no operators is not relevant to this pull request.

PR #1498, opened 2017-10-24
Fix EntityEat events.

Muqsit:
## Introduction

This PR fixes EntityEat events not being able to be handled through plugins.

Currently, trying to handle EntityEatBlockEvent/EntityEatItemEvent would give you the following error:
`[Server thread/CRITICAL]: pocketmine\plugin\PluginException: "pocketmine\event\entity\EntityEatItemEvent does not have a valid handler list" (EXCEPTION) in "src/pocketmine/plugin/PluginManager" at line 805`
This is due to non-existance of `Event::$handlerList` property in EntityEatBlockEvent and EntityEatItemEvent. For an unknown reason, that property existed in EntityEatEvent instead.

[The server also never fired EntityEatItemEvent.](https://github.com/pmmp/PocketMine-MP/blob/18777a90411c4d8baf1cf484676cd49350375255/src/pocketmine/item/Food.php#L53-#L63)

### Behavioural changes

* The server now fires EntityEatItemEvent.
* Plugins can now handle EntityEatBlockEvent and EntityEatItemEvent.

## Tests

```php
public function handleEntityEatItem(EntityEatItemEvent $event){
    $event->getResidue();
}

public function handleEntityEatBlockEvent(EntityEatBlockEvent $event){
    $event->getResidue();
}
```

### DIFF
--- a/src/pocketmine/event/entity/EntityEatBlockEvent.php
+++ b/src/pocketmine/event/entity/EntityEatBlockEvent.php
@@ -28,6 +28,8 @@ use pocketmine\entity\Entity;
 use pocketmine\item\FoodSource;
 
 class EntityEatBlockEvent extends EntityEatEvent{
+	public static $handlerList = null;
+
 	public function __construct(Entity $entity, FoodSource $foodSource){
 		if(!($foodSource instanceof Block)){
 			throw new \InvalidArgumentException("Food source must be a block");
@@ -38,7 +40,7 @@ class EntityEatBlockEvent extends EntityEatEvent{
 	/**
 	 * @return Block
 	 */
-	public function getResidue(){
+	public function getResidue() : Block{
 		return parent::getResidue();
 	}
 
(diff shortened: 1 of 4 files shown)


dktapps:
I actually intended to remove these events entirely, they make things awkward for non-food item implementations.

Muqsit:
Awkward?

dktapps:
yes, because they are poorly thought out (as also shown by this bug). Only players gain hunger and saturation from food - mob eating mechanics are entirely different. Additionally, there are items which may be consumed which are not foods.

Muqsit:
yeah, I was thinking of renaming the events into HumanEat* events. I'll close this, it would be better if the new food eating implementation comes in one piece other than this being mostly future-redundant code.

dktapps:
this is why I didn't already fix these bugs myself, because I don't feel it's worth fixing it only to break code depending on it later on down the line.

dktapps:
also, don't forget that PlayerItemConsumeEvent exists... not at all confusing

Muqsit:
wow, that's 4 events corresponding to food changes.

PR #1499, opened 2017-10-25
Implemented Config->removeNested()

dktapps:
## Introduction
We have `getNested()`, `setNested()`, but no `removeNested()`.

Sometimes we want to delete data from a nested key in a config. Currently there is no way to do so (`remove()` doesn't deal with this).

### API changes
- Added `Config->removeNested(string $key)`

## Tests

```php

		$this->getConfig()->setNested("a.b.c", "wtf"); //set a nested "a.b.c"
		$this->getConfig()->save();
		var_dump($this->getConfig());
		var_dump(file_get_contents($this->getDataFolder() . "config.yml"));

		echo PHP_EOL;
		$this->getConfig()->removeNested("a.b.c"); //removes item "c" from the "a.b" nested key
		$this->getConfig()->save();

		var_dump($this->getConfig());
		var_dump(file_get_contents($this->getDataFolder() . "config.yml"));

		$this->getConfig()->removeNested("a.b"); //removes item "b" from the "a" key
		$this->getConfig()->save();

		var_dump($this->getConfig());
		var_dump(file_get_contents($this->getDataFolder() . "config.yml"));
```

## Follow-up
This change is needed for a change I need to make to #1461 .

### DIFF
--- a/src/pocketmine/utils/Config.php
+++ b/src/pocketmine/utils/Config.php
@@ -376,6 +376,27 @@ class Config{
 		return $this->nestedCache[$key] = $base;
 	}
 
+	public function removeNested(string $key) : void{
+		unset($this->nestedCache[$key]);
+
+		$vars = explode(".", $key);
+
+		$currentNode =& $this->config;
+		while(count($vars) > 0){
+			$nodeName = array_shift($vars);
+			if(isset($currentNode[$nodeName])){
+				if(empty($vars)){ //final node
+					unset($currentNode[$nodeName]);
+					break;
+				}elseif(is_array($currentNode[$nodeName])){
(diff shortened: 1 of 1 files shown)


matcracker (approved):

moskadev (approved):

jasonw4331 (approved):

SOF3 (commented):
The key value cache is not updated if the key points to an array and cache exists for its children. However, this problem exists with setNested() too.

dktapps:
Whether or not to remove the cache (unneeded or no?), and/or change the implementation of Config, is not relevant to this PR. That problem can't be solved without dumb bruteforce.

ahmedhozny (approved):

PR #1515, opened 2017-10-30
migrate to new CompoundTag API

dktapps:


src/pocketmine/block/BurningFurnace.php
```diff
@@ -77,10 +76,8 @@ public function onActivate(Item $item, Player $player = null) : bool{
 				$furnace = Tile::createTile(Tile::FURNACE, $this->getLevel(), TileFurnace::createNBT($this));
 			}
 
-			if(isset($furnace->namedtag->Lock) and $furnace->namedtag->Lock instanceof StringTag){
-				if($furnace->namedtag->Lock->getValue() !== $item->getCustomName()){
-					return true;
-				}
+			if($furnace->namedtag->getString("Lock", "", true) !== $item->getCustomName()){
```
dktapps: this will make furnaces un-openable with items that _do_ have custom names

src/pocketmine/item/Item.php
```diff
@@ -481,9 +481,8 @@ public function clearCustomName(){
 	public function getLore() : array{
 		$display = $this->getNamedTagEntry(self::TAG_DISPLAY);
 		if($display instanceof CompoundTag and ($lore = $display->getListTag(self::TAG_DISPLAY_LORE)) !== null){
-			return array_map(function(StringTag $tag) : string{
-				return $tag->getValue();
-			}, $lore->getValue());
+			/** @var string[] $result */
```
dktapps: $result doesn't exist

dschwartz783:
Shouldn't hasTag be hasString, hasInt, etc, similar to getString/getInt, and setString/setInt

dktapps:
too much boilerplate. The getters and setters are there for type safety only. hasTag() only ever needs to return a bool, so there's no need for many of them.

dktapps:
I would like to get this merged as soon as possible. If anyone would like to help testing, please do.

PR #1517, opened 2017-11-01
Cosmetically improve UUID::equals()

sylvrs:
## Introduction
I felt that this pull request was necessary as the function UUID::equals() seems really messy for absolutely no reason.

### Relevant issues
N/A

### API changes
N/A

### Behavioural changes
N/A

## Backwards compatibility
Only a cosmetic fix

## Tests

Test with my commit.

```
$uuid1 = UUID::fromRandom();
$uuid2 = new UUID(...$uuid1->getParts());
$uuid3 = UUID::fromRandom();
var_dump($uuid1->equals($uuid2));
var_dump($uuid1->equals($uuid3));
```

Result

```
bool(true)
bool(false)
```

[Here](http://sandbox.onlinephpfunctions.com/code/9c9e3d3bdf16576a7266e94bfd2b36f74ec422b9) are some live testing results, just in-case you needed more affirmation. Obviously, I used strings, but evaluating it with integers would provide the same result.


dktapps:
please don't use the web editor, OR web file upload...

IrishPacks:
Ah, my bad. Noticed it when I was away from my computer, so I hoped the web editor wouldn't be dumb, but sure enough it was.

PR #1524, opened 2017-11-07
Fix health overflow after calling setMaxHealth()

Muqsit:
## Introduction

While setting an entity's max health, the server does not check whether the entity's health is > than their maxHealth.

### Behavioural changes

`Entity::setMaxHealth($amount)` will check whether the entity's health is greater than the `$amount` and handle the situation appropriately.

## Tests

Before:
```php
/** @var Entity $entity */
$entity->setMaxHealth(40);
$entity->setHealth(40);
$entity->setMaxHealth(20);
var_dump($entity->getHealth());//float(40)
```
After:
```php
/** @var Entity $entity */
$entity->setMaxHealth(40);
$entity->setHealth(40);
$entity->setMaxHealth(20);
var_dump($entity->getHealth());//float(20)
```

### DIFF
--- a/src/pocketmine/entity/Entity.php
+++ b/src/pocketmine/entity/Entity.php
@@ -943,6 +943,9 @@ abstract class Entity extends Location implements Metadatable, EntityIds{
 	 */
 	public function setMaxHealth(int $amount){
 		$this->maxHealth = $amount;
+		if($this->health > $amount){
+			$this->health = $amount;
+		}
 	}
 
 	/**


Muqsit:
@dktapps but you'll need to wait until their health changes for that to be called https://github.com/pmmp/PocketMine-MP/blob/0e64c3dad89e25a0f00bfea9d9a0bd7ff90ec99e/src/pocketmine/entity/Entity.php#L930

Muqsit:
Here's a possible issue without this PR.
There are two players:
```php
/** @var Player $player1 */
/** @var Player $player2 */
```

`$player1`'s health and max health has been set to 100. Then the max health is reverted back to 20.
```php
$player1->setMaxHealth(100);
$player1->setHealth($player1->getMaxHealth());
$player1->setMaxHealth(20);
var_dump($player1->getHealth()); //float(100)
```

Now `$player2` attacks `$player1`.
Here's the problem, the attack does deduct any HP from `$player1` if `EntityDamageEvent::getFinalDamage()` is lesser than 80 (100 - 20, aka `$player1->getHealth() - $player1->getMaxHealth()`).

So what finally happens?
[This](https://github.com/pmmp/PocketMine-MP/blob/aa399a1109d2213e5df84c5390deeda0746c2717/src/pocketmine/entity/Entity.php#L908) is the only thing from `Entity::setHealth()` that will be executed. The damage dealt is totally ignored. The player's health is set to `20` than to `20 - EntityDamageEvent::getFinalDamage()`.

dktapps:
@Muqsit you're forgetting the health attribute on Living entities. This type of overflow won't just occur with Entity base health.

PR #1538, opened 2017-11-12
Prismarine Metadata fix

Infernus101:
## Introduction

Prismarine's metadata interchange, according to https://minecraft.gamepedia.com/Prismarine


dktapps:
Classic rookie mistake. Don't base your assumptions solely off the wiki - it primarily holds data for Minecraft PC edition. Always test in-game first. The current values are correct for Bedrock.

Issue #1543, opened 2017-11-14
Game mode is not saved after changes in PlayerQuitEvent


zKoz210:
### Issue description

Game mode is not saved after changes in PlayerQuitEvent. **Looks like when execute PlayerQuitEvent ANY action with the player not working as setHealth() and so on...**

### Steps to reproduce the issue

1. Write a small plugin
```
    public function onPlayerQuitEvent(PlayerQuitEvent $event)
    {
        $player = $event->getPlayer();
        $change = $player->setGamemode(0);
        $this->getLogger()->notice($change);
    }
```
2. Join to server and set gamemode 3 (/gamemode 3)
3. Re-join
4. Game mode has not changed, although `setGamemode()` returns `true`

### OS and versions

* PocketMine-MP: https://github.com/pmmp/PocketMine-MP/commit/aa399a1109d2213e5df84c5390deeda0746c2717
* PHP: 7.2
* Server OS: CentOS 7
* Game version: PE/Win10

zKoz210:
https://github.com/pmmp/PocketMine-MP/blob/aa399a1109d2213e5df84c5390deeda0746c2717/src/pocketmine/Player.php#L3377 Save need make after PlayerQuitEvent

dktapps:
#1482

Issue #1482, opened 2017-10-21
Inventory items are duplicated when players are killed in PlayerQuitEvent


Bluzzi:
Hello, when I disconnect, then reconnects, here is what happens (watch gif), kill the player when disconnected.

Sorry for my bad english i'm french :/

Gif: http://www.hostingpics.net/viewer.php?id=205605GifDupli.gif

* PocketMine-MP: https://github.com/pmmp/PocketMine-MP/commit/bc7ba3b3c10d48a37e1e8493e91a72e6e41a4af2
* PHP: 7.2
* Server OS: Debian
* Game version: Win10 and Android phone.

Bluzzi:
(Watch the diamond)

matcracker:
The issue can't be reproduced without plugins. This means that you're probably adding one diamond (and one paper) everytime the player join.

Bluzzi:
@matcracker 

No, I do not give any stuff, and I do not give any stuff when a player joins.

Bluzzi:
No, I do not give any stuff, and I do not give any stuff when a player joins.

Bluzzi:
@matcracker No, I do not give any stuff, and I do not give any stuff when a player joins.

dktapps:
then where does the paper come from?

Bluzzi:
@dktapps The player with me has to drop it, try for yourself just kill the player in onQuit, and deco with stuff, reconnecting, the stuff will be doubled.

dktapps:
This is because player data is now saved before PlayerQuitEvent is called.

Bluzzi:
@dktapps Hmm... and how can I fix that?

optimusprime338:
well this started happening after you fixed double chest glitch thingy

dktapps:
@Bluzzi Try cancelling PlayerDataSaveEvent for that player, then save their data manually in PlayerQuitEvent using `Player->save()` AFTER you kill them.

Bluzzi:
@dktapps It does not work :/

`
public function onSavePlayer(PlayerDataSaveEvent $event){
        if(isset($this->dcombat[$event->getPlayer()->getName()])){
            $event->setCancelled(true);
        }
    }

    public function onQuit(PlayerQuitEvent $event){
        $player = $event->getPlayer();
        $name = $player->getName();

        foreach($this->plugin->getServer()->getOnlinePlayers() as $players){
			$players->sendPopup("§c- ".$name." -");
        }

[...]

Issue #284, opened 2017-01-22
Crash when unloading player level during PlayerQuitEvent


legoboy0215:
### Issue description

The getBlah on null in Level.php issue is back again!

### Steps to reproduce the issue

1. Unload a level.

### OS and versions
* PocketMine-MP: This server is running PocketMine-MP 1.6.2dev 「Unleashed」 implementing API version 3.0.0-ALPHA2 for Minecraft: PE v1.0.0.16 (protocol version 100)
* PHP: 7.0.4
* Server OS: Ubuntu 16.10
* Game version: PE/Win10 (delete as appropriate)

### Crashdump, backtrace or other files

http://pastebin.com/rsH9Sb4b

dktapps:
This is unsurprising, happening since 6fbb9b6083e80f835ce042e658797d3326c189d6. This was a patch to prevent Level objects from leaking, but they are used unsafely everywhere. The first issue that came to light was a double chest memory leak.

Did you see anything in your server log about `Position was holding a reference to an unloaded Level` before the crash occurred?

legoboy0215:
No.

legoboy0215:
```
2017-01-22 [03:52:17] [Server thread/CRITICAL]: Unhandled exception executing command 'kick l' in kick: Call to a member function getSpawn() on null
2017-01-22 [03:52:17] [Server thread/CRITICAL]: Error: "Call to a member function getSpawn() on null" (EXCEPTION) in "/src/pocketmine/level/Level" at line 2543
2017-01-22 [03:52:17] [Server thread/CRITICAL]: Error: "Call to a member function getWorldHeight() on null" (EXCEPTION) in "/src/pocketmine/level/Level" at line 1245
2017-01-22 [03:52:17] [Server thread/EMERGENCY]: An unrecoverable error has occurred and the server has crashed. Creating a crash dump
2017-01-22 [03:52:17] [Server thread/EMERGENCY]: Please upload the "/root/pm-1.0/crashdumps/Sun_Jan_22-03.52.17-EST_2017.log" file to the Crash Archive and submit the link 
```
Another one. I did /kick Lego and this happened?!

dktapps:
Wtf... you are in a world which has a null LevelProvider?

dktapps:
That should be impossible. I don't understand.

legoboy0215:
[...]

PR #1544, opened 2017-11-14
Fixed save player data after PlayerQuitEvent

zKoz210:
## Introduction

Fixed saving of data that did not work before, because the data was saved before the call to PlayerQuitEvent
### Relevant issues
Fixed https://github.com/pmmp/PocketMine-MP/issues/1543

## Tests

Completely tested and everything works fine!

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -3378,12 +3378,12 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 						$this->server->broadcastMessage($ev->getQuitMessage());
 					}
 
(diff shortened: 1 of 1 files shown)


dktapps:
this will again break #284 .

zKoz210:
sorry, fixed

dktapps:
Since this issue solves some big issues, I'm going to merge this, but an alternative (safer) method for level unloading needs to be implemented at a later date. This will cause some minor issues with player data saving relating to #284 , but I think that's a less important and far less common issue than the others mentioned (#1543, #1482).

PR #1552, opened 2017-11-17
Some changes to messages

Johnmacrocraft:
## Introduction

This pull request changes some messages.

### Relevant issues

Nothing.

### API changes

Nothing.

### Behavioural changes

Now you'll be able to see message which says bed is too far away.

## Backwards compatibility

No backwards incompatible changes.

## Follow-up

Bed message works fine, but server sends that message even I'm close to bed. You'll probably want to fix that issue.

Requires translation:

| `tile.bed.tooFar` | `Bed is too far away` |

## Tests

Server startup: Start the server.
Bed: Click bed from far distance.

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -1610,7 +1610,7 @@ class Server{
 
 			$this->logger->info($this->getLanguage()->translateString("pocketmine.server.info", [
 				$this->getName(),
-				($version->isDev() ? TextFormat::YELLOW : "") . $version->get(true) . TextFormat::WHITE,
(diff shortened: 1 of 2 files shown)


jasonwynn10:
to add the Bed too far message, create a PR on https://github.com/pmmp/PocketMine-Language

dktapps (dismissed):
These two changes are unrelated and should be in separate commits.

Johnmacrocraft:
@dktapps Ah, ok.

Johnmacrocraft:
@dktapps Separated them as two commits.

PR #1755, opened 2017-11-30
Make Living::addEffect() return bool

Muqsit:
## Introduction

This will let you know if an effect has been successfully applied or not.

### API changes

Living::addEffect() returns whether the effect has successfully been applied or not.

## Tests

```php
/** @var Living $living */
$effect = Effect::getEffect(Effect::SPEED)->setDuration(INT32_MAX)->setAmplifier(1);
var_dump($living->addEffect($effect));//bool(true)

$effect->setAmplifier(0);
var_dump($living->addEffect($effect));//bool(false)
```

### DIFF
--- a/src/pocketmine/entity/Living.php
+++ b/src/pocketmine/entity/Living.php
@@ -231,9 +231,10 @@ abstract class Living extends Entity implements Damageable{
 	 * Adds an effect to the mob.
 	 * If a weaker effect of the same type is already applied, it will be replaced.
 	 * If a weaker or equal-strength effect is already applied but has a shorter duration, it will be replaced.
-	 * Returns whether the effect has been successfully applied.
 	 *
 	 * @param Effect $effect
+	 *
+ 	 * Returns whether the effect has been successfully applied.
 	 * @return bool
 	 */
 	public function addEffect(Effect $effect) : bool{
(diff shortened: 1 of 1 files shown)


src/pocketmine/entity/Living.php
```diff
@@ -231,17 +231,19 @@ public function hasEffect(int $effectId) : bool{
 	 * Adds an effect to the mob.
 	 * If a weaker effect of the same type is already applied, it will be replaced.
 	 * If a weaker or equal-strength effect is already applied but has a shorter duration, it will be replaced.
+	 * Returns whether the effect has been successfully applied.
```
SOF3: This sentence fits better on line 237 behind `@return bool`.

dktapps:
that change would not be within the scope of this pull request.

Muqsit:
with a change in the title and description

dktapps:
that would be combining two unrelated changes into the same PR, which goes against the guidelines.

Muqsit:
you win bruv 👍

src/pocketmine/entity/Living.php
```diff
@@ -233,15 +233,18 @@ public function hasEffect(int $effectId) : bool{
 	 * If a weaker or equal-strength effect is already applied but has a shorter duration, it will be replaced.
 	 *
 	 * @param Effect $effect
+	 *
+ 	 * Returns whether the effect has been successfully applied.
```
dktapps: I think @SOF3 meant something more like `@return bool Whether the effect has been successfully applied`

Muqsit: ah yeah he probably did

PR #1761, opened 2017-12-02
Update start.cmd

dirouc:
Bedrock*


DaPigGuy:
Might as well change the readme too then.

PR #1762, opened 2017-12-02
Update README.md

dirouc:
Bedrock*


dirouc:
@dktapps What's wrong?

Issue #1767, opened 2017-12-03
Cancelling EntityEffectAddEvent sends client effect particles


Muqsit:
### Issue description

When EntityEffectAddEvent gets cancelled and the effect has Effect::$visible set to true, the effect's particles are still visible, though the entity doesn't get affected by the effect (in case of players, the effect icon isn't visible even though the particles are).

- Expected result: What were you expecting to happen?
- Actual result: What actually happened?
Effect particles aren't visible as the event is cancelled.

### Steps to reproduce the issue

1. Cancel EntityEffectAddEvent
```php
public function onEntityEffectAdd(EntityEffectAddEvent $event){
    $event->setCancelled();
}
```
2. Give yourself any effect. `/effect you blindness 100 1`

### OS and versions

* PocketMine-MP: c301788
* PHP: 7.2.0
* Server OS: Ubuntu 14.04 64-bit
* Game version: Win10

### Plugins
- Test on a clean server without plugins: is the issue reproducible without any plugins loaded?
_This issue is related to PocketMine's plugin (event) API._

PR #1768, opened 2017-12-03
Fixed broken logic in EntityEffectEvents

dktapps:
## Introduction
When these effects were cancelled, they were still added/removed as active effects server-side. This caused bugs such as #1767 .

### Relevant issues
#141 introduced the defective code.

### Behavioural changes
cancelling these events now works as expected (in theory)

## Follow-up
There is a problem with the behaviour when EntityEffectRemoveEvent is cancelled due to the effect duration becoming zero, which would result in the duration consequently becoming negative and causing a crash. This is a relevant issue but a separate bug from the one this PR is targeted at.

## Tests
TBD

### DIFF
--- a/src/pocketmine/entity/Effect.php
+++ b/src/pocketmine/entity/Effect.php
@@ -24,8 +24,6 @@ declare(strict_types=1);
 namespace pocketmine\entity;
 
 use pocketmine\event\entity\EntityDamageEvent;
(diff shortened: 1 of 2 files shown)


Muqsit:
Tested: https://streamable.com/xc415

dktapps:
Possible follow-up would be to have `/effect` check the return value of `addEffect()` and display an appropriate error message if failed.

PR #1775, opened 2017-12-04
NBT streams refactor

dktapps:
## Introduction
Much-needed cleanup, especially since the mess that was 0.16.

Before 0.16, there were 2 ways to read and write NBT (big-endian and little-endian). The former is used by Minecraft PC, the latter by Minecraft PE.

As of 0.16, there is a third way (varint little-endian). This format is used by Minecraft PE over network only and has some small changes related to integers and string lengths.

When there were only two formats, it was acceptable to just use a field to determine the endianness. With the introduction of varint NBT (and the poor implementation by myself) this resulted in network parameters everywhere, which is particularly ugly and messy.

This pull request splits the 3 formats into their own NBT stream classes with a defined interface.

### API changes
- The `NBT` class is now abstract.
- Added `BigEndianNBTStream`, `LittleEndianNBTStream`, and `NetworkLittleEndianNBTStream`, descending from `NBT`.

### Behavioural changes
Some bug fixes to how network NBT is handled.

## Tests
This has not been fully tested yet. Help testing would be appreciated.
This might affect the following things:

- World NBT
- Item NBT
- BlockEntity NBT


dktapps:
This is probably missing some preprocessor includes. It could also possibly benefit from additional optimizations.

jarne:
Doesn't this require a new API version?

dktapps:
It's unreleased, development. It's recorded in the pending changelog for the next release.

PR #1777, opened 2017-12-06
add UUID ban/unban

dschwartz783:
UUIDs are a far more reliable method to uniquely
identify players.

## Introduction
This allows server admins to ban a player based on UUID, using either the player's UUID directly, or the player's name. It also lets them unban an UUID.

### API changes
getUUIDBans is now a function on the Server instance.

### Behavioural changes
N/A

## Backwards compatibility
N/A

## Tests
I've tested all the commands, and they work correctly (in English, at least), and players cannot log in when their UUID is banned.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -331,7 +331,7 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 	}
 
 	public function isBanned() : bool{
-		return $this->server->getNameBans()->isBanned($this->iusername);
+		return $this->server->getNameBans()->isBanned($this->iusername) or $this->server->getUUIDBans()->isBanned($this->uuid->toString());
 	}
 
 	public function setBanned(bool $value){
@@ -340,6 +340,16 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 			$this->kick("You have been banned");
 		}else{
 			$this->server->getNameBans()->remove($this->getName());
+		}
+	}
+
+	// Does nothing if the player is not authenticated
+	public function setUUIDBanned(bool $value){
+		if($value === true){
+			$this->server->getUUIDBans()->addBan($this->getUniqueId(), null, null, null);
+			$this->kick("You have been banned");
+		}else{
+			$this->server->getUUIDBans()->remove($this->getUniqueId());
 		}
 	}
 
@@ -1834,7 +1844,7 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 		}
 
 		if(
-			($this->server->getNameBans()->isBanned($this->iusername) or $this->server->getIPBans()->isBanned($this->getAddress())) and
+			($this->server->getNameBans()->isBanned($this->iusername) or $this->server->getUUIDBans()->isBanned($this->uuid->toString()) or $this->server->getIPBans()->isBanned($this->getAddress())) and
 			$this->kick("You are banned", false)
 		){
 			return true;
--- a/src/pocketmine/command/defaults/BanCommand.php
+++ b/src/pocketmine/command/defaults/BanCommand.php
@@ -28,6 +28,7 @@ use pocketmine\command\CommandSender;
 use pocketmine\command\utils\InvalidCommandSyntaxException;
 use pocketmine\event\TranslationContainer;
 use pocketmine\Player;
+use pocketmine\utils\UUID;
 
 class BanCommand extends VanillaCommand{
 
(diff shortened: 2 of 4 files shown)


src/pocketmine/Player.php
```diff
@@ -343,6 +343,15 @@ public function setBanned(bool $value){
 		}
 	}
 
+	public function setXUIDBanned(bool $value){
```
dktapps: If this is used on a player who is not logged into Xbox Live, then no Xbox Live players will be able to join.

dschwartz783: Should I check for $isAuthenticated, then?

dschwartz783: Wouldn't this just ban all non-xbox live users, since they'd be the only ones with an empty XUID? Checking for isAuthenticated should solve that.

dschwartz783:
So, I should just change this to using UUIDs instead? I can do that.

dschwartz783:
Now it should be working with UUIDs instead.

JackNoordhuis:
Why not just store the players username with the ban so server admins don't have to track IP's or UUID's? Would make a lot of peoples lives a lot easier.

dktapps:
This system is particularly user unfriendly and not intuitive. The average server operator won't know that they should use /ban-uuid to get a secure ban. It should just work.

It's justifiable having a separate command for /ban-ip since IPs aren't necessarily tethered to a player. It is not justifiable to create a separate command for UUID bans. PC servers use UUID bans and don't have anything like this. Users should just be able to punch in a player name and the server should do whatever necessary to prevent that player from joining.

dschwartz783:
Well, IP bans are bad since IP addresses are often ephemeral. This gives server admins an additional, more reliable banning option. UUIDs are not as easy to change as an IP address, or name is.

dktapps:
The point of banning an IP isn't always banning a player. It can also be used to mitigate attacks on the server. That's like saying firewalls are bad since you can just get a different IP address.

dschwartz783:
Generally bans are targeted at an individual troublemaker, not a range addresses belonging Chinese hackers pummeling your Minecraft server.

dschwartz783:
Perhaps I should just replace named bans with UUIDs?

dktapps:
"Generally" but not always. How often it's used in that manner is not the point at hand. On the other hand, UUIDs are used _always_ when specifically banning a player.

UUIDs, XUIDs, names, CIDs, are all directly player associated data. An IP is not, since multiple people can use an IP.

My point is that this system doesn't really make sense. It adds an extra option by which a technical user can ban a player more effectively, but it's not friendly to anyone who doesn't know what a UUID is. Let's not forget that the vast majority of players and promo server mods/ops aren't going to be technical people.

dschwartz783:
I suppose I can just roll the logic for banning UUIDs into the normal ban/pardon commands. Fair?

dktapps:
If I can, as a non-technical user, say `/ban ThisHacker`, and `ThisHacker` can't rejoin after changing their username, then I'm happy. The user doesn't need to know what black magic stops them joining the server. That's how it (in my opinion) ought to work.

dschwartz783:
Makes sense. In that case, I'll close the my locale PR, since it would make sense to just send the normal /ban and /pardon-related messages, without any mention of UUID.

dktapps:
This will however make a requirement that banned UUIDs must be able to be associated with the names of player who's used it, so that they can be properly unbanned.

dschwartz783:
@dktapps I can think if countless things that would break, if players were allowed to use the usernames of previously-banned players, but with different UUIDs. Perhaps that's a can of worms we should open at a later date.

dktapps:
Yeah, I'm not really sure how that can be dealt with right here.

dschwartz783:
Okay, now it works with the regular ban/pardon commands. On top of that, banned player names are now mapped to their UUIDs, so that pardon can work properly with them as well.

src/pocketmine/command/defaults/BanCommand.php
```diff
@@ -56,10 +57,25 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
...
+				$mapFileData = yaml_parse_file($mapFilePath);
+			}
+
+			$mapFileData[strtolower($name)] = $player->getUniqueId()->toString();
+			yaml_emit_file($mapFilePath, $mapFileData);
 		}
 
 		Command::broadcastCommandMessage($sender, new TranslationContainer("%commands.ban.success", [$player !== null ? $player->getName() : $name]));
 
+
```
jasonwynn10: random spacing?

dschwartz783: whoops

dschwartz783:
Honestly, using pure UUID makes more sense than relying on both names and UUIDs, except that most plugins are designed poorly, and simply rely on names to uniquely identify players. Removing the reliance on names, would allow players with different UUIDs to masquerade as another player, at least from the perspective of those poorly-written plugins. Perhaps this is an issue that should be offloaded onto plugin devs for "correctness", but this would likely cause far more issues than ALPHA7 did, and we all know how much server admins loved that.

I'm torn, honestly.

dschwartz783:
I no longer have a use for this, and I haven't received any feedback in a week and a half, so I'm just going to close this, and let @dktapps do the work, if he wants. This doesn't seem like a much wanted addition.

PR #1782, opened 2017-12-07
Match Entity::DATA_FLAG_ONFIRE behaviour with vanilla MCPE

### LINKED ISSUE #1778: Fire Resistance effect allows entity to be set on fire
### Issue description

- Expected result: Entity with fire resistance effect will not be set on fire when in contact with fire.
- Actual result: Entity with fire resistance effect is set on fire, though it doesn't hurt itself. This goes against the _vanilla behaviour_ where _fire resistance never sets an entity on fire_.

1. /effect yourself fire_resistance 100 1
2. Swim in lava
3. Get out of lava
4. Notice how you are set on fire (_but are not being hurt_) even though you have fire resistance on you.

### OS and versions

* PocketMine-MP: 51cec52
* PHP: 7.2.0
* Server OS: Ubuntu 14.04 LTS
* Game version: Win10

### Plugins
- Test on a clean server without plugins: is the issue reproducible without any plugins loaded? Yes

### Vanilla Behaviour
https://streamable.com/gu0ls

Muqsit:
## Introduction

This PR fixes non-vanilla behaviours related to `Entity::DATA_FLAG_ONFIRE` and brings changes in certain entity-fire related API functions.

### Relevant issues
* Fixes #1778

### API changes

* `Entity::setOnFire()` now returns `(boolean)` whether `Entity::$fireTicks`'s value was changed. See Tests.

### Behavioural changes

* `Effect::FIRE_RESISTANCE` behaves the same way it does in vanilla:
  * Entities will not be set on fire while having `Effect::FIRE_RESISTANCE` on them.
  * Entities that were put on fire before having `Effect::FIRE_RESISTANCE` on them will need to wait 2 seconds for the effect to extinguish the fire off them. _[**NOTE:** The wait time of 2 seconds is capped to `Entity::$fireTicks`, meaning that only if `Entity::$fireTicks` > 2 seconds, the entity will be extinguished in 2 seconds)_. During these 2 seconds, the entity will NOT deal any fire damage albeit on fire.
* Living entities won't be set on fire when hurt by another entity that is on fire, unless hurt through `EntityDamageByChildEvent` (called by Projectiles).

## Tests

```php
/** @var Entity $entity */
$entity->extinguish();

var_dump($entity->setOnFire(10));//bool(true)
var_dump($entity->setOnFire(8));//bool(false) #Entity::$fireTicks > 8.
```

### DIFF
--- a/src/pocketmine/entity/Effect.php
+++ b/src/pocketmine/entity/Effect.php
@@ -442,8 +442,8 @@ class Effect{
 				}
 				break;
 			case Effect::FIRE_RESISTANCE:
-				if($entity->getFireTicks() > 2){
-					$entity->setFireTicks(2);
+				if($entity->getFireTicks() > 40){
+					$entity->setFireTicks(40);
 				}
 				break;
 		}
--- a/src/pocketmine/entity/Effect.php
+++ b/src/pocketmine/entity/Effect.php
@@ -441,6 +441,9 @@ class Effect{
 					$entity->setAbsorption($new);
 				}
 				break;
+			case Effect::FIRE_RESISTANCE:
+				$entity->extinguish(2);
+				break;
 		}
 	}
 
--- a/src/pocketmine/entity/Living.php
+++ b/src/pocketmine/entity/Living.php
@@ -384,7 +384,7 @@ abstract class Living extends Entity implements Damageable{
 			$e = $source->getDamager();
 			if($source instanceof EntityDamageByChildEntityEvent){
 				$e = $source->getChild();
-				if($e->isOnFire() > 0){
+				if($e !== null and $e->isOnFire() > 0){
 					$this->setOnFire(2 * $this->level->getDifficulty());
 				}
 			}
--- a/src/pocketmine/entity/Living.php
+++ b/src/pocketmine/entity/Living.php
@@ -384,7 +384,7 @@ abstract class Living extends Entity implements Damageable{
 			$e = $source->getDamager();
 			if($source instanceof EntityDamageByChildEntityEvent){
 				$e = $source->getChild();
-				if($e !== null and $e->isOnFire() > 0){
+				if($e !== null and $e->isOnFire()){
 					$this->setOnFire(2 * $this->level->getDifficulty());
 				}
 			}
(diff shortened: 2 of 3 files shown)


src/pocketmine/entity/Living.php
```diff
@@ -384,13 +384,12 @@ public function attack(EntityDamageEvent $source){
 			$e = $source->getDamager();
 			if($source instanceof EntityDamageByChildEntityEvent){
 				$e = $source->getChild();
-			}
-
-			if($e !== null){
```
dktapps: Why has this condition been removed?

Muqsit: 🤦‍♂️ woops, I thought only `EntityDamageByEntityEvent::getDamager()` calls `Server::findEntity()`.

Muqsit:
Wait a min, this gets weirder...

When an arrow goes through lava and hits an entity, the entity is not set on fire though the arrow is on fire.
But when an arrow is fired from a bow enchanted with `Enchantment::FLAME`, the entity hit by the arrow is set on fire.

Video: https://streamable.com/w7q6h

src/pocketmine/entity/Effect.php
```diff
@@ -441,6 +441,9 @@ public function add(Entity $entity, Effect $oldEffect = null){
 					$entity->setAbsorption($new);
 				}
 				break;
+			case Effect::FIRE_RESISTANCE:
+				$entity->extinguish(2);
```
dktapps: This behaviour is obscure. Modifying the fire ticks directly would be better (I'll add methods for this before the PR is merged).

src/pocketmine/entity/Living.php
```diff
@@ -518,6 +517,10 @@ protected function doEffectsTick(int $tickDiff = 1){
 		}
 	}
 
+	public function setOnFire(int $seconds) : bool{
+		return !$this->hasEffect(Effect::FIRE_RESISTANCE) and parent::setOnFire($seconds);
```
dktapps: This should be in `isFireProof()`.

dktapps:
I'm not actually sure which behaviour set we should match. In PC Java Edition, fire resistance allows you to be set on fire, you simply don't take damage (which would make sense, since it's fire _resistance_ and not fire immunity).

Muqsit:
Then either it's a BR-exclusive (like the _enchantment tables emitting light is a BR exclusive feature_) or they  missed this behaviour.
Another thing I've noticed is how "Zombies on fire" don't set you on fire in BR on any difficulty. The only way you could be set on fire through entities in BR is from Flame bows.

dktapps:
I'm going to close this since I think that the PC behaviour makes more sense. This feels like an overlooked bug to me.

The "burning mobs attacking" issue is not directly relevant to this PR.

PR #1792, opened 2017-12-11
Automatically handle all subclass events

SOF3:
## TL;DR: The changes
* All events are assumed to be listenable (i.e. you can put the event as the function parameter type in a Listener class directly) by default. They no longer need to declare `public static $handlerList = null;`.
* ~~Events that are not listenable should declare a `$noHandle` tag in their class doc comment.~~
  * Abstract event classes cannot be listened to directly unless they declare `@noHandle` in their class doc comment.
* If a Listener function declares a parameter of type `FooEvent`, any event that is `instanceof FooEvent` will be handled by the listener.
  * Vice versa, when an event of type `BarEvent` is called, all listeners with the parameter type of any superclasses of `BarEvent` will handle it.
* In addition, added the `EventPriority::ALL` constant.

## Introduction
There are many issues in the events system, and this pull request (PR) is the first in the series.

This PR targets the issue regarding class inheritance. For example, new developers are often confused about handling EntityDamageEvent, EntityDamageByEntityEvent, etc. This PR makes things clear &ndash; as long as the handler function names an Event subclass and the event can be passed into the handler function without type errors (`$event instanceof FirstParameterType`), the handler should be triggered. If a listener has multiple handler functions where the event can be called upon, all these handler functions will be triggered.

In the following text, three classes namely `FooEvent`, `BarEvent` and `QuxEvent` will be used to elaborate the behaviour, declared as follows:

```php
class FooEvent extends \pocketmine\event\PluginEvent{}
/**
 * @noHandle
 */
class BarEvent extends FooEvent{}
class QuxEvent extends BarEvent{}
```

As in this PR, whether an event class is abstract not concerned by PocketMine as the event calling context is responsible for instantiating it.

Both `FooEvent` and `QuxEvent` can be used as the event parameter, like this:

```
class MyListener implements \pocketmine\event\Listener{
    public function onBarEvent(BarEvent $event){ ... }
    public function onQuxEvent(QuxEvent $event){ ... }
}
```

On the other hand, if `BarEvent` is used as the parameter type, this exception will trigger upon calling registerEvents(), because `@noHandle` forbades handlers to use it as the parameter type:

```
pocketmine\plugin\PluginException: "BarEvent cannot be handled because it declares the @noHandle tag" (EXCEPTION) in "src/pocketmine/plugin/PluginManager" at line 822
```

Among the events declared by PocketMine, the following classes contain the `@noHandle` tag:

```
Event.
block/BlockEvent.
entity/EntityEvent.
inventory/InventoryEvent.
level/ChunkEvent.
level/LevelEvent.
player/PlayerEvent.
plugin/PluginEvent.
server/ServerEvent.
```

Handling them directly is forbidden mainly to prevent abusive wildcard handling of events, such as the belief "I want to stop the player from doing anything, so I handle PlayerEvent and always cancel it if it implements Cancellable".

## Performance concerns
All doc-comments are parsed upon registering event handlers. Considering event handlers should only be registered upon server start (or otherwise, very rarely), this has no impact to runtime performance.

Instead of storing handler lists in the static class property, handler lists are indexed by their class name in the static field `HandlerList::$classMap`.
To prevent using reflections to read up the inheritance of the event class every event call, each entry in $classMap holds an array of reference to each superclass HandlerList instance (if available, i.e. if no `@noHandle` is declared).
These two changes shall have minimal impact on performance <sup>[citation needed: benchmark needed to prove this statement]</sup>.

## Backwards compatibility
Let's use the `FooEvent`, `QuxEvent` and `MyListener` in the Introduction again. Before, when `QuxEvnt` is called, only `onQuxEvent` is triggered. Now, both `onFooEvent` and `onQuxEvent` will be handled. **Developers must pay attention to this potential backward incompatibility because this problem does not raise any warnings or errors.** Event handlers shall be registered as normal, but the handler code may be called twice.

In addition, plugins hacking `XxxEvent::$handlerList` may start crashing as the handler list is no longer filled. However, I have not observed any plugins trying to do so in the past 3 years.

Furthermore, several methods in the `HandlerList` class have been removed, including `HandlerList::bake()` and `HandlerList::getRegisteredListeners()`. This is because the methods are no longer necessary and may result in confusing code, inconsistent with the new behaviour in this PR. A parameter is also required for `HandlerList` constructor.

Also, the new `EventPriority::ALL` constant may lead to plugins using `(new ReflectionClass(EventPriority::class))->getConstants()` failing.

## Follow-up
This PR is **incomplete**. No effort has been made to handle plugin reloads correctly yet &ndash; handler lists may not be unregistered properly.

## Tests
- [x] Basic tests have been carried out to ensure this works under "normal" circumstances. See the foo-bar-quz example above.
- [x] Event handler execution order is now in this sequence: `EventPriority [LOWEST->MONITOR], class inheritance [subclass -> superclass], time of registering listener [first -> last]`
- [x] ~~Handlers for superclasses of `@noHandle` classes (i.e. `FooEvent`) can still handle the subclasses (i.e. `BarEvent` and `QuxEvent`) correctly, even when an event of a `@noHandle` class (not subclass) is called, as of 7b0d332.~~
  - [x] Update: abstract classes are automatically disallowed for being handled directly, unless they declare `@allowHandle`
- [x] Appears to behave correctly with reloaded plugins, asserting source code has not been changed.

Tests can be found here: https://gist.github.com/SOF3/bde9b7acfbc67098b28e21a4f404022a


src/pocketmine/plugin/PluginManager.php
```diff
@@ -792,24 +787,20 @@ public function registerEvents(Listener $listener, Plugin $plugin){
...
-			throw new PluginException($event . " handlerList property is not public");
+		$tags = self::parseDocComment((string) (new \ReflectionClass($event))->getDocComment());
+		if(isset($tags["deprecated"]) and $this->server->getProperty("settings.deprecated-verbose", true)){
+			$this->server->getLogger()->warning($this->server->getLanguage()->translateString("pocketmine.plugin.deprecateEvent", [
+				$plugin->getName(),
+				$event,
+				get_class($listener) . "->" . ($executor instanceof MethodEventExecutor ? $executor->getMethod() : "<unknown>")
+			]));
 		}
+		if(isset($tags[""]))
```
dktapps: ...

dktapps:
For readers' sake, a TL;DR:

This allows you to specify a handler of `EntityDamageEvent` which will handle any instanceof it (as per current behaviour), AND also now allows you to specify a handler specifically for `EntityDamageByEntityEvent`. 

This was not previously possible due to limitations in the way handlers were registered.

SOF3:
There was also one thing that @dktapps and I were arguing: Should events explicitly declare they cannot be handled (`@noHandle` as now), or should they explicitly declare `@allowHandle` (only for abstract classes; non-abstract classes must always be allowed to be handled)?

P.S. why does everyone 🎉 @dktapp's comment but not my PR 🙊

jojoe77777:
> P.S. why does everyone 🎉 @dktapp's comment but not my PR 🙊

@SOF3 No-one likes walls of text with no TL;DR

SOF3:
The mentioned issues in the PR have been fixed. See https://gist.github.com/SOF3/bde9b7acfbc67098b28e21a4f404022a for latest test results. It is now ready for review for merging to master.

dktapps:
Is there a defined order in which handlers will execute? Take this code for example:
```php
	public function onAttack(EntityDamageEvent $event){
		var_dump("handler1");
	}

	public function onAttack2(EntityDamageByEntityEvent $event){
		var_dump("handler2");
	}
```
The `EntityDamageByEntityEvent` handler is called before the `EntityDamageEvent` handler. Is this expected behaviour? If it is, it needs to be documented.

dktapps:
Bump. I'd like to get this merged in the not-too-distant future.

Have we considered events implementing interfaces? I remember it might have been discussed, but I don't remember for sure.

src/pocketmine/event/HandlerList.php
```diff
@@ -65,43 +64,77 @@ public static function unregisterAll($object = null){
...
+		$super = $class;
+		$parentList = null;
+		while($parentList === null && ($super = $super->getParentClass()) !== false){
+			// skip $noHandle events in the inheritance tree to go to the nearest ancestor
+			// while loop to allow skipping $noHandle events in the inheritance tree
+			$parentList = self::getHandlerListFor($super->getName());
+		}
+		$lists = $parentList !== null ? self::$classMap[$parentList->class] : [];
+
+		$list = $noHandle ? null : new HandlerList($event) ;
```
dktapps: nit: formatting issue

src/pocketmine/event/HandlerList.php
```diff
@@ -43,19 +41,20 @@ class HandlerList{
...
 
-	public static function bakeAll(){
-		foreach(self::$allLists as $h){
-			$h->bake();
-		}
-	}
+	/**
+	 * @var HandlerList[][] a 2D array of HandlerList.
+	 *                      The first dimension index is the class name, containing an array of HandlerList for that class and its superclasses.
+	 *                      The second dimension index is an integer >= 0, starting from the HandlerList for that class.
```
dktapps: I don't really understand why a multidimensional array is necessary here. This could be implemented with HandlerLists referencing the HandlerList of its parent.

Since we see further down that this list can be dynamically updated, I wonder why we don't just recursively iterate through handler list parents and call all appropriate handlers instead of using this thing. This would simplify the process of registering handler lists for new events and remove the need for this 2D array.

This 2D array only seems to be used for the purpose of accessing all the handler lists for an event linearly, which doesn't really make sense to me when it can be done recursively with less confusing code.

dktapps:
This handler:
```php
	public function onEntityDamage(EntityDamageEvent $event){
		var_dump(get_class($event));
		//var_dump(1);
	}
```
does not receive `EntityDamageEvent` subclass events such as `EntityDamageByEntityEvent`. This appears to be down to the way handler lists are only initialized for events with handlers listening to that specific class. They are then fetched without initialization for calling. This means that if no plugins directly listen to `EntityDamageByEntityEvent`, then no plugins will receive it at all, despite being a subclass of `EntityDamageEvent`.

dktapps:
> Event handler execution order is now in this sequence: EventPriority [LOWEST->MONITOR], class inheritance [subclass -> superclass], time of registering listener [first -> last]

This appears not to be accurate. From the code I see that we are executing all handlers of X priority traversing through the hierarchy, then moving to the next priority and repeat process. This appears to be better behaviour, but the PR notes it incorrectly.

dktapps (approved):

PR #1794, opened 2017-12-11
Removed Cancellable interface

dktapps:
## Introduction
This interface does not make any sense. The implementations of the methods it defines are already in Event, which gives a magical impression to it. Those methods subsequently check if `$this instanceof Cancellable`, which is poor design.
Additionally, it is easy to accidentally forget to implement this interface since it is not required or suggested by an IDE when creating a new event.

This commit removes the interface and replaces its functionality with an Event->isCancellable() method. This method returns FALSE by default (as per the original behaviour) and must be overridden to return TRUE by cancellable events.

### API changes
- `Cancellable` interface has been removed.
- `Event` now declares an `isCancellable()` method which replaces this functionality.

## Backwards compatibility
This is a BC-break for plugins with custom events implementing this interface, and for plugins using `instanceof Cancellable`.

## Tests
Currently entirely untested. TBD.

## Requests for comments on:
- Concept (removing the Cancellable interface)
- Whether the `isCancelled()` method should:
   - return TRUE by default (since around 70% of core events are cancellable)
   - be abstract to make sure it doesn't get forgotten (would require it to be declared by every event, cancellable or not)


dschwartz783:
I'm not sure it makes sense to check for whether an event is cancellable. An event handling function can only ever execute a single event. Since it's designed to execute that individual event, the logic contained in the function is going to operate on either the assumption that it's cancellable, or that it's not. There shouldn't be a need to check.

dschwartz783:
Wouldn't it be simplest (and more correct) to just have cancellable events inherit from a CancellableEvent abstract class, which contains all the necessary methods (and also inherits from Event), and just remove those methods from Event? That way, developers won't be left wondering if the event can be cancelled. The methods won't be there.

Would also let you remove the exception throwing and whatnot. Much more streamlined.

JackNoordhuis:
A trait would be more appropriate. Having separate abstract classes for server event, entity event, player event, etc that extend the cancelable class is pretty confusing and unneeded.

dschwartz783:
Yes, I suppose a trait does sound more like what we're looking for here. As long as those methods are not available on events which cannot be cancelled. Anything else doesn't make sense.

dktapps:
You can't typehint against a trait. There are things in the core that need these methods as well.

dschwartz783:
@dktapps You don't need to type hint to determine if something's cancellable. The nature of the event itself determines that. Event functions generally assume that an event will either always be cancellable, or never be cancellable, depending on the event. A trait to add the necessary functionality to classes that need it sounds appropriate.

Although, you didn't clarify why you might want to type hint against a trait, so I'm somewhat guessing here. I'm sure you'll tell me if I guessed wrong, one way or another. ;)

Sandertv:
You're missing the point. You can't check if something utilizes a trait or not without hacky code. How would you check if an event is cancellable if you don't want to use reflection?

dschwartz783:
@Sandertv You'll know that it's not cancellable if it doesn't have the setCancelled and related methods available. There is no reason to check if an event is cancellable, because it either always is, or always isn't (unless there's some exception I'm not aware of).

dktapps:
> Although, you didn't clarify why you might want to type hint against an event.

> There are things in the core that need these methods as well.

dschwartz783:
@dktapps Then wrap it in an interface, and your problem's solved.

JackNoordhuis:
The whole point of this PR is to remove the need to implement an interface though.

dschwartz783:
@dktapps It's recommended to pair traits with interfaces anyway. Create an interface which requires conformance to the methods that the trait implements.

I'm aware that this is supposed to remove the interface, but perhaps that decision was premature.

JackNoordhuis:
The `$this instanceof Cancellable` can be removed by using a trait but developers would still have to implement the interface. Although this would prevent some bc changes it is somewhat counter productive as the whole goal of this PR is to remove the interface.

I guess it really comes down to developer education, using the provided API methods (`Event::isCancellable()` in this case) is safer than using a php instanceof check directly.

dschwartz783:
@JackNoordhuis The goal is flawed, as I stated earlier, which means this PR should either be closed, or refactored. My approach, and indeed the approach that you yourself suggested above, would make the code much safer. Developers will not be able to cancel events which cannot be cancelled, because the methods will not be there to use. In no universe is that less safe than requiring isCancellable checks all over the place.

dktapps:
Some food for thought (the motivators behind this PR):

1. subclasses of cancellable cannot become uncancellable with the interface
2. trait + interface does not solve the problem that it's easy to forget to implement it
3. 70% of core events are cancellable. Every cancellable event has to implement this interface. There's no way to make cancellable=true the default with the interface. This is one of the primary RFC points.

dschwartz783:
@dktapps
I’ll provide you with solutions to each of your problems:
1. Then keep your isCancellable method, as a part of the trait/interface. This will allow developers to know that it might be cancellable, but that they still must check to be certain. If it doesn’t conform to cancellable, they don’t have to check, which is convenient.
2. Anything in software development is forgettable. It’s up to the dev to know that that’s how it works, like everything in any programming language.
3. As I said, keep isCancellable in the trait/interface, and you’ll be able to make cancellable=true for every event which uses them. Devs will know to check whether or not it’s actually cancellable (might want to add something in the docs too for this), and if it can’t ever be cancelled (like player death), the methods won’t be there.

dktapps:
> subclasses of cancellable cannot become uncancellable with the interface

If your approach is taken, then the methods will be present for subclass events of cancellables which are not cancellable. This will be a confusing inconsistency.

> 70% of core events are cancellable. Every cancellable event has to implement this interface. There's no way to make cancellable=true the default with the interface. This is one of the primary RFC points.

Your proposal still leaves this problem. If it's trait-ified, still every cancellable event must include the trait.

In short, I don't think your solution solves anything.

dschwartz783:
...and with this proposal, every non-cancellable event must include the isCancellable method, which seems worse to me.

I’ve already mentioned what my proposal solves numerous times, but I’ll mention it again: having methods which are conditionally useless seems like sloppy coding to me. There has to be a way you’ll accept which allows the removal of those methods, at least part of the time.

Thunder33345:
I think having to check if method exist to determine if it's cancelable is even worse

dktapps:
It appears there are several deleted messages here. For historical purposes and for the sake of the discussion not being confusing, let me correct that:

image
image
image
image
image
image

Please do not delete messages in technical discussions. It simply makes them difficult to follow.

SOF3 (approved):

src/pocketmine/plugin/RegisteredListener.php
```diff
@@ -91,7 +90,7 @@ public function getPriority() : int{
 	 * @param Event $event
 	 */
 	public function callEvent(Event $event){
-		if($event instanceof Cancellable and $event->isCancelled() and $this->isIgnoringCancelled()){
+		if($event->isCancellable() and $event->isCancelled() and $this->isIgnoringCancelled()){
```
SOF3: This call looks quite redundant. Why not always return false for non-cancellable `Event::isCancelled()` calls, rather than throwing exceptions?

dktapps:
Due to controversy in the RFC and lack of resolution, I'm going to close this. Re-reading this now I'm more aligned with @dschwartz783 in the mindset that the cancellation handling methods should be moved to a trait instead, although this will break all events implementing `Cancellable`, core and plugin both - but that's a topic for a different RFC.

PR #1796, opened 2017-12-11
Consumables refactor

dktapps:


dktapps:
This PR contains changes that are not directly within the scope of the branch. These should be moved/removed prior to merge.

PR #1798, opened 2017-12-12
Added the EXECUTE priority.

SOF3:
## Description
Under this priority, an event is listened to for cancelling it and overriding its execution.

This priority is useful for plugins that do not change how the incident occurs, i.e. semantically the incident represented by the event still occurs, but it is technically cancelled to prevent the event from being executed by the calling context.

## Example: DiscordMirror + HologramChat
Suppose we have these two plugins:
* DiscordMirror: it mirrors all chat messages to a Discord channel
* HologramChat: it shows chat messages as holograms (FloatingText) rather than using the traditional Player->sendMessage().

The DiscordMirror plugin should listen to PlayerChatEvent at the MONITOR priority. It should ignore cancelled events, e.g. anti-spam plugins and plugins that override chat for other input (e.g. auth plugins may use it for receiving passwords).

What about the HologramChat plugin? It should override default behaviour by cancelling the event and spawning the FloatingText in its handler. But if it cancels the event at LOWEST to HIGHEST, it will stop DiscordMirror from working. At the MONITOR event, since the order is undefined, it may or may not affect DiscordMirror.

This PR allows HologramChat to handle the event at the EXECUTE priority, which will not affect plugins that only concern the final semantic outcome of the event (DiscordChat).

## Backward compatibility
This PR should preserve 100% backward-compatibility, because the number of event priorities has not been guaranteed to remain unchanged.

### DIFF
--- a/src/pocketmine/event/EventPriority.php
+++ b/src/pocketmine/event/EventPriority.php
@@ -28,38 +28,49 @@ namespace pocketmine\event;
  * List of event priorities
  *
  * Events will be called in this order:
- * LOWEST -> LOW -> NORMAL -> HIGH -> HIGHEST -> MONITOR
+ * LOWEST -> LOW -> NORMAL -> HIGH -> HIGHEST -> MONITOR -> EXECUTE
  *
  * MONITOR events should not change the event outcome or contents
  */
 abstract class EventPriority{
 	/**
 	 * Event call is of very low importance and should be ran first, to allow
-	 * other plugins to further customise the outcome
+	 * changes from other plugins to override the outcome
 	 */
-	public const LOWEST = 5;
+	public const LOWEST = 6;
 	/**
 	 * Event call is of low importance
 	 */
-	public const LOW = 4;
+	public const LOW = 5;
 	/**
 	 * Event call is neither important or unimportant, and may be ran normally
 	 */
-	public const NORMAL = 3;
+	public const NORMAL = 4;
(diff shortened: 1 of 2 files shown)


dktapps:
I'm a bit dubious about this, but it's obvious from the diff that this preserves backwards compatibility. Are you sure that this cannot be achieved any other way?

_SOS: I don't understand what you're trying to do here_

dktapps:
As I understand this, the tl;dr is this: when you cancel an event at this priority, you are doing so in order to change the way the event actually happens - you are _not_ actually _cancelling_ the event.

My preferred approach would be to allow installing callback handlers and allow plugins to set these in order to override the event's result.
```php
$event->setExecuteHandler(function(PlayerChatEvent $event){
    $event->getPlayer()->getLevel()->addFloatingText(...);
}
```
however, this would restrict events to single-outcome only.

There exists the potential for unexpected behaviour if multiple handlers handle at execute priority, but I think that is an issue which exists with any priority.

SOF3:
Actually the method as suggested by @dktapps can still allow multiple executions by calling the original dxecutor in the new executor. But this is pointless -- if you aren't gonna change the existing behavior (rather than adding), just use the MONITOR priority.

dktapps:
I'm closing this and #1961 since neither of them properly solve the problem. Both are hacks which would make for confusing API. Additionally, these PRs are both stale.

PR #1799, opened 2017-12-12
PluginManager: Altered regex to allow @ignoreCancelled without a value

dktapps:
## Introduction
This allows specification of `@ignoreCancelled` on event handlers without a `true` or `false` value. If a value is not specified, `true` is used.

### API changes
see intro

## Tests
```php
	/**
	 * @param BlockBreakEvent $event
	 * @ignoreCancelled
         */
	public function onBlockBreak(BlockBreakEvent $event){
		var_dump($event->getPlayer()->lastBreak);
		$event->getPlayer()->sendMessage("Time to break: " . ceil((microtime(true) - $event->getPlayer()->lastBreak) * 20) . " ticks");
		$event->getPlayer()->sendMessage("Expected: " . ceil($event->getBlock()->getBreakTime($event->getItem()) * 20) . " ticks");
	}
```

### DIFF
--- a/src/pocketmine/plugin/PluginManager.php
+++ b/src/pocketmine/plugin/PluginManager.php
@@ -760,7 +760,7 @@ class PluginManager{
 					if(!isset($matches[1])){
 						$ignoreCancelled = true;
 					}else{
-						$matches[1] = strtolower($matches[1]);
+						$matches[1] = strtolower(trim($matches[1]));
 						if($matches[1] === "false"){
 							$ignoreCancelled = false;
 						}elseif($matches[1] === "true"){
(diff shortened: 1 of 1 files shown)


dktapps:
OOOpS, this breaks normal specification.

SOF3:
I don't know why @shoghicp made it so complicated. It could have been `(bool) preg_match("/^[\t ]*\\* @ignoreCancelled[\t ]{1,}true/m")` in the first place; then this PR can also be as simple as `(bool) preg_match("/^[\t ]*\\* @ignoreCancelled([\t ]{1,}true)?/m")`.

SOF3 (changes requested):
The code is made unnecessarily confusing. `@ignoreCancelled false` could be directly ignored.

dktapps:
This was supposed to be a simple change. I separated it from another PR in the hopes of getting it in separately, but since it's more trouble than it's worth, I won't bother.

PR #1825, opened 2017-12-16
Enchantment: Split up enchantment type data and enchantment instance data

dktapps:


zKoz210:
1) Need to add a function that would return the maximum level of enchantment. It would be very convenient
2) You can use a negative level of enchantment. (/enchant knockback zKoz210 -299)
3) /enchant knockback zKoz210 wefwfwef - works, though needs to be check

dktapps:
@zKoz210 that is not relevant to this PR.

zKoz210:
@dktapps Oh, you have changed name pull request, sorry

PR #1826, opened 2017-12-17
change word

nnnlog:
## Introduction

change word

### DIFF
--- a/src/pocketmine/nbt/tag/StringTag.php
+++ b/src/pocketmine/nbt/tag/StringTag.php
@@ -63,7 +63,7 @@ class StringTag extends NamedTag{
 	 *
 	 * @throws \TypeError
 	 */
-	public function setValue(string $value) : void{
+	public function setValue($value) : void{
 		if(!is_string($value)){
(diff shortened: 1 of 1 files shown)


jasonwynn10:
`The short data type is a 16-bit signed two's complement **integer**.` The wording should not be changed.

robske110:
Actually ShortTag in the message is also wrong. He should change ShortTag->StringTag and int->string. Because this is in the StringTag class.

jasonwynn10:
Right... I didn’t notice which file was being changed. I assumed it was ShortTag because of the message 🤔

dktapps:
Copy pasta is so nice... would anyone like to PR and fix this properly?

nnnlog:
@dktapps No **Copy Paste**

SuperMaXAleX:
drink water

dktapps:
https://media.giphy.com/media/jXD7kFLwudbBC/giphy.gif

nnnlog:
@dktapps Live like that.

PR #1830, opened 2017-12-18
Added a getEntry() to BanList

zKoz210:
## Introduction
Adds direct access to BanEntry

### API changes
Added `BanList::getEntry(string $name): ?BanEntry`

## Backward compatibility
This PR should preserve 100% backward-compatibility

### DIFF
--- a/src/pocketmine/permission/BanList.php
+++ b/src/pocketmine/permission/BanList.php
@@ -59,6 +59,23 @@ class BanList{
 	}
 
 	/**
+	 * @param string $name
+	 *
+	 * @return BanEntry|null
+	 */
+	public function getEntry(string $name): ?BanEntry{
+		$name = strtolower($name);
+		
+		$this->removeExpired();
+
+		if (!isset($this->list[$name])){
+			return null;
+		}
+
+		return $this->list[$name];
+	}
+
+	/**
 	 * @return BanEntry[]
 	 */
 	public function getEntries() : array{
--- a/src/pocketmine/permission/BanList.php
+++ b/src/pocketmine/permission/BanList.php
@@ -64,11 +64,9 @@ class BanList{
 	 * @return BanEntry|null
 	 */
 	public function getEntry(string $name) : ?BanEntry{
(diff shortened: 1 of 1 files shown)


src/pocketmine/permission/BanList.php
```diff
@@ -58,6 +58,23 @@ public function setEnabled(bool $flag){
 		$this->enabled = $flag;
 	}
 
+	/**
+	 * @param string $name
+	 *
+	 * @return BanEntry|null
+	 */
+	public function getEntry(string $name): ?BanEntry{
```
dktapps: formatting issue

src/pocketmine/permission/BanList.php
```diff
@@ -58,6 +58,23 @@ public function setEnabled(bool $flag){
...
+		$name = strtolower($name);
+		
+		$this->removeExpired();
+
+		if (!isset($this->list[$name])){
+			return null;
+		}
+
+		return $this->list[$name];
```
dktapps: the `isset()` and this statement could be merged into a single statement using a null coalesce (`$this->list[$name] ?? null`)

dktapps (changes requested):
Looks like a reasonable addition, but please see comments on the diff.

src/pocketmine/permission/BanList.php
```diff
@@ -58,6 +58,19 @@ public function setEnabled(bool $flag){
...
+	 * @param string $name
+	 *
+	 * @return BanEntry|null
+	 */
+	public function getEntry(string $name) : ?BanEntry{
+		$name = strtolower($name);
+		
+		$this->removeExpired();
+
+		return $this->list[$name] ?? null;
```
dktapps: the `strtolower()` can now be moved to here, and remove the first statement, since it's only used once.

SOF3 (approved):

dktapps (approved):

dktapps:
I'd better check the EOF newline before I merge this - not really clear why it still shows up on the diff.

PR #1833, opened 2017-12-18
Fixed wrong execution PlayerPreLoginEvent close #1831

### LINKED ISSUE #1831: Wrong execution PlayerPreLoginEvent
### Issue description
The banlist and whitelist should be checking only after PlayerPreLoginEvent. Otherwise it couldn’t execute my plugin example

### OS and versions

* PocketMine-MP: https://github.com/pmmp/PocketMine-MP/commit/0ee78d24167da734161c6c163ff915819bbbf963

### Example plugin
```
	public function onPlayerPreLoginEvent(PlayerPreLoginEvent $event)
	{
		if ($player->isBanned() && $player->hasPermission("ban.exempt")) {
			$player->setBanned(false);
		}
	}
```

zKoz210:
## Introduction
See https://github.com/pmmp/PocketMine-MP/issues/1831

### API changes
`PlayerPreLoginEvent` will be called before check `BanList` and `WhiteList`

## Backward compatibility
This PR should preserve 100% backward-compatibility

## Tests
```
	public function onPlayerPreLoginEvent(PlayerPreLoginEvent $event)
	{
		if ($player->isBanned() && $player->hasPermission("ban.exempt")) {
			$player->setBanned(false);
		}
	}
```

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -1827,7 +1827,7 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 		}
 
 		$this->setSkin($skin);
-		
+
 		$this->server->getPluginManager()->callEvent($ev = new PlayerPreLoginEvent($this, "Plugin reason"));
 		if($ev->isCancelled()){
 			$this->close("", $ev->getKickMessage());
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -1827,6 +1827,13 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 		}
 
 		$this->setSkin($skin);
+		
+		$this->server->getPluginManager()->callEvent($ev = new PlayerPreLoginEvent($this, "Plugin reason"));
+		if($ev->isCancelled()){
+			$this->close("", $ev->getKickMessage());
+
+			return true;
+		}
 
 		if(!$this->server->isWhitelisted($this->iusername) and $this->kick("Server is white-listed", false)){
 			return true;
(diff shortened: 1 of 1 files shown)


SalmonDE:
Behavioural changes:
In plugins like PurePerms it could be useful to check if the player is banned from the server and not listed on the whitelist before registering them.

zKoz210:
any review?

dktapps:
This can also be achieved by cancelling `PlayerKickEvent`, although there isn't currently a reliable way to determine the kick reason (comparing messages isn't good enough).

I wouldn't say this behaviour is "wrong".

zKoz210:
@dktapps How is it you cannot say? 

Previously it was implemented right

Ex. plugins:
1) EssentialsPE: https://github.com/LegendOfMCPE/EssentialsPE/blob/master/src/EssentialsPE/EventHandlers/PlayerEvents.php#L26-L37
2) CustomAlerts: https://github.com/EvolSoft/CustomAlerts/blob/master/CustomAlerts/src/CustomAlerts/EventListener.php#L76-L110 **(!)**

Similar work is also observed in Bukkit, when we can prevent a kick player appropriate way. You broke so many plugins, unfortunately, and deprived me of the only adequate solution to my problem.

src/pocketmine/Player.php
```diff
@@ -1827,6 +1827,13 @@ public function handleLogin(LoginPacket $packet) : bool{
 		}
 
 		$this->setSkin($skin);
+
```
dktapps: Please strip trailing whitespace.

zKoz210:
Fixed

PR #1837, opened 2017-12-20
Implemented TextFormat::colorize()

SOF3:
## Introduction

See 4c46087ffccc7f26835a1fc5bbf2ac60397884f9 for the initial discussion

### API changes

Added TextFormat::colorize().

### Behavioural changes

nil

## Backwards compatibility

No backward incompatibilities noted.

## Follow-up

nil

## Tests

```php
<?php
include "vendor/autoload.php";
echo TextFormat::colorize("&a&b&c&d&e&f&g&h&i&j&&o& r&");
```

```
§a§b§c§d§e§f&g&h&i&j&§o& r&
```

### DIFF
--- a/src/pocketmine/utils/TextFormat.php
+++ b/src/pocketmine/utils/TextFormat.php
@@ -81,6 +81,18 @@ abstract class TextFormat{
 	}
 
 	/**
+	 * Replaces placeholders of § with the correct character. Only valid codes (as in the constants of the TextFormat class) will be converted.
+	 *
+	 * @param string $string
+	 * @param string $placeholder default "&"
+	 *
+	 * @return string
+	 */
+	public static function colorize(string $string, string $placeholder = "&") : string{
+		return preg_replace('/' . preg_quote($placeholder) . '([0-9a-fk-or])/u', TextFormat::ESCAPE . '$1', $string);
+	}
+
+	/**
 	 * Returns an JSON-formatted string with colors/markup
 	 *
 	 * @param string|array $string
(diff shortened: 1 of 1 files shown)


Thunder33345:
should a colourize for &colourname ex &green to be escape<greencode> considered while we are at this topic?
using if defined(TextFormat::class."::".$colourCode) and constant(TextFormat::class."::".$colourCode)

dktapps:
too much work. This functionality matches (as far as I know) the Bukkit API and nothing more.

Yexeed:
what means to much work?

dktapps:
it means it's not worth the effort.

SOF3:
@Thunder33345 In that case, an HTML-like approach e.g. `&green;` etc. (with a trailing semicolon) would be more appropriate. But in that case, you are inventing a new language.

src/pocketmine/utils/TextFormat.php
```diff
@@ -80,6 +80,18 @@ public static function clean(string $string, bool $removeFormat = true) : string
...
+	/**
+	 * Replaces placeholders of § with the correct character. Only valid codes (as in the constants of the TextFormat class) will be converted.
+	 *
+	 * @param string $string
+	 * @param string $placeholder default "&"
+	 *
+	 * @return string
+	 */
+	public static function colorize(string $string, string $placeholder = "&") : string{
+		return preg_replace('/' . preg_quote($placeholder) . '([0-9a-fk-or])/u', TextFormat::ESCAPE . '$1', $string);
```
dktapps: at least this regex is better than https://github.com/pmmp/PocketMine-MP/pull/1837/files#diff-b1b8c79c93533368678816cecafcbe98R65

dktapps:
This will crash if `/` is used as a placeholder. Is this considered a problem?

dktapps (approved):

PR #1839, opened 2017-12-21
Implement armor damage protection enchantments

dktapps:
## Introduction
This is a rough port of my original work from the `api3/blocks` branch, with some modernisations and better design, and minus the JSON stuff. It is almost completely untested.

It is mostly complete except for some of the finer nuances such as fire protection's behaviour.

## Changes
TBD

## Tests
TBD


dktapps:
The following things are missing from this PR:

- Blast protection reduces knockback from explosions by 15L%.
- Fire protection reduces burn time by 15L% when set on fire.

dktapps:
This would benefit from changes made in #1957 .

PR #1843, opened 2017-12-23
Fixed Block/Item Problem

hmy2001:
## Introduction

Fixed that item should be Item but processed as Block

### API changes

nil

### Behavioural changes

nil

## Backwards compatibility

No backward incompatibilities noted.

## Follow-up

nil

## Tests

I don't have do to test but i think that it is corret...


hmy2001:
Why?

dktapps:
because it's wrong, and you obviously didn't test it.

hmy2001:
i have any question.
how t make phar?

PR #1846, opened 2017-12-24
Implement MinecraftSound

Encritary:
## Introduction

There is PlaySoundPacket, but no user-friendly class for it. So here it is!
### API changes
Implemented new MinecraftSound class, that plays sound by it's ID.
## Tests

It was tested a lot of times on my mini-games and it works perfectly.


dktapps:
This has issues that cannot be resolved because you've deleted the origin repository.

PR #1864, opened 2018-01-01
LevelProviders refactor

dktapps:
## Introduction
Level providers are currently extremely confusing and overcomplicated. This pull request delivers a major refactor to level providers, which drastically simplifies their implementation, makes them independent of Levels, and lots of other cool bonuses.

This is a follow-up to the changes made resulting in the 3.0.0-ALPHA4 API bump (disentanglement of chunks from their level providers to allow them to exist independently).

### API changes
- `LevelProvider`s are no longer dependent on Levels. This is a significant change because it makes it possible to construct `LevelProvider`s on threads without a `Level`, which opens up **future possibilities** to do many cool things:
    - Super-simple world format conversion
    - Asynchronous chunk I/O
    **(DISCLAIMER: These things are NOT implemented yet, but are examples of things that COULD be done.)**
- The `LevelProvider` interface has undergone significant changes. Unused or redundant methods have been stripped out, and the remaining methods have significant changes:
    - `loadChunk()` now directly returns a `Chunk` object read from disk.
    - `saveChunk()` now directly accepts a `Chunk` object instead of coordinates.
    - `__construct()` now **only** accepts a `string $path`, the `Level` parameter has been removed.
    - The following methods have been obsoleted, and therefore removed:
        - `getChunk()`
        - `setChunk()`
        - `saveChunks()`
        - `unloadChunk()`
        - `unloadChunks()`
        - `isChunkLoaded()`
        - `getLoadedChunks()`
        - `isChunkGenerated()`
        - `isChunkPopulated()`
        - `getLevel()`
- `getServer()` method has been removed from `BaseLevelProvider`.

### Behavioural changes
- `LevelProvider`s no longer keep records of loaded chunks. They are now simplified down to an I/O interface. This was changed because `Level`s already keep records of loaded chunks, so the records in the providers were entirely unnecessary, except for excessive overcomplications.
- `LevelProvider`s are no longer circularly dependent on their `Level`s.

Functionally, everything should work the same as it did before to the end user.

## Backwards compatibility
This poses backwards compatibility breaks for reasons described above.

## Follow-up
This will be followed by further changes, including:
- Split level providers into chunk I/O and level-data components
- Making chunk I/O asynchronous.

## Tests
This has been lightly tested with all world formats. However, everything should work fine.

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -2406,6 +2406,10 @@ class Level implements ChunkManager, Metadatable{
 		if($chunk === null){
 			return;
 		}
+
+		$chunk->setX($chunkX);
+		$chunk->setZ($chunkZ);
+
 		$chunkHash = Level::chunkHash($chunkX, $chunkZ);
 		$oldChunk = $this->getChunk($chunkX, $chunkZ, false);
 		if($unload and $oldChunk !== null){
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -2780,7 +2780,7 @@ class Level implements ChunkManager, Metadatable{
 				}
 			}
 
-			$chunk->unload();
+			$chunk->onUnload();
 		}catch(\Throwable $e){
 			$logger = $this->server->getLogger();
 			$logger->error($this->server->getLanguage()->translateString("pocketmine.level.chunkUnloadError", [$e->getMessage()]));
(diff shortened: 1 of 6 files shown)


dktapps:
Travis is being an ass for some reason.

zKoz210:
getServer() has been removed in BaseLevelProvider

dktapps:
thanks @zKoz210 , missed that.

PR #1871, opened 2018-01-02
Fixed Player::sleepOn

Frago9876543210:
## Introduction
```$pos``` is Vector3. Which means that if you send Vector3 containing a float, it will result in an error.
## Tests
- [x] Tested

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -1178,7 +1178,7 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 
 		$this->sleeping = clone $pos;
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by dktapps** [comment]
> [x] Tested

tells me nothing whatsoever about _how_ you tested it.

You might have prevented an error, but you haven't fixed the bug. Additionally this will cause unexpected behaviour in negative coordinates.

**Comment by Frago9876543210** [comment]
> tells me nothing whatsoever about how you tested it.

```php
$player->sleepOn($player);
```

Frago9876543210:
> Additionally this will cause unexpected behaviour in negative coordinates.

However, this hotfix was used repeatedly in the Player.php

dktapps:
This is a hack, and it doesn't fix the bug properly. Other code in Player is not a reason to make more code that also doesn't behave as expected.

Frago9876543210:
Can be used [intval](http://php.net/manual/en/function.intval.php) to solve this problem.

PR #1876, opened 2018-01-03
Refactored entity metadata handling into its own class, with type-saf…

dktapps:
…e methods

This includes several other changes, including:
- SLOT data properties now accept items directly
- POS data properties now accept floored Vector3s (in future this will be block positions) or null for 0,0,0
- VECTOR3F data properties now accept Vector3s or null for 0,0,0

cc @thebigsmileXD who requested some changes which are in this PR - would anyone like to help testing?

### DIFF
--- a/src/pocketmine/entity/DataPropertyManager.php
+++ b/src/pocketmine/entity/DataPropertyManager.php
@@ -1,0 +1,303 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * @author PocketMine Team
+ * @link http://www.pocketmine.net/
+ *
+ *
+*/
+
+declare(strict_types=1);
+
+namespace pocketmine\entity;
+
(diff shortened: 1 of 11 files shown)


thebigsmileXD:
@dktapps I'd definitely prefer a similar way to MiNET instead.
Adding a new manager for it is okay, but having to know the type as an developer is a bad choice.

Dev's should be able, just as in MiNET, to set the value and only get an error if the value type mismatches. 

This can accomplished by registering properties like items/blocks:
```
$manager->registerProperty(id, type, default value);
```

so setting them instead checks if the value type is the property type, and assert/throw an error instead.

dktapps:
MiNET's system is entirely different because the metadata dictionary is built from class properties on the fly when sending to players. That kind of change is out of the scope of this PR.

dktapps:
Additionally MiNET's system does not provide a way to detect changes in metadata, meaning that the dictionary has to be relayed manually. Look at this for example: https://github.com/NiclasOlofsson/MiNET/blob/master/src/MiNET/MiNET/Entities/World/PrimedTnt.cs#L68-L92

PocketMine's system uses a dirty properties method to allow relaying them in bulk on tick, to cut down on packet spam. This relies on being able to know which properties changed.

thebigsmileXD:
Requested/Discussed changes were just skipped?

src/pocketmine/entity/DataPropertyManager.php
```diff
@@ -0,0 +1,303 @@
...
+		return $this->properties;
+	}
+
+	/**
+	 * Returns properties that have changed and need to be broadcasted.
+	 *
+	 * @return array
+	 */
+	public function getDirty() : array{
+		return $this->dirtyProperties;
```
thebigsmileXD: Still sounds like a hack lol

src/pocketmine/network/mcpe/NetworkBinaryStream.php
```diff
@@ -153,23 +153,17 @@ public function getEntityMetadata(bool $types = true) : array{
...
-					$item = $this->getSlot();
-					$value[0] = $item->getId();
-					$value[1] = $item->getCount();
-					$value[2] = $item->getDamage();
+					$value = $this->getSlot();
 					break;
 				case Entity::DATA_TYPE_POS:
-					$value = [0, 0, 0];
-					$this->getSignedBlockPosition(...$value);
+					$value = new Vector3();
```
thebigsmileXD: Ah, thats neat.

PR #1880, opened 2018-01-04
Block bounding boxes

### LINKED ISSUE #169: Block bounding boxes and collision boxes
### Issue description
According to PocketMine-MP, there are a large number of blocks which do not have any bounding box. This includes blocks such as tallgrass, water, signs, and many more.

This is plain wrong.

_All_ blocks have bounding boxes. The only exception to that rule is air, which is a null block. What those blocks don't have is _collision_ boxes, i.e. entities can walk through them.

This issue came to my eye while working on a basic pressure plate implementation. Pressure plates would have to be implemented by making them nonsolid but with an entity collision, like how water works. The issue with this is that because the server thinks these blocks have no bounding box, the intersection of an entity bounding box with that of, for example, a pressure plate, cannot be detected.

The best that can currently be done is to fire Block->onEntityCollide() when the entity enters the block's allocated full block area. That is _wrong_. It should be done when the entity's bounding box intersects with the pressure plate's, otherwise you would be able to trigger pressure plates by hovering above them.

### OS and versions

* PocketMine-MP: https://github.com/pmmp/PocketMine-MP/commit/5ce6c6227ff6ec512d89bf08cb163b3433ccb1ed
* PHP: 7.0.13
* Server OS: irrelevant
* Game version: PE/Win10

### LINKED ISSUE #1892: Lava sets fire to the player if he stands next
### Issue description
Lava sets fire to the player if he stands next

1. Place block of lava
2. Go to a distance of one block

### OS and versions
* PocketMine-MP: https://github.com/pmmp/PocketMine-MP/commit/4d2549b50ab245649b40367c4070294251de6419
* Game version: PE/Win10

Video: [1234.zip](https://github.com/pmmp/PocketMine-MP/files/1610042/1234.zip)

dktapps:
## Introduction
For a long time now there has been a mistaken idea that some blocks have a "null" bounding box or no bounding box. This is inaccurate, as described in #169 .

This PR changes the declaration of this so that a block always has a 1x1x1 bounding box unless explicitly overridden. It also declares the `canPassThrough` property on more blocks.

Now, declaring the `canPassThrough()` property on a block will cause it to not have any _collision_ boxes, which means entities can pass through them without the need for null bounding box hacks.
This allows blocks such as liquid to have a checkable bounding box for intersections, which (finally!!!) allows a solution to #1892 , and also allows properly implementing things like pressure plates.

## Relevant issues
- fixes #169 
- fixes #1892


src/pocketmine/entity/Entity.php
```diff
@@ -1634,7 +1634,7 @@ protected function checkGroundState(float $movX, float $movY, float $movZ, float
 	 */
 	public function getBlocksAround() : array{
 		if($this->blocksAround === null){
-			$bb = $this->boundingBox->grow(0.01, 0.01, 0.01);
+			$bb = clone $this->boundingBox;
```
Muqsit: why clone when it's only being read? 🤔

dktapps: it was a quickfix because I didn't have time to go through and change all the `bb` to `this->boundingBox`. Obviously I wasn't thinking very clearly.

dktapps:
The changes to entity collision in this PR do however break magma blocks, but I thought dysfunctional magma was better than lava setting you on fire when you're not in it.

src/pocketmine/entity/Entity.php
dktapps: Additionally this should actually be shrink() because floating-point errors can result in you "intersecting" with a block that you're only next to. As I recall the grow() was a change for magma blocks to work, but it causes lots of undesired behaviour.

dktapps:
As it turns out, this does not allow correct pressure plate impl because pressure plates use a special bounding box for checking activation, which doesn't occupy the same area as its visual bounding box (it's taller and the sides are shorter). Since this removes the capability to do things like this, it can't be merged in its current form.

PR #1885, opened 2018-01-05
Re-implemented block ray tracing

dktapps:
## Introduction
100% accurate block raytracing is needed for projectiles to work correctly. This PR implements a ray-tracing algorithm as described by http://www.cse.yorku.ca/~amana/research/grid.pdf

A previous block ray tracing method (`BlockIterator`) already existed and was used for `Living->getLineOfSight()`, however `BlockIterator` is excessively confusing, far too complicated and dependent on a `Level` to function.

### API changes
- Removed `BlockIterator` class
- Added `pocketmine\math\VoxelRayTrace` which contains two public generator functions:
  - `inDirection()` allows ray tracing from a start point in a direction, up to a maximum distance
  - `betweenPoints()` allows ray tracing between an arbitrary start and end point.

## Follow-up
Use this code for projectile path ray traces (instead of relying on AABB collisions).


dktapps:
It's come to my attention that `BlockIterator` performs a similar function to this algorithm. However, `BlockIterator` is very and unnecessarily overcomplicated.

The current methods here provide ray tracing for one specific use case (block bounding box intercepts). However, this can be made more useful by either adding some kind of callback mechanism which accepts a current Block (or even just coordinates), or a class abstraction with a method that can be overridden to provide custom handling for blocks.

dktapps:
For example, this would be insufficient to provide ray tracing functionality needed for ender pearls, since they behave differently than any other projectile with regards to collisions.

PR #1891, opened 2018-01-07
Fixed INT32_MIN

Enes5519:
## Tests
I tested on 32-bit :D

### DIFF
--- a/src/pocketmine/PocketMine.php
+++ b/src/pocketmine/PocketMine.php
@@ -22,7 +22,7 @@
 declare(strict_types=1);
 
 namespace {
-	const INT32_MIN = -0x80000000;
+	const INT32_MIN = -0x7fffffff;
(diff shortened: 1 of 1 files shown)


Enes5519:
@dktapps When I test 32 bit, it returns as float not int.

dktapps:
32-bit is not supported anymore...

Enes5519:
@dktapps I know. I just wanted to prove it.

[image](https://hizliresim.com/lOYPyJ)

dktapps:
What you wanted to prove is irrelevant.

> 32-bit is not supported anymore...

We are fully aware that 32-bit does odd things like this. That is one of the reasons I cut support for 32-bit.

dktapps:
https://bugs.php.net/bug.php?id=53934

PR #1895, opened 2018-01-08
Level: Implemented asynchronous chunk I/O

dktapps:
# IMPORTANT!
**Back up your worlds before testing this**. I take no responsibility for any disasters that might happen from using this code without taking appropriate precautions beforehand.

## TL;DR
Synchronous chunk loading and saving is a major performance problem. This PR moves chunk loading and saving onto another thread, which offers major performance bonuses.

## Introduction
For a very long time now, a common and critical problem with PocketMine-MP's performance in real applications has been chunk I/O.

Up until now, chunk I/O has been done synchronously, on the main thread. This has a range of detrimental effects on performance, including lag spikes on chunk load and save (particularly on machines using HDDs). This I/O does not need to be done synchronously.

This pull request splits LevelProviders into two halves. The chunk I/O part is moved onto a separate thread to allow disk I/O to be done asychronously without wasting main thread time.

### Benefits
Major performance improvements due to getting time-consuming disk I/O off the main thread, leaving the CPU free to do more useful things than waiting for disk. This means:
- No more huge lag spikes due to autosave
- No more lag spikes on chunk load. This will be particularly visible on machines with HDDs.
- The server is no longer disk-bound due to level I/O, which opens up the possibility of raising the chunks-per-tick limit throttle, since blocking the main thread with disk I/O is no longer a concern. Current tests show that 16 chunks per tick is easily viable. This leads to better player experience, because chunks will load faster.

### Caveats
There is still some performance penalty to loading and saving chunks, due to the way PHP pthreads works. Non-thread-safe objects must be serialized in order to copy them between threads. This adds some overhead to chunk copying.

Tests are being done to optimize this by using the [igbinary](https://github.com/igbinary/igbinary) extension, which improves deserialization speeds by a factor of 5 (deserialization is the main problem for the main thread, because more chunk loading happens than saving).

### API changes
- **By necessity, chunks can no longer be synchronously autoloaded on the fly from the main thread.** This is an important change, which means that chunk autoloading will no longer occur if something (such as a plugin) tries to acccess a chunk which is not loaded - instead, a RuntimeException will be thrown.
- The LevelProvider interface has changed to allow _requesting_ chunk load/save asychronously, instead of demanding synchronous loading.

### Behavioural changes
- Potentially huge performance improvements due to moving time-consuming disk I/O off the main thread.
- One extra thread is now created per level. Chunk disk I/O now happens on this thread instead of on the main thread, along with decompression and necessary conversions to in-memory format. This offers enormous performance advantages particularly for McRegion worlds and Anvil without the ChunkUtils extension.

## Backwards compatibility
This changes lots of parts of the LevelProvider API, as well as the chunk-loading behaviour of various Level methods, including get/setBlockIdAt() etc.

## Tests
Testing is currently in progress, bugs are being found. Some things are still broken.


Yexeed:
"Back up up" ?

dktapps:
Oops, that's a typo. Thanks.

zKoz210:
1) If you load the world loadLevel(), and then teleport to it, then we get the error.
2) Random teleportation Ex: /tp 99999 8 99999 

Is this normal?
```
 RuntimeException: "Chunk 8 6 is not loaded" (EXCEPTION) in "src/pocketmine/level/Level" at line 2436
```

dktapps:
It's not "normal" obviously, but I'm aware of the bug. Similar issues also exist with lighting updates.

robske110:
Is there any remote possibility to implement a method which still does synchronous chunk loading, to provide an easier way to upgrade old plugins. Because adapting some plugins to async chunk loading can be the same as a complete rewrite.

dktapps:
Not really. The whole point of this was to eliminate that. Making it able to operate in both modes would be a considerable pain in the ass. Even without threads it's designed to be async.

Plugins won't be affected provided that they operate within the bounds of the loaded world. Any plugins that relied on sync autoloading can use ChunkLoaders and such instead. Agreed, it'll require some code changes, but that is life. ¯\_(ツ)_/¯

robske110:
Agreed. I also noticed that actually far less plugins will be affected in a severe way than I initially anticipated. Also a full-rewrite won't be really required to adapt a plugin which breaks because of this, one could hack it quickly together.

Driesboy:
the server keep hanging while stopping the server by closing sessions while alot of players are online and alot of chunks are loaded

dktapps:
That's expected behaviour, unless it hangs indefinitely.

Driesboy:
it hangs indefinitely

dktapps:
Some things that need fixing:

- [ ] Chunks getting ticked without adjacent chunks being loaded (grass bug)
- [ ] Population will populate chunks without waiting for chunks to be loaded (will cause generation bugs)
- [ ] Light updates will crash when attempting to propagate into unloaded chunks (hmm)

daniktheboss:
Sorry for bump, but how stable is this now?

dktapps:
no more so than last time I commented.

dktapps:
This has built up a large array of merge conflicts and hasn't had any recent activity because I haven't had time to work on it.

As a proof of concept, this is fine, but has a LOT of bugs which I am still finding today. This is therefore in no fit state to be merged and will have to be revised in the future.

For the above reasons, I'm going to close this pull request.

PR #1901, opened 2018-01-11
Anvil : Add getVariantBitmask

Enes5519:
## Introduction
Added getVariantBitmask for Anvil

## Tests
No tests


Muqsit:
spaces 👀

dktapps:
"No tests" is unacceptable.

The reason I didn't do this myself is because of differences in the anvil item vs block metadata which a simple bitmask does not account for. There is a bug in the original code, and this is not the correct solution.

PR #1906, opened 2018-01-11
FIX #1903

### LINKED ISSUE #1903: Block cache is no longer counted or cleared correctly
### Issue description
As of the merge of #1823 the Level block cache is now a two-dimensional array. This means that `Level->clearCache()` is now counting a number of _arrays_ of blocks, as opposed to a number of actual blocks.

https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/level/Level.php#L934-L938

dschwartz783:
### Relevant issues

Fixes #1903

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -932,7 +932,7 @@ class Level implements ChunkManager, Metadatable{
 			$this->chunkCache = [];
 			$this->blockCache = [];
 		}else{
-			if(count($this->blockCache) > 2048){
+			if(array_sum(array_map("count", $this->blockCache)) > 2048){
 				$this->blockCache = [];
 			}
 		}


src/pocketmine/level/Level.php
```diff
@@ -932,7 +932,7 @@ public function clearCache(bool $force = false){
 			$this->chunkCache = [];
 			$this->blockCache = [];
 		}else{
-			if(count($this->blockCache) > 2048){
+			if(array_sum(array_map(function ($blockList) { return count($blockList); }, $this->blockCache)) > 2048){
```
dktapps: Is a lambda necessary here? `"count"` as a callable ought to suffice.

dschwartz783: ok, done

dschwartz783: now actually done. :p

dktapps:
While this is an elegant solution, it's very sub-optimal for performance. `array_*` functions tend to be quite slow.

On a testing server, I find that this is about 2x faster in the majority of cases:
```php
			$count = 0;
			foreach($this->blockCache as $list){
				$count += count($list);
				if($count > 2048){
					$this->blockCache = [];
					break;
				}
			}
```
Likely something to do with not allocating a temporary array for `array_sum()` to add up.

Of course, this is all on the order of microseconds, so any savings are very small.

dschwartz783:
PHP is quite the drag... Leave it for code readability, or replace it with yet another PHP performance workaround?

When we're done with that, can we all agree that Shoghi needs to die for choosing PHP in the first place?

dktapps:
To clarify to readers, effectively what's happening in this PR under the hood at runtime is this (when you PHPify the `array_*` functions):
```php
            $counts = [];
            foreach($this->blockCache as $list){
                $counts[] = count($list);
            }

            $count = 0;
            foreach($counts as $c){
                $count += $c;
            }
            
            if($count > 2048){
                $this->blockCache = [];
            }
```

dschwartz783:
I find it sad that PHP developers are so used to making these sacrifices, that they don't even see what's wrong with it anymore. In many languages, map functions and array summations perform well. PHP truly brings out the worst in developers.

dktapps:
This discussion could go on all day if we had the patience. I'm going to lock this before it gets out of hand.

PR #1912, opened 2018-01-13
Enchantment: Implement Silk Touch

dktapps:
The changeset in this PR is fairly simple and self-explanatory. This has been PR'd to make sure there are no special cases with blocks being unaffected by silk touch that I missed.

### DIFF
--- a/src/pocketmine/block/Bed.php
+++ b/src/pocketmine/block/Bed.php
@@ -207,6 +207,10 @@ class Bed extends Transparent{
 		return [];
 	}
 
+	public function isAffectedBySilkTouch() : bool{
+		return false;
+	}
+
 	public function getAffectedBlocks() : array{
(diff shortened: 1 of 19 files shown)


dktapps:
Silk touch is a problem for any block that drops items with NBT from a tile. 
Additionally, some blocks suck as monster eggs do not drop themselves when Silk Touch is used (although this can already be done by declaring `false` for affected by silk touch, and then checking manually in `getDropsForCompatibleTool()`).

TL;DR: there needs to be a better way to handle getting the drops when silk touch is used.

Also, there needs to be a better generic way to prevent default actions happening when ST is used (ice -> water (#1914), monster eggs don't spawn silverfish, etc).

dktapps:
The tiles issue can be resolved with future API updates which will allow blocks to be identified as tile-holders. In PC, tile-holding blocks are not silk-touch affected by default.

PR #1919, opened 2018-01-15
Update README.md

logig:
Let  developers find a link to download API documents.


dktapps:
any particular reason why people would want to download and not just view it...?

robske110:
Probably because you can't view it currently. (CSP)

PR #1929, opened 2018-01-17
Implemented critical hits

Encritary:
## Introduction

Implemented critical hits. It works when player is: not flying, not sprinting, falling, not having blindness effect, not inside of water. It deals +50% of original damage(ex. diamond sword deals 7 damage and with critical it will deal 10.5 damage).

### API changes

It adds new modifier: `EntityDamageEvent::MODIFIER_CRITICAL`.

## Tests

Test's results in screenshots:
Damage
Critical hit


Encritary:
Uhm, wait, what's wrong with Travis CI build? o.o

Encritary:
@dktapps can you please restart Travis CI build?

Muqsit:
But why +50% the original damage? Why not 30 or 40?

dktapps:
@Muqsit https://minecraft.gamepedia.com/Damage#Critical_Hits

dktapps:
Travis is misbehaving again, it's not an issue specific to the PR.

You can run a Travis tests locally if you have git bash by running `./tests/travis.sh` from the repository root.

Muqsit:
@dktapps the wiki says `150% of Item::getAttackPoints() + pots`

`While falling, melee attacks deal a critical hit (150% of the weapon's damage including potion effects, but before enchantments are applied, rounded down to the nearest hit point).`

dktapps:
so 150% of the damage incoming from the attacker (of which the handling is currently very scattered and messy). EvE really does need a lot of cleaning up.

src/pocketmine/Player.php
```diff
@@ -2413,6 +2413,14 @@ public function handleInventoryTransaction(InventoryTransactionPacket $packet) :
 							$ev->setCancelled();
 						}
 
+						if(!$this->isSprinting() and !$this->isFlying() and $this->fallDistance > 0 and !$this->hasEffect(Effect::BLINDNESS) and !$this->isInsideOfWater()){
+							$ev->setDamage($ev->getDamage() / 2, EntityDamageEvent::MODIFIER_CRITICAL);
+							$pk = new AnimatePacket();
+							$pk->action = AnimatePacket::ACTION_CRITICAL_HIT;
+							$pk->entityRuntimeId = $target->getId();
+							$this->server->broadcastPacket($target->getViewers(), $pk);
```
dktapps: This shouldn't be here. If a plugin changes this, then critical particles should not be broadcasted.

dktapps:
@Encritary stop spamming.

Encritary:
Travis CI is ill today ._.

src/pocketmine/Player.php
```diff
@@ -2413,6 +2413,10 @@ public function handleInventoryTransaction(InventoryTransactionPacket $packet) :
 							$ev->setCancelled();
 						}
 
+						if(!$this->isSprinting() and !$this->isFlying() and $this->fallDistance > 0 and !$this->hasEffect(Effect::BLINDNESS) and !$this->isInsideOfWater()){
+							$ev->setDamage($ev->getDamage() / 2, EntityDamageEvent::MODIFIER_CRITICAL);
```
zKoz210: `$ev->getDamage() / 2` - 150%?

Muqsit: yeah, but instead of setting the base damage to 150% of the base damage directly, the extra 50% is handled by a different modifier.

zKoz210: @Muqsit right, I forgot about this, thanks.

src/pocketmine/Player.php
```diff
@@ -2431,6 +2431,9 @@ public function handleInventoryTransaction(InventoryTransactionPacket $packet) :
 							$pk->action = AnimatePacket::ACTION_CRITICAL_HIT;
 							$pk->entityRuntimeId = $target->getId();
 							$this->server->broadcastPacket($target->getViewers(), $pk);
+							if($target instanceof Player){
```
jojoe77777: Or you could broadcast the packet to `$this->getViewers()` and `$this`

Encritary: @jojoe77777 no difference. In most times viewers of both attacker and target are same.

dktapps: Do all players see critical hit particles? I was under the impression that only the attacker saw them.

src/pocketmine/Player.php
dktapps: This should be `getFinalDamage()`.

dktapps:
Looks reasonable. Please rebase the branch on top of current master.

Encritary:
@dktapps done

PR #1930, opened 2018-01-17
Add API methods: Entity::isInvisible(), Entity::setInvisible()

Muqsit:
## Introduction

Many plugins make use of `Entity::setDataFlag(Entity::DATA_FLAGS, Entity::DATA_FLAG_INVISIBLE, bool)` function. This commit adds a short-hand for the functions, just like some other API functions (`Entity::isSprinting()`, `Entity::setSprinting()`, `Entity::isImmobile()`, `Entity::setImmobile()` etc...).

### API changes

Added API methods
`Entity::isInvisible()`: Returns whether entity is invisible.
`Entity::setInvisible()`: Toggle entity's visibility.

## Tests

A simple invisible toggler:
```
/** @var Entity $player */
$isInvis = $player->isInvisible();
$player->setInvisible(!$isInvis);
$player->sendMessage("You are now " . ($player->isInvisible() ? "invisible" : "visible") . ".");
```
video-1-17-2018-12-38-23-pm

### DIFF
--- a/src/pocketmine/entity/Entity.php
+++ b/src/pocketmine/entity/Entity.php
@@ -655,13 +655,13 @@ abstract class Entity extends Location implements Metadatable, EntityIds{
 		$this->setGenericFlag(self::DATA_FLAG_IMMOBILE, $value);
 	}
 
-        public function isInvisible() : bool{
-                return $this->getGenericFlag(self::DATA_FLAG_INVISIBLE);
-        }
+	public function isInvisible() : bool{
+		return $this->getGenericFlag(self::DATA_FLAG_INVISIBLE);
+	}
 
-        public function setInvisible(bool $value = true){
-                $this->setGenericFlag(self::DATA_FLAG_INVISIBLE, $value);
-        }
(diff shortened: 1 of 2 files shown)


src/pocketmine/entity/Entity.php
```diff
@@ -655,6 +655,14 @@ public function setImmobile(bool $value = true){
 		$this->setGenericFlag(self::DATA_FLAG_IMMOBILE, $value);
 	}
 
+        public function isInvisible() : bool{
+                return $this->getGenericFlag(self::DATA_FLAG_INVISIBLE);
+        }
+
+        public function setInvisible(bool $value = true){
```
dktapps: can have a void typehint here

Muqsit: oh totally forgot about tabs, do I strict type other like-functions too?

dktapps: No, keep it to related changes. Others can be changed separately.

PR #1934, opened 2018-01-18
Implemented Elytra

Encritary:
## Introduction
This pull request implements Elytra. It has 431 max durability and it gets 1 damage per second while gliding. Source: https://minecraft.gamepedia.com/Elytra

## Tests
To test:
* Give yourself Elytra
* Find high place or teleport yourself to some big Y-coordinate
* Start gliding and look at damage of Elytra
* Try to fall while gliding: there will be no fall damage.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -63,6 +63,7 @@ use pocketmine\event\player\PlayerPreLoginEvent;
 use pocketmine\event\player\PlayerQuitEvent;
 use pocketmine\event\player\PlayerRespawnEvent;
 use pocketmine\event\player\PlayerToggleFlightEvent;
+use pocketmine\event\player\PlayerToggleGlideEvent;
 use pocketmine\event\player\PlayerToggleSneakEvent;
 use pocketmine\event\player\PlayerToggleSprintEvent;
 use pocketmine\event\player\PlayerTransferEvent;
@@ -1659,6 +1660,15 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 		$this->timings->startTiming();
 
 		if($this->spawned){
+			if($this->isGliding()){
+				$this->resetFallDistance();
+				if($currentTick % 20 === 0){
+                	                $elytra = $this->inventory->getChestplate();
+        	                        $elytra->applyDamage(1);
+	                                $this->inventory->setChestplate($elytra);
+				}
+			}
+
 			$this->processMovement($tickDiff);
 
 			Timings::$timerEntityBaseTick->startTiming();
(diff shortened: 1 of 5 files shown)


src/pocketmine/Player.php
```diff
@@ -1659,6 +1660,15 @@ public function onUpdate(int $currentTick) : bool{
 		$this->timings->startTiming();
 
 		if($this->spawned){
+			if($this->isGliding()){
```
thebigsmileXD: The code for this should not stand above everything else inside the fucntion.

src/pocketmine/Player.php
```diff
@@ -1668,7 +1678,7 @@ public function onUpdate(int $currentTick) : bool{
 			if(!$this->isSpectator() and $this->isAlive()){
 				$this->checkNearEntities($tickDiff);
 
-				if($this->speed !== null){
+				if($this->speed !== null and !$this->isGliding()){
```
thebigsmileXD: In this case you would also need anti-cheating checks for elytra speed modification

src/pocketmine/Player.php
```diff
@@ -2713,8 +2723,27 @@ public function handlePlayerAction(PlayerActionPacket $packet) : bool{
 				}
 				return true;
 			case PlayerActionPacket::ACTION_START_GLIDE:
+				if($this->inventory->getChestplate()->getId() !== Item::ELYTRA){
+					$this->server->getLogger()->warning($this->getName() . " tried to start glide without Elytra");
```
thebigsmileXD: Add sending contents of chestplate to player (in case there were server<->client inventory inconsistencies)

src/pocketmine/Player.php
```diff
@@ -2713,8 +2723,27 @@ public function handlePlayerAction(PlayerActionPacket $packet) : bool{
...
+				$this->server->getPluginManager()->callEvent($ev);
+				if($ev->isCancelled()){
+					$this->sendData($this);
+				}else{
+					$this->setGliding(true);
+				}
+				return true;
 			case PlayerActionPacket::ACTION_STOP_GLIDE:
-				break; //TODO
+				$ev = new PlayerToggleGlideEvent($this, false);
```
thebigsmileXD: So  - no check for elytra here?

Also remember that you can take off / put on the elytra mid flight - which calls the toggle.

src/pocketmine/item/Elytra.php
```diff
@@ -0,0 +1,35 @@
...
+ *
+ *
+*/
+
+declare(strict_types=1);
+
+namespace pocketmine\item;
+
+
+class Elytra extends Durable{
```
thebigsmileXD: Shouldn't it also clarify as a chestplate?

thebigsmileXD:
Reasons to close? @dktapps

dktapps:
I already have local changes for this, quicker to skip the reviewal process (which is very tedious) and just commit my local code. Also, this implementation is sloppy.

PR #1937, opened 2018-01-19
Testing RakLib IPv6 support

dktapps:
## Introduction
see https://github.com/pmmp/RakLib/pull/22 for the library changes.


dktapps:
Not sure why Travis crashed. It's failed to bind to the IPv6 unspecified address for some reason.

dktapps:
```
[12:29:50] [RakLibServer thread/CRITICAL]: InvalidStateException: "Failed to bind to :: 19132: Address already in use" (EXCEPTION) in "vendor/pocketmine/raklib/server/UDPServerSocket" at line 37
```
how irritating...

dktapps:
It seems I'm a little lacking of understanding of socket differences between Linux and Windows. On Windows, IPv6 sockets can't receive IPv4 traffic and vice versa. On Linux, an IPv6 socket can receive _both_ IPv6 and IPv4 traffic, so a dual socket setup is not needed. This is problematic for implementation purposes.

dktapps:
This needs to take into account that PHP can be compiled without IPv6 support. It should either bail or not open the IPv6 interface if IPv6 is not available.

dktapps:
`/ban-ip` appears incapable of banning IPv6 addresses.

TheAz928:
This pull request is very useful if all problems are solved, which is possible to solve. Please consider doing this first because with many vps/dedicated providers give more ipv6 than ipv4, and those default ports are kinda a new priority now. Thanks.

dktapps:
It's not as trivial as it looks. There are a lot of internals that are entirely unprepared for IPv6. While this allows players to join via an IPv6 interface, there are a lot of other things that still need to be addressed.

SOF3:
Please explain what this pull request is blocking on. It is almost two years old.

dktapps:
There are some problems with /ban-ip as already mentioned, but the bigger problem is error handling.

Currently, the server will crash if either IPv4 or IPv6 fails to bind. With this in mind, it should be possible to disable all but one network interface.

PR #1961, opened 2018-01-24
Added Event->finallyCancelled

SOF3:
## Introduction
Quoting the scenario "DiscordMirror + HologramChat" from #1798:

> Suppose we have these two plugins:
> * DiscordMirror: it mirrors all chat messages to a Discord channel
> * HologramChat: it shows chat messages as holograms (FloatingText) rather than using the traditional Player->sendMessage().
> 
> The DiscordMirror plugin should listen to PlayerChatEvent at the MONITOR priority. It should ignore cancelled events, e.g. anti-spam plugins and plugins that override chat for other input (e.g. auth plugins may use it for receiving passwords).
> 
> What about the HologramChat plugin? It should override the default behaviour by cancelling the event and spawning the FloatingText in its handler. But if it cancels the event at LOWEST to HIGHEST, it will stop DiscordMirror from working. At the MONITOR event, since the order is undefined, it may or may not affect DiscordMirror.

This pull request is an alternative to #1798. Instead of allowing plugins to execute things at the EXECUTE priority, where multiple handlers may end up cancelling each other, HologramChat should handle the event at MONITOR priority, but using setFinallyCancelled() instead of setCancelled() to stop the default execution. This will prevent DiscordMirror from wrongly seeing the event is cancelled whilst it is, effectively, not cancelled.

### API changes

Added the getter and setter for `Event->isFinallyCancelled`. If an event is not cancellable, calling such methods will result in an exception, just as the getter and setter for `Event->isCancelled`.

Quoting the documentation for `Event->setFinallyCancelled` in this PR:
> If an event is set to finally-cancelled, the event will not be cancelled immediately, but will be cancelled just before the flow returns to PluginManager::callEvent()'s caller.
>
> In other words, passing true to this function makes sure the event ends up cancelled, but this is not visible to other plugins from Event::isCancelled().

## Backwards compatibility

If plugins use the events API correctly, there should not be backward incompatibilities.

The only exception is in "hacky" plugins that assume a non-cancelled event at MONITOR level will result in the event context being executed. Such cases are too specific and cannot be protected by this PR case-by-case.

## Follow-up

As this PR is related to cancelling events, it is written based on the rfc/remove-cancellable branch, considering there is no major objection towards #1794 and that there are conflicts merging #1794 to master. This PR should be merged to rfc/remove-cancellable if #1794 has not been merged at the time this PR is approved.

## Tests

Not tested yet.

The Travis-CI is failing because of submodule problems. It is not related to this PR.

## Relevant issues:
* Closes #1798

### DIFF
--- a/src/pocketmine/plugin/PluginManager.php
+++ b/src/pocketmine/plugin/PluginManager.php
@@ -25,6 +25,7 @@ namespace pocketmine\plugin;
 
 use pocketmine\command\PluginCommand;
 use pocketmine\command\SimpleCommandMap;
+use pocketmine\event\Cancellable;
 use pocketmine\event\Event;
 use pocketmine\event\EventPriority;
 use pocketmine\event\HandlerList;
@@ -712,6 +713,11 @@ class PluginManager{
 				$currentList = $currentList->getParent();
 			}
 		}
+
+		if($event instanceof Cancellable && !$event->isCancelled() && $event->isFinallyCancelled()){
+			$event->setCancelled();
+		}
+
 		--$this->eventCallDepth;
 	}
 
--- a/src/pocketmine/event/Event.php
+++ b/src/pocketmine/event/Event.php
@@ -79,7 +79,7 @@ abstract class Event{
 	 * @param bool $value
 	 */
 	public function setFinallyCancelled(bool $value = true) : void{
-		if(!$this->isCancellable()){
+		if(!($this instanceof Cancellable)){
 			throw new \BadMethodCallException("Event is not Cancellable");
 		}
 
@@ -90,7 +90,7 @@ abstract class Event{
 	 * @return bool
 	 */
 	public function isFinallyCancelled() : bool{
-		if(!$this->isCancellable()){
+		if(!($this instanceof Cancellable)){
 			throw new \BadMethodCallException("Event is not Cancellable");
 		}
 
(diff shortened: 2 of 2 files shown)


JackNoordhuis:
@dktapps well it doesn't directly change the outcome of the event, which could be interpreted as not changing the outcome from within the handler.

SOF3:
Please refer to my comment in #1798. Adding finally-cancelled allows for multiple execute handlers to co-exist, which then renders the EXECUTE priority unnecessary because running before or after "normal" MONITOR handlers makes no difference.

dktapps:
> As this PR is related to cancelling events, it is written based on the rfc/remove-cancellable branch

but it is not dependent on the said RFC - or more accurately can easily be done without changes made in that RFC. Please rebase this onto master with master compatibility instead.

src/pocketmine/plugin/PluginManager.php
```diff
@@ -730,6 +730,10 @@ public function callEvent(Event $event){
 				$this->server->getLogger()->logException($e);
 			}
 		}
+
+		if($event->isCancellable() && !$event->isCancelled() && $event->isFinallyCancelled()){
```
dktapps: I don't think it's necessary to check `isCancelled() here`.

dktapps:
Concept ACK, although I would be happier if a better name could be found for this than `finallyCancelled`.

PR #1962, opened 2018-01-24
Fixed armor not dropping

Muqsit:
## Introduction

Since the armor-playerinventory split-up in https://github.com/pmmp/PocketMine-MP/commit/6543d969102b6ce761717073fdd858467217b876, armor contents weren't dropping. This commit restores back the return values of `Human::getDrops()` to contain armor contents and Player's armor inventory to get cleared on death (if keepInventory is enabled).

## Tests

Death while wearing armor results in armor contents being dropped. This can also be tested by dumping `Human::getDrops()` to check whether it contains the contents from ArmorInventory.

Here's a dump of the following:
https://pastebin.com/fG1pvnV7
1x

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -3548,6 +3548,7 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 			if($this->inventory !== null){
 				$this->inventory->setHeldItemIndex(0, false); //This is already handled when sending contents, don't send it twice
 				$this->inventory->clearAll();
+				$this->armorInventory->clearAll();
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


src/pocketmine/Player.php
```diff
@@ -3548,6 +3548,7 @@ protected function onDeath(){
 			if($this->inventory !== null){
 				$this->inventory->setHeldItemIndex(0, false); //This is already handled when sending contents, don't send it twice
 				$this->inventory->clearAll();
+				$this->armorInventory->clearAll();
```
dktapps: shouldn't this be in its own null-check?

Issue #1979, opened 2018-01-28
Terminal::$formattingCodes has to be inherited for any thread which wants to use the MainLogger


dktapps:
### Issue description
On threads started using pthreads, the `STDIN`, `STDOUT` and `STDERR` constants are not defined. This is because they cannot be inherited due to them being resources (threads and resources don't mix).

This is a serious problem for any thread which wants to:
a) use the MainLogger, and
b) be run without inheriting all the garbage from the parent thread upstairs

because of https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/utils/Terminal.php#L56-L67
If `$formattingCodes` is not inherited due to thread start options, a crash will occur on the thread when it attempts to use the MainLogger.

dktapps:
Additionally, if this is duct-taped (for example by using a supplementary resource from `fopen('php://stdout', 'w');`, console colours will still not work correctly until `Terminal::init()` is called.

This is sloppy to say the least. The logger shouldn't be dependent on the using code to wipe its ass.

dktapps:
This could be solved by doing logger output on the logger thread itself - push a message containing `TextFormat` colour codes into the output buffer, and then apply formatting codes and echo from the `MainLogger` thread itself. Then `Terminal` only needs to be initialized on the `MainLogger` thread.

Basis: https://github.com/pmmp/PocketMine-MP/tree/more-threaded-logging

dktapps:
Amusingly this might also apply to logger timestamps if the timezone was set by auto-detection, because only the main thread will have the updated timestamps. This means that logger output from all other threads would have the UTC timestamp.

dktapps:
The solution in that commit is not ideal but it's the quickest solution to the problem for now.

PR #1983, opened 2018-01-30
MainLogger: Do output to console on the logger thread itself

dktapps:
This solves a range of issues (existing and potential) with logging, such as:

- echo is like a disk operation (blocking I/O) which may produce main thread slowdown.
- Logging to the console is _not_ currently synchronized - I am not actually sure why console output doesn't become illegible gibberish when multiple threads use the logger at the same time. Maybe something to do with locking when accessing properties of the logger? (implicit synchronization). This change solves that by doing all output to console on a single thread.
- In PowerShell, if the user selects some text in QuickEdit mode, the server will freeze next time it hits an echo call. This is because STDOUT write blocks until the user stops selecting text. This change resolves that problem.
- This also opens a possible solution to #1979 by handling escape code processing on the logger thread itself.

### API changes
- Added 2 new API methods - `MainLogger->pushToOutputBuffer()` and `MainLogger->getOutputBufferContents()`

### Behavioural changes
- Logging now echoes output on the logger thread itself instead of the calling thread, avoiding slowdowns and hangs due to console I/O.
- Title ticking output is now done on the logger thread due to visible slowdown it causes on the main thread.

To be clear, we're usually talking fractions of a millisecond. This is primarily intended to **resolve problems** and not to provide performance improvements.

## Backwards compatibility
This breaks tests as seen below - additionally, logger output can no longer be captured for a thread by using the PHP output buffer functions (`ob_*()`).

## Tests
Has been tested and several bugs fixed (bug fixes have since been ported to master since they were not specific to this change). However, Travis is broken because TesterPlugin needs updating.

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -171,7 +171,7 @@ class Server{
 
 	private $dispatchSignals = false;
 
-	/** @var \AttachableThreadedLogger */
+	/** @var MainLogger */
 	private $logger;
 
 	/** @var MemoryManager */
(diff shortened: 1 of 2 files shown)


dktapps:
The problem with this is that a block-up writing to disk or writing to console will affect each other, which is probably undesirable. It might be better to split up disk I/O and console I/O into two separate threads so that they can't affect each other.

dktapps:
This can be better done under a new planned logger API with isolated threaded logger attachments. As I mentioned above, disk and console I/O should not affect each other, but will if either one blocks.

PR #1984, opened 2018-01-30
Added Living::getEntityLookingAt() method

xBeastMode:
## Introduction
Similar to Living::getTargetBlock() this method is used to get the target entity in the view of a living entity.

### Relevant issues

none.

### API changes
Living::getEntityLookingAt() has two (2) optional paremeters:

- **distance**: an AlignedAxisBB class to select the distance in which the function searches, the bigger the distance the farther the function can search. If this parameter is null, the function will search in a 100 block radius.
- **precision**: a float that indicates how precise the entity has to be poiting to get the exact target entity. Values >= `0.998` indicate that the entity should be looking exactly at another entity, never use a value >= `1.0`

### Behavioural changes
none.

## Backwards compatibility
none.

## Follow-up
none.

## Requires translations:

none.

## Tests
```Living::getEntityLookingAt()``` was tested by swapping positions with the target entity when the client right clicked. The client was not teleported when not looking directly at an entity.

### DIFF
--- a/src/pocketmine/entity/Living.php
+++ b/src/pocketmine/entity/Living.php
@@ -800,24 +800,24 @@ abstract class Living extends Entity implements Damageable{
          * @return null|Entity
          *
          */
-	public function getEntityLookingAt(AxisAlignedBB $distance = null, float $precision = 0.998){
-	        if($distance === null){
-	                $pointA = $this->subtract(100, 100, 100);
-	                $pointB = $this->add(100, 100, 100);
+        public function getTargetEntity2(AxisAlignedBB $distance = null, float $precision = 0.998){
+                if($distance === null){
+                        $pointA = $this->subtract(100, 100, 100);
+                        $pointB = $this->add(100, 100, 100);
 
-	                $distance = new AxisAlignedBB($pointA->x, $pointA->y, $pointA->z, $pointB->x, $pointB->y, $pointB->z);
+                        $distance = new AxisAlignedBB($pointA->x, $pointA->y, $pointA->z, $pointB->x, $pointB->y, $pointB->z);
                 }
 
                 foreach($this->level->getNearbyEntities($distance) as $entity){
-	                if($entity === $this) continue;
+                        if($entity === $this) continue;
 
-	                $toEntity = $this->asVector3()->subtract($entity->asVector3());
-	                $direction = $this->getDirectionVector();
-	                $exact = $toEntity->normalize()->dot($direction);
+                        $toEntity = $entity->asVector3()->subtract($this->asVector3());
+                        $direction = $this->getDirectionVector();
+                        $exact = $toEntity->normalize()->dot($direction);
 
-	                if($exact >= $precision){
+                        if($exact >= $precision){
 
-	                        return $entity;
+                                return $entity;
 
                         }
                 }
--- a/src/pocketmine/entity/Living.php
+++ b/src/pocketmine/entity/Living.php
@@ -800,7 +800,7 @@ abstract class Living extends Entity implements Damageable{
          * @return null|Entity
          *
          */
-        public function getTargetEntity2(AxisAlignedBB $distance = null, float $precision = 0.998){
+        public function getEntityLookingAt(AxisAlignedBB $distance = null, float $precision = 0.998){
                 if($distance === null){
                         $pointA = $this->subtract(100, 100, 100);
                         $pointB = $this->add(100, 100, 100);
--- a/src/pocketmine/entity/Living.php
+++ b/src/pocketmine/entity/Living.php
@@ -37,6 +37,7 @@ use pocketmine\item\Armor;
 use pocketmine\item\Consumable;
 use pocketmine\item\enchantment\Enchantment;
 use pocketmine\item\Item as ItemItem;
+use pocketmine\math\AxisAlignedBB;
 use pocketmine\math\Vector3;
 use pocketmine\math\VoxelRayTrace;
 use pocketmine\nbt\tag\ByteTag;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/pocketmine/entity/Living.php [hunk 1]
```diff
@@ -786,10 +787,44 @@ public function getTargetBlock(int $maxDistance, array $transparent = []){
...
+         * @param AxisAlignedBB|null $distance
+         * @param float              $precision
+         *
+         * @return null|Entity
+         *
+         */
+        public function getEntityLookingAt(AxisAlignedBB $distance = null, float $precision = 0.998){
+                if($distance === null){
+                        $pointA = $this->subtract(100, 100, 100);
+                        $pointB = $this->add(100, 100, 100);
```
dktapps: ??????? Why on earth ???????

It should be more than sufficient to give a floating-point linear distance, and then multiply the entity's direction vector by it to obtain points A and B:

```php
$pointA = $this->asVector3();
$pointB = $this->getDirectionVector()->multiply($linearMaxDistance);
```

This solution is ludicrously inefficient.

src/pocketmine/entity/Living.php
```diff
@@ -786,10 +787,44 @@ public function getTargetBlock(int $maxDistance, array $transparent = []){
...
+
+                        $distance = new AxisAlignedBB($pointA->x, $pointA->y, $pointA->z, $pointB->x, $pointB->y, $pointB->z);
+                }
+
+                foreach($this->level->getNearbyEntities($distance) as $entity){
+                        if($entity === $this) continue;
+
+                        $toEntity = $entity->asVector3()->subtract($this->asVector3());
+                        $direction = $this->getDirectionVector();
+                        $exact = $toEntity->normalize()->dot($direction);
```
dktapps: This shouldn't be necessary. Getting the entity's AABB and checking for an intercept between the points should be sufficient.

src/pocketmine/entity/Living.php
```diff
@@ -786,10 +787,44 @@ public function getTargetBlock(int $maxDistance, array $transparent = []){
...
+        public function getEntityLookingAt(AxisAlignedBB $distance = null, float $precision = 0.998){
+                if($distance === null){
+                        $pointA = $this->subtract(100, 100, 100);
+                        $pointB = $this->add(100, 100, 100);
+
+                        $distance = new AxisAlignedBB($pointA->x, $pointA->y, $pointA->z, $pointB->x, $pointB->y, $pointB->z);
+                }
+
+                foreach($this->level->getNearbyEntities($distance) as $entity){
+                        if($entity === $this) continue;
```
dktapps: The second parameter of `getNearbyEntities()` should be used for this.

src/pocketmine/entity/Living.php
```diff
@@ -786,10 +787,44 @@ public function getTargetBlock(int $maxDistance, array $transparent = []){
...
+         *
+         * @return null|Entity
+         *
+         */
+        public function getEntityLookingAt(AxisAlignedBB $distance = null, float $precision = 0.998){
+                if($distance === null){
+                        $pointA = $this->subtract(100, 100, 100);
+                        $pointB = $this->add(100, 100, 100);
+
+                        $distance = new AxisAlignedBB($pointA->x, $pointA->y, $pointA->z, $pointB->x, $pointB->y, $pointB->z);
```
dktapps: Following from my above comment, the first three parameters should be the lesser of each x,y,z and the second three should be the greater ones.

src/pocketmine/entity/Living.php
```diff
@@ -786,10 +787,44 @@ public function getTargetBlock(int $maxDistance, array $transparent = []){
...
+                foreach($this->level->getNearbyEntities($distance) as $entity){
+                        if($entity === $this) continue;
+
+                        $toEntity = $entity->asVector3()->subtract($this->asVector3());
+                        $direction = $this->getDirectionVector();
+                        $exact = $toEntity->normalize()->dot($direction);
+
+                        if($exact >= $precision){
+
+                                return $entity;
```
dktapps: No distance checking? You want the entity closest to this entity along its line of sight.

src/pocketmine/entity/Living.php
```diff
@@ -786,10 +787,44 @@ public function getTargetBlock(int $maxDistance, array $transparent = []){
...
 
+        /**
+         *
+         * For the exact entity use a precision >= 0.998
+         *
+         * @param AxisAlignedBB|null $distance
+         * @param float              $precision
+         *
+         * @return null|Entity
+         *
```
dktapps: redundant empty line

src/pocketmine/entity/Living.php
```diff
@@ -786,10 +787,44 @@ public function getTargetBlock(int $maxDistance, array $transparent = []){
...
+         */
+        public function getEntityLookingAt(AxisAlignedBB $distance = null, float $precision = 0.998){
```
dktapps: a `?Entity` typehint can be applied here.

src/pocketmine/entity/Living.php
```diff
@@ -786,10 +787,44 @@ public function getTargetBlock(int $maxDistance, array $transparent = []){
 			}
 		}catch(\ArrayOutOfBoundsException $e){
 		}
-
```
dktapps: unrelated formatting change

src/pocketmine/entity/Living.php
```diff
@@ -786,10 +787,44 @@ public function getTargetBlock(int $maxDistance, array $transparent = []){
...
+                }
+
+                foreach($this->level->getNearbyEntities($distance) as $entity){
+                        if($entity === $this) continue;
+
+                        $toEntity = $entity->asVector3()->subtract($this->asVector3());
+                        $direction = $this->getDirectionVector();
+                        $exact = $toEntity->normalize()->dot($direction);
+
+                        if($exact >= $precision){
```
dktapps: This is unnecessary if an AABB ray trace is used.

dktapps (changes requested):

Doreez:
Definitely

dktapps:
Has been discussed with author elsewhere, who doesn't want to continue with it.

PR #1996, opened 2018-02-07
Update for the 1.2.10

CookieCode:
* Update for the last version
* I have tested it


dktapps:
https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/network/mcpe/protocol/ProtocolInfo.php#L32-L36

10/10 for not reading.

CookieCode:
Sorry. 10/10 I've lost my glasses.

PR #2002, opened 2018-02-09
Don't use static function

nnnlog:
## Introduction

Substitute `$this->server->getGamemodeName()` for `Server::getGamemodeName()`

### API changes

No

### Behavioural changes

no

## Tests

I completed test


Muqsit:
Why? It is a static function so why have it called non-statically?

dktapps:
> Why?

this is the billion dollar question... this change is wrong and completely pointless.

PR #2007, opened 2018-02-11
Sami documentation generator

dktapps:
## Introduction
This is intended to eventually replace Doxygen as our source for generating API documentation, since Doxygen doesn't work very well with PHP and is not very pretty to look at.

### Why not phpDocumentor or other alternatives?
phpDocumentor is a very nice tool, but it requires additional extensions which are not packaged as part of our PHP builds, making it awkward for users to generate their own documentation. Sami is easy to use and easy to restyle & customize, and has a nice modern feel.


Bruno99:
Can we already see the site for documentation with sami technology?

nathfreder:
This is still a work in progress?

dktapps:
I haven't worked on this recently because resources were diverted elsewhere.

nathfreder:
What else does it require before merging?

dktapps:
I don't recall off the top of my head. A pretty landing page was one of the things I was working on.

nathfreder:
I may work on that, once the documentation site is complete. So that the themes match.

nathfreder:
Fun fact: Sami is no longer maintained.

PR #2023, opened 2018-02-15
add slime block and make some blocks bouncy(Bed,SlimeBlock) + add EntityBounceOnBlockEvent

djdisodo:
## Tests

Now if the floor is a slime block and the player is not sneaking, it will not take damage
If you fall on a bed or slime block, it will bounce up(only for Livings and Items, not working for players....because It has some Bugs)
show particle when Living falls down more higher than 3 blocks(disable this when adding chicken)

I could not test it. because my pc is using 32bit os
please test it


TheAz928:
Your pull request is fully invaid, syntax error and stuff, learn php syntax before making a pull rec

PR #2054, opened 2018-02-24
Added hasEffects() function

zKoz210:
## Introduction

I know you could do this in a plugin, but this function would be very handy for developers.

### API changes

`Living::hasEffects()` - Returns true if the entity has at least one effect

## Tests
Everything works, there is nothing to test

### DIFF
--- a/src/pocketmine/entity/Living.php
+++ b/src/pocketmine/entity/Living.php
@@ -238,7 +238,7 @@ abstract class Living extends Entity implements Damageable{
 	 * @return bool
 	 */
 	public function hasEffects() : bool{
-		return count($this->effects) > 0;
+		return !empty($this->effects);
 	}
 
 	/**
--- a/src/pocketmine/entity/Living.php
+++ b/src/pocketmine/entity/Living.php
@@ -232,6 +232,14 @@ abstract class Living extends Entity implements Damageable{
 	public function hasEffect(int $effectId) : bool{
 		return isset($this->effects[$effectId]);
 	}
+	
+	/**
+	 * Returns whether the any effect is active on the mob.
+	 * @return bool
+	 */
+	public function hasEffects() : bool{
+		return count($this->effects) > 0;
+	}
 
 	/**
 	 * Adds an effect to the mob.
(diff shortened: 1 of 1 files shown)


dktapps:
> Everything works, there is nothing to test

Please don't take this attitude... if you want to contribute code, then test it first, always.

src/pocketmine/entity/Living.php
```diff
@@ -232,6 +232,14 @@ public function getEffect(int $effectId){
 	public function hasEffect(int $effectId) : bool{
 		return isset($this->effects[$effectId]);
 	}
+	
+	/**
+	 * Returns whether the any effect is active on the mob.
+	 * @return bool
+	 */
+	public function hasEffects() : bool{
+		return count($this->effects) > 0;
```
dktapps: This could also be `!empty($this->effects)`

zKoz210: Yeah, I know, but what's better? To do as you have?

src/pocketmine/entity/Living.php
```diff
@@ -232,6 +232,14 @@ public function getEffect(int $effectId){
 	public function hasEffect(int $effectId) : bool{
 		return isset($this->effects[$effectId]);
 	}
+	
+	/**
+	 * Returns whether the any effect is active on the mob.
```
Muqsit: Grammar?

zKoz210: @Muqsit I'd appreciate it if you fixed my mistake. I don't know english well.

Muqsit: Returns whether the mob has any active effects.

src/pocketmine/entity/Living.php
```diff
@@ -238,7 +238,7 @@ public function hasEffect(int $effectId) : bool{
 	 * @return bool
 	 */
 	public function hasEffects() : bool{
-		return count($this->effects) > 0;
+		return !empty($this->effects);
```
KennFatt: checking an array with empty() is much way better than count it. In this case, we only check if empty or filled. 👍

KennFatt (approved):

dktapps (approved):

PR #2070, opened 2018-03-01
Fixed player can eat cake in creative mode

zKoz210:


dktapps:
The title of this pull request is misleading. It's preferred that you describe the bug that you're fixing as opposed to the the result of the bug fix.

dktapps:
I'm unable to reproduce the bug described here. Please give reproducing steps.

zKoz210:
1. 
image

2. /gamemode 1
3. Tap to Cake

dktapps:
This would appear to affect all eatable items. The only reason the same bug does not affect normal foods is because the client doesn't send the eating action for those.

PR #2089, opened 2018-03-10
Added getRandomTickedBlocks() function

zKoz210:
## Introduction

It would be more convenient to completely disable all possible variations of the blocks.

### API changes

`Level::getRandomTickedBlocks()` - Returns array with random ticked blocks

## Tests
It works, tested on my own server

```
foreach ($this->getServer()->getDefaultLevel()->getRandomTickedBlocks() as $id => $block) {
	if ($block === null) {
		continue;
	}
	var_dump($id);
	$this->getServer()->getDefaultLevel()->removeRandomTickedBlock($id);
}
```

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -966,8 +966,8 @@ class Level implements ChunkManager, Metadatable{
 	public function clearChunkCache(int $chunkX, int $chunkZ){
 		unset($this->chunkCache[Level::chunkHash($chunkX, $chunkZ)]);
 	}
-	
-	public function getRandomTickedBlocks(): SplFixedArray{
+
+	public function getRandomTickedBlocks() : SplFixedArray{
 		return $this->randomTickBlocks;
 	}
 
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -966,6 +966,10 @@ class Level implements ChunkManager, Metadatable{
 	public function clearChunkCache(int $chunkX, int $chunkZ){
 		unset($this->chunkCache[Level::chunkHash($chunkX, $chunkZ)]);
 	}
+	
+	public function getRandomTickedBlocks(): SplFixedArray{
+		return $this->randomTickBlocks;
+	}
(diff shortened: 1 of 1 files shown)


src/pocketmine/level/Level.php
```diff
@@ -966,6 +966,10 @@ public function clearCache(bool $force = false){
 	public function clearChunkCache(int $chunkX, int $chunkZ){
 		unset($this->chunkCache[Level::chunkHash($chunkX, $chunkZ)]);
 	}
+
```
dktapps: trailing whitespace

src/pocketmine/level/Level.php
```diff
@@ -966,6 +966,10 @@ public function clearCache(bool $force = false){
 	public function clearChunkCache(int $chunkX, int $chunkZ){
 		unset($this->chunkCache[Level::chunkHash($chunkX, $chunkZ)]);
 	}
+	
+	public function getRandomTickedBlocks(): SplFixedArray{
```
dktapps: formatting

dktapps (dismissed):
I think the code you offered as an example for this doesn't make any sense. However this could be a useful API method regardless.

zKoz210:
Why? Maybe I want to make a hub where nothing has to grow or change. I'm too lazy to do a list of ID, it is easier to write 3 lines of code.

dktapps:
This function is not needed to achieve that goal. You could simply use [this](https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/block/BlockFactory.php#L402).
```php
foreach(BlockFactory::getBlockStatesArray() as $block){
    $level->removeRandomTickedBlock($block->getId());
}
```

zKoz210:
That may be so, but my code seems to be cleaner.

dktapps:
Your code is an overcomplication.

I'm not saying that this PR is not useful, I am saying that the use you offered for it is pointless.

zKoz210:
I understood. I'm just wondering why my way was pointless. Thanks

Issue #1881, opened 2018-01-04
Asynchronous events


SOF3:
<details open>
<summary>TL;DR</summary>
- Async event allows you to use async task, wait for user response or do other time-consuming actions in an event handler.
- Async event is opt-in. This is not a compatibility-destructive proposal.

### Example
This example shows how a plugin can translate PlayerChatEvent messages with an online API:

```php
// Phase II: changes to the Player class for firing PlayerChatEvent
function Player::handleChat(){
    $event = new PlayerChatEvent(...);
    // $server->getPluginManager()->callEent($event); // the traditional way
    // the new way is like this:
    $server->getPluginManager()->callAsyncEvent($event, function(PlayerChatEvent $event){
        if(!$event->isCancelled()){
            foreach($event->getRecipients() as $recipient){
                if(!$recipient->isOnline()) continue; // Resources may have been invalidated after the event call ends.
                $recipient->sendMessage($event->getMessage()); // yes, I know it has some sprintf() stuff, but this is just an example
            }
    });
}
```

```php
class TranslatePlugin extends PluginBase implements Listener{
    public function onPlayerChat(PlayerChatEvent $event){
        $event->pause();
        $this->getServer()->getScheduler()->scheduleAsyncTask(new TranslateTask($event));
    }
}
```

```php
class TranslateTask extends AsyncTask{
    private $message;

    public function __construct(PlayerChatEvent $event){
        $this->storeLocal($event);
        $this->message = $event->getMessage();
    }

    public function onRun(){
        $result = Utils::getURL("https://example.com/translate?text=" . urlencode($this->message));
        $this->setResult($result);
    }

    public function onCompletion(Server $server){
        $event = $this->fetchLocal($server);
        $result = $this->getResult();
        if($result === false) $event->setCancelled();
        else $event->setMessage($result);
        $event->continue();
    }
}
```

There are no changes to PlayerChatEvent.php
</details>

# Full RFC
**Warning: The following RFC has more than 1500 words.** If you don't want to read it, just read the TL;DR above and the "Expected use cases" at the end.

### Introduction
<details open>
<summary>Introduction: Why async events?</summary>

It is difficult to write a plugin that fetches data asynchronously, because many parts of the PocketMine API expect you to return a value **immediately**.

Currently, plugins usually use either of these two solutions:
- Allow the event to execute, then undo the event after the plugin finds out the event should have been cancelled.
[...]

PR #2096, opened 2018-03-11
Asynchronous events: Phase I

SOF3:
## Introduction
Event handlers can now return asynchronously. By asynchronously, I mean that event listeners can delay the execution of the next event handler, hence the execution of the event calling context.

This is very useful for plugins that wish to set useful values in event handlers, instead of using hacks. For example, a plugin that uses an online agent to check whether an event is cancelled can cancel the event after the online agent returns a value, instead of redoing/undoing the event a few ticks later, causing undesired side effects.

### Relevant issues
* Proposed in #1881
* Based on #1792 
* #1798 is no longer necessary due to this PR
* #1961 is no longer necessary due to this PR

##### pocketmine\event\Event
```php
public final function pause(int $ticks = 200, ?callable $onTimeout = null) : void;
public final function continue() : void;
```

##### pocketmine\plugin\PluginManager
```php
public function callAsyncEvent(Event $event, callable $onCompletion) : void;
```

### Behavioural changes
This is a pure API addition. There shall be no behaviour changes unless a plugin is using the new API.

## Backward compatibility
This is a pure API addition. All backward incompatibility shall be attributed to plugins, not the core.

## Follow-up

The core should use features in this PR. Details are described in #1881.

## Tests

https://gist.github.com/SOF3/a02c48c5ed6f9c69cddc6229bbb68ae3

### DIFF
--- a/src/pocketmine/event/Event.php
+++ b/src/pocketmine/event/Event.php
@@ -26,12 +26,31 @@ declare(strict_types=1);
  */
 namespace pocketmine\event;
 
+use pocketmine\plugin\RegisteredListener;
+use pocketmine\Server;
+use pocketmine\utils\MainLogger;
+
 abstract class Event{
+	public const MAX_PAUSE_TICKS = 6000;
 
 	/** @var string|null */
 	protected $eventName = null;
 	/** @var bool */
 	private $isCancelled = false;
+
+	/** @var bool */
+	private $async = false;
+	/** @var RegisteredListener[]|null */
+	private $asyncQueue = null;
+	/** @var bool */
+	private $asyncComplete = false;
+	/** @var callable|null */
+	private $asyncCompleteFunc = null;
+
+	/** @var int|null */
+	private $pauseTimeout = null;
+	/** @var callable|null */
+	private $pauseTimeoutFunc = null;
 
 	/**
 	 * @return string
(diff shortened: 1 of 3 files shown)


src/pocketmine/event/Event.php
```diff
@@ -77,14 +87,139 @@ public function setCancelled(bool $value = true){
 	}
 
 	/**
-	 * @return HandlerList
+	 * @internal Only to be called from PluginManager->callAsyncEvent
+	 *
+	 * @param RegisteredListener[] $asyncQueue
+	 * @param callable             $onCompletion
+	 */
+	public final function setAsyncQueue(array $asyncQueue, callable $onCompletion) : void{
```
dktapps: nit: code style violation

src/pocketmine/event/Event.php
```diff
@@ -77,14 +87,139 @@ public function setCancelled(bool $value = true){
...
+		$this->async = true;
+		$this->asyncQueue = $asyncQueue;
+		$this->asyncCompleteFunc = $onCompletion;
+	}
+
+	public final function isAsync() : bool{
```
dktapps: same here

src/pocketmine/event/Event.php
```diff
@@ -77,14 +87,139 @@ public function setCancelled(bool $value = true){
...
+
+	/**
+	 * @internal Only to be called from PluginManager->callAsyncEvent
+	 *
+	 * @param int $currentTick
 	 */
-	public function getHandlers() : HandlerList{
-		if(static::$handlerList === null){
-			static::$handlerList = new HandlerList();
+	public final function startAsyncQueue(int $currentTick) : void{
```
dktapps: and here

src/pocketmine/event/Event.php
```diff
@@ -77,14 +87,139 @@ public function setCancelled(bool $value = true){
...
+			return;
+		}
+
+		$this->doAsyncLoop();
+		if($this->asyncComplete){
+			$this->onComplete();
+		}
+	}
+
+	public final function asyncCheck(int $currentTick) : bool{
```
dktapps: and here

src/pocketmine/event/Event.php
```diff
@@ -77,14 +87,139 @@ public function setCancelled(bool $value = true){
...
+		}
+		if($this->pauseTimeout !== null){
+			throw new \InvalidStateException("Event has already been paused");
+		}
+
+		$this->pauseTimeout = Server::getInstance()->getTick() + $ticks;
+		$this->pauseTimeoutFunc = $onTimeout ?? [$this, "onTimeout"];
+	}
+
+	public final function continue() : void{
```
dktapps: and here

src/pocketmine/event/Event.php
```diff
@@ -77,14 +87,139 @@ public function setCancelled(bool $value = true){
...
+				$this->getEventName(),
+				current($this->asyncQueue)->getPlugin()->getDescription()->getFullName(),
+				get_class(current($this->asyncQueue)->getListener())
+			]));
 	}
 
+	/**
+	 * @return bool
+	 */
+	public final function isAsyncComplete() : bool{
```
dktapps: and here

src/pocketmine/event/Event.php
kenygamer: `final public function setAsyncQueue(array $asyncQueue, callable $onCompletion) : void`

SOF3:
I changed the max delay from 1200 ticks to 6000 ticks.

src/pocketmine/event/Event.php
```diff
@@ -67,4 +86,147 @@ public function setCancelled(bool $value = true){
...
+
+		$this->async = false;
+		$this->asyncComplete = false;
+		$this->asyncQueue = null;
+		$this->pauseTimeout = null;
+		$this->pauseTimeoutFunc = null;
+	}
+
+	public function pause(int $ticks = 200, ?callable $onTimeout = null) : void{
+		if(!$this->async){
```
dktapps: Not a fan of conditionally-useless functions.

src/pocketmine/event/Event.php
```diff
@@ -67,4 +86,147 @@ public function setCancelled(bool $value = true){
...
+		// do not execute anything directly after this, to prevent recursion stack overflow or concurrency due to the plugin executing something after this call
+	}
+
+	public function onTimeout() : void{
+		MainLogger::getLogger()->error(
+			Server::getInstance()->getLanguage()->translateString("pocketmine.plugin.eventTimeout", [
+				$this->getEventName(),
+				current($this->asyncQueue)->getPlugin()->getDescription()->getFullName(),
+				get_class(current($this->asyncQueue)->getListener())
+			]));
```
dktapps: I don't like this. It makes the Event base circularly dependent on other things.

dktapps (commented):
I haven't reviewed this in depth, but have left a couple of comments.

SOF3:
Closed due to lack of activity. I will open another pull request when changes such as concurrent execution have been added.

PR #2124, opened 2018-03-27
Implement Player -> Entity interaction

CortexPE:
## Introduction
This PR implements Player to Entity interaction... Which makes ArmorStands, Villager Trading UIs, Operations on Player interact (like on Mineplex) possible without overusing EntityDamageEvent or using a separate Packet Handler for plugins.

### Relevant issues
Currently, No issues address this but it is a 'TODO' in https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/Player.php#L2471

### API changes
This adds Entity::onInteract() and PlayerEntityInteractEvent

### Behavioural changes
None

## Backwards compatibility
Purely API additions, no BC Breaks.

## Follow-up
None

## Tests
This can be tested by registering an entity that uses Entity::onInteract() or by using a plugin and listening for PlayerEntityInteractEvent

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -2469,12 +2469,12 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 
 				switch($type){
 					case InventoryTransactionPacket::USE_ITEM_ON_ENTITY_ACTION_INTERACT:
-						$item = $packet->trData->itemInHand;
+						$item = $this->inventory->getItemInHand();
 						$clickPos = $packet->trData->clickPos;
 						$slot = $packet->trData->hotbarSlot;
 
 						$ev = new PlayerEntityInteractEvent($this, $target, $item, $clickPos, $slot);
-						
+
 						if(!$this->canInteract($target, 8)){
 							$ev->setCancelled();
 						}
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -51,6 +51,7 @@ use pocketmine\event\player\PlayerChatEvent;
 use pocketmine\event\player\PlayerCommandPreprocessEvent;
 use pocketmine\event\player\PlayerDeathEvent;
 use pocketmine\event\player\PlayerEditBookEvent;
+use pocketmine\event\player\PlayerEntityInteractEvent;
 use pocketmine\event\player\PlayerExhaustEvent;
 use pocketmine\event\player\PlayerGameModeChangeEvent;
 use pocketmine\event\player\PlayerInteractEvent;
@@ -2468,7 +2469,23 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 
 				switch($type){
 					case InventoryTransactionPacket::USE_ITEM_ON_ENTITY_ACTION_INTERACT:
-						break; //TODO
+						$item = $packet->trData->itemInHand;
+						$clickPos = $packet->trData->clickPos;
+						$slot = $packet->trData->hotbarSlot;
+
+						$ev = new PlayerEntityInteractEvent($this, $target, $item, $clickPos, $slot);
+						
+						if(!$this->canInteract($target, 8)){
+							$ev->setCancelled();
+						}
+
+						$this->server->getPluginManager()->callEvent($ev);
+
+						if(!$ev->isCancelled()){
+							$target->onInteract($this, $item, $clickPos, $slot);
+						}
+
+						return true;
 					case InventoryTransactionPacket::USE_ITEM_ON_ENTITY_ACTION_ATTACK:
 						if(!$target->isAlive()){
 							return true;
--- a/src/pocketmine/entity/Entity.php
+++ b/src/pocketmine/entity/Entity.php
@@ -990,10 +990,8 @@ abstract class Entity extends Location implements Metadatable, EntityIds{
 	 * @param Item    $item
 	 * @param Vector3 $clickPos
 	 * @param int     $slot
-	 * @return void
 	 */
 	public function onInteract(Player $player, Item $item, Vector3 $clickPos, int $slot) : void{
-		return;
 	}
 
 	public function entityBaseTick(int $tickDiff = 1) : bool{
--- a/src/pocketmine/entity/Entity.php
+++ b/src/pocketmine/entity/Entity.php
@@ -992,6 +992,7 @@ abstract class Entity extends Location implements Metadatable, EntityIds{
 	 * @param int     $slot
 	 */
 	public function onInteract(Player $player, Item $item, Vector3 $clickPos, int $slot) : void{
+
 	}
 
 	public function entityBaseTick(int $tickDiff = 1) : bool{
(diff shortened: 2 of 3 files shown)


src/pocketmine/event/player/PlayerEntityInteractEvent.php
```diff
@@ -0,0 +1,90 @@
...
+	 * @return Vector3
+	 */
+	public function getClickPosition() : Vector3{
+		return $this->clickPos;
+	}
+
+	/**
+	 * @return int
+	 */
+	public function geSlot() : int{
```
fuyutsuki: typo?

fuyutsuki (changes requested):
Please fix `PlayerEntityInteractEvent::geSlot()`

src/pocketmine/event/player/PlayerEntityInteractEvent.php
CortexPE: Fixed on https://github.com/pmmp/PocketMine-MP/pull/2124/commits/79298cfd5a8b8ed7751dc9de136a2181ae3d82e2

fuyutsuki: Thx

src/pocketmine/entity/Entity.php
```diff
@@ -982,6 +983,19 @@ public function getDataPropertyManager() : DataPropertyManager{
...
 	}
 
+	/**
+	 * Called when interacted or tapped by a Player
+	 *
+	 * @param Player $player
+	 * @param Item $item
+	 * @param Vector3 $clickPos
+	 * @param int $slot
+	 * @return bool
```
dktapps: The parameter names should be aligned, please indent them appropriately.

src/pocketmine/Player.php
```diff
@@ -2468,7 +2469,23 @@ public function handleInventoryTransaction(InventoryTransactionPacket $packet) :
 
 				switch($type){
 					case InventoryTransactionPacket::USE_ITEM_ON_ENTITY_ACTION_INTERACT:
-						break; //TODO
+						$item = $packet->trData->itemInHand;
+						$clickPos = $packet->trData->clickPos;
+						$slot = $packet->trData->hotbarSlot;
```
dktapps: These should not be used. The client can lie about these and you're not checking if they are valid. If this was an attack the client could say it had a diamond sword and the server wouldn't know any better.

src/pocketmine/event/player/PlayerEntityInteractEvent.php
```diff
@@ -0,0 +1,90 @@
...
+	 * @return Vector3
+	 */
+	public function getClickPosition() : Vector3{
+		return $this->clickPos;
+	}
+
+	/**
+	 * @return int
+	 */
+	public function getSlot() : int{
```
dktapps: The use of this function is not immediately clear. Documentation should be added (this is the selected hotbar slot, yes?).

src/pocketmine/entity/Entity.php
```diff
@@ -982,6 +983,19 @@ public function getDataPropertyManager() : DataPropertyManager{
...
+	 */
+	public function onInteract(Player $player, Item $item, Vector3 $clickPos, int $slot) : bool {
+		return true;
+	}
+
```
dktapps: nit: trailing whitespace

src/pocketmine/Player.php
```diff
@@ -2468,7 +2469,23 @@ public function handleInventoryTransaction(InventoryTransactionPacket $packet) :
...
+
+						$ev = new PlayerEntityInteractEvent($this, $target, $item, $clickPos, $slot);
+
```
dktapps: nit: trailing whitespace

src/pocketmine/entity/Entity.php
```diff
@@ -982,6 +983,19 @@ public function getDataPropertyManager() : DataPropertyManager{
...
+	 * Called when interacted or tapped by a Player
+	 *
+	 * @param Player $player
+	 * @param Item $item
+	 * @param Vector3 $clickPos
+	 * @param int $slot
+	 * @return bool
+	 */
+	public function onInteract(Player $player, Item $item, Vector3 $clickPos, int $slot) : bool {
+		return true;
```
dktapps: What is the significance of this return value? It appears to be superfluous.

dktapps (changes requested):
I've given this a quick review.

src/pocketmine/Player.php
CortexPE: I'll change L#2472 to use the server-side Item in hand on the next commit...
clickPos will still remain dependent on the client as they may be using 'split touch'.

dktapps: Yes, clickpos is fine. I was referring to the slot and item. Sorry for not being clear.

src/pocketmine/entity/Entity.php
```diff
@@ -982,6 +983,19 @@ public function getDataPropertyManager() : DataPropertyManager{
...
+	 * Called when interacted or tapped by a Player
+	 *
+	 * @param Player  $player
+	 * @param Item    $item
+	 * @param Vector3 $clickPos
+	 * @param int     $slot
+	 * @return void
+	 */
+	public function onInteract(Player $player, Item $item, Vector3 $clickPos, int $slot) : void{
+		return;
```
dktapps: superfluous return statement and `@return`

src/pocketmine/entity/Entity.php
```diff
@@ -982,6 +983,17 @@ public function getDataPropertyManager() : DataPropertyManager{
...
+	/**
+	 * Called when interacted or tapped by a Player
+	 *
+	 * @param Player  $player
+	 * @param Item    $item
+	 * @param Vector3 $clickPos
+	 * @param int     $slot
+	 */
+	public function onInteract(Player $player, Item $item, Vector3 $clickPos, int $slot) : void{
+	}
```
CortexPE: Closing bracket is on the next line following PSR-2 Standards with exceptions to The PocketMine-MP Code Syntax Standards in https://github.com/pmmp/PocketMine-MP/blob/master/CONTRIBUTING.md > "Opening braces MUST go on the same line, and MUST NOT have spaces before."

dktapps: there should be an empty line after `public function ...`, see other methods for examples.

daniktheboss:
I think this is more of an entity event. EntityInteractEvent not PlayerEntityInteractEvent

jojoe77777:
@daniktheboss EntityInteractEvent implies that the entity is interacting.

daniktheboss:
Oops lol

TheAz928:
@CortexPE PlayerInteractEntityEvent should be the event name, it makes more sense

src/pocketmine/Player.php
```diff
@@ -2468,7 +2469,23 @@ public function handleInventoryTransaction(InventoryTransactionPacket $packet) :
 
 				switch($type){
 					case InventoryTransactionPacket::USE_ITEM_ON_ENTITY_ACTION_INTERACT:
-						break; //TODO
+						$item = $this->inventory->getItemInHand();
+						$clickPos = $packet->trData->clickPos;
+						$slot = $packet->trData->hotbarSlot;
```
dktapps: I don't think this makes sense. You're already assuming that the item used for this action is the currently held item, so I don't see why it's necessary to throw the slot around.

dktapps:
Closing due to conflicts and lack of activity.

PR #2125, opened 2018-03-28
Ignore plugin folder name contains (dot) phar

KennFatt:
## Introduction

Fix plugins directory with ".phar" and system detect it as a real phar, but actually its a directory.

### API changes

No API changes.

### Behavioural changes

Nothing affect to main core, its just detect if a $file is directory.

## Backwards compatibility

PharPluginLoader only load Phar files, not directory.

## Follow-up

`UnexpectedValueException: "Cannot create phar 'C:/Users/KENNAN/Desktop/pmmp/latest/1.2/MinigamesEvent/plugins/PlayerAura_v1.2.phar', file extension (or combination) not recognised or the directory does not exist" (EXCEPTION) in "src/pocketmine/plugin/PharPluginLoader" at line 85`

## Tests

Its just skip directory that contain ".phar".

### DIFF
--- a/src/pocketmine/plugin/PharPluginLoader.php
+++ b/src/pocketmine/plugin/PharPluginLoader.php
@@ -81,6 +81,10 @@ class PharPluginLoader implements PluginLoader{
 	 * @return null|PluginDescription
 	 */
 	public function getPluginDescription(string $file){
+		if(is_dir($file)) {
+			return null;
+		}
+
 		$phar = new \Phar($file);
 		if(isset($phar["plugin.yml"])){
 			$pluginYml = $phar["plugin.yml"];


KennFatt:
I thought so. But Im afraid if someday we will use FolderPluginLoader without DevTools 😃 . Ok, wait a minute.

KennFatt:
@dktapps If we put it on top of `loadPlugin()`, then it would break `FolderPluginLoader` too. So, basically its prevent `PharPluginLoader` to load a folder with ".phar" name.

dktapps:
I don't see how that would have any bearing on FolderPluginLoader. I'm saying you should put a check here: https://github.com/pmmp/PocketMine-MP/pull/2125/files#diff-1100e6e5a8fefc58972fee094d9bc673R52

instead of in the plugin description function.

KennFatt:
Ahhh I  see 😢 01:31 AM here, sorry.

dktapps:
again, I said you should put a check in PharPluginLoader->loadPlugin(), **not in the PluginManager**.

PR #2131, opened 2018-04-03
Add /daylock and /alwaysday

lukeeey:
## Introduction
This PR implements the `/alwaysday` command and `/daylock` command. It also removes `/time start` and `/time stop`.

Since both commands act *exactly* the same, i have implemented the logic in the daylock command and just registered the class for both "daylock" and "alwaysday". Please tell me if this will have unexpected issues, but i didnt see the point in duplicating code or adding it as an alias since it should show up as a seperate command like in vanilla.

When locking the time, it initially sets it to 5000 (like vanilla) where it will stay until you change it (and then obviously it will stay there).

### Relevant issues
None that i can find.

### API changes
* Removed permission `pocketmine.command.time.start`
* Removed permission `pocketmine.command.time.stop`
* Added permission `pocketmine.command.daylock`

### Behavioural changes
None.

## Backwards compatibility
This will (probably) have no effect on plugins, unless they stopped the time using `Server->dispatchCommand` (which they shouldn't be doing anyway).

## Follow-up
Translations required, however the command success messages should display when executed from the client.

The start/stop text will need to be removed from the `/time` command usage string.

## Tests
Join a server and test the following commands:

* /daylock
* /alwaysday
* /daylock true
* /daylock false

It all works perfectly.

### DIFF
--- a/src/pocketmine/command/SimpleCommandMap.php
+++ b/src/pocketmine/command/SimpleCommandMap.php
@@ -23,6 +23,7 @@ declare(strict_types=1);
 
 namespace pocketmine\command;
 
+use pocketmine\command\defaults\DayLockCommand;
 use pocketmine\command\defaults\BanCommand;
 use pocketmine\command\defaults\BanIpCommand;
 use pocketmine\command\defaults\BanListCommand;
@@ -89,6 +90,8 @@ class SimpleCommandMap implements CommandMap{
 			new BanCommand("ban"),
 			new BanIpCommand("ban-ip"),
 			new BanListCommand("banlist"),
+			new DayLockCommand("alwaysday"),
+		  new DayLockCommand("daylock"),
 			new DefaultGamemodeCommand("defaultgamemode"),
 			new DeopCommand("deop"),
 			new DifficultyCommand("difficulty"),
--- a/src/pocketmine/command/SimpleCommandMap.php
+++ b/src/pocketmine/command/SimpleCommandMap.php
@@ -90,7 +90,6 @@ class SimpleCommandMap implements CommandMap{
 			new BanCommand("ban"),
 			new BanIpCommand("ban-ip"),
 			new BanListCommand("banlist"),
-			new DayLockCommand("alwaysday"),
 			new DayLockCommand("daylock"),
 			new DefaultGamemodeCommand("defaultgamemode"),
 			new DeopCommand("deop"),
--- a/src/pocketmine/command/SimpleCommandMap.php
+++ b/src/pocketmine/command/SimpleCommandMap.php
@@ -91,7 +91,7 @@ class SimpleCommandMap implements CommandMap{
 			new BanIpCommand("ban-ip"),
 			new BanListCommand("banlist"),
 			new DayLockCommand("alwaysday"),
-		  new DayLockCommand("daylock"),
+			new DayLockCommand("daylock"),
 			new DefaultGamemodeCommand("defaultgamemode"),
 			new DeopCommand("deop"),
 			new DifficultyCommand("difficulty"),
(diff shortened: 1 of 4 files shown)


src/pocketmine/command/defaults/DayLockCommand.php
```diff
@@ -0,0 +1,75 @@
...
+		if(!$this->testPermission($sender)){
+			return true;
+		}
+
+		if(count($args) > 1){
+			throw new InvalidCommandSyntaxException();
+		}
+
+		if(count($args) < 1 or $args[0] === "true"){
+			foreach($sender->getServer()->getLevels() as $level){
```
jasonwynn10: Why does the command stop and start time in every level rather than just the one the player is in?

lukeeey: That's how it was like in TimeCommand.php, i can change it if needed.

dktapps: The rules around commands operating on levels are currently undefined, and adding level arguments increases the overload complexity. Since we have a garbage commands API at the moment, this kind of change is better avoided.

Better to maintain existing behaviour for consistency's sake if nothing else.

src/pocketmine/command/defaults/DayLockCommand.php
```diff
@@ -0,0 +1,76 @@
...
+				$level->stopTime();
+				$level->checkTime();
+			}
+
+			Command::broadcastCommandMessage($sender, new TranslationContainer("commands.always.day.locked"));
+		}elseif($args[0] === "false"){
+			foreach($sender->getServer()->getLevels() as $level){
+				$level->checkTime();
+				$level->startTime();
+				$level->checkTime();
```
dktapps: `checkTime()` call is redundant. I know this was the same in the original code, but you could change it if you want. If not,  I'll fix it separately.

dktapps:
I don't think that the `start` and `stop` subcommands should be removed from /time. It's also debatable as to whether there's any real benefit to having three commands to do things that a single command can already do without any changes. Seems like needless /help pollution to me.

Particularly I strongly disagree with having the same command registered twice under different names. Since they both do the same thing, it seems stupid to me to make them appear separately.

lukeeey:
I agree its stupid, but I thought the goal of PMMP was to completely replicate vanilla behaviour.

dktapps:
I don't agree with replicating stupidity. If you want a vanilla server, get a Realm.

lukeeey:
Then close the PR lol

lukeeey:
What counter arguments could there be? This PR tries to bridge the gap between the lack of commands in PocketMine and vanilla. If we can achieve parity, it will be easy for people to remember commands and their syntax. We also have the issue of things like command blocks. When they are implemented, what if we made a world on vanilla and used a vanilla command, but then loaded the world on the server? You would need to change it all. This isnt quite a concern yet but it could be in the future.

lukeeey:
If this PR is merged and you dont want to clutter /help as much, just add alwaysday as an alias instead

dktapps:
/alwaysday is actually an alias anyway (see [this](https://gist.github.com/dktapps/050fd9b985d993bead412c6934b9538a#file-commands_decoded-json-L784))

lukeeey:
Well PocketMine aliases dont show up in the help page so i registered it again instead

dktapps:
You should keep your commits atomic. If a change can be used on its own without the other things you want to commit, commit it separately.

src/pocketmine/command/SimpleCommandMap.php
```diff
@@ -89,6 +90,8 @@ private function setDefaultCommands(){
 			new BanCommand("ban"),
 			new BanIpCommand("ban-ip"),
 			new BanListCommand("banlist"),
+			new DayLockCommand("alwaysday"),
+			new DayLockCommand("daylock"),
```
dktapps: One of these should be registered as an alias instead of duplicating the command.

dktapps (changes requested):
This should not alter the behaviour of the existing `/time start` and `/time stop`.

dktapps:
Closing this due to lack of activity. There are changes requested but the PR is stale.

PR #2145, opened 2018-04-08
Fix Player DisplayName

hmy2001:
## Introduction
This patch is for fix to display of player display name.

### Relevant issues
none

### API changes
No

### Behavioural changes
No

## Backwards compatibility
Yes

## Follow-up
No

## Tests
A&B: spawn.
A: change display name.
A: move to away(B isn't loaded chunk).
B: move to A.


lukeeey:
Their username is their username... not display name.

PR #2151, opened 2018-04-14
Implementation of xp command

ahmedhozny:
## Introduction

Implement XP command as in vanilla 

### API changes

Added new class XPCommand which has the command basics
### Behavioural changes

N/A
## Backwards compatibility

N/A

Requires translations:

| Name | Value in eng.ini |
| `commands.xp.success.other` | `Gave {%0} to {%1)` |
| `commands.xp.success.self` | `Gave himself {%0}` |
| `pocketmine.command.xp.usage` | `/xp <experience[L]> <player>` |

-->

## Tests
I've tried the command with different situations (Player offline or not found, L letter adds level, Op can give to himself, etc..)


src/pocketmine/command/defaults/XPCommand.php
```diff
@@ -0,0 +1,90 @@
...
+                                ]));
+                        return true;
+                        }else{
+                                $sender->getServer()->getPlayer($sender->getName())->addXpLevels($xp);
+                                Command::broadcastCommandMessage($sender, new TranslationContainer("%commands.xp.success.self",[$xp]));
+                                return true;
+                        }
+                        
+                }elseif(!is_numeric($args[0])){
+                        $sender->sendMessage(TextFormat::RED. "Please write in numbers");
```
lukeeey: "Please write in numbers" doesn't sound like very good English.

src/pocketmine/command/defaults/XPCommand.php
```diff
@@ -0,0 +1,90 @@
...
+
+namespace pocketmine\command\defaults;
+
+use pocketmine\command\Command;
+use pocketmine\command\CommandSender;
+use pocketmine\command\utils\InvalidCommandSyntaxException;
+use pocketmine\lang\TranslationContainer;
+use pocketmine\utils\TextFormat;
+
+class XPCommand extends VanillaCommand{
```
lukeeey: I think this should be `XpCommand` instead of `XPCommand`

src/pocketmine/command/defaults/XPCommand.php
```diff
@@ -0,0 +1,90 @@
...
+	
+        public function __construct(string $name){
+	        parent::__construct(
+		        $name,
+		        "%pocketmine.command.xp.description",
+		        "%pocketmine.command.xp.usage"
+		);
+	$this->setPermission("pocketmine.command.xp");
```
lukeeey: Indentation throughout entire file

lukeeey (changes requested):

src/pocketmine/command/defaults/XPCommand.php
```diff
@@ -0,0 +1,90 @@
...
+	}
+        
+        public function execute(CommandSender $sender, string $commandLabel, array $args){
+	        if(!$this->testPermission($sender)){
+		        return true;
+	        }
+                
+                if(count($args) < 1){
+	                throw new InvalidCommandSyntaxException();
```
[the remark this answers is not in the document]
dktapps: Throwing an `InvalidCommandSyntaxException` makes the usage message be displayed. That's its reason for existence.

src/pocketmine/command/defaults/XPCommand.php
```diff
@@ -0,0 +1,90 @@
...
+                                Command::broadcastCommandMessage($sender, new TranslationContainer("%commands.xp.success.self",[$xp]));
+                                return true;
+                        }
+                        
+                }elseif(!is_numeric($args[0])){
+                        $sender->sendMessage(TextFormat::RED. "Please write in numbers");
+                        return true;
+                } 
+                if(isset($args[1])){
+                        $player->addXp(intval($args[0]));
```
dktapps: use a typecast, not `intval()`.

src/pocketmine/command/defaults/XPCommand.php
dktapps: This should be a generic command syntax error.

src/pocketmine/command/defaults/XPCommand.php
```diff
@@ -0,0 +1,90 @@
...
+                                        $xp,
+                                        $player->getName()
+                                ]));
+                        return true;
+                        }else{
+                                $sender->getServer()->getPlayer($sender->getName())->addXpLevels($xp);
+                                Command::broadcastCommandMessage($sender, new TranslationContainer("%commands.xp.success.self",[$xp]));
+                                return true;
+                        }
+
```
dktapps: The returns in both of the above blocks are redundant and make the `else` useless. They should be placed outside of it. Don't litter the function with lots of return points.

src/pocketmine/command/defaults/XPCommand.php
```diff
@@ -0,0 +1,90 @@
...
+                
+                if(strtolower(preg_replace("/[^a-zA-Z]/", "",$args[0])) == "l"){
+                        $xp = intval(preg_replace("/[^0-9]+/", "", $args[0]), 10);
+                        if(isset($args[1])){
+                                $player->addXpLevels($xp);
+                                Command::broadcastCommandMessage($sender, new TranslationContainer("%commands.xp.success", [
+                                        $xp,
+                                        $player->getName()
+                                ]));
+                        return true;
```
dktapps: ... formatting?

src/pocketmine/command/defaults/XPCommand.php
```diff
@@ -0,0 +1,90 @@
...
+                if(isset($args[1])){
+                        $player = $sender->getServer()->getPlayer($args[1]);
+                        if($player === null){
+	                        $sender->sendMessage(new TranslationContainer(TextFormat::RED . "%commands.generic.player.notFound"));
+	                        return true;
+                        }
+                }
+                
+                if(strtolower(preg_replace("/[^a-zA-Z]/", "",$args[0])) == "l"){
+                        $xp = intval(preg_replace("/[^0-9]+/", "", $args[0]), 10);
```
dktapps: This regex doesn't make sense. If I read correctly `5l0` will be interpreted as 50 levels. And again, use a typecast, not intval().

src/pocketmine/command/defaults/XPCommand.php
```diff
@@ -0,0 +1,90 @@
...
+                
+                if(isset($args[1])){
+                        $player = $sender->getServer()->getPlayer($args[1]);
+                        if($player === null){
+	                        $sender->sendMessage(new TranslationContainer(TextFormat::RED . "%commands.generic.player.notFound"));
+	                        return true;
+                        }
+                }
+                
+                if(strtolower(preg_replace("/[^a-zA-Z]/", "",$args[0])) == "l"){
```
dktapps: this should use strict equality check.

dktapps (changes requested):

src/pocketmine/command/defaults/XPCommand.php
```diff
@@ -0,0 +1,90 @@
...
+                        }
+                        
+                }elseif(!is_numeric($args[0])){
+                        $sender->sendMessage(TextFormat::RED. "Please write in numbers");
+                        return true;
+                } 
+                if(isset($args[1])){
+                        $player->addXp(intval($args[0]));
+                        Command::broadcastCommandMessage($sender, new TranslationContainer("commands.xp.success", [
+                                intval($args[0]),
```
dktapps: You have a tendecy to repeatedly call functions applying operations to the same thing. Apply it once, store it in a variable and use the variable.

src/pocketmine/command/defaults/XpCommand.php
```diff
@@ -0,0 +1,98 @@
...
+                        $lxp = $this->getInteger($sender, substr_replace($args[0], "", -1));
+                        if(isset($args[1])){
+                                $player->addXpLevels($lxp);
+                                Command::broadcastCommandMessage($sender, new TranslationContainer("%commands.xp.success.other", [
+                                        $lxp,
+                                        $player->getName()
+                                ]));
+                        }else{
+                                if(!($sender instanceof Player)){
+                                        throw new InvalidCommandSyntaxException;
```
dktapps: missing `()`

src/pocketmine/command/defaults/XpCommand.php
```diff
@@ -0,0 +1,98 @@
...
+                                $player->addXp($xp);
+                                Command::broadcastCommandMessage($sender, new TranslationContainer("%commands.xp.success.other", [
+                                        $xp,
+                                        $player->getName()
+                                ]));
+                        }else{
+                            if(!($sender instanceof Player)){
+                                throw new InvalidCommandSyntaxException;
+                            }
+                                $sender->getServer()->getPlayer($sender->getName())->addXp($xp);
```
dktapps: ... what in god's name is this?

src/pocketmine/command/defaults/XpCommand.php
```diff
@@ -0,0 +1,98 @@
...
+                                if(!($sender instanceof Player)){
+                                        throw new InvalidCommandSyntaxException;
+                                }
+                                $sender->getServer()->getPlayer($sender->getName())->addXpLevels($lxp);
+                                Command::broadcastCommandMessage($sender, new TranslationContainer("%commands.xp.success.self",[$lxp]));
+                        }
+                        return true;
+                }
+                
+                if(is_numeric($args[0])){
```
dktapps: This should be using `if..elseif`, not abusing return statements.

src/pocketmine/command/defaults/XpCommand.php
```diff
@@ -0,0 +1,98 @@
...
+                    $xp = $this->getInteger($sender, $args[0]);
+                        if(isset($args[1])){
+                                $player->addXp($xp);
+                                Command::broadcastCommandMessage($sender, new TranslationContainer("%commands.xp.success.other", [
+                                        $xp,
+                                        $player->getName()
+                                ]));
+                        }else{
+                            if(!($sender instanceof Player)){
+                                throw new InvalidCommandSyntaxException;
```
dktapps: nit: missing `()`

src/pocketmine/command/defaults/XpCommand.php
```diff
@@ -0,0 +1,98 @@
...
+                if(isset($args[1])){
+                        $player = $sender->getServer()->getPlayer($args[1]);
+                        if($player === null){
+	                        $sender->sendMessage(new TranslationContainer(TextFormat::RED . "%commands.generic.player.notFound"));
+	                        return true;
+                        }
+                }
+                
+                if(strtolower(substr($args[0], -1)) == "l"){
+                        $lxp = $this->getInteger($sender, substr_replace($args[0], "", -1));
```
dktapps: ... why not simply `substr($args[0], 0, -1)`

src/pocketmine/command/defaults/XpCommand.php
```diff
@@ -0,0 +1,98 @@
...
+                
+                if(isset($args[1])){
+                        $player = $sender->getServer()->getPlayer($args[1]);
+                        if($player === null){
+	                        $sender->sendMessage(new TranslationContainer(TextFormat::RED . "%commands.generic.player.notFound"));
+	                        return true;
+                        }
+                }
+                
+                if(strtolower(substr($args[0], -1)) == "l"){
```
dktapps: again, use strict comparison, not weak.

dktapps (changes requested):
I hope you don't expect this to be merged with such abominable formatting...

src/pocketmine/command/defaults/XpCommand.php
lukeeey: what does `nit` mean?

killer549: ## @dktapps Yeah and would you please specify the issue here

dktapps: are you capable of reading?

> missing `()`

dktapps: @lukeeey it means small issue in this context.

src/pocketmine/command/defaults/XpCommand.php
dschwartz783: This code actually made me laugh. xD

src/pocketmine/command/defaults/XpCommand.php
```diff
@@ -58,38 +58,30 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
                         }
                 }
                 
-                if(strtolower(substr($args[0], -1)) == "l"){
-                        $lxp = $this->getInteger($sender, substr_replace($args[0], "", -1));
+                if(strcasecmp(substr($args[0], -1), "l") == 0){
```
dktapps: Why change this?

killer549: I thought it would be a better comparison

dktapps: it's a completely unnecessary alteration.

src/pocketmine/command/defaults/XpCommand.php
```diff
@@ -58,38 +58,30 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
...
+                                        $player->addXp($xp);
+                                        Command::broadcastCommandMessage($sender, new TranslationContainer("commands.xp.success.other", [$xp, $player->getName()]));
+                                        
                         }else{
                             if(!($sender instanceof Player)){
-                                throw new InvalidCommandSyntaxException;
+                                throw new InvalidCommandSyntaxException();
                             }
-                                $sender->getServer()->getPlayer($sender->getName())->addXp($xp);
-                                Command::broadcastCommandMessage($sender, new TranslationContainer("%commands.xp.success.self",[$xp]));
```
lukeeey: Umm didnt someone say to keep the % there? It means it will be translated client side.

src/pocketmine/command/defaults/XpCommand.php
```diff
@@ -58,38 +58,30 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
...
-                if(is_numeric($args[0])){
-                    $xp = $this->getInteger($sender, $args[0]);
-                        if(isset($args[1])){
-                                $player->addXp($xp);
-                                Command::broadcastCommandMessage($sender, new TranslationContainer("%commands.xp.success.other", [
-                                        $xp,
-                                        $player->getName()
-                                ]));
+                }elseif(is_numeric($args[0])){
+                        $xp = $this->getInteger($sender, $args[0]);
```
lukeeey: Why not use a type cast?
```php
$xp = (int) $args[0];
```

lukeeey (changes requested):

dktapps (changes requested):

src/pocketmine/command/defaults/XpCommand.php
```diff
@@ -0,0 +1,90 @@
...
+			$player = $sender->getServer()->getPlayer($args[1]);
+			if($player === null){
+				$sender->sendMessage(new TranslationContainer(TextFormat::RED . "%commands.generic.player.notFound"));
+				return true;
+			}
+		}
+
+		if(strtolower(substr($args[0], -1)) == "l"){
+			$lxp = $this->getInteger($sender, substr($args[0], 0, -1));
+			if(isset($args[1])){
```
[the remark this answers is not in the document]
killer549: @dktapps Sorry for that but is that even possible to do without changing about quarter of the code ?😕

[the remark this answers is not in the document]
dschwartz783: If $args[1] is set, then earlier in your code you got a Player object to work with. Since you're not actually using $args[1] in the following code, but rather the Player object you created earlier, you should just check for that. Although, you could probably refactor this code a bit so that this entire block (as well as the other block where you check for isset($args[1]) appears in the same block where you initially created the Player object. It would reduce code duplication a bit.

dschwartz783:
You destroyed the formatting again. Tabs, not spaces.

killer549:
Not sure why am I using Netbeans

dschwartz783:
Use PHPStorm.

dktapps:
please read... https://github.com/pmmp/PocketMine-MP/blob/master/CONTRIBUTING.md#code-contributions

dschwartz783:
My point about using PHPStorm which I suppose I should have mentioned, is that PHPStorm will automatically handle some formatting issues, by dynamically switching to either tabs or spaces for indentation depending on the file. I don't believe NetBeans does this. The suggestion was not off-topic.

src/pocketmine/command/defaults/XpCommand.php
```diff
@@ -0,0 +1,92 @@
...
+			if(!($player instanceof Player)){
+				throw new InvalidCommandSyntaxException();
+			}
+			
+			$xp = (int) $args[0];
+			if($player !== $sender){
+				$player->addXp($xp);
+				Command::broadcastCommandMessage($sender, new TranslationContainer("%commands.xp.success.other", [$xp, $player->getName()]));
+			}else{
+				$player->addXp($xp);
```
Muqsit: you can move this just after line 80 and remove this and line 82... right?
```php
$player->addXp($xp);
if($player !== $sender){
    Command::broadc...
}else{
    Command::broadc...
}
```

killer549: Nice idea..
Thanks for your suggestion. 😃

robske110:
image
Still mixed tabs&spaces here

src/pocketmine/command/defaults/XpCommand.php
```diff
@@ -0,0 +1,90 @@
...
+				throw new InvalidCommandSyntaxException();
+			}
+			
+			$xp = (int) $args[0];
+			$player->addXp($xp);
+			if($player !== $sender){
+				Command::broadcastCommandMessage($sender, new TranslationContainer("%commands.xp.success.other", [$xp, $player->getName()]));
+			}else{
+				Command::broadcastCommandMessage($sender, new TranslationContainer("%commands.xp.success.self",[$xp]));
+			}
```
dktapps: code copy-paste - also, these messages should not be the same for both levels and XP points.

src/pocketmine/command/defaults/XpCommand.php
```diff
@@ -0,0 +1,90 @@
...
+				return true;
+			}
+		}
+
+		if(strtolower(substr($args[0], -1)) === "l"){
+			if(!($player instanceof Player)){
+				throw new InvalidCommandSyntaxException();			
+			}
+			
+			$lxp = (int) substr($args[0], 0, -1);
```
dktapps: the string after removing the `l` might not be numeric, in which case this will produce undefined behaviour.

src/pocketmine/command/defaults/XpCommand.php
```diff
@@ -0,0 +1,90 @@
...
+			$player->addXpLevels($lxp);
+			if($player !== $sender){
+				Command::broadcastCommandMessage($sender, new TranslationContainer("%commands.xp.success.other", [$lxp, $player->getName()]));
+			}else{
+				Command::broadcastCommandMessage($sender, new TranslationContainer("%commands.xp.success.self",[$lxp]));
```
dktapps: formatting issue here (and in the copy pasta block below)

src/pocketmine/command/defaults/XpCommand.php
```diff
@@ -0,0 +1,90 @@
...
+			if($player === null){
+				$sender->sendMessage(new TranslationContainer(TextFormat::RED . "%commands.generic.player.notFound"));
+				return true;
+			}
+		}
+
+		if(strtolower(substr($args[0], -1)) === "l"){
+			if(!($player instanceof Player)){
+				throw new InvalidCommandSyntaxException();			
+			}
```
dktapps: why duplicate this check in multiple branches? this should be outside of here, where you actually get the sender in the first place.

Additionally I don't think a command-syntax error is accurate in this case - it should be something like `Please specify a player` as per things like /teleport.

dktapps (changes requested):

src/pocketmine/command/defaults/XpCommand.php
lukeeey: Why not allow InvalidCommandSyntaxException to have a string parameter?

dktapps: That is outside the scope of this PR - and regardless, this isn't a command syntax error, it's console not being permitted to use a player-specific overload.

src/pocketmine/command/defaults/XpCommand.php
killer549: Is it better to remove `self` and `other` and just provide on message like this `"%commands.xp.success`?

src/pocketmine/command/defaults/XpCommand.php
```diff
@@ -68,13 +68,34 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
...
 				$level = (int) $check;
-				$player->addXpLevels($level);
-				Command::broadcastCommandMessage($sender, new TranslationContainer("%commands.xp.success.level", [$level, $player->getName()]));
+				if($level < 0){
+					if($player->getXpLevel() + $level < 0){
+						$sender->sendMessage(TextFormat::RED. "Players cannot have negative levels");
+						return true;
+					}
+					
+					$player->subtractXpLevels(-$level);
```
killer549: Not sure if using subtract here (and in the block below) is a great idea but I just didn't want to put `$player->addXpLevels($level);` at last after everything has done.

PR #2158, opened 2018-04-28
Allow command description and usage to depend on sender and time

SOF3:
## Introduction
Commands may want to have dynamic description and usage due to many reasons:
- Localization (although this can be better solved with a better Translations API)
- Sender type-dependent usage (e.g. console must provide a player argument for /tp, optional for players)
- Certain permission-dependent features (including disabling certain subcommands)

**Important note:** The usage will become obsolete in the future, as it shall be replaced by arg info, and it is possible to change the description by using a resource pack, but I feel that this is too heavy for a simple problem.

### API changes

- Added `Command::getDescriptionFor(CommandSender)` and `Command::getUsageFor(CommandSender)`, which redirect to `Command::getDescription()` and `Command::getUsage()` respectively in their implementation.

### Behavioural changes

- `/help` uses description and usage from the new methods
- Throwing `InvalidCommandSyntaxException` will show a message from the `getUsageFor` instead of `getUsage`
- In the client, the command description message will be the one generated from `Command::getDescriptionFor` **when the player logins, just after PlayerLoginEvent** and when the player's permissions are recalculated.
  - This will not change when the client changes its language in-game (although really rare).

## Backwards compatibility

Almost fully backwards-comptaible, unless a Command subclass defines getDescriptionFor()/getUsageFor(), which is highly unlikely.

## Follow-up

There are some follow-ups we can do, but they will surely all be rejected because of the never-coming command argss API 😶 

## Tests

Tested with a plugin overriding getDescriptionFor() in a command class and running `/help` on console.

### DIFF
--- a/src/pocketmine/command/Command.php
+++ b/src/pocketmine/command/Command.php
@@ -239,6 +239,8 @@ abstract class Command{
 	}
 
 	/**
+	 * Returns the generic description message of this command.
+	 *
 	 * @return string
 	 */
 	public function getDescription() : string{
@@ -246,10 +248,32 @@ abstract class Command{
 	}
 
 	/**
+	 * Returns the specific description message of this command for the CommandSender at this moment.
+	 *
+	 * @param CommandSender $sender
+	 * @return string
+	 */
+	public function getDescriptionFor(CommandSender $sender) : string{
+		return $this->getDescription();
+	}
+
+	/**
+	 * Returns the generic usage message of this command.
+	 *
 	 * @return string
 	 */
 	public function getUsage() : string{
 		return $this->usageMessage;
+	}
+
+	/**
+	 * Returns the specific usage message of this command for the CommandSender at this moment.
+	 *
+	 * @param CommandSender $sender
+	 * @return string
+	 */
+	public function getUsageFor(CommandSender $sender) : string{
+		return $this->getUsage();
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


src/pocketmine/command/Command.php
```diff
@@ -239,19 +239,43 @@ public function getPermissionMessage() : ?string{
...
 	}
 
+	/**
+	 * Returns the specific usage message of this command for the CommandSender at this moment.
+	 *
+	 * @param CommandSender $sender
+	 * @return string
+	 */
+	public function getUsageFor(CommandSender $sender) : string{
+		return $this->getUsage();
```
dktapps: Hardcoded usage messages will be removed at some point in the future. I'm not entirely happy with adding API that will inevitably get removed later on.

src/pocketmine/command/Command.php
```diff
@@ -239,19 +239,43 @@ public function getPermissionMessage() : ?string{
...
+	 * Returns the generic description message of this command.
+	 *
 	 * @return string
 	 */
 	public function getDescription() : string{
 		return $this->description;
 	}
 
 	/**
+	 * Returns the specific description message of this command for the CommandSender at this moment.
```
dktapps: I think this is misleading. This doesn't apply for players - they'll see the message that was there when the command data was sent (i.e. during login).

dktapps:
Closing this due to lack of activity.

PR #2162, opened 2018-04-29
Skip method if Listener declares method with undefined argument type

SOF3:
## Introduction

Some plugins may declare a soft dependency to other plugins, i.e. the plugin can still load without the other plugin. These plugins may want to listen to events from these other plugins if they exist, or just ignore the event if they don't exist.

Currently, plugins have to call `PluginManager->registerEvent()` directly, or separate the Listener class, in order not to trigger the error of trying to listen to a non-existent event:

```
[20:31:57] [Server thread/CRITICAL]: ReflectionException: "Class no\such\event does not exist" (EXCEPTION) in "src/pocketmine/plugin/PluginManager" at line 755
```

This pull request allows a Listener class to declare methods requiring parameters of non-existent class types by declaring `@softDepend` in its doccomment. This makes listening events from plugins pssively more convenient.

This is mainly useful for listening to optional events that may exist or not exist. This should not be relied on, say, to execute something when a player is authenticated by a certain auth plugin, because if multiple auth plugins are loaded (for no proper reason though), the upon-auth execution will execute multiple times.

### API changes

Added the `@softDepend` tag to skip Listener methods with undefined class parameters.

The methods should only be skipped when the soft-depend plugin is not loaded; if it is loaded, but the event is still not found, this is most likely a typo in the event class, or an API change in the dependent plugin. In this case, we still want to display the error message, so `@softDepend` can accept a parameter, the plugin name, which will be made sure not to be loaded.

## Backwards compatibility

Fully backward-compatible, as `@softDepend` is not a standard doc comment expected to be found in event listener methods.

## Tests

Tested with a Listener with an undefined parameter, both with and without `@softDepend`, both working as expected.

### DIFF
--- a/src/pocketmine/plugin/PluginManager.php
+++ b/src/pocketmine/plugin/PluginManager.php
@@ -752,7 +752,16 @@ class PluginManager{
 				$ignoreCancelled = isset($tags["ignoreCancelled"]) && strtolower($tags["ignoreCancelled"]) === "true";
 
 				$parameters = $method->getParameters();
-				if(count($parameters) === 1 and $parameters[0]->getClass() instanceof \ReflectionClass and is_subclass_of($parameters[0]->getClass()->getName(), Event::class)){
+				try{
+					$isHandler = count($parameters) === 1 && $parameters[0]->getClass() instanceof \ReflectionClass && is_subclass_of($parameters[0]->getClass()->getName(), Event::class);
+				}catch(\ReflectionException $e){
+					if(isset($tags["softDepend"])){
+						continue;
+					}
+
+					throw new \ClassNotFoundException($e->getMessage(), 0, $e);
+				}
+				if($isHandler){
 					$class = $parameters[0]->getClass()->getName();
 					$this->registerEvent($class, $listener, $priority, new MethodEventExecutor($method->getName()), $plugin, $ignoreCancelled);
 				}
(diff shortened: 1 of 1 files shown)


src/pocketmine/plugin/PluginManager.php
```diff
@@ -752,7 +752,16 @@ public function registerEvents(Listener $listener, Plugin $plugin) : void{
...
 				$parameters = $method->getParameters();
-				if(count($parameters) === 1 and $parameters[0]->getClass() instanceof \ReflectionClass and is_subclass_of($parameters[0]->getClass()->getName(), Event::class)){
+				try{
+					$isHandler = count($parameters) === 1 && $parameters[0]->getClass() instanceof \ReflectionClass && is_subclass_of($parameters[0]->getClass()->getName(), Event::class);
+				}catch(\ReflectionException $e){
+					if(isset($tags["softDepend"])){
+						continue;
+					}
+
+					throw new \ClassNotFoundException($e->getMessage(), 0, $e);
```
dktapps: I would say it's better to just rethrow `$e` here (as per the original behaviour). `ClassNotFoundException` is not expected to be thrown by this function and is not used anywhere else (it remains solely for backwards compatibility and will be removed in future).

dktapps (approved):

PR #2164, opened 2018-05-01
Implemented @notHandler annotation for event handlers - skip register…

dktapps:
…ing any handlers with this annotation

It is somewhat reasonable to have a function in an event handler which accepts an Event parameter, but is not a handler. For example, multiple event handlers can redirect to the same function to process an event, but this function may not want to receive called events.

There are other ways to get around this, such as making the event handler protected/private, or adding a dummy parameter, but this way is cleaner and more explicit.

Relevant old-repo PR: PocketMine/PocketMine-MP#2143


SOF3 (commented):
Looks OK if nobody from Spain still wants to reject this.

dktapps:
Since this takes literally 3 lines of code to implement, I don't think it makes sense to fuss about it.

PR #2171, opened 2018-05-05
Allow threads to signal the server to wake it up during tick sleeps

dktapps:
## Introduction
This pull request changes how server tick sleeping works in order to permit signaling interrupts.

Currently the server hard-sleeps between ticks. This can be problematic for threaded things such as network processing because it adds extra latency. For example, in RakLib:

- RakLib pushes a packet into the main thread processing queue
- At the same time, the server goes to sleep for 49ms
- Packet doesn't get processed until at least 49ms later, adding response latency.

This used to be very apparent when older versions of RakLib used to tick at 20 TPS, which could cause up to 100ms accumulated latency _just for receiving packets_ (server max sleep time + RakLib max sleep time), which is obviously extremely sub optimal. With newer versions of RakLib (which tick at 100 TPS) this is reduced to up to 60ms (server tick 50ms + RakLib tick 10ms), but still very sub-optimal.

Ideally these packets should be processed as soon as they are received. This PR adds a mechanism which allows other threads to wake the server up from sleep to process actions. This allows packet processing latency to be reduced to 10ms if a packet arrives during server sleep time. Further changes to RakLib will allow this to be reduced to near-zero (RakLib still uses a sub-optimal socket polling mechanism).

This mechanism has been implemented with RakLib to demonstrate the use of this pull request. This could be extended to a range of other things to improve performance.

### API changes
- Added classes `TickSleeper` and `SleeperNotifier`
- Added method `Server->getTickSleeper()`

### Behavioural changes
- RakLibServer is now able to immediately wake the server from sleep on packet receive, which improves response time visibly. However, this mechanism is currently a little buggy and needs more work.

## Follow-up
Extend the use of this mechanism to the following things:
- `RCON` command sending (change to blocking architecture)
- `CommandReader` can make use of this to remove the need for polling the thread.

This pull request was created with network thread pools in mind. I intend to add dedicated thread pools for network processing so that compression/decompression can use the whole CPU instead of hogging the main thread. Obviously a polling mechanism for this would be extremely sub optimal as it would add even more latency to a system with too much processing latency already.

## Tests
Has been tested under low and high-load conditions, but there are known bugs with this mechanism currently - further testing is needed.


dktapps:
Notifications will not get processed if a tick takes longer than 50ms to execute.

dktapps:
With the change to make this use closure notification handlers, the library parts of this code could be moved to PocketMine-SPL to allow RakLib to utilize it as well.

PR #2179, opened 2018-05-11
Player: changeSkin() remove unnecessary parameters.

KennFatt:
## Introduction

Since changeSkin() is never used `$newSkinName` and `$oldSkinName` in it's method and Skin class already handled everything for it.

### Relevant issues
* Those parameters are make developers confused and actually mean nothing.

### API changes
Important changes:
Player: changeSkin(Skin $skin, string $newSkinName, string $oldSkinName) to changeSkin(Skin $skin).

### Behavioural changes
There is no effect with server behaves.

## Tests
Already tested with change skin in-game with any types of skin.


dktapps:
I don't see how they confuse developers - this isn't an API method. These were here for some as-yet incomplete work.

PR #2181, opened 2018-05-16
Change global variable to local

rdstonech:
## Introduction
Small change in RaklibInterface

### API changes
No API changes

### Behavioural changes
Perfomance


dktapps:
this change is pointless and unnecessary.

PR #2107, opened 2018-03-16
Implemented fire spread

dktapps:
## Introduction
This PR (re)implements fire handling to produce behaviour matching vanilla standards (or as close an approximation as possible).

### Relevant issues
- fixes #1424 
- fixes #1779 

### API changes
- added API methods:
  - `Block->getFlameEncouragement()`
  - `Block->getFlammability()`
  - `Block->isFlammable()`
  - `Block->burnsInfinitely()` (used by netherrack)
- fire spread can be prevented by cancelling `BlockSpreadEvent`
- block burning away can be prevented by cancelling `BlockBurnEvent`

### Behavioural changes
- Fire now spreads and burns blocks as in vanilla Minecraft
- TNT is ignited by fire
- Flint and steel now work as they do in vanilla
- Fixes the following bugs:
  - fire can be placed in strange locations (#1424 )
  - fire doesn't extinguish in a reasonable space of time (#1779 )
  - can't use flint-steel on things like leaves

## Tests
Tests are still being done. Some things have been noted that I am not entirely sure on the cause of (fire doesn't do as much damage here as vanilla, and similar). Help testing would be appreciated.

PR #2189, opened 2018-05-18
Remove implicit chunk autoloading

dktapps:
## Introduction
For a long time it's been problematic that `Level->getChunk()` automatically loads chunks unasked. This has caused a wide range of difficult-to-find bugs, especially since there is no way to get a chunk _without_ loading it, without boilerplate `isChunkLoaded()` checks everywhere.

Issues stemming from this include:
- The good old entity leak of 2016 - if implicit chunk loading didn't exist, it would never have happened, because the server would have crashed instead
- Entities running off the map will load chunks, which is entirely unexpected behaviour and could cause problems with rogue AI in the future.
- Light updates trigger autoloading of chunks, which is the primary blocker for #2107, because they cause CPU and memory leaks.

Another big reason for this change is that it won't be possible to implement asynchronous chunk I/O without it. As we know there's already #1895 open for months, which has stagnated because there are far too many changes in it and has many bugs. This pull request's changes have been separated from #1895 so that they can be integrated separately.

This pull request changes the core to prefer explicit chunk loading **where appropriate** (such as for Players).

### API changes
- `Level->getChunk()` DOES NOT load chunks anymore and has had the third parameter removed.
- Added `Level->getLoadedChunk()` which will always either return an existing chunk or throw a `RuntimeException`.
- `Level->get*At()` functions (and their corresponding setters) will not automatically sync-load chunks anymore - instead they will throw exceptions.

It is expected that these changes will _not_ affect plugins for the most part, however some world edit plugins might be affected. It is expected that plugins should utilize the [`ChunkLoader`](https://github.com/pmmp/PocketMine-MP/blob/4f8e4f05220ecd56b038204009fbfdfe4a6359e4/src/pocketmine/level/ChunkLoader.php) to keep chunks loaded that they need to operate on.

For smoothness of upgrading, **please remember that chunk loading will become asynchronous in future**, and plan appropriately.

### Behavioural changes
Implicit chunk autoloading is now removed, fixing lots of bugs.

## Backwards compatibility
Noted above in the API changes section.

## Tests
Brief tests have been performed, more extensive testing is needed.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -998,7 +998,7 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 			++$count;
 
 			$this->usedChunks[$index] = false;
-			$this->level->registerChunkLoader($this, $X, $Z, false);
+			$this->level->registerChunkLoader($this, $X, $Z, true);
 
 			if(!$this->level->populateChunk($X, $Z)){
 				continue;
--- a/src/pocketmine/entity/Entity.php
+++ b/src/pocketmine/entity/Entity.php
@@ -520,10 +520,7 @@ abstract class Entity extends Location implements Metadatable, EntityIds{
 		$this->boundingBox = new AxisAlignedBB(0, 0, 0, 0, 0, 0);
 		$this->recalculateBoundingBox();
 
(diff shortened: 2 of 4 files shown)


dktapps:
Known bugs so far:
- Light updates have problems with getting adjacent light and crash.

PR #2191, opened 2018-05-19
Players can now swim correctly 

moskadev:
## Introduction
Added a new player action : Swimming with its respective event.

### API changes
* Added `PlayerToggleSwimEvent`
* Added Entity::`isSwimming()` and `setSwimming()`

## Tests
Tested. Everything seems to work.


dktapps (commented):
Nice, looks simple and straight to the point - this will allow plugin devs to do elytra things without hacks as well while the core doesn't have support for them.

src/pocketmine/Player.php
```diff
@@ -2780,18 +2782,46 @@ public function handlePlayerAction(PlayerActionPacket $packet) : bool{
 				}
 				return true;
 			case PlayerActionPacket::ACTION_START_GLIDE:
+				$ev = new PlayerToggleGlideEvent($this, true);
+				$this->server->getPluginManager()->callEvent($ev);
+				if($ev->isCancelled()){
+					$this->sendData($this);
+				}else{
+					$this->setGliding(true);
```
dktapps: I'm unsure if allowing gliding should be the default behaviour here since it isn't implemented yet. Maybe this should be cancelled by default (until elytra are implemented into the core).

src/pocketmine/Player.php
```diff
@@ -2780,18 +2782,46 @@ public function handlePlayerAction(PlayerActionPacket $packet) : bool{
...
+				}
+				return true;
 			case PlayerActionPacket::ACTION_CONTINUE_BREAK:
 				$block = $this->level->getBlock($pos);
 				$this->level->broadcastLevelEvent($pos, LevelEventPacket::EVENT_PARTICLE_PUNCH_BLOCK, BlockFactory::toStaticRuntimeId($block->getId(), $block->getDamage()) | ($packet->face << 24));
 				//TODO: destroy-progress level event
 				break;
 			case PlayerActionPacket::ACTION_START_SWIMMING:
-				break; //TODO
+				$ev = new PlayerToggleSwimEvent($this, true);
```
dktapps: shouldn't we check if they are actually in water?

src/pocketmine/Player.php
```diff
@@ -2782,34 +2783,22 @@ public function handlePlayerAction(PlayerActionPacket $packet) : bool{
...
 				$this->level->broadcastLevelEvent($pos, LevelEventPacket::EVENT_PARTICLE_PUNCH_BLOCK, BlockFactory::toStaticRuntimeId($block->getId(), $block->getDamage()) | ($packet->face << 24));
 				//TODO: destroy-progress level event
 				break;
 			case PlayerActionPacket::ACTION_START_SWIMMING:
+				$block = $this->level->getBlock($this);
+
 				$ev = new PlayerToggleSwimEvent($this, true);
 				$this->server->getPluginManager()->callEvent($ev);
-				if($ev->isCancelled()){
+				if(!$block instanceof Water or $ev->isCancelled()){
```
dktapps: you'll probably want to put brackets around the instanceof here

src/pocketmine/Player.php
```diff
@@ -64,8 +65,10 @@
 use pocketmine\event\player\PlayerQuitEvent;
 use pocketmine\event\player\PlayerRespawnEvent;
 use pocketmine\event\player\PlayerToggleFlightEvent;
+use pocketmine\event\player\PlayerToggleGlideEvent;
```
dktapps: this class doesn't exist

src/pocketmine/Player.php
```diff
@@ -2788,10 +2791,27 @@ public function handlePlayerAction(PlayerActionPacket $packet) : bool{
 				//TODO: destroy-progress level event
 				break;
 			case PlayerActionPacket::ACTION_START_SWIMMING:
-				break; //TODO
+				$block = $this->level->getBlock($this);
+
+				$ev = new PlayerToggleSwimEvent($this, true);
+				$this->server->getPluginManager()->callEvent($ev);
+				if(!$block instanceof Water or $ev->isCancelled()){
```
dktapps: I think it's better to cancel the event before calling it if the player is not in water, to give plugins a chance to alter the outcome.

Muqsit:
Out of curiosity since no one has pointed it out, is there a reason why we aren't using `Entity::isInsideOfWater()` here?

supermaxalex:
@Muqsit It checks if the entity is under water and not on the surface

Muqsit:
Ah, right. 👍

dktapps (changes requested):

supermaxalex:
@dktapps ^

src/pocketmine/Player.php
```diff
@@ -2788,10 +2790,35 @@ public function handlePlayerAction(PlayerActionPacket $packet) : bool{
 				//TODO: destroy-progress level event
 				break;
 			case PlayerActionPacket::ACTION_START_SWIMMING:
-				break; //TODO
+				$block = $this->level->getBlock($this);
+
+				$ev = new PlayerToggleSwimEvent($this, true);
+				if(!($block instanceof Water)){
```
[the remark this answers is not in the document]
SalmonDE: Not in singleplayer

src/pocketmine/Player.php
```diff
@@ -2788,10 +2790,35 @@ public function handlePlayerAction(PlayerActionPacket $packet) : bool{
...
+					$ev->setCancelled();
+				}
+				$this->server->getPluginManager()->callEvent($ev);
+				if($ev->isCancelled()){
+					$this->sendData($this);
+				}else{
+					$this->propertyManager->setFloat(self::DATA_BOUNDING_BOX_HEIGHT, $this->width);
+					$this->height = $this->width;
```
Muqsit: I think you need to swap the height and width too (`$this->width = $this->height;`).

dktapps: no, it's correct like this. same for elytra, the BB becomes 0.6x0.6.

src/pocketmine/Player.php
```diff
@@ -2788,10 +2790,35 @@ public function handlePlayerAction(PlayerActionPacket $packet) : bool{
...
+
+				$ev = new PlayerToggleSwimEvent($this, true);
+				if(!($block instanceof Water)){
+					$ev->setCancelled();
+				}
+				$this->server->getPluginManager()->callEvent($ev);
+				if($ev->isCancelled()){
+					$this->sendData($this);
+				}else{
+					$this->propertyManager->setFloat(self::DATA_BOUNDING_BOX_HEIGHT, $this->width);
```
dktapps: this will produce astonishing behaviour if a plugin modifies the player's scale (this data property is supposed to be the base height, not rescaled).

src/pocketmine/event/player/PlayerToggleSwimEvent.php
```diff
@@ -0,0 +1,49 @@
...
+		$this->isSwimming = $isSwimming;
+	}
+
+	/**
+	 * @return bool
+	 */
+	public function isSwimming() : bool{
+		return $this->isSwimming;
+	}
+
```
dktapps: unnecessary newline

dktapps (commented):
I would prefer that the logic for toggling swimming was moved to a dedicated method (see https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/Player.php#L2782 for example).

src/pocketmine/Player.php
```diff
@@ -2788,10 +2790,30 @@ public function handlePlayerAction(PlayerActionPacket $packet) : bool{
...
+
+				$ev = new PlayerToggleSwimEvent($this, true);
+				$this->server->getPluginManager()->callEvent($ev);
+				if(!($block instanceof Water)){
+					$ev->setCancelled();
+				}
+				if($ev->isCancelled()){
+					$this->sendData($this);
+				}else{
+					$this->propertyManager->setFloat(self::DATA_BOUNDING_BOX_HEIGHT, $this->width);
```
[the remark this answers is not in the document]
lukeeey: @dktapps If you forget anti cheat, would this code work to swim through 1 block tunnels?

PR #2198, opened 2018-05-20
Item: quick implementation for Totem

ahmedhozny:
## Introduction

This PR implements Totem of undying as in vanilla.
### API changes

Added protected function `consumeTotem()` in `pocketmine\entity\Human` Added `PlayerConsumeTotemEvent`.
Added a separated block that checks whether who died is a human or something else
### Behavioural changes

N/A
## Backwards compatibility

N/A
## Follow-up

Something is wrong in showing the totem animation. The green particles do appear, but the totem that appear on the screen when used doesn't.

## Tests

Tested from all aspects:
1- When using `/kill` or reason of death is void.
2- Tried the PlayerConsumeTotemEvent and worked fine .


src/pocketmine/entity/Entity.php
```diff
@@ -954,6 +958,15 @@ public function setHealth(float $amount) : void{
 
 		if($amount <= 0){
 			if($this->isAlive()){
+				if($this instanceof Player and $this->spawned === true and $this->hasTotemInHand()){
```
killer549: `$this->spawned` check is here to avoid getting Item in hand on null

JackNoordhuis: === true isn't needed

src/pocketmine/Player.php
```diff
@@ -3647,6 +3647,24 @@ protected function onDeathUpdate(int $tickDiff) : bool{
 		return false; //never flag players for despawn
 	}
 
+	protected function onTotem() : void{
```
JackNoordhuis: This method name doesn't make sense -- it contains logic for actually using the totem but the name suggests it contains logic for handling the event. Perhaps `useTotem()` or `consumeTotem()` would be better. 

Also, does this method belong here? Maybe it would be better in the `pocketmine\entity\Human` class.

src/pocketmine/event/player/PlayerTotemEvent.php
```diff
@@ -0,0 +1,34 @@
...
+*/
+
+declare(strict_types=1);
+
+namespace pocketmine\event\player;
+
+use pocketmine\event\Cancellable;
+use pocketmine\Player;
+
+class PlayerTotemEvent extends PlayerEvent implements Cancellable{
```
JackNoordhuis: Same deal with a non-descriptive name, `PlayerTotemEvent` doesn't describe the actual event and sounds more like an abstract class that other events should extend.

src/pocketmine/Player.php
killer549: > Also, does this method belong here? Maybe it would be better in the pocketmine\entity\Human class.

I thought that methods like this are written in `pocketmine\Player`

src/pocketmine/Player.php
```diff
@@ -3647,6 +3648,24 @@ protected function onDeathUpdate(int $tickDiff) : bool{
 		return false; //never flag players for despawn
 	}
 
+	protected function consumeTotem() : void{
```
Muqsit: You can move this function to Human class for code extendability.

src/pocketmine/Player.php
```diff
@@ -3647,6 +3648,24 @@ protected function onDeathUpdate(int $tickDiff) : bool{
...
+		$this->server->getPluginManager()->callEvent($ev = new PlayerConsumeTotemEvent($this));
+		if($ev->isCancelled()){
+			$this->kill();
+			return;
+		}
+
+		$this->inventory->setItemInHand(Item::get(Item::AIR));
```
Muqsit: Any explanation for this? And why not use `Inventory::clear()`?

src/pocketmine/entity/Entity.php
```diff
@@ -954,6 +958,15 @@ public function setHealth(float $amount) : void{
 
 		if($amount <= 0){
 			if($this->isAlive()){
+				if($this instanceof Player and $this->spawned and $this->hasTotemInHand()){
```
Muqsit: `$this instanceof Player` makes declaring `Entity::consumeTotem()` useless. Maybe move this check into Human class? (See comment above)

dktapps:
> quick implementation

please don't ... if you're going to do a job, do it properly, otherwise you're wasting your time.

dktapps:
please also read the [CONTRIBUTING](https://github.com/pmmp/PocketMine-MP/blob/master/CONTRIBUTING.md#code-contributions) guidelines - particularly the top bullet point.

src/pocketmine/entity/Human.php
```diff
@@ -217,6 +217,9 @@ public function addFood(float $amount) : void{
 	public function isHungry() : bool{
 		return $this->getFood() < $this->getMaxFood();
 	}
+	public function hasTotemInHand() : bool{
```
Muqsit: What's the reason for this method's existence? ._.

PR #2213, opened 2018-05-30
Scheduler nuke

dktapps:
## Introduction
Individual commits descriptions explain what is changing here. 

### API changes
- Removed `Server->getScheduler()`. All plugins now have their own scheduler which is accessible using `Plugin->getScheduler()`. Aside from being syntactically more concise and pleasant, this also allows much more effective management of tasks when plugins are disabled.
- Removed `PluginTask` class. Before this PR it was necessary for plugin tasks to descend from `PluginTask` to ensure that the server could clean them up correctly on plugin disable. This is no longer necessary, so the `PluginTask` class has been removed. Plugins may now utilize the `Task` class as a base if they like.
- Added `Server->getAsyncPool()`. Since the global scheduler does not exist any more, it does not manage the server's `AsyncPool` any more. Additionally, `ServerScheduler` was previously bloated by a lot of `AsyncTask` related methods, which are now not necessary because direct access to `AsyncPool` is granted instead.

- `ServerScheduler`:
  - `ServerScheduler` has been renamed to `TaskScheduler` since it is now a general-purpose task scheduler which is non-dependent on the user. This allows much greater flexibility and also makes it possible to unit-test.
  - All `AsyncTask`/`AsyncPool` related methods have been removed - the task scheduler does not manage the async pool anymore.
    - Calls to `Server->getScheduler()->scheduleAsyncTask()` should be replaced with `Server->getAsyncPool()->submitTask()`.
    - Calls to `Server->getScheduler()->scheduleAsyncTaskToWorker()` should be replaced with and `Server->getAsyncPool()->submitTaskToWorker()`.

## Backwards compatibility
This poses significant backwards compatibility breaks for any plugins utilizing Tasks or AsyncTasks. These breaks are described above, along with basic upgrade steps. The upgrade process is quite straightforward.

## Follow-up
A large part of the goal with this pull request is to modularize these parts of the code so that they can be reused and also unit-tested. I would like to remove the existing test set from TesterPlugin at some stage when the AsyncPool can operate without a Server.

Because of the above, I am considering making further backwards incompatible changes directly to `AsyncTask` to remove the `Server` parameters from `onCompletion()` and `onProgressUpdate()`. These shouldn't be too difficult to upgrade from and can be prepared for in advance.

## Tests
All changes have been tested (though not heavily) and the TesterPlugin submodule has been updated to reflect the necessary changes.


fuyutsuki:
+1!

zBearchenPlayz:
WHY!

zBearchenPlayz:
WHY TASK OUT?!!!!

dktapps:
Looks like someone didn't read any further than the title...

@zBearchenPlayz please don't spam the discussion, you're upsetting the bot. If you make it too angry it might ban you from the repo, and what a tragic loss that would be ;-;

I'm sure you can contribute more constructively by actually reading the OP first, and then make rational comments once you've done that.

xXNiceYT:
@dktapps how do u use it

dktapps:
Currently known bugs:
- Disabling and re-enabling a plugin will cause any tasks scheduled by the plugin after re-enable to throw exceptions.

Other things that aren't noted in the OP:
- Plugins can still schedule tasks that won't get deleted by using another plugin's scheduler. However, it's also possible to fool the server by doing something like this with PluginTask, so I don't consider this to be a real concern - it's only something that a malicious plugin would do, and normal plugins are incentivized to use their own schedulers because it's syntactically more pleasant to do so.

zBearchenPlayz:
I DONT Spam!ITS really Bad To delete PluginTask,New API is OK BUT NOT PluginTask

dktapps:
So far I'm only seeing spam and yelling "pls don't remove", reasons please.

TheAz928:
Is it possible to create a custom task scheduler (not under a plugin) with this pr?

TheAz928:
The scheduler must be functional also, ik it's possible to create class object

dktapps:
@TheAz928 yes, but it relies on something to externally pulse it each tick.

dktapps:
having said that, you could pulse it at whatever frequency you wanted ¯\_(ツ)_/¯

TheAz928:
Interesting 👍

dktapps:
This is a big blocker for several other things I'd like to get done. If I could get reviews on this ASAP that would be great. (cc @SOF3 )

dktapps:
Merging this without reviews because it's blocking progress. I'm sure the bugs will only start popping out of the woodwork after it's merged anyway.

Virvolta:
how do we get the task information? 
example $quitTask = new QuitTask(Duel::getInstance(), $player);

PR #2216, opened 2018-05-31
Adding Player::getDeviceOS() function

iiFlamiinBlaze:
## Introduction
This PR adds a new function to the player class, Player::getDeviceOS().

### API changes
 * Adds Player::getDeviceOS()

## Tests
Tested with a clean PMMP server and a test plugin, that sent a message when the player joins with the deviceOS number, to test if the deviceOS function worked.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -166,6 +166,13 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 	public const ADVENTURE = 2;
 	public const SPECTATOR = 3;
 	public const VIEW = Player::SPECTATOR;
+	public const OS_ANDROID = 1;
+	public const OS_IOS = 2;
+	public const OS_MAC = 3;
+	public const OS_FIRE = 4;
+	public const OS_GEARVR = 5;
+	public const OS_HOLOLENS = 6;
+	public const OS_WINDOWS = 7;
 
 	/**
 	 * Checks a supplied username and checks it is valid.
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -3974,6 +3974,6 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 	}
 
 	public function getDeviceOS() : int{
-	    return $this->deviceOS;
-    }
+		return $this->deviceOS;
+	}
 }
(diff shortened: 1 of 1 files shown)


src/pocketmine/Player.php
```diff
@@ -3960,4 +3965,8 @@ public function getLoaderId() : int{
 	public function isLoaderActive() : bool{
 		return $this->isConnected();
 	}
+
+	public function getDeviceOS() : int{
+	    return $this->deviceOS;
```
supermaxalex: formatting

src/pocketmine/Player.php
```diff
@@ -234,6 +234,9 @@ public static function isValidUserName(?string $name) : bool{
 	/** @var string */
 	protected $xuid = "";
 
+	/** @var int */
```
supermaxalex: Use `$deviceOs` instead of `$deviceOS`

moskadev (changes requested):

dktapps (commented):
I'm hesitant to merge this because it's going to conflict with changes I've made on `network-nuke`. I'm working on things like this on that branch (making login data more accessible to plugins).

I don't think this is particularly useful on its own, especially not without constants for the OS types. Without constants it's simply a magic number.

src/pocketmine/Player.php
dktapps: In my opinion this isn't necessary. `deviceOs` looks odd and unintuitive imho.

supermaxalex: But it better fits how variables are called in PMMP.

src/pocketmine/Player.php
iiFlamiinBlaze: The formatting is fine...

src/pocketmine/Player.php
iiFlamiinBlaze: `$deviceOS` is better since the function is named `getDeviceOS()`

src/pocketmine/Player.php
dktapps: no, it's not. this line and the line below have incorrect indentation. It should use tabs, not spaces.

MisteFr:
It might be a good idea to implement all the device stuff in a same PR.
You can get the device model too in the clientData part of the LoginPacket (clientData["DeviceModel"]).

PR #2235, opened 2018-06-12
Added a pocketmine.yml kill switch for development builds

dktapps:
## Introduction
It's very tiresome to keep seeing new users and noobs keep using dev builds when we keep telling them not to for the safety of their data, their plugins and their gameplay. Then they come to us asking why X plugin doesn't work or why the server crashes or whatever.

This pull requests prevents development builds from running out of the box, adding an obstacle to running development builds which should hopefully reduce the amount of turbulence new confused users get hit with when wanting to set up a new server.

Beyond a doubt, those who think they are smart to run dev builds in production today will simply disable this and continue to be morons, but for new users it needs to be made apparent at first use that they are using something unsafe which may have undesirable effects on their server. This is hoped to discourage new users from running unsafe development builds.

### Relevant issues
Too many to count, but not github issues.

### API changes
- A new constant, `\pocketmine\IS_DEVELOPMENT_BUILD` has been added. If this constant is true, then the suicide rules apply. This constant should be set to false prior to tagging a release, and then reverted once the release has been created.

### Behavioural changes
- Development builds will now commit suicide unless the `settings.enable-dev-builds` directive is enabled in pocketmine.yml (or through command-line overrides).

## Follow-up
- Translation of the new strings into other languages (which should take place on Crowdin).
- Jenkins configuration needs to be updated to make sure that it can still build development builds.

## Tests
Tested under the following circumstances:
- Switch on, off and not present in pocketmine.yml
- Command-line override using `--settings.enable-dev-builds=1`

### DIFF
--- a/src/pocketmine/PocketMine.php
+++ b/src/pocketmine/PocketMine.php
@@ -81,6 +81,7 @@ namespace pocketmine {
 	const NAME = "PocketMine-MP";
 	const VERSION = "1.7dev";
 	const API_VERSION = "3.0.0";
+	const IS_DEVELOPMENT_BUILD = true;
 
 	const MIN_PHP_VERSION = "7.2.0";
 
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -1452,6 +1452,19 @@ class Server{
 
 			define('pocketmine\DEBUG', (int) $this->getProperty("debug.level", 1));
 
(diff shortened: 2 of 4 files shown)


SOF3 (approved):

dktapps:
Squash-merged in b5da6b15912ad41f5d1593c43b1bf2f85d6e75b0.

PR #2257, opened 2018-06-22
Implemented Mending enchantment

dktapps:
## Introduction
Implementing Mending enchantment. For those who don't know, this enchantment causes equipped items with the enchantment to become repaired by 2 points for every point of XP collected from an XP orb.

### API changes
No API changes made by this PR. There has been 1 new public method added to `Human`, but this is not considered to be API, and does not break API anyway.

### Behavioural changes
Mending enchantment now works.

## Backwards compatibility
None. This is a non-breaking feature addition which should therefore target the next 3.x minor release (3.1).

## Tests
Tested in-game, repairs held items correctly - this has not been thoroughly tested yet, further testing is needed.

### DIFF
--- a/src/pocketmine/entity/Human.php
+++ b/src/pocketmine/entity/Human.php
@@ -34,6 +34,7 @@ use pocketmine\inventory\EntityInventoryEventProcessor;
 use pocketmine\inventory\InventoryHolder;
 use pocketmine\inventory\PlayerInventory;
 use pocketmine\item\Consumable;
+use pocketmine\item\Durable;
 use pocketmine\item\enchantment\Enchantment;
 use pocketmine\item\FoodSource;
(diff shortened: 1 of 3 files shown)


Muqsit:
What if the keys for `$equipment` were `-1` for item in hand and 0, 1, 2, 3 (synced with ArmorInventory slots) for armors. Definitely not worth the hassle, it's just avoiding an `array_search`.
```php
$repairItem = $equipment[$k = array_rand($equipment)];
if($repairItem->getDamage() > 0){
    //...
    if($k === -1){
        $this->inventory->setItemInHand($repairItem);
    }else{
        $this->armorInventory->setItem($k, $repairItem);
    }
}
```

dktapps:
not a big fan of magic numbers, they break too easily.

PR #2267, opened 2018-06-27
reorganize and optimize start.sh

dschwartz783:
## Introduction
The code in start.sh was a bit sloppy. Now, it is not. This new version makes use of double brackets, which is a shell built-in, rather than the single brackets which requires an extra process launch, and is thus less performant, and isn't as powerful. The loop logic has been refactored to remove redundant checks, and is now a lot more concise. This change also makes it possible to do `env DO_LOOP="yes" ./start.sh` if you wanted to. The initial `DO_LOOP="no"` is unnecessary, and removing it enables this functionality.

### API changes

none.

### Behavioural changes
Equivalent behavior

## Backwards compatibility
Yup.

## Tests
It loops with ./start.sh -l, and doesn't with no arguments. The output is the same as before with the nutty logic.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -2,8 +2,6 @@
 DIR="$(cd -P "$( dirname "${BASH_SOURCE[0]}" )" && pwd)"
 cd "$DIR"
 
-DO_LOOP="no"
-
 while getopts "p:f:l" OPTION 2> /dev/null; do
 	case ${OPTION} in
 		p)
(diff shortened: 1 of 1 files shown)


dktapps:
I'm really unconvinced that the power of double brackets is really relevant to already-working code. Furthermore, double brackets is less widely compatible than single ones (see https://serverfault.com/questions/52034/what-is-the-difference-between-double-and-single-square-brackets-in-bash). As for the performance of start.sh, I really strongly doubt that anyone cares.

Other than that, I can see some minor improvements that this PR can offer, but they should be in separate commits.

dschwartz783:
@dktapps There is already one place in start.sh where double brackets were used, so I can either replace that, leave it like this. Double brackets do have more features available since they're a shell built-in, and are generally a better idea to use by default. They are not "POSIX", but as far as I know, all popular shells support them.

Doesn't matter much to me, though. It's one of those things were, there isn't too much of a reason to switch, but there's really no reason not to.

dschwartz783:
I'll just leave them for now.

dktapps (approved):

dktapps:
It took me quite some time to realize this pull was completely pointless since you can just pass the `-l` flag instead of using an environment variable.

dschwartz783:
What does this mean exactly? This is a pretty standard shell script, setting a variable using the "-l" flag. Did you think this change made it required to use environment variables? It does not. Simply removing the explicit set at the beginning of the script was intended to make the use of environment variables an option, but definitely not required.

dktapps:
I never said anything about it being required. I said this is pointless because we have a flag for it.

dschwartz783:
Not at all correct. There are many uses for setting environment variables over using flags.

Regardless, this pull request was primarily about optimizing the shell script. That change rather a side thing. You made the right choice to merge it, do not worry.

dktapps:
thanks, there's no need to be so condescending. My _point_ was that there is nothing you can do with this PR that wasn't already doable by other means.

dschwartz783:
Nothing condescending at all about it. You did a good job and I'm acknowledging that fact. Have more confidence in yourself, you're good at what you do.

Issue #2249, opened 2018-06-19
Reduce the complexity of the `Plugin` interface


dktapps:
### Issue description
`pocketmine\plugin\Plugin` contains a large number of methods which are not necessary to be declared as interface methods. Most of these methods are only used internally by `PluginBase` (for example `reloadConfig()`, `saveDefaultConfig()`, etc). A lot of these methods should not even be public because they are expected to only be used inside plugins themselves (again, look at the config methods, or even `onEnable()`/`onDisable()`.

This issue proposes changes to the API of `Plugin` and `PluginBase` to:
- reduce the complexity of the public API
- reduce the leakiness of the API

These changes should not affect plugins or plugin interoperability significantly. However, any changes made are likely to break backwards compatibility, and for that reason should target PocketMine-MP 4.

dktapps:
`getScheduler()` was introduced in 3.0 which is only used by the `PluginManager`. There should be a better way of managing schedulers which doesn't require every `Plugin` implementation to have `getScheduler()`.

SOF3:
This issue is too broad and cannot be precisely defined. I recommend converting this to a project rather than an issue.

SOF3:
"Complex" is an abstract adjective. It is impossible to define when this issue can be closed. I am closing this issue since [Project 6](https://github.com/pmmp/PocketMine-MP/projects/6) describes the same thing anyway.

PR #2268, opened 2018-06-29
Remove a whole bunch of crap from the `Plugin` public interface

dktapps:
## Introduction
As discussed in #2249 , there are a whole lot of things declared in the `Plugin` public interface which do not need to be there, because they are only used from within `PluginBase`.

### Relevant issues
#2249 

### API changes
- Removed the following from `Plugin` interface: `onLoad()`, `onEnable()`, `onDisable()`, `getResource()`, `saveResource()`, `getResources()`, `getConfig()`, `saveDefaultConfig()`, `saveConfig()`, `reloadConfig()`.
- Changed visibility of the following `PluginBase` methods from `public` to `protected`: `onLoad()`, `onEnable()`, `onDisable()`, `getResource()`, `saveResource()`, `getResources()`, `getConfig()`, `saveDefaultConfig()`, `saveConfig()`, `reloadConfig()`

### Behavioural changes
None. This is purely an API cleanup.

## Backwards compatibility
This does pose _potential_ BC breaks for things using the `Plugin` interface, but in the vast majority of cases, plugins should not be affected by this change.

This now reduces the unnecessary overcomplexity of writing a custom `Plugin` implementation.

## Follow-up
- Consider making `Plugin` not inherit from `CommandExecutor`.

## Tests
This has been tested with DevTools and several other plugins, but not thoroughly. Since these methods are supposed to be internal anyway, it shouldn't break any good code.

### DIFF
--- a/src/pocketmine/plugin/Plugin.php
+++ b/src/pocketmine/plugin/Plugin.php
@@ -29,8 +29,6 @@ namespace pocketmine\plugin;
 use pocketmine\command\CommandExecutor;
 use pocketmine\scheduler\TaskScheduler;
 use pocketmine\Server;
-use pocketmine\utils\Config;
-
 
 /**
  * It is recommended to use PluginBase for the actual plugin
@@ -92,20 +90,6 @@ interface Plugin extends CommandExecutor{
 	 * @return \SplFileInfo[]
 	 */
 	public function getResources() : array;
-
-	/**
-	 * @return Config
-	 */
-	public function getConfig() : Config;
-
-	public function saveConfig();
-
-	/**
-	 * @return bool
-	 */
-	public function saveDefaultConfig() : bool;
-
-	public function reloadConfig();
 
 	/**
 	 * @return Server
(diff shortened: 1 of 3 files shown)


dktapps:
Unsure on the visibility changes for some `PluginBase` methods. `onLoad()`, `onEnable()` and `onDisable()` should definitely be protected because they are hooks, but I'm a bit less sure about the others.

TheAz928:
@dktapps I think PluginBase::getConfig() should be public method, because in some cases (mostly on plugins coded on multi-files) plugin's internal classes needs to access config. When config is protected, developers would have to create a new class variable or method to access the config. There's no harm leaving `getConfig` method public

dktapps:
@TheAz928 you're right. any other things you'd like to point out?

TheAz928:
@dktapps Nothing else, since `getConfig` method allows both `save` and `reload`

dktapps:
Reversed visibility changes for things that might reasonably be used outside of `PluginBase`. These can be discussed further at a later date.

dktapps:
Note that this does not require any visibility changes to your plugins' `onEnable()` `onDisable()` `onLoad()`. Overriding a protected method with a public one is legal.

@PresentKim Note that changing from public to protected in your plugins breaks their compatibility with 3.x.x APIs. If you want to preserve compatibility with 3.x.x, leave those hooks public.

PR #2274, opened 2018-07-03
Fixed a bug where CursorInventory was not cleared after death

zKoz210:
## Introduction
This pull request fixes bug with clearing CursorInventory after death. It is strange that it did not before.

## Tests
Tests were conducted before and after fixes the bug. Everything works fine, inventory was cleaned.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -77,6 +77,7 @@ use pocketmine\inventory\transaction\InventoryTransaction;
 use pocketmine\inventory\transaction\TransactionValidationException;
 use pocketmine\item\Consumable;
 use pocketmine\item\Durable;
+use pocketmine\item\enchantment\Enchantment;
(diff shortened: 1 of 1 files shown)


dktapps:
Unsure that this should be just deleted outright. That could cause loss of items. The cursor item should also be added to the drops if non-null.

dktapps:
apparently you didn't test the latest patch...
```
[15:03:42] [Server thread/CRITICAL]: Unhandled exception executing command 'kill p' in kill: Class 'pocketmine\Enchantment' not found
[15:03:42] [Server thread/CRITICAL]: Error: "Class 'pocketmine\Enchantment' not found" (EXCEPTION) in "src/pocketmine/Player" at line 1479
```

dktapps:
ebbbc581caf301e76fd2cf9f64259b323c52470a resolves this.

PR #2279, opened 2018-07-05
Fixed double chest inventory desync issues, closes #2261

### LINKED ISSUE #2261: DoubleChestInventory doesn't send updates to all viewers
### Issue description

DoubleChestInventory attempts to send updates to viewers of the left/right side depending on the slot you put or remove an item from.
When you view a DoubleChestInventory, the DoubleChestInventory's `$viewers` count is `1` and the left and right chest inventories' `$viewers` count is `0`. [`DoubleChestInventory::setItem()`](https://github.com/pmmp/PocketMine-MP/blob/30a83544a0d4be47ad3cb77c91fb771d554fe337/src/pocketmine/inventory/DoubleChestInventory.php#L67) calls `ChestInventory::setItem()` (ChestInventory = left chest or right chest) which sends the slot update to viewers of THAT chest.

- Expected result: All viewers of DoubleChestInventory should be sent item updates.
- Actual result: Only the person who set an item in the double chest inventory gets the update.

1. View a double chest from two clients.
2. Set an item in the double chest from any one client.
3. The other client can't see the item unless they re-open the inventory.

### OS and versions

* PocketMine-MP: 76174f192036cad1436e9f8f085d92bf3f472db0
* PHP: 7.2.3
* Server OS: Ubuntu 18.04
* Game version: Windows 10

### Plugins
- Test on a clean server without plugins: is the issue reproducible without any plugins loaded? Yes

dktapps:
## Introduction
chest pairing really needs rewriting... this code really sucks

This is hacky but fixes #2261 .

### Behavioural changes
- fixed bug described in #2261 
- Paired chest halves now share the same `DoubleChestInventory` instance

## Follow-up
rewrite chest pairing... it really, really sucks

## Tests
has been lightly tested, but I'm wary of merging this directly due to the potential for recursion bugs. @Muqsit would you mind testing this?

### DIFF
--- a/src/pocketmine/inventory/DoubleChestInventory.php
+++ b/src/pocketmine/inventory/DoubleChestInventory.php
@@ -64,7 +64,12 @@ class DoubleChestInventory extends ChestInventory implements InventoryHolder{
 	}
 
 	public function setItem(int $index, Item $item, bool $send = true) : bool{
-		return $index < $this->left->getSize() ? $this->left->setItem($index, $item, $send) : $this->right->setItem($index - $this->left->getSize(), $item, $send);
+		$old = $this->getItem($index);
+		if($index < $this->left->getSize() ? $this->left->setItem($index, $item, $send) : $this->right->setItem($index - $this->left->getSize(), $item, $send)){
+			$this->onSlotChange($index, $old, $send);
+			return true;
+		}
+		return false;
 	}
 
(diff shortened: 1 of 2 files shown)


Muqsit:
Yes this fixes it, thanks.
I tested it with two different clients viewing the double chest inventory and one of them placing an item in the double chest (as mentioned in the issue).

PR #2284, opened 2018-07-06
Automatically create plugin data directories

### LINKED ISSUE #2248: Automate creation of plugin data directories
### Issue description
An exception will be thrown if attempting to `getConfig()` when the plugin's data directory has not been created.

This stems from the problem that plugins have to create their data directories manually, which seems stupid to me. Nearly every plugin I've seen has some variation of this code on `onEnable()`:

```php
		if(!is_dir($this->getDataFolder())){
			@mkdir($this->getDataFolder());
		}
		//... whatever else
	}
```
This is bad, generates boilerplate and is fragile. It's easy to forget, resulting in astonishing behaviour.

The same could be said of `saveDefaultConfig()`.

I am thinking that either
a) the server should _always_ pre-create the plugin's data-dir
b) the data dir should be created on the fly (how?)

Use the following code:
```php
    $someValue = $this->getConfig()->get("someValue");
}
```

### OS and versions

* PocketMine-MP: https://github.com/pmmp/PocketMine-MP/commit/30ad3a1705081d7f00ba74fc94963e17249fb7a7
* PHP: 7.2.6
* Server OS: windows
* Game version: irrelevant

dktapps:
## Introduction
This issue is described in #2248.

#### Advantages
- No more boilerplate code for every plugin developer to write

#### Disadvantages
- Every plugin now has a data directory created for it, no matter whether it wants to use it or not. However, this is consistent with other platforms that I care to look at (such as Android) so this isn't something I consider to be an issue.

### Relevant issues
Fixes #2248

### Behavioural changes
- Plugins now don't need to create their data directories themselves, since it's now automatic.

## Backwards compatibility
This does not pose any backwards compatibility issues since plugins should check if their data directory exists before trying to create it anyway.

## Tests
Has been tested with a small range of plugins I have on my development server. This is a fairly trivial change.

### DIFF
--- a/src/pocketmine/plugin/PluginBase.php
+++ b/src/pocketmine/plugin/PluginBase.php
@@ -256,9 +256,7 @@ abstract class PluginBase implements Plugin{
 	}
 
 	public function reloadConfig(){
-		if(!$this->saveDefaultConfig()){
-			@mkdir($this->dataFolder);
-		}
+		$this->saveDefaultConfig();
 		$this->config = new Config($this->configFile);
 		if(($configStream = $this->getResource("config.yml")) !== null){
 			$this->config->setDefaults(yaml_parse(Config::fixYAMLIndexes(stream_get_contents($configStream))));
--- a/src/pocketmine/plugin/PluginManager.php
+++ b/src/pocketmine/plugin/PluginManager.php
@@ -181,6 +181,9 @@ class PluginManager{
 						$this->server->getLogger()->error("Projected dataFolder '" . $dataFolder . "' for " . $description->getName() . " exists and is not a directory");
 						return null;
 					}
+					if(!file_exists($dataFolder)){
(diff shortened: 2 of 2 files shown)


nnnlog:
I think that automatic creating data in folders to creating a mess in plugin folder or the plugin which doesn't need create file waste resource

dktapps:
Please read the initial post properly. I already brought up this point.

SOF3:
@nnnlog Creating an empty folder wastes negligible resource unless you installed millions of plugins on your server.

On the other hand, data folders are now created in `plugin_data` instead of along with plugin phars, so I doubt how it could mess things up. On the bright side, it is even better, because your plugin_data directory list is now identical to the plugins directory list (assuming your plugin files are named with the plugin's name).

SOF3 (approved):

PR #2290, opened 2018-07-08
Level: Make spawn protection non-dependent on op count

dktapps:
## Introduction
I don't care if this matches PC behaviour or not. bugs.mojang.com is full of bug reports about this. Just search for "minecraft spawn protection not working" and you'll see what I mean.

If you want to disable spawn protection, actually disable it. This behaviour is something that most users are not aware of and find astonishing when they discover it.

### Relevant issues
More than I care to list. Many people have complained about spawn protection inexplicably not working over the years.

### Behavioural changes
If enabled, spawn protection will no longer be disabled when there are no operators.

## Follow-up
Consider moving spawn protection to an external plugin.

## Tests
This is a trivial change and shouldn't need extensive testing.

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -1627,7 +1627,7 @@ class Level implements ChunkManager, Metadatable{
 
 			$spawnLocation = $this->getSpawnLocation();
 			$s = new Vector2($spawnLocation->x, $spawnLocation->z);
-			if(count($this->server->getOps()->getAll()) > 0 and $t->distance($s) <= $distance){
+			if($t->distance($s) <= $distance){
 				return true;
 			}
 		}


ZizzyZizzy:
Astonished is an understatement. 5 years and I never knew this was a thing. Holy insecure coding, Batman! 

> "Who needs to lock the doors if the owner isn't home/defined? We'll just leave it wide open."

**If anyone doesn't quite get why this is very important:**
I ran a public server for about a year and it was common practice to remove myself from the ops list to prevent unexpected forced op due to auth, book, and other bugs running around at the time. Nobody else had op perms. (With properly configured permissions, nobody should have op perms but owner anyway.)

Of course plugins like WorldGuard are your friend here, but Pocketmine is nowhere near as mature as Java Minecraft and plugins are not updated fast enough to keep up with Mr. dktWizard's updates. Relying on the built-in spawn protection is a real thing.

dktapps:
For posterity's sake, the commit introducing this "feature": https://github.com/pmmp/pocketmine-mp/commit/eb0525e892219508d0c0e4602e835d5ddbacaf45

If there's any mess that can be used to demonstrate why copying from vanilla is not always a good idea, it's this one.

dktapps:
Unsure whether this should go in 3.0.next as a bug fix or 3.1.0 as a feature change.

PR #2292, opened 2018-07-09
Listener: Add documentation on functionality

dktapps:
## Introduction
The `Listener` interface is one of the most magical parts of PocketMine-MP, and before this pull request it didn't have a single bit of documentation.

### Background
There is a recurring theme I've seen with plugin developers. Developers who previously worked with the Bukkit API (which PocketMine-MP's API is mostly derived from) usually don't have too much difficulty getting to grips with PM because a lot of the concepts are shared. This also means that in a lot of cases Bukkit plugin documentation can also sort-of apply to PM.

This generally results in a gap of understanding. Developers directly introduced to PocketMine-MP have very little documentation to work with, and no idea that they could look to Bukkit for clues. Meanwhile, developers who worked with Bukkit and its vastly superior documentation have a significant advantage. This often results in lower quality code because fresh developers have little to no idea what they are doing and are forced to look at the core code itself to find out how to do things, or to look at how other plugin developers did things.

As PM diverges more and more away from the Bukkit API, it is going to be necessary to develop our own comprehensive documentation to ease the entry of new plugin developers into the space.

### API changes
This is merely addition of documentation and does not affect any API or behaviour.

## Follow-up
Document more things which are sorely lacking in documentation.

### DIFF
--- a/src/pocketmine/event/Listener.php
+++ b/src/pocketmine/event/Listener.php
@@ -23,6 +23,32 @@ declare(strict_types=1);
 
 namespace pocketmine\event;
 
+/**
+ * Classes implementing this interface can be registered to receive called Events when something happens.
+ *
+ * Event handlers do not have to have any particular name - they are detected using reflection.
+ *
+ * A function in a Listener class must meet the following criteria to be registered as an event listener:
+ * - MUST be public
+ * - MUST NOT be static
+ * - MUST accept EXACTLY ONE class parameter which:
+ *   - MUST be a VALID class extending Event
+ *   - MUST NOT be abstract, UNLESS it has an @allowHandle annotation
+ *
+ * Functions which meet the criteria can have the following annotations in their doc comments:
+ *
+ * - `@nonHandler`: Marks a function as NOT being an event handler. Only needed if the function meets the above criteria.
+ * - `@softDepend [PluginName]`: Handler WILL NOT be registered if its event doesn't exist. Useful for soft-depending
+ *     on plugin events. Plugin name is optional.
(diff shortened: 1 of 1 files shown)


src/pocketmine/event/Listener.php
```diff
@@ -23,6 +23,32 @@
 
 namespace pocketmine\event;
 
+/**
+ * Classes implementing this interface can be registered to receive called Events when something happens.
```
SOF3: I recommend a reference to PluginManager->registerEvents() here.

src/pocketmine/event/Listener.php
```diff
@@ -23,6 +23,32 @@
...
+ * Event handlers do not have to have any particular name - they are detected using reflection.
+ *
+ * A function in a Listener class must meet the following criteria to be registered as an event listener:
+ * - MUST be public
+ * - MUST NOT be static
+ * - MUST accept EXACTLY ONE class parameter which:
+ *   - MUST be a VALID class extending Event
+ *   - MUST NOT be abstract, UNLESS it has an @allowHandle annotation
+ *
+ * Functions which meet the criteria can have the following annotations in their doc comments:
```
SOF3: Can be better phrased as "listener functions". Otherwise this sounds like functions not meeting the criteria can't have the annotations due to restrictions.

dktapps: I don't see how that would make it any clearer...

dktapps:
I also wonder whether there should be a restriction applied on listener return types, such that

`public function handleEntityDamage(EntityDamageEvent $event) : int`

would not become a handler.

PR #2297, opened 2018-07-11
Added LOCK_EX flag for async save

Frago9876543210:
## Introduction
Supposing that we have economy plugin and we add 10 coins for steve, after save it with parameter `$async = true`. But what if you need to save data for many players? Without this flag, the config can be destroyed.

### DIFF
--- a/src/pocketmine/utils/Config.php
+++ b/src/pocketmine/utils/Config.php
@@ -217,7 +217,7 @@ class Config{
 				}
 
 				if($async){
-					Server::getInstance()->getAsyncPool()->submitTask(new FileWriteTask($this->file, $content));
+					Server::getInstance()->getAsyncPool()->submitTask(new FileWriteTask($this->file, $content, LOCK_EX));
 				}else{
 					file_put_contents($this->file, $content);
 				}


dktapps:
Honestly I don't see a fit purpose for async saving at all. It should either always be synchronous or always asynchronous, and at the user's own option. However, this isn't currently possible because Config doesn't enable you to get the serialized content without writing it to disk.

This flag is not sufficient to prevent destruction of the config or race conditions. Consider the following code:
```php
		for($i = 0, $size = $this->getServer()->getAsyncPool()->getSize(); $i < $size; ++$i){
			$this->getServer()->getAsyncPool()->submitTask(new class extends AsyncTask{
				public function onRun(){
					sleep(5);
				}
			});
		}
		$config = $this->getConfig();
		$config->set("steve", "hi");
		$config->save(true);
		$config->set("steve", "bye");
		$config->save(false);
```
Output:
```

dktapps:
Not to mention that:
- async save creates a cyclic dependency between Server and Config (making it difficult to unit-test)
- there are far larger performance issues such that a Config saving is basically irrelevant. If your configs ever get big enough that saving them is a drag on performance, then you're either saving too frequently or you're misusing Configs.

SOF3:
Async saving is determined by many other factors, such as the order of execution of AsyncTasks. Applying a LOCK_EX is not guaranteed to resolve the issue mentioned in the introduction.

Meanwhile, applying a file lock would prevent concurrent writing of the config file, but cannot prevent overwriting the config file afterward from another thread/process, as @dktapps mentioned. If you are worried about other processes rather than multi-threading in a single server, this change should not be limited to $async. But I doubt it will be meaningful anyway.

Issue #2288, opened 2018-07-08
Change Player chatting abilities to be controlled by permissions


dktapps:
### Issue description
Currently non-operators are not permitted to use colour codes in chat. When a player is opped, this is not updated. This can be easily fixed of course, but this gave me ideas on ways to improve this.

Additionally, to alter this behaviour it is necessary to have a bespoke plugin which specifically overrides this. I think it would be preferable to have this managed by permissions instead, so that permission management plugins can control it.

Additionally, there is currently no simple way to deny permission to chat ("mute"), short of cancelling `PlayerChatEvent` for every chat message from selected players.

This proposes the addition of a new permission node `pocketmine.chat` to allow permission plugins to control chat abilities.

### Steps to reproduce the issue
1. op a player
2. notice that player can't use colour codes until they rejoin

### OS and versions
* PocketMine-MP: https://github.com/pmmp/PocketMine-MP/commit/9c80e349cede7857154c053450c5de801fdbb5b1

lukeeey:
Or you could implement the built in permissions that are shown when clicking a player in the player list

dktapps:
That's just a frontend. This issue is talking about how to handle the backend permission system.

hartleyterw:
How do you go about overriding this with a plugin? PureChat doesn't seem to cut it anymore.

Landy005:
Is there a way to give non operators the ability to use § then? Because for pureperms operators need to be disabled.

PR #2306, opened 2018-07-14
Chat improvements (#2288)

Frago9876543210:
## Introduction
Fix for #2288 and some improvements about mute

### API changes
Added:
`Player::setMuted()`
`Player::isMuted()`

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -310,6 +310,8 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 	protected $allowFlight = false;
 	/** @var bool */
 	protected $flying = false;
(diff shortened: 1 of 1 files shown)


src/pocketmine/Player.php
```diff
@@ -593,6 +604,7 @@ public function setOp(bool $value){
 		}else{
 			$this->server->removeOp($this->getName());
 		}
+		$this->setRemoveFormat(!$value);
```
dktapps: This can be overridden by plugins, which is why I didn't apply such a fix myself already. Additionally this is not related to mute.

Frago9876543210:
@dktapps You offers to do handle messages as parts? For example, for processing mute, color messages, etc.

PR #2308, opened 2018-07-15
Forms API, part 2: Implementation

dktapps:
## Introduction
This is a revision of #1476 , which was abandoned due to me getting sick of messing about with it and it being full of bugs.

This primarily differs from #1476 due to a large simplification in how form data is handled. Previously form response data would be stored in the form itself, which added a lot of extra complexity to form handling, made them impossible to duplicate, reuse, etc etc. In this revision, forms and their subcomponents do not have any responsibility for storing response data - instead it's passed to `onSubmit()`.

This is not yet complete and will probably see further changes before it is done.

### API changes
- Added new classes under the `pocketmine\form` namespace
- Added API method `Player->sendForm()`.

### Behavioural changes
This is a pure API addition and should not affect any existing behaviour, apart from possibly breaking external form API plugins.

## Backwards compatibility
see Behavioural changes

## Tests
has been briefly tested, test scripts will come.

### DIFF
--- a/src/pocketmine/utils/Utils.php
+++ b/src/pocketmine/utils/Utils.php
@@ -547,4 +547,13 @@ class Utils{
 
 		return true; //stfu operator
 	}
+
+	public static function validateObjectArray(array $array, string $class) : bool{
+		foreach($array as $key => $item){
+			if(!($item instanceof $class)){
+				throw new \TypeError("Element \"$key\" is not an instance of $class");
+			}
+		}
+		return true;
+	}
 }
--- a/src/pocketmine/form/BaseForm.php
+++ b/src/pocketmine/form/BaseForm.php
@@ -26,8 +26,6 @@ declare(strict_types=1);
  */
 namespace pocketmine\form;
 
-use pocketmine\Player;
-
 /**
  * Base class for a custom form. Forms are serialized to JSON data to be sent to clients.
  */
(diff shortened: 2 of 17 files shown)


dktapps:
If I recall correctly the remaining issues with this now lie with the queuing mechanism - it's too fragile and there's no way to detect faults.

A possible enhancement could be to have an `IForm` interface to allow custom Form implementations to have their forms sent using `sendForm()`. `sendForm()` is the main advantage this implementation has over other third party implementations.

dktapps:
With this interface separation, this now resolves a cyclic dependency and makes it possible to separate the implementation of forms from the core code, allowing plugin developers the freedom to implement custom libraries to their hearts' content while still being able to use `sendForm()`.

I'm considering moving the main implementation out of the core into a separate library now that this has been resolved.

src/pocketmine/utils/Utils.php
```diff
@@ -547,4 +547,13 @@ public static function errorExceptionHandler(int $severity, string $message, str
...
 		return true; //stfu operator
 	}
+
+	public static function validateObjectArray(array $array, string $class) : bool{
+		foreach($array as $key => $item){
+			if(!($item instanceof $class)){
+				throw new \TypeError("Element \"$key\" is not an instance of $class");
+			}
+		}
+		return true;
```
dktapps: note to self: document the apparently-useless return value here (it's for `assert()`).

This change should be separated from the other changes since it's not directly related.

PR #2310, opened 2018-07-16
able to change attackTick

djdisodo:
## Introduction
i'm making a gun plugin.
attack() doesn't worked on some guns like shoutgun(it spawns many projectiles)

## Tests
tested

### DIFF
--- a/src/pocketmine/entity/Living.php
+++ b/src/pocketmine/entity/Living.php
@@ -560,7 +560,7 @@ abstract class Living extends Entity implements Damageable{
 			return;
 		}
 
-		$this->attackTime = 10; //0.5 seconds cooldown
+		$this->attackTime = $source->getAttackTick(); //0.5 seconds cooldown
 
 		if($source instanceof EntityDamageByEntityEvent){
 			$e = $source->getDamager();
--- a/src/pocketmine/entity/Living.php
+++ b/src/pocketmine/entity/Living.php
@@ -560,7 +560,7 @@ abstract class Living extends Entity implements Damageable{
 			return;
 		}
 
-		$this->attackTime = $source->getAttackTick(); //0.5 seconds cooldown
+		$this->attackTime = $source->getAttackCooldown();
 
 		if($source instanceof EntityDamageByEntityEvent){
 			$e = $source->getDamager();
--- a/src/pocketmine/event/entity/EntityDamageEvent.php
+++ b/src/pocketmine/event/entity/EntityDamageEvent.php
@@ -202,6 +202,8 @@ class EntityDamageEvent extends EntityEvent implements Cancellable{
 	}
 	
 	/**
+	 * Returns the cooldown in ticks before the target entity can be attacked again.
+	 *
 	 * @return int
 	 */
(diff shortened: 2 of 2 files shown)


Thunder33345:
what about my plugin who use `CAUSE_CUSTOM` and i want attack ticks
you would break my plugin!!!!!!!

djdisodo:
@Thunder33345 
https://github.com/pmmp/PocketMine-MP/compare/master...djdisodo:patch-2?expand=1
how about this

Thunder33345:
looks better
the default should be 10 tho?

Muqsit:
Nice, but `attackDelay`, `getAttackDelay` and `setAttackDelay` sound more meaningful than `attackTick`.

Thunder33345:
more like [quazi]immune tick
blame mojang naming stuff

dktapps:
https://xkcd.com/1172/

src/pocketmine/entity/Living.php
```diff
@@ -560,7 +560,7 @@ public function attack(EntityDamageEvent $source) : void{
 			return;
 		}
 
-		$this->attackTime = 10; //0.5 seconds cooldown
+		if($source->getCause() !== EntityDamageEvent::CAUSE_CUSTOM) $this->attackTime = 10; //0.5 seconds cooldown
```
SOF3: OCD: Use curly braces.

SOF3 (dismissed):

src/pocketmine/event/entity/EntityDamageEvent.php
```diff
@@ -197,4 +200,18 @@ public function canBeReducedByArmor() : bool{
...
+	 * @return int
+	 */
+	public function getAttackTick() : int{
+		return $this->attackTick;
+	}
+	
+	/**
+	 * @param int $attackTick
+	 */
+	public function setAttackTick(int $attackTick) {
```
dktapps: this should have a `void` typehint.

dktapps (commented):
To me it seems more clear to refer to it as an attack cooldown than `attackTick` - the use is more clear from the name.

src/pocketmine/event/entity/EntityDamageEvent.php
```diff
@@ -197,4 +200,18 @@ public function canBeReducedByArmor() : bool{
...
+	 */
+	public function getAttackCooldown() : int{
+		return $this->attackCooldown;
+	}
+	
+	/**
+	 * @param int $attackTick
+	 */
+	public function setAttackCooldown(int $attackTick) : void{
+		$this->attackTick = $attackCooldown;
```
dktapps: looks like you missed a couple of things here

dktapps:
The only nitpick I can make here now is the fact that this will not work for non-living entities. I'm not really sure how that can be dealt with.

djdisodo:
@SOF3 please del change request....

dktapps (approved):
LGTM, will look at integrating this into 3.2.

dktapps:
Squash-merged in the referenced commit. Thanks.

Issue #2219, opened 2018-06-01
Consider removing config default polyfilling


dktapps:
### Issue description
A little-known "feature" in PocketMine-MP that most plugin developers don't know about, is that their `config.yml` will be polyfilled with defaults from `resources/config.yml`.

The intents of this feature may be good, but there are a few unexpected negative side effects that make me question whether this is worth it.

Most notably, consider the following configuration:
config.yml:
```yml
test:
 d: true
 e: true
 f: true
```
resources/config.yml
```
test:
 a: true
 b: true
 c: true
```

will result in the following astonishing behaviour:
```
array(6) {
  ["d"]=>
  bool(true)
  ["e"]=>
  bool(true)
  ["f"]=>
  bool(true)
  ["a"]=>
  bool(true)
  ["b"]=>
  bool(true)
  ["c"]=>
  bool(true)
}
```
This behaviour is highly unexpected and could even be considered a security exploit depending on how the config.yml is used.

This proposes the removal of such polyfilling, or at the least some way to disable it. I do not believe that the majority of plugin developers know about this, and most that I have spoken to would consider this as a bug. It's also confusing because defaults can be given when `get()`ing values from configs, so this doesn't make sense anyway.

This makes me wonder if this is one of the reasons why plugins commonly avoid use of `getConfig()` and instead using their own created `Config` objects.

SOF3:
Another possible case is when you add a line `delete this line: after you have filled the required fields`.

PR #2316, opened 2018-07-21
PluginBase: Do not fill defaults from resources/config.yml

dktapps:
## Introduction
described in referenced issue.

### Relevant issues
#2219 

### API changes
- `PluginBase->reloadConfig()` will no longer fill in defaults from `resources/config.yml`. Instead, default values for `Config->get()` should be used (is there anyone who _doesn't_ do this???)

## Backwards compatibility
- Potential BC break for plugins whose users have messed up their configs where defaults have not been used.

## Tests
This is a trivial change and does not need substantial testing.


dktapps:
This seems to be green so I'm going to go ahead and merge it.

PR #2318, opened 2018-07-22
Player: Separate disconnection logic into disconnect() API method

### LINKED ISSUE #1239: Crash when closing players during some events
### Issue description
When I try to reconnect a player to a server, this happens:

[02:15:50] [Server thread/CRITICAL]: Error: "Call to a member function addChunkPacket() on null" (EXCEPTION) in "/home/data/other/src/pocketmine/Player" at line 2147
[02:15:50] [Server thread/CRITICAL]: [Network] Stopped interface pocketmine\network\mcpe\RakLibInterface due to Undefined index: X.X.X.X:62724 (my ip was here)

I let players reconnect to the server if they are stuck or something.

1. Transfer a player to the same server they're already on using code below.

### OS and versions

* PocketMine-MP: ALPHA7 (is that enough info?)
* PHP: 7.0
* Server OS: ubuntu linux
* Game version: irrelevant

### Plugins
No it isn't reproducable without plugins. but this isnt a plugin error..
	public function transfer(Player $player, string $identifier, $msg = ""){
		$pk = new TransferPacket();
		$pk->address = $this->getAddress(); //server ip
		$pk->port = $this->getPortByIdentifier($identifier); //server port
		$player->directDataPacket($pk);
		if($msg != "") $this->addTransferMessage($player, $msg);
		$player->close();
	}

UPDATE: It doesn't do this anymore with Player->kick(), but I still think this is unexpected behavior..? Shouldn't crash.

dktapps:
## Introduction
close() should NOT be used by plugins anymore.

disconnect() will do the logic that won't cause things to crash. close() will handle breaking cyclic references and cleanup only.

This fixes a multitude of bugs relating to player kicking and closing during events. This fixes #1239 and a range of other related (but not reported on the issue tracker) bugs.

### Relevant issues
#1239

### API changes
- Added new API method `Player->disconnect()`.
- `Player->close()` should now be considered internal, and should not be used.

### Behavioural changes
None known.

## Backwards compatibility
This breaks backwards compatibility by changing the use and functionality of `Player->close()`. It should perhaps be made private, but this isn't currently possible due to limitations in how entity deletion works.

## Follow-up
- Clean up generic Entity deletion in this fashion.
- Make Player only be constructed when the login sequence is completed, to remove the mess of incomplete players.

## Tests
Transferring, kicking and disconnecting during `PlayerInteractEvent` and `PlayerAnimateEvent` have been tested, amongst other things.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -1769,7 +1769,7 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 
 		$this->server->getPluginManager()->callEvent($ev = new PlayerPreLoginEvent($this, "Plugin reason"));
 		if($ev->isCancelled()){
-			$this->close("", $ev->getKickMessage());
+			$this->disconnect("", $ev->getKickMessage());
 
 			return true;
 		}
@@ -1804,7 +1804,7 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 		}
 
 		if($error !== null){
-			$this->close("", $this->server->getLanguage()->translateString("pocketmine.disconnect.invalidSession", [$error]));
+			$this->disconnect("", $this->server->getLanguage()->translateString("pocketmine.disconnect.invalidSession", [$error]));
 			return;
 		}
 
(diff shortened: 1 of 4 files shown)


dktapps:
I also considered other ways to solve this, such as the discussed exception throws, but this appears to be the least BC-breaking way to solve it.

dktapps:
Apparently the server will crash when a player is kicked due to login timeout. 🙄

dktapps:
Going to put this on hold, since it requires a bunch of hacks because of the mess of Player construction.

PR #2322, opened 2018-07-23
Implemented player entity interaction

Spajker7:
## Introduction
This pull request implements basic changes needed for Player interaction with entities.

### Relevant issues
/

### API changes
Adds a new event: PlayerGentlyTouchEntityEvent
Adds a new method Entity::interact()

### Behavioural changes
/

## Backwards compatibility
The changes are backwards compatible.

## Follow-up
Rename the event to something sensible. 

## Tests
/

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -908,7 +908,7 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 		$this->usedChunks[Level::chunkHash($x, $z)] = true;
 		$this->chunkLoadCount++;
 
-		$this->networkSession->getInterface()->putPacket($this->networkSession, $payload);
+		$this->networkSession->sendEncoded($payload);
 
 		if($this->spawned){
 			foreach($this->level->getChunkEntities($x, $z) as $entity){
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -1903,7 +1903,7 @@ class Server{
 	public function broadcastPacketsCallback(string $payload, array $sessions, bool $immediate = false){
 		/** @var NetworkSession $session */
 		foreach($sessions as $session){
-			$session->getInterface()->putPacket($session, $payload, $immediate);
+			$session->sendEncoded($payload, $immediate);
 		}
 	}
 
--- a/src/pocketmine/network/mcpe/NetworkSession.php
+++ b/src/pocketmine/network/mcpe/NetworkSession.php
@@ -192,6 +192,11 @@ class NetworkSession{
 		}
 	}
 
+	public function sendEncoded(string $payload, bool $immediate = false) : void{
+		//TODO: encryption
+		$this->interface->putPacket($this, $payload, $immediate);
+	}
+
 	/**
 	 * Disconnects the session, destroying the associated player (if it exists).
 	 *
--- a/src/pocketmine/network/mcpe/handler/PreSpawnSessionHandler.php
+++ b/src/pocketmine/network/mcpe/handler/PreSpawnSessionHandler.php
@@ -86,7 +86,7 @@ class PreSpawnSessionHandler extends SessionHandler{
 		$this->player->sendAllInventories();
 		$this->player->getInventory()->sendCreativeContents();
 		$this->player->getInventory()->sendHeldItem($this->player);
-		$this->session->getInterface()->putPacket($this->session, $this->server->getCraftingManager()->getCraftingDataPacket());
+		$this->session->sendEncoded($this->server->getCraftingManager()->getCraftingDataPacket());
 
 		$this->server->sendFullPlayerListData($this->player);
 	}


src/pocketmine/Player.php
```diff
@@ -2241,7 +2242,34 @@ public function handleInventoryTransaction(InventoryTransactionPacket $packet) :
...
 				switch($type){
 					case InventoryTransactionPacket::USE_ITEM_ON_ENTITY_ACTION_INTERACT:
-						break; //TODO
+						if(!$target->isAlive()){
+							return true;
+						}
+
+						$heldItem = $this->inventory->getItemInHand();
+
+						$ev = new PlayerGentlyTouchEntityEvent($this, $target, $heldItem, $packet->trData->clickPos);
```
dktapps: I hope you don't expect me to merge this...

src/pocketmine/Player.php
```diff
@@ -2241,7 +2242,34 @@ public function handleInventoryTransaction(InventoryTransactionPacket $packet) :
...
+
+						if(!$this->canInteract($target, 8)){
+							$ev->setCancelled();
+						}
+
+						$this->server->getPluginManager()->callEvent($ev);
+
+						$target->interact($ev);
```
dktapps: what exactly is the point of passing the event here? Seems like extra work for no good reason.

src/pocketmine/event/player/PlayerGentlyTouchEntityEvent.php
```diff
@@ -0,0 +1,78 @@
...
+		return $this->item;
+	}
+
+	/**
+	 * @return Vector3
+	 */
+	public function getClickPos(): Vector3{
+		return $this->clickPos;
+	}
+}
```
dktapps: formatting needs fixing throughout this file

src/pocketmine/entity/Entity.php
```diff
@@ -898,6 +899,13 @@ public function attack(EntityDamageEvent $source) : void{
 		$this->setHealth($this->getHealth() - $source->getFinalDamage());
 	}
 
+	/**
+	 * @param PlayerGentlyTouchEntityEvent $source
+	 */
+	public function interact(PlayerGentlyTouchEntityEvent $source) : void{
```
dktapps: Please let's not have another godawful mess like `Entity->attack()`. The intent of this function is not clear from its name. Is the entity interacting with some other entity, or is it the one getting interacted with?

src/pocketmine/Player.php
```diff
@@ -2241,7 +2242,34 @@ public function handleInventoryTransaction(InventoryTransactionPacket $packet) :
...
+
+						if(!$ev->isCancelled()){
+							$oldItem = clone $heldItem;
+
+							if($heldItem->onClickEntity($this, $target, $packet->trData->clickPos)){
+								if(!$heldItem->equalsExact($oldItem)){
+									$this->inventory->setItemInHand($heldItem);
+									$this->inventory->sendHeldItem($this->hasSpawned);
```
dktapps: I'm surprised this isn't automatic... that's perhaps a bug.

src/pocketmine/Player.php
```diff
@@ -2241,7 +2242,34 @@ public function handleInventoryTransaction(InventoryTransactionPacket $packet) :
...
+
+						if(!$this->canInteract($target, 8)){
+							$ev->setCancelled();
+						}
+
+						$this->server->getPluginManager()->callEvent($ev);
+
+						$target->interact($ev);
+
+						if(!$ev->isCancelled()){
```
dktapps: Seems to be an inconsistency here. The item is only receiving the click event if the event is not cancelled, but the entity will always receive it, cancelled or not.

Additionally, parameter inconsistency exists here.

dktapps (changes requested):
I've given this a quick once-over, but it's not in depth.

dktapps:
Note also that a similar change set is also present in #2124 .

src/pocketmine/Player.php
Spajker7: For Entity::interact() I tried to be consistent with other methods in Entity ( Entity::attack ). 
For Item::onClickEntity, I tried to be consistent with other methods in Item ( Item::onClickAir, Item::onActivate ).

src/pocketmine/event/player/PlayerGentlyTouchEntityEvent.php
Spajker7: Whats wrong exactly ?

Muqsit: leave space after the colon

src/pocketmine/Player.php
dktapps: Entity::attack() calls the event passed to it. Here you're calling it externally and then passing it to the hook, which seems pointless.

iiFlamiinBlaze:
#2124 ...

Spajker7:
@iiFlamiinBlaze What's your point?

Muqsit:
@iiFlamiinBlaze Better? I don't see how any of the PRs are better than the other (neither have you stated any ways one is better over the other), and if you think so, please recommend what needs to be changed rather than starting such silly fights.

The other PR has been inactive for more than 2 months, so why not let someone else implement it rather than waiting an eternity?

iiFlamiinBlaze:
@Muqsit I was not starting a fight, I was just stating on the way it was implemented, that it is better. I also asked why he decided to make this PR when obviously there is one already open with the same exact feature. I get that the PR is inactive, but its still open so there must be some significance to keep it open.

Spajker7:
Again, you haven't explained how the other PR is better? I looked at the code and it seems to be the exact thing I did, with the exception of my implementation also having a method inside the Item class.

PR #2323, opened 2018-07-23
Clear the title bar when the server shutdown.

OGFris:
## Introduction
It always annoys me how when i stop my server and want to keep my terminal open i still have the title of pocketmine.

### Relevant issues
Not a real issue.

## Changes
Now the title bar gets cleared when the server shutdown.

### API changes
No api changes.

### Behavioural changes
No behavioural changes.

## Backwards compatibility
No backwards compatibility.

## Follow-up
No action required after this pull request.

Requires translations:

| Name | Value in eng.ini |
| :--: | :---: |
| `foo.bar` | `Foo bar` |

--> No.

## Tests
This will clear the old pocketmine title bar text off the terminal (or cmd) when pocketmine shutdown just as simple as it sounds like.

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -2077,7 +2077,9 @@ class Server{
 			$this->logger->emergency("Crashed while crashing, killing process");
 			@Utils::kill(getmypid());
 		}
-		echo "\x1b]0; \x07";
+		if($this->doTitleTick){
+				echo "\x1b]0; \x07";
+		}
 	}
 
 	/**
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -2077,7 +2077,7 @@ class Server{
 			$this->logger->emergency("Crashed while crashing, killing process");
 			@Utils::kill(getmypid());
 		}
-
+		echo "\x1b]0; \x07";
 	}
 
 	/**


dktapps:
and what about when the user's console doesn't support ANSI escape codes? you just echoed some garbahe characters into their console...

OGFris:
yea i guess but just do it, i hate it when the title still exists ugh

dktapps:
I can't reproduce the bug you're talking about on any platform I've tested so far.

OGFris:
well i have always got this bug on ubuntu but perhaps it's because of putty

PotatoeTrainYT:
fris bad dev legacy sucks play sylph instead

dktapps:
I tested native Windows, native Ubuntu, and Ubuntu in PuTTY, and have no sign of this bug. How odd.

lukeeey:
i believe it happens on Android in termux when ssh'ing into a server

dktapps:
That's more likely a bug in your terminal app than anything else.

nnnlog:
@dktapps When I used XShell program, I could find this error.
```OS: Ubuntu 18.04, PM Version: lastest```

dktapps:
I'm not merging this since you haven't addressed the issue of what happens when the user's console doesn't support ANSI escape codes. In that case, the console will likely misbehave or garbage characters will be shown on screen.

OGFris:
alright i'll do it

dktapps:
Merged with alterations as 69c54e789a13e426d411f1fd9e1273b75f9ac4f6.

PR #2324, opened 2018-07-24
Move Internet-related functions from Utils into their own class

dktapps:
## Introduction
Haven't decided whether or not to move this to the pocketmine\network namespace, as that's perhaps where it belongs.

This pulls a bunch of Internet-related functions out of the generic `Utils` class into its own dedicated `Internet` class. This provides better organisation and less mess in `Utils`.

Deprecated functions have been left in `Utils` as redirects. These deprecated functions should be removed no later than 4.0.0.

### API changes
- Added `Internet::getIP()`, `Internet::getURL()`, `Internet::postURL()`, and `Internet::simpleCurl()`.
- Deprecated the corresponding functions in `Utils`. Updating to the new functions is as simple as replacing `Utils` with `Internet`, since this doesn't break backwards compatibility.

### Behavioural changes
None known.

## Backwards compatibility
No issues, although the abovementioned deprecations should be catered for by plugin developers. These deprecated redirects will be removed no later than 4.0.0.

## Follow-up
Remove the deprecated functions on `master` once this has been merged.

## Tests
Has been briefly tested.

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -99,6 +99,7 @@ use pocketmine\timings\TimingsHandler;
 use pocketmine\updater\AutoUpdater;
 use pocketmine\utils\Binary;
 use pocketmine\utils\Config;
+use pocketmine\utils\Internet;
 use pocketmine\utils\MainLogger;
 use pocketmine\utils\Terminal;
 use pocketmine\utils\TextFormat;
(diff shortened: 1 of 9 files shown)


dktapps:
Since this pull request is introducing new API methods, it may also be desirable to alter the behaviour of the false-returning functions to allow them to throw exceptions instead, as discussed internally. The legacy deprecated functions can then convert such exceptions back to the old behaviour.

PR #2328, opened 2018-07-25
Remove $handlerList from PlayerExperienceChangeEvent

adeynes:
I noticed that PlayerExperienceChangeEvent was the only event class that still declared a `$handlerList` property. It is no longer needed, and the `$handlerList` declaration for every other event class was removed with commit 49fbbea. This PR removes this small inconsistency.

Nothing in the PMMP code uses `PlayerExperienceChangeEvent::$handlerList` anymore, so there will not be any effects on anything.


dktapps:
Merged as a4965842d6f1c1eb241e0e0bc8d8b542d9468ab9. This will be in the next 3.0.x and 3.1.x releases.

PR #2329, opened 2018-07-25
add implements in OfflinePlayer

djdisodo:
why not?


dktapps:
Well sunshine, a better question is why, because you haven't read the source code, so I'm going to close this. https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/IPlayer.php#L28

PR #2330, opened 2018-07-25
Added ItemDamageEvent

TwistedAsylumMC:
## Introduction

Earlier today i was wondering how to disable items taking damage, and instead of doing it in multiple events i thought one event for it all would be better.

### API changes

Added event\item
Added ItemDamageEvent

## Tests

script:
```php
public function onItemDamage(ItemDamageEvent $event) : void{
    var_dump($event->getDamage());
    $event->setDamage($event->getDamage() + 2);
    var_dump($event->getDamage());
}
```
output:
```php
int(1)
int(3)
```
(tested breaking a block)

### DIFF
--- a/src/pocketmine/event/item/ItemDamageEvent.php
+++ b/src/pocketmine/event/item/ItemDamageEvent.php
@@ -1,0 +1,72 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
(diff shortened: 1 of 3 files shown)


SalmonDE:
Did you ever consider looking at this?
https://github.com/pmmp/PocketMine-MP/blob/98b34aaac8b8b26ccee5ffe856d5237954b3c8fd/src/pocketmine/item/Durable.php#L43

lukeeey:
lol

PotatoeTrainYT:
yeah, but that doesnt help with setting how much damage is taken too

DaPigGuy:
The names of the function getDamage() and setDamage() are slightly misleading. Maybe rename to getAppliedDamage() and setAppliedDamage()?

djdisodo:
how about adding `ItemEvent::setItem(Item $item)`

PotatoeTrainYT:
@djdisodo why would that be necessary? You cant simply change the item that's taking damage, It's like saying add ``PlayerEvent->setPlayer(Player $player)``

dktapps:
This serves little to no purpose since a plugin detecting this is not able to find out who caused the damage or how the damage was done. This would be somewhat alike to having a BlockBreakEvent without any player.

PR #2340, opened 2018-07-28
Allow entities to override createBaseNBT 

Spajker7:
## Introduction
This PR allows entities to override createBaseNBT in order to add additional NBT for entity spawning.
For example Sheep can override createBaseNBT in order to add color randomness.

### Relevant issues
/

-->

### API changes
Added Entity::createBaseNBTForEntity

### Behavioural changes
/

## Backwards compatibility
No backwards incompatible changes.

## Follow-up
/

## Tests
/

### DIFF
--- a/src/pocketmine/entity/Entity.php
+++ b/src/pocketmine/entity/Entity.php
@@ -356,6 +356,25 @@ abstract class Entity extends Location implements Metadatable, EntityIds{
 	}
 
 	/**
+	 * Creates the base NBT for the specified type of entity.
+	 *
+	 * @param int|string   $type
+	 * @param Vector3      $pos
+	 * @param null|Vector3 $motion
+	 * @param float        $yaw
+	 * @param float        $pitch
+	 * @return Entity|null
+	 */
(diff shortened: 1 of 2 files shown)


Muqsit:
The way tiles do it, they've got the same structured method — `Entity::createBaseNBT()` (but it's called `Tile::createNBT()`). They also have a method `Tile::createAdditionalNBT()` which is called through `Tile::createNBT()` for optionally adding any further entity data.

I think an approach like that would be desirable.
```php
class Entity{
    public static function createAdditionalNBT($tag){
    }

    public static function createBaseNBT(...){
        $tag = ...;
        static::createAdditionalNBT($tag);
        return $tag;
    }
}
```

dktapps:
I don't think this makes any sense. `Entity::createNBT()` is intended to create a minimal tag that contains the bare essentials for constructing an entity. Anything else can be filled in by the entity's own constructor.

The reason such methods exist for `Tile`s is because their creation is directly dependent on the arguments of their block placement. Here, this is not true.

dktapps:
read what I said... it makes zero sense here.

> The reason such methods exist for Tiles is because their creation is directly dependent on the arguments of their block placement.

PR #2342, opened 2018-07-29
fix #1551 Slowness Effect levels >= 7 do not reset FOV when removed

### LINKED ISSUE #1551: Slowness Effect messes with the FOV
### Issue description

- Expected result: FOV reverts back to normal after slowness effect is removed/expires.
- Actual result: FOV does not revert back to normal until you rejoin. _This cannot be reproduced in vanilla_.

1. /give <yourself> slowness 1 100
2. Your FOV wont revert back.

### OS and versions

* PocketMine-MP: 48a7627
* PHP: 7.2RC6
* Server OS: Ubuntu 14.04 LTS 64-bit
* Game version: Win10

### Plugins
- Test on a clean server without plugins: Issue still exists

djdisodo:
slowness always has higher priority than speed

tested

### DIFF
--- a/src/pocketmine/entity/Effect.php
+++ b/src/pocketmine/entity/Effect.php
@@ -358,7 +358,7 @@ class Effect{
 				break;
 			case Effect::SLOWNESS:
 				$attr = $entity->getAttributeMap()->getAttribute(Attribute::MOVEMENT_SPEED);
-				$attr->setValue($attr->getValue() / (1 - 0.15 * $instance->getEffectLevel()));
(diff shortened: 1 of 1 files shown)


djdisodo:
#1551

dktapps:
If it was this simple it would have been done already. This is just a hack and will break things such as overlapping speed and slowness effects.

As I wrote in the issue, this requires an attribute modifier implementation.

Please also read the contribution guidelines, particularly the part that says **avoid the web editor**.

PR #2356, opened 2018-08-04
Prepend parent name to permissions starting with a dot

SOF3:
## Introduction
Conventionally <sup>[citation needed]</sup>, permission names are constructed according to the permission inheritance. For example, if a permission is a child of the permission `example.permission`, the child permission name should be `example.permission.` plus its own part of the name. This pull request eases declaring permissions by allowing plugins to skip the parent permission name.

### Targeted version
This pull request is targeted at API 4.0.0 as it contains potential backward-incompatible changes.

### API changes
Permissions declared in plugin.yml can no longer use names starting with `.`. Using `.` in a root permission will trigger an InvalidStateException. Using `.` in a child permission will mean prepending with the parent permission.

### Behavioural changes
nil

## Backwards compatibility
The pull request is backwards-incompatible, as it changes the behaviour of plugins with permissions starting with `.`.

However, none of the latest build of 2377 plugins built on Poggit-CI (1220 of which declares permissions) uses permissions starting with `.`. Therefore, I consider this change to be *marginally* backward-compatible.

## Tests
Tests will be written for this change after reviews have been given on the idea.

### DIFF
--- a/src/pocketmine/permission/Permission.php
+++ b/src/pocketmine/permission/Permission.php
@@ -92,16 +92,24 @@ class Permission{
 	}
 
 	/**
-	 * @param string $name
-	 * @param array  $data
-	 * @param string $default
-	 * @param array  $output
-	 *
+	 * @param string      $name
+	 * @param array       $data
+	 * @param string      $default
+	 * @param array       $output
+	 * @param null|string $parentName
 	 * @return Permission
 	 *
 	 * @throws \Exception
 	 */
-	public static function loadPermission(string $name, array $data, string $default = self::DEFAULT_OP, array &$output = []) : Permission{
+	public static function loadPermission(string $name, array $data, string $default = self::DEFAULT_OP, array &$output = [], ?string $parentName = null) : Permission{
+		if($name{0} === "."){
+			if($parentName === null){
+				throw new \InvalidStateException("Permission names must not start with a dot");
+			}
+
+			$name = $parentName . $name;
+		}
+
 		$desc = null;
 		$children = [];
 		if(isset($data["default"])){
@@ -117,7 +125,7 @@ class Permission{
 			if(is_array($data["children"])){
 				foreach($data["children"] as $k => $v){
 					if(is_array($v)){
-						if(($perm = self::loadPermission($k, $v, $default, $output)) !== null){
+						if(($perm = self::loadPermission($k, $v, $default, $output, $name)) !== null){
 							$output[] = $perm;
 						}
 					}


lukeeey:
So now plugin developers can be even more lazy. Great.

src/pocketmine/permission/Permission.php
```diff
@@ -92,16 +92,24 @@ public static function loadPermissions(array $data, string $default = self::DEFA
...
+	 * @param null|string $parentName
 	 * @return Permission
 	 *
 	 * @throws \Exception
 	 */
-	public static function loadPermission(string $name, array $data, string $default = self::DEFAULT_OP, array &$output = []) : Permission{
+	public static function loadPermission(string $name, array $data, string $default = self::DEFAULT_OP, array &$output = [], ?string $parentName = null) : Permission{
+		if($name{0} === "."){
+			if($parentName === null){
+				throw new \InvalidStateException("Permission names must not start with a dot");
```
dktapps: Not sure this is an invalid state condition. Perhaps invalid argument? It's bad data...

src/pocketmine/permission/Permission.php
```diff
@@ -92,16 +92,24 @@ public static function loadPermissions(array $data, string $default = self::DEFA
...
-	 * @param string $name
-	 * @param array  $data
-	 * @param string $default
-	 * @param array  $output
-	 *
+	 * @param string      $name
+	 * @param array       $data
+	 * @param string      $default
+	 * @param array       $output
+	 * @param null|string $parentName
```
dktapps: there should be an additional newline here, between the @param and @return

dktapps (commented):
I've made a couple of nitpicks.

As discussed internally I think there are conceptual issues with this, and the permission system has been considered for an overhaul for 4.0.

This minor feature addition seems to be merely a workaround for existing bad design.

Given a permission tree like
```
parent:
  children:
    parent.child: blah
```
the `parent` part should **always** be the prefix anyway, otherwise you would not declare it as a child to begin with.
It would be better to be able to simply declare:
```
parent:
  children:
    child: blah
```
instead. Yes, such a change is BC-breaking, but this change also targets the next major version due to BC breaks.

TL;DR: I doubt it's worth merging this unless it goes onto a minor 3.x release (disregarding potential breakages), because on a major there are bigger issues to deal with anyway.

src/pocketmine/permission/Permission.php
SOF3: Idk, blame @shoghicp. All the other `throw`s in this function throw `InvalidStateException`. I do the same just for consistency.

dktapps: there's literally one other throw... that makes no sense. "Let's be consistently wrong because blame @shoghicp"

dktapps:
I think this should be considered for 4.0 integration IF - and only if - a permissions handling rewrite does not fit into timing for 4.0.

SOF3:
Closed as stale. A BC break in 4.0 is most likely undesired as larger scale changes might happen in 5.0

dktapps:
Some additional context for posterity: It later came to my attention that `children` in permissions mixes two entirely different concepts together, causing an interesting variety of permission bugs for plugins because nobody understands how it works.

TL;DR: I intend to remove `children` entirely at a later date, so this PR is entirely obsolete.

PR #2358, opened 2018-08-06
Network buffering

dktapps:
## Introduction
This introduces buffering for packet batching.

Prior to these changes, every packet would be compressed by itself and sent over RakNet directly. This is an enormous waste of bytes due to the overhead of sending a packet:
- 2 bytes (zlib header)
- 1 byte (zlib block header)
- 4 bytes (zlib block length)
- 4 bytes (adler32)
- 8 bytes (encryption checksum)
- 10 bytes (RakNet datagram info)
- 10 bytes IF packet has to be split

This means that sending a single packet of 1 byte can incur up to 29 bytes of extra overhead in the current architecture. This is a huge waste.

These changes reduce this wastage by buffering packets per-session until the end of the tick before sending.

These changes also have a performance advantage due to less compression, more batched compressible data, less data to be sent overall, less checksums, etc.

### Behavioural changes
- Async compression is now reliable and does not cause race conditions due to queuing.
- Bandwidth usage reductions, especially when sending lots of small packets.
- Performance improvements.

This has been PR'd because of concerns over the latency when mixing buffering types (for example level buffering). This should be addressed before this is merged.

## Tests
Has been tested in-game, but needs more work to reduce sending latency where possible.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -90,6 +90,7 @@ use pocketmine\nbt\tag\ByteTag;
 use pocketmine\nbt\tag\CompoundTag;
 use pocketmine\nbt\tag\DoubleTag;
 use pocketmine\nbt\tag\ListTag;
+use pocketmine\network\mcpe\CompressBatchPromise;
 use pocketmine\network\mcpe\NetworkCipher;
 use pocketmine\network\mcpe\NetworkSession;
 use pocketmine\network\mcpe\protocol\AdventureSettingsPacket;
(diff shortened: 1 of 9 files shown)


dktapps:
A quick real-world example:

We all know about the spam bug that occurs when you right-click and hold on W10. This spam causes approximately 9KB/s feedback from the server when exploiting this bug (resending blocks). With these changes, that's reduced to approximately 2.2KB/s.

The main impact these changes have is significantly reduced bandwidth cost when you send a large number of small packets.

dktapps:
Merging this as-is since it's an improvement over the existing state of things - the latency issues with level buffering have been resolved except for async compression, and async compression has always had additional latency associated with it anyway.

PR #2361, opened 2018-08-08
Kill entity runtime NBT

dktapps:


PresentKim:
I found solution, Sorry
- load when call `PlayerLoginEvent`
- save when call `PlayerDataSaveEvent`

PresentKim:
I found a way to load the data from the **player's NBT**.  
 - load when call `PlayerLoginEvent`  
  
--------  
  
But I can't find a way to save the data in the **player's NBT**.  
  
**I tried save when call `PlayerDataSaveEvent`**  
It worked when player quit the server.  
But when server closed, It not work because the plugin is disabled before the player is kicked from server.  

  
How can I save the data in the **player's NBT**?  
Should I give up saving the data there?

PR #2362, opened 2018-08-08
Added experimental support for ZOPFLI chunk compression

dktapps:
## Introduction
This PR introduces support for using ZOPFLI to super-compress chunks.

ZOPFLI can produce outputs 10-20% smaller than ZLIB level 9 does, but at a huge extra CPU cost.
This huge CPU cost (generally a couple of orders of magnitude) may be worth it for big servers which get lots of traffic, whose maps are static (resulting in a very high percentage of cache hits).

If the map is immutable, the tradeoff between CPU cost and bandwidth is massively skewed in favour of saving bandwidth, since the cache can be primed once and then used many times.

Decompression performance is not affected by this, so clients won't experience any performance difference, although they will have less bandwidth used.

### API changes
None.

### Behavioural changes
If the `zopfli` extension is present and `network.zopfli-chunk-compression` is `true`, ZOPFLI compression will be used to compress chunks.

Please be aware that this IS NOT intended for most servers. If you do any kind of building or otherwise modify the map (such as on a creative server) this will just cause a huge loss of performance because the cache in such circumstances has to constantly be recreated.

This is expected to be best in lobby server maps where the world cannot be modified, so that the cache can be primed once and used lots of times.

## Tests
Has been tested using a 1.5.0 client.

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -71,6 +71,7 @@ use pocketmine\nbt\tag\LongTag;
 use pocketmine\nbt\tag\ShortTag;
 use pocketmine\nbt\tag\StringTag;
 use pocketmine\network\AdvancedNetworkInterface;
+use pocketmine\network\mcpe\ChunkRequestTask;
 use pocketmine\network\mcpe\CompressBatchedTask;
 use pocketmine\network\mcpe\NetworkCipher;
 use pocketmine\network\mcpe\NetworkCompression;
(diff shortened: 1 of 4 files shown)


dktapps:
A possible improvement to this could be to make this optional per-world instead of globally (for multi-world sake), but I'm doubtful about mixing network settings with world settings.

JUSTaMod:
If it matters - I'd prefer to keep it per server, not per world.

lukeeey:
@JUSTaMod that would have no benefit whatsoever

dktapps:
Multi world is an anti-feature anyway, so I tend to agree with @JUSTaMod ... but it has less flexibility given that the anti-feature exists.

dktapps:
I don't plan to work on this any further since it's only useful in extremely specialized cases, and I have little interest in doing anything further with it at this point.

PR #2370, opened 2018-08-12
Fireworks

Frago9876543210:
## Introduction
I think it's all clear from the title.

### API changes
For example:
```php
public function onClick(PlayerInteractEvent $e) : void{
	if($e->getAction() === $e::RIGHT_CLICK_BLOCK && $e->getItem()->isNull()){
		$item = new Fireworks;
		$item->setNamedTag(Fireworks::createNBT(new FireworksData(1, [
			new FireworksExplosion(Fireworks::COLOR_ORANGE, Fireworks::COLOR_CYAN, true, true, Fireworks::TYPE_STAR_SHAPED)
		])));
		$e->getPlayer()->getInventory()->addItem($item);
	}
}
```

image


src/pocketmine/item/Fireworks.php
```diff
@@ -0,0 +1,104 @@
...
+
+		$rocket = new FireworksRocket($player->level, $nbt, $player, $this, $random);
+		if($player->isSurvival()){
+			--$this->count;
+		}
+		$rocket->spawnToAll();
+		return true;
+	}
+
+	public static function createNBT(FireworksData $fireworksData) : CompoundTag{
```
dktapps: This appears to be dead code.

src/pocketmine/item/FireworksData.php
```diff
@@ -0,0 +1,55 @@
...
+ * @link http://www.pocketmine.net/
+ *
+ *
+*/
+
+declare(strict_types=1);
+
+namespace pocketmine\item;
+
+class FireworksData{
```
dktapps: This appears to be dead code.

src/pocketmine/item/FireworksExplosion.php
```diff
@@ -0,0 +1,88 @@
...
+ * @link http://www.pocketmine.net/
+ *
+ *
+*/
+
+declare(strict_types=1);
+
+namespace pocketmine\item;
+
+class FireworksExplosion{
```
dktapps: This appears to be dead code.

src/pocketmine/item/FireworksExplosion.php
```diff
@@ -0,0 +1,88 @@
...
+	 */
+	public function getFade() : string{
+		return $this->fade !== null ? chr($this->fade) : "";
+	}
+
+	/**
+	 * @return int
+	 */
+	public function isFlickering() : int{
+		return (int) $this->flicker;
```
dktapps: This is very confusing. A plugin developer using this code will not understand why this doesn't return a boolean.

src/pocketmine/item/FireworksExplosion.php
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	/**
+	 * @return int
+	 */
+	public function hasTrail() : int{
+		return (int) $this->trail;
```
dktapps: Same as above.

src/pocketmine/entity/projectile/FireworksRocket.php
```diff
@@ -0,0 +1,100 @@
...
+		parent::__construct($level, $nbt, $shootingEntity);
+	}
+
+	protected function initEntity(CompoundTag $nbt) : void{
+		parent::initEntity($nbt);
+		$random = $this->random ?? new Random;
+
+		$this->setGenericFlag(self::DATA_FLAG_HAS_COLLISION, true);
+		$this->setGenericFlag(self::DATA_FLAG_AFFECTED_BY_GRAVITY, true);
+		$this->getDataPropertyManager()->setItem(self::DATA_MINECART_DISPLAY_BLOCK, $this->item);
```
dktapps: yes, this reads well indeed...

dktapps (commented):
This appears to be incomplete. There are several dead code paths and the crafting capability is not present. As is, this would have taken me 5 minutes to do myself because the meat isn't here.

src/pocketmine/entity/projectile/FireworksRocket.php
Frago9876543210: wdym?

Muqsit: code readability

dktapps:
Some additional comments on why this was closed (aside from being stale) is that it is far too dependent on Mojang internals for me to feel comfortable touching it. We've seen how easily such things catch fire in the last year or so.

PR #2371, opened 2018-08-12
Removed useless strtolower calls

Frago9876543210:
## Introduction
This PR removes useless strtolower calls, because it is will called earlier.
https://github.com/pmmp/PocketMine-MP/blob/078a2486b5e56fb2e9b4f709d566b255872b80db/src/pocketmine/Server.php#L728-L737

### DIFF
--- a/src/pocketmine/OfflinePlayer.php
+++ b/src/pocketmine/OfflinePlayer.php
@@ -46,7 +46,7 @@ class OfflinePlayer implements IPlayer, Metadatable{
 	public function __construct(Server $server, string $name){
 		$this->server = $server;
 		$this->name = $name;
-		$this->iusername = strtolower($this->name);
+		$this->iusername = strtolower($name);
 		if(file_exists($this->server->getDataPath() . "players/" . $this->iusername . ".dat")){
 			$this->namedtag = $this->server->getOfflinePlayerData($this->name);
 		}else{
--- a/src/pocketmine/OfflinePlayer.php
+++ b/src/pocketmine/OfflinePlayer.php
@@ -32,6 +32,8 @@ class OfflinePlayer implements IPlayer, Metadatable{
 
 	/** @var string */
 	private $name;
+	/** @var string */
+	private $iusername;
 	/** @var Server */
 	private $server;
 	/** @var CompoundTag|null */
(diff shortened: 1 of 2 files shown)


src/pocketmine/OfflinePlayer.php
```diff
@@ -44,7 +44,7 @@ class OfflinePlayer implements IPlayer, Metadatable{
 	public function __construct(Server $server, string $name){
 		$this->server = $server;
 		$this->name = $name;
-		if(file_exists($this->server->getDataPath() . "players/" . strtolower($this->getName()) . ".dat")){
+		if(file_exists($this->server->getDataPath() . "players/" . $this->name . ".dat")){
```
dktapps: It's possible that an `OfflinePlayer` instance may be constructed by something else (such as a unit test), in which case the name might not be passed in lowercase. The class should not rely on the caller to do its work for it. At the least this check should be `strtolower()`d.

dktapps (changes requested):
Looks OK apart from one nitpick.

src/pocketmine/OfflinePlayer.php
```diff
@@ -43,7 +43,7 @@ class OfflinePlayer implements IPlayer, Metadatable{
 	 */
 	public function __construct(Server $server, string $name){
 		$this->server = $server;
-		$this->name = $name;
+		$this->name = strtolower($name);
```
dktapps: I was talking about the `file_exists()` check. The case of the passed name should be preserved on the property.

src/pocketmine/OfflinePlayer.php
```diff
@@ -44,7 +46,8 @@ class OfflinePlayer implements IPlayer, Metadatable{
 	public function __construct(Server $server, string $name){
 		$this->server = $server;
 		$this->name = $name;
-		if(file_exists($this->server->getDataPath() . "players/" . strtolower($this->getName()) . ".dat")){
+		$this->iusername = strtolower($this->name);
```
jasonwynn10: use `$name` here instead of `$this->name`

jasonw4331 (changes requested):
Here's my nitpick

jasonw4331 (approved):

lukeeey:
i thought i already fixed this in my PR?

Frago9876543210:
BTW, the problem is not just strtolower calls. Look at the that code:
```php
public function getMoney(IPlayer $player) : int{
	return (int) $this->money->get(strtolower($player->getName()), self::DEFAULT_MONEY);
}
```
Sometimes, it is quite convenient to use IPlayer as argument. However, if the IPlayer will be OfflinePlayer it return name in lower case, but if the player is online, then his name in will be returned without lower case. This PR fixes this. And adds to the interface `getLowerCaseName()`

PR #2376, opened 2018-08-16
Added CommandEvent, deprecated (Remote)?ServerCommandEvent

SOF3:
## Introduction

Currently, plugins have to handle ServerCommandEvent and PlayerCommandPreprocessEvent separately to catch all commands by players and console/RCON. Not only is this inconvenient, but this also prevents plugins from catching events from other CommandSender types. This pull request provides a generic event `pocketmine\event\server\CommandEvent` that covers all CommandSender types.

### API changes

- Added `pocketmine\event\server\CommandEvent`
- Added an optional `$internal` parameter to `pocketmine\Server::dispatchCommand()`, which prevents `CommandEvent` from getting called
- Deprecated `pocketmine\event\server\{ServerCommandEvent, RemoteServerCommandEvent}`

### Behavioural changes

_nil_

## Backwards compatibility

Two events have been deprecated. This does not (yet) have any impact on backward compatibility.

## Follow-up

_nil_

## Tests

_nil_

### DIFF
--- a/src/pocketmine/event/server/RemoteServerCommandEvent.php
+++ b/src/pocketmine/event/server/RemoteServerCommandEvent.php
@@ -27,6 +27,8 @@ use pocketmine\command\CommandSender;
 
 /**
  * This event is called when a command is received over RCON.
+ *
+ * @deprecated Use CommandEvent instead.
  */
 class RemoteServerCommandEvent extends ServerCommandEvent{
 
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -39,6 +39,7 @@ use pocketmine\event\HandlerList;
 use pocketmine\event\level\LevelInitEvent;
 use pocketmine\event\level\LevelLoadEvent;
 use pocketmine\event\player\PlayerDataSaveEvent;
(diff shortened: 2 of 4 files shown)


dktapps (approved):
This has been one part of command handling that has irritated me for a long time.

I suggest that the deprecated events should be removed in 4.0.

SOF3:
For those wondering: (Remote)?ServerCommandEvent is kept only for backward compatibility concerns. Updates and new plugins should switch to CommandEvent as of 3.2.0. Therefore, whether ServerCommandEvent or CommandEvent is called first will not be documented.

src/pocketmine/event/server/RemoteServerCommandEvent.php
```diff
@@ -27,6 +27,8 @@
 
 /**
  * This event is called when a command is received over RCON.
+ *
+ * @deprecated Use CommandEvent instead.
```
dktapps: will this trigger the appropriate deprecation warnings when a plugin registers a handler for this?

SOF3: Not tested, but I think so

PR #2377, opened 2018-08-16
A Primer for Mobs

95CivicSi:
# This PR is not complete.

## Introduction

Many users of PocketMine-MP expect to have basic mobs spawning for survival type scenarios.  The demand for a plugin that adds this feature has stayed fairly high.  The current mob spawning solutions are either in dire need of updates or do not provide AI movement which leaves much to be desired for the end users

As the current maintainer for one of these, it has become more and more apparent that PMMP may be ready to handle MobAI at some level.  The intent of this PR is to provide the basic requirements for spawning all creatures, and act as an RFC to determine how much functionality should be added in this (or another PR).

I would expect at minimum to implement the following.
 - [ ] Add missing class files for mobs.
   - [x] Include minimum properties for spawning (ID, height, width)
   - [ ] Include item drops
   - [ ] Include XP drops
   - [ ] Set correct max health.
 - [ ] Register all mobs to make them available to plugins.
 - [ ] Implement the /summon command.
 - [ ] Add/Fix Spawn Eggs
 - [ ] Give mobs a very basic AI for general movement (no path finding, following, or other special movement)

I feel that the addition of these items will give a good foundation for other plugins to implement the additional missing features. Please leave comments and suggestions.


dktapps:
this doesn't look so much like "implement mob AI" as "implement anything remotely related to mobs" ... all in one PR, no less.

As we discussed on Discord there is a significant cost associated with big contributions. Most of these changes you're suggesting do not require other things you're targeting for the same PR and as such shouldn't be in this PR - i.e. they are independent features. For example mob spawners do not require all mobs to be registered, and nor do they require that mobs have any AI. Bundling lots of changes into the same PR makes it extremely difficult to work with.

I have zero intention of merging any mega "implement all the things" PR for this reason. Please break down the changes into atomic features and PR them separately.

95CivicSi:
*quote 95CivicSi*
```
The intent of this PR is to provide the basic requirements for spawning all creatures, and act as an RFC to determine how much functionality should be added in .
```
Should the RFC part of this be separated into it's own PR and the title / OP be updated appropriately?

95CivicSi:
Closing this as it contains a change set that is too large to be implemented.

PR #2381, opened 2018-08-16
PluginManager: Add warnings when a plugin declares useless minor/patc…

dktapps:
…h API versions

## Introduction
Far too often I see people writing long lists of API versions like
`3.0.0, 3.0.1, 3.0.2, 3.0.3, 3.0.4 ...` and so on. It seems that the information on how the API versioning system checks compatibility has completely flown over the heads of most people.

This PR adds warnings on plugin load to inform the user that the plugin is declaring useless API versions.

### Relevant issues
No GitHub issues, but many many many Poggit plugin rejections.

### API changes
Added the following methods:
- `PluginManager::compareApi()`
- `PluginManager::findRedundantApis()`

### Behavioural changes
A warning will now be emitted when useless API versions are declared in the plugin manifest.

## Backwards compatibility
This does not pose any BC breaks, but it is a feature addition and as such targets 3.2.

## Follow-up
The language strings for these warnings need to be translated - currently they are hardcoded in English.

## Tests
Has been tested with compliant and non-compliant plugins. Original loading behaviour should be preserved, the only difference should be with the warning emission on bad plugins.

The following `plugin.yml` was used to test:
```yaml
name: test
main: dktapps\test\Main
version: 0.0.1
api: [2.0.0, 2.1.0, 3.0.4, 3.0.0, 3.0.1, 3.0.2, 3.0.3, 3.1.0, 4.0.0, 4.0.1]
load: POSTWORLD
commands:
 hand:
  description: "Skeleton plugin example command"
  usage: "/examplecommand"
  permission: test.hand
permissions:
 test:
  default: true
  description: "Example root permission node"
  children:
   hand:
    default: true
    description: "Allows the use of /examplecommand"
```
The output:
```
[Server thread/WARNING]: test declares redundant API versions: 2.1.0 (already included with 2.0.0), 3.0.1, 3.0.2, 3.0.3, 3.0.4, 3.1.0 (already included with 3.0.0), 4.0.1 (already included with 4.0.0)
```

### DIFF
--- a/src/pocketmine/plugin/PluginManager.php
+++ b/src/pocketmine/plugin/PluginManager.php
@@ -248,6 +248,14 @@ class PluginManager{
 								continue;
 							}
 
+							if(!empty($redundant = self::findRedundantApis(...$description->getCompatibleApis()))){
+								$messages = [];
+								foreach($redundant as $base => $duplicates){
+									$messages[] = implode(", ", $duplicates) . " (already included with $base)";
+								}
+								$this->server->getLogger()->warning($description->getName() . " declares unnecessary API versions: " . implode(", ", $messages));
+							}
+
 							if(count($pluginMcpeProtocols = $description->getCompatibleMcpeProtocols()) > 0){
 								$serverMcpeProtocols = [ProtocolInfo::CURRENT_PROTOCOL];
 								if(count(array_intersect($pluginMcpeProtocols, $serverMcpeProtocols)) === 0){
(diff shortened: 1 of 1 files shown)


src/pocketmine/plugin/PluginManager.php
```diff
@@ -362,39 +370,80 @@ public function loadPlugins(string $directory, array $newLoaders = null){
...
+	/**
+	 * Searches the given list of API versions and filters out unnecessary ones.
+	 *
+	 * @param string ...$versions
+	 *
+	 * @return string[][] base version (string) => unnecessary extra versions (string[])
+	 */
+	public static function findRedundantApis(string ...$versions) : array{
+		$redundant = [];
+		usort($versions, 'version_compare');
```
SOF3: Not sure if `version_compare` really works.

src/pocketmine/plugin/PluginManager.php
```diff
@@ -362,39 +370,80 @@ public function loadPlugins(string $directory, array $newLoaders = null){
...
-				$pluginNumbers = array_map("intval", array_pad(explode(".", $pluginApi[0]), 3, "0")); //plugins might specify API like "3.0" or "3"
+	/**
+	 * Returns whether a given API version is compatible with the required version.
+	 *
+	 * @param string $haveVersion
+	 * @param string $needVersion
+	 *
+	 * @return bool
+	 */
+	public static function compareApi(string $haveVersion, string $needVersion) : bool{
```
SOF3: "Compare" is inaccurate. You aren't comparing versions here; you are checking if they are compatible. Consider renaming this function (and maybe `isCompatibleApi` too)

SOF3 (changes requested):

src/pocketmine/plugin/PluginManager.php
dktapps: `isCompatibleApi` can't be renamed here due to BC concerns.

src/pocketmine/plugin/PluginManager.php
dktapps: It's a bit more loose than I'm happy with, but our version format works with `version_compare()` just fine.

SOF3: I'd rather you add an actual compare function that wraps version_compare(). It's too easy to miss this in the future in case we change the versioning scheme somehow such that it isn't version_compare()-compatible.

jackthehack21:
This sounds very usefull, love to see this in the future

dktapps:
Perhaps this would be more effective if it made the plugin not load instead.

My reasoning for this is that the server cannot identify the actual min required version when there are multiple compatible versions specified, so it should err on the side of caution by not loading the plugin is newer versions are found.

SOF3:
Agreed. However this would become a backward-incompatible change and should target 4.0.0

dktapps:
Meh. This is discouraged behaviour anyway, and doesn't cause backwards incompatibility unless you were doing stupid shit anyway.

src/pocketmine/plugin/PluginManager.php
```diff
@@ -248,6 +248,14 @@ public function loadPlugins(string $directory, array $newLoaders = null){
 								continue;
 							}
 
+							if(!empty($redundant = self::findRedundantApis(...$description->getCompatibleApis()))){
+								$messages = [];
+								foreach($redundant as $base => $duplicates){
+									$messages[] = implode(", ", $duplicates) . " (already included with $base)";
+								}
+								$this->server->getLogger()->warning($description->getName() . " declares unnecessary API versions: " . implode(", ", $messages));
```
dktapps: This might be difficult to create translations for.

SOF3: Meanwhile let's have language-specific punctuation. It should be `、` instead of `, ` in Chinese! 😛

dktapps:
Retargeting this to 4.0 since 3.x line is in feature freeze now.

PR #2383, opened 2018-08-17
Added a hack for MC W10 form JSON empty strings bug

dktapps:
## Introduction
Minecraft Windows 10 Edition is stupid and sends empty strings inside form response JSON without quotes, for example `["hello",,"sunshine"]` if the second input field was left empty.

This pull request adds a hack to fix this bug.

### Relevant issues
None on our issue tracker because this is an unreleased version, but this has been a bug that's plagued third party forms implementations for a while.

### Behavioural changes
Form responses from MCW10 containing empty strings _should_ now decode correctly.

## Tests
Tested with the following JSON blobs:
```
[\n   \"a\",\"b,c,d,e\\\"   \",,0,1,2, false, 0.001]
0
false
NULL
["\",,\"word","a\",,\"word2",]
```

Please test this and try and break it however you can.

### DIFF
--- a/src/pocketmine/network/mcpe/PlayerNetworkSessionAdapter.php
+++ b/src/pocketmine/network/mcpe/PlayerNetworkSessionAdapter.php
@@ -239,7 +239,37 @@ class PlayerNetworkSessionAdapter extends NetworkSession{
 	}
 
 	public function handleModalFormResponse(ModalFormResponsePacket $packet) : bool{
-		return $this->player->onFormSubmit($packet->formId, json_decode($packet->formData, true));
+		return $this->player->onFormSubmit($packet->formId, self::stupid_json_decode($packet->formData, true));
+	}
+
+	/**
+	 * Hack to work around a stupid bug in Minecraft W10 which causes empty strings to be sent unquoted in form responses.
+	 *
+	 * @param string $json
+	 * @param bool   $assoc
+	 *
+	 * @return mixed
+	 */
+	private static function stupid_json_decode(string $json, bool $assoc = false){
+		if(preg_match('/^\[(.+)\]$/s', $json, $matches) > 0){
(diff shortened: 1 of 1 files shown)


src/pocketmine/network/mcpe/PlayerNetworkSessionAdapter.php
```diff
@@ -239,7 +239,37 @@ public function handleBookEdit(BookEditPacket $packet) : bool{
...
+
+	/**
+	 * Hack to work around a stupid bug in Minecraft W10 which causes empty strings to be sent unquoted in form responses.
+	 *
+	 * @param string $json
+	 * @param bool   $assoc
+	 *
+	 * @return mixed
+	 */
+	private static function stupid_json_decode(string $json, bool $assoc = false){
```
Frago9876543210: Why not to make it public? Let other plugins fix it too

dktapps: There's no need for other plugins to fix it.

Frago9876543210: Why? Many have not yet switched from custom forms api to pm api

dktapps: This is a solution for the new API of 3.2. Plugins should be upgrading to use the new API anyway.

dktapps: By "new API" I'm referring to the `Form` interface and `Player->sendForm()`. Here's an example changeset: https://github.com/jojoe77777/FormAPI/pull/41

Frago9876543210: How about third party APIs?

dktapps: > Plugins should be upgrading to use the new API anyway.

PR #2393, opened 2018-08-21
Enable ANSI support for git bash

Frago9876543210:
## Introduction
Before:
default
After:
default

## Tests
Tested on git bash and Debian 9.

### DIFF
--- a/src/pocketmine/utils/Terminal.php
+++ b/src/pocketmine/utils/Terminal.php
@@ -64,7 +64,7 @@ abstract class Terminal{
 						getenv('TERM') !== false or //Console says it supports colours
 						(function_exists('sapi_windows_vt100_support') and sapi_windows_vt100_support($stdout)) //we're on windows and have vt100 support
 					)
-				));
+				) or getenv('MSYSTEM') !== false);
 				fclose($stdout);
 			}
 


dktapps:
this change is now pointless since you can simply write `start.sh --enable-ansi`.

Frago9876543210:
I don't think it's convenient

dktapps:
this is all missing the point... the fault here is that the colour support is not being correctly auto-detected. As I said, look at `Terminal::hasFormattingCodes()`.

dktapps:
It appears that MSYS acts as a pipe, like lots of other console applications. Colour codes are automatically disabled when it's detected that stdout is a pipe.

Frago9876543210:
Sometimes it is not piped
default

SOF3:
Note: using the `winpty` command in Git Bash will also show colors.

dktapps:
There isn't any cover all solution here that won't cause bugs. If this change is adopted, running `./start.sh > somewhere_else` will cause problems with character encoding when writing to a file (which is why it's disabled when a pipe is detected to begin with).

lukeeey:
Add a startup option to force colours

dktapps:
such a thing already exists 👏

lukeeey:
Oh well I don't use this abomination

SOF3:
Use an alias.

```bash
alias pm='./start.sh --enable-ansi`
```

PR #2412, opened 2018-08-29
Fixed crash on join game

Enes5519:
## Introduction
Fixed crash on join game

### API changes

**PlayerListEntry** remove third party and platform property and fields

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -2355,7 +2355,7 @@ class Server{
 		$pk = new PlayerListPacket();
 		$pk->type = PlayerListPacket::TYPE_ADD;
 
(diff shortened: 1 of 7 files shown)


dktapps:
This code is wrong. Only 1.7 has these fields removed.

Enes5519:
Hmm... I don't join game on 1.6 but I take crash but I do this. I join now

dktapps:
Nobody else has reported issues, and this code is still wrong... you must be a special snowflake. ¯\_(ツ)_/¯

Enes5519:
Wow. I'm amazed.

dktapps:
I've asked someone else to independently check this. Here's the code: https://gist.github.com/MisteFr/368891d2d07d9f165f58499c225d549d

Thank you for wasting my time.

PR #2414, opened 2018-08-30
Implemented rail connectivity

dktapps:
## Introduction
I don't need to explain why this is necessary, do I?

### Behavioural changes
Rails (all types) now automatically join up in a vanilla-like manner when placed.

There are known issues with curved rails not following the same precedence as vanilla when there are 3+ possible connection directions. However this is not considered a major issue and is likely just an artifact of Mojang's implementation (to which this is nothing alike) as opposed to a by-design feature.

## Follow-up
Implement minecarts.

## Tests
has been extensively tested in-game, many bugs were found. I can't easily upload images so I will assume reviewers are capable of placing rails for themselves to test.


lukeeey:
Omg, this commit is so epic that i am beginning to go [off the rails](https://www.urbandictionary.com/define.php?term=off%20the%20rails)

dktapps:
don't get too excited, there's a few ... _corner cases_ shall we say

PR #2425, opened 2018-09-03
able to bypass that checks block's Id in Level::useItemOn()

djdisodo:
## Introduction
i was making Hide and Seek minigame
i sent block on each players position(that's fake block used sendBlock)
but i couldn't make player get damaged on hit blocks(without handling packets)

### Relevant issues
able to bypass that checks block's Id in Level::useItemOn()


Muqsit:
Don't use `Level::sendBlocks()` if you don't know what it does?

dktapps:
this is undefined behaviour and as such isn't considered "API"... closing

PR #2428, opened 2018-09-10
Strip anti flight out of the core

dktapps:
## Introduction
Has been on the roadmap for a long time (see associated project).

This may later be developed into a plugin if it is a desired feature, but having it in the core is unnecessary and a pain in the ass.

I think it is reasonable to assume that anyone who is serious about preventing fly hackers is not using this excuse for an anti-cheat, and is instead using their own or a different outsourced implementation.

### API changes
- API method `Server->getAllowFlight()` is now deprecated.

## Tests
Has been tested in-game. A plugin which will replicate this functionality is under development.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -284,14 +284,10 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 
 	/** @var Vector3|null */
 	protected $newPosition;
-	/** @var Vector3|null */
-	public $speed = null;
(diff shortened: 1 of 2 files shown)


matcracker (approved):

jasonw4331 (approved):

Driesboy:
Thanks, Now I can get rid of the hacks to disable it ;)

T14Raptor:
its not anticheat smart one

PR #2429, opened 2018-09-10
Level: Batch light updates at the end of the tick to amortize CPU cost

### LINKED ISSUE #456: On-the-fly block light updating has serious performance issues
### Issue description
Many people will have noticed the crazy numbers on timings next to UseItemPacket and RemoveBlockPacket. This is due to synchronous on-the-fly block light updates when light-emitting blocks are placed or removed.

I added timings locally to monitor the extent of the time taken to execute Level->setBlock() and do lighting updates. An example timings report can be seen [here](https://timings.imgcl.co/?url=24240666). While these timings were being executed I placed and removed maybe half-a-dozen torches in a flat world.

Raw timings results on my machine showed that on average, placing a torch with no nearby light sources caused the server to spend ~90ms updating nearby block light, and ~50ms removing block light.

Based on other timings results I believe it may actually be significantly faster to fast-serialize up to 9 adjacent chunks and perform these operations in an AsyncTask. 

1. place or remove any light-emitting block
2. see CPU spike on that tick

### OS and versions

* PocketMine-MP: https://github.com/pmmp/PocketMine-MP/pull/454 with local additions to add timings for setBlock and lighting updates
* PHP: 7.0.16
* Server OS: w10
* Game version: not relevant

dktapps:
## Introduction
this produces a 5x performance improvement for lighting updates during water flow, and 25% improvement for lava flow.

### Relevant issues
closes #456 

### Behavioural changes
`setBlock()` no longer performs updated light propagation directly. Instead it's added to a queue and updated at the end of the tick in batch. This has major performance advantages when performing block setting en masse. A vanilla example of this is liquid spread.

## Follow-up
Work on parallelizing and further optimizing light updates.

## Tests
Has been spot checked using a debug plugin with light updates caused by liquids. It has not been extensively tested, but the changes do not functionally affect lighting updates, they just amortize the CPU cost by batching.

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -247,6 +247,11 @@ class Level implements ChunkManager, Metadatable{
 	/** @var bool */
 	private $closed = false;
 
+	/** @var BlockLightUpdate|null */
+	private $blockLightUpdate = null;
+	/** @var SkyLightUpdate|null */
+	private $skyLightUpdate = null;
+
 	public static function chunkHash(int $x, int $z) : int{
 		return (($x & 0xFFFFFFFF) << 32) | ($z & 0xFFFFFFFF);
 	}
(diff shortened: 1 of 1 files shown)


dktapps:
It's possible that the same block may be set multiple times in a tick. If this happens, this code will crash when it tries to update the same location multiple times.

PR #2444, opened 2018-09-17
Added Dark Oak Tree 

Iam2smart:
## Introduction

Fixes Dark Oak Trees I hope


Iam2smart:
I Do not know if this will work i am new so yeah code is bad but should work.

dktapps:
have you ever seen a dark oak tree in game? Hint: it's not this simple...

Iam2smart:
Yes I have seen what a Dark Oak Tree Looks like that why i said it was partial done

lukeeey:
I don't get it how is it different other trees

PR #2447, opened 2018-09-18
MainLogger::logException() now logs inducing exceptions

SOF3:
## Introduction
PocketMine implements its own error logging mechanism, which does not log indirect exception causes. This may prevent some errors from being properly displayed.

### Relevant issues

nil

### API changes

Added a `$depth` function to `Utils::printableTrace`

### Behavioural changes

Logged exception causes are now displayed in this manner:

```
$exception
$exception->trace
Caused by:
  $exception->previous
  $exception->previous->trace
  Caused by: $exception->previous->previous
    $exception->previous->previous->trace
...
```

## Backwards compatibility

MainLogger is not supposed to be extended, so I consider there to be no backward compatibility issues.

## Follow-up
This is not planning to get backported to 3.x due to extensive API changes across majors.

## Tests
See https://github.com/pmmp/PocketMine-MP/pull/2447#issuecomment-565790033

### DIFF
--- a/src/utils/MainLogger.php
+++ b/src/utils/MainLogger.php
@@ -153,6 +153,37 @@ class MainLogger extends \AttachableThreadedLogger{
 	 * @param array|null $trace
 	 */
 	public function logException(\Throwable $e, $trace = null){
+		$messages = [];
+		self::logExceptionToArray($e, $trace, 0, $messages);
+
+		$this->synchronized(function() use ($messages) : void{
+			foreach($messages as $depth => [$message, $stack]) {
+				if($depth === 0){
+					$this->critical($message);
+				}else{
+					$this->debug(str_repeat("  ", $depth - 1) . "Caused by:", true);
+					$this->debug(str_repeat("  ", $depth) . $message, true);
+				}
+				foreach($stack as $line){
+					$this->debug($line, true);
+				}
+			}
+		});
+
+		$this->syncFlushBuffer();
+	}
+
+	/**
+	 * Appends exceptions messages recursively to the referenced array.
+	 *
+	 * Each element of the array is a [string, string[]], containing the message and the trace lines.
+	 *
+	 * @param \Throwable                $e
+	 * @param array|null                $trace
+	 * @param int                       $depth
+	 * @param array<string|string[]> &$ret
+	 */
+	private static function logExceptionToArray(\Throwable $e, $trace, int $depth, array &$ret) : void{
 		if($trace === null){
 			$trace = $e->getTrace();
 		}
@@ -170,16 +201,14 @@ class MainLogger extends \AttachableThreadedLogger{
 		$errfile = Filesystem::cleanPath($errfile);
 
 		$message = get_class($e) . ": \"$errstr\" ($errno) in \"$errfile\" at line $errline";
-		$stack = Utils::printableTrace($trace);
+		$stack = Utils::printableTrace($trace, 80, $depth);
 
-		$this->synchronized(function() use ($message, $stack) : void{
-			$this->log(LogLevel::CRITICAL, $message);
-			foreach($stack as $line){
-				$this->debug($line, true);
-			}
-		});
+		$ret[] = [$message, $stack];
 
-		$this->syncFlushBuffer();
+		$inner = $e->getPrevious();
+		if($inner !== null){
+			self::logExceptionToArray($inner, null, $depth + 1, $ret);
+		}
 	}
 
 	public function log($level, $message){
--- a/src/utils/Utils.php
+++ b/src/utils/Utils.php
@@ -409,10 +409,11 @@ class Utils{
 	/**
 	 * @param array $trace
 	 * @param int   $maxStringLength
+	 * @param int   $depth           each depth level indents all resultant messages by 2 spaces
 	 *
 	 * @return array
 	 */
-	public static function printableTrace(array $trace, int $maxStringLength = 80) : array{
+	public static function printableTrace(array $trace, int $maxStringLength = 80, int $depth = 0) : array{
 		$messages = [];
 		for($i = 0; isset($trace[$i]); ++$i){
 			$params = "";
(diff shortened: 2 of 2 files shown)


src/pocketmine/utils/MainLogger.php
```diff
@@ -209,20 +222,20 @@ public function logException(\Throwable $e, $trace = null){
 		$errstr = preg_replace('/\s+/', ' ', trim($errstr));
 		$errfile = Utils::cleanPath($errfile);
 
-		$message = get_class($e) . ": \"$errstr\" ($errno) in \"$errfile\" at line $errline";
+		$message = $induced ? "Caused by: " : "";
+		$message .= get_class($e) . ": \"$errstr\" ($errno) in \"$errfile\" at line $errline";
+		yield [$type, $message];
```
dktapps: Don't think a type is necessary here. Previous exceptions can be logged in debug the same as the rest. Having it as its own error will make it confusing to read, since it's hard to tell if the debug blocks are part of the same log error.

Additionally, consider referring to this as "previous" instead of "caused by" to reduce ambiguity when there are multiple chained exceptions.

dktapps: Also, it's possible to yield the message type as key if you want to keep it.

src/pocketmine/utils/MainLogger.php
```diff
@@ -174,6 +174,19 @@ public function setLogDebug(bool $logDebug){
 	 * @param array|null $trace
 	 */
 	public function logException(\Throwable $e, $trace = null){
+		$generator = $this->logExceptionMessages($e, $trace, false);
+		$messages = iterator_to_array($generator, false);
```
dktapps: I'm not very clear why this is a generator to begin with, since it's only used here.

src/pocketmine/utils/MainLogger.php
SOF3: About the key: I use iterator_to_array() to prevent running the message generation in the synchronized() block, so I'm not using the key here.

src/pocketmine/utils/MainLogger.php
dktapps: You can use `array_push($target, ...$array2)`

SOF3: That would result in O(n^2) performance, which is at the same complexity class of array_merge(). See this benchmark (we are push-right here): https://gist.github.com/SOF3/988c408a92f79a644cc3fbc65727ca8d

Using array_unshift() is possible, but too complicated for such a simple feature where optimization isn't really critical. (You don't have an exception with 100 chained causes every day; if you do, you don't want to log it anyway)

src/pocketmine/utils/MainLogger.php
SOF3: Since iterator_to_array() has been removed, using the type as key is more convenient (but it's the same thing anyway!)

SOF3:
This pull request hsa been rewritten against the latest master.

With this test plugin:

```php
<?php

/**
 * @name abc
 * @main Foo
 * @api 4.0.0
 * @version 4.0.0
 */

class Foo extends \pocketmine\plugin\PluginBase {
	public function onEnable() {
		$e = new \Exception("first", 0);
		$e = new \Exception("second", 0, $e);
		$this->getServer()->getLogger()->logException($e);
	}
}
```

Gives this result:

```
[08:55:49.225] [Server thread/CRITICAL]: Exception: "second" (EXCEPTION) in "plugins/abc" at line 13
[08:55:49.225] [Server thread/DEBUG]: #0 src/plugin/PluginBase(143): Foo->onEnable()
[08:55:49.225] [Server thread/DEBUG]: #1 src/plugin/PluginManager(435): pocketmine\plugin\PluginBase->onEnableStateChange(boolean 1)
[08:55:49.225] [Server thread/DEBUG]: #2 src/Server(1507): pocketmine\plugin\PluginManager->enablePlugin(object Foo)
[08:55:49.225] [Server thread/DEBUG]: #3 src/Server(1259): pocketmine\Server->enablePlugins(object pocketmine\plugin\PluginLoadOrder)
[08:55:49.225] [Server thread/DEBUG]: #4 src/PocketMine(264): pocketmine\Server->__construct(object BaseClassLoader, object pocketmine\utils\MainLogger, string[34] /home/sofe/proj/php/PocketMine-MP/, string[42] /home/sofe/proj/php/PocketMine-MP/plugins/)
[08:55:49.226] [Server thread/DEBUG]: #5 src/PocketMine(287): pocketmine\server()
[08:55:49.226] [Server thread/DEBUG]: Caused by:
[08:55:49.226] [Server thread/DEBUG]:   Exception: "first" (EXCEPTION) in "plugins/abc" at line 12
[08:55:49.226] [Server thread/DEBUG]:   #0 src/plugin/PluginBase(143): Foo->onEnable()
[08:55:49.226] [Server thread/DEBUG]:   #1 src/plugin/PluginManager(435): pocketmine\plugin\PluginBase->onEnableStateChange(boolean 1)
[08:55:49.226] [Server thread/DEBUG]:   #2 src/Server(1507): pocketmine\plugin\PluginManager->enablePlugin(object Foo)
[08:55:49.226] [Server thread/DEBUG]:   #3 src/Server(1259): pocketmine\Server->enablePlugins(object pocketmine\plugin\PluginLoadOrder)
[08:55:49.226] [Server thread/DEBUG]:   #4 src/PocketMine(264): pocketmine\Server->__construct(object BaseClassLoader, object pocketmine\utils\MainLogger, string[34] /home/sofe/proj/php/PocketMine-MP/, string[42] /home/sofe/proj/php/PocketMine-MP/plugins/)
[08:55:49.226] [Server thread/DEBUG]:   #5 src/PocketMine(287): pocketmine\server()
```

src/utils/MainLogger.php
```diff
@@ -153,6 +153,37 @@ public function setLogDebug(bool $logDebug) : void{
...
+
+	/**
+	 * Appends exceptions messages recursively to the referenced array.
+	 *
+	 * Each element of the array is a [string, string[]], containing the message and the trace lines.
+	 *
+	 * @param \Throwable                $e
+	 * @param array|null                $trace
+	 * @param int                       $depth
+	 * @param array<string|string[]> &$ret
```
SOF3: This doc comment is actually wrong. No idea why phpstan didn't pick it up.

Help needed with specifying tuples in phpdoc.

dktapps: phpstan doesn't like reference notation, it might need to be removed

PR #2456, opened 2018-09-23
Add missing corner checks for Stair BB calculation

dktapps:
## Introduction
Stair bounding box calculation did not take corners into account prior to this pull request. This can lead to strange effects with arrows (colliding with nothing or passing through seemingly solid parts of the block) and may also cause problems with movement anti-cheat.

### Behavioural changes
Stair bounding boxes now take corner wrapping into account. This now makes entities (most noticeably arrows) behave correctly when encountering a corner stair.

This targets master instead of 3.3 due to changes in the code responsible for handling this, so I am not pushing this to 3.3 because I don't want to deal with collisions.

## Tests
A test server is live at dev1.ams.pmmp.io:19132. This has been tested in game using arrows to determine collision points.
unknown


dktapps:
Merged in a77e05f875ac645557cd2c15c0e64b694946cf7b.

PR #2474, opened 2018-10-06
Torch: Fix placement issues closes #2453

### LINKED ISSUE #2453: Torch placement does not work correctly when clicking the bottom of a block
### Issue description
When clicking on the bottom of a block with a torch, the game will search for a wall face to attach the torch to below the clicked block, if it doesn't find a block directly below. We don't check this, but only check if there is a block directly below.

1. Try to place a torch at the top of a wall by clicking the ceiling.

Pugzy:
## Introduction
Update the torch placement method to allow for check of all target block sides.
- Torches can be placed on top and sides of block.
- Torches can be placed on top of walls and fences.
- When placed on the under side of a block all sides and bottom are evaluated.
- When placed on a transparent blocks, fence and walls all sides and bottom are also evaluated.

Happy for feedback as I understand it may not be the best way of doing it.

### Relevant issues
Fixes Issue: Torch placement does not work correctly when clicking the bottom of a block #2453 

## Follow-up
Following on from this commit update the blocks that torches can be placed on.
The [Minecraft Wiki](https://minecraft.gamepedia.com/index.php?title=Torch#Placement) states items such as glass and stained glass should be allowed however current method rules them out as transparent blocks.

Requires translations:

## Tests
Video showcasing a demo of the commit https://youtu.be/h3YnFWHcm6E including the issue it resolves.


Pugzy:
I see what you mean regarding the replaceable blocks. I'll investigate further.

Pugzy:
Previous change resolves the replaceable bug.

src/pocketmine/block/Torch.php
```diff
@@ -73,16 +73,28 @@ public function onNearbyBlockChange() : void{
...
-		$below = $this->getSide(Facing::DOWN);
-
-		if(!$blockClicked->isTransparent() and $face !== Facing::DOWN){
-			$this->facing = $face;
-			return parent::place($item, $blockReplace, $blockClicked, $face, $clickVector, $player);
-		}elseif(!$below->isTransparent() or $below->getId() === self::FENCE or $below->getId() === self::COBBLESTONE_WALL){
+		if($blockClicked->canBeReplaced() and !$blockClicked->getSide(Facing::DOWN)->isTransparent()){
 			$this->facing = Facing::UP;
 			return parent::place($item, $blockReplace, $blockClicked, $face, $clickVector, $player);
+		}elseif($face !== Facing::DOWN and (!$blockClicked->isTransparent() or ($face == Facing::UP and ($blockClicked->getId() === self::FENCE or $blockClicked->getId() === self::COBBLESTONE_WALL)))){
```
dktapps: nit: use `===` not `==`.

Pugzy: acceptable nit, resolved

PR #2481, opened 2018-10-16
Create Barrier Block...

Asutartix:
## Introduction

I'm not good at english...
So, i write in korean..

구동기에서 베리어 블럭이 추가가 되지 않아 추가 요청을 합니다...


dktapps:
If you had tested this, you would know it doesn't work. This pull request is therefore pointless since it does not make the necessary core changes to allow this to work as designed.

There is a reason why none of the extended blocks have been added yet.

PR #2496, opened 2018-10-27
Fix #2436

### LINKED ISSUE #2436: Doors don't drop when the block below them is destroyed
1. place a wooden door
2. destroy the block below it

### OS and versions
* PocketMine-MP: all current release/3.* branches + master
* PHP: 7.2.9
* Server OS: windows

daniktheboss:
This fixes the no-drop door issue


dktapps:
If I recall correctly, the bottom condition can be removed with this change, since `useBreakOn()` should delete both parts. You might want to test this (I don't have a computer handy right now).

dktapps:
Additionally note that this won't deal with the problem of iron doors since they require a tool to break, but that can't be fixed without a hack right now.

PR #2497, opened 2018-10-27
Added a ClosureTask implementation for easier task scheduling

dktapps:
## Problem
It's very common for plugins to use scheduled tasks to fire functions within themselves. Before PocketMine-MP API 2.0.0 (henceforth referred to as 2.0 for simplicity), we had `CallbackTask` which allowed plugins to do this. However, in 2.0 the `CallbackTask` was removed.

While `CallbackTask` was never intended for plugin usage, the usefulness of callables in this aspect is undeniable. So, meet the new `ClosureTask`.

Consider the old workflow:

```php
//Main.php
$this->getScheduler()->scheduleDelayedTask(new SimpleTask($this), 1);

//SimpleTask.php
class SimpleTask{
    private $plugin;

    public function __construct(Main $plugin){
        $this->plugin = $plugin;
    }

    public function onRun(int $currentTick) : void{
        $this->plugin->getServer()->broadcastMessage("Hello players");
    }
}
```
This is very annoying especially when the `onRun()` body is merely 1 line. This can be reduced by using anonymous classes, but still causes a lot of boilerplate.

## Solution
Wouldn't it be nice to just be able to do this...
```php
//Main.php
$this->getScheduler()->scheduleDelayedTask(new ClosureTask(function(int $currentTick) : void{
    $this->getServer()->broadcastMessage("Hello players");
}, 1);
```

Closures are extremely powerful. The main reason is because they retain the `this` scope where they were created. They can also keep variables from the call site by `use`ing them, like `function() use($var1, $var2, $var3)`.

### API changes
- Added a new `ClosureTask` class. The constructor accepts a `Closure` which should accept exactly ONE parameter `(int $currentTick)` and SHOULD NOT return a value.

## Tests
```php
<?php

declare(strict_types=1);

require 'vendor/autoload.php';

$sched = new \pocketmine\scheduler\TaskScheduler(new \SimpleLogger());

$sched->scheduleTask(new \pocketmine\scheduler\ClosureTask(function(int $currentTick) : void{
	echo "hello sunshine!" . PHP_EOL;
	echo $currentTick . PHP_EOL;
}));

$sched->mainThreadHeartbeat(0);
```
outputs
```
hello sunshine!
0
```

### Unsolved issues
Timings task names are a problem which is currently unsolved, since they'll all appear as a generic `ClosureTask`.

### DIFF
--- a/src/pocketmine/scheduler/ClosureTask.php
+++ b/src/pocketmine/scheduler/ClosureTask.php
@@ -30,6 +30,7 @@ use pocketmine\utils\Utils;
  *
  * Example usage:
  *
+ * ```
  * TaskScheduler->scheduleTask(new ClosureTask(function(int $currentTick) : void{
  *     echo "HI on $currentTick\n";
  * });
--- a/src/pocketmine/scheduler/ClosureTask.php
+++ b/src/pocketmine/scheduler/ClosureTask.php
@@ -34,6 +34,7 @@ use pocketmine\utils\Utils;
  * TaskScheduler->scheduleTask(new ClosureTask(function(int $currentTick) : void{
  *     echo "HI on $currentTick\n";
  * });
+ * ```
  */
 class ClosureTask extends Task{
 
--- a/src/pocketmine/scheduler/ClosureTask.php
+++ b/src/pocketmine/scheduler/ClosureTask.php
@@ -1,0 +1,50 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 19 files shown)


src/pocketmine/scheduler/ClosureTask.php
```diff
@@ -0,0 +1,50 @@
...
+namespace pocketmine\scheduler;
+
+/**
+ * Task implementation which allows closures to be called by a scheduler.
+ *
+ * Example usage:
+ *
+ * TaskScheduler->scheduleTask(new ClosureTask(function(int $currentTick) : void{
+ *     echo "HI on $currentTick\n";
+ * });
```
SOF3: This code snippet might not render properly in some documentation generators.

dktapps: I'm not sure how else to do it. Suggestions?

src/pocketmine/scheduler/ClosureTask.php
```diff
@@ -0,0 +1,56 @@
...
+namespace pocketmine\scheduler;
+
+use pocketmine\utils\Utils;
+
+/**
+ * Task implementation which allows closures to be called by a scheduler.
+ *
+ * Example usage:
+ *
+ * TaskScheduler->scheduleTask(new ClosureTask(function(int $currentTick) : void{
```
SOF3:
```suggestion
 * ```
 * TaskScheduler->scheduleTask(new ClosureTask(function(int $currentTick) : void{
```

src/pocketmine/scheduler/ClosureTask.php
```diff
@@ -0,0 +1,56 @@
...
+ *     echo "HI on $currentTick\n";
+ * });
```
SOF3:
```suggestion
 * });
 * ```
```

SOF3 (approved):

dktapps:
Considering extending this to allow any callable instead of just Closure objects. The implementation would be nearly identical to the current one.

Issue #2441, opened 2018-09-16
Half-doors can be placed at y=255


dktapps:
### Issue description
![image](https://user-images.githubusercontent.com/14214667/45598078-af94d100-b9cd-11e8-9d13-b7f4947f3ebb.png)

This is an image of /setblock in vanilla, but the same effect can be produced by using a door item at y=255 where there isn't enough space for the top half of the door.

When opening the door, the game crashes. With that said, it shouldn't be possible to place a door up here anyway.

### Steps to reproduce the issue
1. create a platform of blocks at y=254
2. place a door at y=255
3. notice a half-door is created

### OS and versions
* PocketMine-MP: f7f06c993d908b714feb3b10dd92a1f5b0bcad2b
* PHP: 7.2.9
* Server OS: windows

QimieGames:
It is possible to make a check on place? Like check whats the coords of the door is placed at. If its 255. then just cancel event for the placement. Just asking. Im new to this community

dktapps:
It's possible but not the best solution, since every double block would have to have such a check (double tallgrass and tall flowers for example), so this isn't a good solution..

PR #2498, opened 2018-10-28
Door: Prevent placement above height limit.

Pugzy:
## Introduction
Check and prevent if the top block of a door would be at or above the max world height once placed.

### Relevant issues
Fixes: Half-doors can be placed at y=255 #2441

## Tests
Placement now works the same as in vanilla.

### DIFF
--- a/src/pocketmine/block/Door.php
+++ b/src/pocketmine/block/Door.php
@@ -154,7 +154,7 @@ abstract class Door extends Transparent{
 		if($face === Facing::UP){
 			$blockUp = $this->getSide(Facing::UP);
 			$blockDown = $this->getSide(Facing::DOWN);
-			if(!$blockUp->canBeReplaced() or $blockDown->isTransparent()){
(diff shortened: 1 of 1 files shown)


nathfreder:
I am not sure about the actual source, because I am not a core maintainer. However, this was made so that he can get a t-shirt for Hacktoberfest. If this get's closed, you should add a `invalid` tag to prevent this from counting towards a reward.

Muqsit:
There are other blocks that are taller than a block, eg: double plants. I think the correct way would be by handling `Block::getAffectedBlocks()`.

Pugzy:
@Muqsit tested with double plans too, in vanilla they let the player place them and cut off the top half so no checks need to be made. Not too sure how `getAffectedBlocks` could assist with this but happy to modify and test.

@nathfreder no need for drama on a pull request, Hacktoberfest was created to encourage contributions towards open source projects, happy to discuss via Discord dm or likewise.

dktapps:
Thanks for the PR.

The issue I have with this is that it's a special-case exception, and won't prevent the bug happening for other multi-blocks in the future. It would be better to have a general solution for this, so I'm not rushing a merge of this right now.

PR #2508, opened 2018-10-31
Added PermissibleTrait

SOF3:
## Introduction
Implementing Permissible is based on PermissibleBase, but redirecting the method calls produces a lot of boilerplate code. This PR adds a trait to copy-paste this boilerplate.

### API changes

- Added PermissibleTrait

### Behavioural changes

none

## Backwards compatibility

None.
(Plugins that extend Player and set `$player->closed` incorrectly might lead to memory leaks, but "incorrect" is not something to consider for BC)

## Tests

Not yet tested

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -147,6 +147,7 @@ use pocketmine\network\mcpe\protocol\UpdateBlockPacket;
 use pocketmine\network\mcpe\VerifyLoginTask;
 use pocketmine\network\SourceInterface;
 use pocketmine\permission\PermissibleBase;
+use pocketmine\permission\PermissibleTrait;
 use pocketmine\permission\PermissionAttachment;
 use pocketmine\permission\PermissionAttachmentInfo;
 use pocketmine\permission\PermissionManager;
@@ -164,6 +165,10 @@ use pocketmine\utils\UUID;
  * Main class that handles networking, recovery, and packet sending to the server part
  */
 class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
+	use PermissibleTrait {
+		hasPermission as private _hasPermission;
+		recalculatePermissions as private _recalculatePermissions;
+	}
 
 	public const SURVIVAL = 0;
 	public const CREATIVE = 1;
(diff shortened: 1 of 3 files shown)


dktapps:
This feels a bit like it's missing the point as I mentioned earlier.. I understand why you're doing this (backwards compatibility) but it's just cosmetic really, and will make more work if we decide to just go the whole hog and convert the whole `PermissibleBase` into a trait. I'm doubtful if it's worth pushing this to 3.4.

SOF3:
The choice of [implementation by delegation](https://en.wikipedia.org/wiki/Delegation_pattern) is not entirely about backward compatibility. The `PermissibleBase` class contains several class properties that need to be cleaned up after the Player state changes. With implementation by delegation, we could simply do `$this->perm = null` to clean up the resources, but with implement-in-trait, it is not so convenient to clean up unused resources. In the context of PocketMine, a delegation pattern would be convenient for the long term.

The only problem we are encountering is the inconvenient class structure (we end up having 3 classes for the same thing - Permissible, PermissibleBase and PermissibleTrait). But I think this is more like a problem with autoloading. I wonder if we could alter the autoloading mechanism to improve this, since the three structures are more or less the same thing.

Side note: I have created [an issue in PhpStorm's YouTrack](https://youtrack.jetbrains.com/issue/WI-44039) to add delegation pattern as a refactoring feature.

dktapps:
@SOF3 that makes no sense. A trait will know by itself what things it contains and therefore what to clean up when instructed. It even has the clearPermissions() thing right there.

SOF3:
Suppose there are many fields inside the trait, it would be very unmaintainable to have a lot of `unset`/`$this->perm = null`s in the destructor (the clear method). Organizing functionality by delegation allows the referenced resources to be garbage-collected immediately. This reduces the chance of "forgetting to unset" or other problems.

Using a delegation pattern also has other advantages. See: https://stackoverflow.com/a/30424702/3990767

SOF3:
Going to rewrite for a slightly BC-breaking solution to prevent creating 3 files for one feature

PR #2509, opened 2018-10-31
Eliminate the obsolete `pocketmine` subdirectory

dktapps:
# Before anyone says anything stupid
This DOES NOT affect namespaces. As you can see if you look at the `composer.json` file, the `pocketmine` namespace is just being mapped to a different location.

## Introduction
This pull request removes the obsolete `pocketmine` directory, **without affecting namespaces**.

### Rationale
Since this is the only directory under `src` in this repo nowadays (since other modules are using Composer) it doesn't make sense for this directory to continue to exist. Instead, we make use of the PSR-4 autoloading standard to map the `pocketmine` namespace to the `src` directory directly. This has several benefits, including shorter file paths to type, less clicks to find files, less bytes wasted on your disk, and so on.

### API changes
None.

### Behavioural changes
None.

## Backwards compatibility
This may present BC issues for DevTools, so this targets the next major version. Other plugins should not be affected (provided they don't do anything stupid with the source code).

## Follow-up
Allow plugins to use the PSR-4 autoloading standard to remove the need for all these redundant subdirectories.


SOF3:
Note: this may also affect people who just `php src/pocketmine/PocketMine.php`. But this is a devs-only thing

dktapps:
yup, that's also technically an internals thing, so not something I consider worrisome.

SOF3:
Is it necessary to keep `resources` inside `src`?

dktapps:
It would break a bunch of things even more than they are getting broken already (DevTools and the likes).

dktapps:
Relocating `resources` outside of `src` is something that can be done independently of this PR with significantly lower conflict rates, although it will still break DevTools `/makeserver`. I agree it should be done, but not in this pull.

dktapps:
Obsoleted by 5499ac620c61ea335c2151a2750b7f6cc6dc7e75.

PR #2511, opened 2018-11-02
 Changed PermissibleBase to trait to remove boilerplate

SOF3:
## Introduction
Implementing Permissible is based on PermissibleBase, but redirecting the method calls produces a lot of boilerplate code. This PR converts PermissibleBase to an actual trait that implements the Permissible methods.

### API changes

- Added PermissibleTrait
- Removed PermissibleBase

### Behavioural changes

none

## Backwards compatibility

Classes that extend/contain PermissibleBase have to switch to `use PermissibleTrait;`. **Remember to call `initPermissible()` and `clearPermissions()`.**

## Tests

Works when player join and quits with almost no plugins. Not everything has been carefully tested.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -123,7 +123,7 @@ use pocketmine\network\mcpe\protocol\types\ContainerIds;
 use pocketmine\network\mcpe\protocol\types\PlayerPermissions;
 use pocketmine\network\mcpe\protocol\UpdateAttributesPacket;
 use pocketmine\network\mcpe\ProcessLoginTask;
-use pocketmine\permission\PermissibleBase;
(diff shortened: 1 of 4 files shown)


dktapps:
It's been discussed to instead keep this as a subcomponent and remove the Permissible implementation from Player. This would solve the problem as well, along with others related to code complexity.

SOF3:
This change is better combined with the removal of other implementations from Player, hopefully including Vector3, Entity, etc.

dktapps:
3c677bd3ec3584572f84c4728d65179302705ea2 resolves the boilerplate issue, but it also creates conflicts with this pull. Since the majority of its purpose has been obsoleted, I'm closing this.

PR #2523, opened 2018-11-11
Return Item instance for enchant related functions

TwistedAsylumMC:
## Introduction

When coding some things - for example adding items to an array - if you want one of the items to be enchanted, you need to do it outside of the array because ``Item::addEnchantment()`` returns void, but things like ``Item::setLore()`` and ``Item:;setCustomName()`` return an Item instance.

### API changes

``Item::removeEnchantment()`` returns an Item instance instead of void
``Item::removeEnchantments()`` returns an Item instance instead of void
``Item::addEnchantment()`` returns an Item instance instead of void

### Behavioural changes

No

## Backwards compatibility

Don't think it will cause any errors with existing plugins...

## Tests

```php
$item = Item::get(Item::DIAMOND_SWORD)->setCustomName("Sword")->addEnchantment(new EnchantmentInstance(Enchantment::getEnchantment(Enchantment::SHARPNESS), 5));
/** @var Player $player */
$player->getInventory()->addItem($item);
```
image

### DIFF
--- a/src/pocketmine/item/Item.php
+++ b/src/pocketmine/item/Item.php
@@ -305,7 +305,7 @@ class Item implements ItemIds, \JsonSerializable{
 
 		$this->setNamedTagEntry($ench);
 		
-		return $this
+		return $this;
 	}
 
 	/**
--- a/src/pocketmine/item/Item.php
+++ b/src/pocketmine/item/Item.php
@@ -292,7 +292,7 @@ class Item implements ItemIds, \JsonSerializable{
 	public function removeEnchantment(int $id, int $level = -1) : Item{
 		$ench = $this->getNamedTagEntry(self::TAG_ENCH);
 		if(!($ench instanceof ListTag)){
-			return;
+			return $this;
 		}
 
 		/** @var CompoundTag $entry */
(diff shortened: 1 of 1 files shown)


dktapps (changes requested):

src/pocketmine/item/Item.php
```diff
@@ -302,16 +304,25 @@ public function removeEnchantment(int $id, int $level = -1) : void{
...
 
 		$this->setNamedTagEntry($ench);
+		
+		return $this;
 	}
 
-	public function removeEnchantments() : void{
+	/**
+	* @return Item
+	*/
```
dktapps: formatting issue here

src/pocketmine/item/Item.php
```diff
@@ -302,16 +304,25 @@ public function removeEnchantment(int $id, int $level = -1) : void{
...
+	public function removeEnchantments() : Item{
 		$this->removeNamedTagEntry(self::TAG_ENCH);
+
```
dktapps: trailing whitespace here

src/pocketmine/item/Item.php
```diff
@@ -339,6 +350,8 @@ public function addEnchantment(EnchantmentInstance $enchantment) : void{
 		}
 
 		$this->setNamedTagEntry($ench);
+
```
dktapps: trailing whitespace here

dktapps (changes requested):

dktapps (approved):

PR #2526, opened 2018-11-14
Fixed registerEvents() throwing exceptions on candidate event handler…

dktapps:
…s with missing classes

If the class is missing, we can't make any assumptions about the class' inheritance chain. While this error might seem nice at first glance, it doesn't make sense for it to only crop up here, when registering handlers. Instead, it would make more sense to have a debugging autoloader which checks ALL the things irrelevant of whether they are handlers or not.

TL;DR: It's not the responsibility of registerEvents() to barf on this, since it will only barf when it's hitting something it likes the look of. While I think crashing is better than bad behaviour, this validation should happen somewhere else where it is more effective.

### DIFF
--- a/src/pocketmine/event/Listener.php
+++ b/src/pocketmine/event/Listener.php
@@ -43,9 +43,6 @@ use pocketmine\plugin\PluginManager;
  * Functions which meet the criteria can have the following annotations in their doc comments:
  *
  * - `@notHandler`: Marks a function as NOT being an event handler. Only needed if the function meets the above criteria.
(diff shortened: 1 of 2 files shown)


dktapps:
Test case:
```php
<?php

declare(strict_types=1);

namespace dktapps\test;

	}

	public function thisIsntAnEventHandler(SomeClassThatDoesntExist $thing) : void{

	}
}
```

SOF3:
I don't recommend merging this before "this validation should happen somewhere else where it is more effective" actually happens somewhere trivial.

PR #2527, opened 2018-11-14
Exploit PHP copy-on-write behaviour to reduce chunk memory usage by 20-40%

### LINKED ISSUE #2492: Memory usage of unnecessary light arrays
Recently it came to my attention that, since the majority of subchunks have 0 light in one or both layers, we could gain significant memory savings (20-40% per subchunk) by deallocating subchunk light layers which have all-zero light. This would amount to a significant memory saving for many worlds.

dktapps:
This takes advantage of two key behaviours of PHP:

1. Assigning a string does not copy the string.
2. Changing an offset in a string causes the string to be copied.

These two factors combined, along with the fact that blocklight and skylight arrays are usually all-zeros, allow us to produce a significant memory usage reduction of loaded chunks.
A freshly generated PM world with 3,332 chunks loaded drops from 310MB to 200MB memory usage with these changes applied.

This closes #2492 .

### Behavioural changes
No behavioural changes are expected except for a 20-40% drop in memory usage (perhaps up to 60% in edge cases).

## Backwards compatibility
This should have zero effects on BC, so it targets 3.4.0.

## Tests
Tested in-game with a freshly-generated PMgen world.
- Before changes, 3,332 chunks occupy 310 MB of RAM.
- After changes, 3,332 chunks occupy 200 MB of RAM.

### DIFF
--- a/src/pocketmine/level/format/Chunk.php
+++ b/src/pocketmine/level/format/Chunk.php
@@ -840,8 +840,12 @@ class Chunk{
 	 */
 	public function collectGarbage() : void{
 		foreach($this->subChunks as $y => $subChunk){
-			if(!($subChunk instanceof EmptySubChunk) and $subChunk->isEmpty()){ //normal subchunk full of air, remove it and replace it with an empty stub
-				$this->subChunks[$y] = $this->emptySubChunk;
+			if($subChunk instanceof SubChunk){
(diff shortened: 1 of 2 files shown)


dktapps:
Retargeting this to 3.4 since it isn't a bugfix.

PR #2593, opened 2018-12-20
Use 2 spaces in all YAML files for consistency

SOF3:
## Introduction

The number of indents in PocketMine is inconsistent, and it is not specified in .editorconfig, leading to confusion when editing new YAML files.

The common candidates are 1 spaces, 2 spaces, 4 spaces and 8 spaces.

### 1 space
1 space is the current format used in pocketmine.yml, resource_packs.yml and .travis.yml. Nevertheless, in my opinion, 1 space looks ugly and hard to spot, and I bet most people will agree with me here. In addition, commenting out one space has a confusing appearance:

```yaml
uncommented:
 still: uncommented
#This line of comment is not supposed to be removed
# commented from: the line start
 #commented after: the indent
#no indent: but commented
```

The prose comment on line 3 does not use any spaces, but some may find `#` look ugly when followed by text without a space in between.
The one-indent line 4 is unclear whether the space in `# commented...` is part of the `#` or an indent.
The one-indent line 5 is more apparent, but I don't see this convention being used in pocketmine.yml comments. It also appears that PhpStorm <kbd><kbd>Ctrl</kbd>+<kbd>/</kbd></kbd> places the comment symbol `#` at the line start rather than between indents and text start.

Since one space behind `#` is trivial but two spaces behind it isn't, this can be considered a disadvantage of one indent.

In short, one indent is ugly, and should be changed.

### 2 spaces
2 spaces is the default output from `yaml_emit()`. Using this format is consistent with autogenerated YAML files. Since both editors display tabs with size of 4 or 8, using 2 spaces can also help distinguish from tabs.

### 4 spaces
4 spaces is consistent with the tab indent size in the `[*]` entry in .editorconfig, but I don't see it used anywhere within the project.

### 8 spaces
I haven't seen anyone use 8 spaces in YAML files, except vim users who are too lazy to setup their indent size.

As such, 2 spaces would be the best choice, especially given the yaml_emit() convention.

## Changes
All resource .yml files use 2 spaces instead of 1 space. This may change the generated pocketmine.yml.

## Follow-up
May have impacts on updating pocketmine.yml automatically. @dktapps please check.

This change is targetted on master due to resources directory moving.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,3 +4,7 @@ root = yes
 [*]
 indent_size = 4
 indent_style = tab
+
+[*.yml]
+indent_size = 2
+indent_style = space
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,22 +4,22 @@ php:
  - 7.2
 
 before_script:
-# - pecl install channel://pecl.php.net/pthreads-3.1.6
- - echo | pecl install channel://pecl.php.net/yaml-2.0.2
- - pecl install channel://pecl.php.net/crypto-0.3.1
- - git clone https://github.com/pmmp/pthreads.git
- - cd pthreads
- - git checkout c8cfacda84f21032d6014b53e72bf345ac901dac
- - phpize
- - ./configure
- - make
- - make install
- - cd ..
- - echo "extension=pthreads.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
- - composer install
+#  - pecl install channel://pecl.php.net/pthreads-3.1.6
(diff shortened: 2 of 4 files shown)


dktapps:
This will cause additional conflicts any time a change is made to the 3.x branches. I'd prefer this targeted a 3.x branch and be merged upwards so we don't have to deal with conflicts over and over again.

By itself, the change looks OK.

SOF3:
Not gonna rebase towards 3.x since there are unlikely to be changes there.

PR #2596, opened 2018-12-21
Cancelable ClosureTask

Frago9876543210:
## Introduction
Adds the ability to cancel ClosureTask

## Backwards compatibility
Yes

## Tests
```php
$this->getScheduler()->scheduleDelayedTask(new ClosureTask(function(int $currentTick) : void{
	$this->getLogger()->info("Hello / currentTick = $currentTick");
}), 1); //singe call

//for example, now it will allow make a simple timer
$ticks = 5;
$this->getScheduler()->scheduleRepeatingTask(new ClosureTask(function(int $_) use (&$ticks) : bool{
	if($v = $ticks-- !== 0){
		$this->getLogger()->info("*ticking*");
	}else{
		$this->getLogger()->warning("I'm cancelled!");
	}
	return $v; //if false, it will be cancelled
}, true), 20);
```

### DIFF
--- a/src/pocketmine/scheduler/ClosureTask.php
+++ b/src/pocketmine/scheduler/ClosureTask.php
@@ -48,7 +48,7 @@ class ClosureTask extends Task{
 	 * @param \Closure|null $cancel Will be called if the first argument will return false
 	 */
 	public function __construct(\Closure $closure, ?\Closure $cancel = null){
-		Utils::validateCallableSignature(($nullable = is_null($cancel)) ? function(int $currentTick) : void{} : function(int $currentTick) : bool{ return true; }, $closure);
+		Utils::validateCallableSignature(($nullable = $cancel === null) ? function(int $currentTick) : void{} : function(int $currentTick) : bool{ return true; }, $closure);
 		if(!$nullable){
 			Utils::validateCallableSignature(function() : void{}, $cancel);
 		}
@@ -63,7 +63,10 @@ class ClosureTask extends Task{
 	public function onRun(int $currentTick){
 		if(($this->closure)($currentTick) === false && $this->cancel !== null){
 			$this->getHandler()->cancel();
-			($this->cancel)();
 		}
+	}
+
+	public function onCancel(){
+		($this->cancel)();
 	}
 }
--- a/src/pocketmine/scheduler/ClosureTask.php
+++ b/src/pocketmine/scheduler/ClosureTask.php
@@ -40,13 +40,20 @@ class ClosureTask extends Task{
 
 	/** @var \Closure */
 	private $closure;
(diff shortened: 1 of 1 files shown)


src/pocketmine/scheduler/ClosureTask.php
```diff
@@ -40,20 +40,30 @@ class ClosureTask extends Task{
...
 
 	public function getName() : string{
 		return Utils::getNiceClosureName($this->closure);
 	}
 
 	public function onRun(int $currentTick){
-		($this->closure)($currentTick);
+		if(($this->closure)($currentTick) === false && $this->cancel !== null){
+			$this->getHandler()->cancel();
+			($this->cancel)();
```
dktapps: This can be called from inside `onCancel()` in line with standard tasks.

Frago9876543210: Okey, I'll move it tommorow. That about is_null?

dktapps: Prefer comparison to `null` instead of that.

src/pocketmine/scheduler/ClosureTask.php
```diff
@@ -40,20 +40,33 @@ class ClosureTask extends Task{
...
 	 */
-	public function __construct(\Closure $closure){
-		Utils::validateCallableSignature(function(int $currentTick) : void{}, $closure);
+	public function __construct(\Closure $closure, ?\Closure $cancel = null){
+		Utils::validateCallableSignature(($nullable = $cancel === null) ? function(int $currentTick) : void{} : function(int $currentTick) : bool{ return true; }, $closure);
+		if(!$nullable){
+			Utils::validateCallableSignature(function() : void{}, $cancel);
+		}
 		$this->closure = $closure;
+		$this->cancel = $cancel;
```
dktapps: I've been thinking about this callback and I don't think it makes a whole lot of sense. A self-cancelling task is capable of running actions directly when it does so.

dktapps (commented):
I wonder if this logic might be better suited to a separate class instead of butchering `ClosureTask`. I'm not comfortable with the variable signature requirement for the `run` closure.

src/pocketmine/scheduler/ClosureTask.php
Frago9876543210: Do you mean that this should be transferred to another class to not to create confusion with the returned type?

dktapps: I mean that this `cancel` closure doesn't make sense.

dktapps:
Obsoleted by 808259dccd650d41fad9c9adfdf5d936a0a7e292.

PR #2613, opened 2018-12-27
Allow plugins to disable background generation

Frago9876543210:
## Introduction
See title

### API changes
```diff
- Server::generateLevel(string $name, int $seed = null, $generator = null, array $options = [])
+ Server::generateLevel(string $name, int $seed = null, $generator = null, array $options = [], bool $backgroundGeneration = true)
```

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -1052,7 +1052,7 @@ class Server{
 	 *
 	 * @return bool
 	 */
-	public function generateLevel(string $name, int $seed = null, $generator = null, array $options = [], $backgroundGeneration = true) : bool{
+	public function generateLevel(string $name, int $seed = null, $generator = null, array $options = [], bool $backgroundGeneration = true) : bool{
 		if(trim($name) === "" or $this->isLevelGenerated($name)){
 			return false;
 		}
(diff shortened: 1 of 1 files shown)


src/pocketmine/Server.php
```diff
@@ -1048,10 +1048,11 @@ public function loadLevel(string $name) : bool{
 	 * @param int|null    $seed
 	 * @param string|null $generator Class name that extends pocketmine\level\generator\Generator
 	 * @param array       $options
+	 * @param bool        $backgroundGeneration
 	 *
 	 * @return bool
 	 */
-	public function generateLevel(string $name, int $seed = null, $generator = null, array $options = []) : bool{
+	public function generateLevel(string $name, int $seed = null, $generator = null, array $options = [], $backgroundGeneration = true) : bool{
```
dktapps:
```suggestion
	public function generateLevel(string $name, int $seed = null, $generator = null, array $options = [], bool $backgroundGeneration = true) : bool{
```

dktapps (changes requested):
Looks fine after nits are fixed.

dktapps (approved):

dktapps:
This might have unintended side effects on world loading. If no spawn chunks are generated, the initial region on Region-based worlds will not be created, which will prevent the format selector detecting the correct world format. Plugins would have to set chunks into the world before unloading it, or risk the empty world getting overwritten with a new one.

Frago9876543210:
Yes, it has side effects, but the responsibility lies with the plugins. This can simplify the format conversion, which is required later.

PR #2646, opened 2019-01-06
Player: Experimental new movement processing system

### LINKED ISSUE #1215: Movement buffering system causes various bugs on net lag spikes
### Issue description
When a player sprints and jumps at the same time continuously, the anti-cheat gets triggered.

Expected result: What were you expecting to happen? Sprint jump with no problems
Actual result: What actually happened? Lots of players being pulled to the ground after sprint jumping continuously

1. Join the server
2. Start sprint jumping continuously

### OS and versions

* PocketMine-MP: 4bd65b4
* PHP: 7.0.3
* Server OS: Linux
* Game version: PE/Win10

### Plugins
- Test on a clean server without plugins: is the issue reproducible without any plugins loaded? Yes, it is.

* N/A

dktapps:
## Introduction
The existing movement system is problematic because it only processes the last movement received on a tick. This can be a problem on laggy connections. Even when not on a laggy connection, this can cause problems like anti-cheat triggering falsely (anti fly not seeing players touch ground), the infamous "moved too fast", and more.

This PR changes movement processing to use a pending movement queue instead of only processing the last movement. This allows much more effective recovery from network lag spikes without rubberbanding. With this in place, appearances of `moved too fast` will be significantly reduced.

This has been carefully designed to avoid exposing the server to attack vectors, but it is not complete on the security front yet.

### Relevant issues
- Fixes #1215 - touchdown should now work correctly (assuming the client sends appropriate movements). `moved too fast` is also significantly reduced.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -161,11 +161,14 @@ use pocketmine\timings\Timings;
 use pocketmine\utils\TextFormat;
 use pocketmine\utils\UUID;
 use function abs;
+use function array_fill;
 use function array_merge;
+use function array_sum;
 use function assert;
 use function base64_decode;
 use function ceil;
 use function count;
+use function end;
 use function explode;
 use function floor;
 use function fmod;
@@ -182,6 +185,7 @@ use function lcg_value;
 use function max;
 use function microtime;
 use function min;
+use function mt_rand;
 use function preg_match;
 use function round;
 use function spl_object_hash;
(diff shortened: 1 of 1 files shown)


dktapps:
This can be enhanced by ignoring movements which only involve rotation compared to the previous movement, since rotation has no effect on movement distance.

Wumpotamus:
I've been testing this and the only bug I've noticed was when falling from a very large height (32,000 to less than 20,000) you lose hunger, which isn't like that in vanilla.

dktapps:
#2708

dktapps:
This pull is overcomplicated. This change could be better implemented by removing buffering entirely and implementing a simple network-layer rate-limit, which would also cooperate better with things like block placement.

I'm closing this due to being stale and having conflicts.

PR #2668, opened 2019-01-16
Added Dockerfile

### LINKED ISSUE #928: Create Docker container for easier distribution
### Issue description
PMMP Team: I've created a docker image for PMMP that may be useful to include in the broader distribution. If you think its worth to have as part of the main distribution I can fork and submit a PR.

https://github.com/crscheid/pocketmine-mp

None

### OS and versions
N/A

N/A

SOF3:
## Introduction
Closes #928

## Discussion needed
The configuration currently only builds the following:
```
branch master -> pmmp/pocketmine-mp:dev
tag .* -> pmmp/pocketmine-mp:{sourceref}
tag .* -> pmmp/pocketmine-mp:latest
```

Should other branches be built?

## Follow-up
Advertise the docker repository `pmmp/pocketmine-mp` in README, etc.

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -1,0 +1,44 @@
+FROM ubuntu:bionic
+MAINTAINER PMMP Team <team@pmmp.io>
+
+EXPOSE 19132/tcp
+EXPOSE 19132/udp
+
+RUN apt-get update && apt-get --no-install-recommends -y install \
+	sudo \
+	ca-certificates \
+	jq \
+	curl \
+	unzip
+
+RUN groupadd pocketmine && useradd -r -d /pocketmine -p "" -m -g pocketmine -g sudo pocketmine
+
+WORKDIR /pocketmine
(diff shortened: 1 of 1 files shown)


dktapps:
I wonder if we should make use of `apt` and system libraries here since we have them guaranteed available in this environment.

SOF3:
Would using `apt` introduce inconvenience since we might have custom extensions?

SOF3:
To minimize the maintenance cost, using the php;-build-scripts artifact is more preferrable, unless we no longer ship non-Docker builds.

dktapps:
The maintenance cost of php-build-scripts is very high. This is why I suggested using system libraries inside the image.

SOF3:
This docker image is not going to remove the maintenance cost of php-build-scripts as we are not deprecating the latter. Right now it only relies on the public API of the Jenkins server that is unlikely to change. I don't see how using system libraries would reduce maintenance cost.

PR #2697, opened 2019-01-27
make addMotion more comfortable

rdstonech:
## Introduction
with current addMotion if we need to add only X coord need to write Entity->addMotion(1, 0, 0);
This pull request make it more easy to use: Entity->addMotion(1);

### API changes
Entity->addMotion(float, float, float) -> Entity->addMotion(float, float = 0, float = 0)

### Behavioural changes
No.

## Backwards compatibility
Full backward compatibility

### DIFF
--- a/src/pocketmine/entity/Entity.php
+++ b/src/pocketmine/entity/Entity.php
@@ -1687,7 +1687,7 @@ abstract class Entity extends Location implements Metadatable, EntityIds{
 	 * @param float $y
 	 * @param float $z
 	 */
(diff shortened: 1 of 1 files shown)


dktapps:
This is unnecessary and will most likely result in bugs when used since it becomes possible to specify only 1 argument by mistake (anyone remember the old DestroyBlockParticle bug where it didn't spawn at the right position?).
It's also potentially misleading because this could be misinterpreted as allowing a scalar value, when in fact the first parameter only applies to 1 axis.

In addition, a plugin is not supposed to use this.

PR #2736, opened 2019-02-04
Throw TerrainNotLoadedException when accessing unloaded terrain, fixes lots of consistency bugs in chunk handling

dktapps:
## Introduction
This was originally integrated directly into master and then reversed due to stability issues. This is still being worked on and problems still need to be addressed.

The main issue is that chunks are synchronously loaded on the fly whenever `Level->getChunk()` is called (unasked) which causes a whole lot of hidden problems to manifest, including unexplained memory leaks, DoS attack vectors and more.

Second, there is an inconsistency around how unloaded chunks are handled in general. Sometimes they are synchronously loaded, and in other cases they are just skipped and treated like a chunk full of air. This causes a lot of problems with performance, as well as problems with incorrect cached items when caching results of accessing the world.

This PR aims to solve these problems by making unloaded chunk handling 100% consistent.

I opted not to have chunks always autoloaded, since this is not always possible anyway. In addition, it is not possible to synchronously request the generation of a chunk, which causes the same problems under the current architecture as assuming air for an unloaded chunk.

These changes are necessary to facilitate asynchronous I/O, which has been a desired feature for over 2 years.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -1497,7 +1497,7 @@ class Player extends Human implements CommandSender, ChunkLoader, ChunkListener,
 			$this->server->getLogger()->warning($this->getName() . " moved too fast, reverting movement");
 			$this->server->getLogger()->debug("Old position: " . $this->asVector3() . ", new position: " . $this->newPosition);
 			$revert = true;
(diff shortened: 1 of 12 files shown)


dktapps:
(for posterity) There are a series of problems which manifest with this PR which required that it be removed from the master branch. These include:

- Crashes and incorrectly calculated lighting during chunk loading (this could be fixed by not light-populating chunks whose adjacent chunks are missing)
- Crashes when an entity moves or teleports into a chunk where the immediately adjacent chunks that it overlaps are not loaded
- maybe more I don't remember

PR #2744, opened 2019-02-08
Improvements for Splash Potion

Vrysed:
Improved Motion

## Introduction
This pull request is necessary because these values are closer to vanillas values 

### Relevant issues
Force is too far


dktapps:
This force value matches that found in vanilla code. If the distance is different, that indicates there is something wrong elsewhere and this will simply remove the symptom rather than fixing the cause.

PR #2747, opened 2019-02-10
Implement thread safe composer class loading

JackNoordhuis:
## Introduction
Makes use of a thread-safe/thread friendly version of the composer autoloader. This removes the need to use the `BaseClassLoader` which is currently only used for plugin class loading and is a step towards having plugins loaded from a composer.json instead of the traditional plugin.yml manifest.

### API changes
Exposes composer class loader methods from the `ServerClassLoader` which extends the `ThreadedClassLoader` and implements `\ClassLoader` to maintain backwards compatibility.

### Behavioural changes
The server no longer maintains two class loaders. The server performance shouldn't have been impacted, although pthreads inefficencies with `\Threaded` objects may have an effect.

## Backwards compatibility
Removes the protected `$composerAutoloaderPath` property from the `Thread` and `Worker` classes. This could possibly be considered an internal change given the lack of plugins using threading at this level let alone using this property.

## Tests
I ran the travis tests locally before committing and the [threaded-class-loader repository](https://github.com/JackNoordhuis/threaded-class-loader) has it's own tests available [here](https://github.com/JackNoordhuis/threaded-class-loader/blob/master/tests/Autoload/ClassLoaderTest.php).

### DIFF
--- a/src/pocketmine/PocketMine.php
+++ b/src/pocketmine/PocketMine.php
@@ -29,6 +29,7 @@ namespace {
 namespace pocketmine {
 
 	use pocketmine\utils\MainLogger;
+	use pocketmine\utils\ServerClassLoader;
 	use pocketmine\utils\ServerKiller;
 	use pocketmine\utils\Terminal;
 	use pocketmine\utils\Timezone;
@@ -156,7 +157,7 @@ namespace pocketmine {
 	define('pocketmine\COMPOSER_AUTOLOADER_PATH', $bootstrap);
 
 	if(\pocketmine\COMPOSER_AUTOLOADER_PATH !== false and is_file(\pocketmine\COMPOSER_AUTOLOADER_PATH)){
-		require_once(\pocketmine\COMPOSER_AUTOLOADER_PATH);
+		$autoloader = require_once(\pocketmine\COMPOSER_AUTOLOADER_PATH);
 	}else{
 		critical_error("Composer autoloader not found at " . $bootstrap);
 		critical_error("Please install/update Composer dependencies or use provided builds.");
@@ -168,8 +169,7 @@ namespace pocketmine {
 	/*
 	 * We now use the Composer autoloader, but this autoloader is still for loading plugins.
 	 */
-	$autoloader = new \BaseClassLoader();
-	$autoloader->register(false);
+	$autoloader = ServerClassLoader::fromComposerLoader($autoloader);
 
 	set_time_limit(0); //Who set it to 30 seconds?!?!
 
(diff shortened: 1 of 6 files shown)


src/pocketmine/PocketMine.php
```diff
@@ -168,8 +169,7 @@ function check_platform_dependencies(){
 	/*
 	 * We now use the Composer autoloader, but this autoloader is still for loading plugins.
```
JackNoordhuis: Should probably modify or remove this comment

src/pocketmine/utils/ServerClassLoader.php
```diff
@@ -0,0 +1,66 @@
...
+	 * Creates a new threaded class loader from a default composer class loader.
+	 *
+	 * @param \Composer\Autoload\ClassLoader $loader The composer class loader.
+	 * @param array $includeFiles Array of files to be included by the loader.
+	 * @param bool $register      If the new autoloader should be registered.
+	 * @param bool $unregister    If the composer autoloader should be unregistered.
+	 *
+	 * @return \pocketmine\utils\ServerClassLoader
+	 */
+	public static function fromComposerLoader(ComposerClassLoader $loader, array $includeFiles = [], bool $register = true, bool $unregister = true){
```
dktapps: It appears this function is duplicated from your own project, or am I missing something?

dktapps:
Unless I understood incorrectly, your library is a near perfect replica of the traditional composer loader, but for pthreads. What I don't understand is why you have a nest of reflection hacks to extract the information from the previous loader.

The way I previously approached this was to have a Threaded loader with the same PSR-0/4 capabilities which was registered _after_ the regular composer loader. This worked in a fashion like:
```php
$loader = new ThreadedComposerLoader("vendor/autoload.php");
```
and then the loader would register itself (the dynamic version) after the static composer one. Then, paths would be added to the threaded loader in the same fashion that they are today. Nearly the same design as today, but the threaded loader simply `require`s the composer loader path when registered, then registers itself for dynamic loading. The only reason this was never implemented is because I didn't have the time to replicate the classic Composer loader and make it thread safe.

JackNoordhuis:
I added the methods to my threaded loader implementation to make using multiple composer autoloaders easier. The methods aren't currently covered by tests and imo shouldn't really be in the loader class itself, so I'll do something about that when I get the chance.

As for your method with passing in the composer autoloaders path directly, this behaviour can already be achieved with `ThreadedClassLoader::addFile()` (which is copied over from the composer loader) rather than merging with the composer loader directly. However, my thought process was to just have all the classes tracked and registered by one loader.

PR #2748, opened 2019-02-10
 Split up plugin loader and plugin manifest loader

JackNoordhuis:
## Introduction
This pull request introduces separation of a plugins storage format and it's manifest. The simplest way of explaining the rationale behind this change is that now .phar plugins won't always have to have a plugin.yml for the server to load the plugin, every loader (usually a different format .phar, .zip, etc) can have many registered manifests (plugin.yml, composer.json). This makes the code easier to test and much easier for plugin developers to implement their own loaders and manifest formats.

### API changes
* Core plugin loaders and related classes are now under the `pocketmine\plugin\loader` namespace.
* Plugins can now register new plugin manifest classes to core plugin loaders using `AbstractPluginLoader->addManifestLoader()`.
* `PluginManager->registerInterface()` has been renamed to `PluginManager->registerLoader()` for consitency.
* `PluginLoader->getDescription()` has been replaced with `PluginLoader->getManifest()` and the description.

### Behavioural changes
An extra step has been added to accommodate registering new manifest formats. First the server checks if a plugin loader exists for the format, then the loader searches for a suitable manifest to load the plugin,

## Backwards compatibility
All plugin loaders provided by plugins will need updating, this process should be pretty straight forward and will reduce the amount of code needed to create new loaders and manifest formats.

## Follow-up
~~I will open a pull request to update devtools to accommodate these changes.~~
DevTools pull request: https://github.com/pmmp/PocketMine-DevTools/pull/40

## Tests
Added test cases for detecting phar and script plugins with their respective loaders, along with tests for finding and parsing the yaml and php doc manifests.


src/pocketmine/plugin/manifest/PluginManifest.php
```diff
@@ -0,0 +1,66 @@
...
+	 * @return PluginDescription|null
+	 */
+	public function getPluginDescription() : ?PluginDescription;
+
+	/**
+	 * Register the plugin into the runtime.
+	 *
+	 * @param \ClassLoader $loader
+	 */
+	public function registerPlugin(\ClassLoader $loader) : void;
```
dktapps: It's also unclear what business a manifest has loading a plugin. This is almost certainly not the job of the manifest to handle.

src/pocketmine/plugin/manifest/PluginManifest.php
```diff
@@ -0,0 +1,66 @@
...
+*/
+
+declare(strict_types=1);
+
+namespace pocketmine\plugin\manifest;
+
+use pocketmine\plugin\PluginDescription;
+use pocketmine\plugin\loader\PluginLoader;
+
+interface PluginManifest{
```
dktapps: This interface has no documentation.

While this is not a terrible thing, I am unable to come up with a succint description of what this interface does. Documenting it would be good to help confidence in the work.

src/pocketmine/Server.php
```diff
@@ -1606,8 +1606,8 @@ public function reload(){
 			$this->getNetwork()->blockAddress($entry->getName(), -1);
 		}
 
-		$this->pluginManager->registerInterface(new PharPluginLoader($this->autoloader));
-		$this->pluginManager->registerInterface(new ScriptPluginLoader());
+		$this->pluginManager->registerLoader(new PharPluginLoader());
+		$this->pluginManager->registerLoader(new ScriptPluginLoader());
```
dktapps: This change seems to be unrelated to the goal of the PR.

src/pocketmine/plugin/loader/PharPluginLoader.php
```diff
@@ -0,0 +1,52 @@
...
+	 * @inheritdoc
+	 */
+	public function getAccessProtocol() : string{
+		return "phar://";
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function getFileExtension(): string{
```
dktapps: formatting issue here

dktapps (changes requested):
I've given the code a first pass, will re-review it later.

PR #2750, opened 2019-02-10
Calculate server uptime correctly

### LINKED ISSUE #2713: Server startup time includes time took on setup wizard too
### Issue description
Currently, PocketMine-MP calculates startup time from when it is run. This is a problem for first-run since it doesn't care how much time did user spent from setup wizard. In other words, when a user spent 10 minutes from the wizard, the server thinks it took 10 minutes and some startup time to run the server.
While it is unclear whether it is a bug or not, I believe it should be considered as a bug since technically the server isn't even running because it is just a setup wizard.

1. Setup a new server
2. Wait for some time before finishing wizard to make it easier to see the issue
3. Finish wizard
4. See that startup time includes time took on wizard too

### OS and versions

* PocketMine-MP: 3.5.7
* PHP: 7.2.14
* Server OS: Windows 10
* Game version: PE/Win10 (delete as appropriate)

### Plugins

JackNoordhuis:
Fixes #2713.

### DIFF
--- a/src/pocketmine/PocketMine.php
+++ b/src/pocketmine/PocketMine.php
@@ -244,11 +244,15 @@ namespace pocketmine {
 	$exitCode = 0;
 	do{
 		if(!file_exists(\pocketmine\DATA . "server.properties") and !isset($opts["no-wizard"])){
+			$start = microtime(true);
 			$installer = new SetupWizard();
 			if(!$installer->run()){
 				$exitCode = -1;
 				break;
 			}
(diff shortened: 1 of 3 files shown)


dktapps:
It seems to me that there's a simpler fix for this without writing all this additional code.

JackNoordhuis:
It could do without the getters but I thought it'd be a good idea to add them to keep it consistent and expose the constants through the methods. We could drop the setup duration constant too but my thought process was plugins may want to use it for some reason.

Issue #2692, opened 2019-01-25
Make `gamemode` key in server.properties accept gamemode names instead of magic numbers


dktapps:
Currently users are forced to look up magic numbers to change the `gamemode` property. This is sub optimal. They should be able to use string names instead.

PR #2752, opened 2019-02-10
Allow gamemode string in server.properties

JackNoordhuis:
Addresses #2692.

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -430,7 +430,7 @@ class Server{
 	 * @return int
 	 */
 	public function getGamemode() : int{
-		return $this->getConfigInt("gamemode", 0) & 0b11;
+		return GameMode::fromString($this->getConfigString("gamemode", "s"));
 	}
 
 	/**
@@ -466,7 +466,7 @@ class Server{
 	 * @return int
 	 */
 	public function getDefaultGamemode() : int{
-		return $this->getConfigInt("gamemode", 0) & 0b11;
+		return GameMode::fromString($this->getConfigString("gamemode", "s"));
 	}
 
 	/**


dktapps:
There is no error checking in this PR, which will cause the server to crash if an invalid gamemode is found in server.properties. This is why I did not apply this change myself, as trivial as it looks.

JackNoordhuis:
I wasn't sure what behaviour to implement for catching the error. I don't think the server should start given a bad configuration but a warning and defaulting back to survival could be an alternative?

dktapps:
The configuration crashing issue is a more general problem. The point is that I explicitly avoided applying this change myself because of the additional problems it would cause without a proper solution for config handling.

src/pocketmine/Server.php
```diff
@@ -466,7 +466,7 @@ public function isHardcore() : bool{
 	 * @return int
 	 */
 	public function getDefaultGamemode() : int{
-		return $this->getConfigInt("gamemode", 0) & 0b11;
+		return GameMode::fromString($this->getConfigString("gamemode", "s"));
```
dktapps: I would be happier if `(string) GameMode::SURVIVAL` was used here. It's too easy to miss this if refactors happen.

dktapps (changes requested):

src/pocketmine/Server.php
```diff
@@ -430,7 +430,7 @@ public function getServerUniqueId(){
 	 * @return int
 	 */
 	public function getGamemode() : int{
-		return $this->getConfigInt("gamemode", 0) & 0b11;
+		return GameMode::fromString($this->getConfigString("gamemode", "s"));
```
dktapps: and here too (ofc)

dktapps:

Issue #2667, opened 2019-01-16
Missing API method to set despawn time of items


artulloss:
## Description

It should be possible to easily set the despawn time of Items, why would you want to have to run a separate task to do this when there is already something that will despawn the items? I could see setting this lower as a better alternative to "clear lag" plugins.

## Justification

You can currently set the pickup delay but not the despawn delay via Event, this feels like bad design and should probably be implemented.

## Alternative methods

Extend the Item class then register it with

Entity::registerEntity();

and tweak the class to have a lower despawn time.

Edit: Complied  with the new Enhancement template and gave a way to work around this current limitation.

PR #2753, opened 2019-02-10
Add getter and setter for item entity age

JackNoordhuis:
## Introduction
Somewhat addresses #2667.

### API changes
* `ItemEntity->getAge()`
* `ItemEntity->setAge(int)`

### DIFF
--- a/src/pocketmine/entity/object/ItemEntity.php
+++ b/src/pocketmine/entity/object/ItemEntity.php
@@ -58,6 +58,9 @@ class ItemEntity extends Entity{
 	/** @var int */
 	protected $age = 0;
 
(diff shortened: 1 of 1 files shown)


jasonw4331 (approved):

dktapps:
The issue I have with this is that this "age" is more accurately counting up to a despawn limit, which is hardcoded, or in this PR, const. It seems to me it would be more useful to be able to change the despawn _timeout_ instead of the entity's despawn _ticker_. As such, it does not really resolve #2667 .

Gianluxx:
We could add a Item's MAX_AGE in pocketmine.yml config 
+ this pr for custom items despawn

PR #2756, opened 2019-02-11
Fix a crash and unnecessary packet sending

EmreTr1:
## Introduction

Fixes a crash and unnecessary packet sending

### Relevant issues

- Fixes a crash

## Tests

Tested and working.

### Proof
When resource packs are done. networksession starts to construct entity:
ekran alintisi56565

When `PlayerLoginEvent` was cancelled, server closes player:

ekran alintisi4343
And, Player::close function destroys player and now player's level is null.
ekran alintisi96794534
but this function doesnt stop and it continues to setup `PreSpawnSessionHandler`
ekran alintisi564564564564
and here, this tries to get time of level but level is null and crashes.
ekran alintisi5645646456456

### DIFF
--- a/src/pocketmine/network/mcpe/NetworkSession.php
+++ b/src/pocketmine/network/mcpe/NetworkSession.php
@@ -411,8 +411,10 @@ class NetworkSession{
 
 	public function onResourcePacksDone() : void{
 		$this->player->_actuallyConstruct();
-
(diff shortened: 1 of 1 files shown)


SalmonDE:
Firstly:
**Explain existing problems or why this pull request is necessary**

Secondly:
_Details should be provided of tests done. Simply saying "tested" or equivalent is not acceptable._

PR #2759, opened 2019-02-13
Allow to create custom entities

EmreTr1:
## Introduction

I created a resource pack and i added custom entities to it and i want to spawn this entities in server but i cant, so i made this commit. Also this may be wrong or hack. we should use entity namespaces(ex: minecraft:zombie) instead of magic numbers with 1.8-1.9, i think

### Relevant issues

We cant spawn our custom entities with resource packs

## Tests

```php
$pk = new AddEntityPacket;
$pk->type = "custom:car";
$pk->position = $pos;
$pk->entityRuntimeId = EntityFactory::nextRuntimeId();

$player->sendDataPacket($pk);
```


dktapps:
You clearly have no idea what you're doing. That example code will most likely crash the client or do nothing whatsoever.

It is planned to transition away from legacy IDs for 4.0 release. This hack is present for the existing production releases because of backwards compatibility issues. It doesn't make sense to preserve it like this on bleeding edge.

PR #2768, opened 2019-02-17
Fixed PaintingItem exception after interacting

eddir:
## Introduction

Fixes a crash with PaintingItem while attempting to change their.

```
Error: Return value of pocketmine\item\PaintingItem::onActivate() must be an instance of pocketmine\item\ItemUseResult, boolean returned
File: src/pocketmine/item/PaintingItem
Line: 102
Type: TypeError
```

### Relevant issues

 * Fixes a crash

## Tests

 Now the PentingItem is working correctly when the player interacts with it.


dktapps:
Ouch, I didn't see this until after I pushed my own fixes. Thank you anyway.

PR #2797, opened 2019-03-05
Update PlayerRespawnEvent.php

Thunder33345:
removed obsolete comment

## Introduction
because the comment is obsolete, i think we should remove it

### Relevant issues

?(unsure)

### API changes

none, other the comments should not affect the code itself
### Behavioural changes

none, other the comments which also should not affect how it behaves or performance
## Backwards compatibility

none, other the comments which should not affect backward compatibility
## Follow-up
Nothing
## Tests

comment changes should not necessary test or affect code behavior right?

### DIFF
--- a/src/pocketmine/event/player/PlayerRespawnEvent.php
+++ b/src/pocketmine/event/player/PlayerRespawnEvent.php
@@ -27,7 +27,7 @@ use pocketmine\level\Position;
 use pocketmine\Player;
 
 /**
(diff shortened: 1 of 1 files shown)


dktapps (approved):

dktapps:
I'd merge this already, but it seems GitHub isn't working...

dktapps:
I'd merge this already, but it seems GitHub isn't working...

PR #2812, opened 2019-03-12
Implement device info methods

lukeeey:
## Introduction
This PR adds the ability to view the players device model and platform type in PlayerPreLoginEvent. 
Due to some issues this is not yet available in the regular Player class.

This is slightly different to #2216 but maybe that can be closed if this is merged.

## Changes
* Added a PlatformTypes interface with all current platform ids.
* Added `PlayerInfo->getDeviceModel()` and `getPlatformType()`  

### Behavioural changes
None

### Follow up
Unfuck the player class and have access to these methods from there.

### DIFF
--- a/src/pocketmine/PlayerInfo.php
+++ b/src/pocketmine/PlayerInfo.php
@@ -43,14 +43,20 @@ class PlayerInfo{
 	private $xuid;
 	/** @var int */
 	private $clientId;
+	/** @var string */
+	private $deviceModel;
+	/** @var int */
+	private $platformType;
 
-	public function __construct(string $username, UUID $uuid, Skin $skin, string $locale, string $xuid, int $clientId){
+	public function __construct(string $username, UUID $uuid, Skin $skin, string $locale, string $xuid, int $clientId, string $deviceModel, int $platformType){
 		$this->username = $username;
 		$this->uuid = $uuid;
 		$this->skin = $skin;
 		$this->locale = $locale;
 		$this->xuid = $xuid;
 		$this->clientId = $clientId;
+		$this->deviceModel = $deviceModel;
+		$this->platformType = $platformType;
 	}
 
 	/**
@@ -93,5 +99,19 @@ class PlayerInfo{
 	 */
 	public function getClientId() : int{
 		return $this->clientId;
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getDeviceModel(): string {
+		return $this->deviceModel;
+	}
+
+	/**
+	 * @return int
+	 */
+	public function getPlatformType(): int {
+		return $this->platformType;
 	}
 }
--- a/src/pocketmine/PlayerInfo.php
+++ b/src/pocketmine/PlayerInfo.php
@@ -104,14 +104,14 @@ class PlayerInfo{
 	/**
 	 * @return string
 	 */
-	public function getDeviceModel(): string {
+	public function getDeviceModel() : string {
 		return $this->deviceModel;
 	}
 
 	/**
 	 * @return int
 	 */
-	public function getPlatformType(): int {
+	public function getPlatformType() : int {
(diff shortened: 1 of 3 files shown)


src/pocketmine/PlayerInfo.php
```diff
@@ -94,4 +100,18 @@ public function getXuid() : string{
 	public function getClientId() : int{
 		return $this->clientId;
 	}
+
+	/**
+	 * @return string
+	 */
+	public function getDeviceModel(): string {
```
dktapps: formatting issue here, should be `() : string{`

src/pocketmine/PlayerInfo.php
```diff
@@ -94,4 +100,18 @@ public function getXuid() : string{
...
+		return $this->deviceModel;
+	}
+
+	/**
+	 * @return int
+	 */
+	public function getPlatformType(): int {
```
dktapps: similar here

src/pocketmine/network/mcpe/protocol/LoginPacket.php
```diff
@@ -196,7 +199,9 @@ protected function decodeConnectionRequest() : void{
 			),
 			$this->clientData[self::I_LANGUAGE_CODE],
 			$this->extraData[self::I_XUID],
-			$this->clientData[self::I_CLIENT_RANDOM_ID]
+			$this->clientData[self::I_CLIENT_RANDOM_ID],
+			$this->clientData[self::I_DEVICE_MODEL],
+			$this->clientData[self::I_DEVICE_OS]
```
dktapps: These fields are not checked for presence, and it's possible for them to be omitted (for example by a custom client). If you want them to be mandatory, their presence should be checked in the `Validator` above.

src/pocketmine/network/mcpe/protocol/types/PlatformTypes.php
```diff
@@ -0,0 +1,36 @@
...
+
+	public const ANDROID = 1;
+	public const IOS = 2;
+	public const MAC = 3;
+	public const FIRE = 4;
+	public const GEARVR = 5;
+	public const HOLOLENS = 6;
+	public const WINDOWS = 7;
+
+}
```
dktapps: This table is incomplete - you can find a more complete list in Steadfast2.

src/pocketmine/network/mcpe/protocol/LoginPacket.php
lukeeey: I was thinking about that. I dont see think it should be mandatory and could maybe set it to null or "Unknown" if it doesnt exist, which would you suggest?

dktapps: https://github.com/pmmp/PocketMine-MP/pull/2812/files#diff-5527e13f8305c41ad658af11214531e0R186

lukeeey: Is that supposed to link to a specific line because it doesnt

Wumpotamus:
Don't forget Xbox :laughing:, the ID is 13

lukeeey:
Is that confirmed? Its not in Steadfast2.

Wumpotamus:
@lukeeey That's the DeviceOS that I received when I joined on a test server.

MisteFr:
@Wumpotamus Oh boiiiii, good to see you again

src/pocketmine/network/mcpe/protocol/LoginPacket.php
dktapps: https://github.com/pmmp/PocketMine-MP/blob/0dec932df0e3cc1b2d97a43d0d7db330ecbf47c1/src/pocketmine/network/mcpe/protocol/LoginPacket.php#L186

PR #2823, opened 2019-03-21
Remove /reload

dktapps:
For many years, this has been the cause of many users (particularly plugin devs) confusion. The reality is that /reload has little to no practical value. It does not reload plugin source code (contrary to popular belief), and produces inconsistent behaviour as far as data reloading goes (since it depends on plugins to handle reloading correctly, which many don't).

## Tests
There isn't much to test. Executing `reload` now produces an `unknown command` message.


SOF3:
This might have further impacts. For example, could I further assume that plugins won't be disabled without server stop? Could I assume that a plugin is only run single-instance in each server runtime? /reload has been the favorite excuse for such reasoning.

dktapps:
Not necessarily. There are other methods by which plugins can be disabled.

SOF3 (approved):
The pull request itself is OK, but I would go for the more aggressive change of removing post-startup plugin enabling completely.

PR #2827, opened 2019-03-23
Allow script plugins without the manifest, resolves #2815

### LINKED ISSUE #2815: Allow script plugins without metadata
Script plugins are designed to allow developers to test something with a plugin context very quickly. Typing the metadata is a waste of time and a big distraction.

It is proposed that PocketMine automatically detects the namespace and main class name of a script plugin (by scanning `extends PluginBase` from the lexer function `token_get_all()`) and generate a name automatically according to the script name (e.g. `fooBar.php` -> `fooBar_Script`). Version can be auto-generated, and API can be automatically matched, since this is not the main point of a script plugin.

On the other hand, to prevent abuse of script plugins, PocketMine may show a big warning/notice about the script plugin being loaded.

SOF3:
I used a very simple algorithm for ~~detecting~~ guessing the main class that is prone to false-positives and false-negatives, but I believe that, for most cases, developers would not need more than this. If they really have uncommon setups, they can always use @main again.

See #2815 for details.

### DIFF
--- a/src/pocketmine/plugin/ScriptPluginLoader.php
+++ b/src/pocketmine/plugin/ScriptPluginLoader.php
@@ -23,15 +23,18 @@ declare(strict_types=1);
 
 namespace pocketmine\plugin;
 
+use function basename;
 use function file;
 use function is_file;
 use function preg_match;
 use function strlen;
 use function strpos;
+use function strtolower;
 use function substr;
 use function trim;
 use const FILE_IGNORE_NEW_LINES;
 use const FILE_SKIP_EMPTY_LINES;
+use const pocketmine\BASE_VERSION;
 
 /**
  * Simple script loader, not for plugin development
@@ -61,6 +64,28 @@ class ScriptPluginLoader implements PluginLoader{
 	 * @return null|PluginDescription
 	 */
 	public function getPluginDescription(string $file) : ?PluginDescription{
+		$data = self::findHeader($file);
+		if(isset($data["notscript"])){
+			return null;
+		}
+
+		$data += [
+			"name" => basename($file, ".php"),
+			"version" => "1.0.0",
+			"api" => BASE_VERSION,
+		];
+
+		if(!isset($data["main"])){
+			$data["main"] = $main = self::detectMainClass($file);
+			if($main === null){
+				return null;
+			}
+		}
+
+		return new PluginDescription($data);
+	}
+
+	private static function findHeader(string $file) : array{
 		$content = file($file, FILE_IGNORE_NEW_LINES | FILE_SKIP_EMPTY_LINES);
 
 		$data = [];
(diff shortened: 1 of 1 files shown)


src/pocketmine/plugin/ScriptPluginLoader.php
```diff
@@ -72,24 +97,33 @@ public function getPluginDescription(string $file) : ?PluginDescription{
...
+	private static function detectMainClass(string $file) : ?string{
+		// It is possible to use token_get_all() to analyze the file syntactically,
+		// but it is usually not necessary, and is too complex for a simple problem.
+		$namespace = "";
+		foreach(file($file) as $line){
+			if(preg_match('/^namespace[ \t]+([A-Za-z0-9_\\]+)/', $line, $match) === 0){
+				$namespace = $match[1] . "\\";
+				continue;
+			}
+			if(preg_match('/^class[ \t]+([A_Za-z0-9_]+)[ \t]+extends\b/', $line, $match) === 0){
```
SOF3: Should change to `^[ \t]+class` to allow indented classes (e.g. under a namespace block)

SOF3:
Since script plugins are designed for immediate use and testing only, it is valid to argue that the poorly defined behaviour of main class detection does not have severe impact because the developer can fix it very quickly. Guessing the main class could never be accurate, and if it ever fails to detect the main class, it is very easy to add `@main` manually and ignore the changes introduced by this pull request.

Nevertheless, I would provide this specification for no-manifest script plugins:
- The file name SHOULD BE `<PluginName>.php`.
  - I am not concerned about portability issues here, because script plugins are not supposed to be portable. If you want to share it with someone else, you should package it as a phar plugin.

- The namespace, if any, must be in this exact syntax:
```php
namespace Name\Space;
// or
namespace Name\Space{
```
While it is also valid syntax to have something like `Name \ Space`, this will not be detected by the script.

- The main class SHOULD BE the first class that appears in the file, and it must follow this syntax:

```php
class ClassName extends PluginBase{
```

whereas `PluginBase` may be changed to other aliases or adopt the FQN, but it must not contain spaces within.

This specification is the very normal practice for, I suppose, 99.9% script plugins (and even normal non-obfuscated phar plugins). If you are worried, just put back the `@main` all the time.

Remember that script plugin is only a feature intended for immediate testing, and should not be used in the long term, much less in production, due to its lack of portability.

It is of course possible for me to use [`token_get_all`](https://php.net/token-get-all) or even a library/extension like [PHP-Parser](https://github.com/nikic/PHP-Parser to accurately detect the namespace and the actual superclass, but this would be an overkill and totally unnecessary for an auxiliary feature.

dktapps:
It almost feels like you're going out of your way to avoid making the auxiliary feature work properly all the time...

SOF3:
Closing as the class name detection algorithm assumes too much unexpectable behaviour.

Issue #2825, opened 2019-03-22
Disallow non-startup plugin loading


SOF3:
## Description
Following #2823, it is now possible to further regularize plugin loading lifecycle.

Plugins must not be (or requested to be) enabled after server startup.

It is still possible for plugins to request enabling another plugin as long as the server is still starting up (generally equivalent to the `Done (%gs)` message not yet emitted). However, plugins should not be enabled after server starts. In particular, no plugins can be loaded after any player has joined.

Plugins being disabled before server stop is another business.

## Pros
- More optimizations become possible
  - E.g. Handler lists might no longer need to support online insertion
- Reduces code complexity
  - Can force more things to happen before startup, less cases to consider
- Plugins no longer need to check for players already online when the server starts
- Plugin mutual dependency can be addressed properly
  - E.g. if A uses a feature from B and B uses a feature from A, their load order doesn't matter; instead, they can initialize the components required from each other during enable and perform interaction in a just-before-startup hook.
- Plugin load order can be correctly tuned. Plugins can now be confident that they are never enabled while their `loadbefore` plugins have already been enabled, and never have `softdepend` plugins starting up after they have been enabled.
- STARTUP plugins should not be allowed to load after server start anyway

## Cons
- Plugin managers can no longer immediately enable a plugin after a plugin is installed
  - But note that there aren't many people using plugin managers
  - Even less people want to enable plugins immediately
  - Not to mention cases where plugin does not work the first time (e.g. need configuration)
  - It is not a big deal to restart a server immediately or schedule a restart, given that many servers restart regularly (although they don't really need to)
  - Someone is usually on the console in case the plugin doesn't work

SOF3:
Another issue is that plugins often cache their dependencies in startup. It is common to see something like this:

```php
public function onEnable() {
    $this->purePerms = $this->getServer()->getPluginManager()->getPlugin("PurePerms");
}
```

This may end up using the incorrect instance if *only* PurePerms is reloaded.

dktapps:
To summarise internals discussions on this topic:

The real problem that this RFC is intended to solve is that there is currently no way to say "this plugin must ONLY load at startup and no other time". It's not a very good solution for this problem; it imposes restrictions on all plugins to prevent issues with a small subset of plugins which are actually affected by this problem. TL;DR: It's a hamfisted approach.

The solution I would suggest instead is this:

[...]

PR #2828, opened 2019-03-23
Disallow plugin loading after startup

SOF3:
## Introduction
See #2825 for the motivation

### API changes
`PluginManager->loadPlugins()` and `PluginManager->loadPlugin()` can no longer be called when the server starts ticking.

### Behavioural changes
New plugins cannot be loaded without restarting the server.

## Backwards compatibility
See #2825 for discussion

## Follow-up
This behaviour should be documented.

## Tests
None yet.

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -179,6 +179,8 @@ class Server{
 
 	/** @var bool */
 	private $isRunning = true;
+	/** @var bool */
+	private $startedTicking = false;
 
 	/** @var bool */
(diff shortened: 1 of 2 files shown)


dktapps:
I'm not a fan of this. This introduces another circular dependency between PluginManager and Server.

In addition, these methods should simply not be available at all after they aren't allowed to be used anymore.

dktapps:
https://github.com/pmmp/PocketMine-MP/issues/2825#issuecomment-1015568551

PR #2833, opened 2019-03-25
Added ChestPairEvent

### LINKED ISSUE #2829: Add a ChestPairEvent
## Description
See title.

## Justification
Currently it's difficult for a plugin to detect when a chest pair is created. It may be desirable to intercept this or prevent it, for example to prevent the formation of double chests.

StarryN1ght:
## Introduction
Calls when 2 chests are about to pair, is cancellable.

Closes #2829

## Changes

* Added ChestPairEvent.php

### DIFF
--- a/src/pocketmine/block/Chest.php
+++ b/src/pocketmine/block/Chest.php
@@ -79,7 +79,7 @@ class Chest extends Transparent{
 						$pair = $this->level->getTile($c);
 						$ev = new ChestPairEvent($this, $c);
                         $ev->call();
-						if($pair instanceof TileChest and !$ev->isCanceled() and !$pair->isPaired()){
+						if($pair instanceof TileChest and !$ev->isCancelled() and !$pair->isPaired()){
 							$pair->pairWith($tile);
 							$tile->pairWith($pair);
 							break;
--- a/src/pocketmine/block/Chest.php
+++ b/src/pocketmine/block/Chest.php
@@ -77,15 +77,15 @@ class Chest extends Transparent{
 					$c = $this->getSide($side);
 					if($c instanceof Chest and $c->isSameType($this) and $c->facing === $this->facing){
 						$pair = $this->level->getTile($c);
-                        if($pair instanceof TileChest and !$pair->isPaired()){
-                        	$ev = new ChestPairEvent($this, $c);
+						if($pair instanceof TileChest and !$pair->isPaired()){
+							$ev = new ChestPairEvent($this, $c);
 							$ev->call();
 							if(!$ev->isCancelled()){
 								$pair->pairWith($tile);
 								$tile->pairWith($pair);
-                            }
-                            break;
-                        }
+							}
+							break;
+						}
 					}
 				}
 			}
--- a/src/pocketmine/block/Chest.php
+++ b/src/pocketmine/block/Chest.php
@@ -24,6 +24,7 @@ declare(strict_types=1);
 namespace pocketmine\block;
 
 use pocketmine\block\utils\BlockDataValidator;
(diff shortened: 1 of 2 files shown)


src/pocketmine/block/Chest.php
```diff
@@ -76,7 +77,9 @@ public function place(Item $item, Block $blockReplace, Block $blockClicked, int
 					$c = $this->getSide($side);
 					if($c instanceof Chest and $c->isSameType($this) and $c->facing === $this->facing){
 						$pair = $this->level->getTile($c);
-						if($pair instanceof TileChest and !$pair->isPaired()){
+						$ev = new ChestPairEvent($this, $c);
+                        $ev->call();
+						if($pair instanceof TileChest and !$ev->isCancelled() and !$pair->isPaired()){
```
dktapps: This logic order makes no sense. A plugin will receive this event when a pairing is not going to happen due to the target already being paired.

src/pocketmine/event/block/ChestPairEvent.php
```diff
@@ -0,0 +1,69 @@
...
+     * @param Chest $pair
+     */
+    public function __construct(Chest $chest, Chest $pair){
+        parent::__construct($chest);
+        $this->chest = $chest;
+        $this->pair = $pair;
+    }
+    
+    /**
+     * Returns the chest thats about to pair with the chest thats already placed.
```
dktapps: Grammar errors here - `thats` should be `that's` or `that is`.

src/pocketmine/event/block/ChestPairEvent.php
```diff
@@ -0,0 +1,69 @@
...
+     *
+     * @return Chest
+     */
+    public function getChest() : Chest{
+        return $this->chest;
+    }
+    
+    /**
+     * Returns the chest thats already placed.
```
dktapps: Same as above.

src/pocketmine/block/Chest.php
```diff
@@ -76,7 +77,9 @@ public function place(Item $item, Block $blockReplace, Block $blockClicked, int
 					$c = $this->getSide($side);
 					if($c instanceof Chest and $c->isSameType($this) and $c->facing === $this->facing){
 						$pair = $this->level->getTile($c);
-						if($pair instanceof TileChest and !$pair->isPaired()){
+						$ev = new ChestPairEvent($this, $c);
+                        $ev->call();
```
dktapps: Formatting issue here. Use tabs, not spaces.

dktapps (changes requested):
Looks interesting. I've left some comments.

StarryN1ght:
Btw, do you have any suggestions for the function `getPair()` name?

I was just going off the variables name when making it.

dktapps:
I would prefer getLeft() and getRight(). Chest pairing is not instigated by any specific node from a plugin's perspective.

src/pocketmine/event/block/ChestPairEvent.php
```diff
@@ -0,0 +1,69 @@
...
+namespace pocketmine\event\block;
+
+use pocketmine\block\Chest;
+use pocketmine\event\Cancellable;
+use pocketmine\event\CancellableTrait;
+
+/**
+ * Called when a chest pairs with another chest.
+ */
+class ChestPairEvent extends BlockEvent implements Cancellable{
```
dktapps: Extending BlockEvent doesn't make sense in this case I think, since there are 2 blocks involved. There may be confusion over the return value of `getBlock()` so I think it's better to omit it from the inheritance and extend `Event` directly.

fuyutsuki:
looks nice!

PR #2840, opened 2019-03-28
Call BlockTeleportEvent when dragon egg teleports

Frago9876543210:
## Introduction
Now when the dragon egg tries to teleport it is calls BlockTeleportEvent


dktapps:
This is an abuse of BlockUpdateEvent and not what it's designed for.

Frago9876543210:
What event do you suggest?

HimbeersaftLP:
DragonEggTeleportEvent obviously

Frago9876543210:
One event per feature? Okey, will commit tomorrow

dktapps:
The logic is specialized to this specific block. Go with `BlockTeleportEvent` if you're bothered. This logic will be separated into a component in the future.

PR #2852, opened 2019-03-31
Changed block and item IDs to Chinese numerals

SOF3:
Following our partnership with NetEase, our development will be outsourced to China. We should change our constants to Chinese numerals to enhance collaboration, as well as to prevent developers in the western world to hardcode block and item IDs. After all, we will only provide free support to Chinese developers, so it's easier to help Chinese developers who hardcode the item IDs.


dktapps:
This is ahead of its time since we haven't made an official announcement yet.

SOF3:
Authorities have called an end to our partnership with NetEase due to political incorrectness as pointed out by @PEMapModder. I am sorry to see that nobody is taking up the development after @dktapps' exit from our organization.

nathfreder:
@Encritary Of course not, @pmmp is the most serious development team.

PR #2857, opened 2019-04-05
Update SplashPotion.php

Vrysed:
## Introduction
When you splash the potion on the floor directly below you it gives you full effect when it shouldn't. The only way you can get the full effect is when the potion collides with entity.

### DIFF
--- a/src/pocketmine/entity/projectile/SplashPotion.php
+++ b/src/pocketmine/entity/projectile/SplashPotion.php
@@ -93,7 +93,7 @@ class SplashPotion extends Throwable{
 							continue;
 						}
 
-						$distanceMultiplier = 1 - (sqrt($distanceSquared) / 4);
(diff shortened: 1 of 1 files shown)


dktapps:
Have you tested this in vanilla? This code is wrong...

dktapps:
I've just tested vanilla vs the existing code, and I must conclude you did not test it yourself, because the existing code matches vanilla almost perfectly (give or take a few seconds).

Vrysed:
In Pocketmine when you splash the potion directly underneath you, you get the full effect. In Vanilla when you splash it directly underneath you, you do not get the full effect. And Yes I tested it.

dktapps:
I did test it. My results do not support your conclusion.

lukeeey:
Now we just search through github to find which spoon this came from...

dktapps:
We need a spoon index.

lukeeey:
spoondex.org

PR #2863, opened 2019-04-08
TextFormat: Added compression function

dktapps:
## Introduction
This function will reduce a given string to the minimum amount of formatting codes possible to produce the smallest possible string with the same appearance as the input.

This is needed because formatting code spam can be abused to produce physically large but visually small strings.

### API changes
Added new function `TextFormat::compress()`. This is not yet used, but is planned to be used for chat and sign text compression.

## Follow-up
Use this function to compress formatting codes given by users where appropriate, such as in chat, signs, books and item names.

## Tests
PHPUnit tests are included with some test cases, but it would be appreciated if this test case set could be expanded.

### DIFF
--- a/src/pocketmine/utils/TextFormat.php
+++ b/src/pocketmine/utils/TextFormat.php
@@ -59,12 +59,39 @@ abstract class TextFormat{
 	public const YELLOW = TextFormat::ESCAPE . "e";
 	public const WHITE = TextFormat::ESCAPE . "f";
 
+	public const COLORS = [
+		self::BLACK => true,
+		self::DARK_BLUE => true,
+		self::DARK_GREEN => true,
+		self::DARK_AQUA => true,
+		self::DARK_RED => true,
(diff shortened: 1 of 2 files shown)


dktapps:
There have been concerns raised about the removal of &r formatting codes when the string is going to be used as a parameter inside another string, such as a translation. This should be discussed prior to merge.

SOF3:
Is it possible to introduce the nonstandard tags §( and §) that PocketMine preprocesses internally, for the sake of plugin compatibility?

dktapps:
It's theoretically possible, but unwise since it might cause side effects when used by clients.

dktapps:
Closing this due to inactivity, conflicts and little practical value.

dktapps:
The problems this faced are closely linked to #4116 , and the solution to #4116 will likely involve a formatting code preprocessing function very similar to the one written here.

PR #2900, opened 2019-05-02
Implement anvil destorying torches whilst falling down

### LINKED ISSUE #2895: Sand falling on top of the cactus does not drop
### Issue description

- Expected result: What were you expecting to happen?
Sand falling on top of the cactus does not drop/break the sand block.
- Actual result: What actually happened?
Sand is placed on top of the cactus.

1. Place down a cactus (on sand of course).
2. Place a sand block 2 blocks above the cactus.
3. The sand drops right on top of the cactus and stays there contrary to vanilla. (Screenshots attached)

### OS and versions

* PocketMine-MP: 3.8.1
* PHP: 7.2
* Server OS: Windows
* Game version: PE/Win10

### Plugins

`none`

### Screenshots

- On PocketMine-MP:

- In Vanilla:
a

geNAZt:
Fix cactus not having vanilla behaviour when sand falls on top.

## Introduction
Added a first entity behaviour. DestroyWhilstFalling can destroy/drop blocks whilst touching them.

### Relevant issues

* Fixes #2895 

### Behavioural changes
The server now drops all sort of torches when a falling anvil touches them. Also sand gets dropped when hit by a cactus.

### DIFF
--- a/src/pocketmine/block/Block.php
+++ b/src/pocketmine/block/Block.php
@@ -849,7 +849,7 @@ class Block extends Position implements BlockLegacyIds, Metadatable{
 	 * 			be destroyed and falling should continue
 	 */
 	public function canBlockLand(Fallable $block) : int{
-		return ($this->getId() > 0 and !$this->isTransparent() and !$this->canBeReplaced()) ?
+		return (!$this->isTransparent() and !$this->canBeReplaced()) ?
 			FallingBlock::STATE_SOLIDIFY : FallingBlock::STATE_DROP_ITEM;
 	}
 
--- a/src/pocketmine/block/Block.php
+++ b/src/pocketmine/block/Block.php
@@ -849,8 +849,7 @@ class Block extends Position implements BlockLegacyIds, Metadatable{
 	 * 			be destroyed and falling should continue
 	 */
 	public function canBlockLand(Fallable $block) : int{
-		return (!$this->isTransparent() and !$this->canBeReplaced()) ?
-			FallingBlock::STATE_SOLIDIFY : FallingBlock::STATE_DROP_ITEM;
+		return FallingBlock::STATE_SOLIDIFY;
 	}
 
 }
--- a/src/pocketmine/block/Block.php
+++ b/src/pocketmine/block/Block.php
@@ -26,8 +26,10 @@ declare(strict_types=1);
  */
 namespace pocketmine\block;
 
+use pocketmine\block\utils\Fallable;
 use pocketmine\block\utils\InvalidBlockStateException;
 use pocketmine\entity\Entity;
+use pocketmine\entity\object\FallingBlock;
 use pocketmine\item\enchantment\Enchantment;
 use pocketmine\item\Item;
 use pocketmine\item\ItemFactory;
@@ -837,4 +839,18 @@ class Block extends Position implements BlockLegacyIds, Metadatable{
 			$this->level->getBlockMetadata()->removeMetadata($this, $metadataKey, $owningPlugin);
 		}
 	}
+
+	/**
+	 * Check if the falling block can be made solid again when hitting this block
+	 *
+	 * @param Fallable $block which is currently falling
+	 *
+	 * @return 	int 0 when block can be solidfy, 1 when the block should drop as item, 2 when this block should
+	 * 			be destroyed and falling should continue
+	 */
+	public function canBlockLand(Fallable $block) : int{
+		return ($this->getId() > 0 and !$this->isTransparent() and !$this->canBeReplaced()) ?
+			FallingBlock::STATE_SOLIDIFY : FallingBlock::STATE_DROP_ITEM;
+	}
+
 }
--- a/src/pocketmine/block/Cactus.php
+++ b/src/pocketmine/block/Cactus.php
@@ -24,7 +24,9 @@ declare(strict_types=1);
 namespace pocketmine\block;
 
 use pocketmine\block\utils\BlockDataValidator;
+use pocketmine\block\utils\Fallable;
 use pocketmine\entity\Entity;
+use pocketmine\entity\object\FallingBlock;
 use pocketmine\event\block\BlockGrowEvent;
 use pocketmine\event\entity\EntityDamageByBlockEvent;
 use pocketmine\event\entity\EntityDamageEvent;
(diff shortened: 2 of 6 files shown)


src/pocketmine/block/Cactus.php
```diff
@@ -127,4 +129,18 @@ public function place(Item $item, Block $blockReplace, Block $blockClicked, int
...
 		return false;
 	}
+
+	public function canBlockLand(Fallable $block) : int{
+		$parentDecision = parent::canBlockLand($block);
+		if ($parentDecision == FallingBlock::STATE_DROP_ITEM) {
+			return $parentDecision;
+		}
+
+		if ($block instanceof Sand) {
```
dktapps: what about concrete?

src/pocketmine/block/Block.php
```diff
@@ -837,4 +839,18 @@ public function removeMetadata(string $metadataKey, Plugin $owningPlugin) : void
...
+	/**
+	 * Check if the falling block can be made solid again when hitting this block
+	 *
+	 * @param Fallable $block which is currently falling
+	 *
+	 * @return 	int 0 when block can be solidfy, 1 when the block should drop as item, 2 when this block should
+	 * 			be destroyed and falling should continue
+	 */
+	public function canBlockLand(Fallable $block) : int{
+		return ($this->getId() > 0 and !$this->isTransparent() and !$this->canBeReplaced()) ?
```
dktapps: We're moving away from legacy blockIDs, so please avoid this usage.

src/pocketmine/entity/behaviour/Behaviour.php
```diff
@@ -0,0 +1,38 @@
...
+namespace pocketmine\entity\behaviour;
+
+use pocketmine\entity\Entity;
+
+abstract class Behaviour{
+
+	/** @var Entity */
+	protected $entity;
+
+	protected function __construct(Entity $entity) {
```
dktapps: formatting issue here.

src/pocketmine/entity/object/FallingBlock.php
```diff
@@ -70,6 +81,10 @@ protected function initEntity(CompoundTag $nbt) : void{
 		$damage = $nbt->getByte("Data", 0);
 
 		$this->block = BlockFactory::get($blockId, $damage);
+		if ($this->block instanceof Anvil) {
+			$this->behaviours = [new DestroyWhileFalling($this,	[Block::TORCH, Block::COLORED_TORCH_RG, Block::COLORED_TORCH_BP,
```
dktapps: Not a fan of this. This should be a property of the blocks themselves.

dktapps (changes requested):
I've given this a quick look, but not in depth.

src/pocketmine/block/Cactus.php
geNAZt: is there a list of blocks which has the same behaviour? does it work for all falling blocks?

dktapps: As far as I know, yes.

src/pocketmine/entity/object/FallingBlock.php
dktapps: There's also a rogue tab in the argument list here.

geNAZt: There is no property system for blocks in PMMP so i would have to generate a getter for it. Also this would not allow for any dynamic lists (since the behaviour is then dependent on the block property) of destoryable blocks.

lukeeey:
I see you have given up on GoMint

src/pocketmine/block/Block.php
```diff
@@ -837,4 +839,18 @@ public function removeMetadata(string $metadataKey, Plugin $owningPlugin) : void
...
+	/**
+	 * Check if the falling block can be made solid again when hitting this block
+	 *
+	 * @param Fallable $block which is currently falling
+	 *
+	 * @return 	int 0 when block can be solidfy, 1 when the block should drop as item, 2 when this block should
+	 * 			be destroyed and falling should continue
+	 */
+	public function canBlockLand(Fallable $block) : int{
+		return (!$this->isTransparent() and !$this->canBeReplaced()) ?
```
dktapps: this will cause unintended side effects when dropped on glass.

src/pocketmine/block/Cactus.php
```diff
@@ -127,4 +129,14 @@ public function place(Item $item, Block $blockReplace, Block $blockClicked, int
...
 		return false;
 	}
+
+	public function canBlockLand(Fallable $block) : int{
+		$parentDecision = parent::canBlockLand($block);
+		if ($parentDecision == FallingBlock::STATE_DROP_ITEM) {
+			return $parentDecision;
+		}
+
+		return FallingBlock::STATE_DROP_ITEM;
```
dktapps: The logic in this function seems redundant. It seems like this could be reduced to a single line.

src/pocketmine/entity/object/FallingBlock.php
```diff
@@ -70,6 +81,10 @@ protected function initEntity(CompoundTag $nbt) : void{
 		$damage = $nbt->getByte("Data", 0);
 
 		$this->block = BlockFactory::get($blockId, $damage);
+		if ($this->block instanceof Anvil) {
+			$this->behaviours = [new DestroyWhileFalling($this, [Block::TORCH, Block::COLORED_TORCH_RG, Block::COLORED_TORCH_BP,
+					Block::LIT_REDSTONE_TORCH, Block::UNLIT_REDSTONE_TORCH, Block::UNDERWATER_TORCH], true)];
```
dktapps: I don't like this. I'm trying to get rid of legacy block IDs. In addition, you're creating a whole new system here for one thing, which makes no sense and is inconsistent with the rest of the code.

It seems like this logic can be dropped and just add `canBlockLand()` implementations for these blocks.

dktapps: In fact, the logic for this already exists, so this code can be removed entirely.

src/pocketmine/entity/object/FallingBlock.php
```diff
@@ -105,20 +120,42 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
...
+						break;
+
+					case self::STATE_DROP_ITEM:
+						$this->flagForDespawn();
+						$this->getLevel()->dropItem($this, $this->block->asItem());
+						break;
+
+					case self::STATE_DESTROY_BLOCK:
+						$this->level->setBlock($blockDown, Block::get(Block::AIR));
+						$this->getLevel()->dropItem($blockDown, $blockDown->asItem());
```
dktapps: This should use `Level->useBreakOn()`.

src/pocketmine/block/Cactus.php
dktapps: This problem has not been resolved.

dktapps:
Closing due to lack of activity on requested changes.

PR #2901, opened 2019-05-02
Implement a locking variable to circumvent a infinite regression

### LINKED ISSUE #2876: Infinite loop when using Entity->close() during EntityDespawnEvent
Plugin developer stupidity could lead to this circumstance which can be problematic, because `Entity->closed` is not set to true until after the event.

geNAZt:
## Introduction
The current problem is that a plugin could enter a infinite loop when calling Entity::close in a EntityDespawnEvent.

### Relevant issues

* Fixes #2876

### API changes
Entity::close may not fire any additional events when called from inside the EntityDespawnEvent. 

## Backwards compatibility
I haven't touched the $closed ordering to fix this issue since it would break the event contract on the entity. That should maybe be done on 4.0 but not on 3.x

## Follow-up
Think about reordering close variable state inside Entity::close

### DIFF
--- a/src/pocketmine/entity/Entity.php
+++ b/src/pocketmine/entity/Entity.php
@@ -562,6 +562,8 @@ abstract class Entity extends Location implements Metadatable, EntityIds{
 	/** @var bool */
 	protected $constructed = false;
 
+	/** @var bool */
+	private $closeInFlight = false;
 
 	public function __construct(Level $level, CompoundTag $nbt){
 		$this->constructed = true;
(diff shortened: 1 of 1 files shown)


src/pocketmine/entity/Entity.php
```diff
@@ -2108,7 +2110,12 @@ public function isClosed() : bool{
 	 * WARNING: Entities are unusable after this has been executed!
 	 */
 	public function close() : void{
+		if($this->closeInFlight) {
```
dktapps: just one nit here (formatting)

geNAZt: Tell me pls :(

dktapps: pls remove the space between ) and {

dktapps:
you just introduced a bunch of unrelated whitespace changes.

lukeeey:
like all his prs lol

PR #2906, opened 2019-05-05
Changed ignoreCancelled opt-in to handleCancelled opt-in

SOF3:
## Introduction
Ignoring cancelled events is usually the desired behaviour, and should be the default if undetermined (or casually missed).

In addition, the `@ignoreCancelled` tag has been ambiguous. Some people tend to take the double-negation interpretation of "ignore whether an event is cancelled and fire it anyway" rather than "ignore cancelled events and don't handle it". Calling it `@handleCancelled` removes the ambiguity.

### API changes
Declaring `@ignoreCancelled` will now trigger error.

Declaring `@handleCancelled` will now cause cancelled events to be handled.

Not declaring `@handleCancelled` (or `@handleCancelled false`) will now cause cancelled events to not be handled. This is a BC break.

## Backwards compatibility
All current plugins will no longe rreceive cancelled events unless explicitly declaring `@handleCancelled`.

## Tests
To be added.

### DIFF
--- a/src/pocketmine/event/Listener.php
+++ b/src/pocketmine/event/Listener.php
@@ -46,7 +46,7 @@ use pocketmine\plugin\PluginManager;
  * - `@softDepend [PluginName]`: Handler WILL NOT be registered if its event doesn't exist. Useful for soft-depending
  *     on plugin events. Plugin name is optional.
  *     Example: `@softDepend SimpleAuth`
- * - `@ignoreCancelled`: Cancelled events WILL NOT be passed to this handler.
+ * - `@handleCancelled`: Cancelled events will STILL invoke this handler.
  * - `@priority <PRIORITY>`: Sets the priority at which this event handler will receive events.
  *     Example: `@priority HIGHEST`
  *     @see EventPriority for a list of possible options.
--- a/src/pocketmine/plugin/PluginManager.php
+++ b/src/pocketmine/plugin/PluginManager.php
@@ -532,6 +532,10 @@ class PluginManager{
 					throw new PluginException("Event handler " . Utils::getNiceClosureName($handlerClosure) . "() declares invalid/unknown priority \"" . $tags["priority"] . "\"");
 				}
 
+				if(isset($tags["ignoreCancelled"])){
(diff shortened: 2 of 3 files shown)


src/pocketmine/event/Listener.php
```diff
@@ -46,7 +46,7 @@
  * - `@softDepend [PluginName]`: Handler WILL NOT be registered if its event doesn't exist. Useful for soft-depending
  *     on plugin events. Plugin name is optional.
  *     Example: `@softDepend SimpleAuth`
- * - `@ignoreCancelled`: Cancelled events WILL NOT be passed to this handler.
+ * - `@handleCancelled`: Cancelled events will STILL invoke this handler.
```
dktapps: Perhaps mention that the default will be to NOT pass this.

src/pocketmine/plugin/PluginManager.php
```diff
@@ -532,22 +532,25 @@ public function registerEvents(Listener $listener, Plugin $plugin) : void{
 					throw new PluginException("Event handler " . Utils::getNiceClosureName($handlerClosure) . "() declares invalid/unknown priority \"" . $tags["priority"] . "\"");
 				}
 
-				$ignoreCancelled = false;
 				if(isset($tags["ignoreCancelled"])){
-					switch(strtolower($tags["ignoreCancelled"])){
+					throw new PluginException("Event handler " . Utils::getNiceClosureName($handlerClosure) . "() declares obsolete tag @ignoreCancelled");
```
dktapps: This doesn't make a whole lot of sense. The new behaviour is to "ignore cancelled" anyway, so this is just a patchy way to warn of a behavioural change that didn't have any effect in this case anyway.

SOF3: This is like a shouty warning to let people notice about this change, so although it usually targets victims that are not affected by the new behavioiur, it is still ok.

dktapps: It's pointless since it will only shout at people who aren't affected.

SOF3:
I expected more response to this pull request as this ~~breaks~~ changes the behaviour of almost every plugin that existed.

PR #2907, opened 2019-05-06
First shot renaming Level -> World

dktapps:
This has been discussed for a while, but I decided to do it today because YOLO.


dktapps:
Due to the problems created by chunkutils2, namespaces for classes presented by chunkutils2 are untouched by this PR. Those can be changed later on once we're happy with this.

lukeeey:
We need some more of this YOLO attitude in PM :smile:

dktapps:
This change is huge, so it had to take top priority before it develops millions of merge conflicts.

TheShermanTanker:
> 
> 
> We need some more of this YOLO attitude in PM 😄

If only Cloudburst had the same attitude :/

PR #2946, opened 2019-05-31
Removed PluginLoader (resolves #2834)

### LINKED ISSUE #2825: Disallow non-startup plugin loading

### LINKED ISSUE #2834: Removal of PluginLoader
Remove PluginLoader interface entirely. Instead, third party plugin loaders can actively pass the plugins they found into the PluginManager, in the format of a `(PluginDescription, function(): Plugin)` tuple, and let PluginManager decide when to load the plugins.

This reduces the complexity caused by the PluginLoader interface, and opens for higher flexibility of loading plugins through removing the constraint of load-from-path.

This proposal can be more easily implemented based on #2825.

Plugin loaders such as devtools will require backward-incompatible changes.

### LINKED ISSUE #2838: Misleading naming of PluginLoadOrder and related directives
## Description
This issue proposes that `PluginLoadOrder` should be renamed to `PluginEnableOrder` and its plugin.yml directive renamed to `enable` instead of `load`.

## Justification
To add another thing to the list of poorly-chosen names in the API (did this also come from Bukkit?), `PluginLoadOrder` and its related `load` setting in plugin manifests do not make the plugin **load** at these time points, but rather makes them be **enabled** at these points. This can cause confusion amongst plugin developers using `onLoad()`.

In addition, `onLoad()` is entirely useless if the #2825 proposal is adopted.

### LINKED ISSUE #2942: Extract a ResourceLoader unit from PluginBase
## Description
The methods `getResource()` and `getResources()` of `PluginBase` should be extracted to a separate unit / interface.

## Justification
Currently resource handling is hard-baked into the core of PluginBase. This is problematic because resource data accessing requires specialized logic to retrieve information which depends on the plugin loader.

This has been partially addressed with the use of access protocol on plugin loaders, but this limits the methods by which resources can be loaded to things addressable as file paths relative to the plugin's own root file.

It has been discussed that a plugin may not _have_ a "root file" to begin with, since this is actually not necessary for the design of a plugin.

In addition, it is currently not possible to modify the way that, for example, a script plugin accesses its resources (it can't).

We propose extracting a `ResourceLoader` interface, implemented by a `FilesystemResourceLoader` which is used for folder and phar plugins. A `NullResourceLoader` can be implemented for script plugins. This will be injected via the constructor, replacing the path parameter.

### LINKED ISSUE #2944: Plugin dependency is not reliable
Due to complex problems such as plugin load orders, late plugin enabling, etc., plugin dependencies might sometimes not be satisfied.

In addition, disabling a dependency plugin does not disable the corresponding dependents, leading to instability.

### LINKED ISSUE #2945: Shuffle plugins before loading
In general, plugin load order is undefined unless constrained by `load`, `depend`, `softDepend`, `loadBefore`.
However, due to OS/filesystem level reasons, plugin load order is usually based on lexicographical order. This causes some bugs to stop being noticed, e.g. a plugin starting with "Z" rarely needs to explicitly declare dependency on other plugins.
This issue proposes the shuffling of plugin loading. By randomizing the order that plugin files are scanned by PocketMine, lexicographical order of plugin files/names would have no effect on loading order, thus revealing these long-undisclosed issues.

SOF3:
- Plugin loading is performend by actively calling PluginManager->loadPlugin()
- Plugin main is now only loaded just before onEnable
- onLoad() is deprecated; overriding leads to crash
  - This results in "load" and "enable" having the same meaning, thus resolves #2838
- Script plugins must extend pocketmine\plugin\ScriptPlugin instead of pocketmine\plugin\PluginBase
- Plugins cannot be loaded beyond PluginLoadOrder::SCAN (resolves #2825)
- Takes into account the contradiction of PluginLoadOrder and dependencies, fixes #2944
- Replaces and closes #2748 as PluginLoader is removed
- Shuffles plugins before toposort, resolves #2945

### API changes
TODO

### Behavioural changes

TODO

## Backwards compatibility

TODO

## Follow-up
Requires translations:

| Name | Value in eng.ini |
| :--: | :---: |
| `foo.bar` | `Foo bar` |

TODO

## Tests
TODO


SOF3:
This pull request is a bit bloated. Any suggestions to split it down? Most changes seem to be very inter-related.

Issue #2865, opened 2019-04-09
Introduce a new Form interface (or break the existing one)


dktapps:
## Description
The current `pocketmine\form\Form` interface sucks because it does not allow forms to serialize themselves in a context-dependent manner. For example, it is currently difficult to do translatable forms, which creates complications for third party implementations.

SOF3:
In my opinion, this is the responsibility of the forms library. In fact, we don't even define what a `Form` is &mdash; Is it just a representation of content, like HTML? Or is it supposed to be dynamic? Is it supposed to be used on just a single player for a single time, or should we allow the same instance to be used on multiple players simultaneously, or the same player multiple times, or both? Now this gets us back to those debates we had while writing the `forms-api` branch.

Recall that the `Form` interface is there for minimalism. I don't think we can easily reach a consensus to have more than what we currently have, because that leads to a lot of decisions like object states, object lifetime, etc. Currently, libraries are doing just well. If people want to hold the same `$form` object to use on multiple players/occasions, they could just use a library that provides something like this:

```php
class DynamicForm{
    private $data; // and code to initialize $this->data
    public function send(Player $player){
        $translated = translate($data, $player->getLanguage());
        $player->sendForm(new class implements Form{
            // boilerplate to delegate variabies
            public function jsonSerialize(){ return $this->translated; }
        });
    }
}
```

There is no need to increase the complexity of what we already have just to accomplish something that could easily be implemented through libraries. (And almost everyone uses libraries for sending forms anyway)

I am voting to close this issue.

dktapps:
It turns out that this is unnecessary because a translation-supporting form doesn't have to implement `Form` directly and can have a `Form` implementation whose sole purpose is to contain post-processed data. Forms would need to stick something over the top of `sendForm()`, but this isn't a big deal.

PR #2964, opened 2019-06-14
Removed the Form interface

SOF3:
## Introduction

The Form interface is redundant. What PocketMine needs is just a JSON blob and a callback.

### Relevant issues

Previously discussed in #2865

### API changes
- Removed pocketmine\form\Form
- `Player->sendForm(array $data, Closure $callback)`
- `NetworkSession->onFormSubmit(int $id, array $data)`

## Backwards compatibility

API BC-breaking change.

## Follow-up

The `pocketmine\form\` namespace now only contains one class. This is a bit fishy.


src/pocketmine/Player.php
```diff
@@ -2277,14 +2277,15 @@ public function sendTip(string $message){
...
 	 * Sends a Form to the player, or queue to send it if a form is already open.
 	 *
-	 * @param Form $form
+	 * @param array $data
+	 * @param \Closure $callback
 	 *
 	 * @throws \InvalidArgumentException
 	 */
-	public function sendForm(Form $form) : void{
+	public function sendForm(array $data, \Closure $callback) : void{
```
dktapps: pls add a signature check here with `Utils::validateCallableSignature()`

src/pocketmine/Player.php
```diff
@@ -2295,18 +2296,19 @@ public function sendForm(Form $form) : void{
...
 			return false;
 		}
 
 		try{
-			$this->forms[$formId]->handleResponse($this, $responseData);
+			$closure = $this->formHandlers[$formId];
+			$closure($this, $responseData);
 		}catch(FormValidationException $e){
-			$this->logger->critical("Failed to validate form " . get_class($this->forms[$formId]) . ": " . $e->getMessage());
+			$this->logger->critical("Failed to validate form: " . $e->getMessage());
```
dktapps: might be better to use `Utils::getNiceClosureName()` here for debug info instead of removing it completely.

dktapps (commented):
Concept ACK, I've made note of some minor nits.

HimbeersaftLP:
Imo this reduces code readability because the Form interface made it more clear what you're passing to a function.

SOF3:
Using your logic, the `Array.reduce(initialValue, reduceFn)` function should be changed into `Array.reduce(Reducer)`, e.g. they should have `[1, 2, 3].reduce(new MultiplyReducer())` rather than `[1, 2, 3].reduce(1, (a,b)=>a*b)`.
This apparently doesn't make sense. The sendForm function is a low level function; it should be called from form libraries, but directly on the user land. There is no reason to coerce two barely related parameters into a single object.

dktapps:
To clarify the motivation for this: The form interface has 2 jobs:
1) being json-encodable
2) providing a result handling function

These two jobs are unrelated in functionality. Having them bound together creates a needless complexity.

For example, I wanted to implement a form type which was translated on the fly a la #2865 . This would require that the `Form` interface be changed to allow a `Player` context to be passed.
As it turns out, it's possible to do this with the following glue code:
https://media.discordapp.net/attachments/373199722573201410/589111049962258455/unknown.png
However, this glue code is only necessary because of a needless restriction in the API that enforces the JSON data provider and the response handler must both be in the same class.

The bottom line is that the Form interface forces you to do two jobs at once in the same, which is an unnecessary limitation.

HimbeersaftLP:
oh

dktapps:
The main thing that stopped me from merging this is that `sendForm()` was introduced to make it simple to send forms, regardless of the implementation used. By implementing this change, it would be necessary for plugin devs to change styles again, creating confusion.

For example, in the old days, `FormAPI` had a `sendToPlayer()` method. `sendForm()` eliminated the need for this, so now we had `player->sendForm(form)`. This change would a reversion back to those days, which calls into question why we added this method at all. It still exposes the protocol internals to plugins (by way of the JSON schema for forms, which might be different from version to version), so it's not really providing a lot of value other than just a shiny wrapper around `sendDataPacket(ModalFormRequestPacket::create(...))`.

The short of this is that I think removing this interface ruins the usability of this method, while also not being very useful for anything else.

SOF3:
The usability of this method would actually be enhanced if we convert the callback parameter to some sort of Promise instead.

dktapps:
I'm not sure how practical it would be to return a Promise for this without a builtin form implementation (we wouldn't be able to statically infer types otherwise).

dktapps:
I'm not sure how practical it would be to return a Promise for this without a builtin form implementation (we wouldn't be able to statically infer types otherwise).

SOF3:
Perhaps we keep the form interface and have it process the data then?

```php
/**
 * @phpstan-template R processed client data
 */
interface Form {
    public function getData();
    /**
     * @phpstan-return R
     */
    publc function processResponse($rawClientResponse);
}

class Player {
    /**
     * @phpstan-template R
     * @phpstan-param Form<R>
     * @phpstan-return Promise<R>
     */
    public function sendForm(Form $form) : Promise { ... }
}
```

SOF3:
Is it true that this issue does not cater for client response validation? In that case, it might be meaningful to keep the interface after all.

dktapps:
The response validation is done by the implementation.

dktapps:
This PR is now heavily stale.

Considering the relatively tiny benefits offered by this PR, and the large scale disruption it would cause to existing plugins, I don't think it makes much sense to proceed.

PR #2970, opened 2019-06-19
ChunkSerializer: added deserialize method

Frago9876543210:
## Introduction
It is assumed that this thing can be used in PocketMine-related software

### API changes
Added ```ChunkSerializer::deserialize(chunkX, chunkZ, buffer)```


dktapps:
This code is useless as is because the block compression dictionary won't be the same across server nodes.

Frago9876543210:
wdym?

dktapps:
The block mapping table isn't designed to run in a non-server context, and it's not possible to provide block mappings for deserialization. This code is therefore useless, since the server will never be decoding its own chunks.

PR #2974, opened 2019-06-20
Implemented StructureGrowEvent

matcracker:
## Introduction

Currently is not possible to detect when a tree (or better a structure) is grew.

### Relevant issues

- Need to implement Huge Mushroom **(nothing to do with this addition)**.
- Need to get a list of all blocks associated with the structure.
**Currently is not possible to get all structure blocks due to internal usage of BlockTransaction in Tree::placeObject()**

### API changes

**Added StructureGrowEvent.**
Event that is called when an organic structure attempts to grow (Sapling -> Tree), (Mushroom -> Huge Mushroom), naturally or using fertilizer (bonemeal).

### Behavioural changes

Doesn't change server behavior.

## Backwards compatibility

It's backward compatible.

## Tests

Tested actions:
Event cancelled -> Tree is not grew.

### DIFF
--- a/src/pocketmine/event/world/StructureGrowEvent.php
+++ b/src/pocketmine/event/world/StructureGrowEvent.php
@@ -1,0 +1,92 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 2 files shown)


src/pocketmine/event/world/StructureGrowEvent.php
```diff
@@ -0,0 +1,92 @@
...
+        $this->species = $species;
+        $this->fertilizer = $fertilizer;
+        $this->player = $player;
+    }
+
+    /**
+     * Returns the structure coordinates.
+     * @return Vector3
+     */
+    public function getVector3(): Vector3{
```
dktapps: Mushrooms don't have a tree specie.

matcracker: According to wiki, it says that is a 'tree-like', so I considered like a tree.
https://minecraft.gamepedia.com/Huge_mushroom

lukeeey:
May i suggest adding this as a subclass of BlockGrowEvent. Also, "Structure" is a bit ambiguous.

And if you dont want to make a subclass of BlockGrowEvent you could incorporate it into the event itself.

matcracker:
I was thinking too, but this is more related to a world event.
BlockGrowEvent is related to grow the block self (block to block).
StructureGrowEvent is related to grow a block to a pre-set structure (tree in that case).

lukeeey:
I think it would work as both as, while it is a structure, it is also the result of a block being grown

dktapps:
BlockGrowEvent doesn't fit. Look at its API.

dktapps:
Closing due to inactivity on review comments.

PR #2977, opened 2019-06-21
 Add API methods to change a players' player list name

lukeeey:
## Introduction
Since the ability to update the player list was removed when calling `setDisplayName` (and rightly so), we need another method update the name. 

### API changes
Add the following methods to Player:
* `getPlayerListName()`
* `setPlayerListName(string $name)`

## Backwards compatibility
It is but does it even really matter at this point? lol

## Follow-up
Maybe use the player list name instead of the username in `NetworkSession->onPlayerAdded()` as the plain username will be displayed until the player actually spawns (if you use my join event example, i'm not sure if it would be possible to change the name in an earlier event)

## Tests
```php
public function onJoin(PlayerJoinEvent $event): void {
        $player->setPlayerListName(TF::LIGHT_PURPLE . "[PLN] " . $player->getName());
    }
```

### DIFF
--- a/src/network/mcpe/NetworkSession.php
+++ b/src/network/mcpe/NetworkSession.php
@@ -983,7 +983,7 @@ class NetworkSession{
 	}
 
 	public function onPlayerAdded(Player $p) : void{
-		$this->sendDataPacket(PlayerListPacket::add([PlayerListEntry::createAdditionEntry($p->getUniqueId(), $p->getId(), $p->getDisplayName(), SkinAdapterSingleton::get()->toSkinData($p->getSkin()), $p->getXuid())]));
+		$this->sendDataPacket(PlayerListPacket::add([PlayerListEntry::createAdditionEntry($p->getUniqueId(), $p->getId(), $p->getPlayerListName(), SkinAdapterSingleton::get()->toSkinData($p->getSkin()), $p->getXuid())]));
 	}
 
 	public function onPlayerRemoved(Player $p) : void{
--- a/src/network/mcpe/NetworkSession.php
+++ b/src/network/mcpe/NetworkSession.php
@@ -978,7 +978,7 @@ class NetworkSession{
 	 */
 	public function syncPlayerList(array $players) : void{
 		$this->sendDataPacket(PlayerListPacket::add(array_map(function(Player $player) : PlayerListEntry{
-			return PlayerListEntry::createAdditionEntry($player->getUniqueId(), $player->getId(), $player->getDisplayName(), SkinAdapterSingleton::get()->toSkinData($player->getSkin()), $player->getXuid());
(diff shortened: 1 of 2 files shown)


ErkamKahriman:
i like that

lukeeey:
Hmm what about removing the entry and then adding it again like frago did

lukeeey:
Sorry about formatting but im about to go out i can fix later if needed and ive allowed edits from maintainers

Frago9876543210 (changes requested):
onPlayerAdded also should use getPlayerListName()

Wertzui123:
What do you mean with listname?

lukeeey:
@Frago9876543210 Dylan wasnt sure about that which is why i added it in the follow up

Frago9876543210:
@dktapps because getDisplayName used for chat what can cause conflict

dktapps:
That doesn't answer my question. Display name was used for player list for years without any complaints.

ErkamKahriman:
But if you setDisplayName then the name in playerlist will be like §4Owner §8» §f123456789123456789
and no one can ready that

dktapps:
Player list is an insignificant part of gameplay. The time you spend looking at it is minimal. It could use standard username and nobody would notice.

lukeeey:
Well an alternative reason for this is for either third party clients like alex that implement a tablist or if it gets implemented into the client itself.

Also a players display name may be something quite long, like
```
§b[Owner] §cAQuiteLongUsername
```
This means most of the players actual username may be cut out of the player list entry. 

I may want to just set the players player list name to the players name as the colour red so people can see i have the owner rank but without the rank name in there taking up space.


Also ive just thought that when implementing a nickname system, you generally setthe players display name. This would change it in the player list too, which may not he expected behaviour.

lukeeey:
And while yes its insignificant, that doesnt mean we shouldnt give users the choice to change it

SOF3 (approved):

SOF3:
> Player list is an insignificant part of gameplay. The time you spend looking at it is minimal. It could use standard username and nobody would notice.

I don't think this is a valid reason to reject this pull request. If there is nothing wrong, why can't we get it merged? If someone created this pull request, it's here for some reason.

The only issue I see with this pull request is that `getDisplayName` now has an ambiguous meaning, and that `setDisplayName()` now doesn't change the player list value. This is indeed an unresolved question and a BC break, but I don't see this mentioned above.

Suggestion: Reorient this pull request into "split display name into chat name and player list name", and deprecate/remove `getDisplayName` and `setDisplayName`.

dktapps:
> If there is nothing wrong, why can't we get it merged? 

because it adds extra complexity for a fringe feature that doesn't matter 99.9% of the time anyway.

PR #2979, opened 2019-06-22
Add Level->getTimeOfDay(), closes #2908

### LINKED ISSUE #2908: World: add API to get current time of day
## Description
This proposes adding some sugar like `getTimeOfDay()` to `World` (sorry, `Level`).

## Justification
Currently it's necessary to do `getTime() % Level::TIME_FULL` whenever you want the time of day and not the global time. This is easy to screw up and less easy to refactor when used frequently.

lukeeey:
## Introduction
Implements #2908 

### API changes
Added the method `getTimeOfDay()` to `Level`

### Behavioural changes
None.

## Backwards compatibility
Compatible.

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -936,8 +936,10 @@ class Level implements ChunkManager, Metadatable{
 		}
 
 		if($resetTime){
-			if($this->getTimeOfDay() >= Level::TIME_NIGHT and $this->getTimeOfDay() < Level::TIME_SUNRISE){
-				$this->setTime($this->getTime() + Level::TIME_FULL - $this->getTimeOfDay());
+			$time = $this->getTimeOfDay();
+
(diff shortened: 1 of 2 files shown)


src/pocketmine/level/Level.php
```diff
@@ -936,10 +936,8 @@ public function checkSleep(){
 		}
 
 		if($resetTime){
-			$time = $this->getTime() % Level::TIME_FULL;
-
-			if($time >= Level::TIME_NIGHT and $time < Level::TIME_SUNRISE){
-				$this->setTime($this->getTime() + Level::TIME_FULL - $time);
+			if($this->getTimeOfDay() >= Level::TIME_NIGHT and $this->getTimeOfDay() < Level::TIME_SUNRISE){
+				$this->setTime($this->getTime() + Level::TIME_FULL - $this->getTimeOfDay());
```
dktapps: it will be better to keep the change minimal here, replace the original `$time` variable assignment. It requires three less calls.

PR #2980, opened 2019-06-22
Implemented ChestPairEvent

### LINKED ISSUE #2829: Add a ChestPairEvent

T14Raptor:
## Introduction
Redo of #2833 

Adds a new event called ChestPairEvent, is cancellable.

Closes #2829

### API changes
Adds a new cancellable event called ChestPairEvent.

## Tests

ChestPairEvent Cancelled

### DIFF
--- a/src/pocketmine/block/Chest.php
+++ b/src/pocketmine/block/Chest.php
@@ -25,6 +25,7 @@ namespace pocketmine\block;
 
 use pocketmine\block\tile\Chest as TileChest;
 use pocketmine\block\utils\BlockDataValidator;
+use pocketmine\event\block\ChestPairEvent;
 use pocketmine\item\Item;
 use pocketmine\math\AxisAlignedBB;
 use pocketmine\math\Facing;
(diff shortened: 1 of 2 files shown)


lukeeey:
Could it be useful to get which chest initiated the event?

Thunder33345:
maybe more like "original" chest as in the chest that "was there"
and a "new pair" as in the newly placed one
maybe also get original state, like orientation before pairing

lukeeey:
@Thunder33345 thats what i meant

src/pocketmine/block/Chest.php
```diff
@@ -77,8 +78,26 @@ public function onPostPlace() : void{
...
+								$ev = new ChestPairEvent($this, $c);
+							}
+						}
+
+						$ev->call();
+						if(!$ev->isCancelled()){
+							$pair->pairWith($tile);
+							$tile->pairWith($pair);
+						}
+
```
dktapps: I think there is a simpler way to implement this logic. A chest to the clockwise of the chest's `facing` is on its left, and anticlockwise to its right.

dktapps:
This appears to be 2 months stale with no attempt to address comments.

PR #2989, opened 2019-07-01
Created math-related methods which return Vector3 instances from Position objects

TwistedAsylumMC:
Previously, Position::add() and several other methods would return a Vector3 instance as they are from the parent (Vector3) class

## Introduction

This pull request implements a method multiple methods to the Position class which were inherited from the parent Vector3 class, but returning an instance of Position instead.

### API changes

Created ``Position::add()``
Created ``Posisition::subtract()``
Created ``Posisition::multiply()``
Created ``Posisition::divide()``
Created ``Posisition::ceil()``
Created ``Posisition::floor()``
Created ``Posisition::round()``
Created ``Posisition::abs()``

### Behavioural changes

None :)

## Backwards compatibility

There is no backwards compatability changes, only people currently using Position::add() and the other methods will now receive a Position instance, but this will not affect anything.

## Tests

```php
$position = new Position(0, 0, 0, Server::getInstance()->getLevelByName("Test"));
$newPosition = $position->add(1, 1, 1);
var_dump($newPosition->__toString());
```
Output: ``string(32) "Position(level=Test,x=1,y=1,z=1)"``

### DIFF
--- a/src/pocketmine/level/Position.php
+++ b/src/pocketmine/level/Position.php
@@ -103,6 +103,47 @@ class Position extends Vector3{
 			return new Position($this->x + $x, $this->y + $y, $this->z + $z);
 		}
 	}
+	
+	/**
+	 * @param Position|int $x
+	 * @param int         $y
+	 * @param int         $z
+	 *
+	 * @return Vector3
+	 */
+	public function subtract($x = 0, $y = 0, $z = 0) : Position{
+		if($x instanceof Position){
+			return $this->add(-$x->x, -$x->y, -$x->z);
+		}else{
+			return $this->add(-$x, -$y, -$z);
+		}
+	}
+	
+	public function multiply(float $number) : Position{
+		return new Position($this->x * $number, $this->y * $number, $this->z * $number, $this->level);
+	}
+	
+	public function divide(float $number) : Position{
+		return new Position($this->x / $number, $this->y / $number, $this->z / $number, $this->level);
+	}
+	
+	public function ceil() : Position{
+		return new Position((int) ceil($this->x), (int) ceil($this->y), (int) ceil($this->z), $this->level);
+	}
+	
+	public function floor() : Position{
+		return new Position((int) floor($this->x), (int) floor($this->y), (int) floor($this->z), $this->level);
+	}
+	
+	public function round(int $precision = 0, int $mode = PHP_ROUND_HALF_UP) : Position{
+		return $precision > 0 ?
+			new Position(round($this->x, $precision, $mode), round($this->y, $precision, $mode), round($this->z, $precision, $mode), $this->level) :
+			new Position((int) round($this->x, $precision, $mode), (int) round($this->y, $precision, $mode), (int) round($this->z, $precision, $mode), $this->level);
+	}
+	
+	public function abs() : Position{
+		return new Position(abs($this->x), abs($this->y), abs($this->z), $this->level);
+	}
 
 	/**
 	 * Checks if this object has a valid reference to a loaded Level
--- a/src/pocketmine/level/Position.php
+++ b/src/pocketmine/level/Position.php
@@ -90,41 +90,33 @@ class Position extends Vector3{
 	}
 
 	/**
-	 * @param Position|int $x
+	 * @param Vector3|int $x
 	 * @param int         $y
 	 * @param int         $z
 	 *
-	 * @return Vector3
+	 * @return Position
 	 */
 	public function add($x, $y = 0, $z = 0) : Position{
-		if($x instanceof Position){
-			return new Position($this->x + $x->x, $this->y + $x->y, $this->z + $x->z);
+		if($x instanceof Vector3){
+			return new Position($this->x + $x->x, $this->y + $x->y, $this->z + $x->z, $this->level);
 		}else{
-			return new Position($this->x + $x, $this->y + $y, $this->z + $z);
+			return new Position($this->x + $x, $this->y + $y, $this->z + $z, $this->level);
 		}
 	}
 	
 	/**
-	 * @param Position|int $x
+	 * @param Vector3|int $x
 	 * @param int         $y
 	 * @param int         $z
 	 *
-	 * @return Vector3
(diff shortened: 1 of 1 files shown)


Sandertv:
It seems rather inconsistent to do this only for the `add()` method, whereas methods such as `subtract()` still return Vector3s.

TwistedAsylumMC:
Oh oops, will add that now. Forgot about that haha

src/pocketmine/level/Position.php
```diff
@@ -89,6 +89,62 @@ public function setLevel(Level $level = null){
...
+	 */
+	public function subtract($x = 0, $y = 0, $z = 0) : Position{
+		if($x instanceof Position){
+			return $this->add(-$x->x, -$x->y, -$x->z);
+		}else{
+			return $this->add(-$x, -$y, -$z);
+		}
+	}
+	
+	public function multiply(float $number) : Position{
```
SOF3: I don't see the point of multiplying a `Position`. A `Position`, by definition, is a positional vector. There is no point of scalar-multiplying a positional vector.

src/pocketmine/level/Position.php
```diff
@@ -89,6 +89,62 @@ public function setLevel(Level $level = null){
...
+		return new Position((int) floor($this->x), (int) floor($this->y), (int) floor($this->z), $this->level);
+	}
+	
+	public function round(int $precision = 0, int $mode = PHP_ROUND_HALF_UP) : Position{
+		return $precision > 0 ?
+			new Position(round($this->x, $precision, $mode), round($this->y, $precision, $mode), round($this->z, $precision, $mode), $this->level) :
+			new Position((int) round($this->x, $precision, $mode), (int) round($this->y, $precision, $mode), (int) round($this->z, $precision, $mode), $this->level);
+	}
+	
+	public function abs() : Position{
```
SOF3: Same as multiply &mdash; there is no reason to take the absolute value of a positional vector. The origin or the octant doesn't really mean anything in the positional sense.

src/pocketmine/level/Position.php
```diff
@@ -89,6 +89,62 @@ public function setLevel(Level $level = null){
...
 
+	/**
+	 * @param Position|int $x
+	 * @param int         $y
+	 * @param int         $z
+	 *
+	 * @return Vector3
+	 */
+	public function add($x, $y = 0, $z = 0) : Position{
+		if($x instanceof Position){
```
SOF3: A Position object is a positional vector by definition. There is no reason to add two positional vectors.

src/pocketmine/level/Position.php
```diff
@@ -89,6 +89,62 @@ public function setLevel(Level $level = null){
...
+			return new Position($this->x + $x->x, $this->y + $x->y, $this->z + $x->z);
+		}else{
+			return new Position($this->x + $x, $this->y + $y, $this->z + $z);
```
SOF3: Shouldn't the level field be carried over?

src/pocketmine/level/Position.php
```diff
@@ -89,6 +89,62 @@ public function setLevel(Level $level = null){
...
+	/**
+	 * @param Position|int $x
+	 * @param int         $y
+	 * @param int         $z
+	 *
+	 * @return Vector3
+	 */
+	public function subtract($x = 0, $y = 0, $z = 0) : Position{
+		if($x instanceof Position){
+			return $this->add(-$x->x, -$x->y, -$x->z);
```
SOF3: The difference between two positions is actually a relative vector. A relative vector should not be represented with the `Position` class.

src/pocketmine/level/Position.php
```diff
@@ -89,6 +89,62 @@ public function setLevel(Level $level = null){
...
+		}else{
+			return $this->add(-$x, -$y, -$z);
+		}
+	}
+	
+	public function multiply(float $number) : Position{
+		return new Position($this->x * $number, $this->y * $number, $this->z * $number, $this->level);
+	}
+	
+	public function divide(float $number) : Position{
```
SOF3: Same problem with multiply().

SOF3 (changes requested):
We have been looking to remove the inheritance relationship between Position and Vector3, preferring composition over inheritance. This pull request is actually one reason why it is necessary to separate &mdash; inheritance is not an appropriate approach between these two classes.
Merging this pull request implies that every future method in Vector3 has to be overridden by an equivalent method in Position, otherwise leading to inconsistency. This does not seem to be a sustainable approach.
In addition, manipulating vectors is usually something mathematical, and should not be mixed up with the semantic role of the Position class.

src/pocketmine/level/Position.php
```diff
@@ -89,6 +89,62 @@ public function setLevel(Level $level = null){
 		return $this;
 	}
 
+	/**
+	 * @param Position|int $x
```
SOF3: This is backwards-incompatible, disallowing passing a `Vector3` to `Position->add()`.
This also violates covariance, because this method is expected to work but conditionally doesn't:
```php
function foo(Vector3 $a) : Vector3 {
    return $a->add(new Vector3(1, 2, 3));
}
```
Calling `foo` with a `Position` object leads to a crash (trying to add `$this->x` to a `Vector` object), while `foo` did not even need to depend on pmmp.

SOF3:
> ## Tests
> 
> ```
> $position = new Position(0, 0, 0, Server::getInstance()->getLevelByName("Test"));
> $newPosition = $position->add(1, 1, 1);
> var_dump($newPosition->__toString());
> ```
> 
> Output: `string(32) "Position(level=Test,x=1,y=1,z=1)"`

Is this a real test, or did you just made this up? I don't see a level being passed to `$newPosition` in any way.

Eren5960:
`Position::fromObject($position->add(1, 1, 1), $level);` is not it enough?

dktapps:
This has been discussed on and off for years, and we've always decided against it because of the amount of extra complexity it creates for any descendent class. The benefit is not worth the extra mess it creates. `Position` should never have extended `Vector3` to begin with.

PR #2990, opened 2019-07-02
Fix Entity::isNameTagAlwaysVisible() was returning the wrong value

Encritary:
## Introduction

While Entity::setNameTagAlwaysVisible() was setting the byte data value, Entity::isNameTagAlwaysVisible() was still returning the unused flag value.
## Tests

Before this commit:
```php
$entity->setNameTagAlwaysVisible(true);
var_dump($entity->isNameTagAlwaysVisible()); //bool(false)
```
After:
```php
$entity->setNameTagAlwaysVisible(true);
var_dump($entity->isNameTagAlwaysVisible()); //bool(true)
```

### DIFF
--- a/src/pocketmine/entity/Entity.php
+++ b/src/pocketmine/entity/Entity.php
@@ -662,7 +662,7 @@ abstract class Entity extends Location implements Metadatable, EntityIds{
 	 * @return bool
 	 */
 	public function isNameTagAlwaysVisible() : bool{
-		return $this->getGenericFlag(self::DATA_FLAG_ALWAYS_SHOW_NAMETAG);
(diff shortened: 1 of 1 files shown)


src/pocketmine/entity/Entity.php
```diff
@@ -662,7 +662,7 @@ public function isNameTagVisible() : bool{
 	 * @return bool
 	 */
 	public function isNameTagAlwaysVisible() : bool{
-		return $this->getGenericFlag(self::DATA_FLAG_ALWAYS_SHOW_NAMETAG);
+		return $this->propertyManager->getByte(self::DATA_ALWAYS_SHOW_NAMETAG) > 0;
```
dktapps: There's more than 1 mode. I would be more comfortable with a `=== 1` check.

Issue #1760, opened 2017-12-02
BlockPlaceEvent block doesn't have the meta value it would have when placed


dktapps:
### Issue description
In effect, `BlockPlaceEvent->getBlock()` might as well simply alias to `BlockPlaceEvent->getItem()->getBlock()` since the resulting set block is unknown before `Block->place()` is called.

This requires a separation of logic of the block meta determination code from the actual setting the block into the world.

### OS and versions

* PocketMine-MP: alpha9
* PHP: irrelevant
* Server OS: irrelevant
* Game version: irrelevant

### Plugins
This is an API issue.

dktapps:
This is planned to be fixed for 4.0.

dktapps:
This can be addressed by using the resulting `BlockTransaction` produced by `Block->place()`. Also, it might be a good idea to provide the transaction to `BlockPlaceEvent` directly so that plugins can properly understand stuff like bed placement.

PR #2998, opened 2019-07-08
Fixed BlockPlaceEvent firing when the block can't be placed.

matcracker:
## Introduction

When you try to place a item/block in a block where it can't be placed (e.g. flowers on sand), the event BlockPlaceEvent is always fired.

### Relevant issues

* Fixes BlockPlaceEvent always fired when the block can't be placed.
### API changes

* Added BlockTransaction::validate() - Must be used before BlockTransaction::apply() else it throws an UnexpectedValueException.

### Behavioural changes

No.
## Backwards compatibility

Yes.
## Tests

Before:
```php
public function onPlace(BlockPlaceEvent $event): void{
    var_dump($event->getBlock()->getName()); //Always prints the block name.
}
```
After:
```php
public function onPlace(BlockPlaceEvent $event): void{
    var_dump($event->getBlock()->getName()); //It prints only when the block can be placed.
}
```

### DIFF
--- a/src/pocketmine/world/World.php
+++ b/src/pocketmine/world/World.php
@@ -1754,7 +1754,11 @@ class World implements ChunkManager{
 			}
 		}
 
-
+		$tx = new BlockTransaction($this);
+		if(!$hand->place($tx, $item, $blockReplace, $blockClicked, $face, $clickVector, $player) or !$tx->apply()){
+			return false;
+		}
+		
 		if($player !== null){
 			$ev = new BlockPlaceEvent($player, $hand, $blockReplace, $blockClicked, $item);
 			if($player->isAdventure(true) and !$ev->isCancelled()){
@@ -1780,11 +1784,7 @@ class World implements ChunkManager{
 				return false;
 			}
 		}
-
-		$tx = new BlockTransaction($this);
-		if(!$hand->place($tx, $item, $blockReplace, $blockClicked, $face, $clickVector, $player) or !$tx->apply()){
-			return false;
-		}
+		
 		foreach($tx->getBlocks() as [$x, $y, $z, $_]){
 			$tile = $this->getTileAt($x, $y, $z);
 			if($tile !== null){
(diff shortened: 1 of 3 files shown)


src/pocketmine/world/World.php
```diff
@@ -1754,7 +1754,11 @@ public function useItemOn(Vector3 $vector, Item &$item, int $face, ?Vector3 $cli
 			}
 		}
 
-
+		$tx = new BlockTransaction($this);
+		if(!$hand->place($tx, $item, $blockReplace, $blockClicked, $face, $clickVector, $player) or !$tx->apply()){
```
dktapps: This will write the blocks to the world before the event is called.

I did mention previously that the return of this function is not intended to signify anything with regards to events.

matcracker: You're right. Check last commit, I fix it.

src/pocketmine/world/generator/object/Tree.php
```diff
@@ -112,7 +112,9 @@ public function placeObject(ChunkManager $world, int $x, int $y, int $z, Random
 		$this->placeTrunk($x, $y, $z, $random, $this->generateChunkHeight($random), $transaction);
 		$this->placeCanopy($x, $y, $z, $random, $transaction);
 
-		$transaction->apply(); //TODO: handle return value on failure
+		if($transaction->validate()){
+			$transaction->apply(); //TODO: handle return value on failure
+		}
```
dktapps: This boilerplate is now going to be needed everywhere a transaction has been used. For one specific use-case, this is unacceptable.

dktapps:
Part of the objective with using a transaction in this code was to have the transaction injected into `BlockPlaceEvent`, for the following reasons:
- fixing #1760 
- make plugins aware that multiple blocks may be created, which is something `BlockPlaceEvent` doesn't give any indication of
- allow plugins to alter the result of placement and possibly make invalid transactions become valid

As such, this PR doesn't make a whole lot of sense because it increases the code complexity and would be replaced anyway.

matcracker:
I can't find any other solution, goodbye. lol

PR #3007, opened 2019-07-11
MetadataProperties: id, data => static runtime id

Frago9876543210:


Enes5519 (approved):

src/pocketmine/network/mcpe/protocol/types/EntityMetadataProperties.php
```diff
@@ -49,7 +49,7 @@ private function __construct(){
 	public const PADDLE_TIME_LEFT = 13; //float
 	public const PADDLE_TIME_RIGHT = 14; //float
 	public const EXPERIENCE_VALUE = 15; //int (xp orb)
-	public const MINECART_DISPLAY_BLOCK = 16; //int (id | (data << 16))
+	public const MINECART_DISPLAY_BLOCK = 16; //int (static runtime id)
 	public const HORSE_FLAGS = 16; //int
```
dktapps: this is not a static ID

dktapps (changes requested):

dktapps:
I don't want to merge this on master because it's going to get overwritten the next time I have to integrate changes from stable. You should make your change there instead.

PR #3042, opened 2019-07-18
Avoid useless loop into plugin description

matcracker:
## Introduction

The $plugin["authors"] already returns an array if is set, so it's useless iterate it.
### Relevant issues
Any issue. Just a micro optimization.

### API changes

No
### Behavioural changes

No
## Backwards compatibility

Yes

### DIFF
--- a/src/pocketmine/plugin/PluginDescription.php
+++ b/src/pocketmine/plugin/PluginDescription.php
@@ -140,7 +140,7 @@ class PluginDescription{
 			$this->authors[] = $plugin["author"];
 		}
 		if(isset($plugin["authors"])){
-			$this->authors = (array) $plugin["authors"];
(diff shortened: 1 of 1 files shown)


dktapps (changes requested):
This will overwrite the `author` directive if both are used.

matcracker:
> This will overwrite the `author` directive if both are used.

Wooops, I accidentaly deleted the brackets.

dktapps:
The latest commit makes the problem worse. Now it's appending an array to an array.

lukeeey (commented):
Shouldve probably used `array_merge`

matcracker:
Yes, I did that but Dylan closed the PR before my commit. He's so cruel :(

PR #3047, opened 2019-07-19
Add test for item's display properties' persistence after deserialization

Muqsit:
## Introduction

Items lose their lores during serialization/deserialization.

## Tests

Results of the following tests at commit b5b4133c5d0763284fcd9a0e432b2eb26c4f7114
```php
bool(true) // self::assertTrue($item->getCustomName() === $name);
bool(false) // self::assertTrue($item->getLore() === $lore);
```

### DIFF
--- a/tests/phpunit/item/ItemTest.php
+++ b/tests/phpunit/item/ItemTest.php
@@ -62,6 +62,21 @@ class ItemTest extends TestCase{
 		self::assertTrue($item1->equals($item2));
 	}
 
+	/**
+	 * Tests whether items retain their display properties
+	 * after being deserialized
+	 */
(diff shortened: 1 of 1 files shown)


dktapps:
have you made any progress on pinpointing the bug yet?

Muqsit:
Found it.
https://github.com/pmmp/PocketMine-MP/blob/b5b4133c5d0763284fcd9a0e432b2eb26c4f7114/src/pocketmine/item/Item.php#L286
```php
$lore = $tag->getListTag(self::TAG_DISPLAY_LORE);
```
is supposed to be
```php
$lore = $display->getListTag(self::TAG_DISPLAY_LORE);
```

PR #3049, opened 2019-07-19
Item: fix for setCustomName

Eren5960:


Wumpotamus:
What exactly does this fix? You’ve given 0 detail on what issue is there and what it fixes.

PR #3054, opened 2019-07-23
FlowerPot: added `PlantBlock` support

Frago9876543210:
## Introduction
This PR adds support of `PlantBlock` tag which used in bedrock worlds


dktapps:
This should be implemented in a more general way. This logic isn't only needed for flower pots.

Frago9876543210:
For what else it needed?

benda95280:
Any news ?

PR #3058, opened 2019-07-26
Load Doxygen Mathjax from https to prevent unsafe scripts error

HimbeersaftLP:
## Introduction

Most browsers will block loading scripts over the http protocol when the site that loads it is served using https.
A warning like this will be displayed and the script will not load:
image

## Changes
I added a letter

## Tests
https://cdn.mathjax.org/mathjax/latest/MathJax.js does work


lukeeey:
thats a shit browser then

PR #3064, opened 2019-07-31
Scan load order

### LINKED ISSUE #2825: Disallow non-startup plugin loading

SOF3:
### Relevant issues
* Resolves #2825

### API changes
- Added a `SCAN` load order
- `$newLoaders` parameter in `loadPlugins` is automatically detected based on the time it is called

### Behavioural changes
- `loadPlugins` is automatically called again after `SCAN` order plugins are enabled
- Plugins can no longer be loaded after all the second call to `loadPlugins`.

## Backwards compatibility
Plugins that load plugins need to be updated. Refer to the Changes section above.

## Tests
Not yet tested. Needs a modified version of DevTools to test.

### DIFF
--- a/src/plugin/PluginManager.php
+++ b/src/plugin/PluginManager.php
@@ -430,6 +430,8 @@ class PluginManager{
 	public function enablePlugins(PluginLoadOrder $type) : void{
 		if($this->scanLoaded){
 			$this->scanEnabled = true;
+		}elseif($type->equals(PluginLoadOrder::SCAN())){
+			$this->scanLoaded = true;
 		}
 
 		foreach($this->plugins as $plugin){
--- a/src/plugin/PluginManager.php
+++ b/src/plugin/PluginManager.php
@@ -78,6 +78,9 @@ class PluginManager{
 	 */
 	protected $enabledPlugins = [];
 
+	protected $scanLoaded = false;
+	protected $scanEnabled = false;
+
 	/**
 	 * @var PluginLoader[]
 	 */
@@ -148,6 +151,10 @@ class PluginManager{
 	 * @return Plugin|null
 	 */
 	public function loadPlugin(string $path, ?array $loaders = null) : ?Plugin{
+		if($this->scanEnabled){
+			throw new \InvalidStateException("Cannot load plugins after SCAN-order plugins are enabled");
+		}
+
 		foreach($loaders ?? $this->fileAssociations as $loader){
 			if($loader->canLoadPlugin($path)){
 				$description = $loader->getPluginDescription($path);
@@ -203,6 +210,10 @@ class PluginManager{
 	 * @return Plugin[]
 	 */
 	public function loadPlugins(string $directory) : array{
+		if($this->scanEnabled){
+			throw new \InvalidStateException("Cannot load plugins after SCAN-order plugins are enabled");
+		}
+
 		if(!is_dir($directory)){
 			return [];
 		}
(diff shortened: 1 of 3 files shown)


SOF3:
DevTools update has been pushed to the `pmmp/PocketMine-DevTools:scan-load-order` branch. Tests are passing, and plugins are loading correctly. Ready for review.

src/plugin/PluginManager.php
```diff
@@ -429,9 +433,26 @@ public function isPluginEnabled(Plugin $plugin) : bool{
...
+		foreach($this->plugins as $plugin){
+			if(!$plugin->isEnabled() and $plugin->getDescription()->getOrder()->equals($type)){
+				$this->enablePlugin($plugin);
+			}
+		}
+
+		if($type->equals(PluginLoadOrder::POSTWORLD())){
+			$this->server->getCommandMap()->registerServerAliases();
+			DefaultPermissions::registerCorePermissions();
+		}
```
dktapps: I think this block should be located in the Server startup routine. I have no idea why it was here to begin with, and it certainly doesn't make sense to reinitialize all this crap if something decides to add an extra plugin.

Perhaps this change should be done separately, but I'm logging it here for posterity since you've touched the code.

src/plugin/PluginManager.php
```diff
@@ -77,6 +78,9 @@ class PluginManager{
 	 */
 	protected $enabledPlugins = [];
 
+	protected $scanLoaded = false;
+	protected $scanEnabled = false;
```
dktapps: please add `@var` annotations to these fields

dktapps: The name `scanEnabled` is ambiguous. I initially took it to mean "scanning is enabled", but it seems to actually mean "scan load-order plugins have been enabled".

src/plugin/PluginManager.php
```diff
@@ -77,6 +78,9 @@ class PluginManager{
 	 */
 	protected $enabledPlugins = [];
 
+	protected $scanLoaded = false;
```
dktapps: This field is never written to.

dktapps (changes requested):

src/plugin/PluginManager.php
SOF3: `@var` annotations for types or for actual documentation? It's a private field (it's protected, but effectively private).

What other name would you expect?

dktapps: For the types. All fields should have type annotations, regardless of visibility.

dktapps:
This has since been obsoleted by #3971.

PR #3073, opened 2019-08-02
remove name change case when looking for an offline player

jasonw4331:
## Introduction

OfflinePlayer instances always return a lowercase name rather than the (possibly)mixed case argument.

### Relevant issues

N/A

### API changes

N/A
### Behavioural changes

OfflinePlayer instances returned by this function can now have mixed case names.
## Backwards compatibility

Is this considered backwards incompatible?
## Follow-up
N/A

## Tests
```php
$name = "Steve101";
echo $name;
$offlinePlayer = $pluginBase->getServer()->getOfflinePlayer($name);
echo $offlinePlayer->getName();
```
Names should match.

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -780,7 +780,6 @@ class Server{
 	 * @return OfflinePlayer|Player
 	 */
 	public function getOfflinePlayer(string $name){
-		$name = strtolower($name);
 		$result = $this->getPlayerExact($name);
 
 		if($result === null){


xISRAPILx:
Player names are kept in lowercase, I think you've broken everything.

dktapps:
by a lucky coincidence, no, but I'm not happy with this change because we have no way of knowing what the true case of an offline player's name is. It's better to consistently use lower case than whatever bullshit the user decides to put in.

PR #3097, opened 2019-08-14
Scoreboard added

AlexBrin:
## Introduction
Scoreboard added

### API changes

* `Server->getScoreboardManager`
* `ScoreboardManager->create(Player $player, ?string $title = null)`
* `ScoreboardManager->destroy(Player $player)`
* `ScoreboardManager->setScoreboardClass(string $newScoreboard)`
* `Scoreboard->create(string $title, bool $oldRemove = false)`
* `Scoreboard->destroy()`
* `Scoreboard->setLines(string[] $lines)`
* `Player->addScoreboard(string title, array $lines = [])`
* `Player->removeScoreboard()`

### Send scoreboard:
```php
$player->sendScoreboard("Awesome scoreboard", [
  "My scoreboard",
  "Veeery long scoreboard line",
]);
```

### Custom scoreboard:
```php
class MyScoreboard extends \pocketmine\scoreboard\Scoreboard {
// todo
}
```
And install a new instance:
```php
$server->getScoreboardManager()->setScoreboardClass(MyScoreboard::class);
```

### DIFF
--- a/src/pocketmine/scoreboard/ScoreboardManager.php
+++ b/src/pocketmine/scoreboard/ScoreboardManager.php
@@ -1,0 +1,93 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * @author PocketMine Team
+ * @link http://www.pocketmine.net/
+ *
+ *
+*/
+
+namespace pocketmine\scoreboard;
+
+use pocketmine\Player;
+
+class ScoreboardManager
+{
+	/** @var Scoreboard[] */
+	private $scoreboards = [];
+
+	/**
+	 * @var string
+	 */
+	private $baseClass = Scoreboard::class;
+
+	/**
+	 * @var
+	 */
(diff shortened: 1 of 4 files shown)


Encritary:
There's already a scoreboard PR and, don't get me wrongly, IMHO it's much better.

src/pocketmine/scoreboard/ScoreboardManager.php
```diff
@@ -0,0 +1,93 @@
...
+	 * @return string
+	 */
+	public function getScoreboardClass(): string {
+		return $this->currentClass;
+	}
+
+	/**
+	 * @param $class
+	 */
+	public function setScoreboardClass($class) {
```
SOF3: I hope you aren't trying to make another PlayerCreationEvent...

SOF3:
This looks more like a statically-linked library than an API. I don't see any evidence of attempts to maintain plugin compatibility.

src/pocketmine/scoreboard/ScoreboardManager.php
AlexBrin: Of course not. May be necessary to use your own scoreboard.
E.g. to change the line numbering.

SOF3: ScoreboardManager is a singleton. This is generating the same problem as PlayerCreationEvent.

AlexBrin:
@SOF3, what compatibility are we talking about? 
Yes, it statically-linked, but only to avoid problems when working with the scoreboard. E.g., from different plugins. But it is, nevertheless, API.

src/pocketmine/scoreboard/ScoreboardManager.php
```diff
@@ -0,0 +1,93 @@
...
+	 * @param Player $player
+	 * @param string|null $title
+	 * @return Scoreboard
+	 */
+	public function create(Player $player, ?string $title = null): Scoreboard {
+		if(isset($this->scoreboards[$player->getName()])) {
+			$scoreboard = $this->scoreboards[$player->getName()];
+		} else {
+			$scoreboard = new $this->currentClass($player);
+			$this->scoreboards[$player->getName()] =& $scoreboard;
```
dktapps: Objects are always assigned by reference. This reference-to-reference makes zero sense.

src/pocketmine/scoreboard/ScoreboardManager.php
dktapps: This entire class is pointless. It's just a reduction of flexibility.

PR #3100, opened 2019-08-17
added missing setText method for Sign block

Frago9876543210:
## Introduction
Now there is no way to change sign text without player. Well, you can only use block entity. However, this is marked as deprecated

### DIFF
--- a/src/block/Sign.php
+++ b/src/block/Sign.php
@@ -143,6 +143,14 @@ class Sign extends Transparent{
 	}
 
 	/**
+	 * @param SignText $text
+	 */
+	public function setText(SignText $text) : void{
+		$this->text = $text;
+		$this->pos->getWorld()->setBlock($this->pos, $this);
+	}
+
+	/**
 	 * Called by the player controller (network session) to update the sign text, firing events as appropriate.
 	 *
 	 * @param Player   $author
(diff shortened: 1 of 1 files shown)


dktapps:
SignText is mutable, this is redundant.

Frago9876543210:
Since e61a08a56bb71c77c49cb070feeefa83e86397a2 SignText is immutable and this PR make sense

src/block/Sign.php
```diff
@@ -142,6 +142,14 @@ public function getText() : SignText{
 		return $this->text;
 	}
 
+	/**
+	 * @param SignText $text
+	 */
+	public function setText(SignText $text) : void{
+		$this->text = $text;
+		$this->pos->getWorld()->setBlock($this->pos, $this);
```
dktapps, line 147: please remove this, it's possible that we may want to set the text of a block which doesn't have a position.

src/block/Sign.php
```diff
@@ -165,8 +173,7 @@ public function updateText(Player $author, SignText $text) : bool{
 		}, $text->getLines())));
 		$ev->call();
 		if(!$ev->isCancelled()){
-			$this->text = clone $ev->getNewText();
-			$this->pos->getWorld()->setBlock($this->pos, $this);
```
dktapps, line 169: please restore the setBlock() call here

PR #3115, opened 2019-09-05
Add doc for unknown field of CommandParameter

Frago9876543210:
## Introduction
if false:
изображение
if true:
изображение


dktapps (changes requested):
The type should be a byte, not a boolean. It's more likely a set of flags rather than a single on/off.

Frago9876543210:
Did you sure about it? 
> unknown, always zero except for in /gamerule command

dktapps:
Yes I'm sure. It wouldn't be written as a byte if I wasn't.

PR #3117, opened 2019-09-08
YOU CAN SPRINT AND DEAL CRITS NOW

Vrysed:
BEFORE YOU CANT DEAL CRITS WHEN SPRINTING NOW YOU CAN

## Introduction
In PocketMine you cannot deal crits when sprinting, but in vanilla you can.


DaPigGuy:
... Are you stupid?

PR #3118, opened 2019-09-08
InstantEnchantParticle: add Color into constructor

Frago9876543210:
## Introduction
These particles are similar to DustParticle and also have color


Erkam246:
add nullable default 0

dktapps:
could you provide some screenshots?

Frago9876543210:
i

Frago9876543210:
Initially, these particles were black, so it was suggested that they work by analogy with dust

towuk118:
как на линухе майн запустил?)

dktapps:
Those don't look like enchanting particles to me, more like coloured critical particles.

PR #3123, opened 2019-09-20
Improve bucket API

Frago9876543210:
## Introduction
Added method for obtaining kind of liquid

### API changes
Added `LiquidBucket::getLiquid()` method

### DIFF
--- a/src/event/player/PlayerBucketEvent.php
+++ b/src/event/player/PlayerBucketEvent.php
@@ -26,9 +26,7 @@ namespace pocketmine\event\player;
 use pocketmine\block\Block;
 use pocketmine\event\Cancellable;
 use pocketmine\event\CancellableTrait;
-use pocketmine\item\Bucket;
 use pocketmine\item\Item;
-use pocketmine\item\LiquidBucket;
 use pocketmine\player\Player;
 
 /**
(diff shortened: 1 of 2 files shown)


src/event/player/PlayerBucketEvent.php
```diff
@@ -62,7 +64,7 @@ public function __construct(Player $who, Block $blockClicked, int $blockFace, It
 	/**
 	 * Returns the bucket used in this event
 	 *
-	 * @return Item
+	 * @return LiquidBucket|Bucket|Item
```
dktapps: This is orthogonal to adding liquid type API, and also unnecessary since `Item` includes these types.

Frago9876543210: Yes, but how IDE understand that Item has getLiquid method?

dktapps: `instanceof`, same as anything.

Frago9876543210: Well, method called "getBucket" and should already it, not any item

dktapps (changes requested):
There are two unrelated changes in this PR. Please remove one of them.

PR #3145, opened 2019-10-11
Added encode/decode for some packets

yoraze:
## Introduction
Encode / decode methods have been added for two new packets that may be used in the near future.

### API changes
Without API changes.

### Behavioural changes
Without behavioural changes

## Tests
I will write when I can test it.(I will try to do it very soon)
At the moment, I am busy with a bit of my own affairs, not related to this PR.


dktapps:
readers: anyone with protocol info who would like to give this a peer review? I can't at the moment.

lukeeey (commented):
I wish we had a library like lombok for php, this class would be much better with setters and getters but it would be a lot more code

yoraze:
In my opinion, this is not necessary, since StructureTemplateDataExportRequestPacket is ServerboundPacket.

dktapps:
It makes no difference whether it is serverbound or not. Regardless, in the light of the migrations underway on 4.0, it would be preferred to have the fields private, with public getters and a constructor (or ::create() method).

Sandertv (approved):

PR #3148, opened 2019-10-15
StructureBlockUpdatePacket: Added encode/decode

yoraze:
# Introduction

Added encode decode for StructureBlockUpdatePacket, which can be used for StructureBlockActor!

The player sends it to the server when changing the settings of StructureBlockActor.

### API changes

Without API changes

### Behavioural changes

Without behavioural changes

## Tests

I don’t think that he is really needed here. However, I will try later to add StructureBlockActor and using the packets I will test it myself.


src/pocketmine/network/mcpe/protocol/StructureBlockUpdatePacket.php
```diff
@@ -26,16 +26,32 @@
...
+	/** @var int */
+	public $structureBlockX;
+	/** @var int */
+	public $structureBlockY;
+	/** @var int */
+	public $structureBlockZ;
+	/** @var StructureEditorData */
+	public $structureEditorData;
+	/** @var bool */
+	public $isStructureBlockPowered;
```
dktapps: These names don't need to be so verbose. It's already implied that they target a `structure block` because of the packet name.

dktapps:
It would be nice if someone could peer-review this and confirm the changes.

src/pocketmine/network/mcpe/NetworkBinaryStream.php
```diff
@@ -629,4 +630,30 @@ protected function putStructureSettings(StructureSettings $structureSettings) :
...
 		$this->putInt($structureSettings->integritySeed);
 	}
+
+	protected function getStructureEditorData() : StructureEditorData{
+		$result = new StructureEditorData();
+
+		$result->structureName = $this->getString();
+		$result->structureDataField = $this->getString();
+
+		$result->includePlayers = $this->getBool();
```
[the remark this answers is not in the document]
yoraze: I hope the question is settled?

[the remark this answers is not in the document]
yoraze: Also this name can be seen in the namedtag of this block.

(If you have a PC and IDA - see StructureBlockActor::save(StructureBlockActor *this, CompoundTag *a2))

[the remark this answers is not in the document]
dktapps: That doesn't mean that's actually what it does. It wouldn't be the first time something was completely misleadingly named by Microjang.

yoraze:
> It would be nice if someone could peer-review this and confirm the changes.

I have no acquaintances who could confirm this. Usually I check that others write about the protocol, and for this reason they often started contacting me for advice. :^/

yoraze:
I don’t know if you are interested or not, but I found similar information about this packet from Sandertv.
(https://github.com/Sandertv/gophertunnel/blob/master/minecraft/protocol/packet/structure_block_update.go)

Maybe ask him to confirm or deny this PR?
I just think you know each other and he could test this PR.

Sandertv (approved):
I've had a look over it and it looks good. We had a talk about it on Discord.

src/pocketmine/network/mcpe/NetworkBinaryStream.php
```diff
@@ -629,4 +630,30 @@ protected function putStructureSettings(StructureSettings $structureSettings) :
...
+		$result->showBoundingBox = $this->getBool();
+
+		$result->structureBlockType = $this->getVarInt();
```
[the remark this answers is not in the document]
thebigsmileXD: https://github.com/thebigsmileXD/libstructure/blob/master/src/xenialdan/libstructure/packet/StructureEditorData.php
There are 5 types. https://github.com/thebigsmileXD/libstructure/blob/master/src/xenialdan/libstructure/tile/StructureBlockTags.php#L11-L16

src/pocketmine/network/mcpe/protocol/StructureBlockUpdatePacket.php
```diff
@@ -26,16 +26,32 @@
 #include <rules/DataPacket.h>
 
 use pocketmine\network\mcpe\NetworkSession;
+use pocketmine\network\mcpe\protocol\types\StructureEditorData;
 
-class StructureBlockUpdatePacket extends DataPacket{
+class StructureBlockUpdatePacket extends DataPacket/* implements ServerboundPacket*/{
```
thebigsmileXD: This should not be here

src/pocketmine/network/mcpe/protocol/StructureBlockUpdatePacket.php
```diff
@@ -26,16 +26,32 @@
...
+	/** @var StructureEditorData */
+	public $structureEditorData;
+	/** @var bool */
+	public $isPowered;
+
 	protected function decodePayload(){
-		//TODO
+		$this->getBlockPosition($this->x, $this->y, $this->z);
+		$this->structureEditorData = $this->getStructureEditorData();
+		$this->isPowered = $this->getBool();
```
thebigsmileXD: There is another bool here. i haven't been able to figure out what it is used for though

src/pocketmine/network/mcpe/protocol/StructureBlockUpdatePacket.php
yoraze: I believe that it has a place to be.
It is a "little help" for merging into the master branch.

dktapps: It's been added to all the new stuff since that interface was added on the master branch. It saves having to dig into the binary twice when we implement a bunch of new stuff.

src/pocketmine/network/mcpe/protocol/StructureBlockUpdatePacket.php
yoraze: You're wrong.

Yes, he is, but starting with protocol 370.

dktapps:
This should have been targeted at `stable`.
We really want to keep the protocol as consistent as possible between the branches, and since PM is a reference implementation, it's confusing to people if our implementation isn't available on the default branch.

This also needs to be rechecked against Minecraft 1.14 to ensure that it's still in line with latest changes.

yoraze:
> This also needs to be rechecked against Minecraft 1.14 to ensure that it's still in line with latest changes.

Okay, I rechecked the read/write methods of StructureBlockUpdatePacket in BDS 1.14.30.2 and made sure that nothing has changed there. I also rechecked serialize/unserialize methods of StructureEditorData and StructureSettings - there are no changes either.

src/pocketmine/network/mcpe/protocol/StructureBlockUpdatePacket.php
thebigsmileXD: The bool is still missing here. If not included, you have unread bytes errors

thebigsmileXD: The bool is still missing here. If not included, you have unread bytes errors

inxomnyaa (changes requested):

PR #3160, opened 2019-10-23
Plugin::getLogger() interface no longer depends on PluginLogger

JackNoordhuis:
## Introduction
This change serves to allow the `pocketmine\plugin\Plugin` interface to be implemented more easily by only depending on the underlying interface it currently relies on –– the `pocketmine\plugin\PluginLogger` class is only an implementation of `\AttachableLogger` and adds no extra functionality.

### API changes
`pocketmine\plugin\Plugin::getLogger()` now enforces a return type of `\AttachableLogger` instead of `pocketmine\plugin\PluginLogger`.

### Behavioural changes
None – the type hint has been widened so all existing behavioural should continue to work.

## Backwards compatibility
All classes implementing `pocketmine\plugin\Plugin` will need to change the return type of `getLogger()` from `pocketmine\plugin\PluginLogger` to `\AttachableLogger` (see changes to `pocketmine\plugin\PluginBase`).

### DIFF
--- a/src/plugin/Plugin.php
+++ b/src/plugin/Plugin.php
@@ -81,9 +81,9 @@ interface Plugin{
 	public function getName() : string;
 
 	/**
-	 * @return PluginLogger
+	 * @return \AttachableLogger
 	 */
-	public function getLogger() : PluginLogger;
+	public function getLogger() : \AttachableLogger;
 
 	/**
 	 * @return PluginLoader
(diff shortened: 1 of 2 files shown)


dktapps:
The `PluginBase` implementation needs updating.

dktapps:
This is BC-breaking because return types are invariant.

JackNoordhuis:
Updated PR description (lack of sleep really impedes your ability to think) and committed `pocketmine\plugin\PluginBase` changes,

Issue #2730, opened 2019-02-02
Synchronized building doesn't work correctly


dktapps:
### Issue description
In recent versions of Bedrock the devs added a really nice (really!) feature which I refer to as "synchronized building". This makes block placement (and breaking in creative) speed synchronized with walking speed.

Sadly, this feature has some issues in PocketMine-MP due to how the movement processing system currently works. When the player is walking backwards placing blocks, sometimes the block will not be placed because the player's current position is inside the block. On the client side, they are _not_ inside the block, and the bug is happening because the server only processes movements once per tick, but processes block place/break immediately.

### Steps to reproduce the issue
1. Walk backwards placing blocks.
2. Notice that if you get too close to the target block the blocks you're placing will glitch and disappear.

### OS and versions

* PocketMine-MP: .3.5.9
* PHP: 7.2.14
* Game version: Win10

dktapps:
related to #1215

dktapps:
https://github.com/pmmp/PocketMine-MP/commit/485f573955fdfda52395e15fc497066d6ef77040 improves on this but doesn't completely fix it. It'll be necessary to handle position updates from the transaction packet itself to completely resolve this.

dktapps:
For posterity: This was caused by a bug in the handling for InventoryTransactionPacket, in a hack designed to reduce the overhead caused by spammy right-clicks. It did not account for player movements (or aiming at the same position on a different block). The movement changes did not have any effect on the problem (placebo).

PR #3167, opened 2019-10-27
Player: remove move buffering, implement simple rate limited movement…

### LINKED ISSUE #1215: Movement buffering system causes various bugs on net lag spikes

dktapps:
… processing

this needs some work to fix move sync spam in some areas, but this is nothing that wasn't already a problem to begin with.

## Introduction
This PR is a second attempt at improving movement processing to fix #1215 , #2730 and more.

This is significantly less complex than the previous attempt #2646 -- it gets rid of the movement buffering system entirely and instead relies on a simple rate limit counter to restrict on-the-fly movement processing.

Movement is rate limited to a max average of 2 per tick. It allows up to 5 seconds' backlog to accommodate network lag. The rate limit counter is increased by 2 per tick and decreased once for every movement processed. This prevents movement processing being abused for denial of service attacks.

### API changes
This PR, while obviously highly beneficial for most current users, poses some BC-breaking issues because of changes to the internal Player API. 
- `Player->processMovement()` (protected) has been removed. This is a BC concern for custom player classes which overrode it and called it as a parent. In addition, child implementations won't be called every tick any more, which could break some custom movement processing systems.
- `Player->newPosition` (protected) has been removed. This internal field may have been accessed by custom movement implementations.
- `Player->isTeleporting` (protected) has been removed. BC concern is same as previous point.
- `Player->getNextPosition()` (public) has been `@deprecated`.
- Added the following protected `Player` class members:
  - `int $moveRateLimit`
  - `?Vector3 $forceMoveSync`
  - `handleMovement()`
  - `processMostRecentMovements()`
  - `revertMovement()`

### Behavioural changes
 Player movement is now subject to less rubberbanding and has more reliable behaviour.

## Tests
Basic testing has been done. There are some corner cases which I'm not completely happy with yet, but those are equivalent to problems which already existed.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -337,6 +337,17 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 	/** @var Vector3|null */
 	protected $forceMoveSync = null;
 
+	/**
+	 * @deprecated this remains solely to preserve BC with plugins that have custom movement systems
+	 * @var Vector3|null
+	 */
+	protected $newPosition;
+	/**
+	 * @deprecated this remains solely to preserve BC with plugins that have custom movement systems
+	 * @var bool
+	 */
+	protected $isTeleporting = false;
+
 	/** @var int */
 	protected $inAirTicks = 0;
 	/** @var float */
@@ -873,7 +884,7 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 	 * @deprecated
 	 */
 	public function getNextPosition() : Position{
-		return $this->getPosition();
+		return $this->newPosition !== null ? Position::fromObject($this->newPosition, $this->level) : $this->getPosition();
 	}
 
 	public function getInAirTicks() : int{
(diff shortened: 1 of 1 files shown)


SatanaDev:
Wow

dktapps:
Considering the high value of the changes in this PR, I've decided to try and make this backwards compatible.

dktapps:
never mind, seems like I can't hack this perfectly... so, it's time to bend the rules a little bit.

PR #3183, opened 2019-11-07
Update to 1.13

buchwasa:
## Introduction

Huge thanks to NetherGames and Nukkit ~~Steadfast2~~ for providing most of the information needed to get a basic implementation done. This PR is needed to update PM to 1.13. Most things that are needed to update PM to 1.13 have been implemented to a working state. Quite honestly, I wouldn't say this is 100% safe to merge yet, I haven't had sufficient testing.

### API changes

- NetworkBinaryStream has 4 new methods: getSkin() : SkinData, putSkin(SkinData), getSkinImage(), and putSkinImage(SkinImage)
- Player->sendRespawnPacket accepts the respawnState parameter
- Added the `SkinImage` and `SkinAnimation` classes
- Added the SkinAdapter interface to deal with the data received.
- Added the LegacySkinAdapter to convert all incoming skins to legacy (for BC reasons)
- Added the SkinAdapterSingleton class with get : SkinAdapter and set(SkinAdapter) allowing developers to create their own skin adapters
- PlayerListEntry no longer accepts a Skin and accepts only SkinData, as well as PlayerSkinPacket

### Behavioural changes

It's still a server, no worries
## Follow-up
- Vanilla resources will need updated

## Tests

This hasn't be thoroughly tested, I've just done some here and there testing and it seems to work okay.


src/pocketmine/Player.php
```diff
@@ -2638,38 +2668,14 @@ public function handleInventoryTransaction(InventoryTransactionPacket $packet) :
...
-
-								if($this->isSurvival()){
-									$slot->pop();
-									$this->inventory->setItemInHand($slot);
-									$this->inventory->addItem($slot->getResidue());
-								}
-
-								return true;
-							}
-
```
dktapps: Why remove this whole block of code, and yet leave behind the case?

src/pocketmine/Player.php
```diff
@@ -834,12 +839,10 @@ public function getLocale() : string{
 	 * Plugin developers should not use this, use setSkin() and sendSkin() instead.
 	 *
 	 * @param Skin   $skin
-	 * @param string $newSkinName
-	 * @param string $oldSkinName
 	 *
 	 * @return bool
 	 */
-	public function changeSkin(Skin $skin, string $newSkinName, string $oldSkinName) : bool{
+	public function changeSkin(Skin $skin) : bool{
```
dktapps: These are here for future usage. If their removal isn't related to a protocol change, leave them alone.

src/pocketmine/Player.php
```diff
@@ -1913,12 +1925,22 @@ public function handleLogin(LoginPacket $packet) : bool{
...
+			base64_decode($packet->clientData["SkinResourcePatch"] ?? ""),
+			new SerializedImage($packet->clientData["SkinImageHeight"], $packet->clientData["SkinImageWidth"], base64_decode($packet->clientData["SkinData"] ?? "")),
+			$animations,
+			new SerializedImage($packet->clientData["CapeImageHeight"], $packet->clientData["CapeImageWidth"], base64_decode($packet->clientData["CapeData"] ?? "")),
+			base64_decode($packet->clientData["SkinGeometryData"] ?? ""),
+			base64_decode($packet->clientData["AnimationData"] ?? ""),
+			$packet->clientData["PremiumSkin"] ?? false,
+			$packet->clientData["PersonaSkin"] ?? false,
+			$packet->clientData["CapeOnClassicSkin"] ?? false,
+			$packet->clientData["CapeId"] ?? ""
```
dktapps: RIP backwards compatibility.

src/pocketmine/Player.php
```diff
@@ -285,8 +290,8 @@ public static function isValidUserName(?string $name) : bool{
 	protected $windowIndex = [];
 	/** @var bool[] */
 	protected $permanentWindows = [];
-	/** @var PlayerCursorInventory */
-	protected $cursorInventory;
+	/** @var PlayerUIInventory */
+	protected $playerUIInventory;
```
dktapps: BC-breaking change.

src/pocketmine/Player.php
```diff
@@ -2520,7 +2542,15 @@ public function handleInventoryTransaction(InventoryTransactionPacket $packet) :
 							}
 						}
 
-						$this->setUsingItem(true);
+						if($this->startAction === -1){
```
dktapps: `isUsingItem()` should be used here.

src/pocketmine/Player.php
```diff
@@ -3804,16 +3819,19 @@ protected function addDefaultWindows(){
...
 		$this->addWindow($this->getArmorInventory(), ContainerIds::ARMOR, true);
 
-		$this->cursorInventory = new PlayerCursorInventory($this);
-		$this->addWindow($this->cursorInventory, ContainerIds::CURSOR, true);
-
-		$this->craftingGrid = new CraftingGrid($this, CraftingGrid::SIZE_SMALL);
+		$this->playerUIInventory = new PlayerUIInventory($this);
+		$this->addWindow($this->playerUIInventory, ContainerIds::UI, true);
 
+		$this->craftingGrid = new CraftingGrid($this->playerUIInventory, CraftingGrid::SIZE_SMALL);
```
dktapps: BC-breaking change.

src/pocketmine/entity/Human.php
```diff
@@ -117,10 +120,16 @@ public function __construct(Level $level, CompoundTag $nbt){
 	protected static function deserializeSkinNBT(CompoundTag $skinTag) : Skin{
 		$skin = new Skin(
 			$skinTag->getString("Name"),
-			$skinTag->hasTag("Data", StringTag::class) ? $skinTag->getString("Data") : $skinTag->getByteArray("Data"), //old data (this used to be saved as a StringTag in older versions of PM)
-			$skinTag->getByteArray("CapeData", ""),
-			$skinTag->getString("GeometryName", ""),
-			$skinTag->getByteArray("GeometryData", "")
+			$skinTag->getString("SkinResourcePatch", ""),
+			new SerializedImage($skinTag->getInt("SkinImageHeight"), $skinTag->getInt("SkinImageWidth"), $skinTag->getByteArray("Data")),
```
dktapps: `SkinImageHeight`, `SkinImageWidth` won't exist in worlds loaded from 1.12. This will cause a crash.

There is a backwards compatibility patch for `Data` that has been removed here that will also cause crashes on old data.

src/pocketmine/entity/Human.php
```diff
@@ -117,10 +120,16 @@ public function __construct(Level $level, CompoundTag $nbt){
...
+			"", //TODO: animations
+			new SerializedImage($skinTag->getInt("CapeImageHeight"), $skinTag->getInt("CapeImageWidth"), $skinTag->getByteArray("CapeData")),
```
dktapps: `CapeImageHeight`, `CapeImageWidth` won't exist in worlds loaded from 1.12. This will cause a crash.

src/pocketmine/entity/Human.php
```diff
@@ -117,10 +120,16 @@ public function __construct(Level $level, CompoundTag $nbt){
...
+			$skinTag->getByteArray("GeometryData", ""),
+			$skinTag->getByteArray("AnimationData", ""),
+			$skinTag->getByte("PremiumSkin") === 1,
```
dktapps: This won't exist on 1.12 world data.

src/pocketmine/entity/Human.php
```diff
@@ -117,10 +120,16 @@ public function __construct(Level $level, CompoundTag $nbt){
...
+			$skinTag->getByte("PersonaSkin") === 1,
```
dktapps: nor this

src/pocketmine/entity/Human.php
```diff
@@ -117,10 +120,16 @@ public function __construct(Level $level, CompoundTag $nbt){
...
+			$skinTag->getByte("CapeOnClassic") === 1,
```
dktapps: nor this

src/pocketmine/entity/Skin.php
```diff
@@ -24,6 +24,8 @@
 namespace pocketmine\entity;
```
dktapps: BC in this file is completely trashed. I don't know where to start. It doesn't look like any attempt has been made to preserve BC whatsoever.

src/pocketmine/inventory/BaseUIInventory.php
```diff
@@ -0,0 +1,116 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\inventory;
+
+use pocketmine\item\Item;
+use pocketmine\item\ItemFactory;
+use pocketmine\Player;
+
+class BaseUIInventory extends BaseInventory{
```
dktapps: The introduction of this class pushes network internals off onto plugin developers for the sake of catering for one change: the addressing of fake inventories is now mapped to window ID `124` and using magic slot numbers, instead of negative window IDs and normal slot numbers.

This is an overcomplicated solution for such a problem which would be better solved at a network-local level with a more isolated hack that doesn't touch the API.

src/pocketmine/inventory/CraftingGrid.php
```diff
@@ -29,134 +29,23 @@
 use function min;
 use const PHP_INT_MAX;
 
-class CraftingGrid extends BaseInventory{
-	public const SIZE_SMALL = 2;
-	public const SIZE_BIG = 3;
+class CraftingGrid extends BaseUIInventory{
+	public const SIZE_SMALL = 4;
```
dktapps: This is a BC-break.

src/pocketmine/inventory/CraftingGrid.php
```diff
@@ -29,134 +29,23 @@
...
+	public const SIZE_BIG = 9;
```
dktapps: and this

src/pocketmine/inventory/CraftingGrid.php
```diff
@@ -29,134 +29,23 @@
...
-
-	/**
-	 * Returns the item at offset x,y, offset by where the starts of the recipe rectangle are.
-	 *
-	 * @param int $x
-	 * @param int $y
-	 *
-	 * @return Item
-	 */
-	public function getIngredient(int $x, int $y) : Item{
```
dktapps: BC break here

src/pocketmine/inventory/CraftingGrid.php
```diff
@@ -29,134 +29,23 @@
...
-
-		throw new \InvalidStateException("No ingredients found in grid");
-	}
-
-	/**
-	 * Returns the width of the recipe we're trying to craft, based on items currently in the grid.
-	 *
-	 * @return int
-	 */
-	public function getRecipeWidth() : int{
```
dktapps: BC break here

src/pocketmine/inventory/CraftingGrid.php
```diff
@@ -29,134 +29,23 @@
...
-		return $this->xLen ?? 0;
-	}
-
-	/**
-	 * Returns the height of the recipe we're trying to craft, based on items currently in the grid.
-	 * @return int
-	 */
-	public function getRecipeHeight() : int{
```
dktapps: BC break here.

src/pocketmine/network/mcpe/protocol/types/RuntimeBlockMapping.php
```diff
@@ -50,20 +53,27 @@ private function __construct(){
 	public static function init() : void{
 		$legacyIdMap = json_decode(file_get_contents(\pocketmine\RESOURCE_PATH . "vanilla/block_id_map.json"), true);
 
-		$compressedTable = json_decode(file_get_contents(\pocketmine\RESOURCE_PATH . "vanilla/required_block_states.json"), true);
+		try{
+			/** @var CompoundTag $tag */
+			$tag = (new BigEndianNBTStream())->read(file_get_contents(\pocketmine\RESOURCE_PATH . "vanilla/runtime_block_states.dat"));
+		}catch(BinaryDataException $e){
+			throw new \RuntimeException("", 0, $e);
```
dktapps: what exactly is the point of this?

src/pocketmine/utils/SkinAnimation.php
```diff
@@ -0,0 +1,52 @@
...
+declare(strict_types=1);
+
+namespace pocketmine\utils;
+
+class SkinAnimation{
+
+	/** @var SerializedImage */
+	private $image;
+	/** @var int */
+	private $type;
```
dktapps: This is undocumented and has no constants.

src/pocketmine/network/mcpe/protocol/types/ResourcePackType.php
```diff
@@ -30,9 +30,13 @@ private function __construct(){
...
-	public const SKINS = 5;
+	public const ADDON = 1;
+	public const CACHED = 2;
+	public const COPY_PROTECTED = 3;
+	public const BEHAVIOR = 4;
+	public const PERSONA_PIECE = 5;
+	public const RESOURCES = 6;
+	public const SKINS = 7;
+	public const WORLD_TEMPLATE = 8;
+	public const COUNT = 9;
```
dktapps: Cargo-cult code.

It's common practice in C++ to add a dummy member to the end of an enum for range-checking, like so:

```
enum things: uint8_t {
    thing1,
    thing2,
    thing3,
    number_of_things
}
```
Here, this doesn't make any sense, because this is not an enum.

src/pocketmine/network/mcpe/protocol/types/ResourcePackType.php
```diff
@@ -30,9 +30,13 @@ private function __construct(){
...
 	public const INVALID = 0;
-	public const RESOURCES = 1;
-	public const BEHAVIORS = 2;
-	public const WORLD_TEMPLATE = 3;
-	public const ADDON = 4; //scripts?
-	public const SKINS = 5;
+	public const ADDON = 1;
+	public const CACHED = 2;
+	public const COPY_PROTECTED = 3;
+	public const BEHAVIOR = 4;
```
dktapps: BC-break here. This should be `BEHAVIORS`.

src/pocketmine/network/mcpe/protocol/types/PlayerListEntry.php
```diff
@@ -34,12 +34,18 @@ class PlayerListEntry{
...
-	/** @var Skin */
-	public $skin;
 	/** @var string */
 	public $xboxUserId;
 	/** @var string */
 	public $platformChatId = "";
+	/** @var int */
+	public $buildPlatform = 0; //Unknown
+	/** @var Skin */
+	public $skin;
```
dktapps: This change is utterly redundant.

src/pocketmine/network/mcpe/protocol/ProtocolInfo.php
```diff
@@ -185,5 +185,13 @@ interface ProtocolInfo{
...
 	public const CLIENT_CACHE_BLOB_STATUS_PACKET = 0x87;
 	public const CLIENT_CACHE_MISS_RESPONSE_PACKET = 0x88;
+	public const EDUCATION_SETTINGS_PACKET = 0x89;
+	public const EMOTE_PACKET = 0x8a;
+	public const MULTIPLAYER_SETTINGS_PACKET = 0x8b;
+	public const SETTINGS_COMMAND_PACKET = 0x8c;
+	public const ANVIL_DAMAGE_PACKET = 0x8d;
+	public const COMPLETED_USING_ITEM_PACKET = 0x8e;
+	public const NETWORK_SETTINGS_PACKET = 0x8f;
+	public const PLAYER_AUTH_INPUT_PACKET = 0x90;
```
dktapps: These packets seem to be missing.

src/pocketmine/network/mcpe/protocol/PlayerListPacket.php
```diff
@@ -29,6 +29,8 @@
 use pocketmine\entity\Skin;
 use pocketmine\network\mcpe\NetworkSession;
 use pocketmine\network\mcpe\protocol\types\PlayerListEntry;
+use pocketmine\utils\SerializedImage;
+use pocketmine\utils\SkinAnimation;
```
dktapps: These imports appear to be redundant.

src/pocketmine/network/mcpe/protocol/PacketPool.php
```diff
@@ -23,6 +23,7 @@
 
 namespace pocketmine\network\mcpe\protocol;
 
+use pocketmine\network\mcpe\protocol\CompletedUsingItemPacket;
```
dktapps: This import is redundant.

src/pocketmine/network/mcpe/protocol/CompletedUsingItemPacket.php
```diff
@@ -0,0 +1,66 @@
...
+	public const ACTION_RETRIEVED = 12;
+	public const ACTION_DYED = 13;
+	public const ACTION_TRADED = 14;
+
+	/** @var int */
+	public $itemId;
+	/** @var int */
+	public $action;
+
+	public function decodePayload(){
```
dktapps: Please apply `void` typehints here for easier integration with `master`.

src/pocketmine/network/mcpe/protocol/CompletedUsingItemPacket.php
```diff
@@ -0,0 +1,66 @@
...
+		$this->itemId = $this->getShort();
+		$this->action = $this->getLInt();
+	}
+
+	public function encodePayload(){
```
dktapps: Please apply `void` typehints here for easier integration with `master`.

src/pocketmine/network/mcpe/protocol/CompletedUsingItemPacket.php
```diff
@@ -0,0 +1,66 @@
...
+	public const ACTION_USE_TOOL = 10;
+	public const ACTION_INTERACT = 11;
+	public const ACTION_RETRIEVED = 12;
+	public const ACTION_DYED = 13;
+	public const ACTION_TRADED = 14;
+
+	/** @var int */
+	public $itemId;
+	/** @var int */
+	public $action;
```
dktapps: Recent style changes prefer private fields and public getters & a `::create()` function, e.g. https://github.com/pmmp/PocketMine-MP/blob/cac3c356a5a83cf6cada0269e86f8d8d11e80c52/src/pocketmine/network/mcpe/protocol/ClientCacheStatusPacket.php#L33-L47

src/pocketmine/network/mcpe/NetworkSession.php
```diff
@@ -23,6 +23,7 @@
 
 namespace pocketmine\network\mcpe;
 
+use pocketmine\network\mcpe\protocol\CompletedUsingItemPacket;
```
dktapps: Imports should be sorted lexicographically ascending.

src/pocketmine/item/Consumable.php
```diff
@@ -26,6 +26,7 @@
 use pocketmine\block\Block;
 use pocketmine\entity\EffectInstance;
 use pocketmine\entity\Living;
+use pocketmine\Player;
```
dktapps: This change is superfluous.

dktapps (changes requested):
I haven't given this a full review, but I've seen enough to haze it so far.

I want to say that this is a worthwhile first attempt at an update. However, several things have significantly overblown complexity and this PR makes no effort whatsoever to preserve backwards compatibility in any arena.

src/pocketmine/Player.php
```diff
@@ -2520,7 +2542,15 @@ public function handleInventoryTransaction(InventoryTransactionPacket $packet) :
...
 						}
 
-						$this->setUsingItem(true);
+						if($this->startAction === -1){
+							$this->setUsingItem(true);
+							return true;
+						}
+
+						$this->setUsingItem(false);
+						if($item->onUse($this)){
```
dktapps: This change creates a behavioural break for food consumption. The events and benefits are supposed to apply after the item is finished being eaten (during the CONSUME transaction type), not when the player just started using the item.

yoraze:
@dktapps, I don’t think he will understand your comments on this PR. As I understand it, this is a "very good copy+paste from NetherGames software".

dktapps:
Thanks, there's no need for that. I think if that's the case it will make itself apparent pretty quickly without toxicity.

Driesboy:
If we wouldn't like our code being used we would have made it opensource. Simple it is.

SOF3:
Please stop spamming this issue with irrelevant flame war. If you have unsettled personal arguments, please resolve them in private.

DaPigGuy:
Don't forget to update the AvailableCommandsPacket constants :)

SleepSpace9:
ItemFrame blocks should also not be forgotten ("Can now be placed on the top and bottom of blocks"). The new metadata for their sides needs to be updated, otherwhise they behave strangely when placed and idk if it will save wrong nbt data to the world too.

yoraze:
The author of this PR copied a broken block palette from SteadFast2, because of this, possibly, similar problems arise.

Wumpotamus:
The CraftingGrid API needs more testing, but PlayerCursorInventory should work just fine, I was able to test that. I'm working on removing the getCompletedAction() stuff, since it was a bad implementation and isn't actually needed to work at all.

Encritary:
@Wumpotamus On my test server, I check if item is consumable and if it is, I do consume logic and set action to ``ACTION_CONSUME``, otherwise the action is ``ACTION_UNKNOWN``. No changes in ``pocketmine\item`` at all.

dktapps:
CompleteUsingItemPacket is superfluous.

Encritary:
@dktapps is it?

dktapps:
It's useless. According to my inside sources, it's only used for statistics.

The bug with food is because the CONSUME type for ReleaseItemTransaction doesn't exist anymore. It's now merged with the regular RELEASE type. The changes needed to address that are significantly less than seen here.

strich3:
when is the 1.13 update Best regards Potato.net Network! 😉

src/pocketmine/network/mcpe/protocol/types/PlayerListEntry.php
```diff
@@ -40,6 +40,12 @@ class PlayerListEntry{
 	public $xboxUserId;
 	/** @var string */
 	public $platformChatId = "";
+	/** @var int */
+	public $buildPlatform = 0; //Unknown
```
yoraze: Could you rename this to `deviceOs`? I think this is a more suitable name for this field.
And it's better to use -1 as the default value: -1 = UNKNOWN_OS

src/pocketmine/network/mcpe/protocol/StartGamePacket.php
```diff
@@ -130,6 +134,8 @@ class StartGamePacket extends DataPacket{
 	/** @var bool */
 	public $onlySpawnV1Villagers = false;
 
+	/** @var string */
+	public $vanillaVersion = ProtocolInfo::MINECRAFT_VERSION_NETWORK;
```
yoraze: I think it’s better to leave an empty string here - the buffer will be a couple of bytes less. This field is useless.

Also `vanillaVersion` refers to LevelSettings, you put newline in the wrong place.

src/pocketmine/network/mcpe/NetworkBinaryStream.php
```diff
@@ -74,6 +77,62 @@ public function putUUID(UUID $uuid) : void{
...
+		$this->putString($skin->getGeometryData());
+		$this->putString($skin->getAnimationData());
+		$this->putBool($skin->getPremium());
+		$this->putBool($skin->getPersona());
+		$this->putBool($skin->getCapeOnClassic());
+		$this->putString($skin->getCapeId());
+		$this->putString($skin->getFullSkinId());
+	}
+
+	public function getImage() : SerializedImage{
```
yoraze: But this does not exist in the protocol...
If you have a PC and can download the IDA, take a look at the serialize method in SerializedSkin.

src/pocketmine/entity/Skin.php
```diff
@@ -120,6 +152,48 @@ public function getGeometryData() : string{
...
+	 */
+	public function getCapeId() : string{
+		return $this->capeId;
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getFullSkinId() : string{
+		return $this->skinId . "_" . $this->capeId;
```
yoraze: You could leave an empty string here, it is useless anyway.
However, I want to note that the client sends SkinId here again.

Sandertv: The cape ID sent is (currently) always empty.

src/pocketmine/network/mcpe/NetworkBinaryStream.php
dktapps: This comment is nonsense. The presence and location of this function have absolutely zero bearing on the protocol.

src/pocketmine/network/mcpe/protocol/StartGamePacket.php
SupremeMortal: From what I've heard, sending anything other than the version will crash the client.

yoraze: Personally, I send an empty string there and everything works fine...

dktapps: works for you != works for everyone

haggling over a premature optimisation like that is a waste of everyone's time.

yoraze: > works for you != works for everyone

But what if I say that everyone who uses my fork with 1.13 support works fine?
By the way, I want to note that BDS itself sends an empty string there.

dktapps: It doesn't matter either way. This is just bikeshedding over a micro optimisation. Let it go.

src/pocketmine/network/mcpe/protocol/PlayerSkinPacket.php
```diff
@@ -38,15 +38,23 @@ class PlayerSkinPacket extends DataPacket{
 	public $uuid;
 	/** @var Skin */
 	public $skin;
+	/** @var string */
+	public $newSkinName = "";
+	/** @var string */
+	public $unknownString = ""; //Sent as empty, assuming it is the old skin name
```
dktapps: This should be `oldSkinName` as per the original code.

src/pocketmine/network/mcpe/PlayerNetworkSessionAdapter.php
```diff
@@ -248,7 +253,7 @@ public function handleResourcePackChunkRequest(ResourcePackChunkRequestPacket $p
 	}
 
 	public function handlePlayerSkin(PlayerSkinPacket $packet) : bool{
-		return $this->player->changeSkin($packet->skin, $packet->newSkinName, $packet->oldSkinName);
+		return $this->player->changeSkin($packet->skin, "", "");
```
dktapps: This change should be removed.

Wumpotamus:
As of the latest commit, legacy skins are the only working skins, persona skins are invisible and there's a weird bug that if you pause the screen, your skin goes invisible. As discussed on discord, the plan is to convert persona skins to a legacy skin, of a random color

src/pocketmine/network/mcpe/NetworkBinaryStream.php
```diff
@@ -74,6 +77,62 @@ public function putUUID(UUID $uuid) : void{
...
+		$capeOnClassic = $this->getBool();
+		$capeId = $this->getString();
+		$fullSkinId = $this->getString();
+
+		return new Skin($skinId, $skinData->getData(), $capeData->getData(), $skinResourcePatch, $geometryData);
+	}
+
+	public function putSkin(Skin $skin){
+		$this->putString($skin->getSkinId());
+		$this->putString($skin->getGeometryName()); //resource patch
```
dktapps: doesn't this need converting to a skin resource patch?

src/pocketmine/utils/SerializedImage.php
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @author PocketMine Team
+ * @link http://www.pocketmine.net/
+ *
+ *
+*/
+
+declare(strict_types=1);
+
+namespace pocketmine\utils;
```
dktapps: This should be moved to the `pocketmine\network\mcpe\protocol\types` namespace.

src/pocketmine/utils/SkinAnimation.php
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * @author PocketMine Team
+ * @link http://www.pocketmine.net/
+ *
+ *
+*/
+
+declare(strict_types=1);
+
+namespace pocketmine\utils;
```
dktapps: This should be moved to the `pocketmine\network\mcpe\protocol\types` namespace.

dktapps (changes requested):
My recommendation was to provide a separate protocol-specific `Skin` type, which would be converted to a legacy skin at packet handle and before sending.

We don't want to lose information from packets when debugging. The current BC impl simply discards information it doesn't find useful.

src/pocketmine/utils/SerializedImage.php
```diff
@@ -0,0 +1,67 @@
...
+
+class SerializedImage{
```
dktapps: This probably should be named `SkinImage`.

src/pocketmine/network/mcpe/NetworkBinaryStream.php
```diff
@@ -74,6 +77,62 @@ public function putUUID(UUID $uuid) : void{
...
+		$this->putImage(SerializedImage::fromLegacy($skin->getSkinData()));
+		/** @var SkinAnimation[] $animations */
+		$animations = [];
+		$this->putLInt(count($animations));
+		foreach($animations as $animation){
+			$this->putImage($animation->getImage());
+			$this->putLInt($animation->getType());
+			$this->putLFloat($animation->getFrames());
+		}
+		$this->putImage(new SerializedImage(0, 0, $skin->getCapeData()));
```
dktapps: This looks likely to cause problems for non-null capes.

src/pocketmine/Player.php
```diff
@@ -2913,6 +2911,14 @@ public function handleAnimate(AnimatePacket $packet) : bool{
 		return true;
 	}
 
+	public function handleRespawn(RespawnPacket $packet) : bool{
+		if(!$this->isAlive() && $packet->respawnState === RespawnPacket::CLIENT_READY_TO_SPAWN){
+			$this->sendRespawnPacket($this, RespawnPacket::READY_TO_SPAWN);
+		}
+
+		return true;
```
dktapps: This should return unhandled if it arrived in an unexpected condition.

src/pocketmine/entity/Skin.php
```diff
@@ -55,6 +55,10 @@ public function __construct(string $skinId, string $skinData, string $capeData =
 		$this->geometryData = $geometryData;
 	}
 
+	public static function convertToLegacyName(string $name) : string{
+		return '{"geometry" : {"default" : "' . $name . '"}}';
```
dktapps: This looks susceptible to injection bugs.
```suggestion
		return json_encode(["geometry" => ["default" => $name]]);
```

src/pocketmine/level/particle/FloatingTextParticle.php
```diff
@@ -96,7 +96,7 @@ public function encode(){
 
 			$add = new PlayerListPacket();
 			$add->type = PlayerListPacket::TYPE_ADD;
-			$add->entries = [PlayerListEntry::createAdditionEntry($uuid, $this->entityId, $name, new Skin("Standard_Custom", str_repeat("\x00", 8192)))];
+			$add->entries = [PlayerListEntry::createAdditionEntry($uuid, $this->entityId, $name, new Skin("Standard_Custom", str_repeat("\x00", 8192), "", Skin::convertToLegacyName("geometry.humanoid.custom"), ""))];
```
dktapps: Is this change still necessary with BC core skins?

src/pocketmine/network/mcpe/NetworkBinaryStream.php
```diff
@@ -74,6 +77,62 @@ public function putUUID(UUID $uuid) : void{
...
 	}
 
+	public function getSkin() : Skin{
+		$skinId = $this->getString();
+		$skinResourcePatch = $this->getString();
+		$skinData = $this->getImage();
+		$animationCount = $this->getLInt();
+		$animations = [];
+		for($i = 0; $i < $animationCount; ++$i){
+			$animations[] = new SkinAnimation($this->getImage(), $this->getLInt(), $this->getLFloat());
```
dktapps: For the sake of code clarity, please make the following change:
```suggestion
			$animations[] = new SkinAnimation(
				$image = $this->getImage(),
				$animationType = $this->getLInt(),
				$animationFrames = $this->getLFloat()
			);
```
This makes it clear what the purpose of the fields being read are, and refactoring will persist these assignments.

src/pocketmine/network/mcpe/NetworkBinaryStream.php
```diff
@@ -74,6 +77,62 @@ public function putUUID(UUID $uuid) : void{
...
+		$this->putString($skin->getGeometryData());
+		$this->putString(""); //animation data
+		$this->putBool(false); //isPremium
+		$this->putBool(false); //isPersona
+		$this->putBool(false); //isCapeOnClassic
+		$this->putString(""); //capeId
+		$this->putString(""); //fullskinId
+	}
+
+	public function getImage() : SerializedImage{
```
dktapps: This should be renamed to `getSkinImage()` to avoid confusion.
It's also not clear to me that this needed to be public.

src/pocketmine/network/mcpe/NetworkBinaryStream.php
```diff
@@ -74,6 +77,62 @@ public function putUUID(UUID $uuid) : void{
...
+		$width = $this->getLInt();
+		$height = $this->getLInt();
+		$data = $this->getString();
+		return new SerializedImage($height, $width, $data);
+	}
+
+	public function putImage(SerializedImage $image) : void{
```
dktapps: This should be renamed to `putSkinImage()` to avoid confusion.
It's also not clear to me that this needed to be public.

src/pocketmine/network/mcpe/NetworkBinaryStream.php
```diff
@@ -74,6 +77,66 @@ public function putUUID(UUID $uuid) : void{
...
+		$capeData = $this->getSkinImage();
+		$geometryData = $this->getString();
+		$animationData = $this->getString();
+		$premium = $this->getBool();
+		$persona = $this->getBool();
+		$capeOnClassic = $this->getBool();
+		$capeId = $this->getString();
+		$fullSkinId = $this->getString();
+
+		return new Skin($skinId, $skinData->getData(), $capeData->getData(), $skinResourcePatch, $geometryData);
```
dktapps: the geometry name should be extracted from the resource patch `default` field, as far as I understand.

dktapps:
Latest commits are starting to look a little more interesting, but the geometry name handling is still incorrect. It should be extracted from `skinResourcePatch["geometry"]["default"]`.

src/pocketmine/network/mcpe/protocol/types/LegacySkinAdapter.php
```diff
@@ -32,6 +32,10 @@ public function toSkinData(Skin $skin) : SkinData{
 	}
 
 	public function fromSkinData(SkinData $data) : Skin{
-		return new Skin($data->skinId, $data->skinImage->getData(), $data->capeImage->getData(), $data->resourcePatch, $data->geometryData);
+		$skinData = $data->skinImage->getData();
+		if($data->persona){
+			$skinData = str_repeat(random_bytes(3) . "\xff", 2048);
+		}
+		return new Skin($data->skinId, $skinData, $data->capeImage->getData(), $data->resourcePatch, $data->geometryData);
```
dktapps: There may need to be additional checks for `personaCapeOnClassicSkin`.

src/pocketmine/network/mcpe/protocol/PlayerListPacket.php
```diff
@@ -89,13 +79,12 @@ protected function encodePayload(){
...
 				$this->putString($entry->username);
-				$this->putString($entry->skin->getSkinId());
-				$this->putString($entry->skin->getSkinData());
-				$this->putString($entry->skin->getCapeData());
-				$this->putString($entry->skin->getGeometryName());
-				$this->putString($entry->skin->getGeometryData());
 				$this->putString($entry->xboxUserId);
 				$this->putString($entry->platformChatId);
+				$this->putLInt($entry->buildPlatform);
+				$this->putSkin($entry->skinData);
```
dktapps: this looks like a bug

src/pocketmine/network/mcpe/protocol/PlayerListPacket.php
```diff
@@ -57,22 +57,12 @@ protected function decodePayload(){
...
-					$skinId,
-					$skinData,
-					$capeData,
-					$geometryName,
-					$geometryData
-				);
 				$entry->xboxUserId = $this->getString();
 				$entry->platformChatId = $this->getString();
+				$entry->buildPlatform = $this->getLInt();
+				$entry->skin = $this->getSkin();
```
dktapps: this looks like a bug

src/pocketmine/network/mcpe/protocol/types/LegacySkinAdapter.php
```diff
@@ -0,0 +1,41 @@
...
+	public function toSkinData(Skin $skin) : SkinData{
+		return new SkinData($skin->getSkinId(), json_encode(["geometry" => ["default" => $skin->getGeometryName()]]), SkinImage::fromLegacy($skin->getSkinData()), [], new SkinImage(32, 64, $skin->getCapeData()), $skin->getGeometryData());
+	}
+
+	public function fromSkinData(SkinData $data) : Skin{
+		$skinData = $data->skinImage->getData();
+		if($data->persona){
+			$skinData = str_repeat(random_bytes(3) . "\xff", 2048);
+		}
+		return new Skin($data->skinId, $skinData, $data->capeImage->getData(), json_decode($data->resourcePatch, true)["geometry"]["default"], $data->geometryData);
```
dktapps: I wouldn't assume the `resourcePatch` is valid; it might result in an uncaught error and crash.

src/pocketmine/level/particle/FloatingTextParticle.php
```diff
@@ -96,7 +97,9 @@ public function encode(){
 
 			$add = new PlayerListPacket();
 			$add->type = PlayerListPacket::TYPE_ADD;
-			$add->entries = [PlayerListEntry::createAdditionEntry($uuid, $this->entityId, $name, new Skin("Standard_Custom", str_repeat("\x00", 8192)))];
+			$add->entries = [PlayerListEntry::createAdditionEntry($uuid, $this->entityId, $name, SkinAdapterSingleton::get()->toSkinData(new Skin(
+				"Standard_Custom", str_repeat("\x00", 8192), "", '{"geometry" : {"default" : "geometry.humanoid.custom"}}' //TODO: Remove, hack to fix ftp
```
dktapps: It's not clear to me what the purpose of this hack is, and to my understanding of the code, this hardcoded JSON should not be here.

src/pocketmine/network/mcpe/protocol/types/SkinAdapterSingleton.php
```diff
@@ -0,0 +1,40 @@
...
+ *
+*/
+
+declare(strict_types=1);
+
+namespace pocketmine\network\mcpe\protocol\types;
+
+class SkinAdapterSingleton{
+
+	private static $skinAdapter = null;
```
dktapps: Missing `@var` doc comment here.

src/pocketmine/network/mcpe/protocol/types/LegacySkinAdapter.php
```diff
@@ -0,0 +1,41 @@
...
+class LegacySkinAdapter implements SkinAdapter{
+
+	public function toSkinData(Skin $skin) : SkinData{
+		return new SkinData($skin->getSkinId(), json_encode(["geometry" => ["default" => $skin->getGeometryName()]]), SkinImage::fromLegacy($skin->getSkinData()), [], new SkinImage(32, 64, $skin->getCapeData()), $skin->getGeometryData());
+	}
+
+	public function fromSkinData(SkinData $data) : Skin{
+		$skinData = $data->skinImage->getData();
+		if($data->persona){
+			$skinData = str_repeat(random_bytes(3) . "\xff", 2048);
```
dktapps: This is not sufficient. The geometry model information, persona cape and more will persist, which is likely the reason for your headless players.
```suggestion
			return new Skin("Standard_Custom", str_repeat(random_bytes(3) . "\xff", 2048), "", "geometry.humanoid.custom");
```

src/pocketmine/network/mcpe/protocol/types/LegacySkinAdapter.php
dktapps: It's possible to have a Persona cape attached to a classic skin. Capes should be removed if they aren't classic.

src/pocketmine/network/mcpe/protocol/types/SkinData.php
```diff
@@ -0,0 +1,64 @@
...
+	/** @var string */
+	public $geometryData;
+	/** @var string */
+	public $animationData;
+	/** @var bool */
+	public $persona;
+	/** @var bool */
+	public $premium;
+	/** @var bool */
+	public $capeOnClassic;
```
dktapps: This would be better named `personaCapeOnClassicSkin`.

src/pocketmine/network/mcpe/protocol/types/SkinData.php
```diff
@@ -0,0 +1,64 @@
...
+	/** @var string */
+	public $capeId;
```
dktapps: These fields shouldn't be public; please make them private and generate getters for them.

src/pocketmine/network/mcpe/protocol/types/SkinData.php
```diff
@@ -0,0 +1,64 @@
...
+
+	public function __construct(string $skinId, string $resourcePatch, SkinImage $skinImage, array $animations = [], SkinImage $capeImage = null, string $geometryData = "", string $animationData = "",  bool $premium = false, bool $persona = false, bool $capeOnClassic = false, string $capeId = ""){
```
dktapps: Please provide `@param` docs here; there is an `array` parameter whose type isn't clear from the native typehints.

dktapps: Again here: `capeOnClassic` would be better named `personaCapeOnClassicSkin`.

dktapps:
There are some additional protocol changes which have not been addressed, which can be seen in this diff: https://github.com/Sandertv/gophertunnel/commit/139aa27d2bc8ea4755b6cf2e4ace71c647aab461

- `CraftingDataPacket` has two additional varint-prefixed lists before the end of the packet containing `PotionMix` and `PotionContainerMix` structures (paraphrasing, pls choose better names if you can think of any):
  - `PotionRecipe` (for brewing) has an input potion ID (item metadata), ingredient item ID, and output potion ID (item metadata)
  - `PotionContainerMix` (for e.g. converting a normal potion to splash potion) has an input item ID, ingredient item ID, and output item ID
- There is an extra list of shit at the end of `AvailableCommandsPacket` that doesn't make much sense right now. At the least, it should be filled out with a zero length and a TODO placed.
- `isEduMode` in `StartGamePacket` appears to have changed; see https://github.com/Sandertv/gophertunnel/commit/139aa27d2bc8ea4755b6cf2e4ace71c647aab461#diff-daacd64fd2e506c3704bcabd09f4636aR65

src/pocketmine/network/mcpe/protocol/StartGamePacket.php
```diff
@@ -83,8 +83,8 @@ class StartGamePacket extends DataPacket{
 	public $hasAchievementsDisabled = true;
 	/** @var int */
 	public $time = -1;
-	/** @var bool */
-	public $eduMode = false;
+	/** @var int */
+	public $eduEditionOffer = 0;
```
yoraze: I ask you to make some changes to Player about this. (In Player you can find sending StartGamePacket and do something with it, otherwise there will be extra crashdumps :P)

Wumpotamus: Thanks

src/pocketmine/network/mcpe/protocol/types/LegacySkinAdapter.php
```diff
@@ -43,11 +43,16 @@ public function toSkinData(Skin $skin) : SkinData{
 
 	public function fromSkinData(SkinData $data) : Skin{
 		$capeData = $data->getCapeImage()->getData();
+		$geometryName = "";
+		$resourcePatch = json_decode($data->getResourcePatch(), true);
+		if(isset($resourcePatch["geometry"]) && isset($resourcePatch["geometry"]["default"])){
```
dktapps: This isn't sufficient. The `geometry` key might not be an array, or the `default` key might not be a string.

dktapps: In addition, the player should be rejected with `Invalid skin` if the resource patch is invalid.

Wumpotamus:
You don't, this isn't ready for production and shouldn't be compiled

Wumpotamus:
note for later:

Inventory offsets go as:
```
Cursor - 0
Anvil - 1
Enchant - 14
Beacon - 27
Crafting Small - 28
Crafting Big - 32
```

dktapps:
seems like there are some gaps?

nituzhazha:
I used the method of setSkin() to set the skin, but the skin could not be shown in the test. Is this the core problem or I may have written it wrong? Please tell me

dktapps:
This isn't a plugin support forum.

nituzhazha:
> This isn't a plugin support forum.

I'm sorry...

dktapps:
Doors and item frame placement is broken; doors don't hinge correctly and item frames are ... screwed.

dktapps:
There has been further discussion on the skins issue, and it's come to light that the `SkinAdapter` may not be sufficient for third party developers to bolt on Persona support properly. In addition, players think that the solid colour skins are a bug.

Given that there is no way to implement Persona properly without a clean BC break, it may be necessary to release a new major version to support 1.13, which has skin changes and nothing else.

superbobby2000:
I don't understand how PlayerListEntry work now

PR #3192, opened 2019-11-21
Compatible Operating Systems

jasonw4331:
## Introduction

This PR is here to add a feature which will allow developers to include compatible operating systems within their plugin.yml where the plugin will be allowed to load.

### Relevant issues
N/A

### API changes
`PluginDescription::getCompatibleOperatingSystems()` is created for retrieving the compatible operating systems listed.
`PluginDescription::checkCompatibleOperatingSystem()` is created for testing whether the OS of the host is listed.

### Behavioural changes
The PluginManager class now checks for compatible Server OS before any required extensions

## Backwards compatibility
The inclusion of the compatible operating systems is completely optional and does not break any previous API.

## Follow-up

Requires translations:

| Name | Value in eng.ini |
| :--: | :---: |
| `pocketmine.plugin.incompatibleOS` | `Incompatible Operating System (plugin requires one of: {%0})` |

Newly supported operating systems should be documented and added to `Utils::getOS()` for manifest-checking support.

Any plugin.yml schema documentation should be updated.

## Tests

```yaml
name: GitRollbacks
main: jasonwynn10\GitRollbacks\Main
version: 0.1.0
api: 3.0.0
description: "just a plugin"
author: "jasonwynn10"
load: STARTUP
os:
- win
- linux
- mac
```


src/pocketmine/plugin/PluginDescription.php
```diff
@@ -170,6 +174,24 @@ public function getCompatibleApis() : array{
...
+	
+	/**
+	 * @return string[]
+	 */
+	public function getCompatibleOperatingSystems() : array{
+		return $this->compatibleOperatingSystems;
+	}
+	
+	/**
+	 * Checks if the current operating system is in the compatible systems list.
```
SOF3: Consider linking to `Utils::getOS()` for the list of systems.

dktapps:
The names could be reduced to `OS` instead of `OperatingSystem`. The current ones are long and not particularly readable.

JackMD:
Also how about `os` key in plugin.yml instead of `operating-systems`?

lukeeey:
wouldnt this encourage plugin developers to be lazy and not support multiple operating systems?

jasonwynn10:
I think they would be too lazy in the first place to know this exists as part of the plugin.yml if that were the case

dktapps:
Please rebase your branch onto the updated `next-minor` (or merge it if you prefer). it fixes a problem with the CI that's currently impacting this PR.

src/pocketmine/plugin/PluginManager.php
```diff
@@ -158,6 +158,7 @@ public function loadPlugin(string $path, array $loaders = null) : ?Plugin{
 				if($description instanceof PluginDescription){
 					$this->server->getLogger()->info($this->server->getLanguage()->translateString("pocketmine.plugin.load", [$description->getFullName()]));
 					try{
+						$description->checkCompatibleOperatingSystem();
 						$description->checkRequiredExtensions();
```
dktapps: `checkRequiredExtensions()` was already misplaced here; your check should more likely be placed in [this](https://github.com/pmmp/PocketMine-MP/pull/3192/files#diff-10c30a902a542d91471a1a004b245c13R256-R286) block, following the style of other compatibility checks.

src/pocketmine/plugin/PluginDescription.php
```diff
@@ -170,6 +174,24 @@ public function getCompatibleApis() : array{
...
+	 *
+	 * @throws PluginException if the current operating system is not compatible
+	 */
+	public function checkCompatibleOperatingSystem() {
```
dktapps: I think this function is redundant: it's only used in 1 place and it can be checked in the same fashion as everything else [here](https://github.com/pmmp/PocketMine-MP/pull/3192/files#diff-10c30a902a542d91471a1a004b245c13R256-R286).

src/pocketmine/plugin/PluginDescription.php
```diff
@@ -97,6 +100,7 @@ private function loadMap(array $plugin){
 
 		$this->api = array_map("\strval", (array) ($plugin["api"] ?? []));
 		$this->compatibleMcpeProtocols = array_map("\intval", (array) ($plugin["mcpe-protocol"] ?? []));
+		$this->compatibleOperatingSystems = array_map("\strval", (array) ($plugin["operating-systems"] ?? []));
```
dktapps: I was under the impression we were using `os` for this key. Maybe you lost some changes when you force-pushed.

src/pocketmine/plugin/PluginManager.php
```diff
@@ -274,6 +273,14 @@ public function loadPlugins(string $directory, array $newLoaders = null){
 						continue;
 					}
 
+					if(!empty($this->compatibleOperatingSystems) and !in_array(Utils::getOS(), $this->compatibleOperatingSystems)) {
```
dktapps: bad copy pasta

src/pocketmine/plugin/PluginManager.php
```diff
@@ -274,6 +273,14 @@ public function loadPlugins(string $directory, array $newLoaders = null){
...
+						$this->server->getLogger()->error($this->server->getLanguage()->translateString("pocketmine.plugin.loadError", [
+							$name,
+							$this->server->getLanguage()->translateString("%pocketmine.plugin.incompatibleOS", [implode(", ", $this->compatibleOperatingSystems)])
```
dktapps: and again here

src/pocketmine/plugin/PluginManager.php
dktapps: please avoid usage of empty(), it includes isset() logic, which is concealing a bug in this code.

src/pocketmine/plugin/PluginManager.php
```diff
@@ -273,6 +273,14 @@ public function loadPlugins(string $directory, array $newLoaders = null){
 						continue;
 					}
 
+					if(!empty($description->getCompatibleOperatingSystems()) and !in_array(Utils::getOS(), $description->getCompatibleOperatingSystems())) {
```
dktapps:
```suggestion
					if(count($description->getCompatibleOperatingSystems()) > 0 and !in_array(Utils::getOS(), $description->getCompatibleOperatingSystems(), true)) {
```

src/pocketmine/plugin/PluginDescription.php
```diff
@@ -97,6 +100,7 @@ private function loadMap(array $plugin){
 
 		$this->api = array_map("\strval", (array) ($plugin["api"] ?? []));
 		$this->compatibleMcpeProtocols = array_map("\intval", (array) ($plugin["mcpe-protocol"] ?? []));
```
SOF3: Who added the `\` here? That's both wrong string literal syntax and unnecessary.

dktapps: It's not wrong and it's not unnecessary either, since imports aren't respected at this scope.

dktapps: apparently PhpStorm has a bug and this is not necessary, because callback references are evaluated in the root namespace scope anyway. However, it isn't technically wrong.

how about we just replace all callables with closures instead? lol...

SOF3 (approved):
Looks OK. Although I am not very happy about the long `OperatingSystems` in the field/getter name, it would be even more odd to have `Os`/`OS`/`OSs`/`OSes`/`Oss`/`Oses`. (wtf am I writing...)

Otherwise, in general, this change is consistent with other plugin.yml attributes and can be merged.

SOF3:
@dktapps If you are merging this, please also review whatever documentation we have about plugin.yml format.

That is... if we have any that is even updated.

dktapps:
https://github.com/pmmp/DeveloperDocs/blob/master/plugin-manifest-spec.md#plugin-loading-controls

dktapps:
There is a problem with this that just jumped into my head: if someone tries to load a plugin that uses this directive on a version older than whatever supports the `os` directive, your plugin will explode. I'm not sure how this issue can be resolved short of manually requiring the minimum API version that this directive is supported on.

dktapps:
Merging this as-is; it's not possible to fix the problem I mentioned.

HimbeersaftLP:
Well considering that the developer declaring an older API guarantees compatibility with said version, it's to be assumed that they also consider plugin.yml properties in this.

dktapps:
Please submit a PR to https://github.com/pmmp/PocketMine-Language to add the required language strings to `eng.ini`.

jasonwynn10:
https://github.com/pmmp/PocketMine-Language/pull/54

dktapps:
My bad. I didn't realize.

PR #3194, opened 2019-11-22
FIx possible grammar error

rjworks:
FIx possible grammar error


rjworks:
or is pmmp/PocketMine supposed to be pmmp and PocketMine? Are they different?

PR #3197, opened 2019-11-29
Add php directive to plugin.yml

### LINKED ISSUE #3195: Add a `php` directive to plugin.yml
## Description
We need a directive in plugin.yml to constrain the versions of PHP that a plugin will load on.

## Justification
Recently it's become more and more apparent that some plugin devs are using 7.3+ only features, while PM itself is still supporting 7.2. This is a problem because users running 7.2 won't know there is a problem with the plugin until it crashes.

buchwasa:
## Introduction
This PR adds support for php versions within plugin.yml. The plugin will only load under the conditions that it matches the exact version of php you're running and the version(s) you specify in plugin.yml

### Relevant issues
Closes #3195 

### API changes

Adds `getCompatiblePhpVersions()` to PluginDescription

### Behavioural changes
The php version will be checked right after mcpe-protocol is checked.

## Backwards compatibility
This is backwards compatible with the rest of 3.x.

## Follow-up
I would like to add translations, but decided to get this PR checked first.

## Tests
```yaml
name: test
author: Drew
version: 1
main: test\Main
api: 3.0.0
php: [7.3, 7.3.9]
```


dktapps:
Newer PHP versions might cause incompatibilities as well, a minversion might not be sufficient.

SOF3:
What makes this more significant than other directives like `os` is that new PHP versions might introduce language features that would lead to syntax error or other critical error on the main class. Developers would then have the burden to maintain backward compatibility with older PHP versions by not using such language features on the main class. On the contrary, plugin.yml directives can prevent the main class from being loaded.

src/pocketmine/plugin/PluginDescription.php
```diff
@@ -171,6 +176,13 @@ public function getCompatibleMcpeProtocols() : array{
 		return $this->compatibleMcpeProtocols;
 	}
 
+	/**
+	 * @return string
+	 */
+	public function getMinPHPVersion() : string{
```
SOF3: Please use `Php` not `PHP` in camelCase.

src/pocketmine/plugin/PluginManager.php
```diff
@@ -284,6 +284,16 @@ public function loadPlugins(string $directory, array $newLoaders = null){
 						}
 					}
 
+					if($description->getMinPHPVersion() !== ""){
+						if(version_compare($description->getMinPHPVersion(), PHP_VERSION) > 0){
+							$this->server->getLogger()->error($this->server->getLanguage()->translateString("pocketmine.plugin.loadError", [
```
SOF3: It is a better idea to use semver-based version matching, just like how we match API versions. Supporting 7.3.2 should only imply support for 7.3.{≥2} and 7.{≥4}.?, but not 8.?.?.

SOF3 (changes requested):

src/pocketmine/plugin/PluginManager.php
dktapps: for posterity, recreating the same comment here that I wrote on discord: PHP does not use semver, breaks can happen on minor versions

SatanaDev:
Tf...:dizzy_face:

JaxkDev:
Just an idea here, as dylan said any version can break things what about having php tag an array of supported PHP v's and if non present assume all.

lukeeey:
That sounds like the best idea ^

SOF3:
I think just having minPhpVersion is fine. PHP is mostly backward-compatible across minors, and if incompatibilities are encountered, it would be some really rare case that it's either very obvious (leading to error) or very unobvious (e.g. corner case precedence issues). Hence it's safe to assume forward compatibility. If the developer knows there is a max version, he'd simply fix the problem rather than adding the max version.

As for very major BC breaks from PHP, it is also likely that PM has to bump major API version as well (such as PHP 5.6 -> 7.0). Then there is no need to check anyway.

dktapps:
> If the developer knows there is a max version, he'd simply fix the problem rather than adding the max version.

this logic is flawed and assumes that the burden of updating is smaller than the burden of adding a maximum bound or locking into a specific version. Reality usually doesn't work this way.

Wumpotamus:
I've switched it to the array system, under the impression that was the verdict for it.

SOF3:
I don't think it's practical to list out every PHP patch version compatible, since there are really lots of them.

PR #3220, opened 2019-12-10
Bump protocol version to 389

Kaooot:


yoraze:
Are you completely sure that the packets and "protocol constants" have not changed at all?

dktapps:
the capacity for humans to not read information that is right in their faces never fails to amaze me ...
https://github.com/pmmp/PocketMine-MP/pull/3220/files#diff-5a5fa064352c66bac5e4e6fade5bcb63R36

PR #3244, opened 2019-12-15
added GameRule name constants

Saehing:
This innovation will add GameRules values to the kernel, they can be used in plugins

### DIFF
--- a/src/network/mcpe/protocol/StartGamePacket.php
+++ b/src/network/mcpe/protocol/StartGamePacket.php
@@ -30,6 +30,7 @@ use pocketmine\math\Vector3;
 use pocketmine\nbt\tag\ListTag;
 use pocketmine\nbt\TreeRoot;
 use pocketmine\network\mcpe\handler\PacketHandler;
+use pocketmine\network\mcpe\protocol\types\GameRules;
 use pocketmine\network\mcpe\protocol\types\PlayerPermissions;
 use pocketmine\network\mcpe\protocol\types\RuntimeBlockMapping;
 use pocketmine\network\mcpe\serializer\NetworkBinaryStream;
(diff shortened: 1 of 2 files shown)


SOF3:
Everything in the `protocol` namespace is internal and is not expected to be used from plugins.

Saehing:
> Everything in the `protocol` namespace is internal and is not expected to be used from plugins.

your proposal where can be this to place?

dktapps:
I think the point is that plugins are not supposed to use it, being internal.

I have my doubts about the gamerule name casing. `naturalregeneration` != `naturalRegeneration`.

Saehing:
> I think the point is that plugins are not supposed to use it, being internal.
> 
> I have my doubts about the gamerule name casing. `naturalregeneration` != `naturalRegeneration`.

I checked, on this occasion all is well

SOF3:
> there are times when you need to include for example the display of coordinates during the game

Then PocketMine should expose an API for that. Exposing protocol-specific constants does not change anything for plugins.

dktapps:
since this pull request is only adding a few constants it can't really be described as implementing game rules. Nonetheless, the addition of such constants can be valuable, but we can generate these from command data anyway; they don't need to be created by hand.

dktapps:
> I checked, on this occasion all is well

does that mean you checked if the client crashes, or does it mean you checked if the bug it is supposed to fix still stays fixed?

Saehing:
> > I checked, on this occasion all is well
> 
> does that mean you checked if the client crashes, or does it mean you checked if the bug it is supposed to fix still stays fixed?

I took all the values from minecraft: bedrock edition 1.14, after checking everything, the player understands them perfectly. But the value of RandomTickSpeed the client does not want to perceive :/
image

Saehing:
> > I checked, on this occasion all is well
> 
> does that mean you checked if the client crashes, or does it mean you checked if the bug it is supposed to fix still stays fixed?

Is there any problem in adding constants , or is it a matter of time?

dktapps:
As far as gamerules go, this is one of the areas that least needs attention, particularly since any skriptkiddie gamerule hacking right now will likely involve hardcoded strings anyway.
This file is not automatically generated, meaning that the value that might be granted by having these as constants is currently nullified, and there's also no type information associated with the gamerule definitions.

dktapps:
Closing this since it's stale and has no practical value right now. When/if it's needed, it can be generated.

PR #3247, opened 2019-12-17
PlayerAuthInputPacket: Fix assigning variable in static context

lukeeey:
pretty self explanatory


dktapps:
this should be targeted at the stable line since that's the lowest branch where the bug appears.

SOF3:
Merged in 216138a37e52e545abdba5a3600eabbcced867be

PR #3256, opened 2019-12-22
Add comment to make what setDuration for effect is in more obvious

Zedstar16:
## Introduction

Noob might not realise that setDuration() for an effect instance is in ticks rather than seconds,
Comment makes it more obvious

### DIFF
--- a/src/pocketmine/entity/EffectInstance.php
+++ b/src/pocketmine/entity/EffectInstance.php
@@ -82,7 +82,7 @@ class EffectInstance{
 	}
 
 	/**
-     * Sets the duration to the given number of ticks
+	 * Sets the duration to the given number of ticks
      * 
 	 * @param int $duration
 	 *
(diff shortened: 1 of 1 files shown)


dktapps (commented):
looks OK, but pls use tabs for indentation instead of spaces

src/pocketmine/entity/EffectInstance.php
```diff
@@ -82,6 +82,8 @@ public function getDuration() : int{
 	}
 
 	/**
+	 * Sets the duration to the given number of ticks
+     *
```
SOF3: and the second line too...

SOF3 (changes requested):

src/pocketmine/entity/EffectInstance.php
dktapps: why is this marked as resolved?

dktapps:
Closing this due to too much hassle for such a small change; it can be applied by a maintainer without waiting for review comments to be addressed.

PR #3268, opened 2020-01-02
changed the execution sequence of the getCurrentWindow() function

duphless:
due to the fact that the onCurrentWindowChange() function is faster than $inventory->open($this); what happens is that the actions in onOpen() do not have time to be processed before sending the inventory to the player.

This prevents the implementation of custom inventory.

### DIFF
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -2468,8 +2468,8 @@ class Player extends Human implements CommandSender, ChunkLoader, ChunkListener,
 		$this->removeCurrentWindow();
 
 		$this->logger->debug("Opening inventory " . get_class($inventory) . "#" . spl_object_id($inventory));
-		$this->networkSession->getInvManager()->onCurrentWindowChange($inventory);
 		$inventory->onOpen($this);
+		$this->networkSession->getInvManager()->onCurrentWindowChange($inventory);
 		$this->currentWindow = $inventory;
 		return true;
 	}


dktapps:
please provide more detail, the description of the problem is not clear...

duphless:
> please provide more detail, the description of the problem is not clear...

here I create my custom inventory:
https://gist.github.com/duphless/3c876968059bcc55a22ab2ea088c3e79

in onOpen (), I have a block created for this inventory, but these actions are performed after the player is sent a ContainerOpenPacket

dktapps:
I should point out that the whole reason this code was split up like this is to make sure that network and core are isolated, so it's not expected for onOpen() to be doing network logic.

duphless:
> I should point out that the whole reason this code was split up like this is to make sure that network and core are isolated, so it's not expected for onOpen() to be doing network logic.

okay, thanks! can I close the PR?

dktapps:
I'm still trying to decide whether this is useful or not. While it looks simple it can have side effects.

duphless:
> I'm still trying to decide whether this is useful or not. While it looks simple it can have side effects.

well, I won't interfere. according to my tests, there is no harm from this permutation, but this very permutation can help very much in the future

dktapps:
Yes, but it breaks potential assumptions that this is called before, so if there is logic that a plugin wants to do after the inventory has been opened, then it can't do that with this change.

PR #3270, opened 2020-01-04
Added the new dark yellow color format

TwistedAsylumMC:
## Introduction
In either 1.13 or 1.14 (Don't know which), a new color code was introduced (``g``), which is dark_yellow

### API changes
Added ``TextFormat::DARK_YELLOW`` constant

## Backwards compatibility
There is no BC breaks in this change

## Tests
```php
use pocketmine\Player;

/** @var Player $player */
$player->sendMessage(TextFormat::DARK_YELLOW . "This message is dark yellow!");
```


SOF3:
Technically, color codes are part of the client protocol. I wouldn't regard this as a BC break for the API, as it only updates protocol information.

dktapps:
as discussed on discord, this does cause potential BC breaks for things which previously used § and g adjacently without expecting them to form a colour code. The protocol excuse doesn't work here since this is documented API used in many areas, and not just for the protocol

TheBlackPlague (approved):
Great! This PR should be merged into **`stable`**.

SOF3:
How does this PR work with the terminal? I don't see ANSI codes for dark yellow being added. It appears that the &sect;g code would be emitted on console directly. Please resolve this issue first.

TwistedAsylumMC:
Can anyone on linux figure out the tput thing to put in ``Terminal::getEscapeCodes()``? I'm on windows and cannot test values.

Encritary:
@TwistedAsylumMC if colors >= 256, ``tput setaf 178`` looks fine

Encritary:
Screenshot_20200113-133123

dktapps:
This may be considered for 3.12 if you can rebase onto `next-minor` and fix the build failures.

TwistedAsylumMC:
Closing this because I screwed it up, will open a new pull request with the fixed changes

PR #3275, opened 2020-01-15
Remove parameter $isAdmin from Player::kick()

Muqsit:
## Introduction

The `$isAdmin` parameter in `Player->kick()` servers not much of a purpose besides further modifying the `$reason` parameter. As by default, the parameter is `true` this could result in a misleading kick message.

Plugins may want to kick players for various other reasons which do not require an admin intervention (eg: anticheat and anti-afk plugins) but the `$isAdmin` parameter is dedicated to serve only the purposes that were resulted by an admin kicking, among the several other purposes.
So the "Kicked by admin." prefix is better handled by the method caller than the method itself.

At the moment, `Player->kick()` is called in 5 places:
```php
BanCommand.php:61:           $player->kick($reason !== "" ? "Banned by admin. Reason: " . $reason : "Banned by admin.");
BanIpCommand.php:84:         $player->kick($reason !== "" ? $reason : "IP banned.");
KickCommand.php:61:          $player->kick($reason);
Player.php:390:              $this->kick("You have been banned");
Player.php:1774:             $this->kick("Attempting to attack an invalid entity");
```
1. The first one, in BanCommand serves no purpose as there's already a "Banned by admin." prefix before the ban reason.
2. The second usage is arguable. Adding a "Banned by admin." over there might be helpful assuming the majority users set a vague ban reason. For example: `/ban-ip player flying` would send the player a disconnect screen with only `flying` displayed. (RFC?)
3. Same as point 2, assuming the majority users `/kick player flying`.
4. Arguable since a ban is most likely the result of an admin intervention. This could be misleading when the server is set to hardcore mode where a player isn't banned as a result of admin intervention.
5. Results in a misleading kick screen as player isn't kicked by an administrator.

### API changes

Removes optional parameter `$isAdmin` from `Player::kick()`.

### Behavioural changes

1. The "Kicked by admin." prefix is no longer automatically added when plugins call `Player::kick()`.
2. `/ban-ip` now adds an "IP banned by admin." prefix before the reason instead of displaying only the reason on the disconnect screen when a reason is supplied.

## Backwards compatibility

Breaks plugins that supplied more than 1 parameter to `Player::kick()`.
Plugins that would like to maintain the previous behaviour may migrate as follows:

| Before | After |
| --- | --- |
| `$player->kick($reason, false);` | `$player->kick($reason);` |
| `$player->kick($reason, false, $quitMessage);` | `$player->kick($reason, $quitMessage);` |
| `$player->kick($reason, true);` | `$player->kick("Kicked by admin." . ($reason !== "" ? " Reason: " . $reason : ""));` |
| `$player->kick($reason, true, $quitMessage);` | `$player->kick("Kicked by admin." . ($reason !== "" ? " Reason: " . $reason : ""), $quitMessage);` |

## Follow-up
Updating changelog

## Tests

Script plugin that tries to kick and ban player by executing commands via console.
https://gist.github.com/Muqsit/645e1e54fc9593416d7fcb5bed29d792

### DIFF
--- a/src/command/defaults/BanIpCommand.php
+++ b/src/command/defaults/BanIpCommand.php
@@ -81,7 +81,7 @@ class BanIpCommand extends VanillaCommand{
 
 		foreach($sender->getServer()->getOnlinePlayers() as $player){
 			if($player->getNetworkSession()->getIp() === $ip){
-				$player->kick($reason !== "" ? $reason : "IP banned.");
+				$player->kick($reason !== "" ? "IP banned by admin. Reason: " . $reason : "IP banned by admin.");
 			}
 		}
 
--- a/src/command/defaults/KickCommand.php
+++ b/src/command/defaults/KickCommand.php
@@ -58,7 +58,7 @@ class KickCommand extends VanillaCommand{
 		$reason = trim(implode(" ", $args));
 
 		if(($player = $sender->getServer()->getPlayer($name)) instanceof Player){
-			$player->kick($reason);
+			$player->kick($reason !== "" ? "Kicked by admin. Reason: " . $reason : "");
 			if($reason !== ""){
 				Command::broadcastCommandMessage($sender, new TranslationContainer("commands.kick.success.reason", [$player->getName(), $reason]));
 			}else{
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -2056,28 +2056,17 @@ class Player extends Human implements CommandSender, ChunkLoader, ChunkListener,
 	 * Kicks a player from the server
 	 *
 	 * @param string               $reason
-	 * @param bool                 $isAdmin
 	 * @param TextContainer|string $quitMessage
 	 *
 	 * @return bool
 	 */
-	public function kick(string $reason = "", bool $isAdmin = true, $quitMessage = null) : bool{
+	public function kick(string $reason = "", $quitMessage = null) : bool{
 		$ev = new PlayerKickEvent($this, $reason, $quitMessage ?? $this->getLeaveMessage());
 		$ev->call();
 		if(!$ev->isCancelled()){
 			$reason = $ev->getReason();
-			$message = $reason;
-			if($isAdmin){
-				if(!$this->isBanned()){
-					$message = "Kicked by admin." . ($reason !== "" ? " Reason: " . $reason : "");
-				}
-			}else{
-				if($reason === ""){
-					$message = "disconnectionScreen.noReason";
-				}
-			}
+			$message = $reason === "" ? "disconnectionScreen.noReason" : $reason;
 			$this->disconnect($message, $ev->getQuitMessage());
-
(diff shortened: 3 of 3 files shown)


SOF3 (approved):
I would prefer merging this after we have consensus on what to do with the `TextContainer|string|null` mess.

dktapps:
I think the `/kick` command should add the `Kicked by admin` prefix explicitly, similar to how `/ban` does it.

Muqsit:
The same goes for BanIpCommand too, right?

dktapps:
disconnect messages should be making sense regardless of API changes

dktapps:
this is now problematic to merge because there are needed changes, but you've deleted the downstream repository.

src/command/defaults/BanIpCommand.php
```diff
@@ -81,7 +81,7 @@ private function processIPBan(string $ip, CommandSender $sender, string $reason)
 
 		foreach($sender->getServer()->getOnlinePlayers() as $player){
 			if($player->getNetworkSession()->getIp() === $ip){
-				$player->kick($reason !== "" ? $reason : "IP banned.");
+				$player->kick($reason !== "" ? "IP banned by admin. Reason: " . $reason : "IP banned by admin.");
```
dktapps: For some reason I'm not able to provide suggested changes, but the following would be better:
```php
$player->kick("IP banned by admin." . ($reason !== "" ? " Reason: " . $reason : ""));
```

src/command/defaults/KickCommand.php
```diff
@@ -58,7 +58,7 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
 		$reason = trim(implode(" ", $args));
 
 		if(($player = $sender->getServer()->getPlayer($name)) instanceof Player){
-			$player->kick($reason);
+			$player->kick($reason !== "" ? "Kicked by admin. Reason: " . $reason : "");
```
dktapps:
```php
$player->kick("Kicked by admin." . ($reason !== "" ? " Reason: " . $reason : ""));
```

dktapps (changes requested):
These lines should also be updated: https://github.com/pmmp/PocketMine-MP/pull/3275/files#diff-6e3ab584fcfae48a1f070adc1de82abdR2080-R2081

dktapps:
> I would prefer merging this after we have consensus on what to do with the TextContainer|string|null mess.

My current plan is to make `TranslationContainer|null` the accepted type, and box raw strings inside a special `raw` translation key. `TextContainer` has been scrapped because it doesn't make any sense (see 4c51f1dda348d9d1c53c89185ceeb909a1cad21a).

lukeeey:
Also rename it to `Translation`

dktapps:
Implemented in 5a33dbd4c6a88e86c2b0b2f4365429e615c26d3a.

Issue #3139, opened 2019-10-05
Server crashes while trying transfer player


Frago9876543210:
### Issue description
The player is transferred to the server, but the current server crashes

### Steps to reproduce the issue
1. Use code below
2. Join to server
3. Spawn some entity
4. Damage it

### OS and versions
* PocketMine-MP: 37763685aa4244535414d77240f29d42a045a490
* PHP: 7.3.9
* Server OS: Ubuntu 18.04

### Plugins
```php
<?php

declare(strict_types=1);

/**
 * @name    Example
 * @main    Example
 * @api     4.0.0
 * @version 1.0.0
 */

use pocketmine\event\entity\EntityDamageByEntityEvent;
use pocketmine\event\Listener;
use pocketmine\player\Player;
use pocketmine\plugin\PluginBase;

class Example extends PluginBase implements Listener{

	protected function onEnable() : void{
		$this->getServer()->getPluginManager()->registerEvents($this, $this);
	}

	public function onDamage(EntityDamageByEntityEvent $e) : void{
		$attacker = $e->getDamager();
		if($attacker instanceof Player){
			$e->setCancelled();
			$attacker->transfer("play.lbsg.net", 19132);
		}
	}
}
```

### Crashdump, backtrace or other files
```[21:38:25.813] [Server thread/INFO]: [NetworkSession: Frago9876543210] Session closed due to transfer
[21:38:25.814] [Server thread/CRITICAL]: TypeError: "Return value of pocketmine\network\mcpe\NetworkSession::getInvManager() must be an instance of pocketmine\network\mcpe\InventoryManager, null returned" (EXCEPTION) in "src/network/mcpe/NetworkSession" at line 810
[21:38:25.814] [Server thread/DEBUG]: #0 src/network/mcpe/handler/InGamePacketHandler(334): pocketmine\network\mcpe\NetworkSession->getInvManager()
[21:38:25.814] [Server thread/DEBUG]: #1 src/network/mcpe/handler/InGamePacketHandler(182): pocketmine\network\mcpe\handler\InGamePacketHandler->handleUseItemOnEntityTransaction(object pocketmine\network\mcpe\protocol\types\inventory\UseItemOnEntityTransactionData)
[...]

PR #3277, opened 2020-01-16
Close NetworkSession later in the kick/disconnect sequence

TwistedAsylumMC:
## Introduction
When a player is kicked, their NetworkSession is closed too early. If a plugin currently tries modifying data related to their NetworkSession (E.g. Their player inventory) in PlayerQuitEvent and the player was kicked, it causes errors and crashes the server.

### Behavioural changes
NetworkSessions close after PlayerQuitEvent and PlayerDataSaveEvent has been called

## Backwards compatibility
This pull request should not cause any BC breaks

### Before this pull request:
 - When a player leaves, their NetworkSession gets closed after PlayerQuitEvent:
image
 - When a player is kicked, their NetworkSession gets closed before PlayerQuitEvent:
image
### After this pull request:
 - When a player leaves, their NetworkSession gets closed after PlayerQuitEvent:
image
 - When a player is kicked, their NetworkSession gets closed after PlayerQuitEvent
image

### Code used for testing:
```php
<?php
declare(strict_types=1);

use pocketmine\event\player\PlayerKickEvent;
use pocketmine\event\player\PlayerQuitEvent;
use var_dump;

class Test extends PluginBase implements Listener{

	}

	public function onPlayerKick(PlayerKickEvent $event) : void{
		var_dump("Player Kick Event");
	}

	public function onPlayerQuit(PlayerQuitEvent $event) : void{
		var_dump("Player Quit Event");
	}
}```

### DIFF
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -2100,8 +2100,6 @@ class Player extends Human implements CommandSender, ChunkLoader, ChunkListener,
 			return;
 		}
 
-		$this->networkSession->onPlayerDestroyed($reason, $notify);
-
 		//prevent the player receiving their own disconnect message
 		PermissionManager::getInstance()->unsubscribeFromPermission(Server::BROADCAST_CHANNEL_USERS, $this);
 		PermissionManager::getInstance()->unsubscribeFromPermission(Server::BROADCAST_CHANNEL_ADMINISTRATIVE, $this);
@@ -2112,6 +2110,8 @@ class Player extends Human implements CommandSender, ChunkLoader, ChunkListener,
 			$this->server->broadcastMessage($ev->getQuitMessage());
 		}
 		$this->save();
+
+		$this->networkSession->onPlayerDestroyed($reason, $notify);
 
 		$this->spawned = false;
 


dktapps:
This is the wrong fix. This will cause the player to receive garbage packets before the connection has been terminated.

TwistedAsylumMC:
What is the right fix then? A sollution to your issue could be something like flagging the session to be closed, so it doesnt receive any more packets but still has its data until its properly closed

dktapps:
The source of the problem is this: https://github.com/pmmp/PocketMine-MP/blob/master/src/network/mcpe/NetworkSession.php#L459

There are a couple of possible solutions to this:
1) Remove this cyclic reference destruction and allow the GC to deal with it. This might increase memory pressure.
2) Replace this with a no-op InventoryManager implementation which doesn't contain a reference to the player or networksession (similarly to how NullPacketHandler is used for this purpose).

dktapps:
relates to #3139

dktapps:
This doesn't actually address the bug at hand and only masks its symptoms, while also creating undesirable side effects. The issues associated with this bug will remain open until the proper fix has been applied.

PR #3288, opened 2020-01-28
add some PHPStan-specific callable prototypes

dktapps:
## Introduction
This pull requests adds prototype information where callables are used. This is useful for static analysers to detect bugs in code both core and plugin.

Since these annotations are not subject to any code styling standard, the style must be manually maintained; therefore, discussion on the rules to apply on their formatting needs to be had.

### Behavioural changes
PHPStan may detect additional flaws in code relating to closure/callable usage.

## Tests
Since this is a PHPStan focused change, tests are not required, since the CI will do the work.


dktapps:
Merged as 2b402e525a3f0015b294961a5100e2c32efb12b2.

PR #3300, opened 2020-02-09
Async worker init closures

JackNoordhuis:
## Introduction
Allows registering a closure that is run on async worker threads. Useful for static registries (blocks, items, etc).

### API changes
* Added `\pocketmine\scheduler\AsyncPool::addWorkerInitHook(\Closure)` method for registering a closure that is run once on all current workers and any new workers.
* Added `\pocketmine\scheduler\AsyncPool::removeWorkerInitHook(\Closure)` method for removing a closure that has previously been registered as an init hook.

## Backwards compatibility
No BC changes, only API additions.

## Tests
https://gist.github.com/JackNoordhuis/936df0444262d10538239c0e985f7d04

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,8 @@
       "pocketmine/classloader": "dev-master",
       "daverandom/callback-validator": "dev-master",
       "adhocore/json-comment": "^0.1.0",
-      "particle/validator": "^2.3"
+      "particle/validator": "^2.3",
+      "opis/closure": "^3.5"
    },
    "require-dev": {
       "phpstan/phpstan": "^0.12.9",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "bb9233cea6e62cc28ea22863681e066e",
+    "content-hash": "a85ce5a659bb2d131f96167be134ace2",
     "packages": [
         {
             "name": "adhocore/json-comment",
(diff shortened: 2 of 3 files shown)


dktapps:
what exactly is the use case?

dktapps:
I don't really see what you can do with this that can't be done with a worker start hook + outside async task, a la world generator registration. This seems redundant to me.

kenygamer:
I was just looking at the title of this issue to be able to answer that this is completely useless. Doing something when the task is handled can be done in your plugin appending code to your task in `AsyncTask::onRun()`, this only adds spare functionality.

As a side note I know this explanation of how working with tasks is unnecessary for you but I'm still making it for other readers.

JackNoordhuis:
@kenygamer this was intended to be used for registering things like blocks and items (unimplemented vanilla + custom), packets, etc. into pocketmines static registries so that you don't need to schedule your own tasks for every thread. It was not intended to be used to replace any part of async tasks.

Although there currently exists other ways to registering these static mappings, this PR isn't completely useless, more superfluous and in any likelihood should hopefully be redundant by the time PM4 is finally released if all the planned changes are implemented.

dktapps:
@kenygamer I think you don't understand the purpose of the PR. The purpose is to have some code automatically run when a new worker is started in the pool. With that said ...

To reiterate what I said on Discord, the PR does not bring any new functionality to the table, because it's possible use a regular worker start hook to schedule an `AsyncTask` to register your custom blocks as needed. This isn't a great solution due to the memory impact, but it's doable.

JackNoordhuis:
Closing due to the same functionality being achievable with the existing API methods and no external libs (https://gist.github.com/JackNoordhuis/2be6bce14de17aee36c794af78299207).

PR #3305, opened 2020-02-11
Fix for player movement on 1.14.30 clients

Erkam246:


Olybear9:
I recommend changing this in the broadcastMovement function because I'm sure the issue is present with walking entities.

jasonwynn10:
Do the two functions not do the same thing?

dktapps:
why closed?

Erkam246:
Its a hack like you said its flickering

PR #3314, opened 2020-02-20
Make it easier to prevent an object from being consumed

### LINKED ISSUE #3306: You can drink buckets
### Issue description
This is actually the problem I had in [this issue](https://github.com/pmmp/PocketMine-MP/issues/3291), I never noticed the bucket actually disappeared.
- Expected result: Not be able to drink buckets.
- Actual result: Drank a bucket.

1. Get a water or lava bucket.
2. Go in survival.
3. Right click air with the bucket.
4. Potion sound effects and bucket is empty.
5. Profit!

### OS and versions

* PocketMine-MP: 3.11.4
* PHP: 7.3.13
* Server OS: Windows 10
* Game version: Win10

### Plugins

No plugins are needed

GiantQuartz:
## Introduction

At the moment, to prevent an item from being consumed the function `Living->consumeObject()` has to be overwritten. This PR encapsulates that part into Consumable, making it easier to prevent an object from being consumed or to identify if it can be consumed.

### Relevant issues

Fixes #3306 

### API changes

Adds `canBeConsumedBy(Living $consumer)` to the interface `Consumable`
```php
	public function canBeConsumedBy(Living $consumer) : bool;
```

## Backwards compatibility

All classes implementing `Consumable` must be declared abstract or add the function `canBeConsumedBy(Living $consumer)`

## Tests

- Potions can be consumed.
- All food sources work as expected.
- Water/Lava buckets cannot be consumed anymore.

[Testing the update](https://www.youtube.com/watch?v=peaw3k31hPg)

### DIFF
--- a/src/pocketmine/block/Cake.php
+++ b/src/pocketmine/block/Cake.php
@@ -24,6 +24,7 @@ declare(strict_types=1);
 namespace pocketmine\block;
 
 use pocketmine\entity\EffectInstance;
+use pocketmine\entity\Human;
 use pocketmine\entity\Living;
 use pocketmine\item\FoodSource;
 use pocketmine\item\Item;
@@ -128,7 +129,19 @@ class Cake extends Transparent implements FoodSource{
 		return [];
 	}
 
+	/**
+	 * This method always receives a Player object
+	 *
+	 * @param Human|Living $consumer
+	 * @return bool
+	 */
+	public function canBeConsumedBy(Living $consumer) : bool{
+		return $consumer->isHungry();
+	}
+
 	public function onConsume(Living $consumer) : void{
+		$consumer->addFood($this->getFoodRestore());
+		$consumer->addSaturation($this->getSaturationRestore());
 		$this->level->setBlock($this, $this->getResidue());
 	}
 }
--- a/src/pocketmine/item/Consumable.php
+++ b/src/pocketmine/item/Consumable.php
@@ -46,6 +46,11 @@ interface Consumable{
 	public function getAdditionalEffects() : array;
 
 	/**
+	 * @return bool
+	 */
+	public function canBeConsumedBy(Living $consumer) : bool;
+
+	/**
 	 * Called when this Consumable is consumed by mob, after standard resulting effects have been applied.
 	 *
 	 * @return void
--- a/src/pocketmine/item/Food.php
+++ b/src/pocketmine/item/Food.php
@@ -23,6 +23,7 @@ declare(strict_types=1);
 
 namespace pocketmine\item;
 
+use pocketmine\entity\Human;
 use pocketmine\entity\Living;
 
 abstract class Food extends Item implements FoodSource{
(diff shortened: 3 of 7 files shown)


Olybear9:
If you're going to do this, add an event and call it so it can be managed by plugins.
Something similar to: 
```php
class EntityConsumeItemEvent extends EntityEvent {
  private $item;

  public function __construct(Entity $ent, Item $item) {
    $this->item = $item;
    $this->entity = $ent;
  }

  public function getItem(): Item {
    return $this->item;
  }
}
```

src/pocketmine/item/Food.php
```diff
@@ -41,7 +42,14 @@ public function getAdditionalEffects() : array{
...
-	public function onConsume(Living $consumer){
+	public function canBeConsumedBy(Living $consumer): bool{
+		return ($this->requiresHunger() and $consumer instanceof Human and !$consumer->isHungry()) ? false : true;
+	}
 
+	public function onConsume(Living $consumer){
+		if($consumer instanceof Human) {
+			$consumer->addFood($this->getFoodRestore());
+			$consumer->addSaturation($this->getSaturationRestore());
+		}
```
Olybear9: This should be an event, in my opinion. However I do see the `canBeConsumedBy()` pretty useless as one, mobs aren't implemented with functionality, two, if they were they wouldn't be triggering this event. (At least to my knowledge). 
I also know that PM is heavily based on the Bukkit api, and this event is not present for entities there either (http://bukkit.luricos.de/api/1.7.9-R0.2/d9/da5/namespaceorg_1_1bukkit_1_1event_1_1entity.html).

src/pocketmine/item/Consumable.php
```diff
@@ -45,6 +45,11 @@ public function getResidue();
 	 */
 	public function getAdditionalEffects() : array;
 
+	/**
+	 * @return bool
+	 */
+	public function canBeConsumedBy(Living $consumer) : bool;
```
dktapps: This breaks BC.

src/pocketmine/block/Cake.php
```diff
@@ -128,7 +129,19 @@ public function getAdditionalEffects() : array{
 		return [];
 	}
 
+	/**
+	 * This method always receives a Player object
+	 *
+	 * @param Human|Living $consumer
```
dktapps: This assumption is flawed.

src/pocketmine/item/Food.php
dktapps: The placement of this logic is now inconsistent with other consume handling. I also don't see why it's necessary to move this.

src/pocketmine/item/Food.php
```diff
@@ -41,7 +42,14 @@ public function getAdditionalEffects() : array{
 		return [];
 	}
 
-	public function onConsume(Living $consumer){
+	public function canBeConsumedBy(Living $consumer): bool{
+		return ($this->requiresHunger() and $consumer instanceof Human and !$consumer->isHungry()) ? false : true;
```
dktapps: nonsensical ternary

src/pocketmine/item/Food.php
```diff
@@ -41,7 +42,14 @@ public function getAdditionalEffects() : array{
 		return [];
 	}
 
-	public function onConsume(Living $consumer){
+	public function canBeConsumedBy(Living $consumer): bool{
```
dktapps: ws issue here.

src/pocketmine/item/Consumable.php
```diff
@@ -45,6 +45,11 @@ public function getResidue();
 	 */
 	public function getAdditionalEffects() : array;
 
+	/**
+	 * @return bool
```
dktapps: This annotation is redundant. This docblock should be removed.

PR #3331, opened 2020-02-28
Protocol: MultiplayerSettingsPacket can be sent to the player

yoraze:
### API changes

Now MultiplayerSettingsPacket can be sent to the player.

### Behavioural changes

No performance changes.

## Backwards compatibility

This PR contains only protocol-related changes.

## Tests
I can approve that MultiplayerSettingsPacket is ClientboundPacket:
Bedrock Dedicated Server (BDS) sends this packet after successful handling MultiplayerSettingsPacket. This is a screenshot from IDA:


dktapps:
Hi, this needed to be applied to stable as well, but in a slightly different form, so the changes have been applied manually. Thanks.

PR #3337, opened 2020-03-03
Pass old item(s) to inventory change listeners (#3323)

### LINKED ISSUE #3323: 4.0.0: Unable to easily identify old item in inventory changes
## Description
Previously on 3.0.0, the old item in the changed slot could be easily obtained through the EntityInventoryChangeEvent. With 4.0.0, the event has been replaced with [inventory change listeners](https://github.com/pmmp/PocketMine-MP/blob/master/src/inventory/InventoryChangeListener.php), which lacks the ability to easily get the old item. 

## Justification
PiggyCE currently has 3 types of custom enchantments: reactive, ticking, & toggling. Toggling enchantments rely on three different events on 3.0.0: PlayerItemHeldEvent for held enchantments, EntityInventoryChangeEvent for inventory enchantments, & EntityArmorChangeEvent for armor enchantments. EntityInventoryChangeEvent, and EntityArmorChangeEvent which extends EntityInventoryChangeEvent, have since been removed, and with them, the ability to easily get the old item. The old items are used to toggle off the enchantments. It's possible to replicate this without having the old item accessible, albeit quite messily. The feature may also benefit other plugin developers when switching to 4.0

DaPigGuy:
## Introduction

This PR solves the issue discussed in issue #3323 where old items can  not being accessed by inventory change listeners

### Relevant issues

* Fixes #3323 

### API changes

* Signatures of InventoryChangeListener::onSlotChange() and InventoryChangeListener::onContentChange() have been changed
* Signatures of callables used by CallbackInventoryChangeListener have also been changed

## Backwards compatibility

This will break plugins currently using inventory change listeners.

## Tests

```php
    public function onJoin(PlayerJoinEvent $event): void
    {
        $player->getInventory()->addChangeListeners(new CallbackInventoryChangeListener(function (Inventory $inventory, int $slot, Item $oldItem) use ($player): void {
            $player->sendMessage("Change in slot " . $slot . " Old Item: " . $oldItem->__toString() . " New Item: " . $inventory->getItem($slot)->__toString());
        }, function (Inventory $inventory, array $oldContents) use ($player): void {
            foreach ($oldContents as $slot => $oldContent) {
                if (!$inventory->getItem($slot)->equals($oldContent)) {
                    $player->sendMessage("[CONTENT CHANGE] Change in slot " . $slot . " Old Item: " . $oldContent->__toString() . " New Item: " . $inventory->getItem($slot)->__toString());
                }
            }
        }));
    }
```

image
Image shows changes caused by dropping an arrow, picking up the arrow, and placing the arrow in its original slot (5)

image
Image shows content changes when dying with an inventory of 1x arrows

### DIFF
--- a/src/inventory/CallbackInventoryChangeListener.php
+++ b/src/inventory/CallbackInventoryChangeListener.php
@@ -75,6 +75,10 @@ class CallbackInventoryChangeListener implements InventoryChangeListener{
 		($this->onSlotChangeCallback)($inventory, $slot, $oldItem);
 	}
 
+	/**
+	 * @param Inventory $inventory
+	 * @param Item[] $oldContents
+	 */
 	public function onContentChange(Inventory $inventory, array $oldContents) : void{
 		($this->onContentChangeCallback)($inventory, $oldContents);
 	}
--- a/src/inventory/CallbackInventoryChangeListener.php
+++ b/src/inventory/CallbackInventoryChangeListener.php
@@ -76,7 +76,6 @@ class CallbackInventoryChangeListener implements InventoryChangeListener{
 	}
 
 	/**
-	 * @param Inventory $inventory
 	 * @param Item[] $oldContents
 	 */
 	public function onContentChange(Inventory $inventory, array $oldContents) : void{
--- a/src/inventory/CallbackInventoryChangeListener.php
+++ b/src/inventory/CallbackInventoryChangeListener.php
@@ -23,6 +23,7 @@ declare(strict_types=1);
 
 namespace pocketmine\inventory;
 
+use pocketmine\item\Item;
 use pocketmine\utils\Utils;
 
 class CallbackInventoryChangeListener implements InventoryChangeListener{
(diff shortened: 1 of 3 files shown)


src/inventory/CallbackInventoryChangeListener.php
```diff
@@ -75,6 +75,10 @@ public function onSlotChange(Inventory $inventory, int $slot, Item $oldItem) : v
 		($this->onSlotChangeCallback)($inventory, $slot, $oldItem);
 	}
 
+	/**
+	 * @param Inventory $inventory
```
JackNoordhuis: the @param tag is only needed on iterable types, this line isn't needed

src/inventory/InventoryChangeListener.php
```diff
@@ -36,5 +36,9 @@ interface InventoryChangeListener{
 
 	public function onSlotChange(Inventory $inventory, int $slot, Item $oldItem) : void;
 
+	/**
+	 * @param Inventory $inventory
```
JackNoordhuis: same as above

src/inventory/CallbackInventoryChangeListener.php
dktapps: more accurately, @param and @return are now only required when the native typehint doesn't fully describe the type.

PR #3344, opened 2020-03-09
Fix #3336

### LINKED ISSUE #3336: Return value of pocketmine\player\Player::getNetworkSession() must be an instance of pocketmine\network\mcpe\NetworkSession, null returned
Link to crashdump: 
https://pastebin.com/wrB1BKZn

I am using pocketmine 4.0.0 from this commit https://github.com/pmmp/PocketMine-MP/commit/2ca10bf84b62838dada79214e3fe17377af84d2c

The issue occurs regularly on the server and I don't know what causes this other than that its when someone leaves the server.

DenielWorld:
## Introduction

An additional check to make sure that the network session exists and is connected before sending the packet. The lack of this check leads to server crashes at times.

### API changes

N/A

### Behavioural changes

N/A

## Backwards compatibility

N/A

## Follow-up

N/A

### DIFF
--- a/src/entity/Entity.php
+++ b/src/entity/Entity.php
@@ -1556,7 +1556,7 @@ abstract class Entity{
 	public function despawnFrom(Player $player, bool $send = true) : void{
 		$id = spl_object_id($player);
 		if(isset($this->hasSpawned[$id])){
-			if($send){
+			if($send and $player->isConnected()){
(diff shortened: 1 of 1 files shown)


dktapps:
This isn't fixing the bug, just removing the symptoms.

dktapps:
Since this PR doesn't actually address any bug and merely masks its symptoms, I'm going to close this. Feel free to pull a real fix if you discover the cause of the bug.

PR #3345, opened 2020-03-09
Stop overriding item in hand if it has changed

### LINKED ISSUE #2010: Using PlayerInventory->setItemInHand() during BlockBreakEvent later gets overridden
the issue there is because pocketmine sets your item in hand to the item you used with an increased item damage. It overrides your setItemInHand.

PocketMine-MP: PocketMine-MP version 1.7dev-698 "[REDACTED]" (API 3.0.0-ALPHA11)
PHP: 7.2
Game version: PE
Tested on a clean server with only the plugin and it still didn't work

GiantQuartz:
## Introduction

When a player in survival mode breaks a block, the item in his hand is immediately overwritten by another one with a different durability. This causes unexpected behavior: if the item the player has in his hand changes during BlockBreakEvent, it will reappear. 
This PR fixes that issue.

### Relevant issues

Fixes #2010

## Changes
Checks if the player's item in hand is different before overriding it.

## Backwards compatibility

This PR should maintain backwards compatibility.

## Tests
I made a plugin to clean the inventory when BlockBreakEvent was called and then checked if #2010 was still occurring, it wasn't.

[Youtube video](https://www.youtube.com/watch?v=oL_XSPNikzc)

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -2420,7 +2420,7 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 
 						if($this->canInteract($blockVector->add(0.5, 0.5, 0.5), $this->isCreative() ? 13 : 7) and $this->level->useBreakOn($blockVector, $item, $this, true)){
 							if($this->isSurvival()){
-								if(!$item->equalsExact($oldItem)){
+								if(!$item->equalsExact($oldItem) and $item->equals($this->inventory->getItemInHand(), false, false)){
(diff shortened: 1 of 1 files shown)


dktapps:
This fix will produce unexpected behaviour if the held item was changed to have different damage, NBT or count.

GiantQuartz:
Thanks for the feedback, I didn't consider that possibility. I fixed it.

dktapps:
There are a few more places where this bug takes effect. Feel free to make further PRs if you'd like to address those too.

PR #3347, opened 2020-03-09
Added the new dark yellow color code

TwistedAsylumMC:
## Introduction
Minecraft: Bedrock Edition v1.13 introduced a new colour code, which is identified by the letter "g". The colour connected to this code is dark yellow.

### API changes
- The constant ``TextFormat::DARK_YELLOW`` has been created with the value ``§g``
- The static property ``Terminal::$COLOR_DARK_YELLOW `` has been created with the following values:
  - Fallback Escape Code: ``\x1b[38;5;226m``
  - Tput colors <= 8: ``tput setaf 3``
  - Tput colors > 8: ``tput setaf 11``
  - Tput colors > 256: ``tput setaf 178``

### Behavioural changes
Any current occurrences of ``§g`` in-game or in a console will start to be displayed as a dark yellow colour

## Tests
```php
use pocketmine\Player;

/** @var Player $player */
$player->sendMessage(TextFormat::DARK_YELLOW . "This message is dark yellow!");
```

### DIFF
--- a/src/pocketmine/utils/Terminal.php
+++ b/src/pocketmine/utils/Terminal.php
@@ -77,6 +77,8 @@ abstract class Terminal{
 	public static $COLOR_YELLOW = "";
 	/** @var string */
 	public static $COLOR_WHITE = "";
+	/** @var string */
+	public static $COLOR_DARK_YELLOW = "";
 
 	/** @var bool|null */
 	private static $formattingCodes = null;
@@ -129,6 +131,7 @@ abstract class Terminal{
 		self::$COLOR_LIGHT_PURPLE = "\x1b[38;5;207m";
 		self::$COLOR_YELLOW = "\x1b[38;5;227m";
 		self::$COLOR_WHITE = "\x1b[38;5;231m";
+		self::$COLOR_DARK_YELLOW = "\x1b[38;5;226m";
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/pocketmine/utils/TextFormat.php
```diff
@@ -387,6 +392,10 @@ public static function toHTML($string) : string{
 					$newString .= "<span style=color:#FFF>";
 					++$tokens;
 					break;
+				case TextFormat::DARK_YELLOW:
+					$newString .= "<span style=color:#DDD605>";
```
dktapps: Is there no 3-digit code that can represent this?

src/pocketmine/utils/Terminal.php
```diff
@@ -129,6 +131,7 @@ protected static function getFallbackEscapeCodes(){
 		self::$COLOR_LIGHT_PURPLE = "\x1b[38;5;207m";
 		self::$COLOR_YELLOW = "\x1b[38;5;227m";
 		self::$COLOR_WHITE = "\x1b[38;5;231m";
+		self::$COLOR_DARK_YELLOW = "\x1b[38;5;226m";
```
dktapps: I'm a little unclear why different codes are used for `tput` escape codes and ANSI escape codes. All the others match except this one, but here we have `226` vs `178`.

dktapps (changes requested):
`Terminal::toANSI()` needs to be updated.

src/pocketmine/utils/TextFormat.php
TwistedAsylumMC: No there is not, using an HTML colour picker you can see that #DDD605 is the correct colour code for it. This was also [confirmed by Damian](https://discordapp.com/channels/373199722573201408/373199722573201410/663031808593494026) when discussing the original pull request on discord.

src/pocketmine/utils/Terminal.php
TwistedAsylumMC: I was not aware that the two were linked in that way, I'll update it now.

dktapps: well both are doing the rounds with 256color so they should be the same.

TwistedAsylumMC: I fixed it in the recent commit

SuperAdam47:
why still we waiting for merge?

SOF3:
This pull request breaks backward compatibility. Originally `&g` doesn't get replaced but now it does. This may break some behaviour.

PJZ9n:
Is it possible to incorporate this change in the 4.0.0 release line?

PR #3356, opened 2020-03-13
When ClickAir and clear your inventory, the item in your hand remains.

Saehing:
This problem refers to https://github.com/pmmp/PocketMine-MP/pull/3345, also related to https://github.com/pmmp/PocketMine-MP/pull/3345 and https://github.com/pmmp/PocketMine-MP/pull/3348

If you clear your inventory while the player is using the item through the air (PlayerItemUseEvent in the pmmp 4.0), the item will remain in your hand

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -2498,7 +2498,7 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 
 						if($item->onClickAir($this, $directionVector)){
 							$this->resetItemCooldown($item);
-							if($this->isSurvival()){
+							if($this->isSurvival() and $item->equalsExact($this->inventory->getItemInHand())){
 								$this->inventory->setItemInHand($item);
 							}
 						}


Saehing:
Should I make a separate PR for pmmp4.0?

Saehing:
> pls re-read this code ... it makes zero sense

Try clearing the player's inventory during PlayerItemUseEvent. ( tested on pmmp 4.0 )

dktapps:
It works because it breaks the code entirely. This code only replaces the held item if the changed item is the same as the held item, which makes the code entirely redundant and breaks existing functionality.

Saehing:
> It works because it breaks the code entirely. This code only replaces the held item if the changed item is the same as the held item, which makes the code entirely redundant and breaks existing functionality.

Your suggestions for this problem?

dktapps:
If I have to spoonfeed you the solution, I might as well apply the change directly myself and save the hassle.

CortexPE:
@Saehing to fix this, you just have to re-send inventory contents "half-a-tick" after you clear the inventory.

or you could even just resend the currently held slot and not the entire inventory if that's the only issue. PM3: `PlayerInventory->sendHeldItem()`

Saehing:
> @Saehing to fix this, you just have to re-send inventory contents "half-a-tick" after you clear the inventory.
> 
> or you could even just resend the currently held slot and not the entire inventory if that's the only issue. PM3: `PlayerInventory->sendHeldItem()`

I do $inventoryManager->syncAll () on PMMP 4.0, and it doesn 't help. 
I noticed that this only happens when the player has a GameMode Spectator, I think this is due to this condition: https://github.com/pmmp/PocketMine-MP/blob/093a7c239ee61c72314d281ef2592f63146ee726/src/player/Player.php#L1433

Saehing:
> @Saehing to fix this, you just have to re-send inventory contents "half-a-tick" after you clear the inventory.
> 
> or you could even just resend the currently held slot and not the entire inventory if that's the only issue. PM3: `PlayerInventory->sendHeldItem()`

I do $inventoryManager->syncAll () on PMMP 4.0, and it doesn 't help.

PR #3357, opened 2020-03-15
Fix RCON socket lingering

CortexPE:
## Introduction
This fixes an issue where RCON is unable to start after being abruptly stopped.

## Tests
1. Start PocketMine-MP server with RCON enabled.
2. Use an RCON client to execute `/stop` on the server.
3. When the server reboots, RCON will be able to bind the socket. Without this PR, RCON will exit with the error below:
image

### DIFF
--- a/src/pocketmine/network/rcon/RCON.php
+++ b/src/pocketmine/network/rcon/RCON.php
@@ -41,6 +41,7 @@ use function socket_getsockname;
 use function socket_last_error;
 use function socket_listen;
 use function socket_set_block;
+use function socket_set_option;
 use function socket_strerror;
 use function socket_write;
 use function trim;
--- a/src/pocketmine/network/rcon/RCON.php
+++ b/src/pocketmine/network/rcon/RCON.php
@@ -46,9 +46,11 @@ use function socket_write;
 use function trim;
 use const AF_INET;
 use const AF_UNIX;
+use const SO_REUSEADDR;
 use const SOCK_STREAM;
 use const SOCKET_ENOPROTOOPT;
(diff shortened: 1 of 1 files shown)


dktapps:
I don't like this (it feels like a hack), but it does seem like this is the most common use case of REUSEADDR.

CortexPE:
The only other way to fix this is to make the server restart delay longer, or at least delay the restart of the RCON instance itself to be able to use the dead socket's address again.

Setting this flag is the simplest least invasive way to fix the TCP socket lingering issue without affecting the current 5 second restart delay, while ensuring that RCON is reliable enough because its startup is not delayed by an arbitrary TCP `TIME_WAIT` period.

dktapps:
In case of actual forcible termination, I might agree. But in the case you mentioned in the OP, it's not forcible termination and there's enough time to gracefully halt RCON and kick out clients. So, I think there is a bug here that needs to be addressed properly before we think about merging this change.

CortexPE:
In that case, checking if the server is not running anymore or shutting down... Right after a command is ran, then breaking the RCON server thread loop would be the solution.

dktapps:
It should and does, which is why I don't understand where this bug is coming from.

dktapps:
I have a feeling this doesn't have anything to do with a remote client sending /stop at all. My feeling is that the problem actually stems from having clients connected during shutdown in the first place.

dktapps:
pls add an import for `socket_set_option` and this will be good to go.

dktapps:
this also needs to be applied to the `RconServer` plugin if you want to do that.

PR #3366, opened 2020-03-19
fix for projectile items like snowball

Eren5960:
## Introduction
Projectiles such as snowball could not be thrown in series

### Relevant issues
* Fixed  useHeldItem in Player.php

### API changes
no

### Behavioural changes
no

## Backwards compatibility
no

-->

## Tests
throw the snowball

### DIFF
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -74,6 +74,7 @@ use pocketmine\item\enchantment\EnchantmentInstance;
 use pocketmine\item\enchantment\MeleeWeaponEnchantment;
 use pocketmine\item\Item;
 use pocketmine\item\ItemUseResult;
+use pocketmine\item\ProjectileItem;
 use pocketmine\lang\TextContainer;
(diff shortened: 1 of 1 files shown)


dktapps:
this fix is incorrect. If it was that simple, why do you think the change would not have already been applied?

this change breaks bows and any other item that has a release action.

Eren5960:
You are right. I just realized this. I will try other ways

Eren5960:
I tested with:
- Bow
- Shovel
- Snowball
- Ender Pearl

and working successfully

dktapps:
this is still wrong. A bow is not a projectile, and projectiles don't have release actions. In addition, this breaks eating.

dktapps:
never mind, i misread your code. Nonetheless, this flag should not be set if there isn't a release action and most items don't have that, hence the TODO.

dktapps:
Fixed by #3664 .

PR #3367, opened 2020-03-19
New bubbles colors

YoungFad:
## Introduction

Just a new feature
### API changes

No

### DIFF
--- a/src/pocketmine/command/defaults/ParticleCommand.php
+++ b/src/pocketmine/command/defaults/ParticleCommand.php
@@ -223,7 +223,7 @@ class ParticleCommand extends VanillaCommand{
 		}elseif(strpos($name, "instantspell") === 0){
 			$d = explode("_", $name);
 			if(count($d) === 3){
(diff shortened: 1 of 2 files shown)


YoungFad:
Actually it's not water bubbles, it's just the potions bubble colors.

dktapps:
this breaks backwards compatibility.

lukeeey:
its still *technically* backwards incompatible as plugins made for the latest stable build won’t work on the older ones

dktapps:
yes this is what we have minor versions for.

dktapps:
I think it would be better if this used the Color class instead of a bunch of integers.

dktapps:
now we are back to a backwards compatibility break.

YoungFad:
No longer backwards compatibility break.

dktapps:
I have no idea what you did in the last few commits but it's definitely not what I expected.

YoungFad:
I'll close the PR and create new one to show the final result of it.

dktapps:
I can see the final results of it just fine. You've made a mess in the last few commits.

Issue #3216, opened 2019-12-08
Implement native UPnP support


dktapps:
Currently UPnP forwarding is only supported on Windows because we use .NET bindings to use C# API for forwarding (com_dotnet). It's possible to implement forwarding natively in PHP without .NET, by introducing dependency on XML extensions (which are now provided with our PHP builds by default).

HimbeersaftLP:
> Currently UPnP forwarding is only supported on Linux

I think you wanted to say "Windows"?

dktapps:
thanks, brain bug caught me out yet again...

PR #3378, opened 2020-03-29
Implement UPnP support without dotNET 

Yosshi999:
## Introduction
Implement UPnP forwarding by using socket and XML module instead of .NET bindings.

### Relevant issues
Solve #3216

## Backwards compatibility
.NET is no longer used for UPnP

## Tests
Tested on WSL (Ubuntu 18.04.2 LTS) and a user was able to connect to the server from external.


src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,195 @@
...
+		$location = trim($matches[1]);
+		$url = parse_url($location);
+		$response = Internet::getURL($location, 3, [], $err, $headers, $httpCode);
+		if(!$response){
+			throw new \RuntimeException("Unable to access XML: {$err}");
+		}
+		if($httpCode !== 200){
+			throw new \RuntimeException("Unable to access XML: {$response}");
+		}
+		$root = new \SimpleXMLElement($response);
```
dktapps: A dependency on `ext-simplexml` needs to be added to `composer.json` for this.

dktapps:
nice, this looks interesting! I can't test this myself currently, but maybe we could get tester feedback from MacOS and Windows users.

I'll review this in closer detail later.

matcracker (commented):
Question: why the function names are not camelCase?

dktapps:
blame @shoghicp

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -138,22 +138,22 @@ public static function Setup() : void{
...
+		|| !self::getChildDevice($device, "urn:schemas-upnp-org:device:WANConnectionDevice:1", $device)
+		|| !self::getChildService($device, "urn:schemas-upnp-org:service:WANIPConnection:1", $service)){
 			throw new \RuntimeException("Your router does not support portforwarding");
 		}
 		$controlURL = (string)$service->controlURL;
 		self::$serviceURL = "{$url['host']}:{$url['port']}/{$controlURL}";
 	}
 
-	public static function PortForward(int $port) : void{
+	public static function portForward(int $port) : void{
```
dktapps: This is BC-breaking.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -182,7 +182,7 @@ public static function PortForward(int $port) : void{
 		}
 	}
 
-	public static function RemovePortForward(int $port) : bool{
+	public static function removePortForward(int $port) : bool{
```
dktapps: This is also BC-breaking.

dktapps (changes requested):

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,195 @@
...
 use function is_object;
+use const AF_INET;
+use const SOCK_DGRAM;
+use const SOL_UDP;
+use const IPPROTO_IP;
+use const IP_MULTICAST_IF;
+use const IP_MULTICAST_LOOP;
+use const IP_MULTICAST_TTL;
+use const SOL_SOCKET;
+use const SO_RCVTIMEO;
```
SOF3, line 70: Add the corresponding function and class imports too

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+		socket_sendto($socket, $contents, strlen($contents), 0, "239.255.255.250", 1900);
+		socket_recvfrom($socket, $buffer, 1024, 0, $responseHost, $responsePort);
+
+		if(!$buffer || !preg_match('/location\s*:\s*(.+)\n/i', $buffer, $matches)){
+			throw new \RuntimeException("Unable to find the router. Ensure that network discovery is enabled in Control Panel.");
+		}
+		$location = trim($matches[1]);
+		$url = parse_url($location);
+		$response = Internet::getURL($location, 3, [], $err, $headers, $httpCode);
+		if(!$response){
```
dktapps, line 138: Prefer `response === false` check here.

src/pocketmine/network/upnp/UPnP.php
dktapps: This is going to spew a bunch of warnings and an `Exception` if the data isn't valid, which might crash the server. We want a nice error to be logged, we don't want a crash.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+			throw new \RuntimeException("Unable to access XML: {$err}");
+		}
+		if($httpCode !== 200){
+			throw new \RuntimeException("Unable to access XML: {$response}");
+		}
+		$root = new \SimpleXMLElement($response);
+		$device = $root->device;
```
dktapps, line 145: This might not exist, or it might be of the wrong type.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+		if(!((string)$device->deviceType === "urn:schemas-upnp-org:device:InternetGatewayDevice:1")
+		|| !self::getChildDevice($device, "urn:schemas-upnp-org:device:WANDevice:1", $device)
+		|| !self::getChildDevice($device, "urn:schemas-upnp-org:device:WANConnectionDevice:1", $device)
+		|| !self::getChildService($device, "urn:schemas-upnp-org:service:WANIPConnection:1", $service)){
+			throw new \RuntimeException("Your router does not support portforwarding");
+		}
+		$controlURL = (string)$service->controlURL;
```
dktapps, line 152: This might not exist, or it might be of the wrong type.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+use const IPPROTO_IP;
+use const IP_MULTICAST_IF;
+use const IP_MULTICAST_LOOP;
+use const IP_MULTICAST_TTL;
+use const SOL_SOCKET;
+use const SO_RCVTIMEO;
 
 abstract class UPnP{
+	/** @var string */
+	private static $serviceURL;
```
dktapps, line 80: This should be `string|null` with a default of `null`.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+	 * Search the device node with given type.
+	 *
+	 * @param \SimpleXMLElement $device
+	 * @param string $deviceType
+	 * @param \SimpleXMLElement $foundDevice reference parameter
+	 *
+	 * @return bool
+	 */
+	private static function getChildDevice(\SimpleXMLElement $device, string $deviceType, &$foundDevice) : bool{
+		foreach($device->deviceList->device as $child){
```
dktapps, line 92: Again, no guarantee that `deviceList` will be of the correct type, or that it'll have a `device` member.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+			if((string)$child->deviceType === $deviceType){
```
dktapps, line 93: No guarantee of presence or type here.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+	 * Search the service node with given type.
+	 *
+	 * @param \SimpleXMLElement $device
+	 * @param string $serviceType
+	 * @param \SimpleXMLElement $foundService reference parameter
+	 *
+	 * @return bool
+	 */
+	private static function getChildService(\SimpleXMLElement $device, string $serviceType, &$foundService) : bool{
+		foreach($device->serviceList->service as $child){
```
dktapps, line 111: Same as above - no guarantee of type or field presence.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+			if((string)$child->serviceType === $serviceType){
```
dktapps, line 112: Same as above.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+			}
+		}
+		return false;
+	}
+	
+	public static function setup() : void{
+		$socket = socket_create(AF_INET, SOCK_DGRAM, SOL_UDP);
+		socket_set_option($socket, SOL_SOCKET, SO_RCVTIMEO, array("sec"=>3,"usec"=>0));
+		$contents =
+			"M-SEARCH * HTTP/1.1\r\n".
```
dktapps, line 124: Formatting - there should be whitespace between the string and the concat operator.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+			"MX: 2\r\n".
```
dktapps, line 125: as above

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+			"HOST: 239.255.255.250:1900\r\n".
```
dktapps, line 126: as above

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+			"MAN: \"ssdp:discover\"\r\n".
```
dktapps, line 127: as above.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+				$foundService = $child;
+				return true;
+			}
+		}
+		return false;
+	}
+	
+	public static function setup() : void{
+		$socket = socket_create(AF_INET, SOCK_DGRAM, SOL_UDP);
+		socket_set_option($socket, SOL_SOCKET, SO_RCVTIMEO, array("sec"=>3,"usec"=>0));
```
dktapps, line 122: prefer the short array syntax `[]`, and also please fix the whitespace.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+		$contents =
+			"M-SEARCH * HTTP/1.1\r\n".
+			"MX: 2\r\n".
+			"HOST: 239.255.255.250:1900\r\n".
+			"MAN: \"ssdp:discover\"\r\n".
+			"ST: upnp:rootdevice\r\n\r\n";
+		socket_sendto($socket, $contents, strlen($contents), 0, "239.255.255.250", 1900);
+		socket_recvfrom($socket, $buffer, 1024, 0, $responseHost, $responsePort);
+
+		if(!$buffer || !preg_match('/location\s*:\s*(.+)\n/i', $buffer, $matches)){
```
dktapps, line 132: prefer an `=== false` check here.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+	public static function setup() : void{
+		$socket = socket_create(AF_INET, SOCK_DGRAM, SOL_UDP);
+		socket_set_option($socket, SOL_SOCKET, SO_RCVTIMEO, array("sec"=>3,"usec"=>0));
+		$contents =
+			"M-SEARCH * HTTP/1.1\r\n".
+			"MX: 2\r\n".
+			"HOST: 239.255.255.250:1900\r\n".
+			"MAN: \"ssdp:discover\"\r\n".
+			"ST: upnp:rootdevice\r\n\r\n";
+		socket_sendto($socket, $contents, strlen($contents), 0, "239.255.255.250", 1900);
```
dktapps, line 129: This might emit E_WARNING, which will turn into `ErrorException` that will crash the server.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+		socket_recvfrom($socket, $buffer, 1024, 0, $responseHost, $responsePort);
```
dktapps, line 130: same here

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+
+		if(!$buffer || !preg_match('/location\s*:\s*(.+)\n/i', $buffer, $matches)){
+			throw new \RuntimeException("Unable to find the router. Ensure that network discovery is enabled in Control Panel.");
+		}
+		$location = trim($matches[1]);
+		$url = parse_url($location);
```
dktapps, line 136: `parse_url()` may return false if the match is invalid

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+		$response = Internet::getURL($location, 3, [], $err, $headers, $httpCode);
+		if(!$response){
+			throw new \RuntimeException("Unable to access XML: {$err}");
+		}
+		if($httpCode !== 200){
+			throw new \RuntimeException("Unable to access XML: {$response}");
+		}
+		$root = new \SimpleXMLElement($response);
+		$device = $root->device;
+		if(!((string)$device->deviceType === "urn:schemas-upnp-org:device:InternetGatewayDevice:1")
```
dktapps, line 146: This might not exist, or it might have the wrong type.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+		foreach($device->serviceList->service as $child){
+			if((string)$child->serviceType === $serviceType){
+				$foundService = $child;
+				return true;
+			}
+		}
+		return false;
+	}
+	
+	public static function setup() : void{
```
dktapps, line 120: I think it would make more sense if this was named `getControlUrl()` and returned `string`. Then have it use the static field as a cache.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+			'<u:AddPortMapping xmlns:u="urn:schemas-upnp-org:service:WANIPConnection:1">'.
+				'<NewRemoteHost></NewRemoteHost>'.
+				'<NewExternalPort>' . $port . '</NewExternalPort>'.
+				'<NewProtocol>UDP</NewProtocol>'.
+				'<NewInternalPort>' . $port . '</NewInternalPort>'.
+				'<NewInternalClient>' . Internet::getInternalIP() . '</NewInternalClient>'.
+				'<NewEnabled>1</NewEnabled>'.
+				'<NewPortMappingDescription>PocketMine-MP</NewPortMappingDescription>'.
+				'<NewLeaseDuration>0</NewLeaseDuration>'.
+			'</u:AddPortMapping>';
```
dktapps, line 174: Formatting here needs whitespace between strings and concat operators.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
 
-		$myLocalIP = Internet::getInternalIP();
+		$contents =
+			'<?xml version="1.0"?>'.
+			'<s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/" s:encodingStyle="http://schemas.xmlsoap.org/soap/encoding/">'.
```
dktapps, line 178: Needs space between string and concat operators (this line and the one above).

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
-			}
-			/** @noinspection PhpUndefinedFieldInspection */
-			$com->StaticPortMappingCollection->Remove($port, "UDP");
-		}catch(\Throwable $e){
+		$body =
+			'<u:DeletePortMapping xmlns:u="urn:schemas-upnp-org:service:WANIPConnection:1">'.
+				'<NewRemoteHost></NewRemoteHost>'.
+				'<NewExternalPort>' . $port . '</NewExternalPort>'.
+				'<NewProtocol>UDP</NewProtocol>'.
+			'</u:DeletePortMapping>';
```
dktapps, line 204: whitespace

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,201 @@
...
+
+		$contents =
+			'<?xml version="1.0"?>'.
+			'<s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/" s:encodingStyle="http://schemas.xmlsoap.org/soap/encoding/">'.
```
dktapps, line 208: whitespace

dktapps (changes requested):
I've left a few comments, mostly about validation and formatting.

src/pocketmine/network/upnp/UPnP.php
Yosshi999: So should I check the existence of `device` by using `isset()`?

dktapps: Yes, and the type.

Yosshi999: I found `SimpleXMLElement::xpath`, which enable us to explore inside a XML without executing null check or type check in each XML node (All we need will be just checking `empty($root->xpath(...))` is true or not, I think.)
I'm now rewriting the code by using this xpath so it will take a little longer to commit.

jasonwynn10: `empty()` has some issues with phpstan. It's advised to use `is_null()` where applicable for more accurate analysis.

dktapps: `is_null()` is a useless function, and `isset()` has the same problems as `empty()`, so I'm a little confused what the point is here.

dktapps: To be clear, the reason `empty()` is banned by PHPStan is because it disguises an `isset() and != null` check behind a misleading name.

Yosshi999: I mean, since xpath() returns false if some error occurs and returns array(0) if nothing is found, I tried to use empty() to check these two cases.
Now I'm checking them by `$xpathResult === false` and `count($xpathResult) === 0`

dktapps: Yes, don't use weak type checks. That goes for `!$result` on functions that return `something|false` as well.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,183 @@
...
+			$contents =
+				"M-SEARCH * HTTP/1.1\r\n" .
+				"MX: 2\r\n" .
+				"HOST: 239.255.255.250:1900\r\n" .
+				"MAN: \"ssdp:discover\"\r\n" .
+				"ST: upnp:rootdevice\r\n\r\n";
+			socket_sendto($socket, $contents, strlen($contents), 0, "239.255.255.250", 1900);
+			socket_recvfrom($socket, $buffer, 1024, 0, $responseHost, $responsePort);
+			socket_close($socket);
+		}catch(\Exception $e){
```
dktapps, line 95: Not a fan of global `Exception` caught here.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,183 @@
...
+			'/deviceList/device[deviceType="urn:schemas-upnp-org:device:WANDevice:1"]' .
+			'/deviceList/device[deviceType="urn:schemas-upnp-org:device:WANConnectionDevice:1"]' .
+			'/serviceList/service[serviceType="urn:schemas-upnp-org:service:WANIPConnection:1"]' .
+			'/controlURL'
+		);
+		if($xpathResult === false || count($xpathResult) === 0){
+			// if result is false or [], i.e. error or notfound
+			throw new \RuntimeException("Your router does not support portforwarding");
+		}
+		$controlURL = (string)($xpathResult[0]);
```
dktapps, line 133: is this guaranteed to be a string at the type level?

src/pocketmine/network/upnp/UPnP.php
dktapps: please provide a `null` default, it'll be needed for PHP 7.4.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,183 @@
...
 		if(!Internet::$online){
 			throw new \RuntimeException("Server is offline");
 		}
-		if(Utils::getOS() !== "win"){
-			throw new \RuntimeException("UPnP is only supported on Windows");
-		}
-		if(!class_exists("COM")){
-			throw new \RuntimeException("UPnP requires the com_dotnet extension");
+
+		if(!isset(self::$serviceURL)){
```
dktapps, line 143: please use a `!== null` check here, `isset()` should be avoided where possible.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,183 @@
...
-		/** @noinspection PhpUndefinedFieldInspection */
-		$com->StaticPortMappingCollection->Add($port, "UDP", $port, $myLocalIP, true, "PocketMine-MP");
 	}
 
 	public static function RemovePortForward(int $port) : bool{
 		if(!Internet::$online){
 			return false;
 		}
-		if(Utils::getOS() != "win" or !class_exists("COM")){
+		if(!isset(self::$serviceURL)){
```
dktapps, line 177: null-check here please

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,183 @@
...
+			'/deviceList/device[deviceType="urn:schemas-upnp-org:device:WANConnectionDevice:1"]' .
+			'/serviceList/service[serviceType="urn:schemas-upnp-org:service:WANIPConnection:1"]' .
+			'/controlURL'
+		);
+		if($xpathResult === false || count($xpathResult) === 0){
+			// if result is false or [], i.e. error or notfound
+			throw new \RuntimeException("Your router does not support portforwarding");
+		}
+		$controlURL = (string)($xpathResult[0]);
+		$serviceURL = "{$url['host']}:{$url['port']}/{$controlURL}";
```
dktapps, line 134: I'd prefer if this used concat operators or even `sprintf()`, this is difficult to read

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,183 @@
...
+			throw new \RuntimeException("Unable to access XML: {$err}");
+		}
+		if($httpCode !== 200){
+			throw new \RuntimeException("Unable to access XML: {$response}");
+		}
+
+		$defaultInternalError = libxml_use_internal_errors(true);
+		try{
+			$root = new \SimpleXMLElement($response);
+		}catch(\Exception $e){
```
dktapps, line 117: please check a more specific exception if possible

dktapps (dismissed):

src/pocketmine/network/upnp/UPnP.php
Yosshi999: I need to put `set_error_handler([Utils::class, 'errorExceptionHandler']);` before try and catch `ErrorException` and put `restore_error_handler();` in finally, right?

dktapps: No, it's already set by default. That said, it's better to use @ and check for false returns, because static analysers can't make sense of error handlers.

src/pocketmine/network/upnp/UPnP.php
Yosshi999: xpath() returns an array of SimpleXMLElement, and SimpleXMLElement has __toString() so I think casting to string always works.

src/pocketmine/network/upnp/UPnP.php
Yosshi999: `get_class($e)` returned 'Exception', so I think it raises Exception itself.

dktapps: That's fair enough. I just get uncomfortable seeing generalized catches like this.

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -131,7 +137,7 @@ public static function getControlUrl() : string{
 			throw new \RuntimeException("Your router does not support portforwarding");
 		}
 		$controlURL = (string)($xpathResult[0]);
```
dktapps, line 139:
```suggestion
		$controlURL = (string) $xpathResult[0];
```

Yosshi999:
Unfortunately my network environment has been changed and I can't access the UPnP setting of my router, so I can't test the current implementation. I'd like someone to help for testing this.
After making sure that it works, I will squash the commits since they are messy.

dktapps:
don't worry too much about squashing, I can squash merge :slight_smile:

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,232 @@
...
+		if($sendbyte !== strlen($contents)){
+			throw new \RuntimeException("Socket error: Unable to send the entire contents.");
+		}
+		if(@socket_recvfrom($socket, $buffer, 1024, 0, $responseHost, $responsePort) === false){
+			throw new \RuntimeException("Socket error: " . trim(socket_strerror(socket_last_error($socket))));
+		}
+		socket_close($socket);
+		$pregResult = preg_match('/location\s*:\s*(.+)\n/i', $buffer, $matches);
+		if($pregResult === false){
+			// you can use preg_last_error_msg() since PHP 8.
```
dktapps, line 110:
```suggestion
			//TODO: replace with preg_last_error_msg() in PHP 8.
```

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,232 @@
...
+				case PREG_BAD_UTF8_OFFSET_ERROR:
+					$errormsg = "bad utf8 offset error";
+					break;
+				case PREG_JIT_STACKLIMIT_ERROR:
+					$errormsg = "jit stacklimit error";
+					break;
+				default:
+					$errormsg = "unknown error";
+			}
+			throw new \RuntimeException("Preg error: " . $errormsg);
```
dktapps, line 137: This would take much less space if you do something like this: https://github.com/pmmp/PocketMine-MP/blob/15401d740f11f3a543abaecc7a449bcba3c2fc91/src/pocketmine/utils/TextFormat.php#L77-L88

src/pocketmine/network/upnp/UPnP.php
```diff
@@ -19,63 +19,232 @@
...
+		
+		$xpathResult = $root->xpath(
+			'//device[deviceType="urn:schemas-upnp-org:device:InternetGatewayDevice:1"]' .
+			'/deviceList/device[deviceType="urn:schemas-upnp-org:device:WANDevice:1"]' .
+			'/deviceList/device[deviceType="urn:schemas-upnp-org:device:WANConnectionDevice:1"]' .
+			'/serviceList/service[serviceType="urn:schemas-upnp-org:service:WANIPConnection:1"]' .
+			'/controlURL'
+		);
+		if($xpathResult === false || count($xpathResult) === 0){
+			// if result is false or [], i.e. error or notfound
```
dktapps, line 179:
```suggestion
```

composer.json
```diff
@@ -23,6 +23,7 @@
       "ext-yaml": ">=2.0.0",
       "ext-zip": "*",
       "ext-zlib": ">=1.2.11",
+      "ext-simplexml": "*",
```
dktapps, line 26: The extensions should be alphabetically ordered.

dktapps (commented):
I can't test this right now due to network configuration.

dktapps:
This appears to be completely dysfunctional. The `xpath()` call is not able to find even basic elements like `//device`, despite them obviously being present.

dktapps:
This is now working correctly on my Hub 3.0.

dktapps (approved):

PR #3385, opened 2020-04-01
Change Level::TIME_ constants to match client

inxomnyaa:
## Introduction

The constants in `Level.php` do not match the clients representation of the vanilla `/time set` command arguments

In Minecraft Java 1.7 there was a change in values
- Day from 0 to 1000
- Night from 12500 to 13000
See: https://minecraft.gamepedia.com/Commands/time and https://minecraft.gamepedia.com/Java_Edition_1.7

### API changes

- Added new Level::TIME_ constants
- Match Level::TIME_ constants with client values

## Backwards compatibility

- Probably, because i.e. `TIME_DAY` was changed from 0 to 1000, some plugins might depend on the value `0`

## Tests

- Ran PMMP from source
- Tried accessing the new constants
- Returned proper time
- Tried locking level time to MIDNIGHT and NOON, functions well.
- Watched the sun. The sun angle now matches the client for i.e. TIME_DAY (world already bright, no more sunrise)

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -128,9 +128,11 @@ class Level implements ChunkManager, Metadatable{
 	public const Y_MASK = 0xFF;
 	public const Y_MAX = 0x100; //256
 
-	public const TIME_DAY = 0;
+	public const TIME_DAY = 1000;
+	public const TIME_NOON = 6000;
 	public const TIME_SUNSET = 12000;
-	public const TIME_NIGHT = 14000;
+	public const TIME_NIGHT = 13000;
+	public const TIME_MIDNIGHT = 18000;
 	public const TIME_SUNRISE = 23000;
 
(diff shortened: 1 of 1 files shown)


dktapps:
thank you travis, don't drop the ball next time ...

dktapps (approved):

dktapps:
There's an additional PR possible if someone wants to update `/time` to add additional aliases.

thebigsmileXD:
> There's an additional PR possible if someone wants to update `/time` to add additional aliases.

I'd do it, but i would prefer the time command to become more similar to the vanilla syntax instead.
Delay syntax change to another time (PM4 update?) and only add aliases now?

dktapps:
I'm not doing the change, I'm just the gatekeeper 😛

thebigsmileXD:
I was trying to say that i would create that PR :P

dktapps:
Sure.

PR #3386, opened 2020-04-01
beatiful api?

Eren5960:
## Introduction

This pull request was opened so that we could get the crops ages. I need this. Maybe others will need it too.

### API changes

added `Crops->getAge()`

### DIFF
--- a/src/block/Crops.php
+++ b/src/block/Crops.php
@@ -45,6 +45,10 @@ abstract class Crops extends Flowable{
 		return $this->age;
 	}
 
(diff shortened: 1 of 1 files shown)


Eren5960:
or we can make `writeStateToMeta()` public instead of protected

dktapps:
or you can use `getMeta()` which is a specifically designated API method ...

Eren5960:
I asked in the Discord server. When no one answered, I thought there was no

Issue #3390, opened 2020-04-02
Cobblestone generators breaks down after a while


Eren5960:
### Issue description
- Cobblestone generators breaks down after a while

### Steps to reproduce the issue
1. Make a cobblestone generator [Like](https://cdn.discordapp.com/attachments/480650036972224513/695299077734137937/unknown.png)
2. Start to break it ... after [see](https://cdn.discordapp.com/attachments/480650036972224513/695299579481948220/unknown.png)

### OS and versions

* PocketMine-MP: ` 4.0.0+dev for Minecraft: Bedrock Edition v1.14.0 (protocol version 389`
* PHP: 7.3
* Server OS: Windows 10
* Game version: PE & Win10 

### Plugins

no plugins

junxiesky:
Did you install the red stone?

IceCruelStuff:
> Did you install the red stone?

Cobblestone generators don't need redstone

junxiesky:
> > Did you install the red stone?
> 
> Cobblestone generators don't need redstone

This Redstone plugin will cause the cobblestone generator to not work 
https://poggit.pmmp.io/ci/tedo0627/RedstoneCircuit_PMMP-Plugin/RedstoneCircuit

junxiesky:
> > Did you install the red stone?
> 
> Cobblestone generators don't need redstone

This Redstone plug-in will cause the cobblestone generator to stop working after working for a while 
https://poggit.pmmp.io/ci/tedo0627/RedstoneCircuit_PMMP-Plugin/RedstoneCircuit

IceCruelStuff:
> > > Did you install the red stone?
> > 
> > 
> > Cobblestone generators don't need redstone
> 
> This Redstone plug-in will cause the cobblestone generator to stop working after working for a while
> https://poggit.pmmp.io/ci/tedo0627/RedstoneCircuit_PMMP-Plugin/RedstoneCircuit
It says no plugins

junxiesky:
> > > > 您安装了红色石头吗？
> > > 
> > > 
> > > 鹅卵石发生器不需要红石
> > 
> > 
> > 此Redstone插件会在工作一段时间后使鹅卵石生成器停止工作
> > https://poggit.pmmp.io/ci/tedo0627/RedstoneCircuit_PMMP-Plugin/RedstoneCircuit
> 
> 它说没有插件
> 
> > 没有插件

oh....okay

PR #3392, opened 2020-04-05
PM4 Lava fix

Eren5960:
## Introduction
https://github.com/pmmp/PocketMine-MP/issues/3390

### DIFF
--- a/src/block/Lava.php
+++ b/src/block/Lava.php
@@ -77,7 +77,7 @@ class Lava extends Liquid{
 	}
 
 	protected function flowIntoBlock(Block $block, int $newFlowDecay, bool $falling) : void{
-		if($block instanceof Water){
+		if($block instanceof Water && ($falling || $block->isSource())){
 			$block->liquidCollide($this, VanillaBlocks::STONE());
 		}else{
 			parent::flowIntoBlock($block, $newFlowDecay, $falling);


dktapps:
This fix is more of a duct tape than an actual fix, but don't worry, I figured out the real bug.

This appears to be a rare (1/30) race condition which appears only on PM4 due to a change in the way local blockupdates are handled on placement.

When liquid flows into a nearby block, a nearby block update is scheduled on that location. That local block update unconditionally schedules a delayed block update (which causes liquid flow, amongst other things), regardless of whether the block hardened or not. On PM3, this isn't a problem, because the local block update happens synchronously with the setBlock call, so the lava never exists for long enough for the bug to occur on PM3.
However, on PM4, local block updates do not happen in sync with setBlock(), but rather are deferred to the next tick, to avoid block update duplications. Since the delayed scheduled updates are processed before the local block update queue, this means that if there was a ghost update still persisting from a previous lava block that flowed into that space (you'd need to break the cobblestone with very precise timing), then the scheduled update would take place before the local update that hardens the lava, causing it to hit the water before the water hits the lava.

dktapps:
Fixed by the referenced commit. While your PR was not merged, it's nonetheless been helpful towards fixing the bug, so thank you for the effort 🙂

PR #3403, opened 2020-04-15
Fix player interaction when sneaking.

### LINKED ISSUE #1904: buckets can't be used while sneaking

### LINKED ISSUE #2539: Paintings cannot be placed while sneaking
### Issue description
Title describes. Same cause as #1904 .

### OS and versions
* PocketMine-MP: 3.3.3
* PHP: 7.2.9
* Server OS: windows
* Game version: Win10

### LINKED ISSUE #3401: Crouched Players Can't Place Certain Items
### Issue description
When players are crouched, they can't interact with certain items. For example, using fence gates, doors, placing water, and picking up water.
- Expected result: We are expecting it to be able to open and close these doors, and place/retrieve liquid with buckets.
- Actual result: Expected action gets reverted.

1. Crouch and stay crouched,
2. Attempt to interact with doors and liquid items.

### OS and versions

* PocketMine-MP: 3.11.7
* PHP: PHP 7.2.24-0ubuntu0.18.04.3 (cli) (built: Feb 11 2020 15:55:52) ( NTS )
* Server OS: ubuntu18.04
* Game version: PE/Win10 

### Plugins

None.
Yes.

95CivicSi:
## Introduction

This changes behavior of block interactions and item usage to be more like vanilla behavior when the player is crouched/sneaking.  Things such as using buckets, turning items in item frames, placing paintings, and opening/closing doors should now work properly despite the player's stance.

### Relevant issues

* Fixes #3401 (Crouched players can't place certain items)
* Fixes #1904 (Buckets can't be used when sneaking)
* Fixes #2539 (Paintings cannot be placed while sneaking)

### API changes

This does not change API.

### Behavioural changes

No noticeable performance or behavioral impact was observed other than fixing intended behavior.
## Backwards compatibility

This does not impact backwards compatibility.
## Tests

Tested interaction with trapdoors, water buckets, paintings, and item frame. 
https://youtu.be/m8p-KXOfauw

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -1862,11 +1862,15 @@ class Level implements ChunkManager, Metadatable{
 
 			$ev->call();
 			if(!$ev->isCancelled()){
-				if(!$player->isSneaking() and $blockClicked->onActivate($item, $player)){
+				if(!$player->isSneaking()){
+					if($blockClicked->onActivate($item, $player)){
+						return true;
+					}elseif($item->onActivate($player, $blockReplace, $blockClicked, $face, $clickVector)){
+						return true;
+					}
+				}elseif($item->getId() === Item::AIR and $blockClicked->onActivate($item, $player)){
 					return true;
-				}
-
-				if(!$player->isSneaking() and $item->onActivate($player, $blockReplace, $blockClicked, $face, $clickVector)){
+				}elseif($item->onActivate($player, $blockReplace, $blockClicked, $face, $clickVector)){
 					return true;
 				}
 			}else{
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -1868,7 +1868,7 @@ class Level implements ChunkManager, Metadatable{
 					}elseif($item->onActivate($player, $blockReplace, $blockClicked, $face, $clickVector)){
 						return true;
 					}
-				}elseif($item->getId() === Item::AIR and $blockClicked->onActivate($item, $player)){
+				}elseif($item->isNull() and $blockClicked->onActivate($item, $player)){
 					return true;
 				}elseif($item->onActivate($player, $blockReplace, $blockClicked, $face, $clickVector)){
 					return true;


src/pocketmine/level/Level.php
```diff
@@ -1862,11 +1862,15 @@ public function useItemOn(Vector3 $vector, Item &$item, int $face, Vector3 $clic
...
 			$ev->call();
 			if(!$ev->isCancelled()){
-				if(!$player->isSneaking() and $blockClicked->onActivate($item, $player)){
+				if(!$player->isSneaking()){
+					if($blockClicked->onActivate($item, $player)){
+						return true;
+					}elseif($item->onActivate($player, $blockReplace, $blockClicked, $face, $clickVector)){
+						return true;
+					}
+				}elseif($item->getId() === Item::AIR and $blockClicked->onActivate($item, $player)){
```
dktapps, line 1871: you can use `item->isNull()` here

dktapps:
I'm a little confused what the exact problem is that you're trying to address here. You mention empty handed, but also mention the problem of buckets in the same sentence.

I am not currently aware of any issues regarding interacting empty-handed while sneaking. I am aware that buckets aren't usable while sneaking, but that has to do with the fact that it's considered an item activation rather than a block activation, which needs a more specialized fix.

Please clarify what the exact goal(s) of this PR are so I can better understand the changes proposed.

95CivicSi:
Updated original post to better define the target of this PR.

dktapps:
how about if I sneak and click a door with a stick. Does that match vanilla behaviour?

I don't think this solution is correct, particularly since PM buckets have their own logic that does not match vanilla's own, which is the reason for the behavioural difference of buckets (in vanilla using a bucket is the same as placing a block).

src/pocketmine/level/Level.php
```diff
@@ -1862,11 +1862,15 @@ public function useItemOn(Vector3 $vector, Item &$item, int $face, Vector3 $clic
...
 			$ev->call();
 			if(!$ev->isCancelled()){
-				if(!$player->isSneaking() and $blockClicked->onActivate($item, $player)){
+				if(!$player->isSneaking()){
+					if($blockClicked->onActivate($item, $player)){
+						return true;
+					}elseif($item->onActivate($player, $blockReplace, $blockClicked, $face, $clickVector)){
+						return true;
+					}
+				}elseif($item->isNull() and $blockClicked->onActivate($item, $player)){
```
dktapps, line 1871: can't this be simplified into the previous condition `(!$player->isSneaking() or $item->isNull())` , or am I missing some subtlety here?

dktapps:
I think the referenced fix is a simpler version of what you were going for. Thanks for putting the time into this.

PR #3420, opened 2020-04-17
Added missing Window Types

TwistedAsylumMC:
## Introduction

PocketMine has been missing a few window types, but this pull request adds their IDs to the WindowTypes class.

### API changes

- The constant ``WindowTypes::LOOM`` has been added with the value 24
- The constant ``WindowTypes::LECTERN`` has been added with the value 25
- The constant ``WindowTypes::GRINDSTONE`` has been added with the value 26
- The constant ``WindowTypes::CARTOGRAPHY`` has been added with the value 30

### Behavioural changes

There is no behavioural changes
## Backwards compatibility

There are no backwards compatibility breaks

### DIFF
--- a/src/pocketmine/network/mcpe/protocol/types/WindowTypes.php
+++ b/src/pocketmine/network/mcpe/protocol/types/WindowTypes.php
@@ -51,9 +51,12 @@ interface WindowTypes{
 	public const ELEMENT_CONSTRUCTOR = 21;
 	public const MATERIAL_REDUCER = 22;
 	public const LAB_TABLE = 23;
-
+	public const LOOM = 24;
+	public const LECTERN = 25;
+	public const GRINDSTONE = 26;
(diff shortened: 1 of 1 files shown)


jasonwynn10:
Aren't all these going to be removed eventually though?

dktapps:
what gives you that idea?

jasonwynn10:
Window ids dont exist in API 4

DaPigGuy:
https://github.com/pmmp/PocketMine-MP/blob/master/src/network/mcpe/protocol/types/inventory/WindowTypes.php

jasonwynn10:
Oh I'm wrong, my bad. 👍

dktapps:
these are protocol constants, so I don't see why you would think that.

dktapps:
closing and reopening to try and get travis to cooperate

PR #3426, opened 2020-04-18
support 1.14.60+&1.14.0~1.14.30

lovetwice1012:
This pull request is intended to be supported in the latest version of pmmp from 1.14.0 to 1.14.30. This pull request change was achieved by conditional branching by client protocol. This code is running on my server and has been verified to work.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -146,6 +146,8 @@ use pocketmine\network\mcpe\protocol\types\CommandEnum;
 use pocketmine\network\mcpe\protocol\types\CommandParameter;
 use pocketmine\network\mcpe\protocol\types\ContainerIds;
 use pocketmine\network\mcpe\protocol\types\DimensionIds;
+use pocketmine\network\mcpe\protocol\types\PersonaPieceTintColor;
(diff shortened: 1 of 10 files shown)


lovetwice1012:
> Just close this now, I guarentee its not getting merged
ok. I closed it. But I don’t know why it’s not getting merged. I’m very sorry and cloud you tell me about it for me?

dktapps:
If it's been "verified to work" you obviously haven't done very much testing, since I can see at least one place where decoding packets will fail.

We have a long term policy of avoiding multi version because the system we have is simply not designed to handle it. It's a slippery slope that I have no intention of traversing, there's too many issues that you won't consider.

lovetwice1012:
I'm sorry to have requested something that does not follow the policy.
I want to be careful in the future.
I also read the point that the test wasn't enough.
I found that I had to be careful about how to write.
I think it's bad because I neglected to carefully read the policy and the text to post before making a request.
We apologize for making you feel uncomfortable.

dktapps:
It's not a written policy, you don't have anything to apologize for there. However we do require that code is properly tested, and this is clearly not.

Driesboy:
I'm also pretty sure $packet->protocol isn't defined.

PR #3434, opened 2020-04-23
Change "logged in from another location" to localized strings

Encritary:
## Introduction

These reasons are present in the game localization, so it's better to use localized strings.

Reference (1.14.60.5 en_US.lang disconnectionScreen section): https://hastebin.com/exugozoyez.sql

## Backwards compatibility
Maybe some plugins check for the exact ``"logged in from another location"`` string in ``PlayerKickEvent``.

## Follow-up

Requires translations:

| Name | Value in eng.ini |
| :----: | :-----: |
| `disconnectionScreen.loggedinOtherLocation` | `Logged in from other location` |
| `disconnectionScreen.serverIdConflict` | `Cannot join world. The account you are signed in to is currently playing in this world on a different device.` |

## Tests
Try to join from different devices. Also, do the same thing with ``PlayerKickEvent->setCancelled()``.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -1981,8 +1981,8 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 	protected function processLogin(){
 		foreach($this->server->getLoggedInPlayers() as $p){
 			if($p !== $this and ($p->iusername === $this->iusername or $this->getUniqueId()->equals($p->getUniqueId()))){
-				if(!$p->kick("logged in from another location")){
-					$this->close($this->getLeaveMessage(), "Logged in from another location");
+				if(!$p->kick("disconnectionScreen.loggedinOtherLocation")){ // "Logged in from other location"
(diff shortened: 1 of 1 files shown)


src/pocketmine/Player.php
```diff
@@ -1981,8 +1981,8 @@ public function onVerifyCompleted(LoginPacket $packet, ?string $error, bool $sig
 	protected function processLogin(){
 		foreach($this->server->getLoggedInPlayers() as $p){
 			if($p !== $this and ($p->iusername === $this->iusername or $this->getUniqueId()->equals($p->getUniqueId()))){
-				if(!$p->kick("logged in from another location")){
-					$this->close($this->getLeaveMessage(), "Logged in from another location");
+				if(!$p->kick("disconnectionScreen.loggedinOtherLocation")){ // "Logged in from other location"
+					$this->close($this->getLeaveMessage(), "disconnectionScreen.serverIdConflict"); // "Cannot join world. The account you are signed in to is currently playing in this world on a different device."
```
dktapps, line 1985: I don't think this is suitable. This might be fired because of two completely different players with the same username.

dktapps:
Closing due to conflicts and lack of activity.

PR #3439, opened 2020-04-25
Use env to locate bash

dschwartz783:
This change uses env to locate bash, instead of hard-coding it. This is necessary on FreeBSD. Tested on Linux as well.

## Introduction
It's better to not hardcode the shell's path. Env is always in the same place, and can locate it.

### Relevant issues
N/A

### API changes
N/A

### Behavioural changes
None

## Backwards compatibility
Completely

## Tests
Script still executes on Linux, but it also executes on FreeBSD as well now.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 DIR="$(cd -P "$( dirname "${BASH_SOURCE[0]}" )" && pwd)"
 cd "$DIR"
 


SOF3:
Was PocketMine even supposed to support BSD?

SOF3:
Also consider systems with no bash at all. Does the current script use any bash features? If not, what if we just use /bin/sh?

dschwartz783:
PocketMine doesn't support any operating system really, it support PHP configured with the necessary dependencies, all of which I've compiled, and I've verified that PocketMine does in fact run on FreeBSD, with this sole modification.

Your point is valid though, the issue is that there is no definition of exactly what shell /bin/sh is, and there can be differences. On FreeBSD, for example, /bin/sh is a "Bourne shell-compatible" shell, the default for users, while tcsh is used for root. I've verified that start.sh doesn't run on either.

Installing bash is as trivial on FreeBSD as it is on any Linux distro, and it's required by many other packages anyway, so it's likely to be installed regardless.

dktapps:
This is all just bikeshedding in my opinion. I'll accept the pull as long as it doesn't break any existing platforms, but I don't have time to run around and test all the stuff. It's enough of a pain in the ass when I keep getting complaints about CentOS PHP binaries.

dschwartz783:
> This is all just bikeshedding in my opinion. I'll accept the pull as long as it doesn't break any existing platforms, but I don't have time to run around and test all the stuff. It's enough of a pain in the ass when I keep getting complaints about CentOS PHP binaries.

FreeBSD users are usually more advanced than the common Linux user. It's unlikely you'll hear complaints about anything not related directly to PocketMine such as PHP binaries. As I said, I built all the dependencies myself. The PHP build scripts don't even work right, and I didn't feel like figuring out why.

This is actually quite common in shell scripts, specifically because there isn't any guarantee exactly where bash is located (if compiled manually, Linux distros actually default to /usr/local/bin).

dktapps:
My point is that my focus is keeping everything working for the majority of users. So, as long as it doesn't break for those majority of users, a change that benefits a minority is OK by me. But since I can't easily test it, I'll have to let other people do that.

SOF3:
It's worth noting that `/bin/bash` is not guaranteed to exist as per the Linux Filesystem Hierarchy Standard. Therefore the existence of `/bin/bash` is not much more reliable than the existence of `/usr/bin/env`.

SOF3 (approved):

dschwartz783:
> It's worth noting that `/bin/bash` is not guaranteed to exist as per the Linux Filesystem Hierarchy Standard. Therefore the existence of `/bin/bash` is not much more reliable than the existence of `/usr/bin/env`.

In reality nothing is guaranteed, but you do have to make assumptions at some point. /usr/bin/env is always in the same place on all Unix-like operating systems, which is why this is often what people put in their shebangs.

SOF3:
The point is that this change is not even BSD-specific. It is never guaranteed on Linux that bash is in /bin/bash.
Nevertheless, nor is there guarantee that bash exists in $PATH at all.

dschwartz783:
> The point is that this change is not even BSD-specific. It is never guaranteed on Linux that bash is in /bin/bash.
> Nevertheless, nor is there guarantee that bash exists in $PATH at all.

The important point here is the location bash was originally assumed to be in this script (/usr/bin) should always be in $PATH, otherwise a lot of things would break (and you'd have far bigger problems than just locating bash). If it's located anywhere else, this might help. Otherwise, it will do nothing, and work exactly as it was working prior to the change.

PR #3443, opened 2020-04-26
Getter for generator addition.

ZloyNick:
Add a method Level::getGenerator() : Generator

## Introduction
No

### Relevant issues
No

### API changes
src\pocketmine\level\Level::getGenerator() method addition

### Behavioural changes
No

## Backwards compatibility
no

## Follow-up
No

## Tests
Works.


dktapps:
This has no practical value, particularly since the generator doesn't actually exist on the main thread here, it's actually a `::class`. It's obvious you didn't test this.

Issue #2885, opened 2019-04-26
Constants: Move to using final classes + private constructors instead of interfaces


dktapps:
## Description
Interfaces which only contain constants should be converted to final classes with private constructors.

## Justification
Constant interfaces commonly cause confusion amongst plugin developers. An interface is designed to be implemented, but here they are being abused to store constants. This misleads plugin devs into thinking such interfaces are necessary to implement in places where they aren't.

lukeeey:
How about abstract classes

dktapps:
Abstract invites extension, which creates the same confusion problem.

PR #3456, opened 2020-04-29
Added constants for devices

buchwasa:
## Introduction

As of right now, there are classes for input mode, dimensions, containers, but the device os constants from LoginPacket (and PlayerAuthInputPacket) have been left out. This PR adds all known constants of devices.

### API changes
Nope
### Behavioural changes
None
## Backwards compatibility
Still backwards compatible.

### DIFF
--- a/src/pocketmine/network/mcpe/protocol/types/DeviceIds.php
+++ b/src/pocketmine/network/mcpe/protocol/types/DeviceIds.php
@@ -1,0 +1,42 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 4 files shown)


dktapps:
This should be a final class instead of an interface. See #2885

Encritary:
This should be named DeviceOS, not just Device

yoraze (changes requested):

src/pocketmine/network/mcpe/protocol/types/DeviceIds.php
```diff
@@ -0,0 +1,44 @@
...
+	public const IOS = 2;
+	public const OSX = 3;
+	public const AMAZON = 4;
+	public const GEAR_VR = 5;
+	public const HOLOLENS = 6;
+	public const WINDOWS_10 = 7;
+	public const WIN32 = 8;
+	public const DEDICATED = 9;
+	public const TVOS = 10;
+	public const PLAYSTATION = 11;
```
TwistedAsylumMC, line 39: Isn't play station two words?

dktapps: I don't think anyone is going to think to use `PLAY_STATION`. My first thought would be `PS4`.

src/pocketmine/network/mcpe/protocol/types/DeviceIds.php
```diff
@@ -0,0 +1,44 @@
...
+ *
+*/
+
+declare(strict_types=1);
+
+namespace pocketmine\network\mcpe\protocol\types;
+
+interface DeviceIds{
+
+	public const UNKNOWN = -1;
```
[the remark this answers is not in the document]
Wumpotamus: Good point, thank you.

dktapps:
I've merged this into `stable` because it's a protocol system change rather than an API addition, and we want to keep the protocol layer as consistent as possible between branches to minimize upgrading work. Since there is no compatibility guarantee for the protocol layer, this doesn't have to wait for a minor version release.

PR #3465, opened 2020-05-03
the plugin config folder is no longer created

forsterik:
Removed annoying empty folders that are created when the plugin is launched, but the plugin does not use this folder

### DIFF
--- a/src/plugin/PluginManager.php
+++ b/src/plugin/PluginManager.php
@@ -148,9 +148,6 @@ class PluginManager{
 						$this->server->getLogger()->error("Projected dataFolder '" . $dataFolder . "' for " . $description->getName() . " exists and is not a directory");
 						return null;
 					}
-					if(!file_exists($dataFolder)){
-						mkdir($dataFolder, 0777, true);
-					}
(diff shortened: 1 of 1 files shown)


SOF3:
See discussion in https://github.com/pmmp/PocketMine-MP/pull/2284

dktapps:
Hi, this behaviour is intentional. It was introduced in 3.2.0 because it became apparent that it was easier to have some empty directories instead of having to have every single plugin duplicate the same three lines at the top of their `onEnable()` routine. Reversing this now would have backwards-compatibility ramifications. See #2284 for further insights on this.

SOF3:
In particular, could you describe how "annoying" it is? The plugin_data folder should contain as many folders as number of files in your plugins folder. How is that actually annoying?

dktapps:
I don't think there's any point rehashing this since it was already discussed in #2284. Unless there's new material to be brought to the table, I think it would be better to let #2284 speak for itself.

forsterik:
I apologize for my inattention, thank you all!

PR #3470, opened 2020-05-06
Players no longer burn when hit by other players

### LINKED ISSUE #3170: Action not found in vanilla MCPE
(I use Google Translate)
### Issue description

- Expected result: When attacked by a burned player, the attacked player does not burn in vanilla MCPE
- Actual result: PocketMine-MP burns the attacked player

1. ... The attacking side must burn, so burn with flint
2. ... Attack the player while maintaining that state
3. ... Attacked player burns

### OS and versions

* PocketMine-MP: 3.9.3
* PHP: 7.2.19
* Server OS: Linux
* Game version: PE/Win10

### Plugins
`Plugins (0):`

Wumpotamus:
## Introduction

As explained in issue #3170, in neither Minecraft Bedrock nor Java, does the action exist anymore that players can set each other on fire by hitting each other, this PR allows mobs that are on fire to still set players on fire with damage (as that still exists).
### Relevant issues
Fixes #3170 

## Follow-up
Looking at this, the fix could be better, I originally wanted to check if it were an instanceof Monster, however looking at 4.0, that class no longer exists, so I don't want to add the extra burden of dealing with this in 4.0 changes. Ideally, you could actually just remove this if statement in general if you didn't care about mobs, because projectiles are already handled within the [projectile](https://github.com/pmmp/PocketMine-MP/blob/master/src/entity/projectile/Projectile.php#L301-L308) class.

EDIT: 
It seems much more logical to have the setting on fire be handled within Human (or Player), but not targeting a branch that can take this, limits that. The only entities capable of being set on fire through contact are players, ex: a zombie hitting a player whilst on fire will set the player on fire, but a player hitting a zombie on fire whilst on fire does nothing.

### DIFF
--- a/src/pocketmine/entity/Living.php
+++ b/src/pocketmine/entity/Living.php
@@ -559,7 +559,10 @@ abstract class Living extends Entity implements Damageable{
 			if($e !== null){
 				if((
 					$source->getCause() === EntityDamageEvent::CAUSE_PROJECTILE or
-					$source->getCause() === EntityDamageEvent::CAUSE_ENTITY_ATTACK
+					(
+						$source->getCause() === EntityDamageEvent::CAUSE_ENTITY_ATTACK &&
+						!$e instanceof Player
+					)
 				) and $e->isOnFire()){
 					$this->setOnFire(2 * $this->level->getDifficulty());
 				}
--- a/src/pocketmine/entity/Living.php
+++ b/src/pocketmine/entity/Living.php
@@ -556,7 +556,7 @@ abstract class Living extends Entity implements Damageable{
 			if($e !== null){
 				if((
 					$source->getCause() === EntityDamageEvent::CAUSE_ENTITY_ATTACK and
-					!$e instanceof Player
+					$e instanceof Zombie
 				) and $e->isOnFire()){
 					$this->setOnFire(2 * $this->level->getDifficulty());
 				}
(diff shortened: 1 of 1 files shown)


dktapps:
It feels off that this has to be specialized to include players.

jasonwynn10:
How about specializing against Human entities?

dktapps:
I think this needs further investigation but I don't currently have the time. If someone else would like to dig into it, please feel free.

dktapps:
[for context, I'm referring to digging into the java source to see if anyone can make sense of this behaviour]

src/pocketmine/entity/Living.php
```diff
@@ -552,18 +552,8 @@ public function attack(EntityDamageEvent $source) : void{
 
 		if($source instanceof EntityDamageByEntityEvent){
 			$e = $source->getDamager();
-			if($source instanceof EntityDamageByChildEntityEvent){
-				$e = $source->getChild();
```
dktapps, line 556: this is still needed for projectile knockback

dktapps:
for posterity: after some internal discussion we decided the best course of action is to simply scrap this code, since we don't have AI anyway and custom implementations can handle this for themselves.

dktapps:
I wonder if anyone ever noticed that cancelling `EntityCombustByEntityEvent` doesn't prevent projectiles getting set on fire ...

PR #3478, opened 2020-05-09
Made So /ver doesnt leak plugins!

NopeNotDark:


dktapps:
Use permissions if you don't want your users to see this information.

PR #3492, opened 2020-05-13
Fix walk sounds

### LINKED ISSUE #3438: No hear walking sounds other players
No hear walking sounds other players
But i know how to fix it
MovePlayerPacket has global variable "onGround", default = false
Because of this, the client thinks that the player is in the air and does not play the sound of walking
Sorry for my English :(

PocketMine-MP: 3.12.0
Server OS: Ubuntu 18.04
Game version: PE/Win10

Govdim:
Relevant issues
Fixes #3438 

Tests
Tested in my Ubuntu 18.04

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -3738,6 +3738,7 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 		$pk->headYaw = $yaw;
 		$pk->yaw = $yaw;
 		$pk->mode = $mode;
+		$pk->onGround = $this->onGround;
 
 		if($targets !== null){
 			$this->server->broadcastPacket($targets, $pk);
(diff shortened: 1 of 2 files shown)


src/pocketmine/Player.php
```diff
@@ -3738,6 +3738,7 @@ public function sendPosition(Vector3 $pos, float $yaw = null, float $pitch = nul
 		$pk->headYaw = $yaw;
 		$pk->yaw = $yaw;
 		$pk->mode = $mode;
+		$pk->onGround = $this->onGround;
```
dktapps, line 3741: I don't know how reliable or safe this is. Players are considered "on ground" when they are within 0.2 blocks of landing vertically. I also don't know if this is used for anything besides teleporting.

Encritary: Walking sounds appear if you send onGround properly.

dktapps: That's not the question here. The question is whether this flag (which is set when the player is within 0.2 blocks of the ground), is accurate enough, and also whether it has any side effects.

Govdim: This fix is ​​safe, no problems were observed

PR #3499, opened 2020-05-16
Fixes #3495

### LINKED ISSUE #3495: Enchantments bug
Weapon enchantments affect entities when the EntityDamageByEntityEvent is cancelled

Example:
I play the BedWars minigame, and I have a bow with the "PUNCH" enchantment. I shoot at my teammate and the damage does not pass, but the player knockback.

PocketMine-MP: 3.12.1
PHP: 7.2
Server OS: Ubuntu 18.04
Game version: PE/Win10

kenygamer:
## Introduction

Fixes #3495
## Changes
This fixes the  issue mentioned. When `Entity::flagForDespawn` is used it will tick the arrow one time before despawning. With this fix Punch and Infinity enchantments will stop processing that tick if the EntityShootBowEvent event is cancelled
## Tests

Tested in my Ubuntu 16.04.3 LTS server.

### DIFF
--- a/src/pocketmine/item/Bow.php
+++ b/src/pocketmine/item/Bow.php
@@ -93,7 +93,7 @@ class Bow extends Tool{
 			$entity = $ev->getProjectile(); //This might have been changed by plugins
 
 			if($ev->isCancelled()){
-				$entity->flagForDespawn();
+				$entity->close();
 				$player->getInventory()->sendContents($player);
 			}else{
 				$entity->setMotion($entity->getMotion()->multiply($ev->getForce()));


Muqsit:
This doesn't fix the bug where `EntityDamageByEntityEvent` being cancelled causes the punch enchantment to still process.

Notice that `EntityDamageByEntityEvent` is usually called several ticks after `EntityShootBowEvent`, making the two events completely independent of each other at the time of hit.

Your solution passes this test:
```php
public function onShootBow(EntityShootBowEvent $event) : void{
    $event->setCancelled();
}
```

But does it pass this test?
```php
public function onDamageByChild(EntityDamageByChildEntityEvent $event) : void{
    $event->setCancelled();
}
```

dktapps:
This doesn't fix the bug in question, I know that because I know the exact code that causes it. It also creates potential new bugs because of an entity being closed at a time when there might still be things using it.

kenygamer:
I've always known, this commit was stupid...

@Muqsit This passed my EntityDamageByChildEntityEvent and EntityShootBowEvent setCancelled() tests: https://github.com/kenygamer/PocketMine-MP/commit/976c8638309df90134d8e77643cec526f001f4b7

PR #3500, opened 2020-05-17
Fix timings calculation bug in PermissionManager

Muqsit:
## Introduction

`Timings::$permissionDefaultTimer->stopTiming()` didn't get called, leading to inaccurate timings report.


dktapps:
would you mind making a pull against stable?

dktapps:
#3502 will be integrated into master during branch merges, so I'm going to close this. Thanks.

PR #3502, opened 2020-05-17
Fix timings calculation bug in PermissionManager

Muqsit:
## Introduction

`Timings::$permissionDefaultTimer->stopTiming()` didn't get called, leading to inaccurate timings report. (#3500)


dktapps:
pls consider creating a new branch on your fork for each PR, it's easier to manage... anyway, thanks for the fix!

PR #3503, opened 2020-05-17
Ipv6

dschwartz783:
I'd like to work on this. Merging stable into the ipv6 branch and fixing a few things.


dschwartz783:
I'm not sure what else is needed here. Even my rather complicated plugin (makes pretty extensive use of PocketMine's features) works flawlessly with this.

lukeeey:
You're trying to merge a branch that is over **SIX HUNDRED** commits behind the stable branch...thats not gonna work.

dschwartz783:
> You're trying to merge a branch that is over **SIX HUNDRED** commits behind the stable branch...thats not gonna work.

It will, because I corrected the merge conflicts.

dktapps:
It might help to look at the original PR #1936.

dschwartz783:
I believe this is the best solution. Existing installations which may or may not have IPv6 support, will keep working, but if anyone needs and can compile PHP with it, they'll get it. Let me know.

dktapps:
IPv6 support is mandatory anyway. Also, I cannot see the forest for the trees in this PR because it's littered with a whole bunch of unrelated stuff that appears to be my own commits.

dschwartz783:
@dktapps Most of that is merging stable into IPv6. It was necessary.

dschwartz783:
Wanted to emphasize this, because there seems to be confusion from a couple people here. The top of your screen reads like this:
Screen Shot 2020-05-18 at 09 57 46
Have to merge stable to get a working IPv6 branch, which was very outdated. over 600 commits is correct.

dktapps:
@dschwartz783 RakLib requires IPv6 in all conditions because it's necessary to be able to decode IPv6 addresses, which appear in RakLib even when the connection is not itself IPv6.

dktapps:
An additional note: there's currently no dependency listed for ext-filter. Validating IP could alternatively be done using `inet_pton()`, which is available in the stdlib.

dschwartz783:
> @dschwartz783 RakLib requires IPv6 in all conditions because it's necessary to be able to decode IPv6 addresses, which appear in RakLib even when the connection is not itself IPv6.

Well, at the very least PocketMine should be able to handle not having V6, should I work on ensuring RakLib can handle it as well, or do we not care?

dktapps:
RakLib cannot handle it any other way, that's exactly why it's mandatory. Therefore, trying to beat around the bush about a 20-year-old standard that's available in any recent PHP build (and all builds that PM already runs on) is a complete waste of development resources.

dschwartz783:
> RakLib cannot handle it any other way, that's exactly why it's mandatory. Therefore, trying to beat around the bush about a 20-year-old standard that's available in any recent PHP build (and all builds that PM already runs on) is a complete waste of development resources.

Well, you directed me to look at the original PR, and it seems the original PR is misleading. If that's the case, then this should be good to go. I'll revert that one commit which moved around some IPv6 checking logic.

src/pocketmine/Server.php
```diff
@@ -409,10 +409,17 @@ public function getPort() : int{
 		return $this->getConfigInt("server-port", 19132);
 	}
 
+	/**
+	 * @return int
+	 */
```
dktapps, line 414: pls remove this change

src/pocketmine/Server.php
```diff
@@ -409,10 +409,17 @@ public function getPort() : int{
...
 	public function getPortv6() : int{
 		return $this->getConfigInt("server-portv6", 19133);
 	}
 
+
+	/**
+	 * @return int
+	 */
```
dktapps, line 422: and this

src/pocketmine/command/defaults/BanIpCommand.php
```diff
@@ -56,7 +56,7 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
 		$value = array_shift($args);
 		$reason = implode(" ", $args);
 
-		if(preg_match("/^([01]?\\d\\d?|2[0-4]\\d|25[0-5])\\.([01]?\\d\\d?|2[0-4]\\d|25[0-5])\\.([01]?\\d\\d?|2[0-4]\\d|25[0-5])\\.([01]?\\d\\d?|2[0-4]\\d|25[0-5])$/", $value)){
+		if(inet_pton($value)){
```
dktapps, line 59: this needs an explicit `!== false` check since it can return `string|false`. Not sure why PHPStan didn't fail the build.

src/pocketmine/command/defaults/PardonIpCommand.php
```diff
@@ -51,10 +51,12 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
 			throw new InvalidCommandSyntaxException();
 		}
 
-		if(preg_match("/^([01]?\\d\\d?|2[0-4]\\d|25[0-5])\\.([01]?\\d\\d?|2[0-4]\\d|25[0-5])\\.([01]?\\d\\d?|2[0-4]\\d|25[0-5])\\.([01]?\\d\\d?|2[0-4]\\d|25[0-5])$/", $args[0])){
-			$sender->getServer()->getIPBans()->remove($args[0]);
-			$sender->getServer()->getNetwork()->unblockAddress($args[0]);
-			Command::broadcastCommandMessage($sender, new TranslationContainer("commands.unbanip.success", [$args[0]]));
+		$value = $args[0];
+
+		if(inet_pton($value)){
```
dktapps, line 56: same again here

dktapps:
The biggest remaining issue is that the server will crash if either IPv4 or IPv6 is not available. Because of pthreads, handling these error cases is quite a huge pain in the ass, so I'd planned to wait until PM4 before introducing IPv6 support. If you have a solution to that that isn't going to make me rip my eyeballs out, please go ahead.

dschwartz783:
@dktapps Probably not worth the effort for now, then. Pthreads is dead as of PHP 7.4 anyway. Might be worth merging in what I've done so far, though.

dktapps:
Merged your changes as 5627b4c6f. Thank you.

PR #3507, opened 2020-05-18
rename server.phar to server-phar.php

padreon:
## Introduction
server.phar not found on folder build
https://github.com/pmmp/PocketMine-MP/blob/master/build/server-phar.php

## Tests
create PocketMine-MP.phar api 4.0 on my pc with server-phar.php and works


dktapps:
I've merged this by hand since it needed to be applied to stable as well as master. Thank you.

PR #3508, opened 2020-05-18
Update TimeCommand.php to add new time constants

inxomnyaa:
@dktapps What happens if getInteger() is not an integer?

## Introduction

As mentioned in #3385

### API changes

No

### Behavioural changes

Yesn't. Only additional command syntax.

## Backwards compatibility

Fully BC

## Follow-up

Requires change in translations:

| Name | Value in eng.ini |
| :--: | :---: |
| `%pocketmine.command.time.usage` | Should display additional parameters |

## Tests

Its so simple (and i kind of don't want to bother to set up pmmp from source) that it should work right away

### DIFF
--- a/src/pocketmine/command/defaults/TimeCommand.php
+++ b/src/pocketmine/command/defaults/TimeCommand.php
@@ -97,33 +97,26 @@ class TimeCommand extends VanillaCommand{
 			}
 
 			switch ($args[1]){
-				case "day":{
+				case "day":
 					$value = Level::TIME_DAY;
 					break;
-				}
-				case "noon":{
+				case "noon":
 					$value = Level::TIME_NOON;
 					break;
-				}
-				case "sunset":{
+				case "sunset":
 					$value = Level::TIME_SUNSET;
 					break;
-				}
-				case "night":{
+				case "night":
 					$value = Level::TIME_NIGHT;
 					break;
-				}
-				case "midnight":{
+				case "midnight":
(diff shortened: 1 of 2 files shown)


thebigsmileXD:
I actually thought about using an array-key-value map instead, but i was too lazy:
`$possibleValues = ["day"=>Level::TIME_DAY ...]` & `$value = $possibleValues[$args[0]] ?? getInteger ...`
This would've made future changes easier.

matcracker:
In case of non-integer you should throw an InvalidCommandSyntaxException.

dktapps:
stay in your lane

src/pocketmine/command/defaults/TimeCommand.php
```diff
@@ -117,6 +117,7 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
 					break;
 				default:
 					$value = $this->getInteger($sender, $args[1], 0);
+					break;
```
thebigsmileXD, line 120: Is this just for style consistency?

dktapps: Not just style. It's not likely someone will add a case after the default, but in case we do for some reason, this will save that someone a headache. A minute spent now can save an hour later.

dktapps:
Merged as 45c89d084cad6976941205fba45c54d41100a7f8, thanks!

PR #3509, opened 2020-05-18
Burning players no longer set arrows on fire

### LINKED ISSUE #3504: Burning Players Ignite Arrows
### Issue description
To reproduce this issue, I would step into lava, walk away and shoot an arrow. I've done some testing on where the issue could be located, no results yet.

- Expected result: Burning players have no effect on arrows.
- Actual result: Burning players set arrows on fire.

1. Catch fire.
2. Shoot a bow.
3. Profit!

### OS and versions

* PocketMine-MP: 3.12 (Happens on any 3.x and 4.0 build)
* PHP: 7.3.14
* Server OS: Windows 10
* Game version: PE/Win10

### Plugins
Can be reproduced without plugins.

Wumpotamus:
## Introduction
In vanilla Minecraft, arrows do not catch fire whilst a player is on fire. This PR removes that intentional behavior, as it does not exist in Java 1.8 or Bedrock.

### Relevant issues
Fixes #3504

## Tests
Testing consisted of setting a player on fire, the player does not set the arrow on fire.

### DIFF
--- a/src/pocketmine/item/Bow.php
+++ b/src/pocketmine/item/Bow.php
@@ -59,7 +59,6 @@ class Bow extends Tool{
 			($player->yaw > 180 ? 360 : 0) - $player->yaw,
 			-$player->pitch
 		);
-		$nbt->setShort("Fire", $player->isOnFire() ? 45 * 60 : 0);
 
 		$diff = $player->getItemUseDuration();
 		$p = $diff / 20;


dktapps:
oh look, this behaviour came from the same place as #3470 https://github.com/pmmp/PocketMine-MP/commit/fb257c16f13cc271f3ad23e982d82225ca72fb94

PR #3510, opened 2020-05-18
Update LevelEvent, LevelSoundEvent, and Particle constants.

Wumpotamus:
## Introduction
All of this data (excluding particles, those were recently updated) have been out of date for a while. I went ahead and autogenerated the new data for all of those, and left Particle backwards compatible. The script I used does do some stupid things, for example, it changes "SOUND_LEVELUP" to "SOUND_LEVEL_UP," however that is due to the naming of it (It's LevelUp in vanilla) within the game.


Wumpotamus:
Don't worry everyone, I know what I'm doing /s

PR #3511, opened 2020-05-18
Updated LevelEvent, LevelSoundEvent, and Particle constants.

Wumpotamus:
## Introduction
~~Take 2~~ All of this data (excluding particles, those were recently updated) have been out of date for a while. I went ahead and autogenerated the new data for all of those, and left Particle backwards compatible. The script I used does do some stupid things, for example, it changes "SOUND_LEVELUP" to "SOUND_LEVEL_UP," however that is due to the naming of it (It's LevelUp in vanilla) within the game.


dktapps:
the LevelSoundEventPacket IDs are typically generated from the game's `soundEventFromString` map. Since this requires using BDS to extract since 1.13, they haven't been regularly updated.

If I read this correctly, these IDs were generated from the 1.14.60 BDS header leaks. I don't see the value in doing that considering that we're extremely unlikely to get a similar leak again, so mostly this PR will just cause lots of backwards compatibility breaks for no real benefit.

PR #3524, opened 2020-05-24
Remove duplicate call to Entity->resetFallDistance()

Muqsit:
## Introduction

Removes duplicate call to `Entity::resetFallDistance()` introduced in this commit https://github.com/pmmp/PocketMine-MP/commit/8169803bb49a0b3315915188aeb8650a6c3bda64#diff-cf4efef19226a2cd7efd37f88350b474
(`Entity::resetFallDistance()` is called in Water#L69)

## Follow-up

Unsure if the commit referenced above makes this redundant as well: https://github.com/pmmp/PocketMine-MP/blob/stable/src/pocketmine/block/Lava.php#L103

### DIFF
--- a/src/pocketmine/block/Water.php
+++ b/src/pocketmine/block/Water.php
@@ -70,8 +70,6 @@ class Water extends Liquid{
 		if($entity->isOnFire()){
 			$entity->extinguish();
 		}
-
-		$entity->resetFallDistance();
 	}
 
 	public function place(Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, Player $player = null) : bool{


dktapps:
Yeah, I'm not sure what the intended logic was here. This commit is flawed because `resetFallDistance()` and `onGround = true` are not synonymous, so I have no idea what shoghi was doing.
https://github.com/pmmp/PocketMine-MP/commit/8169803bb49a0b3315915188aeb8650a6c3bda64

dktapps:
If you can figure out what the falling logic should be like for lava, feel free to tidy that up as well :)

PR #3527, opened 2020-05-25
WorldManager: Add access to WorldProviderManager

### LINKED ISSUE #3526: Access to WorldProviderManager
API 4.0.0

Make access to WorldProviderManager from WorldManager
Without this it is impossible to make custom world formats

I would like to see something like this way to get the class WorldProviderManager...
$server->getWorldManager()->getProvider(); //return WorldProviderManager

Govdim:
Relevant issues
Fixes #3526


src/world/WorldManager.php
```diff
@@ -81,6 +81,10 @@ public function __construct(Server $server, string $dataPath, WorldProviderManag
 		$this->providerManager = $providerManager;
 	}
 
+	public function getProvider() : WorldProviderManager{
```
dktapps, line 84: this should be `getProviderManager()`

dktapps (changes requested):

PR #3530, opened 2020-05-26
BlockFactory: Open access to array $fullList

Govdim:
I recently started making custom items, blocks, and entities using the plugin and quite successfully. There is only one problem that forces me to edit the server core.
The range of indexes in the $fullList array does not allow adding blocks with ID > 512. This fix will allow increasing the size of the array if necessary.

This is one way to solve this problem. If there are other ways, write about it.


TwistedAsylumMC:
You could use a ReflectionClass to access that property if this PR does not get accepted

lukeeey:
It makes sense considering the other fields are public.

evelire:
Maybe just implement method like 'setListSize'? You can increase size of splfixedarray. https://www.php.net/manual/ru/splfixedarray.setsize.php

dktapps:
> It makes sense considering the other fields are public.

That's a nonsense rationale. The other fields have an entirely different use case and are much less breakable. They are public because they are used for fast property reads that don't require creating an object.

Govdim:
Can do as Evelire suggests?

Govdim:
Can do as evelire suggests?

Govdim:
A

Govdim:
Can do as evelire suggests?
> > It makes sense considering the other fields are public.
> 
> That's a nonsense rationale. The other fields have an entirely different use case and are much less breakable. They are public because they are used for fast property reads that don't require creating an object.

Frago9876543210:
I may be wrong, but BlockFactory will be copied into each thread and might be destroyed when PM does GC on async workers. Therefore, states might be different on main and other threads.

evelire:
Yeah, that's problem. Maybe create methods for registering custom items? (Which iterates over all the threads and adds our block there)

TwistedAsylumMC:
Custom items, blocks & entities can all be done via plugins. Blocks require some reflection classes but it is still possible.

dktapps:
> Yeah, that's problem. Maybe create methods for registering custom blocks? (Which iterates over all the threads and adds our block there). Same for setSize

not possible sadly.

> I may be wrong, but BlockFactory will be copied into each thread and might be destroyed when PM does GC on async workers. Therefore, states might be different on main and other threads.
>
> Update from github channel on Discord:
> setSize() most likely there will be no effect for other threads. Need to do something like object init closure I think

Yes, you're correct. This is why it's difficult to make custom blocks work at all.

PR #3536, opened 2020-05-28
Make use of Generator->getSpawn() instead of using hardcoded values

### LINKED ISSUE #1258: Generator::getSpawn is not used
### Issue description
Expected result: When override Generator::getSpawn and return any coordinates, level spawn will be there.
Actual result: Level spawn is always on coordinates 256x70x256(https://github.com/pmmp/PocketMine-MP/blob/master/src/pocketmine/level/format/io/region/McRegion.php#L255)

1. Create custom generator with getSpawn returns `new Vector3(0, 35, 0)`
2. Generate world with the generator
3. Try to login to the server: you will be spawned on 256x70x256

### OS and versions
* PocketMine-MP: https://github.com/pmmp/PocketMine-MP/commit/aa6eef26e985ab9fc0c10e641666a4c5d0c37992
* PHP: PHP 7.0.17 (cli) (built: Mar 18 2017 15:12:04) ( ZTS )
* Server OS: Linux Ubuntu
* Game version: PE/Win10

### Plugins
- VoidGeneratorTest

* [VoidGeneratorTest](https://gist.github.com/Tarik02/f7dc05c81553635965b52e16ddb5fe71)

CortexPE:
## Introduction
It is currently impossible to set the spawn point of a level as it generates as Generator->getSpawn()'s value is never used.

### Relevant issues
* Fixes #1258

### API changes
None

### Behavioural changes
Returning a Vector3 value for Generator->getSpawn() now affects the newly generated world instead of being always at approximately XZ: 256, 256 (on the default PMAnvil world format)

## Backwards compatibility
None

## Follow-up
None

## Tests
Start a new world with the default `Normal` world generator, it should initially set your spawnpoint to (127, 127) as declared at:

https://github.com/pmmp/PocketMine-MP/blob/c3a795e876071601ee31ecfddc6218d8c58b79fa/src/pocketmine/level/generator/normal/Normal.php#L258-L260

### DIFF
--- a/src/pocketmine/level/format/io/leveldb/LevelDB.php
+++ b/src/pocketmine/level/format/io/leveldb/LevelDB.php
@@ -30,6 +30,7 @@ use pocketmine\level\format\io\exception\CorruptedChunkException;
 use pocketmine\level\format\io\exception\UnsupportedChunkFormatException;
 use pocketmine\level\format\SubChunk;
 use pocketmine\level\generator\Flat;
+use pocketmine\level\generator\Generator;
 use pocketmine\level\generator\GeneratorManager;
 use pocketmine\level\Level;
 use pocketmine\level\LevelException;
@@ -211,6 +212,9 @@ class LevelDB extends BaseLevelProvider{
 			//TODO: add support for limited worlds
 		}
 
+		/** @var Generator $gen */
+		$gen = new $generator($options);
+		$spawn = $gen->getSpawn();
 		$levelData = new CompoundTag("", [
 			//Vanilla fields
 			new IntTag("DayCycleStopTime", -1),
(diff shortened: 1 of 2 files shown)


dktapps:
iirc, the ability to set world spawn via generator was deleted in PM4.

CortexPE:
I think it is better if there's the ability to set world generator especially if you'd want to generate terrain that is supposed to be centered at one point (0, 0).

And why take off a feature from the API that was already there but not used ever since years ago just because of old shogcode and not a technical limitation lol

Commit e9311f5: shoghicp - Added base McRegion classes
https://github.com/pmmp/PocketMine-MP/blob/e9311f5ceba5852ab7b34a7e3064cb0bee7237ec/src/pocketmine/level/format/mcregion/McRegion.php#L69-L71

Also from the same commit, the Generator->getSpawn() abstract method:
https://github.com/pmmp/PocketMine-MP/blob/e9311f5ceba5852ab7b34a7e3064cb0bee7237ec/src/pocketmine/level/generator/Generator.php#L72

dktapps:
I might also add that this area of code changed significantly in PM4, which lowers my interest in merging it due to all the merge conflicts it'll create.

CortexPE:
Perhaps I could create a separate PR that implements this same exact fix for PM4, and re-adds `Generator->getSpawn()`?

dktapps:
Closing this due to being stale, conflicted and obsolete.

I should point out that initializing an entire chunk generator just to get the spawn position is a shitty idea.
Also, anywhere you find yourself using late static binding, there's most likely a better way to do it.

My preferred method would be (in PM4) to either add a `Closure(int $seed) : Vector3 $getSpawnPoint` parameter to `GeneratorManager->registerGenerator()`, or possibly a more flexible `Closure(): WorldCreationOptions $createDefaultOptions` parameter.

PR #3540, opened 2020-05-30
Update ExperienceManager.php

VladYuziuk:
## Introduction
This will give a chance to override ExperienceManager class, before you could not get the entity due to private access.

### API changes
I do not think that it should be documentated


kezum1:
юзак з дарова

VladYuziuk:
There is a better way... xD

PR #3543, opened 2020-06-01
Player: Fixed wrong item use logic

Govdim:
It was possible to use an item only once. It was necessary to switch slots in the hotbar in order to reuse the item. This change fixes this.


Eren5960:
This replacement bow etc. will break the tools

dktapps:
This logic is more wrong than the original. See the comment directly above the line you changed.

PR #3544, opened 2020-06-01
Glass: Block type changed to solid

Govdim:
On the glass it was impossible to put blocks requiring a solid block underneath

### DIFF
--- a/src/pocketmine/block/Glass.php
+++ b/src/pocketmine/block/Glass.php
@@ -25,7 +25,7 @@ namespace pocketmine\block;
 
 use pocketmine\item\Item;
 
-class Glass extends Transparent{
+class Glass extends Solid{
 
 	protected $id = self::GLASS;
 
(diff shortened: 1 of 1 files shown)


dktapps:
This makes mobs suffocate inside glass.

Govdim:
> This makes mobs suffocate inside glass.

Leaving everything as it was also not right. Such a problem not only with glass, but also with slabs

dktapps:
This is just trading one problem for another.

dktapps:
I might also point out that all blocks are already solid by default, so this PR is useless.
https://github.com/pmmp/PocketMine-MP/blame/d554d8060b509fad43a8aa353eccd5a72516c301/src/pocketmine/block/Block.php#L320
https://github.com/pmmp/PocketMine-MP/blob/d554d8060b509fad43a8aa353eccd5a72516c301/src/pocketmine/block/Transparent.php#L26

PR #3545, opened 2020-06-01
Fix wrong item use logic

Govdim:
It was possible to use an item only once. It was necessary to switch slots in the hotbar in order to reuse the item. This change fixes this.

UPD: Changed the previous fix, now the tools do not break

### DIFF
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -1472,7 +1472,7 @@ class Player extends Human implements CommandSender, ChunkLoader, ChunkListener,
 		}
 
 		//TODO: check if item has a release action - if it doesn't, this shouldn't be set
-		$this->setUsingItem(false);
+		$this->setUsingItem(true);
 
 		return true;
 	}
@@ -1507,7 +1507,7 @@ class Player extends Human implements CommandSender, ChunkLoader, ChunkListener,
 			return true;
 		}
 
-		return false;
+		return true;
 	}
 
 	/**
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -1472,7 +1472,7 @@ class Player extends Human implements CommandSender, ChunkLoader, ChunkListener,
 		}
 
 		//TODO: check if item has a release action - if it doesn't, this shouldn't be set
-		$this->setUsingItem(true);
+		$this->setUsingItem(false);
 
 		return true;
 	}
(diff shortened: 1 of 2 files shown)


src/player/Player.php
```diff
@@ -1507,7 +1507,7 @@ public function consumeHeldItem() : bool{
 			return true;
 		}
 
-		return false;
+		return true;
```
TwistedAsylumMC, line 1510: This should not be changed because if this line was ever reached then it means the item was not consumed, and therefore it should return false

dktapps:
You're chasing around hacks that recklessly break other behaviour while still missing the fundamental point that was written directly above the line you changed in your first commit. See discussion in #3366 .

Govdim:
> You're chasing around hacks that recklessly break other behaviour while still missing the fundamental point that was written directly above the line you changed in your first commit. See discussion in #3366 .

I read this discussion and it has nothing to do with my second fix. I tested everything before doing a repeated pull request. Everything worked perfectly, bow, food, and projectileitems. I fixed my first mistake in the second commit.

dktapps:
Both PRs are trying to (incorrectly) fix the same problem.
> //TODO: check if item has a release action - if it doesn't, this shouldn't be set

This is what needs to be addressed.

src/player/Player.php
Govdim: The function returns false in case of unsuccessful food consumption, then to send the previous state of the item to the client (see InGamePaketHandler). Why send a network packet if the item is not food at all and there were no actions on the client side?

Govdim:
> Both PRs are trying to (incorrectly) fix the same problem.
> 
> > //TODO: check if item has a release action - if it doesn't, this shouldn't be set
> 
> This is what needs to be addressed.

The same logic was in a stable branch. Is she wrong too?

dktapps:
Yes. But on the stable branch there are more bugs that cancel this bug out.

Govdim:
> Yes. But on the stable branch there are more bugs that cancel this bug out.

What kind of bug are you talking about? Even if my second fix does not completely solve the problem, it at least corrects the projectile items and did not break other item

src/player/Player.php
dktapps: because this handler isn't supposed to fire at all when food items are not used, and it's expected that it will only do so in the case where the client does something we don't expect it to. This expectation is broken because the use-item flag is set in wrong conditions.

Govdim: use-item flag should be set to false if the item is not used?

dktapps: that's what I've been trying to tell you the whole time.
> //TODO: check if item has a release action - if it doesn't, this shouldn't be set

dktapps: On stable, this bug manifests in a less obvious way: the famous arms-stuck-in-air bug, but the items are still usable because MobEquipmentPacket is handled unconditionally.

On the master branch, MobEquipmentPacket is ignored when the hotbar slot didn't change, which is why this bug appears on master and not on stable.

Govdim: ok, i have an idea how to implement this. Can I add a commit to this PR? Or do I need to open a new one?

PR #3550, opened 2020-06-03
Switch "Auto: Spam" trap to direct links

SOF3:
GitHub supports `contact_links` so that we don't need to use bot label traps anymore.

`blank_issues_enabled: false` forces users to choose one of the templates instead of using none of them. This may or may not be preferrable; RFC.

The `mailto:` link is currently not supported by GitHub. I am not sure what HTTP link we should provide in place of this.

A preview of changes in this PR is available at https://github.com/sofe-pm-pl/refactored-spoon/issues/new/choose


dktapps:
https://github.com/sofe-pm-pl/refactored-spoon/issues/1
blank issues are still a thing

SOF3:
It just forces people to choose one of the templates if I understand correctly

dktapps:
well it didn't force me to do anything. If you go to /new you can still create a blank one.

dktapps:
I think we should leave the security one alone until/unless github gets support for `mailto`.

SOF3:
I updated the link for security vulnerabilities to CONTRIBUTING.md since GitHub does not support email links.

The blank issues property can be left alone since it does not make anything worse.

SOF3:
@dktapps Please review this pull request so that we can stop getting notified for `Support Request` idiots every day.

Issue #2988, opened 2019-06-30
Transition away from Minecraft formatting codes for internal usage


dktapps:
## Description
This issue proposes that we should change the standard for text formatting within PocketMine-MP to use a different non-crap system like HTML. This includes `CommandSender->sendMessage()`, logging, the title bar, and so on.

As appropriate, HTML formatting would be translated to Minecraft formatting on the fly at the network layer.

## Justification
Minecraft's standard formatting codes suck. There are only 16 colours, the formatting set is limited, and there's no way to enclose text.

The primary problem that motivated this issue is the fact that it's not possible to revert to the previous formatting using Minecraft colour codes. In HTML you can do things like this:
```
<blue>This text is blue, but <red>these words are red</red>, and these words are blue again.</blue>
```
There's no way to "close" or encapsulate a block of text with formatting codes with Minecraft formatting. For example:
```
&l&aBold green text, &0black text&r and "default" text.
```
The `&r` formatting code isn't fit for purpose because it removes _all_ formatting.
In addition, the enclosement problem is particularly a headache with translations when translation parameters contain their own formatting codes.

In addition, IDEs like PhpStorm natively support HTML formatting codes and show a nice colour marker in the left column of the editor.

## Alternative methods
It's possible we could alter the existing system and extend it to add more formatting capabilities, but this would also necessitate on-the-fly network translation, and it's not worth the effort.

Consider the following:
```
&4red text&r
```
vs
```
TextFormat::RED . "red text" . TextFormat::RESET
```
vs
```
<red>red text</red>
```

nathfreder:
I'm pretty sure `<red></red>` isn't valid HTML.

SOF3:
Nobody said it's HTML. Let's call it XML instead.

SOF3:
Note the potential problem of unclosed tags in individual return strings. It might be appropriate to box each string in containers, e.g.
```php
function _(string $xml) : FormattedString{
  validate_xml_or_throw($xml);
  return new FormattedString($xml);
}
```

SOF3:
Due to the infamous verbosity of XML, it might be worth considering, at the cost of lower IDE support, to use an alternative syntax for markup, such as `{red(some red text)}`, etc.

lukeeey:
Hmmm
```
{yellow(Hello {red(%s)}!) How are you doing {light_purple(today)}?}
```

UnknownOre:
The html tabgs really looks good

[...]

PR #3551, opened 2020-06-03
Improving the language system

Saehing:
## Introduction
This PR contributes to improving the server language system, also correcting some inconsistencies.

### Relevant issues
* Fixed translation of the text for the player is used.  Previously, a player took as a basis the language that is selected on the server, now for each player his language pack is automatically selected, based on his language from LoginPacket.

## Introduction
This PR helps improve the server's language system, as well as correcting some inconsistencies.

### Related issues
* Fixed text translation is used for the player.  If previously the player used the language selected on the server as the basis, now each player automatically selects their language pack based on their language from LoginPacket.

### API changes
Using the function "loadLanguages()" in LanguageManager you can now load your language packs, this can be used in plugins.

## Backward compatibility
This PR is compatible with old plugins.

## Tests
Proof of Fixing Problem Number One:
image

Test plugin for showing loading of your language packs:
https://gist.github.com/def6b3c3661b4941561a9000f6aad9d7

Also launched phpstan, which did not swear at this PR: 
image

### DIFF
--- a/src/command/SimpleCommandMap.php
+++ b/src/command/SimpleCommandMap.php
@@ -64,6 +64,7 @@ use pocketmine\command\defaults\VanillaCommand;
 use pocketmine\command\defaults\VersionCommand;
 use pocketmine\command\defaults\WhitelistCommand;
 use pocketmine\command\utils\InvalidCommandSyntaxException;
+use pocketmine\player\Player;
 use pocketmine\Server;
 use function array_shift;
 use function count;
@@ -243,7 +244,10 @@ class SimpleCommandMap implements CommandMap{
 		try{
 			$target->execute($sender, $sentCommandLabel, $args);
 		}catch(InvalidCommandSyntaxException $e){
-			$sender->sendMessage($this->server->getLanguage()->translateString("commands.generic.usage", [$target->getUsage()]));
+			if($sender instanceof Player) {
+				$text = $sender->getLanguage()->translateString("commands.generic.usage", [$target->getUsage()]);
+			} else $text = $this->server->getLanguage()->translateString("commands.generic.usage", [$target->getUsage()]);
+			$sender->sendMessage($text);
 		}finally{
 			$target->timings->stopTiming();
 		}
--- a/src/command/defaults/TimeCommand.php
+++ b/src/command/defaults/TimeCommand.php
@@ -50,7 +50,7 @@ class TimeCommand extends VanillaCommand{
 
 		if($args[0] === "start"){
 			if(!$sender->hasPermission("pocketmine.command.time.start")){
-				$sender->sendMessage($sender->getServer()->getLanguage()->translateString(TextFormat::RED . "%commands.generic.permission"));
+				$sender->sendMessage(TextFormat::RED . "%commands.generic.permission");
 
 				return true;
 			}
@@ -61,7 +61,7 @@ class TimeCommand extends VanillaCommand{
 			return true;
 		}elseif($args[0] === "stop"){
 			if(!$sender->hasPermission("pocketmine.command.time.stop")){
-				$sender->sendMessage($sender->getServer()->getLanguage()->translateString(TextFormat::RED . "%commands.generic.permission"));
+				$sender->sendMessage(TextFormat::RED . "%commands.generic.permission");
 
 				return true;
 			}
@@ -72,16 +72,19 @@ class TimeCommand extends VanillaCommand{
 			return true;
 		}elseif($args[0] === "query"){
 			if(!$sender->hasPermission("pocketmine.command.time.query")){
-				$sender->sendMessage($sender->getServer()->getLanguage()->translateString(TextFormat::RED . "%commands.generic.permission"));
+				$sender->sendMessage(TextFormat::RED . "%commands.generic.permission");
 
 				return true;
 			}
+
 			if($sender instanceof Player){
 				$world = $sender->getWorld();
+				$message = $sender->getLanguage()->translateString("commands.time.query", [$world->getTime()]);
 			}else{
 				$world = $sender->getServer()->getWorldManager()->getDefaultWorld();
+				$message = $sender->getServer()->getLanguage()->translateString("commands.time.query", [$world->getTime()]);
 			}
-			$sender->sendMessage($sender->getServer()->getLanguage()->translateString("commands.time.query", [$world->getTime()]));
+			$sender->sendMessage($message);
 			return true;
 		}
 
@@ -91,7 +94,7 @@ class TimeCommand extends VanillaCommand{
 
 		if($args[0] === "set"){
 			if(!$sender->hasPermission("pocketmine.command.time.set")){
-				$sender->sendMessage($sender->getServer()->getLanguage()->translateString(TextFormat::RED . "%commands.generic.permission"));
+				$sender->sendMessage(TextFormat::RED . "%commands.generic.permission");
 
 				return true;
 			}
@@ -126,7 +129,7 @@ class TimeCommand extends VanillaCommand{
 			Command::broadcastCommandMessage($sender, new TranslationContainer("commands.time.set", [$value]));
 		}elseif($args[0] === "add"){
 			if(!$sender->hasPermission("pocketmine.command.time.add")){
-				$sender->sendMessage($sender->getServer()->getLanguage()->translateString(TextFormat::RED . "%commands.generic.permission"));
+				$sender->sendMessage(TextFormat::RED . "%commands.generic.permission");
 
 				return true;
 			}
--- a/src/lang/Language.php
+++ b/src/lang/Language.php
@@ -23,126 +23,56 @@ declare(strict_types=1);
 
 namespace pocketmine\lang;
 
-use function array_filter;
-use function array_map;
-use function explode;
-use function file_exists;
-use function is_dir;
 use function ord;
-use function parse_ini_file;
-use function scandir;
 use function str_replace;
 use function strlen;
 use function strpos;
-use function strtolower;
 use function substr;
-use const INI_SCANNER_RAW;
-use const SCANDIR_SORT_NONE;
 
-class Language{
-
-	public const FALLBACK_LANGUAGE = "eng";
-
-	/**
-	 * @return string[]
-	 * @phpstan-return array<string, string>
-	 *
-	 * @throws LanguageNotFoundException
-	 */
-	public static function getLanguageList(string $path = "") : array{
-		if($path === ""){
-			$path = \pocketmine\RESOURCE_PATH . "locale/";
-		}
-
-		if(is_dir($path)){
-			$allFiles = scandir($path, SCANDIR_SORT_NONE);
-
-			if($allFiles !== false){
-				$files = array_filter($allFiles, function(string $filename) : bool{
-					return substr($filename, -4) === ".ini";
-				});
-
-				$result = [];
-
-				foreach($files as $file){
-					$code = explode(".", $file)[0];
-					$strings = self::loadLang($path, $code);
-					if(isset($strings["language.name"])){
-						$result[$code] = $strings["language.name"];
-					}
-				}
-
-				return $result;
-			}
-		}
-
-		throw new LanguageNotFoundException("Language directory $path does not exist or is not a directory");
-	}
-
+class Language
+{
 	/** @var string */
-	protected $langName;
+	private $langCode;
+	/** @var string[] */
+	private $langValue;
 
 	/**
-	 * @var string[]
-	 * @phpstan-var array<string, string>
+	 * Language constructor.
+	 * @param string $langCode
+	 * @param string[] $langValue
 	 */
-	protected $lang = [];
-	/**
-	 * @var string[]
-	 * @phpstan-var array<string, string>
-	 */
-	protected $fallbackLang = [];
-
-	/**
-	 * @throws LanguageNotFoundException
-	 */
-	public function __construct(string $lang, ?string $path = null, string $fallback = self::FALLBACK_LANGUAGE){
-		$this->langName = strtolower($lang);
-
-		if($path === null){
-			$path = \pocketmine\RESOURCE_PATH . "locale/";
-		}
-
-		$this->lang = self::loadLang($path, $this->langName);
-		$this->fallbackLang = self::loadLang($path, $fallback);
-	}
-
-	public function getName() : string{
-		return $this->get("language.name");
-	}
-
-	public function getLang() : string{
-		return $this->langName;
+	public function __construct(string $langCode, array $langValue)
+	{
+		$this->langCode = $langCode;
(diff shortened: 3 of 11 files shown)


SOF3:
Please explain in prose what you have changed, why you are changing, impacts to existing code, and the benefits and drawbacks of these changes.

SOF3:
Please rewrite your pull request description based on the pull request template:
https://github.com/pmmp/PocketMine-MP/raw/stable/.github/PULL_REQUEST_TEMPLATE.md

Other pull requests that do not follow the template are either too simple or have prior discussion on Discord. Otherwise, just reading code is insufficient for reviewing.

Saehing:
> Please rewrite your pull request description based on the pull request template:
> https://github.com/pmmp/PocketMine-MP/raw/stable/.github/PULL_REQUEST_TEMPLATE.md
> 
> Other pull requests that do not follow the template are either too simple or have prior discussion on Discord. Otherwise, just reading code is insufficient for reviewing.

Thank you for correcting your comment. I will continue to be guided by it

Saehing:
> Please rewrite your pull request description based on the pull request template:
> https://github.com/pmmp/PocketMine-MP/raw/stable/.github/PULL_REQUEST_TEMPLATE.md
> 
> Other pull requests that do not follow the template are either too simple or have prior discussion on Discord. Otherwise, just reading code is insufficient for reviewing.

What is your opinion about this PR?

src/command/SimpleCommandMap.php
```diff
@@ -243,7 +244,10 @@ public function dispatch(CommandSender $sender, string $commandLine) : bool{
 		try{
 			$target->execute($sender, $sentCommandLabel, $args);
 		}catch(InvalidCommandSyntaxException $e){
-			$sender->sendMessage($this->server->getLanguage()->translateString("commands.generic.usage", [$target->getUsage()]));
+			if($sender instanceof Player) {
```
jasonwynn10, line 247: This check is a bit sloppy in my opinion and can be avoided by moving the `getLanguage()` function to the CommandSender class

src/command/defaults/TimeCommand.php
```diff
@@ -72,16 +72,19 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
 			return true;
 		}elseif($args[0] === "query"){
 			if(!$sender->hasPermission("pocketmine.command.time.query")){
-				$sender->sendMessage($sender->getServer()->getLanguage()->translateString(TextFormat::RED . "%commands.generic.permission"));
+				$sender->sendMessage(TextFormat::RED . "%commands.generic.permission");
 
 				return true;
 			}
+
```
jasonwynn10, line 79: This formatting change can be reverted

src/command/defaults/TimeCommand.php
```diff
@@ -72,16 +72,19 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
...
 			if($sender instanceof Player){
 				$world = $sender->getWorld();
+				$message = $sender->getLanguage()->translateString("commands.time.query", [$world->getTime()]);
 			}else{
 				$world = $sender->getServer()->getWorldManager()->getDefaultWorld();
+				$message = $sender->getServer()->getLanguage()->translateString("commands.time.query", [$world->getTime()]);
```
jasonwynn10, line 85: The message duplication here can also be reverted by moving the `getLanguage()` function to the CommandSender class

src/lang/Language.php
```diff
@@ -23,115 +23,71 @@
...
 
-class Language{
-
-	public const FALLBACK_LANGUAGE = "eng";
+class Language
+{
+	/** @var string */
+	private $langCode;
+	/** @var string[] */
+	private $langValue;
```
jasonwynn10, line 37: Nitpick: could be better named `$langValues` plural

src/lang/LanguageManager.php
```diff
@@ -0,0 +1,169 @@
...
+use function is_dir;
+use function parse_ini_file;
+use function scandir;
+use function substr;
+use function is_array;
+use const INI_SCANNER_RAW;
+use const SCANDIR_SORT_NONE;
+
+class LanguageManager
+{
```
jasonwynn10, line 39: Formatting: brace indentation should match the rest of PocketMine's classes

src/lang/LanguageManager.php
```diff
@@ -0,0 +1,169 @@
...
+		'sk_SK' => 'slv',
+		'sv_SE' => 'swe',
+		'tr_TR' => 'tur',
+		'uk_UA' => 'ukr',
+		'zh_CN' => 'chs',
+		'zh_TW' => 'zho'
+	];
+
+	public function __construct(string $path)
+	{
```
jasonwynn10, line 80: Formatting: brace indentation should match the rest of PocketMine's classes

src/lang/LanguageManager.php
```diff
@@ -0,0 +1,169 @@
...
+
+				return $result;
+			}
+		}
+
+		throw new LanguageNotFoundException("Language directory $path does not exist or is not a directory");
+	}
+
+	public function get(string $langCode): ?Language
+	{
```
jasonwynn10, line 121: Formatting: brace indentation should match the rest of PocketMine's classes

src/lang/LanguageManager.php
```diff
@@ -0,0 +1,169 @@
...
+		return $this->languages[$langCode] ?? $this->languages[self::$localeToLangCode[$langCode]] ?? $this->languages[self::FALLBACK_LANGUAGE] ?? null;
+	}
+
+	public function loadLanguages(string $path = ''): void
+	{
```
jasonwynn10, line 126: Formatting: brace indentation should match the rest of PocketMine's classes.

src/player/Player.php
```diff
@@ -271,6 +275,9 @@ public function __construct(Server $server, NetworkSession $session, PlayerInfo
 		$this->displayName = $this->username;
 		$this->locale = $this->playerInfo->getLocale();
 
+		if(($language = $server->getLanguageManager()->get($playerInfo->getLocale())) !== null)
```
jasonwynn10, line 278: The server crashes if this is ever null. The language property must have a default to be used.

src/player/Player.php
```diff
@@ -1859,7 +1874,7 @@ public function sendMessage($message) : void{
 			return;
 		}
 
-		$this->networkSession->onRawChatMessage($this->server->getLanguage()->translateString($message));
+		$this->networkSession->onRawChatMessage($this->getLanguage()->translateString($message));
```
jasonwynn10, line 1877: Also missing changes on `Player::sendTranslation()`

src/command/SimpleCommandMap.php
Saehing: I agree that the design looks terrible , but what about the usual sendMessage that is used in Player.php? Not applicable, if you move the $language variable to CommandSender, you can use this variable from Player, but it is not logically related that the player's language is in CommandSender

src/lang/LanguageManager.php
Saehing: This is not applicable you can fix it

src/lang/LanguageManager.php
Saehing: This is not applicable you can fix it

src/lang/LanguageManager.php
Saehing: This is not applicable you can fix it

src/lang/LanguageManager.php
Saehing: This is not applicable you can fix it

src/player/Player.php
Saehing: This method can return NULL if the fallback language is missing, but this happens when the server starts, which causes an error and the server stops. Accordingly, if there is a spare language, the server will start successfully, and this method will not be able to return NULL

src/player/Player.php
Saehing: Imperfectly understood. I immediately apologize, because I use a translator ;)

src/lang/Language.php
Saehing: This is not applicable you can fix it

src/command/defaults/TimeCommand.php
Saehing: If you go to extreme measures, you can pass the Language when initializing CommandSender, so that it is stored there in public access

src/command/SimpleCommandMap.php
Saehing: If you go to extreme measures, you can pass the Language when initializing CommandSender, so that it is stored there in public access

src/lang/LanguageManager.php
SOF3: Code style is one of the primary concerns when deciding whether to merge a pull request. The responsibility of correcting code style is at the pull request author not the maintainer.

src/command/SimpleCommandMap.php
SOF3: `CommandSender` is an interface not a class and thus "move the $language variable to CommandSender" does not make sense.

The goal of moving `getLanguage()` to `CommandSender` is so that, by Liskov Substitution Principle, one can send a localized message to a command sender regardless whether the command sender is a Player. For example, if the command was sent by some `CommandSender` implementation not defined by pocketmine, it is still possible to let them have their own language.

Always keep in mind that `CommandSender` is not a sealed interface, and any plugin may add their implementations.

src/command/defaults/TimeCommand.php
SOF3: To be exact, please do not include irrelevant formatting changes at your discrete so as not to corrupt the commit diff.

src/command/SimpleCommandMap.php
Saehing: > `CommandSender` is an interface not a class and thus "move the $language variable to CommandSender" does not make sense.
> 
> The goal of moving `getLanguage()` to `CommandSender` is so that, by Liskov Substitution Principle, one can send a localized message to a command sender regardless whether the command sender is a Player. For example, if the command was sent by some `CommandSender` implementation not defined by pocketmine, it is still possible to let them have their own language.
> 
> Always keep in mind that `CommandSender` is not a sealed interface, and any plugin may add their implementations.

In CommandSender, you can make the getLanguage () method, which is already available in Player and Server

src/player/Player.php
```diff
@@ -649,6 +656,14 @@ public function getLocale() : string{
 		return $this->locale;
 	}
 
+	/**
+	 * @return Language
+	 */
+	public function getLanguage(): Language
```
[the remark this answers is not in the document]
SOF3: Are you sure it should return a `Language` instance instead of a `string`? `Language` currently implies a set of mapping from translation keys to translation values rather than a "language" in its intuitive definition. Is it really ideal to require `CommandSender` to return a set of translations rather than just its own preferred language type?

src/command/SimpleCommandMap.php
jasonwynn10: What if I want to localize the language for the console command sender? Such as a situation where a command is sent via console?

SOF3: In that case, why even require a `Language` object? Why not just expose a method that takes translation key + arguments and returns the translated value, without exposing the actual language type?

src/player/Player.php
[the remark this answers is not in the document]
Saehing: > Are you sure it should return a `Language` instance instead of a `string`? `Language` currently implies a set of mapping from translation keys to translation values rather than a "language" in its intuitive definition. Is it really ideal to require `CommandSender` to return a set of translations rather than just its own preferred language type?

Why is this logic not done in `Server`?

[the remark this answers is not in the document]
SOF3: That's exactly what I'm saying. `CommandSender` should just return a locale name and let server resolve the locale name into a `Language` object.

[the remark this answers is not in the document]
Saehing: > That's exactly what I'm saying. `CommandSender` should just return a locale name and let server resolve the locale name into a `Language` object.

I don't understand why this is, if you can just take `Language` from `Server` and `Player`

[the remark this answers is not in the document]
SOF3: This requires `Player` implementations to manage a reference to the global translation data, or otherwise involves a dependency on `Server` in `Player::getLanguage()`.

In particular, if this method is moved to `CommandSender`, It is an anti-pattern to expect an interface method *always* be implemented by delegating to another function (in particular, a stateful one that depends on irrelevant parts, most likely undefined, of the implementation), while the method could be alternatively defined to return the sole differentiating data such that the caller calls such delegation on their own (in particular using their own state if the delegation is stateful).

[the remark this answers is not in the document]
Saehing: > This requires `Player` implementations to manage a reference to the global translation data, or otherwise involves a dependency on `Server` in `Player::getLanguage()`.
> 
> In particular, if this method is moved to `CommandSender`, It is an anti-pattern to expect an interface method _always_ be implemented by delegating to another function (in particular, a stateful one that depends on irrelevant parts, most likely undefined, of the implementation), while the method could be alternatively defined to return the sole differentiating data such that the caller calls such delegation on their own (in particular using their own state if the delegation is stateful).

I think we should do this: https://github.com/lunaste/PocketMine-MP/commit/13d1775f129f9f0bd37d7f239ca906439abcfdac, the `Player class` does not clog this PR from the word, because only one variable was added, which is a link

src/command/SimpleCommandMap.php
dktapps: There are cases where it's necessary to pre-translate, due to the translation system not supporting nested translations, such as [this](https://github.com/pmmp/pocketmine-mp/blob/3473254d01e5012a39bf9f714a6c87d58907c4ab/src/command/Command.php#L232).

Saehing:
@dktapps, what is your attitude to this PR? What changes do you think should be present in addition to code-style?

SOF3:
The complexity of the language system is not just about simple API changes. It is about what we really treat the language concepts as. We are unlikely to move on anywhere without resolving the concerns above.

Saehing:
> The complexity of the language system is not just about simple API changes. It is about what we really treat the language concepts as. We are unlikely to move on anywhere without resolving the concerns above.

what exactly are the problems here? From the big changes you can embed the translation system in html (https://github.com/pmmp/PocketMine-MP/issues/2988). Here is only this processing machine is not yet available, as far as I know, so these small changes will at least slightly improve this language system

dktapps:
This PR tries to do too many things at the same time, and it's also stale (review comments have not been addressed).

SOF3:
Note: To avoid stale/overloaded pull requests from happening in the future, contributors are strongly recommended to discuss their planned changes with us on Discord first if it involves major API changes. Alternatively, submit an issue that describes what exact issues you you want to address and what exactly changes you want to make, instead of writing code. Planning and peer evaluation is a very important step before changing something big.

PR #3556, opened 2020-06-05
Glass: changed block type to Solid

Govdim:
On the glass it was impossible to put blocks requiring a solid block
UPD: removed suffocate #3544

### DIFF
--- a/src/pocketmine/block/Glass.php
+++ b/src/pocketmine/block/Glass.php
@@ -25,7 +25,7 @@ namespace pocketmine\block;
 
 use pocketmine\item\Item;
 
-class Glass extends Transparent{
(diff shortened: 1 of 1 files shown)


dktapps:
This is even more pointless than the first PR. There's no point extending Solid if you're going to override the only thing it does (which in itself is useless, too).

Govdim:
> This is even more pointless than the first PR. There's no point extending Solid if you're going to override the only thing it does (which in itself is useless, too).

Why immediately closed the pull request if I can make a change? .-.

I can change the Solid extension to Block, but this will not affect the final result.
And yet, my correction solves the problem of placing blocks on glass and does not create new problems. Therefore, do not call PR pointless

PR #3559, opened 2020-06-06
Fixed use of items

Govdim:
After two unsuccessful attempts to solve this problem, I came to the right solution (I hope ._.)

### DIFF
--- a/src/item/Bow.php
+++ b/src/item/Bow.php
@@ -44,6 +44,10 @@ class Bow extends Tool{
 		return 385;
 	}
 
+	public function canBeUsed() : bool{
+		return true;
+	}
+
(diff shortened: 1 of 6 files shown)


TwistedAsylumMC:
Might be better if you had like a ``Useable`` class/interface or something which returns true in that function

Govdim:
> Might be better if you had like a `Useable` class/interface or something which returns true in that function

I think this is not a good idea :/

TwistedAsylumMC:
Why is that?

Govdim:
> Why is that?

Personally, it will be inconvenient for me to implement new items that perform some kind of action.

Govdim:
And what about bow, crossbow and other tools? How will it look like?

PR #3570, opened 2020-06-09
Add get LoginPacket method on Player class

AkmalFairuz:
## Introduction

As we know, we can handling this on DataPacketReceiveEvent. But we need alternative method to getting LoginPacket of player without creating Event.

## Changes

Add method getLoginPacket on Player class.

### DIFF
--- a/src/pocketmine/network/mcpe/protocol/SetScorePacket.php
+++ b/src/pocketmine/network/mcpe/protocol/SetScorePacket.php
@@ -61,6 +61,7 @@ class SetScorePacket extends DataPacket{
 						throw new \UnexpectedValueException("Unknown entry type $entry->type");
 				}
 			}
+			$this->entries[] = $entry;
 		}
 	}
 


AkmalFairuz:
This is very useful for getting client client data such as devices used by players.

Encritary:
It's just a duct tape, not proper API implementation. If you need LoginPacket, just store it after you handle DataPacketReceivePacket.

AkmalFairuz:
That alternative method

lukeeey:
@SOF3 You day that but there has been countless PRs to implement device os functions

dktapps:
There's no standard way to represent "device OS" except the protocol, which is volatile and which we've expressed we don't want to further expose to the API exactly because of said volatility. Considering the very finite value of exposing such information anyway (the only valid use case is for analytics), I don't consider it worth the time.

dktapps:
Another factor to consider is that people have been working around this limitation for so long already that very few people will care if there is an API or not, so it doesn't really make sense to introduce API to Player at this late stage of PM3's lifespan. On PM4 it might make more sense, but my intent for PM4 is to expose it by way of "player info metadata" which isn't bound by any API contract.

SOF3:
> @SOF3 You day that but there has been countless PRs to implement device os functions

Your comment is suggesting that, if PRs adding slightly unstable APIs are on hold, we should instead get PRs adding very unstable APIs.

PR #3577, opened 2020-06-11
add Player::getDeviceID,Player::getDeviceOS

Blackjack200:
add Player::getDeviceID,Player::getDeviceOS

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -385,6 +385,13 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 	/** @var Vector3|null */
 	protected $lastRightClickPos = null;
 
+	/** @var string */
+	protected $deviceId;
+	/** @var string */
+	private $deviceModel;
+	/** @var int */
+	private $deviceOS;
+
 	/**
 	 * @return TranslationContainer|string
 	 */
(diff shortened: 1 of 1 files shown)


src/pocketmine/Player.php
```diff
@@ -385,6 +385,13 @@ public static function isValidUserName(?string $name) : bool{
 	/** @var Vector3|null */
 	protected $lastRightClickPos = null;
 
+	/** @var string */
+	protected $deviceId;
+	/** @var string */
+	private $deviceModel;
+	/** @var int */
+	private $deviceOS;
```
TwistedAsylumMC, line 393: Why is deviceId protected and device model and deviceOS private?

src/pocketmine/Player.php
```diff
@@ -4078,4 +4089,16 @@ public function getLoaderId() : int{
...
+		return $this->deviceOS;
+	}
+
+	public function getDeviceModel() : string{
+		return $this->deviceModel;
+	}
+
+	public function getDeviceId() : string{
+		return $this->deviceId;
+	}
```
TwistedAsylumMC, line 4103: These are in reverse order of the properties

TwistedAsylumMC:
As well as that, you also did not state you added the "getDeviceModel" method. This is also still targeted at the "stable" branch so the same response from your last PR still applies:
> In addition, this PR is targeted at the stable branch, which is not accepting changes apart from protocol changes and bugfixes.

dktapps:
...

mmm545:
dktapps: *sigh*

PR #3580, opened 2020-06-12
Registry Trait: fix strtoupper error

Eren5960:
Fixed incorrect strtoupper of incorrect letters on operating systems using languages other than English. 
Like (Stair => STAİR)

## Introduction
```
Error: No such registry member: pocketmine\block\utils\StairShape::STRAIGHT
File: src/utils/RegistryTrait
Line: 102
Type: Error
```

### DIFF
--- a/src/utils/RegistryTrait.php
+++ b/src/utils/RegistryTrait.php
@@ -30,7 +30,7 @@ use function implode;
 use function sprintf;
 use function strlen;
 use function strpos;
-use function strtoupper;
+use function mb_strtoupper;
 use function substr;
 
 trait RegistryTrait{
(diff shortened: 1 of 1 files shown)


dktapps:
Please add unit tests to demonstrate the problem and to show that your change addresses it.

Eren5960:
Test #1 with English locale:
DeepinScreenshot_select-area_20200613105256

Test #1 with Turkish locale:
DeepinScreenshot_select-area_20200613105332

SOF3:
Please commit the unit test files not the screenshots.

Eren5960:
I don't know which directory to upload

dktapps:
I can't reproduce this on Windows.

dktapps:
I've written unit tests for this bug and got it to reproduce on Travis:  https://travis-ci.com/github/pmmp/PocketMine-MP/builds/171222517#L1521

It seems like this won't be the only place bugs like this show up; for example, it's likely they'll appear  if you for some reason use `@priority highest` instead of `@priority HIGHEST` when the system language is set to Turkish. It may also happen when using `ItemFactory::fromString()`, `Enchantment::fromString()` and various other places.

TL;DR: PHP sucks.

PR #1, opened 2016-10-01
Builtin thread-local object storage for AsyncTask

SOF3:
@dktapps deleted my original pull request. Blame him and ask him for description.

PR #3581, opened 2020-06-12
let attack like vanilla

DylanMarkon:
## Introduction
let attack like vanilla
### Relevant issues
none

* Fixes #1 hit entity with break sprinting

### API changes
none
### Behavioural changes
hit entity with break sprinting
## Backwards compatibility
none

## Follow-up
none
Requires translations:
none
## Tests
tested , and its works :).

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -2620,6 +2620,7 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 							}
 						}
 
+						$this->setSprinting(false);
 						$ev = new EntityDamageByEntityEvent($this, $target, EntityDamageEvent::CAUSE_ENTITY_ATTACK, $heldItem->getAttackPoints());
 
 						$meleeEnchantmentDamage = 0;
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -2618,8 +2618,7 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 							if(!$this->server->getConfigBool("pvp")){
 								$cancelled = true;
 							}
-						}
-						
+						}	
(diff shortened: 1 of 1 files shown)


TwistedAsylumMC:
How does this relate at all to the mentioned pull request?

Blackjack200:
nice

dktapps:
This breaks the critical hit check just a few lines below, and it also shouldn't take effect if the event is cancelled.

TL;DR: this is in the wrong place.

DylanMarkon:
> This breaks the critical hit check just a few lines below, and it also shouldn't take effect if the event is cancelled.
> 
> TL;DR: this is in the wrong place.

kk I fixed it in the recently commit

src/pocketmine/Player.php
```diff
@@ -2619,7 +2619,7 @@ public function handleInventoryTransaction(InventoryTransactionPacket $packet) :
 								$cancelled = true;
 							}
 						}
-
+
```
dktapps, line 2622: please remove this change

Encritary:
This might be an unpleasant change for PvP servers, so it's better to add an option to disable this through API (maybe ``EntityDamageByEntityEvent::setSprintBreak(false)``)

dktapps:
oh noes

dktapps:
Something tells me half of the PvP community is going to complain about this change. The issue is that it's possible to implement this using a plugin right now without any core changes, but not possible to stop it being activated if it becomes core behaviour.

Blackjack200:
> Something tells me half of the PvP community is going to complain about this change. The issue is that it's possible to implement this using a plugin right now without any core changes, but not possible to stop it being activated if it becomes core behaviour.

i will add a function to handle sprint

BenHeroPH (approved):
this is good.

ghost:
> This might be an unpleasant change for PvP servers, so it's better to add an option to disable this through API (maybe ``EntityDamageByEntityEvent::setSprintBreak(false)``)

A developer can just use ``Player::setSprinting(false)``

PR #3582, opened 2020-06-13
Support the ServerSettings.

PJZ9n:
## Introduction
ServerSettings are not currently supported by PMMP.

### Relevant issues
none

### API changes
Yes. API is changing. Information about this is included in the commit.

### Behavioural changes
There is essentially no performance impact.

## Backwards compatibility
There are no changes that are not backward compatible.

## Follow-up
none

## Tests
[Test with script plugin (serversettings-test.php)](https://gist.github.com/PJZ9n/a35ab569d0aeb757c0bc0e84ca2614b5/revisions#diff-cb688af410f1f190b85bf8e94c326b19)
This succeeded successfully. Confirm that the Server Settings are displayed like this.
screenshot_20200613_152935
And of course you can get the response.
screenshot_20200613_153100
*I am sorry that the language setting of the client and PMMP is not English. However, this does not affect the test.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -135,6 +135,7 @@ use pocketmine\network\mcpe\protocol\ResourcePackDataInfoPacket;
 use pocketmine\network\mcpe\protocol\ResourcePacksInfoPacket;
 use pocketmine\network\mcpe\protocol\ResourcePackStackPacket;
 use pocketmine\network\mcpe\protocol\RespawnPacket;
+use pocketmine\network\mcpe\protocol\ServerSettingsResponsePacket;
 use pocketmine\network\mcpe\protocol\SetPlayerGameTypePacket;
 use pocketmine\network\mcpe\protocol\SetSpawnPositionPacket;
 use pocketmine\network\mcpe\protocol\SetTitlePacket;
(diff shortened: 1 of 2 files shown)


SOF3:
Have you considered issues on inter-plugin compatibility? If multiple plugins set server settings, there is no proper way to merge them together. The abstraction of the Form interface makes it difficult for one plugin to know about the structural of another.

PJZ9n:
I considered it, but there is no good way. I thought that the approach that can be done from PMMP is to be able to send ServerSettings.
Currently, the only option is to use the protocol directly, and I don't think this is a good situation.
If you want, you can write a plugin to support it. Should this be built into the PM core?

SOF3:
See also #2964.

PJZ9n:
Well, isn't there any other way to use ServerSettings with stable PMMP than using protocol? Would you like to provide a way to issue a ServerSettings based on the stable version specifications?

SOF3:
The state of form support in PocketMine is still experimental. The current `Form` interface is actually protocol-dependent (because the `jsonSerialize` return value is protocol-dependent), which is not the ultimate API we want to provide. The goal is still to provide a simple API that retains backward comptaibility.

PJZ9n:
Hmm, sure. I think something like pmforms from dktapps should be built into the PM core.

PR #3587, opened 2020-06-15
Add closure task arguments.

PJZ9n:
## Introduction
I can't find any way to use TaskHandler in ClosureTask.
ClosureTask is a Task, so you need to be able to access the TaskHandler.
The only way I know is to use pass by reference.
```php
$handler = $this->getScheduler()->scheduleRepeatingTask(new ClosureTask(function (int $currentTick) use (&$handler, $player): void {
```
However, this technique does not guarantee that $handler is a TaskHandler.
I think this is much more useful than having a currentTick argument :)

### Relevant issues
none

### API changes
Yes, there are API changes. This is clear if you look at the commits.

### Behavioural changes
Does not affect performance.

## Backwards compatibility
It is fully backward compatible.

## Follow-up
none

Requires translations:
none

## Tests
[Test with Script plugin](https://gist.github.com/PJZ9n/9d12bd6a1cb86dac7a3ca58ea2c15cb2/revisions#diff-af582cb50ea11ee45a420a345cd31ae1)
screenshot_20200615_203209
screenshot_20200615_203239
screenshot_20200615_203303

### DIFF
--- a/src/pocketmine/scheduler/ClosureTask.php
+++ b/src/pocketmine/scheduler/ClosureTask.php
@@ -40,16 +40,24 @@ class ClosureTask extends Task{
 
 	/**
 	 * @var \Closure
-	 * @phpstan-var \Closure(int) : void
+	 * @phpstan-var \Closure(int) : void|\Closure(int, TaskHandler) : void
 	 */
 	private $closure;
 
+	private $useTaskHandler = false;
+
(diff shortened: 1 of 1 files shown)


PJZ9n:
Hmmm... I can't think of a way to eliminate this Lint.....

PR #3593, opened 2020-06-16
Fixed wrong meta value when pick cake block.

matcracker:
## Introduction

The PR aims to fix an issue regarding the cake block picking that gives the item with the same block meta value instead of 0.

https://streamable.com/to40zq
 
### API changes

No.
### Behavioural changes

No.
## Backwards compatibility

Yes.
## Tests

Obviously it has been tested, I don't think it's necessary to give the demonstration given the simple solution of the problem.

### DIFF
--- a/src/pocketmine/block/Cake.php
+++ b/src/pocketmine/block/Cake.php
@@ -109,7 +109,7 @@ class Cake extends Transparent implements FoodSource{
 		return true;
 	}
 
-	public function getVariant() : int{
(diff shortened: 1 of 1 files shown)


dktapps:
better solution is to provide `0` for the variant bitmask.

matcracker:
That's right. I already edited 👍

dktapps:
The bitmask, not the variant itself

matcracker:
Done

PR #3607, opened 2020-06-22
Fix "Ender pearl still teleporting to Vector after level change"

### LINKED ISSUE #3397: Ender pearl still teleporting to Vector after level change
### Issue description
I threw an ender pearl on level "pvp" and died before it fell down. I respawned at level "spawn" and got teleported to a ~random~ location which I think are the coords where the ender pearl might have fallen.

1. Throw ender pearl
2. Die before it falls
3. Respawn at your original spawn point (in another level)

### OS and versions

* PocketMine-MP: 3.11.6

### Plugins

- If you remove all plugins, does the issue still occur? No, because I believe you can't teleport between worlds without plugins.

knzv97:
## Introduction

Fix #3397 

### Relevant issues

- Ender pearl still teleports to Vector if player was teleported to another world or died

### API changes

- Added:
`Player::hasActivePearls()` to check if player has an active pearls
`Player::increaseActivePearls()` to increase the count of items
`Player::decreaseActivePearls()` to decrease the count of items
`Player::resetActivePearls()` to reset existing pearls

### Behavioural changes

- Ender pearl now doesn't teleport if player has been teleported or died

## Backwards compatibility

Compatible

## Tests
Without fix: https://youtu.be/Ueijz2fbt1E
With fix: https://youtu.be/7b6uRJbpSsg

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -386,6 +386,9 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 	/** @var Vector3|null */
 	protected $lastRightClickPos = null;
 
+	/** @var int */
+	protected $activePearlsCounter = 0;
+
 	/**
 	 * @return TranslationContainer|string
 	 */
@@ -950,6 +953,8 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 			$this->loadQueue = [];
 			$this->level->sendTime($this);
 			$this->level->sendDifficulty($this);
+
+			$this->resetActivePearls();
 
 			return true;
 		}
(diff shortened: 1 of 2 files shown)


dktapps:
This is kinda missing the real problem, which is that epearls will still teleport you after you were teleported (e.g. respawning in the same world before your epearl hits will still have a problem)

dktapps:
what about /tp?

Muqsit:
How about having an `$teleportCounter` property in `Entity` and storing it in the EnderPearl class to verify the thrower didn't teleport since the launch? Enderpearl's dependency on Player can then be removed by checking if getOwningEntity()->isAlive() before teleportation.

dktapps:
Closing due to lack of activity on maintainer comments.

> How about having an `$teleportCounter` property in `Entity` and storing it in the EnderPearl class to verify the thrower didn't teleport since the launch? Enderpearl's dependency on Player can then be removed by checking if getOwningEntity()->isAlive() before teleportation.

This is the solution I was considering. PR to implement it welcome.

PR #3619, opened 2020-06-24
Release server.lock file when shutdown the server.

matcracker:
## Introduction

When the server is going to be turned off, it is a good idea to release the server.lock file.

### API changes

No.

### Behavioural changes

No.

## Backwards compatibility

Yes.

## Tests

Tried to start and stop the server. No errors occurred.

### DIFF
--- a/src/pocketmine/PocketMine.php
+++ b/src/pocketmine/PocketMine.php
@@ -293,6 +293,10 @@ namespace pocketmine {
 			critical_error("Failed to release the server.lock file.");
 		}
 
+		if(!fclose(\pocketmine\LOCK_FILE)){
+			critical_error("Could not close server.lock resource.");
+		}
+
 		exit($exitCode);
 	}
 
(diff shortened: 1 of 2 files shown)


dktapps:
You need to regenerate the `l7-baseline.neon` file to make the build pass.
Run `vendor/bin/phpstan analyze -l7 --generate-baseline tests/phpstan/configs/l7-baseline.neon` and commit the result.

src/pocketmine/PocketMine.php
```diff
@@ -289,6 +289,10 @@ function server(){
 
 		echo Terminal::$FORMAT_RESET . PHP_EOL;
 
+		if(!flock(\pocketmine\LOCK_FILE, LOCK_UN)){
```
dktapps, line 292: It's cleaner to use `fclose()` to avoid mandatory locking problems on Windows. `fclose()` will automatically release locks.

matcracker: No. I tried to replaced the `flock()` with `fclose()` and the resource has not been released. To be sure of closing the resource, I first did the `flock()` and then `fclose()`.

dktapps: hmm, what platform is that on? I was sure fclose releases locks ...

dktapps: image
never mind, i see now... my mistake

PR #3631, opened 2020-06-26
Fix projectile motion being changed by the ladder (#3602)

### LINKED ISSUE #3602: Arrow accelerates or decelerates when it flies near the ladder
### Issue description

- Expected result: The arrow should fly near the ladder with no change in movement
- Actual result: The arrow accelerates at full stretch of the bow or falls down with a weak stretch

1. Place the ladder
2. Take the bow and shoot so that the arrow flies near the ladder (check video)

### OS and versions

* PocketMine-MP: 3.13.0
* PHP: 7.3.18
* Server OS: Windows 10
* Game version: PE/Win10

### Plugins

Plugins (0):

Video: https://youtu.be/E2-ZjhvE0XY

saeharo:
## Introduction
The arrow(projectile) motion being changed by the ladder

### Relevant issues

* Fixes #3602 

## Changes
Entities that belong to projectiles have been modified as exceptions.

## Backwards compatibility
All projectiles motion are not affected by the ladder 

## Tests
https://youtu.be/tu6KeAWclPs


T14Raptor:
Should be checking if the entity isn't Living, instead of checking if the entity isn't a Projectile, xp orbs, item entities etc. are affected too by ladders, which shouldn't happen

saeharo:
@T14Raptor 
But don't only players use ladders in the game? If so, why not check the player

`!$entity instanceof Projectile` To `!$entity instanceof Player`

dktapps:
Any mob can climb a ladder.

saeharo:
Only human, Animal and Monster has extended Living. So general entities don't extend Living. So, why not check the Living?

`!$entity instanceof Projectile` to `$entity instanceof Living`

dktapps:
that's what T14 Raptor suggested.

saeharo:
Oh.. I didn't know that.

PR #3635, opened 2020-06-27
EmoteListPacket can be ServerboundPacket

alvin0319:
## Introduction

In 4.0, a packet processing error occurs when the player applies an emote to a custom skin and tries to connect.

Like this, 
image

## Changes
*  EmoteListPacket is now implemented as ServerboundPacket.


JaxkDev:
Btw It’s `ServerboundPacket` not `ServerBoundPacket`

dktapps:
This is fine apart from the case mismatch.

PR #3659, opened 2020-06-29
New blocks and items

xerenahmed:
## Introduction
Added 
- Basalt
- Block of Netherite
- Ancient Debris
- Chain
- Netherite Scrap
- Netherite Ingot
- Netherite Tools
- Netherite Armors

### DIFF
--- a/src/block/BlockFactory.php
+++ b/src/block/BlockFactory.php
@@ -571,6 +571,10 @@ class BlockFactory{
 		}
 
 		$this->registerElements();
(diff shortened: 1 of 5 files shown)


dktapps:
BedrockData is auto generated. Don't modify it by hand.

dktapps:
These items shouldn't be implemented yet, because the server is still using an old save format. Adding 1.16 things now will complicate the data migration process once the newer format is actually supported.

Anyway, this PR is stale.

PR #3661, opened 2020-06-29
Multi Folder World Support

padreon:
## Introduction
Server can load/create/save the world on another folder, This is very useful when having multiple servers with the same world (minigame server). And can reduce the amount of memory used

### API changes

- Changes

1. `loadLevel(string $name)` to `loadLevel(string $name, string $path = null)`

2.  `generateLevel(string $name, int $seed = null, $generator = null, array $options = [])`  to  `generateLevel(string $name, int $seed = null, $generator = null, array $options = [], string $path = null)`

3. `isLevelGenerated(string $name)` to `isLevelGenerated(string $name, string $path = null)`

## Tests
Create New Worlds
`Server::getInstance()->generateLevel('NewWorld', null, null, [], 'to/your/path/NewWorld/');`
Load World
`Server::getInstance()->loadLevel('MyWorld', 'to/your/path/MyWorld/');`


dktapps:
> This is very useful when having multiple servers with the same world (minigame server)

This is a bad idea. There's no guarantee that multiple servers won't write to the world simultaneously and corrupt it.

> And can reduce the amount of memory used

This is unsubstantiated nonsense.

padreon:
> This is a bad idea. There's no guarantee that multiple servers won't write to the world simultaneously and corrupt it.

Yes, it will corrupt when 2 servers save their world. But in minigames uses `$ world->setAutoSave(false)` like Skywars/BedWars/ etc. , it will work fine. But this will be a problem when multi folder world is used for survival world which is accessed by 2 or more servers simultaneously

> This is unsubstantiated nonsense.

I mean hdd memory, because all servers using 1 world not 1 server 1 world. 
Example: you have 3 servers skywars minigame, and 1 server has 10 worlds. So on your VPS/PC you have 30 worlds, but if you using multi folder world your server just have 10 worlds.

Frago9876543210:
It is rather sound as an idea for readonly worlds.

Encritary:
If you want to do these hacks, you should know the ``ln`` command. It makes no sense to add a path argument.

SOF3:
The rationale described in the PR is not acceptable (symlinks or network drives can do the same thing, and worlds are not concurrency-safe).

However, it is worth thinking whether we really want to stick with the current system where world list is bijective to folders in `${DATA_PATH}/worlds`.

But if a world does not have a folder, who would determine which world provider to use?

Improvements in this direction would most likely be BC-breaking, requiring the `loadLevel` caller to determine the level format, and probably another `tryLoadLevel` method to auto detect world format.

This is kind of similar to what I have been trying to do with plugin loaders, as in #2834.

AkmalFairuz:
As we know, PocketMine-MP cannot handle hundreds of players on one server, and we have to separate the server and then make it connected to the same world

padreon:
> It is rather sound as an idea for readonly worlds. But then you need a mechanism to prevent damage for R/W worlds.

I made this affected by my skyblock server, I want to make players able to access the same skyblock world (1 island 1 world) without having to choose which server he used before. instead of having to save into a database (load-> extract-> move to worlds-> save-> compress-> save to db-> delete folder in worlds)

> If you want to do these hacks, you should know the ln command. It makes no sense to add a path argument.

yes, this must-have access to use `ln` to know the folder path used, but if you cannot use` ln` you cannot have a multi-server. Multi folder world is good for multi server in same vps but if you only have 1 server it's better to use `${DATA_PATH}/worlds/`

> reply message from @SOF3 

as i said before, this will be broken when 2 / more servers save the same world. and we must still leave 1 world folder in `${DATA_PATH}/worlds/` for the default level and this will not be a collision. multi world folder is more recommended when having 2 or more servers on the same vps, this will be useless when only having 1 server. if the server doesn't have special needs, maybe I prefer to use `${DATA_PATH}/worlds/` because it has a private world in a folder, it won't corrupt the world and look neat.  Multi-world folder in addition to when the server has special needs to access the world folder elsewhere (for multi servers) so if it's not needed, there's no need to add the path to the load/generate function `Server::getInstance()->loadLevel('MyWorld');`

> As we know, PocketMine-MP cannot handle hundreds of players on one server, and we have to separate the server and then make it connected to the same world.

Multi world folder will be useful when having 2 or more servers, but multi world folder cannot connect the server to the same world, only makes the server use the shared world base and when using the same 2 world must turn off setAutoSave so as not to get corrupted.

dktapps:
The problems that I see are:
1. Collisions in `Level->getFolderName()`
2. No guarantee of readonly access
3. Server has no way to know if another server already uses this world (short of server.lock, which covers all server data; this would bypass server.lock)
4. Some formats (i.e. leveldb) already disallow multi-process access. Region worlds don't because the implementation sucks and nobody really thought about it. This ties back into 3) you don't know who else, if anyone, is already using the world.

In principle loading worlds from anywhere is an OK idea, but the rationale doesn't make sense.

dktapps:
I should also mention that there's a 0% chance of getting this integrated into stable at this point because the risk/reward balance isn't good enough to justify changes to a branch I've already said I'm going to discontinue, especially considering that this area has had significant changes on PM4.

SOF3:
> > It is rather sound as an idea for readonly worlds. But then you need a mechanism to prevent damage for R/W worlds.
> 
> I made this affected by my skyblock server, I want to make players able to access the same skyblock world (1 island 1 world) without having to choose which server he used before. instead of having to save into a database (load-> extract-> move to worlds-> save-> compress-> save to db-> delete folder in worlds)
> 
> > If you want to do these hacks, you should know the ln command. It makes no sense to add a path argument.
> 
> yes, this must-have access to use `ln` to know the folder path used, but if you cannot use` ln` you cannot have a multi-server. Multi folder world is good for multi server in same vps but if you only have 1 server it's better to use `${DATA_PATH}/worlds/`
> 
> > reply message from @SOF3
> 
> as i said before, this will be broken when 2 / more servers save the same world. and we must still leave 1 world folder in `${DATA_PATH}/worlds/` for the default level and this will not be a collision. multi world folder is more recommended when having 2 or more servers on the same vps, this will be useless when only having 1 server. if the server doesn't have special needs, maybe I prefer to use `${DATA_PATH}/worlds/` because it has a private world in a folder, it won't corrupt the world and look neat. Multi-world folder in addition to when the server has special needs to access the world folder elsewhere (for multi servers) so if it's not needed, there's no need to add the path to the load/generate function `Server::getInstance()->loadLevel('MyWorld');`
> 
> > As we know, PocketMine-MP cannot handle hundreds of players on one server, and we have to separate the server and then make it connected to the same world.
> 
> Multi world folder will be useful when having 2 or more servers, but multi world folder can't connect player on a different server to the same world, only makes the server use the shared world base and when using the same 2 world must turn off setAutoSave so as not to get corrupted.

You can use symlinks transparently as long as you are on Unix... Not many people run large scale servers on an actual windows **server** (not a windows desktop), and I don't think we should waste time on this small portion of users (unless they are willing to contribute, and... Since they're paying windows server license anyway I guess they are more financially capable of setting their own system up?)

Skyblock is not a read-only game mode. You would need to reset the world anyway, so I don't see why you are trying to share world folders.



> The problems that I see are:
> 
> 1. Collisions in `Level->getFolderName()`
> 2. No guarantee of readonly access
> 3. Server has no way to know if another server already uses this world (short of server.lock, which covers all server data; this would bypass server.lock)
> 4. Some formats (i.e. leveldb) already disallow multi-process access. Region worlds don't because the implementation sucks and nobody really thought about it. This ties back into 3) you don't know who else, if anyone, is already using the world.
> 
> In principle loading worlds from anywhere is an OK idea, but the rationale doesn't make sense.

Technically speaking, getFolderName() is something we can circumvent by using the full URI instead of just the folder name, making it effectively a path resolved at `${DATA_PATH}/worlds`.

And to be honest, there is a very simple alternative solution... Just reinvent the wheel of symlinks if you're on Windows. Make a regular file with the same format as Unix symlinks (i.e. the file contents contain exactly the path to the linked world), then add a special LinkLebelProvider which delegates to the backing path. Simple. No need for changes on PocketMine's side.
Except... Come on, use symlinks, really. That's what they're designed for.

dktapps:
> Technically speaking, getFolderName() is something we can circumvent by using the full URI instead of just the folder name, making it effectively a path resolved at ${DATA_PATH}/worlds.

Yes, but not without a BC break. Many plugins rely on getFolderName() to return a simple identifier which doesn't change, because it's commonly used for data storage.

dktapps:
> Make a regular file with the same format as Unix symlinks (i.e. the file contents contain exactly the path to the linked world), then add a special LinkLebelProvider which delegates to the backing path. Simple. No need for changes on PocketMine's side.

I think this settles it; it doesn't require any core changes and sidesteps the getFolderName() issue.

SOF3:
> > Technically speaking, getFolderName() is something we can circumvent by using the full URI instead of just the folder name, making it effectively a path resolved at ${DATA_PATH}/worlds.
> 
> Yes, but not without a BC break. Many plugins rely on getFolderName() to return a simple identifier which doesn't change, because it's commonly used for data storage.

It doesn't break most stuff. It breaks for people who rely it for constructing file names or other stuff that disallow slashes, colons, etc. But it doesn't break storage format since URI is still a string. The normal folders can continue to have their short name.


> > Make a regular file with the same format as Unix symlinks (i.e. the file contents contain exactly the path to the linked world), then add a special LinkLebelProvider which delegates to the backing path. Simple. No need for changes on PocketMine's side.
> 
> I think this settles it; it doesn't require any core changes and sidesteps the getFolderName() issue.

I guess we can close this issue then.

dktapps:
> It doesn't break most stuff. It breaks for people who rely it for constructing file names or other stuff that disallow slashes, colons, etc. But it doesn't break storage format since URI is still a string. The normal folders can continue to have their short name.

It breaks for plugins that assume getFolderName() on version X can be used in getWorldByName() in version X+1. e.g. saved positions in playerdata (SpawnLevel uses folder name; all player spawn points would break if getFolderName() started returning something else).

PR #3664, opened 2020-06-30
Add Releasable interface

Govdim:
## Introduction
Fixed using items by adding a Releasable interface

## Tests
In my Ubuntu 18.04

### DIFF
--- a/src/item/Bow.php
+++ b/src/item/Bow.php
@@ -34,7 +34,7 @@ use pocketmine\world\sound\BowShootSound;
 use function intdiv;
 use function min;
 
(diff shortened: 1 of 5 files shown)


dktapps:
This looks good! although I'm not sure `Useable` is the best name, I think it's a bit ambiguous.

Govdim:
> This looks good! although I'm not sure `Useable` is the best name, I think it's a bit ambiguous.

Suggest a different name, I will change

dktapps:
What do you think of `Releasable`?

PR #3708, opened 2020-07-16
Fix for PlayerItemConsumeEvent

### LINKED ISSUE #3707: Cancelling PlayerItemConsumeEvent will still give effects
### Issue description

- Expected result: When cancelling the event (e.g. Gold Apple) should not give extra gold hearts.
- Actual result: Cancelling will stop the item getting removed/stack count getting lowered, but not stop giving the player the effect

1. Cancel PlayerItemConsumeEvent
2. Eat golden apple
3. You will still receive the effects, without item count going down/item getting removed

### OS and versions

* PocketMine-MP: PocketMine-MP 3.14.2+dev
* PHP: 7.3
* Server OS: Windows 10

sero583:
Remove effects which may get applied when consuming something (e.g. golden apple), when event has been cancelled.

## Introduction

Cause it is a expected behaviour, when cancelling an event that it should fall back to it's origin state.

### Relevant issues
* Fixes #3707

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -2577,6 +2577,10 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 								}
 								$ev->call();
 								if($ev->isCancelled() or !$this->consumeObject($slot)){
+									foreach($slot->getAdditionalEffects() as $effect) {
+										$this->removeEffect($effect->getId());
+									}
+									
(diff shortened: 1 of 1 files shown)


dktapps:
`consumeObject` is not called if the event is cancelled, so this is impossible anyway.

sero583:
I understand, thought it may be it. Anyway, would it be a good idea to re-send some packets so the client actually knows for sure it should not do that?

PR #3709, opened 2020-07-16
Fix wrong hardness value for Podzol

TheNewHEROBRINEX:
## Introduction

The value of hardness was wrong for the Podzol block resulting in breaking particles not being shown correctly.
### Relevant issues

No relevant issues.

## Changes
This PR changes the hardness value for the Podzol block from 2.5 to 0.5. Source: https://minecraft.gamepedia.com/Podzol

## Tests

I tried breaking blocks of podzol with bare hands both before and after this change. Before the change, the blocks would have been broken before the breaking particles had reached its edges, while after the change they perfectly match client breaking time.

### DIFF
--- a/src/pocketmine/block/Podzol.php
+++ b/src/pocketmine/block/Podzol.php
@@ -40,6 +40,6 @@ class Podzol extends Solid{
 	}
 
 	public function getHardness() : float{
-		return 2.5;
+		return 0.5;
 	}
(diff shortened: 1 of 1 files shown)


dktapps:
Thanks. I bet this stems from the same refactor that caused the issue with redstone and netherrack back in 2015.

PR #3712, opened 2020-07-17
Add 1.16 netherite items

andreashgk:
## Introduction

Pocketmine has not yet implemented any of the new items introduced in 1.16 . This PR adds some of those items to Pocketmine. 
### Relevant issues

/

### API changes

/
### Behavioural changes

/
## Backwards compatibility

/
## Follow-up

The armor does not yet have a working **knockback resistance** value. This can be added in the future.
Requires translations:

### DIFF
--- a/src/pocketmine/item/ItemFactory.php
+++ b/src/pocketmine/item/ItemFactory.php
@@ -279,6 +279,24 @@ class ItemFactory{
 		//TODO: RECORD_WARD
 		//TODO: RECORD_11
 		//TODO: RECORD_WAIT
+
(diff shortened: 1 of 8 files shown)


Shelly7w7:
I think dktapps already said 4.0 is feature-locked, all prs regarding features should be for 4.0

StarsTeams:
We recommend adding new items from the Netherlands so we can relax when creating a server

dktapps:
Stable is in feature freeze. Please target feature additions at PM4 (master).

PR #3723, opened 2020-07-23
Move the call of RespawnEvent 

enricoangelon:
This fix will for example let people to set player's xp on respawn, because before it was resetted to 0 by the attribute map (called after the event).

## Introduction
People can't modify player xp on respawn


dktapps:
I'm unsure of the BC effects this might have. It changes some behaviour.
I don't think this is suitable for a stable release. It would have to be 3.15 or 4.0.

dktapps:
Obsoleted by changes made in #4015

PR #3729, opened 2020-07-25
Introduce a ChunkPos type

dktapps:
## Introduction
Chunk positions are used in many places, and pretty much always as an X/Z pair.
This PR unifies these X/Z coordinate pairs behind a `ChunkPos` type.

This has the following advantages:
- `chunkHash` doesn't need to be recomputed all the time (it's calculated in the constructor and stored on `ChunkPos->hash`.
- No more getting X and Z mixed up, since they are now unified into one type.
- More self-descriptive function signatures.
- Enables getting rid of specializations like `getViewersForPosition()`, since now a `ChunkPos` can be constructed using a `Vector3`.

### API changes
- Almost all cases where a chunk X/Z pair would be accepted as integers, now accept a `ChunkPos` argument instead.

### Behavioural changes
It's possible this might have negative performance impacts - this has not yet been assessed. However, considering that chunk coordinates are not heavily used in the core code, I doubt that the overhead of instantiating the occasional `ChunkPos` object will be a problem.

Reducing the number of calls to `World::blockHash()` should have a marginally positive performance impact, but again I don't believe the difference is significant anyway.

## Backwards compatibility
This is BC-breaking.

### DIFF
--- a/src/block/Block.php
+++ b/src/block/Block.php
@@ -40,6 +40,7 @@ use pocketmine\math\Vector3;
 use pocketmine\nbt\tag\CompoundTag;
 use pocketmine\player\Player;
 use pocketmine\world\BlockTransaction;
+use pocketmine\world\ChunkPos;
 use pocketmine\world\Position;
 use pocketmine\world\World;
 use function assert;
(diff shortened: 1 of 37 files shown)


SOF3:
Is there a reason why `ChunkPos->hash` is public? Is it merely for preventing the performance overhead of function calls? Is such optimization really significant?

dktapps:
Yes, it's for performance. Since Shoghi made efforts to inline World::chunkHash() I was wary to alter the performance of that too much, but it needs to have a proper impact assessment.

SOF3:
This statement is quite contradictory with the claim in "Behavioural changes" above. Benchmarking would be desirable.

dktapps:
Closing this due to being stale and having many merge conflicts.

PR #3742, opened 2020-08-01
Jukebox & Records

JaxkDev:
## Introduction
This PR was mainly so I could ditch my current plugin JukeBox, also gives everyone the relevant items & blocks.
*(Excluding the new `PIGSTEP` record)*

### IMPORTANT NOTE
**Mobile** users **WILL NOT** hear the records being played until they download the `Music` DLC by Mojang in the marketplace.

### Relevant issues
N/A

### API changes

Added world sounds:
- RecordSound
- RecordStopSound

Added Blocks:
- Jukebox

Added Items:
- Record
  - 13
  - Cat
  - Blocks
  - Chirp
  - Far
  - Mall
  - Mellohi
  - Stal
  - Strad
  - Ward
  - 11
  - Wait

Added Enums:
- RecordType
  - DISK_13
  - DISK_CAT
  - DISK_BLOCKS
  - DISK_CHIRP
  - DISK_FAR
  - DISK_MALL
  - DISK_MELLOHI
  - DISK_STAL
  - DISK_STRAD
  - DISK_WARD
  - DISK_11
  - DISK_WAIT

Added functions:
 - `Player->sendJukeboxPopup(string key, string[] args)`
 - `NetworkSession->onJukeboxPopup(string key, string[] args)`

### Behavioural changes
N/A

## Backwards compatibility
N/A

## Tests
[Travis log](https://api.travis-ci.com/v3/job/367679902/log.txt)
All tests completed via travis-ci.com


dktapps:
You need to run `tests/phpunit/block/regenerate_consistency_check.php` and commit the changes it makes.

JaxkDev:
Changes added, all travis tests passed, [new log](https://api.travis-ci.com/v3/job/367641964/log.txt)

src/block/BlockFactory.php
```diff
@@ -230,6 +231,7 @@ public function getSilkTouchDrops(Item $item) : array{
 		$this->register(new Opaque(new BID(Ids::IRON_ORE), "Iron Ore", new BlockBreakInfo(3.0, BlockToolType::PICKAXE, ToolTier::STONE()->getHarvestLevel())));
 		$this->register(new Trapdoor(new BID(Ids::IRON_TRAPDOOR), "Iron Trapdoor", new BlockBreakInfo(5.0, BlockToolType::PICKAXE, ToolTier::WOOD()->getHarvestLevel(), 25.0)));
 		$this->register(new ItemFrame(new BID(Ids::FRAME_BLOCK, 0, ItemIds::FRAME, TileItemFrame::class), "Item Frame"));
+		$this->register(new Jukebox(new BID(Ids::JUKEBOX, 0, 0, TileJukebox::class), "Jukebox"));
```
dktapps, line 234: Jukebox item ID is not zero. This will cause bugs when breaking jukeboxes.

src/block/tile/Jukebox.php
```diff
@@ -0,0 +1,99 @@
...
+
+use pocketmine\world\World;
+use pocketmine\item\Record;
+use pocketmine\math\Vector3;
+use pocketmine\nbt\tag\CompoundTag;
+use pocketmine\world\sound\RecordSound;
+use pocketmine\world\sound\RecordStopSound;
+
+class Jukebox extends Spawnable{
+	public const TAG_RECORD = "record";
```
dktapps, line 34: This should be `private`. Anything outside of this class has no business using it.
Please also comment next to it what the expected type of tag is.

src/block/tile/Jukebox.php
```diff
@@ -0,0 +1,99 @@
...
+		}
+	}
+
+	protected function writeSaveData(CompoundTag $nbt) : void{
+		if($this->record !== null){
+			$nbt->setTag(self::TAG_RECORD, $this->record->nbtSerialize());
+		}
+	}
+
+	protected function addAdditionalSpawnData(CompoundTag $nbt): void{}
```
dktapps, line 98: You're sure about this?

src/block/tile/Jukebox.php
dktapps: The wiki says that the tag is `RecordItem`. Please check that PC jukeboxes can be decoded correctly.

src/item/Record.php
```diff
@@ -0,0 +1,36 @@
...
+declare(strict_types=1);
+
+namespace pocketmine\item;
+
+class Record extends Item{
+
+	public function __construct(ItemIdentifier $identifier, string $name)
+	{
+		parent::__construct($identifier, $name);
+	}
```
dktapps, line 31: This constructor seems to be redundant.

dktapps (changes requested):

dktapps:
Hi, this looks OK functionally, but it's not quite in line with development practices of PM4.

- Tiles shouldn't do anything more than just handle data storage. Any APIs should be exposed on the corresponding Block class and should give data to the tile in `readStateFromWorld`/`writeStateToWorld` (see for example `ItemFrame`.
- Avoid relying on item IDs or protocol IDs. I recommend creating a `RecordType` enum (like `DyeColor`) and inject the `RecordType` into the item's constructor.

src/block/tile/Jukebox.php
```diff
@@ -0,0 +1,99 @@
...
+use pocketmine\world\sound\RecordStopSound;
+
+class Jukebox extends Spawnable{
+	public const TAG_RECORD = "record";
+
+	/** @var Record|null */
+	private $record = null;
+
+	public function __construct(World $world, Vector3 $pos){
+		parent::__construct($world, $pos);
```
dktapps, line 40: useless constructor here

src/block/tile/Jukebox.php
JaxkDev: ?

src/item/Record.php
JaxkDev: Sorry had them originally pass things but stripped out later.

src/block/tile/Jukebox.php
dktapps: It seems unusual that it wouldn't be needed. Doesn't it show the disc in the slot?

JaxkDev: No, unless the particles is client side *shrug*

src/block/tile/Jukebox.php
JaxkDev: Sorry, just copied other tile format most are public with no type.

JaxkDev:
Love you too phpstan :/

src/block/Jukebox.php
```diff
@@ -0,0 +1,115 @@
...
+		return 300;
+	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
+		if($player instanceof Player){
+			if($this->hasRecord()){
+				$this->ejectRecord();
+			} else {
+				if($item instanceof Record) {
+					$player->getInventory()->remove($item);
```
dktapps, line 52: This will remove **all** record items of this type from the player's inventory. Probably not what you intended.

src/block/Jukebox.php
```diff
@@ -0,0 +1,115 @@
...
+
+		return true;
+	}
+
+	public function hasRecord() : bool{
+		return $this->record !== null;
+	}
+
+	public function ejectRecord() : void{
+		if($this->hasRecord()){
```
dktapps, line 68: Use `record !== null` here. It's better for static analysis.

src/block/Jukebox.php
```diff
@@ -0,0 +1,115 @@
...
+		}
+	}
+
+	public function stopSound() : void{
+		$this->getPos()->getWorld()->addSound($this->getPos(), new RecordStopSound());
+	}
+
+	public function onBreak(Item $item, ?Player $player = null): bool
+	{
+		$this->ejectRecord();
```
dktapps, line 94: It might be better to just add this to the result of `getDrops()` so that plugins can see it in `BlockBreakEvent`.

dktapps (commented):
This is starting to look better :thumbsup:

src/block/Jukebox.php
```diff
@@ -89,10 +89,8 @@ public function stopSound() : void{
 		$this->getPos()->getWorld()->addSound($this->getPos(), new RecordStopSound());
 	}
 
-	public function onBreak(Item $item, ?Player $player = null): bool
-	{
-		$this->ejectRecord();
-		return parent::onBreak($item, $player);
+	public function getDrops(Item $item): array{
+		return $this->record !== null ? [$this->record] : [];
```
dktapps, line 93: This breaks the regular jukebox block drop.

dktapps: You also don't want to return the original item, since the caller might modify it.

JaxkDev: I will get there 😊

src/block/Jukebox.php
```diff
@@ -90,7 +90,7 @@ public function stopSound() : void{
 	}
 
 	public function getDrops(Item $item): array{
```
dktapps, line 92: This really ought to use `getDropsForCompatibleTool()`, in case a plugin changes the tool requirements.

src/item/Record.php
```diff
@@ -0,0 +1,44 @@
...
+
+namespace pocketmine\item;
+
+use pocketmine\block\utils\RecordType;
+
+class Record extends Item{
+	/** @var RecordType */
+	private $recordType;
+
+	public function __construct(ItemIdentifier $identifier, RecordType $recordType, string $name = "Unknown"){
```
dktapps, line 32: I don't think it makes sense to allow the name to be optional here. It's just asking for bugs.

dktapps:
I'll review this again tomorrow.

JaxkDev:
Reference: LevelSoundEventPacket pigste[ record sound ID is 314

With that i can fully implement the pigstep record to complete all records but its very new and haven't seen any new items or constants is this for a reason ?

dktapps:
Avoid introducing any items that were added in 1.13+. World save handling might get real complicated if we introduce those things to an old world format.

dktapps:
In-game testing revealed the following problems:

- [x] Note particles weren't being displayed while a disk was active. (Missing client-sided blockentity.)
- [x] Records keep playing when the jukebox playing them is destroyed
- [ ] Records are stopped when any jukebox is interacted with or destroyed, regardless of whether the target jukebox is the one currently playing a disc.
- [ ] `Now playing` tip is not displayed
- [x] Disc items were consumed in creative mode
- [ ] Break info is wrong - breaking animation doesn't match when broken by hand (unsure why, haven't investigated).

dktapps:
Seems like the hardness is 0.8 on MCPE and not 2.0.

JaxkDev:
Will update PR info to reflect latest commit changes, Cannot use the record name translations for some reason *shrug* i assume its because its a parameter for the jukebox ('playing now') translation

PR #3744, opened 2020-08-02
Better crash report

GamakCZ:
## Introduction

Developers aren't able to find exact entity class after the crash...

### Behavioural changes

Better crash report when server crashes because of unregistered entity

## Tests- Better crash report when server crashes because of unregistered entity
 Tested


dktapps:
This looks OK, but if you just tell me "tested" again I won't be so forgiving. I'll give it a pass this time since the change is trivial and the CI is green.

> Details should be provided of tests done. Simply saying "tested" or equivalent is not acceptable.

PR #3747, opened 2020-08-04
Use objects for internal structures created in TextFormat::toJSON()

dktapps:
## Introduction
I stumbled across some difficult-to-navigate code today as I was playing with some new PHPStan features and decided it would be a good idea to replace some assoc arrays with objects. Since it's impractical to define recursive assoc array structures with PHPStan, objects were the only option.
### API changes
No API changes.

### Behavioural changes
Behaviour should be the same as before, barring any bugs which I haven't yet tested for.

## Backwards compatibility
Fully backwards compatible.

## Tests
Not yet tested, but it passes a PHPStan run.

### DIFF
--- a/src/pocketmine/timings/Timings.php
+++ b/src/pocketmine/timings/Timings.php
@@ -149,7 +149,7 @@ abstract class Timings{
 	}
 
 	public static function getScheduledTaskTimings(TaskHandler $task, int $period) : TimingsHandler{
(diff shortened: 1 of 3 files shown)


dktapps:
An additional commit shows up in the diff which was pushed to this branch prior to stable. It won't go away seemingly because of some kind of GitHub glitch.

PR #3750, opened 2020-08-05
Emerald ore in Mountains biome

IceCruelStuff:
## Introduction

Adds emerald ore in mountains biome.

### DIFF
--- a/src/world/biome/MountainsBiome.php
+++ b/src/world/biome/MountainsBiome.php
@@ -41,12 +41,13 @@ class MountainsBiome extends GrassyBiome{
 		$tallGrass = new TallGrass();
 		$tallGrass->setBaseAmount(1);
 
+		$this->addPopulator($tallGrass);
+
 		$ores = new Ore();
 		$ores->setOreTypes([
 			new OreType(VanillaBlocks::EMERALD_ORE(), 11, 1, 0, 32)
 		]);
 
-		$this->addPopulator($tallGrass);
 		$this->addPopulator($ores);
 
 		$this->setElevation(63, 127);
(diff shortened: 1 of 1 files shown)


src/world/biome/MountainsBiome.php
```diff
@@ -38,9 +41,13 @@ public function __construct(){
 		$tallGrass = new TallGrass();
 		$tallGrass->setBaseAmount(1);
 
-		$this->addPopulator($tallGrass);
+		$ores = new Ore();
+		$ores->setOreTypes([
+			new OreType(VanillaBlocks::EMERALD_ORE(), 11, 1, 0, 32)
+		]);
```
dktapps, line 47: please move these lines below the tallGrass line

dktapps:
This looks OK now. Can I ask where you got 11 for the cluster size?

There's a minor problem here because this PR can't be tested because of [this bug](https://github.com/pmmp/pocketmine-mp/blob/5a56f68991eaf6781c129174e3cf057f813249ce/src/world/generator/normal/Normal.php#L99). I'd like to see the results of this PR because I don't understand why @shoghicp didn't add this directly if it's this simple.

IceCruelStuff:
Isn't 11 for cluster count?
https://github.com/pmmp/PocketMine-MP/blob/master/src/world/generator/object/OreType.php#L42

IceCruelStuff:
I got all the values here
https://minecraft.gamepedia.com/Emerald_Ore#Natural_generation

dktapps:
Yeah, I meant count, my bad. (Some part of my brain insists on flipping words that it shouldn't.)

IceCruelStuff:
For the cluster count, I got it from the [Minecraft wiki](https://minecraft.gamepedia.com/Emerald_Ore#Natural_generation). It says emerald ore will attempt to generate 11 times per chunk.

PR #3772, opened 2020-08-12
Remove 'reload' command

Xenophilicy:
## Introduction

Plugin developers constantly get useless issues thrown at them when users execute `/reload` and tell them they "restarted the server" and nothing works. This will mitigate the issue by removing the ability to reload the server in the first place.

### Relevant issues

None in PMMP

### API changes

No, this was never in the documentation other than in the `pocketmine\command\defaults` namespace.

### Behavioural changes

The server should behave the exact same and without any more reloading issues related to plugins.

## Backwards compatibility

The only backwards compatibility issues would be server owners implementing this command into their server. In which case, they should just remove the command from their server as a fix.

## Follow-up

Warn people and tell them that this was necessary due to invalid usage of the command.

## Tests
The server runs from the branch with no issues (because hardly anything was changed) and the command shows up as "unknown" which is the desired behavior. Since all I did was remove a command, there should be no further testing required.


ZizzyZizzy:
I second this addition.

We've resorted to re-mapping the reload command to the restart command in the Minecraft control panel to prevent server owners from shooting themselves in the foot with this dangerous command.

jasonwynn10:
API 3 is in feature freeze so this will be rejected, and API 4 already had this done a long time ago. Please do more research before creating PRs.

Xenophilicy:
Don't take everything so seriously lol, it was clearly a joke if you read on Discord :)

dktapps:
A non-API-breaking way to do this would be to just not register the command. There's a few things that might break if it was removed entirely (unknown command permission, plugins trying to use ReloadCommand for some reason, etc).

dktapps:
I should also mention that it's still possible to "reload" the server using Server->reload(), so this still doesn't make the problem go away fully.

PR #3773, opened 2020-08-13
Changed Minecraft: PE to Minecraft: BE in ProtocolInfo

ipad54:
Changed Minecraft:PE to Minecraft: BE in ProtcolInfo


src/pocketmine/network/mcpe/protocol/ProtocolInfo.php
```diff
@@ -36,9 +36,9 @@ interface ProtocolInfo{
 	 * This file is generated automatically, do not edit it manually.
```
dktapps, line 36: did you not read this line?

dktapps:
This change is completely superfluous. The game itself is still referred to as MCPE internally.

PR #3779, opened 2020-08-18
Added ApiMap (basic dependency injection container)

### LINKED ISSUE #3087: Niche Registry
A niche refers to a role where exactly one implementation is required.

This issue proposes that a generic API be added to allow plugins to require/provide niches from other modules, with an intelligent mechanism of resolving conflicts.

Examples of niches:
- Certain interface implementations are niches. For example, if we allow customization of ban list implementations or player data providers, there can only be exactly one niche.
- Each command name is a niche. While we have fallback prefixes, [almost nobody understands/cares/uses them](https://forums.pmmp.io/threads/enforcement-of-rule-c2a-fallback-prefix.6995/).

For the former type, it can easily be added with an interface-based registry-style mapping object. Plugins can collect the required niche impl in some post-enable stage (after all plugins have been enabled). If competing impls are found, the server suspends, and the server owner is explicitly requested to select an impl on console for the server to continue running. (As a result, niche registration is restricted to startup)

I have no intention to mess up with the commands API, but they might reuse the same mechanism to resolve conflicts, eventually eliminating the need for fallback prefixes. As for commands registered beyond server start (such as some plugin that allows users to add new commands), they might want to intercept `CommandEvent` rather than directly registering commands. (Sounds like a bad idea. Any suggestions?)

It is more practical to allow the server owner to decide which implementation to use, compared with letting the plugins fight with each other internally.
I am not on the side for allowing plugins to forcefully override another implementation, because that is commonly associated with bloated plugins that introduce a bunch of features, such as EssentialsPE trying to override /gamemode, /msg and /kill. (This is so brutal and uncivilized!) Troubling the user to select an implementation is a nice sign to remind the user that an existing feature will be backward-incompatibly changed.

SOF3:
## Introduction
This provides a type-based approach for plugins to expose APIs to each other.

For example, PocketMine can provide a `BanList` interface with a default implementation (using the current banned-players.txt etc), then plugins can set it to their own implementation.

This class also handles conflict conditions.

For details, please check the documentation in the code.

### Relevant issues
Resolves #3087

### API changes
Added the `ApiMap` class.

Its methods are directly re-exported in `Server`.

### Behavioural changes
None yet.

## Backwards compatibility
This is a pure API addition with no BC issues.

## Follow-up
Improve error messages.

## Tests
See `ApiMapTest`.

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -69,6 +69,7 @@ use pocketmine\player\GameMode;
 use pocketmine\player\OfflinePlayer;
 use pocketmine\player\Player;
 use pocketmine\player\PlayerInfo;
+use pocketmine\plugin\ApiMap;
 use pocketmine\plugin\PharPluginLoader;
 use pocketmine\plugin\Plugin;
 use pocketmine\plugin\PluginEnableOrder;
@@ -182,6 +183,9 @@ class Server{
 
 	/** @var PluginManager */
 	private $pluginManager;
+
+	/** @var ApiMap */
+	private $apiMap;
 
 	/** @var float */
 	private $profilingTickRate = 20;
@@ -420,6 +424,80 @@ class Server{
 	 */
 	public function getPluginManager(){
 		return $this->pluginManager;
+	}
+	
+	/**
+	 * Returns the underlying server-scoped API map
+	 */
+	public function getApiMap() : ApiMap{
+		return $this->apiMap;
+	}
+
+	/**
+	 * Provides an API implementation of $interface.
+	 *
+	 * $interface can be either an (abstract, open or final) class or interface.
+	 * $impl must implement $interface.
+	 *
+	 * The $interface is used to identify various API types,
+	 * and users should not try to provide APIs for two $interfaces that extend one another.
+	 * For example, PocketMine can provide a default BanList interface provides an interface to track user bans,
+	 * which calls `provideApi(BanList::class)`, but it does not call `provideApi(DefaultBanList::class)`.
+	 * Nevertheless, if alternative implementations are not intended,
+	 * it is fine to provide $interface as the final class.
+	 * (TODO: change "can provide" to "provides" when the BanList API is changed)
+	 *
+	 * If the declaration of `$interface` has a `@purpose` tag in its documentation,
+	 * it is provided as part of the error message to describe the purpose of the interface
+	 * when two conflicting implementations are provided,
+	 * e.g. `@purpose ban list` would result in the error:
+	 *
+	 * > Multiple plugins are providing ban list. Please disable one of them or check configuration
+	 *
+	 * The default implementation (usually provided by the module declaring the interface)
+	 * should call `provideDefaultApi` so that other plugins can override it without triggering errors.
+	 *
+	 * @phpstan-template T of object
+	 * @phpstan-param class-string<T> $interface
+	 * @phpstan-param T $impl
+	 *
+	 * @throws \InvalidArgumentException if $impl is not an instance of $interface
+	 * @throws \RuntimeException if two non-default APIs are provided for the same interface
+	 *
+	 * @see Server::provideDefaultApi()
+	 */
+	public function provideApi(string $interface, Plugin $plugin, object $impl) : void{
+		$this->apiMap->provideApi($interface, $plugin, $impl, false);
+	}
+
+	/**
+	 * Provides a *default* API implementation of $interface.
+	 *
+	 * `provideDefaultApi` must only be called exactly once, by the module that declared `$interface`.
+	 *
+	 * @phpstan-template T of object
+	 * @phpstan-param class-string<T> $interface
+	 * @phpstan-param T $impl
+	 *
+	 * @throws \InvalidArgumentException if $impl is not an instance of $interface
+	 *
+	 * @see Server::provideApi() for detailed semantics of API provision
+	 */
+	public function provideDefaultApi(string $interface, Plugin $plugin, object $impl) : void{
+		$this->apiMap->provideApi($interface, $plugin, $impl, true);
+	}
+
+	/**
+	 * Retrieves the current implementation of `getApi`.
+	 *
+	 * Callers can check whether this implementation is default by getting `$default` by reference.
+	 *
+	 * @phpstan-template T of object
+	 * @phpstan-param class-string<T> $interface
+	 * @phpstan-return T|null
+	 */
+	public function getApi(string $interface, bool &$default = false) : ?object{
+		return $this->apiMap->getApi($interface, $default);
 	}
 
 	/**
@@ -958,6 +1036,8 @@ class Server{
 			$this->pluginManager = new PluginManager($this, ((bool) $this->configGroup->getProperty("plugins.legacy-data-dir", true)) ? null : $this->getDataPath() . "plugin_data" . DIRECTORY_SEPARATOR, $pluginGraylist);
 			$this->pluginManager->registerInterface(new PharPluginLoader($this->autoloader));
 			$this->pluginManager->registerInterface(new ScriptPluginLoader());
+
+			$this->apiMap = new ApiMap;
 
 			$providerManager = new WorldProviderManager();
 			if(
--- a/src/plugin/ApiMap.php
+++ b/src/plugin/ApiMap.php
@@ -1,0 +1,104 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * @author PocketMine Team
+ * @link http://www.pocketmine.net/
+ *
+ *
+*/
+
+declare(strict_types=1);
+
+namespace pocketmine\plugin;
+
+use function get_class;
+use function is_string;
+use InvalidArgumentException;
+use ReflectionClass;
+use RuntimeException;
+use pocketmine\utils\Utils;
+
+/**
+ * Alows different modules to expose APIs using types.
+ */
+final class ApiMap{
+	/**
+	 * @var ApiMapEntry[]
+	 * @phpstan-var array<class-string, ApiMapEntry>
+	 */
+	private $apiMap = [];
+
+	/**
+	 * @see Server::provideApi()
+	 *
+	 * @phpstan-template T of object
+	 * @phpstan-param class-string<T> $interface
+	 * @phpstan-param T $impl
+	 *
+	 * @throws InvalidArgumentException if $impl is not an instance of $interface
+	 * @throws RuntimeException if two non-default APIs are provided for the same interface
+	 */
+	public function provideApi(string $interface, ?Plugin $plugin, object $impl, bool $default = false) : void{
+		if(!($impl instanceof $interface)){
+			$class = get_class($impl);
+			throw new InvalidArgumentException("\$impl is an instance of $class, which does not extend/implement $interface");
+		}
+
+		if(isset($this->apiMap[$interface])){
+			if(!$this->apiMap[$interface]->default && !$default) {
+				// two non-default implementations
+				$otherPlugin = $this->apiMap[$interface]->plugin;
+				$otherPluginName = $otherPlugin !== null ? $otherPlugin->getName() : "PocketMine";
+				$pluginName = $plugin !== null ? $plugin->getName() : "PocketMine";
+
+				$class = new ReflectionClass($interface);
+				$doc = $class->getDocComment();
+				$tags = is_string($doc) ? Utils::parseDocComment($doc) : [];
+				$purpose = $tags["purpose"] ?? "an implementation of $interface";
+
+				// TODO switch to user-friendly exceptions
+				throw new RuntimeException("Multiple plugins ($otherPluginName, $pluginName) are providing $purpose. Please disable one of them or check configuration.");
+			}
+
+			if($default) {
+				return;
+			}
+		}
+
+		$entry = new ApiMapEntry;
+		$entry->plugin = $plugin;
+		$entry->impl = $impl;
+		$entry->default = $default;
+		$this->apiMap[$interface] = $entry;
+	}
+
+	/**
+	 * @see Server::getApi()
+	 *
+	 * @phpstan-template T of object
+	 * @phpstan-param class-string<T> $interface
+	 *
+	 * @phpstan-return T|null
+	 */
+	public function getApi(string $interface, bool &$default = false) : ?object {
+		if(!isset($this->apiMap[$interface])) {
+			return null;
+		}
+		$default = $this->apiMap[$interface]->default;
+		/** @phpstan-var T $impl */
+		$impl = $this->apiMap[$interface]->impl;
+		return $impl;
+	}
+}
--- a/src/plugin/ApiMapEntry.php
+++ b/src/plugin/ApiMapEntry.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * @author PocketMine Team
+ * @link http://www.pocketmine.net/
+ *
+ *
+*/
+
+declare(strict_types=1);
+
+namespace pocketmine\plugin;
+
+/**
+ * @internal
+ */
+final class ApiMapEntry{
+	/** @var Plugin|null */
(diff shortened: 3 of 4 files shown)


src/ApiMap.php
```diff
@@ -0,0 +1,112 @@
...
+use pocketmine\plugin\Plugin;
+use pocketmine\utils\Utils;
+
+/**
+ * Alows different modules to expose APIs using types.
+ */
+final class ApiMap{
+	/**
+	 * @var object[][]|Plugin[][]|null[][]
+	 * @phpstan-var array<class-string, array{plugin: Plugin|null, impl: object, default: bool}>
```
dktapps, line 40: I'd prefer a class was added for this. It's easier to reason around.

src/ApiMap.php
```diff
@@ -0,0 +1,112 @@
...
+	 * @phpstan-param class-string<T> $interface
+	 * @param Plugin|null $plugin
+	 * @param object $impl
+	 * @phpstan-param T $impl
+	 * @param bool $default
+	 *
+	 * @throws InvalidArgumentException if $impl is not an instance of $interface
+	 * @throws RuntimeException if two non-default APIs are provided for the same interface
+	 */
+	public function provideApi(string $interface, ?Plugin $plugin, object $impl, bool $default = false) : void{
```
dktapps, line 58: A `provideDefaultApi()` wrapper would be better for readability than boolean parameters.

dktapps (commented):
I haven't reviewed this in detail, just made a couple of comments.

src/ApiMap.php
SOF3: Is that really necessary for a type only used internally within the class?

src/Server.php
```diff
@@ -441,20 +441,40 @@ public function getPluginManager(){
...
 	 * @param object $impl
 	 * @phpstan-param T $impl
-	 * @param bool $default
 	 *
 	 * @throws \InvalidArgumentException if $impl is not an instance of $interface
 	 * @throws \RuntimeException if two non-default APIs are provided for the same interface
+	 *
+	 * @see #provideDefaultApi
 	 */
 	public function provideApi(string $interface, Plugin $plugin, object $impl, bool $default = false) : void{
```
dktapps, line 458: forgot something?

src/Server.php
```diff
@@ -416,6 +420,80 @@ public function getPluginManager(){
...
+	 * Callers can check whether this implementation is default by getting `$default` by reference.
+	 *
+	 * @template T of object
+	 * @param string $interface
+	 * @phpstan-param class-string<T> $interface
+	 * @param bool $default
+	 * @return object|null
+	 * @phpstan-return T|null
+	 */
+	public function getApi(string $interface, bool &$default = false) : ?object {
```
DerDevHD, line 493: Why is there an unsed parameter? _$default_ is never used

SOF3: Oops right, it's supposed to passthru to $default.

DerDevHD: Will you update or should I make a PR?

SOF3: just fyi, there is a "suggest change" button on GitHub, but I'm going to edit this myself for now

DerDevHD: Okay, fine.

src/Server.php
```diff
@@ -416,6 +420,80 @@ public function getPluginManager(){
...
+	 * when two conflicting implementations are provided,
+	 * e.g. `@purpose ban list` would result in the error:
+	 *
+	 * > Multiple plugins are providing ban list. Please disable one of them or check configuration
+	 *
+	 * The default implementation (usually provided by the module declaring the interface)
+	 * should call `provideDefaultApi` so that other plugins can override it without triggering errors.
+	 *
+	 * @template T of object
+	 * @param string $interface
```
dktapps, line 448:
```suggestion
```

src/Server.php
```diff
@@ -416,6 +420,80 @@ public function getPluginManager(){
...
+	 * @phpstan-param class-string<T> $interface
+	 * @param object $impl
```
dktapps, line 450:
```suggestion
```

src/Server.php
```diff
@@ -416,6 +420,80 @@ public function getPluginManager(){
...
+		$this->apiMap->provideApi($interface, $plugin, $impl, false);
+	}
+
+	/**
+	 * Provides a *default* API implementation of $interface.
+	 *
+	 * `provideDefaultApi` must only be called exactly once, by the module that declared `$interface`.
+	 *
+	 * @template T of object
+	 * @param string $interface
```
dktapps, line 468:
```suggestion
```

src/Server.php
```diff
@@ -416,6 +420,80 @@ public function getPluginManager(){
...
+	 * @phpstan-param class-string<T> $interface
+	 * @param object $impl
```
dktapps, line 470:
```suggestion
```

src/Server.php
```diff
@@ -416,6 +420,80 @@ public function getPluginManager(){
...
+		$this->apiMap->provideApi($interface, $plugin, $impl, true);
+	}
+
+	/**
+	 * Retrieves the current implementation of `getApi`.
+	 *
+	 * Callers can check whether this implementation is default by getting `$default` by reference.
+	 *
+	 * @template T of object
+	 * @param string $interface
```
dktapps, line 487:
```suggestion
```

src/Server.php
```diff
@@ -416,6 +420,80 @@ public function getPluginManager(){
...
+	 * @phpstan-param class-string<T> $interface
+	 * @param bool $default
```
dktapps, line 489:
```suggestion
```

src/Server.php
```diff
@@ -416,6 +420,80 @@ public function getPluginManager(){
...
+	 * @return object|null
```
dktapps, line 490:
```suggestion
```

src/Server.php
dktapps, line 493:
```suggestion
	public function getApi(string $interface, bool &$default = false) : ?object{
```

src/Server.php
```diff
@@ -416,6 +420,80 @@ public function getPluginManager(){
...
+	 * it is provided as part of the error message to describe the purpose of the interface
+	 * when two conflicting implementations are provided,
+	 * e.g. `@purpose ban list` would result in the error:
+	 *
+	 * > Multiple plugins are providing ban list. Please disable one of them or check configuration
+	 *
+	 * The default implementation (usually provided by the module declaring the interface)
+	 * should call `provideDefaultApi` so that other plugins can override it without triggering errors.
+	 *
+	 * @template T of object
```
dktapps, line 447:
```suggestion
	 * @phpstan-template T of object
```

src/Server.php
```diff
@@ -416,6 +420,80 @@ public function getPluginManager(){
...
+	public function provideApi(string $interface, Plugin $plugin, object $impl) : void{
+		$this->apiMap->provideApi($interface, $plugin, $impl, false);
+	}
+
+	/**
+	 * Provides a *default* API implementation of $interface.
+	 *
+	 * `provideDefaultApi` must only be called exactly once, by the module that declared `$interface`.
+	 *
+	 * @template T of object
```
dktapps, line 467:
```suggestion
	 * @phpstan-template T of object
```

src/plugin/ApiMap.php
```diff
@@ -0,0 +1,111 @@
...
+	/**
+	 * @var ApiMapEntry[]
+	 * @phpstan-var array<class-string, ApiMapEntry>
+	 */
+	private $apiMap = [];
+
+	/**
+	 * @see Server#provideApi
+	 *
+	 * @template T of object
```
dktapps, line 46:
```suggestion
	 * @phpstan-template T of object
```

src/plugin/ApiMap.php
```diff
@@ -0,0 +1,111 @@
...
+	 * @param string $interface
```
dktapps, line 47:
```suggestion
```

src/plugin/ApiMap.php
```diff
@@ -0,0 +1,111 @@
...
+	 * @phpstan-param class-string<T> $interface
+	 * @param Plugin|null $plugin
```
dktapps, line 49:
```suggestion
```

src/plugin/ApiMap.php
```diff
@@ -0,0 +1,111 @@
...
+ */
+final class ApiMap{
+	/**
+	 * @var ApiMapEntry[]
+	 * @phpstan-var array<class-string, ApiMapEntry>
+	 */
+	private $apiMap = [];
+
+	/**
+	 * @see Server#provideApi
```
dktapps, line 44:
```suggestion
	 * @see Server::provideApi()
```
Also, this won't be properly resolved because `Server` isn't imported.

src/plugin/ApiMap.php
```diff
@@ -0,0 +1,111 @@
...
+	 *
+	 * @template T of object
+	 * @param string $interface
+	 * @phpstan-param class-string<T> $interface
+	 * @param Plugin|null $plugin
+	 * @param object $impl
```
dktapps, line 50:
```suggestion
```

src/plugin/ApiMap.php
```diff
@@ -0,0 +1,111 @@
...
+	 * @phpstan-param T $impl
+	 * @param bool $default
```
dktapps, line 52:
```suggestion
```

src/plugin/ApiMap.php
```diff
@@ -0,0 +1,111 @@
...
+
+		$entry = new ApiMapEntry;
+		$entry->plugin = $plugin;
+		$entry->impl = $impl;
+		$entry->default = $default;
+		$this->apiMap[$interface] = $entry;
+	}
+
+	/**
+	 * @see Server#getApi
```
dktapps, line 92:
```suggestion
	 * @see Server::getApi()
```
Same as above.

src/plugin/ApiMap.php
```diff
@@ -0,0 +1,111 @@
...
+	 *
+	 * @template T of object
```
dktapps, line 94:
```suggestion
	 * @phpstan-template T of object
```

src/plugin/ApiMap.php
```diff
@@ -0,0 +1,111 @@
...
+	 * @param string $interface
```
dktapps, line 95:
```suggestion
```

src/plugin/ApiMap.php
```diff
@@ -0,0 +1,111 @@
...
+	 * @phpstan-param class-string<T> $interface
+	 * @param bool $default
```
dktapps, line 97:
```suggestion
```

src/plugin/ApiMap.php
```diff
@@ -0,0 +1,111 @@
...
+	 * @return object|null
```
dktapps, line 98:
```suggestion
	 *
```

src/plugin/ApiMap.php
```diff
@@ -0,0 +1,111 @@
...
+	 * @phpstan-return T|null
+	 */
+	public function getApi(string $interface, bool &$default = false) : ?object {
+		if(!isset($this->apiMap[$interface])) {
+			return null;
+		}
+		$default = $this->apiMap[$interface]->default;
+		/** @var T $impl */
```
dktapps, line 106:
```suggestion
		/** @phpstan-var T $impl */
```

src/plugin/ApiMapEntry.php
```diff
@@ -0,0 +1,36 @@
...
+ *
+ *
+*/
+
+declare(strict_types=1);
+
+namespace pocketmine\plugin;
+
+/**
+ * @internal This is an internal, semver-exempt class.
```
dktapps, line 27:
```suggestion
 * @internal
```

src/plugin/ApiMapEntry.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+final class ApiMapEntry{
+	/** @var Plugin|null */
+	public $plugin;
+	/** @var object */
```
dktapps, line 32: I guess this whole class should be generic.

tests/phpunit/plugin/ApiMapTest.php
```diff
@@ -0,0 +1,79 @@
...
+namespace pocketmine\plugin;
+
+use InvalidArgumentException;
+use PHPUnit\Framework\TestCase;
+use RuntimeException;
+use stdClass;
+
+/**
+ * @purpose (api map test purpose)
+ */
```
dktapps, line 33:
```suggestion
```

src/Server.php
```diff
@@ -416,6 +420,80 @@ public function getPluginManager(){
...
+	public function provideDefaultApi(string $interface, Plugin $plugin, object $impl) : void{
+		$this->apiMap->provideApi($interface, $plugin, $impl, true);
+	}
+
+	/**
+	 * Retrieves the current implementation of `getApi`.
+	 *
+	 * Callers can check whether this implementation is default by getting `$default` by reference.
+	 *
+	 * @template T of object
```
dktapps, line 486:
```suggestion
	 * @phpstan-template T of object
```

src/Server.php
```diff
@@ -416,6 +420,80 @@ public function getPluginManager(){
...
+	 * @param string $interface
+	 * @phpstan-param class-string<T> $interface
+	 * @param object $impl
+	 * @phpstan-param T $impl
+	 *
+	 * @throws \InvalidArgumentException if $impl is not an instance of $interface
+	 *
+	 * @see #provideApi for detailed semantics of API provision
```
dktapps, line 475:
```suggestion
	 * @see Server::provideApi() for detailed semantics of API provision
```

src/Server.php
```diff
@@ -416,6 +420,80 @@ public function getPluginManager(){
...
+	 * @template T of object
+	 * @param string $interface
+	 * @phpstan-param class-string<T> $interface
+	 * @param object $impl
+	 * @phpstan-param T $impl
+	 *
+	 * @throws \InvalidArgumentException if $impl is not an instance of $interface
+	 * @throws \RuntimeException if two non-default APIs are provided for the same interface
+	 *
+	 * @see #provideDefaultApi
```
dktapps, line 456:
```suggestion
	 * @see Server::provideDefaultApi()
```

tests/phpunit/plugin/ApiMapTest.php
dktapps: I'm not entirely sure what the point of this doc comment is, but using a test class to test itself is a bit dubious ...

dktapps:
Is there a reason to not just add a `getApiMap()` method to `Server()` instead of adding a bunch of proxy methods?

dktapps (commented):
I'm also not sure what the point of "default" APIs is. Allowing there to be a difference just allows plugins to fight with each other, which seems to defeat the objective.

dktapps:
I also have my doubts that a centralized approach for this is necessary or needed. Some plugins already have their own ways to do this with different behaviour.

tests/phpunit/plugin/ApiMapTest.php
SOF3: Again, I was just lazy to create another class. But this looks fine to me since ApiMap does not have any requirements on `$interface`. I would say no if this forces the class to implement a certain interface or something.

SOF3:
> 
> 
> I also have my doubts that a centralized approach for this is necessary. Some plugins already have their own ways to do this with different behaviour.

This is a move away from singleton approach. I am hoping to provide this as the basis for better DI pattern that can replace stuff like `EconomyAPI::getInstance()` with `$server->getApi(EconomyApi::class)`.

I think centralization is necessary since I plan to implement this with stuff like `PlayerGrayList` (whitelist + ban list) in the future.

SOF3:
> Is there a reason to not just add a `getApiMap()` method to `Server()` instead of adding a bunch of proxy methods?

Just for convenience. For now I don't see any actual reason to expose the ApiMap object directly.

SOF3:
> I'm also not sure what the point of "default" APIs is. Allowing there to be a difference just allows plugins to fight with each other, which seems to defeat the objective.

The default API is the implementation provided by the interface declarer, such as by PocketMine itself, explicitly intended for overriding. I don't think there is a problem since there is only one interface declarer.

SOF3:
Suggested addition:

```
public function onFinalize(string $class, Closure $closure) {
    $this->onStartupHook[] = function() {
        $instance = $this->getApi($class);
        if($instance) $closure($instance);
    }
}
```

This allows decoupling plugin dependencies by only running functions like "register xxx handler" on certain APIs *after* they are all available. In this case, `$class` does not have to be loaded at the time of calling `onFinalize`, and the closure is only executed if the API is available.

Finalization is possible if #2825 is accepted.

tests/phpunit/plugin/ApiMapTest.php
SOF3: I don't think this is really a significant issue. We could always change it to another class if we happen to have issues in the future, since this is not part of the API.

src/plugin/ApiMapEntry.php
SOF3: The `ApiMapEntry` class is supposed to hold an object of unknown type. The type information is stored in the ApiMap array key. I don't think it is correct to make this class generic.

SOF3:
Side note (previously discussed on discord): I would also like to use `ApiMap` to replace MetadataStore, so that plugins do not have to handle sessions themselves. E.g. an economy plugin:

```php
public function onLogin(PlayerLoginEvent $event) {
    $event->getPlayer()->getApiMap()->provideApi(EconomyData::class, EconomyData::loadForPlayer($event->getPlayer());
}
```

It is also worth considering whether we should rename `ApiMap` to `TypeMap` or `MetadataMap` for more generic applications in the future; it's unreasonable to call plugin player sessions an "API" when it's supposed to be internal.

dktapps:
My thoughts on this:

- For provision of plugin APIs to the core, simple getters and setters in the Server are sufficient. I don't see the need for more complication in that regard. It'll only make call chains longer, and plugin devs don't like that.
- For provision of plugin APIs to other plugins, I don't really see the point. Plugins that want to depend on EconomyAPI are not gonna prefer this new way of doing things because there's no obvious advantage. If anything, it's actually worse on account of needing a Server context to retrieve it.

The only way I see this change having significant value is if plugins wanted to fulfil API contracts set out by (multiple) other plugins.
Essentially the workflow being:
- several plugins want some economy provider that implements a shared `EconomyInterface`
- one plugin wants to provide an implementation for this interface without other plugins needing to know about its existence directly & provides an implementation of `EconomyInterface` to the API map

However, in this case, multiple plugins would end up depending on a plugin or package that provides `EconomyInterface` anyway (similar to a Composer virtual package), and at that point, the virtual plugin might as well just provide a `GenericEconomySingleton` anyway.

Long story short: While I think the basic idea is OK, I think there are other ways to implement it that plugin devs will naturally gravitate towards anyway. Unfortunately, it's pretty hard to tempt plugin devs away from `Singleton::getInstance()` when the alternative is `$this->getServer()->getApi(SingletonImpl::class)` (and probably longer still in other packages).

Since this is a basic DI container, if we introduced support for autowiring dependencies into `Plugin::__construct()`, this might break plugin devs of the habit of depending on singletons. However, if that day ever comes, I think it'll make more sense to just use a DI library rather than inventing our own in-house.

PR #3781, opened 2020-08-19
Implemented Swimming

buchwasa:
## Introduction
This pull request aims to add the missing swimming functionality that was added back in 1.4.

### API changes
Added
Living->swimming (protected)
Living->setSwimming(bool $value) : void;
PlayerToggleSwimEvent
Player->toggleSwim(bool $swim) : bool
### Behavioural changes
Hunger now takes account for swimming, 0.015, as stated in the comment.

## Backwards compatibility
This is backwards compatible with current API.

## Follow-up
I'm not too certain how well I like the setSwimming() function.

## Tests
I tested by going underwater and sprinting and going to the surface for air. I've also included a video to show it in action.
[Streamable Video](https://streamable.com/a97q6n)

### DIFF
--- a/src/entity/Living.php
+++ b/src/entity/Living.php
@@ -25,6 +25,7 @@ namespace pocketmine\entity;
 
 use pocketmine\block\Block;
 use pocketmine\block\BlockLegacyIds;
+use pocketmine\block\Water;
 use pocketmine\data\bedrock\EffectIdMap;
 use pocketmine\entity\animation\DeathAnimation;
 use pocketmine\entity\animation\HurtAnimation;
@@ -607,6 +608,14 @@ abstract class Living extends Entity{
 
 			if($this->doAirSupplyTick($tickDiff)){
 				$hasUpdate = true;
+			}
+
+			if(
+				$this->isSwimming() && !$this->isUnderwater() &&
+				$this->getWorld()->getBlock($this->getPosition()->add(0, 1, 0))->getId() === BlockLegacyIds::AIR &&
+				!$this->getWorld()->getBlock($this->getPosition()->subtract(0, 1, 0)) instanceof Water
+			){
+				$this->setSwimming(false);
 			}
 		}
 
--- a/src/entity/Living.php
+++ b/src/entity/Living.php
@@ -119,6 +119,13 @@ abstract class Living extends Entity{
 	/** @var bool */
 	protected $swimming = false;
 
+	/** @var float */
+	protected $originalWidth;
+	/** @var float */
+	protected $originalHeight;
+	/** @var float */
+	protected $originalEyeHeight;
+
 	abstract public function getName() : string;
 
 	protected function initEntity(CompoundTag $nbt) : void{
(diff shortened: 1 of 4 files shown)


src/entity/Living.php
```diff
@@ -221,6 +223,22 @@ public function setSprinting(bool $value = true) : void{
 		}
 	}
 
+	public function setSwimming(bool $value = true) : void{
+		if($value !== $this->isSwimming()) {
+			$this->swimming = $value;
+			if($this->swimming){
+				$this->height = $this->eyeHeight = $this->width;
```
dktapps, line 230: This will mess up the bounding box irreparably.

src/entity/Living.php
```diff
@@ -221,6 +223,22 @@ public function setSprinting(bool $value = true) : void{
...
+				$this->recalculateBoundingBox();
+			}else{
+				$this->setScale($this->getScale());
```
dktapps, line 233: Not sure what this is supposed to do.

src/event/player/PlayerToggleSwimEvent.php
```diff
@@ -0,0 +1,44 @@
...
+
+	/** @var bool */
+	protected $isSwimming;
+
+	public function __construct(Player $player, bool $isSwimming){
+		$this->player = $player;
+		$this->isSwimming = $isSwimming;
+	}
+
+	public function isFlying() : bool{
```
GiantQuartz, line 41: Is this name intentional? It doesn't seem to fit with the context.

dresnite (changes requested):

src/event/player/PlayerToggleSwimEvent.php
buchwasa: oops thanks for catching my laziness

src/entity/Living.php
```diff
@@ -568,6 +600,14 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
 			if($this->doAirSupplyTick($tickDiff)){
 				$hasUpdate = true;
 			}
+
+			if(
+				$this->isSwimming() && !$this->isUnderwater() &&
```
thebigsmileXD, line 605: isUnderwater() should finally be changed to check if the bounding box is submerged.
I suggest a change in syntax as follows:
`function isUnderwater(bool $checkSubmerged = false)`
This way old behaviour will be kept by default, but also allowing checks like the ones below to be removed

src/entity/Living.php
```diff
@@ -237,15 +237,14 @@ public function setSwimming(bool $value = true) : void{
...
 			$this->swimming = $value;
 			if($this->swimming){
-				$this->originalWidth = $this->width;
-				$this->originalHeight = $this->height;
-				$this->originalEyeHeight = $this->eyeHeight;
-				$this->height = $this->eyeHeight = $this->width;
+				$width = $this->size->getWidth();
+				$this->originalWidth = $width;
+				$this->originalHeight = $this->size->getHeight();
+				$this->originalEyeHeight = $this->size->getEyeHeight();
```
dktapps, line 243: why not just make an originalSize field? ...

buchwasa: great idea dktapps

dktapps:
Actually, I think originalSize isn't needed; getInitialSize()->scale() can be used, like this: https://github.com/pmmp/PocketMine-MP/blob/master/src/entity/Entity.php#L297

You might also want to adjust setScale() to account for the changes in this PR.

PR #3798, opened 2020-08-26
Fix op knockback

GamakCZ:
## Introduction

Knockback enchantment is so overpowered in pocketmine

### Relevant issues
N/A

### API changes
N/A

### Behavioural changes
Fixes knockback (check the videos to see exact changes)

## Backwards compatibility
N/A

## Follow-up
N/A

Requires translations:
N/A

## Tests
- Vanilla knockback: https://streamable.com/c1h4kb
- Knockback before PR: https://streamable.com/tq0hlb
- Knockback after PR: https://streamable.com/9m4ytp


dktapps:
this is the wrong fix. the problem is that the y-axis velocity is not capped.

GamakCZ:
It is not possible with current api

Issue #3676, opened 2020-07-03
Player cannot join to server after he play


b1zey:
### Issue description

- Actual result: 
Player playing on the server. And in one moment he kick from server and it cannot join. To solve the problem i remove player_nick_name.dat file on the players directory of the server and after this player can join.
I check packets and when player cannot join the client send packet - PacketWiolationWarning. I think it can help.

### Steps to reproduce the issue
This issue don't have specific steps to reproduce. I think, this can be error item, but a lot of players have this issue.

### OS and versions

* PocketMine-MP:  PocketMine-MP 3.14.1+dev for Minecraft: Bedrock Edition v1.16.0
* PHP: 7.3
* Server OS: Debian 9
* Game version: PE and Win 10, Minecraft Bedrock Edition 1.16.1 and 1.16.0. It happens everywhere.

### Plugins
This issue also happens without any plugins.

### Crashdump, backtrace or other files
No.

dktapps:
I need steps to reproduce in order to investigate, otherwise this issue is a waste of time.
You say a PacketViolationWarningPacket is sent. It might be useful to know the content.

radovinetch:
> I need steps to reproduce in order to investigate, otherwise this issue is a waste of time.
> You say a PacketViolationWarningPacket is sent. It might be useful to know the content.

PacketViolationWarningPacket have a severity with value 2 (terminating connection). The message or others variables are empty. Few minutes, i check all variables and drop the content.

b1zey:
> I need steps to reproduce in order to investigate, otherwise this issue is a waste of time.
> You say a PacketViolationWarningPacket is sent. It might be useful to know the content.

My friend also have this bug on empty server. This bug dont have a steps to reproduce, but this bug very important. A lot of players have this issue.

string(28) "PacketViolationWarningPacket"
string(48) "Type: 0 | Severity: 2 | PacketId: 31 | Message: "

This is a dump of PacketViolationWarningPacket. My friend and all players, who have this bug have the same.
Unfortunately, Minecraft Bedrock Edition have a lot of bugs in this release and even this packet not send everytime, i attempt to join 10 times and 2 times i have this packet. This is very bad, but what to do....

b1zey:
This is full dump of PacketViolationWarningPacket.
`object(pocketmine\network\mcpe\protocol\PacketViolationWarningPacket)#15605 (10) {
[...]

PR #3806, opened 2020-08-30
Provide a default for health

buchwasa:
## Introduction
When a new player would join without this, after commit https://github.com/pmmp/PocketMine-MP/commit/09eb904f6b818756e59325bf71ea72e2e3a12a2b, they would be kicked because there was no default provided for Health.

## Tests
There isn't much to say about this other than I tried to join after resetting my player data, and it worked fine.


HerryYT:
Maybe solution for #3676?

dktapps:
No, this bug was introduced after 3.15.0.

PR #3821, opened 2020-09-09
LevelDB: use 0 in val instead error

Eren5960:


dktapps:
This produces unexpected behaviour instead of erroring.

PR #3847, opened 2020-10-01
Introduce PlayerDisplayNameChangeEvent

Ifera:
## Introduction

This PR aims at providing a way to detect when the display name of a player changes. The display name of the player might be used in variety of places like for setting nicks, in chat, for displaying on scoreboard and the likes. Currently you can't tell if the name was updated or not. 

PM provides the function to change display name so its only fitting for a way to detect name changes as well. Rather than using custom player class to detect name change.
### Relevant issues

N/A
### API changes

- Introduced new `PlayerDisplayNameChangeEvent`
- `PlayerDisplayNameChangeEvent->getOldName()`
- `PlayerDisplayNameChangeEvent->getNewName()`
- `PlayerDisplayNameChangeEvent->setNewName()`
- Changes done to `Player::setDisplayName()` to call the event

### Behavioural changes

N/A

## Backwards compatibility

N/A

## Follow-up

N/A
## Tests

https://gist.github.com/Ifera/257e1a82f8dc2020e3fe970da92893fb
That gist also has the test result.

### DIFF
--- a/src/event/player/PlayerDisplayNameChangeEvent.php
+++ b/src/event/player/PlayerDisplayNameChangeEvent.php
@@ -45,8 +45,4 @@ class PlayerDisplayNameChangeEvent extends PlayerEvent{
 	public function getNewName() : string{
 		return $this->newName;
 	}
-
-	public function setNewName(string $newName) : void{
-		$this->newName = $newName;
-	}
 }
--- a/src/event/player/PlayerDisplayNameChangeEvent.php
+++ b/src/event/player/PlayerDisplayNameChangeEvent.php
@@ -1,0 +1,52 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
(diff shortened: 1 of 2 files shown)


src/event/player/PlayerDisplayNameChangeEvent.php
```diff
@@ -0,0 +1,52 @@
...
+
+	public function getOldName() : string{
+		return $this->oldName;
+	}
+
+	public function getNewName() : string{
+		return $this->newName;
+	}
+
+	public function setNewName(string $newName) : void{
```
dktapps, line 49: I'd prefer this wasn't a thing. It's not clear what the implications of this would be, same as cancellation.
I can approve this much more quickly if the event can't modify the outcome.

dktapps:
Test failures are unrelated.

src/event/player/PlayerDisplayNameChangeEvent.php
Ifera: Yeah true..

dktapps: To be clear, I don't like this because it permits behavioural changes that could break existing plugins, which makes it a much more review-intensive change because we need to figure out whether it makes sense to allow this to be altered (you don't expect your player's name to become Fred when you set it to Steve).

Events that only permit listening to stuff are guaranteed to not have any effect on the code flow and therefore count as pure addition of features. Cancellable and mutable events are much more troublesome.

PR #3848, opened 2020-10-02
expose server instance for Entity class

ghost:

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -560,13 +560,6 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 		}
 	}
 
-	/**
-	 * @return Server
-	 */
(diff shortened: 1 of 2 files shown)


lukeeey:
why?

DimBis:
> why?

why not?

dktapps:
The burden of supporting an argument is on the person making the argument.

dktapps:
I don't see any reasons to do this and I can think of a couple of reasons not to (forces Entity to have Server dependency, potentially breaks BC with custom entity classes), and so forth.

SOF3:
Player provides getServer because it is required by CommandSender. I don't think it's appropriate to require getServer in CommandSender anyway, but removing it from CommandSender is another business. Meanwhile, adding it to more classes doesn't make sense.

PR #3849, opened 2020-10-02
Clear command

### LINKED ISSUE #3840: /clear command
## Description

/clear 

## Justification

i know many servers including my own have this most likely written in our private cores, however i think it should be add here since it is a vanilla mc command

DimBis:
### Relevant issues
* Fixes #3840

## Follow-up

Requires translations:

| Name | Value in eng.ini |
| :--: | :---: |
| `commands.clear.successful` | `Cleared the inventory of {%0}, removing {%1} items` |
| `pocketmine.command.clear.description` | `Clears items from player inventory` |
| `pocketmine.command.clear.usage` | `/clear [player]` |

### DIFF
--- a/src/pocketmine/command/defaults/ClearCommand.php
+++ b/src/pocketmine/command/defaults/ClearCommand.php
@@ -1,0 +1,102 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * @author PocketMine Team
+ * @link http://www.pocketmine.net/
+ *
+ *
+*/
+
+declare(strict_types=1);
+
+namespace pocketmine\command\defaults;
+
+use pocketmine\command\Command;
+use pocketmine\command\CommandSender;
+use pocketmine\command\utils\InvalidCommandSyntaxException;
+use pocketmine\lang\TranslationContainer;
+use pocketmine\Player;
+use pocketmine\utils\TextFormat;
+use function count;
+
+class ClearCommand extends VanillaCommand{
+
+	public function __construct(string $name){
+		parent::__construct(
+			$name,
+			"%pocketmine.command.clear.description",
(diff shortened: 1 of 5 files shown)


src/pocketmine/command/defaults/ClearCommand.php
```diff
@@ -0,0 +1,102 @@
...
+				$sender->sendMessage($sender->getServer()->getLanguage()->translateString(TextFormat::RED . "%commands.generic.permission"));
+
+				return true;
+			}
+
+			$player = $sender->getServer()->getPlayer($args[0]);
+
+			if($player instanceof Player){
+
+				$cleared = 0;
```
lukeeey, line 66:
```php
$cleared = count($player->getInventory()->getContents();
```
same below

lukeeey (changes requested):

lukeeey:
Can you please at least follow the template or remove the parts that don't apply (there are no API changes here so you can get away with not having that section)

also
image

src/pocketmine/command/defaults/ClearCommand.php
dktapps: It's not the same. This loop adds together all of the itemstacks' counts in the inventory (e.g. 1 stack of 64 diamonds cleared will report 64 items).

lukeeey: oh sorry, my bad

src/pocketmine/command/defaults/ClearCommand.php
```diff
@@ -0,0 +1,102 @@
...
+
+				foreach ($player->getInventory()->getContents() as $content) {
+					$cleared += $content->getCount();
+				}
+
+				$player->getInventory()->clearAll();
+				Command::broadcastCommandMessage($sender, new TranslationContainer("commands.clear.successful", [$player->getName(), $cleared]));
```
dktapps, line 73: The code inside this block is repeated.

dktapps:
I can't accept this on stable due to feature freeze. You'll need to target the master branch instead.

dktapps:
This appears to be a near-perfect copy-paste of `KillCommand` with adjustments, explaining why the target selection makes no sense.

Ifera:
Also.. It's only clearing players inventory. Shouldn't this also clear it's armor and cursor inventory as well?

DimBis:
> Also.. It's only clearing players inventory. Shouldn't this also clear it's armor and cursor inventory as well?

I have searched on google for the behavior of the command and i found an article says it ONLY clears the inventory contents and not the armor inventory.

DimBis:
> This appears to be a near-perfect copy-paste of `KillCommand` with adjustments, explaining why the target selection makes no sense. @shoghicp is cursing us from beyond the grave.
Well nearly all of the built in commands are the same exact shit so i didn't bother myself to unshit it.

dktapps:
> Well nearly all of the built in commands are the same exact shit so i didn't bother myself to unshit it.

Bad code isn't an excuse to write more bad code. I won't merge something that I have to rewrite.

DimBis:
> > Well nearly all of the built in commands are the same exact shit so i didn't bother myself to unshit it.
> 
> Bad code isn't an excuse to write more bad code. I won't merge something that I have to rewrite.

k i don't remember asking tho

DimBis:
Also if you don't want a bad code to rewrite why did you decide to maintain pocketmine in the first place?!?!?!

PR #3854, opened 2020-10-03
Added support for missing clear command

### LINKED ISSUE #3840: /clear command

Ifera:
## Introduction

This PR aims to provide support for vanilla command `clear`. With another [PR](https://github.com/pmmp/Language/pull/58) to `pmmp/Language` repo adding support for the missing strings. Exact vanilla-like functionality was achieved.

### Relevant issues

* Fixes #3840 

## Changes
- Added `/clear` command
- Registered the command to command map
- Registered permissions for the command
### API changes

`N/A`
### Behavioural changes

`N/A`
## Backwards compatibility

`Yes`
## Follow-up

`N/A` 
_(translations already added as explained above)_
## Tests

Not quite sure what to add in tests, anyways here are some screenshots in [this gist](https://gist.github.com/Ifera/c6c5d39b99321a148c42fde62c80d737).


src/command/defaults/ClearCommand.php
```diff
@@ -0,0 +1,178 @@
...
+
+			//vanilla behaviour, removes item from armor inventory, regardless of the count specified
+			if(($index = $target->getArmorInventory()->first($item)) !== -1){
+				$cleared++;
+				$target->getArmorInventory()->clear($index);
+			}
+		}
+
+		if($cleared > 0){
+			$sender->sendMessage(new TranslationContainer("%commands.clear.success", [$target->getName(), $cleared]));
```
dktapps, line 171: This should use `Command::broadcastCommandMessage()` for auditing.

dktapps, line 171: This should use `Command::broadcastCommandMessage()` for auditing.

src/command/defaults/ClearCommand.php
```diff
@@ -0,0 +1,178 @@
...
+					}
+
+					$cleared += $i->getCount();
+					$maxCount -= $i->getCount();
+					$target->getInventory()->clear($index);
+				}
+			}
+
+			//vanilla behaviour, removes item from armor inventory, regardless of the count specified
+			if(($index = $target->getArmorInventory()->first($item)) !== -1){
```
dktapps, line 164: I don't like this. This behaviour looks buggy and it doesn't make sense to replicate vanilla bugs.

src/command/defaults/ClearCommand.php
```diff
@@ -0,0 +1,178 @@
...
+				$maxCount = -1;
+			}
+
+			if($item !== null){
+				$item->setCount($maxCount = $this->getInteger($sender, $args[2], 0));
+			}
+		}
+
+		//checking players inventory for all the items matching the criteria
+		if($item !== null and $item->getMeta() === 0 and $maxCount === 0){
```
dktapps, line 108: why is there a `getMeta()` check here?

src/command/defaults/ClearCommand.php
```diff
@@ -0,0 +1,178 @@
...
+		$target = null;
+		if($sender instanceof Player){
+			if(!$sender->hasPermission("pocketmine.command.clear.self")){
+				$sender->sendMessage(new TranslationContainer(TextFormat::RED . "%commands.generic.permission"));
+				return true;
+			}
+
+			$target = $sender;
+		}else{
+			if(count($args) < 1){
```
dktapps, line 66: This can be reduced to `elseif`.

src/command/defaults/ClearCommand.php
```diff
@@ -0,0 +1,178 @@
...
+			}
+
+			if($count > 0){
+				$sender->sendMessage(new TranslationContainer("%commands.clear.testing", [$target->getName(), $count]));
+			}else{
+				$sender->sendMessage(new TranslationContainer(TextFormat::RED . "%commands.clear.failure.no.items", [$target->getName()]));
+			}
+
+			return true;
+		}
```
dktapps, line 122: I don't understand the purpose of this entire block. The point of `/clear` is to remove items.

src/command/defaults/ClearCommand.php
Ifera: Vanilla `clear` command does more than just simply clearing the inventory of the player. Vanilla syntax is, 
```clear [player] [item name] [data] [max count]```
Where item being the item name and data being the damage/meta. If the meta and the max count are set to zero then it reports back all the items matching the items specified in the players inventory. 
More info [here](https://www.digminecraft.com/game_commands/clear_command.php) or see the attached screenshot.

image

File: src/command/defaults/ClearCommand.php (hunk 2 above)
Comment (Ifera) [reply src/command/defaults/ClearCommand.php:164]: So what do you suggest as a replacement..? Just check if the item is

File: src/command/defaults/ClearCommand.php (hunk 5 above)
Comment (Ifera) [reply src/command/defaults/ClearCommand.php:122]: Discussed in above comment.

File: src/command/defaults/ClearCommand.php (hunk 3 above)
Comment (dktapps) [reply src/command/defaults/ClearCommand.php:108]: This seems out of character for a `/clear` command...

File: src/command/defaults/ClearCommand.php (hunk 2 above)
Comment (dktapps) [reply src/command/defaults/ClearCommand.php:164]: I think a more sensible alternative would be to check all the available inventories for the item in a specific order. Maybe the armour inventory could be prioritized, but it doesn't make sense to remove items from multiple inventories when only a count of one was desired.

File: src/command/defaults/ClearCommand.php (hunk 3 above)
Comment (Ifera) [reply src/command/defaults/ClearCommand.php:109]: True but that's how its done in vanilla... Just to be sure, I checked this in-game as well.

File: src/command/defaults/ClearCommand.php (hunk 2 above)
Comment (Ifera) [reply src/command/defaults/ClearCommand.php:165]: Alright, seems fair :+1:

File: src/command/defaults/ClearCommand.php (hunk 3 above)
Comment (dktapps) [reply src/command/defaults/ClearCommand.php:109]: OK, so /clear has some behaviour that's not relevant to clearing when using count zero. Fine. But this still doesn't explain the metadata zero check.

Comment (dktapps) [reply src/command/defaults/ClearCommand.php:109]: EX: `/clear steve bucket 8 0` has the same behaviour as `/clear steve bucket 0 0`.

Comment (Ifera) [reply src/command/defaults/ClearCommand.php:109]: Ah true, forgot about that.

File: src/command/defaults/ClearCommand.php (hunk 2 above)
Comment (Ifera) [reply src/command/defaults/ClearCommand.php]: Fixed, please review.

File: src/command/defaults/ClearCommand.php [hunk 6]
```diff
@@ -0,0 +1,182 @@
...
+			}
+		}
+
+		$maxCount = -1;
+		if(isset($args[2])){
+			if(!is_numeric($args[2])){
+				$maxCount = -1;
+			}
+
+			if($item !== null){
```
dktapps, line 101: This will always be true (presence of args[2] guarantees the presence of args[1], which is the item type).

src/command/defaults/ClearCommand.php
```diff
@@ -0,0 +1,182 @@
...
+				$sender->sendMessage(new TranslationContainer(TextFormat::RED . "%commands.give.item.notFound", [$args[1]]));
+				return true;
+			}
+		}
+
+		$maxCount = -1;
+		if(isset($args[2])){
+			if(!is_numeric($args[2])){
+				$maxCount = -1;
+			}
```
dktapps, line 99: This code is useless (because the condition below is always true).

src/command/defaults/ClearCommand.php
```diff
@@ -0,0 +1,182 @@
...
+		if(!$this->testPermission($sender)){
+			return true;
+		}
+
+		if(count($args) > 3){
+			throw new InvalidCommandSyntaxException();
+		}
+
+		$target = null;
+		if($sender instanceof Player){
```
dktapps, line 59: It doesn't make sense to check this before checking if args[0] is set. A more intuitive logic flow would be
```
if(args[0] is set){
    //try to find player
}elseif(i am a player){
    //use current player as target
}else{
    //invalid syntax
}
```

src/command/defaults/ClearCommand.php
```diff
@@ -0,0 +1,173 @@
...
+				$sender->sendMessage(new TranslationContainer(TextFormat::RED . "%commands.generic.permission"));
+				return true;
+			}
+
+			$target = $sender;
+		}else{
+			throw new InvalidCommandSyntaxException();
+		}
+
+		if($target === null){
```
dktapps, line 76: This block could be moved inside the block which calls Server->getPlayer().

src/command/defaults/ClearCommand.php
```diff
@@ -0,0 +1,173 @@
...
+				}
+			}else{
+				//clear only the given amount of that particular item from targets inventory
+				if(($index = $target->getArmorInventory()->first($item)) !== -1){
+					$cleared++;
+					$maxCount--;
+					$target->getArmorInventory()->clear($index);
+				}
+
+				foreach($target->getInventory()->all($item) as $index => $i){
```
dktapps, line 146: An underflow will occur here if only 1 item wanted to be cleared and it was already removed from the armor inventory.

src/command/defaults/ClearCommand.php
```diff
@@ -0,0 +1,174 @@
...
+				//vanilla checks this at argument parsing layer, can't come up with a better alternative
+				$sender->sendMessage(new TranslationContainer(TextFormat::RED . "%commands.give.item.notFound", [$args[1]]));
+				return true;
+			}
+		}
+
+		$maxCount = -1;
+		if(isset($args[2])){
+			$item->setCount($maxCount = $this->getInteger($sender, $args[2], 0));
+		}
```
dktapps, line 94: You can move this block inside the previous one to fix the PHPStan error.

src/command/defaults/ClearCommand.php
```diff
@@ -0,0 +1,174 @@
...
+			$target->getCursorInventory()->clearAll();
+		}else{
+			//clear the item from targets inventory irrelevant of the count
+			if($maxCount === -1){
+				if(($index = $target->getArmorInventory()->first($item)) !== -1){
+					$cleared++;
+					$target->getArmorInventory()->clear($index);
+				}
+
+				foreach($target->getInventory()->all($item) as $index => $i){
```
dktapps, line 133: `$index` already exists outside the loop, so PHPStan doesn't like you overwriting it. You can use a different variable name to get around it.

src/command/defaults/ClearCommand.php
```diff
@@ -0,0 +1,174 @@
...
+			}else{
+				//clear only the given amount of that particular item from targets inventory
+				if(($index = $target->getArmorInventory()->first($item)) !== -1){
+					$cleared++;
+					$maxCount--;
+					$target->getArmorInventory()->clear($index);
+				}
+
+				if($maxCount > 0){
+					foreach($target->getInventory()->all($item) as $index => $i){
```
dktapps, line 146: Same as above.

src/command/defaults/ClearCommand.php
Ifera: Done

src/command/defaults/ClearCommand.php
```diff
@@ -0,0 +1,174 @@
...
+		//checking players inventory for all the items matching the criteria
+		if($item !== null and $maxCount === 0){
+			$count = 0;
+			$contents = array_merge($target->getInventory()->all($item), $target->getArmorInventory()->all($item));
+			foreach($contents as $content){
+				$count += $content->getCount();
+			}
+
+			if($count > 0){
+				Command::broadcastCommandMessage($sender, new TranslationContainer("%commands.clear.testing", [$target->getName(), $count]));
```
dktapps, line 105: This doesn't need to be a `broadcast` since it doesn't actually do anything, it's only viewing contents.

src/command/defaults/ClearCommand.php
```diff
@@ -0,0 +1,174 @@
...
+			}
+
+			return true;
+		}
+
+		$cleared = 0;
+
+		//clear everything from the targets inventory
+		if($item === null){
+			$contents = array_merge($target->getInventory()->getContents(), $target->getArmorInventory()->getContents());
```
dktapps, line 117: Count of cursor inventory is missing here.

dktapps: I'm not sure if `clear` is intended to clear the cursor or not.

Ifera: Does the count of cursor inventory matter tho?

dktapps: Yes, I would think so. Either don't clear it, or include it in the total.

Ifera: I opted, not to clear it.

dktapps:
ugh, github... all these extra commits don't show in local `git log`, it's ok...

dktapps:
Merged manually as c4b47c6c067528454b685d68af61c705ddd8ce49. Thank you.

PR #3855, opened 2020-10-04
Add all entity ids

Matze997:
## Introduction
This pull request aims to add all minecraft entity ids

### Relevant issues
N/A

## Changes
- Added all entity ids to AddActorPacket.php and EntityIds.php

### API changes
N/A

### Behavioural changes
N/A

## Backwards compatibility
Yes

## Follow-up
N/A

Requires translations:
N/A

## Tests

grafik
grafik
grafik

### DIFF
--- a/src/pocketmine/entity/EntityIds.php
+++ b/src/pocketmine/entity/EntityIds.php
@@ -74,6 +74,7 @@ interface EntityIds{
 	public const AGENT = 56, LEARN_TO_CODE_MASCOT = 56;
 	public const VINDICATOR = 57;
 	public const PHANTOM = 58;
+	public const RAVAGER = 59;
 
 	public const ARMOR_STAND = 61;
 	public const TRIPOD_CAMERA = 62;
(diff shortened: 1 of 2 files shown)


SOF3:
What is the reason of adding these IDs if they are not used in PocketMine itself?

Matze997:
Simply to make it easier for developers to spawn the new entities

buchwasa:
> What is the reason of adding these IDs if they are not used in PocketMine itself?

If we stuck with that mindset, then the EntityIds list would be pretty empty.

lukeeey:
The only reason I could see for this is for documentation purposes so that plugin developers can find the IDs.

I might document them on a sub page at https://wiki.vg/Bedrock_Protocol tomorrow.

dktapps:
New numeric IDs are useless anyway.

dktapps:
This is no longer useful since 3.13.0, since PM's AddActorPacket impl accepts string IDs.

PR #3861, opened 2020-10-08
Force xuid to be empty if `xbox-auth = off` in server.properties

### LINKED ISSUE #3089: Xbox players who failed to login to XBLA have illegal XUIDs
### Issue description

``[13:32:28.267] [Server thread/INFO]: Starting Minecraft: Bedrock Edition server version v1.12.0
[13:32:28.296] [Server thread/WARNING]: Online mode is disabled. The server will not verify that players are authenticated to Xbox Live.
[13:32:28.296] [Server thread/WARNING]: While this makes it possible to connect without internet access, it also allows hackers to connect with any username they choose.
[13:32:28.297] [Server thread/WARNING]: To enable authentication, set "xbox-auth" to "true" in server.properties.
[13:32:28.642] [Server thread/INFO]: This server is running PocketMine-MP version 4.0.0+dev

[13:33:16.405] [Server thread/INFO]: [NetworkSession: 000059555] Session opened
[13:33:16.536] [Server thread/INFO]: [NetworkSession: 000059555] Player: Eren5960
[13:33:17.308] [Server thread/INFO]: [NetworkSession: Eren5960] Session closed due to Invalid session. Reason: Unexpected XUID for non-XBOX-authenticated player``

1. turn off xbox-auth in server.properties
2. sign out your xbox and join server

### OS and versions

* PocketMine-MP: 056c5ed6cd0e59160c85f9ca6788f70210bac007
* PHP: 7.3.7
* Server OS: Windows/Linux
* Game version: PE/Win10

### Plugins
No plugins

no crash

95CivicSi:
## Introduction

Some proxy software leaves the XUID in the login packet for reference but PM4 will reject the login because the keychain is not signed by Mojang.  Additionally, there may be some use cases where an invalid xuid is presented by the client when the client isn't authenticated.

This update causes the server to ignore existing xuid information in the login packet, if `xbox-auth` is set to off in server.properties.

### Relevant issues

* Fixes #3089 

### API changes

None

### Behavioural changes

PocketMine-MP will now ignore any existing xuid in the login packet if authentication is not required.

## Backwards compatibility

None
## Follow-up

None

### DIFF
--- a/src/network/mcpe/handler/LoginPacketHandler.php
+++ b/src/network/mcpe/handler/LoginPacketHandler.php
@@ -121,7 +121,7 @@ class LoginPacketHandler extends PacketHandler{
 			$uuid,
 			$skin,
 			$clientData->LanguageCode,
-			$extraData->XUID,
+			$this->server->requiresAuthentication() ? $extraData->XUID : "",
 			(array) $clientData
 		);
 		($this->playerInfoConsumer)($playerInfo);


dktapps:
This will have side effects. Plugins may assume that players who are authenticated will have XUIDs (they are still checked even when xbox-auth is disabled, they just don't get kicked). This unconditionally removes **all** XUIDs.

95CivicSi:
Would it be acceptable then to mark all logins as not authenticated when xbox-auth = off?

dktapps:
The referenced commit restores the PM3 behaviour. Thanks for providing the push to do this 🙂

PR #3867, opened 2020-10-08
Entity: properly account for upwards motion when calculating fall dis…

### LINKED ISSUE #3866: Players take a large amount of fall damage when comboed
### Issue description

- Expected result: Players to not take a lot of fall damage
- Actual result: Players take a lot of fall damage.

1. Join with a friend
2. Combo them for a while (10-15s is a good amount)
3. Stop, and let them take fall damage

### OS and versions

* PocketMine-MP: 3.x & 4.0
* PHP: 7.4
* Server OS: Linux
* Game version: PE & Win10

### Plugins
Can be done without plugins

dktapps:
…tance

Previously, upwards movement wouldn't be considered, but downwards would, so if an entity bobbed up and down in the air for a while (e.g. while being comboed in PvP), the downwards distance would accumulate and deal a large amount of fall damage on touchdown.
This commit changes fall distance measurement to correctly account for upwards movement.

A better way of measuring fall distance would simply be to record the highest Y coordinate reached while in the air, and then measure the distance between that and the point of contact when landing. This would also remove the need to constantly update the fallDistance field. However, this would involve a BC break and will therefore have to wait until PM4.

This should fix #3866, although I have not yet tested that specific scenario.

### DIFF
--- a/src/pocketmine/entity/Entity.php
+++ b/src/pocketmine/entity/Entity.php
@@ -1451,8 +1451,14 @@ abstract class Entity extends Location implements Metadatable, EntityIds{
 				$this->fall($this->fallDistance);
 				$this->resetFallDistance();
 			}
-		}elseif($distanceThisTick < 0){
+		}elseif($distanceThisTick < $this->fallDistance){
+			//we've fallen some distance (distanceThisTick is negative)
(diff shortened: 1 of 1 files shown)


buchwasa:
Just tested a game with it, everything seems to be working alright.

PR #3870, opened 2020-10-11
Isolate and always show IP details on install

jasonw4331:
## Introduction

Skipping the installation wizard does not display any IP information in the console. I think it would be more usable if it does.

### Relevant issues

N/A

### API changes

1 private function created within the Wizard class.

### Behavioral changes

On first startup, the IP information for the server will be displayed in the console.

## Backwards compatibility

N/A

## Follow-up

Could possibly be replicated on API 4.

## Tests

image

### DIFF
--- a/src/pocketmine/wizard/SetupWizard.php
+++ b/src/pocketmine/wizard/SetupWizard.php
@@ -92,6 +92,7 @@ class SetupWizard{
 		$config->save();
 
 		if(strtolower($this->getInput($this->lang->get("skip_installer"), "n", "y/N")) === "y"){
+			$this->IPDetails();
 			return true;
 		}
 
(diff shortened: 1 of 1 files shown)


src/pocketmine/wizard/SetupWizard.php
```diff
@@ -218,7 +220,9 @@ private function networkFunctions() : void{
 		}
 
 		$config->save();
-
+	}
+	
+	private function IPDetails() : void{
```
SOF3, line 225: Please use camelCase. IP doesn't need to be always uppercase.

SOF3 (changes requested):
Polling the IP is sometimes very slow, depending on the endpoint availability. I am not sure it's a good idea to do this without making it async. In particular, some configurations such as Docker might require frequent startups without initial data, so this change will slow down startup greatly.

jasonwynn10:
The `no-wizard` command line argument can still be used as always to skip over the wizard output completely.

PR #3871, opened 2020-10-12
Introduce new command overload system

alvin0319:
## Introduction

This pull request will rewrite some command execution processes and will provide developers with a more accurate and diverse way to execute the commands.

### Relevant issues

There are no relevant issues.

### API changes

- The following classes have been added:
  - `Overload`
  - `Parameter`
  - `EnumParameter`
  - `PlayerParameter`
  - `StringParameter`
  - `TextParameter`
  - `Vector3Parameter`
  - `NumericParameter`
  - `FloatParameter`
  - `IntegerParameter`
- The following methods have been added:
  - `Command->addOverload()`
  - `Command->getOverloads()`
  - `Command->parseArguments()`
- The following methods have been changed:
  - `Command->execute()` is no longer abstract

### Behavioural changes

There are no behavior changes.

## Backwards compatibility

There are no backwards compatibility issues.

Developers can still override `Command->execute()` and there is no problem with using it.

## Tests

- PHPStan analyze test

```
Note: Using configuration file C:\Users\user\Desktop\MCBE\PocketMine-MP-command-overload\phpstan.neon.dist.
 1282/1282 [============================] 100%

 ------ ---------------------------------------------------------------
  Line   src\command\Overload.php                                      
 ------ ---------------------------------------------------------------
  41     Anonymous function should have native return typehint "void".
  109    Anonymous function should have native return typehint "void".
 ------ ---------------------------------------------------------------

                                                                                                                        
 [ERROR] Found 2 errors
```

- Example plugin

```php
<?php

/**
 * @name TestCommandPlugin
 * @author alvin0319
 * @main alvin0319\TestCommandPlugin\TestCommandPlugin
 * @version 1.0.0
 * @api 4.0.0
 */

declare(strict_types=1);

namespace alvin0319\TestCommandPlugin;

use pocketmine\command\Overload;
use pocketmine\command\parameter\defaults\PlayerParameter;
use pocketmine\command\parameter\defaults\Vector3Parameter;

class TestCommandPlugin extends PluginBase{

		$this->getServer()->getCommandMap()->register("test", new class extends Command{

			public function __construct(){
				parent::__construct("test", "Test command");
				$this->addOverload((new Overload())
					->addParameter((new Vector3Parameter("vector")))
					->setCommandHandler(function(CommandSender $sender, array $args){
						$vector = $args["vector"];

						$sender->sendMessage("Vector: $vector");
					})
				);
				$this->addOverload((new Overload(function(CommandSender $sender, array $args){
					$target = $args["target"]->getName();
					$sender->sendMessage("Target: {$target}");
				}))
					->addParameter(new PlayerParameter("target"))
				);
			}
		});
	}
}
```

### DIFF
--- a/src/command/Command.php
+++ b/src/command/Command.php
@@ -119,7 +119,7 @@ abstract class Command{
 		$offset = 0;
 		$parameters = $overload->getParameters();
 		$argCount = count($parameters);
-		usort($parameters, function (Parameter $a, Parameter $b): int {
+		usort($parameters, function(Parameter $a, Parameter $b) : int{
 			if($a->getLength() === PHP_INT_MAX){
 				return 1;
 			}
--- a/src/command/Command.php
+++ b/src/command/Command.php
@@ -101,7 +101,6 @@ abstract class Command{
 			return false;
 		}
 		foreach($this->overloads as $overload){
-			var_dump($overload->canParse($sender, $args));
 			if($overload->canParse($sender, $args)){
 				$handler = $overload->getCommandHandler();
 				if($handler !== null){
--- a/src/command/Command.php
+++ b/src/command/Command.php
@@ -101,6 +101,7 @@ abstract class Command{
 			return false;
 		}
 		foreach($this->overloads as $overload){
+			var_dump($overload->canParse($sender, $args));
 			if($overload->canParse($sender, $args)){
 				$handler = $overload->getCommandHandler();
 				if($handler !== null){
@@ -139,9 +140,7 @@ abstract class Command{
 			$argument = implode(" ", array_slice($args, $offset, $parameter->getLength()));
 			if($parameter->canParse($sender, $argument)){
 				$result[$parameter->getName()] = $parameter->parse($sender, $argument);
-				if(!$parameter->isOptional){
-					$offset += $parameter->getLength();
-				}
+				$offset += $parameter->getLength();
 			}
 		}
 		return $result;
--- a/src/command/Command.php
+++ b/src/command/Command.php
@@ -175,7 +175,7 @@ abstract class Command{
 	 */
 	public function getOverloads() : array{
 		if(count($this->overloads) === 0){
-			$this->overloads[] = new Overload($this);
+			$this->overloads[] = new Overload();
 		}
 		return $this->overloads;
 	}
--- a/src/command/Command.php
+++ b/src/command/Command.php
@@ -102,7 +102,10 @@ abstract class Command{
 		}
 		foreach($this->overloads as $overload){
 			if($overload->canParse($sender, $args)){
-				$this->onRun($sender, $this->parseArguments($sender, $overload, $args));
+				$handler = $overload->getCommandHandler();
+				if($handler !== null){
+					($handler)($sender, $this->parseArguments($sender, $overload, $args));
+				}
 				return true;
 			}
 		}
@@ -142,13 +145,6 @@ abstract class Command{
 			}
 		}
 		return $result;
-	}
-
-	/**
-	 * @param mixed[] $args An array containing objects or values
-	 */
-	public function onRun(CommandSender $sender, array $args){
-
 	}
 
 	public function getName() : string{
--- a/src/command/Command.php
+++ b/src/command/Command.php
@@ -33,8 +33,12 @@ use pocketmine\Server;
 use pocketmine\timings\Timings;
 use pocketmine\timings\TimingsHandler;
 use pocketmine\utils\TextFormat;
+use function array_slice;
+use function count;
 use function explode;
+use function implode;
 use function str_replace;
+use const PHP_INT_MAX;
 
 abstract class Command{
 
@@ -68,6 +72,9 @@ abstract class Command{
 	/** @var string|null */
 	private $permissionMessage = null;
 
+	/** @var Overload[] */
+	protected $overloads = [];
+
 	/** @var TimingsHandler|null */
 	public $timings = null;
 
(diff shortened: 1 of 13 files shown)


dktapps:
It would be preferable to have callbacks bound to each overload instead of a central `onRun()`. That's how I'd personally planned to implement this.

src/command/Overload.php
```diff
@@ -0,0 +1,107 @@
...
+
+class Overload{
+	/** @var Command */
+	protected $command;
+
+	/** @var Parameter[] */
+	protected $parameters = [];
+
+	public function __construct(Command $command){
+		$this->command = $command;
```
dktapps, line 41: This circular dependency seems completely unnecessary.

src/command/parameter/Parameter.php
```diff
@@ -0,0 +1,121 @@
...
+
+	/**
+	 * Returns the name of parameter
+	 */
+	abstract public function getTargetName() : string;
+
+	/**
+	 * Returns the translation of fail message
+	 */
+	public function getFailMessage(CommandSender $sender) : string{
```
dktapps, line 79: Better to pass a `Language` instance in here directly. There's no need to depend on `CommandSender` (nor `Server`).

src/command/parameter/Parameter.php
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	public function getPostfix() : ?string{
+		return $this->postfix;
+	}
+
+	public function prepare() : void{
+	}
+
+	public function setLength(int $length) : self{
```
dktapps, line 104: what exactly is the purpose of this?

src/command/parameter/Parameter.php
```diff
@@ -0,0 +1,121 @@
...
+	public function getEnum() : ?CommandEnum{
+		return $this->enum;
+	}
+
+	public function setPostfix(?string $postfix) : Parameter{
+		$this->postfix = $postfix;
+		return $this;
+	}
+
+	public function getPostfix() : ?string{
```
dktapps, line 97: Postfixes only apply to integer parameters. Why is this here?

src/command/parameter/Parameter.php
```diff
@@ -0,0 +1,121 @@
...
+	abstract public function getTargetName() : string;
+
+	/**
+	 * Returns the translation of fail message
+	 */
+	public function getFailMessage(CommandSender $sender) : string{
+		return $sender->getServer()->getLanguage()->translateString("%commands.generic.usage", [$this->getOverload()->getCommand()->getUsage()]);
+	}
+
+	public function setEnum(?CommandEnum $enum) : self{
```
dktapps, line 83: Again, I don't see why this is embedded directly into the base `Parameter`. Also, the protocol types should not be exposed to the API.

src/command/parameter/defaults/IntegerParameter.php
```diff
@@ -0,0 +1,42 @@
...
+*/
+
+declare(strict_types=1);
+
+namespace pocketmine\command\parameter\defaults;
+
+use pocketmine\command\CommandSender;
+use pocketmine\network\mcpe\protocol\AvailableCommandsPacket;
+
+class IntegerParameter extends FloatParameter{
```
dktapps, line 29: This is misleading. `instanceof FloatParameter` will return `true` in this case.

src/player/Player.php
```diff
@@ -832,6 +832,10 @@ public function doFirstSpawn() : void{
 		if($this->getHealth() <= 0){
 			$this->respawn();
 		}
+
+		foreach($this->server->getOnlinePlayers() as $player){
+			$player->getNetworkSession()->syncAvailableCommands();
+		}
```
dktapps, line 838: what is the purpose of this?

src/player/Player.php
```diff
@@ -1973,6 +1977,10 @@ public function disconnect(string $reason, $quitMessage = null, bool $notify = t
 		}
 		$this->save();
 
+		foreach($this->server->getOnlinePlayers() as $player){
+			$player->getNetworkSession()->syncAvailableCommands();
+		}
```
dktapps, line 1982: what is the purpose of this?

dktapps (changes requested):
I haven't reviewed this in detail yet.

lukeeey:
> It would be preferable to have callbacks bound to each overload instead of a central `onRun()`. That's how I'd personally planned to implement this.

That is preferable in most cases but that is a big change to how everyone is used to commands being handled. I think the way the PR implements it right now isn’t that bad. 

Maybe a way to optionally have callbacks if they wish would be better? I say this because when I was implementing brigadier into Nukkit (and brigadier also uses a sort of callback system but iirc it’s optional and you can just use a single method to handle everything), I ran into the fact that with the /title command you have options title, subtitle and actionbar. These things are all pretty much the same (they accept a text parameter) they just call different methods. 

With the callback style, I had three different methods with some duplicated code (sanity checks and whatnot) to do these simple tasks but I ended up making it shorter by putting them all into the same method and using a switch.

dktapps:
@lukeeey it's possible to introduce some kind of legacy mapping which just registers `Command::execute()` as the only overload of a command to preserve backwards compatibility.

src/command/parameter/Parameter.php
alvin0319: https://github.com/pmmp/PocketMine-MP/blob/master/src/network/mcpe/protocol/types/command/CommandParameter.php#L43

If you look at the explanation here, postfix can only be set to null and string.

lukeeey:
@dktapps That sounds like a good idea 👍

src/command/Overload.php
```diff
@@ -0,0 +1,107 @@
...
+	public function getCommand() : Command{
+		return $this->command;
+	}
+
+	/**
+	 * @return Parameter[]
+	 */
+	public function getParameters() : array{
+		if(count($this->parameters) === 0){
+			return [new TextParameter("args", true)]; //Prevents parameters from being sent with empty values
```
lukeeey, line 66: But some commands may not have any parameters? As far as i am aware you can send a command to the client without any parameters and it will display just fine.

lukeeey: I may be understanding this wrong, actually. I'm guessing this wont be sent if there are no overloads added to the command.

src/player/Player.php
alvin0319: PlayerParameter needs to synchronize every time the players join or leaves.

lukeeey: But since the type is `target` the client will show a list of players for you. This is the same with the `Item` enum name, which will show a list of items available on the client (although BDS does actually send the entire item list for some reason, you don't actually need to)

alvin0319:
Fixed it by review, Please check the commits.

lukeeey (commented):
An optional parameter doesn't need a default value... if the player has provided the parameter, it shouldn't return anything (because it doesnt exist)

src/command/Command.php
```diff
@@ -101,6 +101,7 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
 			return false;
 		}
 		foreach($this->overloads as $overload){
+			var_dump($overload->canParse($sender, $args));
```
SOF3: remove this?

alvin0319:
Is it okay to add `CommandSender` and `$params(array)` parameters to `Parameter->getFailMessage()` so that they can be converted from `Language`?

dktapps:
Why not just take `Language` as a parameter?

alvin0319:
Currently I used `Language` as a parameter, but something like `commands.generic.usage` do not require additional arguments?

dktapps:
I don't understand what you're talking about.

PR #3878, opened 2020-10-23
Handle InvalidArgumentException thrown by VanillaEnchantments::fromString

Muqsit:
## Introduction

Commit https://github.com/pmmp/PocketMine-MP/commit/a01c086481da988ab3409bc399d82f72ed9134ce causes `/enchant` to throw unhandled `\InvalidArgumentException` when supplied an unregistered string enchantment ID.

## Tests

Prior to this patch:
* Run `/enchant steve aqua_affinity 1` (`aqua_affinity` is not a registered enchantment)
* Server throws `Error: No such registry member: pocketmine\item\enchantment\VanillaEnchantments::AQUA_AFFINITY` (https://pastebin.com/raw/UDqLFC2i)

With this patch:
* Run `/enchant steve aqua_affinity 1`
* An error message is displayed in chat
image

### DIFF
--- a/src/command/defaults/EnchantCommand.php
+++ b/src/command/defaults/EnchantCommand.php
@@ -69,13 +69,17 @@ class EnchantCommand extends VanillaCommand{
 
 		if(is_numeric($args[1])){
 			$enchantment = VanillaEnchantments::byMcpeId((int) $args[1]);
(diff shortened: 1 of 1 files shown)


dktapps:
I think it's also time to lose the integer IDs in the command, but that can wait for another time I guess.

PR #3883, opened 2020-10-27
Fix sugarcane growing past the top of the world on random tick

Unickorn:
## Introduction
Sugarcane blocks were still growing past the top of the world, I think Dylan forgot to add this check to `onRandomTick`.

## Changes
Sugarcane no longer crashes the server, trying to grow past the top of the world.

## Backwards compatibility
`Yes`


dktapps:
Thanks. The growth code is duplicated, so I fixed this by deduplicating it instead (see 014317381).

PR #3894, opened 2020-11-07
Update Item.php

davidgamingzz:
## Introduction

Removing a nonexistent tag.


IceCruelStuff:
https://github.com/pmmp/PocketMine-MP/blob/47976bac34bb03453ea3258d35b1bce808ee3a87/src/item/Item.php#L61

kenygamer:
How is this an unused tag? `Item::TAG_DISPLAY` is the CompoundTag tag to store `TAG_DISPLAY_NAME` and `TAG_DISPLAY_LORE` StringTag tags.

PR #3896, opened 2020-11-08
Match vanilla armor equiping

Driesboy:
## Introduction

Since a few MCPE versions ago, fast armor equipping has changed:

- Both clicking air & blocks trigger this behavior. 
- You can replace armor that's already in your armor inventory

### Relevant issues

* None

### API changes

None

### Behavioural changes

Matches vanilla again

## Backwards compatibility

Nothing has to be changed

## Tests

Join the server and fast equip an item from your hotbar.

### DIFF
--- a/src/item/Armor.php
+++ b/src/item/Armor.php
@@ -122,13 +122,25 @@ class Armor extends Durable{
 		return 0;
 	}
 
-	public function onActivate(Player $player, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector) : ItemUseResult{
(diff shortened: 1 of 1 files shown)


dktapps:
Clicking a block should already fire block and air interactions simultaneously, if my memory is correct.

Driesboy:
Well it didn't work when I tested it.

dktapps:
This doesn't match vanilla for several reasons:
- The action will fire twice when right-clicking on a block
- The item used should be directly swapped for the item in the corresponding armour slot. This won't happen with this code because the activated item still occupies the target slot when the removed armour item is added to the main inventory.

PR #3923, opened 2020-11-22
Added support for emotes

buchwasa:
## Introduction
This pull request aims to add support for emotes.

### API changes
Added:
- PlayerEmoteEvent with method, getEmoteId()

## Follow-up
- I wasn't too sure if I should've implemented the event into the Player class, similar to PlayerActionPacket, would like some feedback on if I should still do that.
- EmoteListPacket is sent, but I ignore it, I couldn't think of a good reason to keep track of the emote list, unless you're forcing the player to emote.

## Tests
There isn't much to show that I tested, I joined a server, emoted, works.


src/event/player/PlayerEmoteEvent.php
```diff
@@ -0,0 +1,43 @@
...
+namespace pocketmine\event\player;
+
+use pocketmine\event\Cancellable;
+use pocketmine\event\CancellableTrait;
+use pocketmine\player\Player;
+
+class PlayerEmoteEvent extends PlayerEvent implements Cancellable{
+	use CancellableTrait;
+
+	private $emoteId;
```
Jviguy, line 34: would recommend phpdocs for properties dylan will **REEEE** at you for not having them

dktapps:
This is a potential secvuln since it's not rate limited or validated, but directly rebroadcasted.

src/event/player/PlayerEmoteEvent.php
dktapps: They aren't recommended, they are required for PHPStan analysis.

PR #3934, opened 2020-11-28
Inventory variables may be null

MCPE-PC:
## Introduction
I got a `TypeError` when I called `Living::getArmorInventory()` in `PlayerDeathEvent` and using `catch ($e)` is annoying

## Changes
 * `Living::getArmorInventory()`: Made the return type nullable
 * `Human::getInventory()`: Moved type hinting from the document to PHP for type checking, with nullable
 * No need to `try...catch`: With proper type hinting

## Backwards compatibility
Compatible


dktapps:
These methods won't return null unless you kicked a player during an event, which is a problem of its own.

PR #3937, opened 2020-11-29
Removal of permission defaults (in favour of permission cascading)

dktapps:
## Introduction
Permissions have supported cascading forever. Any permission is able to grant (or deny) any other arbitrary permission it wants to.
This also means that any permission can make itself be granted by another permission.

Introducing the following new permission nodes:
- `pocketmine.group.operator`: all `op` permissions (including plugin permissions) have this permission as their parent. If this permission is granted, it's equivalent to having `isOp() === true`.
- `pocketmine.group.everyone`: all `true` permissions (including plugin permissions) have this permission as their parent.

In addition, all references to `Permission` `default` are removed. The default availability of a permission now depends on which permission node it's granted by.

`default` in `plugin.yml` is still supported, since it essentially just means "this permission belongs to `pocketmine.group.operator` by default".

## Rationale
It came to my attention that the whole concept of "op" vs "non op" could be implemented in a far more flexible and powerful way using functionality that the permission system **already offered since day one**.

As it's no doubt clear from the size of the diff, this change makes a significant simplification to the internals of `PermissionManager`, as well as opening the gates for a much more flexible system of permission assignment that's no longer limited to op/notop/true/false.

### API changes
The following API methods have been removed:
- `Permissible->isOp()`: use `Permissible->hasPermission(DefaultPermissions::ROOT_OPERATOR)` instead, **but you really shouldn't directly depend on a player's op status, add your own permissions instead!**
- `Permissible->setOp()`: use `addAttachment($plugin, DefaultPermissions::ROOT_OPERATOR, true)` instead to add, and `removeAttachment()` to remove it (or addAttachment() with false to explicitly deny it, just like any other permission)
- `Permission->getDefault()`
- `Permission->setDefault()`
- `PermissionManager->getDefaultPermissions()`
- `PermissionManager->recalculatePermissionDefaults()`
- `PermissionManager->subscribeToDefaultPerms()`
- `PermissionManager->unsubscribeFromDefaultPerms()`
- `PermissionManager->getDefaultPermSubscriptions()`
The following fields have been removed:
- `Permission::$DEFAULT_PERMISSION`
The following constants have been moved:
- `Permission::DEFAULT_FALSE` -> `PermissionParser::DEFAULT_FALSE`
- `Permission::DEFAULT_TRUE` -> `PermissionParser::DEFAULT_TRUE`
- `Permission::DEFAULT_OP` -> `PermissionParser::DEFAULT_OP`
- `Permission::DEFAULT_NOT_OP` -> `PermissionParser::DEFAULT_NOT_OP`
The following new permission nodes have been introduced:
- `pocketmine.group.everyone`
- `pocketmine.group.operator`

### Behavioural changes
The server should behave the same way it did before this change.

## Backwards compatibility
This is BC breaking.

## Tests
Testing has been done (including `notop`) and everything is found to be working. Moreover, the `/dumpmemory` command's permission is now consistently denied to operators (as intended). I intend to introduce a CLI permission which will be assigned to `ConsoleCommandSender` for that.

## Follow-up
- In the future there will be a more powerful system for permission defaults in `plugin.yml`, to allow for permissions to belong to multiple groups simultaneously, and also allow for permissions to inherit from permissions in another plugin with no additional code.
- I also intend to implement a `pocketmine.group.console` permission which will allow permissions to be assigned exclusively for the console. This will be used for stuff like `/dumpmemory`, as well as DevTools' `/makeplugin`, `/extractplugin` etc.

### DIFF
--- a/src/permission/PermissibleBase.php
+++ b/src/permission/PermissibleBase.php
@@ -46,10 +46,15 @@ class PermissibleBase implements Permissible{
 
 	public function __construct(?Permissible $permissible, bool $isOp){
 		$this->parent = $permissible;
-		$this->setCorePermission(DefaultPermissions::ROOT_USER);
+
+		//TODO: we can't setBasePermission here directly due to bad architecture that causes recalculatePermissions to explode
+		//so, this hack has to be done here to prevent permission recalculations until it's fixed...
+		$rootPerms = [DefaultPermissions::ROOT_USER => true];
 		if($isOp){
-			//this HAS to be set AFTER the user perm, so that notop overrides work correctly
-			$this->setCorePermission(DefaultPermissions::ROOT_OPERATOR);
+			$rootPerms[DefaultPermissions::ROOT_OPERATOR] = true;
+		}
+		foreach($rootPerms as $perm => $isGranted){
+			$this->rootPermissions[$perm] = new PermissionAttachmentInfo($this->getRootPermissible(), $perm, null, $isGranted);
 		}
 		//TODO: permissions need to be recalculated here, or inherited permissions won't work
 	}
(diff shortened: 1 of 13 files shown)


dktapps:
The issue with `/dumpmemory`'s unreliable permission is caused by its parent permission `pocketmine.command` implicitly having a default value of `op`, which overrides `pocketmine.command.dumpmemory`'s permission of `false` after permission recalculation.
It seems the reason this PR fixes the problem is because `pocketmine.group.operator` does not grant `pocketmine.command` permission.

On a related note, I'm considering removing `pocketmine.command` and other related "parent" permission nodes, since they serve as nothing more than predefined permission groups. It would be better to standardise the permission name format so that permission plugins can use pattern matching to implement their own permission grouping mechanisms (e.g. `pocketmine.command.*`) and remove these predefined parent permission nodes, which are not logically guaranteed to be consistent anyway and are hardcoded.

dktapps:
I've tested the remaining elements of this PR and they are working correctly (including permissions with `notop` defaults).

src/permission/PermissibleBase.php
```diff
@@ -137,13 +155,10 @@ public function recalculatePermissions() : void{
...
 	public function clearPermissions() : void{
-		$permManager = PermissionManager::getInstance();
-		$permManager->unsubscribeFromAllPermissions($this->getRootPermissible());
-
-		$permManager->unsubscribeFromDefaultPerms(false, $this->getRootPermissible());
-		$permManager->unsubscribeFromDefaultPerms(true, $this->getRootPermissible());
+		PermissionManager::getInstance()->unsubscribeFromAllPermissions($this->getRootPermissible());
 
 		$this->permissions = [];
+		$this->rootPermissions = [];
```
dktapps, line 161: Need to re-review this. I'm not sure it makes sense, but it's here to break cyclic dependencies between `PermissibleBase` and `PermissionAttachmentInfo`.

PR #3940, opened 2020-11-30
Fixed crash when executing command /version with multiple authors declared in 'author'

### LINKED ISSUE #3902: /version crashes server when author is an array
### Issue description
When a plugin declares `author: [array, of, strings]`, PocketMine still loads the plugin, but it crashes the server when `/version $PLUGIN_NAME` is run.

This opens a vulnerability for adversaries to induce crashes on a server by running `/version` on plugins known to have incorrectly formatted plugin.yml

Reported by @Kygekraqmak

1. Create a plugin with this line in plugin.yml: `author: [array]`
2. Start the server
3. Run command `/version $PLUGIN_NAME`

### OS and versions

* PocketMine-MP: 3.15.4
* PHP: PHP 7.4.11 (cli) (built: Nov 10 2020 17:50:51) ( ZTS )
* Server OS: Linux (Docker image)

### Plugins
Reproduced with only DevTools + aforementioned example plugin

Example crash: https://crash.pmmp.io/view/4634552

DerDevHD:
## Introduction
I made this pull request to fix an issue with the /version command.
When providing a plugin with an array value in 'author', it would crash, because the array won't be parsed into `authors` property.

To fix this, I've made a check if its an array, if so, the array gets "parsed" into `authors` property, like said above.

### Relevant issues
- Fixes #3902 

### API changes
/

### Behavioural changes
/

## Backwards compatibility
Everything fine with it.

## Follow-up
/

## Tests
The tests with some plugin which have an array-value in author section went successful.
The `plugin.yml` can look like this:
```yaml
name: "TestPlugin"
version: 1.0.0
description: "Just for testing the API..."
author: ["FirstDev", "SecondDev"]
authors: ["Another", "AndAnother"]
api:
  - 3.0.0
main: TestPlugin\TestPlugin
```
Before my fix, the server would crash, now, everything goes well.
The if-case for the grammar with "Author" and "Authors" is still working like before.

You can perform the test also, just run `/version TestPlugin` in the console/ingame.
## Some last words
I hope you are fine with my pull request... please don't worry too much about it, its my first PR for PocketMine itself.

Best Regards.

### DIFF
--- a/src/pocketmine/plugin/PluginDescription.php
+++ b/src/pocketmine/plugin/PluginDescription.php
@@ -157,9 +157,7 @@ class PluginDescription{
 		$this->authors = [];
 		if(isset($plugin["author"])){
 			if(is_array($authors = $plugin["author"])){
-				foreach($authors as $author){
-					$this->authors[] = $author;
-				}
+				$this->authors = $authors;
 			}else{
 				$this->authors[] = $plugin["author"];
 			}
--- a/src/pocketmine/plugin/PluginDescription.php
+++ b/src/pocketmine/plugin/PluginDescription.php
@@ -157,7 +157,7 @@ class PluginDescription{
 		$this->authors = [];
 		if(isset($plugin["author"])){
 			if(is_array($plugin["author"])){
-				$this->authors = $authors["author"];
+				$this->authors = $plugin["author"];
 			}else{
 				$this->authors[] = $plugin["author"];
 			}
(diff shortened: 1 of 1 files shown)


TwistedAsylumMC:
"Author" isn't plural so there shouldn't be multiple authors inside of it. The "authors" property should be used if multiple authors need to be listed

dktapps:
As I said to @SOF3 on discord, I don't actually know why we have two author properties to begin with. It's been discussed to make them both the same.

Since it's a choice between turning this into an error (just so you can add an extra `s` to do the same thing) or making it work the same as `authors`, I think this is OK.

PR #3966, opened 2020-12-10
Setup GitHub Actions

dktapps:
Travis CI has given OSS a big middle finger, suspending our builds without any notification at all.

I'd avoided migrating to GH Actions for a while because I knew it was going to be a colossal pain in the ass (like setting up a CI always is), but the time has come whether I like it or not.


dktapps:
Blocking issue: https://github.com/shivammathur/setup-php/issues/363

dktapps:
Looks like we're going to have to build our own PHP binaries for this.

Issue #3965, opened 2020-12-10
Hoes doesn't loose durability when breaking sponges


JohnnyX071:
Hello, I found a bug where when you break sponges with a hoe, it doesn't lose durability. I tried with 0 plugins and this was still happening.

Pocketmine version: 3.17.0
PHP: newest (7)
Game version: Win10 Pro, Mc version 1.16.200

Mohagames205:
Hey, I have successfully identified the cause of this problem. It seems like the Hoe item class does not have an onDestroyBlock method, which means no damage will be applied when it's used. I'll try to make a PR to fix this.

JohnnyX071:
ok good!

PR #3967, opened 2020-12-11
Fix that a hoe gets damage applied to it, when it's used to break a block

Mohagames205:
## Introduction
Right now when a hoe is used to break a block, no damage is applied to the hoe. This PR aims to fix that.

### Relevant issues
#3965 

* Damage is added when the hoe is used

### API changes
Nope no API changes

### Behavioural changes
Damage is added to a hoe when it is used to break a block. 

## Backwards compatibility
No BC break

## Follow-up
~~* We probably still need to add the hoe tool to the BlockToolType interface. However, I do not know its value. I assume the values are just random bit values. But I am not sure~~ I added the hoe to the BlockToolType interface
* I don't know the correct damage value for the hoe

## Tests
Before the changes:
https://file.mohamedelyousfi.be/taRO5/seMADENI98.mp4
After the changes:
https://file.mohamedelyousfi.be/taRO5/SIYOCUzu30.mp4

The "dura" in chat indicates the damage on the item and not the durability.

This is one of my first PR's, so feedback is appreciated :)

### DIFF
--- a/src/pocketmine/block/BlockToolType.php
+++ b/src/pocketmine/block/BlockToolType.php
@@ -35,5 +35,6 @@ interface BlockToolType{
 	public const TYPE_PICKAXE = 1 << 2;
 	public const TYPE_AXE = 1 << 3;
 	public const TYPE_SHEARS = 1 << 4;
+	public const TYPE_HOE = 1 << 5;
 
 }
--- a/src/pocketmine/block/Sponge.php
+++ b/src/pocketmine/block/Sponge.php
@@ -31,6 +31,10 @@ class Sponge extends Solid{
 		$this->meta = $meta;
 	}
 
+	public function getToolType() : int{
+		return BlockToolType::TYPE_HOE;
+	}
+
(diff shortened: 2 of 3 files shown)


dktapps:
Do hoes take damage when breaking any block?

Mohagames205:
Yes, I should've mentioned that I tried this in vanilla Minecraft. I'll attach a video

Mohagames205:
As you can see, when I break the blocks damage gets applied to the hoe

https://file.mohamedelyousfi.be/taRO5/meYaSaPI16.mp4

dktapps:
Travis CI has been suspended for this repo; does it work on your fork?

Mohagames205:
Yup, travis works for my fork.

https://travis-ci.com/github/Mohagames205/PocketMine-MP/builds/208210441 and https://travis-ci.com/github/Mohagames205/PocketMine-MP/builds/208212990

dktapps:
Perfect, thanks.

PR #3968, opened 2020-12-14
A hook function for plugins to use newly loaded soft dependencies

jasonw4331:
## Introduction

PocketMine's existing plugin loading system doesn't inform already loaded plugins if a soft dependency is loaded.
This PR intends to resolve that issue by providing plugin developers with an API method which is called whenever a soft dependency is loaded.
### Relevant issues

Issue #2825 can be considered a counterpoint which resolves the issue by removing the root cause entirely

### API changes

A new hook function `onDependencyResolution()` is added to the `Plugin` class for developers to hook into for dependency tracking.

## Backwards compatibility

No backwards compatibility changes are made. The new hook function is optional to use.
## Follow-up

N/A
## Tests

```php
<?php
declare(strict_types=1);
/**
 * @name DependHookTest
 * @main jasonwynn10\DependHookTest\Main
 * @version 0.1.0
 * @api 4.0.0
 * @description A test script
 * @author jasonwynn10
 * @load STARTUP
 * @softdepend unloadedPlugin
 */
namespace jasonwynn10\DependHookTest {

	class Main extends PluginBase {
		public $softDepend = null;
		public function onLoad() : void {
			// will return null because plugin is unloaded
			$this->softDepend = $this->getServer()->getPluginManager()->getPlugin("unloadedPlugin");
			if($this->softDepend !== null) {
				// enable economy features idk
			}
		}
		public function onDependencyResolution() : void {
			// will return plugin instance because plugin is now loaded
			$this->softDepend = $this->getServer()->getPluginManager()->getPlugin("unloadedPlugin");
			if($this->softDepend !== null) {
				// enable economy features
			}
		}
	}
}
```

### DIFF
--- a/src/plugin/PluginManager.php
+++ b/src/plugin/PluginManager.php
@@ -198,6 +198,7 @@ class PluginManager{
 					 */
 					$plugin = new $mainClass($loader, $this->server, $description, $dataFolder, $prefixed, new DiskResourceProvider($prefixed . "/resources/"));
 					$this->plugins[$plugin->getDescription()->getName()] = $plugin;
+					$this->resolveLoadedDependency($plugin);
 
 					return $plugin;
 				}
@@ -385,6 +386,19 @@ class PluginManager{
 		}
 
 		return $loadedPlugins;
+	}
+
+	public function resolveLoadedDependency(Plugin $plugin) : int {
+		$resolved = 0;
+		foreach($this->plugins as $loadedPlugin) {
+			foreach($loadedPlugin->getDescription()->getSoftDepend() as $pluginName) {
+				if($plugin->getName() === $pluginName) {
+					$loadedPlugin->onDependencyResolution();
+					$resolved += 1;
+				}
+			}
+		}
+		return $resolved;
 	}
 
 	public function isPluginEnabled(Plugin $plugin) : bool{
--- a/src/plugin/Plugin.php
+++ b/src/plugin/Plugin.php
@@ -39,6 +39,13 @@ interface Plugin{
 	public function isEnabled() : bool;
 
 	/**
+	 * Called when a plugin soft dependency is resolved
+	 *
+	 * @return void
+	 */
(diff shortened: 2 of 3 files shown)


src/plugin/PluginManager.php
```diff
@@ -387,6 +388,19 @@ public function loadPlugins(string $directory, ?array $newLoaders = null) : arra
 		return $loadedPlugins;
 	}
 
+	public function resolveLoadedDependency(Plugin $plugin) : int {
```
Jviguy, line 391: should there be php-doc on this?

TheNewHEROBRINEX: PocketMine code uses phpdocs on functions only when it is necessary to add information that is not already provided by type hints and return types (for example the type of elements of an array).

jasonwynn10: I will probably add a phpdoc to indicate what the return value means

Jviguy: Dylan said to me that it was needed for phpstan so I just figured it's needed on functions too

dktapps: > only when it is necessary to add information that is not already provided by type hints and return types (for example the type of elements of an array).

dktapps:
IIRC, this PR arose from a discussion about possible loading of soft dependencies after dependents despite best efforts in #3971 .
However, this is a very exotic edge case that only comes into play when a plugin-loading plugin has a soft dependency that gets resolved by another plugin-loading plugin after it decides to give up and not bother trying to resolve the soft dependency.

My verdict is that this is a rare enough situation that it's probably not worth adding a hook for it (besides, it's already possible to do something similar to this using PluginEnableEvent).

dktapps:
I'm closing this for reasons mentioned in my comment above - it's not worthwhile. In addition, a better solution to the overall problem is #2825 , which addresses a wide array of problems including this one.

PR #3973, opened 2020-12-17
Fixed a bucket of lava disappearing when used in a furnace

### LINKED ISSUE #2385: Lava bucket disappeared when using Furnace
### Issue description

Putting a lava bucket in the Furnace will not leave the empty bucket as in Vanilla

- Expected result: I expect that the lava bucket will turn into empty bucket like Vanilla
- Actual result: Lava bucket is gone, along with the bucket

1. Putting the Lava Bucket inside the Furnace
2. The Lava Bucket disappeared, not leaving any trace behind

### OS and versions

* PocketMine-MP:  3.1.4 commit 17f7dc3
* PHP: 7.2
* Server OS: Windows Server 2008 R2 64-bit
* Game version: PE/Win10

### Plugins

- Do not paste crashdumps into an issue - please use our Crash Archive at https://crash.pmmp.io for submitting crash reports to not spam the issue tracker. Add links to your reports in the Crash Archive here.
- Please use gist or anything else to add other files and add links here

* ...

Govdim:
## Introduction

This pull request fixes the disappearance of a lava bucket when used in a furnace.

It was a little strange to add a method for the sake of one item
But maybe someone wants to create custom items and this method will definitely come in handy :)

### Relevant issues

* Fixes #2385

### DIFF
--- a/src/pocketmine/item/Item.php
+++ b/src/pocketmine/item/Item.php
@@ -656,6 +656,14 @@ class Item implements ItemIds, \JsonSerializable{
 	}
 
 	/**
+	 * Returns an item after burning fuel
+	 */
+	public function getFuelResidue() : Item{
+		$this->pop();
+		return $this;
+	}
+
+	/**
 	 * Returns how many points of damage this item will deal to an entity when used as a weapon.
 	 */
 	public function getAttackPoints() : int{
--- a/src/pocketmine/item/Item.php
+++ b/src/pocketmine/item/Item.php
@@ -659,8 +659,10 @@ class Item implements ItemIds, \JsonSerializable{
 	 * Returns an item after burning fuel
 	 */
 	public function getFuelResidue() : Item{
-		$this->pop();
-		return $this;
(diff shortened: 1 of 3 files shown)


src/pocketmine/item/Item.php
```diff
@@ -655,6 +655,14 @@ public function getFuelTime() : int{
 		return 0;
 	}
 
+	/**
+	 * Returns an item after burning fuel
+	 */
+	public function getFuelResidue() : Item{
+		$this->pop();
```
dktapps, line 662: This will produce unexpected behaviour when called multiple times. It should clone itself first.

dktapps (changes requested):
Looks good, just some minor nits.

src/pocketmine/item/Item.php
Govdim: I also thought about it at first, but then I forgot XD

dktapps (approved):

PR #3981, opened 2020-12-26
Let SingletonTrait Idea friendly

Blackjack200:
This commit add document in order to let this trait friendly to PHPStorm

## Introduction
Let SingletonTrait Idea friendly

That add document in order to let this trait friendly to PHPStorm
## Tests
tested

### DIFF
--- a/src/utils/SingletonTrait.php
+++ b/src/utils/SingletonTrait.php
@@ -27,10 +27,12 @@ trait SingletonTrait{
 	/** @var self|null */
 	private static $instance = null;
 
+	/** @return static */
 	private static function make() : self{
 		return new self;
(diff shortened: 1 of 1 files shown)


dktapps:
PhpStorm sucks. `self` in a trait is supposed to refer to the class including the trait.
In addition, this breaks PHPStan analysis, and PHPStan analysis is far more useful than PhpStorm's broken ass.

Ifera:
Or just simply downgrade phpstorm.. 2020.3 is not that good afterall..

dktapps:
2020.3 is hands-down the most broken PhpStorm update I've seen in 4 years.

Issue #3952, opened 2020-12-04
TNT launch sound missing


Blackjack200:
### Issue description

- Expected result: ignite TNT with launch sound
- Actual result: tnt actived but launch sound is missing

### Steps to reproduce the issue
1. ... place tnt
2. ... use steel to execute it and listen to sound

### OS and versions
* PocketMine-MP: 4.0.0 c5e12bb11f6b810e1059e580fda9fb726cda6284
* PHP: `PHP 7.4.12 (cli) (built: Nov 21 2020 16:18:37) ( ZTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies`
* Server OS: Ubuntu 18.06 `Linux root 4.15.0-124-generic #127-Ubuntu SMP Fri Nov 6 10:54:43 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`
* Game version: PE `without any resource packs 1.16.100`

### Plugins
None

Mohagames205:
I just tested this on Windows with

- PocketMine-MP 4.0.0 (5282ae3) 
- Windows 10 
- PHP 7.4
- Minecraft Bedrock Edition 1.16.200

To be specific, the explosion sound is still there. The sound you can't hear is when you ignite the TNT, so the sound right before the actual explosion. 

https://github.com/pmmp/PocketMine-MP/blob/master/src/entity/object/PrimedTNT.php#L77

The issue that the getViewers() method returns 0 Players even though there are players nearby. So this means that when the sound is broadcasted, it should be broadcasted to the viewers. But there aren't any according to PocketMine-MP so no one hears the sound.

https://github.com/pmmp/PocketMine-MP/blob/master/src/entity/Entity.php#L1622



I didn't have enough time to further investigate this.

dktapps:
Nice work @Mohagames205 , that's perfect.

When `initEntity()` is called, `Entity->spawnToAll()` has not yet been called, so nobody will be able to see (or hear) the entity at that point. This wouldn't previously have happened since sounds were broadcasted irrespective of the current viewers, so you'd hear the sound regardless of whether you could see the TNT or not.

Yosshi999:
Moving `broadcastSound(new IgniteSound());` to https://github.com/pmmp/PocketMine-MP/blob/master/src/block/TNT.php#L104 will solve this, right?

ConnorHopper:
no

dktapps:
Yes.

PR #3996, opened 2021-01-07
Fixed TNT broadcasting ignition sound

Mohagames205:
## Introduction
This PR fixes the issue of the TNT ignite-sound not being broadcasted by the server correctly.

### Relevant issues
#3952 

## Changes
The broadcast sound method has been moved to the TNT class AFTER the spawnToAll() method has been called. This way the getViewer() method will be used correctly and the sound will be broadcasted to all the viewers, instead of no one.

### Behavioural changes
You can hear the TNT ignite

## Tests
See video
https://user-images.githubusercontent.com/40402787/103944544-1953be80-5134-11eb-88ce-aeb83a48c86a.mp4


dktapps (approved):
Aside from the bug, it's generally a bad idea to interact with the world during initEntity() anyway, so this kills two birds with one stone.

PR #4005, opened 2021-01-18
Create task with callable

AkmalFairuz:
## Introduction

You can create task with this method:
```php
$scheduler = $this->getScheduler();
$scheduler->scheduleDelayedTask(function ($currentTick) {
      $this->getLogger()->info("Hello World");
}, 20);
```
To fix developer lazy to create new class.

### DIFF
--- a/src/pocketmine/scheduler/TaskScheduler.php
+++ b/src/pocketmine/scheduler/TaskScheduler.php
@@ -60,30 +60,34 @@ class TaskScheduler{
 	}
 
 	/**
+	 * @param Task|callable $task
 	 * @return TaskHandler
 	 */
-	public function scheduleTask(Task $task){
+	public function scheduleTask($task){
 		return $this->addTask($task, -1, -1);
 	}
 
 	/**
+	 * @param Task|callable $task
 	 * @return TaskHandler
 	 */
-	public function scheduleDelayedTask(Task $task, int $delay){
+	public function scheduleDelayedTask($task, int $delay){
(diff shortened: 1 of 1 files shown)


SOF3:
Not exactly relevant, but in the long term I expect to eliminate delayed tasks completely in favor of async-based sleeping.

SOF3:
There is already [`ClosureTask`](https://github.com/pmmp/PocketMine-MP/blob/stable/src/pocketmine/scheduler/ClosureTask.php), so developer does not have to create a new class anyway.

```php
$scheduler = $this->getScheduler();
$scheduler->scheduleDelayedTask(new ClosureTask(function ($currentTick) {
      $this->getLogger()->info("Hello World");
}), 20);
```

I don't think there is further need to simplify this code.

I agree that the use of single-abstract-method (SAM) interfaces is cargo cult in PHP, but since it is already widely used in the ecosystem, I don't think it's worth the cost of breaking backward compatibility to remove Tasks. As for union parameter types, I believe it is frowned upon in PocketMine code style due to confusion with type checking.

dktapps:
Union types are frowned on because they are generally an anti-feature whose usage suggests an API design deficit.
The presence of union types in code is usually an indication that method overloading is needed, not unions.

In this case, neither overloading nor unions are needed because there's already `ClosureTask`, so this is just adding extra complexity for no real benefit. Anyone who would benefit from this change is already benefiting from `ClosureTask` which was added 2 years ago.

PR #4007, opened 2021-01-19
InventoryPickupItemEvent: rename getItem() method to getItemEntity()

SalmonDE:
## Introduction

This is pretty much a continuation of b39bbffdc55f40898ec1d0b0cb31eec416b7aa02, where the former class of an item entity was renamed to ItemEntity, which distinguished it from the Item class, the representation of an item within an inventory.

The ambiguity from back then is still evident in the InventoryPickupItemEvent. This example line of code should illustrate why:

`$event->getItem()->getItem()->getName()` (whereas $event is an instance of InventoryPickupItemEvent)

As you can see, getting the name of an item involves the strange chain of two "->getItem()" calls. The first one returns an instance of ItemEntity, while the second call returns an instance of Item.

### API changes

`InventoryPickupItemEvent->getItem()` renamed to `InventoryPickupItemEvent->getItemEntity()`

### Behavioural changes

none

## Backwards compatibility

This is clearly a backwards-incompatible change, hence the pr is targeted at the master branch in order to take effect in the 4.0 release.

## Follow-up

None that I can think of.

## Tests

I am not aware of any relevant tests.

### DIFF
--- a/src/event/inventory/InventoryPickupItemEvent.php
+++ b/src/event/inventory/InventoryPickupItemEvent.php
@@ -32,14 +32,14 @@ class InventoryPickupItemEvent extends InventoryEvent implements Cancellable{
 	use CancellableTrait;
 
 	/** @var ItemEntity */
-	private $item;
+	private $itemEntity;
 
-	public function __construct(Inventory $inventory, ItemEntity $item){
-		$this->item = $item;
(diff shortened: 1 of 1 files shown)


dktapps:
CI failures are unrelated.

Issue #4002, opened 2021-01-12
Add PlayerDeathEvent->setKeepXp()


dktapps:
## Description
As per title.

## Justification
It's currently cumbersome to implement this from a plugin - you have to capture the XP value at death, prevent it being dropped, and then re-set it after respawn (which still doesn't work because PlayerRespawnEvent is called too early).

## Alternative methods
A stateful mess in a plugin?

PR #4015, opened 2021-01-28
Add PlayerDeathEvent->setKeepXp() as per #4002

TheNewHEROBRINEX:
## Introduction
This PR aims to reduce the complexity needed for a plugin to prevent loss of experience upon player death by adding the method setKeepXp to PlayerDeathEvent.

### Relevant issues

The need for this addition was expressed in #4002.

### API changes

Added PlayerDeathEvent->setKeepXp().

### Behavioural changes

None when no plugin is using the new API method.

## Backwards compatibility

Yes

## Follow-up

None

## Tests

This PR was tested using [this script plugin](https://gist.github.com/TheNewHEROBRINEX/fc01eceb0a6cabb068814343e344d14b) with the related video that can be seen [here](https://youtu.be/o2tqXO8zsfo).


TheNewHEROBRINEX:
In the video, I also take damage on purpose before doing /kill to check that other attributes are still being reset properly.

src/event/player/PlayerDeathEvent.php
```diff
@@ -41,6 +41,8 @@ class PlayerDeathEvent extends EntityDeathEvent{
 	private $deathMessage;
 	/** @var bool */
 	private $keepInventory = false;
+	/** @var bool */
+	private $keepXp = false;
```
dktapps, line 45: You can use a native type for this. master is 7.4+ only now.

TheNewHEROBRINEX: Do I have to keep the getter and the setter anyway for code consistency?

dktapps: I'm not telling you to remove those.

TheNewHEROBRINEX: I know, I was just asking

src/player/Player.php
```diff
@@ -2135,6 +2138,10 @@ public function respawn() : void{
 		$this->setHealth($this->getMaxHealth());
 
 		foreach($this->attributeMap->getAll() as $attr){
+			if($attr->getId() === Attribute::EXPERIENCE or $attr->getId() === Attribute::EXPERIENCE_LEVEL){ //we have already reset both of those if needed when the player died
+				$this->getNetworkSession()->syncAttributes($this, [$attr]);
```
dktapps, line 2142: why is this necessary?

TheNewHEROBRINEX: Because the client on respawn assumes vanilla behaviour and automatically resets experience progress and level to 0. Without this when your respawn it will appear that you have lost your experience but if you leave and join again it appears back.

dktapps: I think markSynchronized(false) should be sufficient.

dktapps: Note to self: this reset might also be causing some other problems (for example setting health to max before resetting attribute defaults might lead to health overflow).

TheNewHEROBRINEX:
I tested the PR again and confirmed that it is working

dktapps (approved):
Tested and found working, apologies for the long delay ...

PR #4030, opened 2021-02-09
Added Hopper::setFacing() and Hopper::getFacing()

Govdim:
## Introduction
These methods are very important if someone wants to implement the hopper behavior.

## API changes
Added:
`Hopper::getFacing()`
`Hopper::setFacing()`

### DIFF
--- a/src/block/Hopper.php
+++ b/src/block/Hopper.php
@@ -38,7 +38,7 @@ class Hopper extends Transparent{
 	use PoweredByRedstoneTrait;
 
 	/** @var int */
-	private $facing = Facing::DOWN;
+	protected $facing = Facing::DOWN;
 
 	public function readStateFromData(int $id, int $stateMeta) : void{
 		$facing = BlockDataSerializer::readFacing($stateMeta & 0x07);
@@ -55,6 +55,17 @@ class Hopper extends Transparent{
 
 	public function getStateBitmask() : int{
 		return 0b1111;
+	}
+
+	public function getFacing() : int{ return $this->facing; }
+
+	/** @return $this */
+	public function setFacing(int $facing) : self{
+		if($facing === Facing::UP){
+			throw new \InvalidArgumentException("Hopper may not face upward");
+		}
+		$this->facing = $facing;
+		return $this;
 	}
 
 	protected function recalculateCollisionBoxes() : array{
--- a/src/block/Hopper.php
+++ b/src/block/Hopper.php
@@ -38,7 +38,7 @@ class Hopper extends Transparent{
 	use PoweredByRedstoneTrait;
 
 	/** @var int */
-	protected $facing = Facing::DOWN;
+	private $facing = Facing::DOWN;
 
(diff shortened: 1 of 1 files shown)


src/block/Hopper.php
```diff
@@ -38,7 +38,7 @@ class Hopper extends Transparent{
 	use PoweredByRedstoneTrait;
 
 	/** @var int */
-	private $facing = Facing::DOWN;
+	protected $facing = Facing::DOWN;
```
SOF3, line 41: Why do you change it to `protected`?

src/block/Hopper.php
```diff
@@ -57,6 +57,17 @@ public function getStateBitmask() : int{
 		return 0b1111;
 	}
 
+	public function getFacing() : int{ return $this->facing; }
```
SOF3, line 60: PocketMine code style. One line per statement.

src/block/Hopper.php
```diff
@@ -57,6 +57,17 @@ public function getStateBitmask() : int{
...
+
+	/** @return $this */
+	public function setFacing(int $facing) : self{
+		if($facing === Facing::UP){
+			throw new \InvalidArgumentException("Hopper may not face upward");
+		}
+		$this->facing = $facing;
+		return $this;
```
SOF3, line 68: Is there a particular reason to make this a fluent setter?

src/block/Hopper.php
Govdim: Because in all such cases the variable is protected
Or is there a special reason for the private of this variable in the hopper?

src/block/Hopper.php
Govdim: Something is wrong? : /

src/block/Hopper.php
SOF3: Is there a special reason to be protected?

src/block/Hopper.php
SOF3, line 60:
```suggestion
	public function getFacing() : int{
		return $this->facing;
	}
```

src/block/Hopper.php
dktapps: Unless there is a reason for increasing the visibility, the visibility should not be increased.

src/block/Hopper.php
dktapps: It's standard.

src/block/Hopper.php
dktapps: This isn't necessary. It's consistent with other implementations.

dktapps:
This looks OK minus the `facing` visibility change (don't change it unless necessary). I actually have an almost identical changeset in my local workspace already (I've been working on these recently as I prepare for 1.13 world support).

Govdim:
> This looks OK minus the `facing` visibility change (don't change it unless necessary). I actually have an almost identical changeset in my local workspace already (I've been working on these recently as I prepare for 1.13 world support).

Ok I'll fix it

PR #4037, opened 2021-02-15
Update BUILDING.md

mmm545:
A work around to fix such error
```
Fatal error: Uncaught BadMethodCallException: unable to create temporary file in PocketMine-MP/build/server-phar.php:119
```

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -34,11 +34,11 @@ Run `composer make-server` using your preferred PHP binary. It'll drop a `Pocket
 
 You can also use the `--out` option to change the output filename.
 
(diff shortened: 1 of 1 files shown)


dktapps:
Please also mention that this is a bug in PHP, so that people don't come nagging me to fix it.

dktapps:
For posterity, it was fixed in this commit: https://github.com/php/php-src/commit/1bb2a4f91cfde36ba343184f3d88189e646cc363
It appears on the 7.4 and 8.0 branches, so I assume the next set of patches releases for 7.4 and 8.0 will have this fix.

PR #4081, opened 2021-03-15
Update argument type constants

Unickorn:
## Introduction
This PR updates the argument type constants in AvailableCommandsPacket and adds the new `WILDCARD_TARGET` argument type, as they were once again shifted by Mojang with their, and I quote, "some kind of dynamic ID bullshit".

### API changes
N/A

### Behavioural changes
Command argument type constants now work fine.

## Backwards compatibility
No issues with backwards compatibility.


dktapps:
This should target stable.

Have you tested these?

Unickorn:
I have just tested them by sending them one by one, not sure how I can provide proper test results.

Unickorn:
Uh oh

dktapps:
> I have just tested them by sending them one by one, not sure how I can provide proper test results.

that's fine it's not really practical to unit test these anyway. I'm just double-checking that someone has actually verified these are correct and not just providing random copy pasted values.

PR #4085, opened 2021-03-17
Fix: Items can't be removed from item frames in Survival mode

HighTecDev:


src/pocketmine/Player.php
```diff
@@ -2941,6 +2942,13 @@ public function handlePlayerAction(PlayerActionPacket $packet) : bool{
 						$this->level->broadcastLevelEvent($pos, LevelEventPacket::EVENT_BLOCK_START_BREAK, (int) (65535 / $breakTime));
 					}
 				}
+				if($target instanceof ItemFrameBlock){
+					$pk = new ItemFrameDropItemPacket();
+					$pk->x = (int) $target->x;
+					$pk->y = (int) $target->y;
+					$pk->z = (int) $target->z;
+					$this->handleItemFrameDropItem($pk);
```
dktapps, line 2950: This block should be placed higher up, since the breaking animation shouldn't be broadcast for removing items from an item frame. I'm also not sure whether extinguishing fire takes precedence over removing an item from a frame or not.

dktapps (changes requested):
This will fire PlayerInteractEvent::LEFT_CLICK_BLOCK twice, since the ItemFrameDropItemPacket handler also fires it. You should separate the item-frame-specific code from the ItemFrameDropItemPacket into a private function used by both handlers.

HighTecDev:
Closed accidentally. I’ll open it up and take care of it next week, provided you don’t want to take the other fix-PR

dktapps:
The other one was closer to mergeable last time I looked at it, but I haven't reviewed it since.

dktapps:
This was superseded by #4101.

PR #4086, opened 2021-03-18
Fixed SingletonTrait not being recognized after PhpStorm 2.3 version

alvin0319:
## Introduction

Since `PhpStorm 2.3` and later, there is a bug where functions or variables in a class are marked as undefined when using the `getInstance()` method of `SingletonTrait`.

image

After this commit, `SingletonTrait` works fine even with `PhpStorm 2.3` and above.

## Tests

image


dktapps:
PhpStorm is broken. It's not PM's problem. Duplicate of #3981 .

PR #4100, opened 2021-03-22
get rid of the useless do

UnknownOre:


UnknownOre:
i should stop and go to sleep.

PR #4101, opened 2021-03-22
fix ItemFrames

UnknownOre:


dktapps:
This looks more or less OK, minus the formatting (tabs should be used for indentation.

src/pocketmine/Player.php
```diff
@@ -2924,6 +2924,18 @@ public function handlePlayerAction(PlayerActionPacket $packet) : bool{
 					$this->inventory->sendHeldItem($this);
 					break;
 				}
+				
+				if($target->getId() == Block::ITEM_FRAME_BLOCK){
```
dktapps, line 2928: I think this check can be removed.

src/pocketmine/Player.php
```diff
@@ -2924,6 +2924,18 @@ public function handlePlayerAction(PlayerActionPacket $packet) : bool{
...
+					$tile = $this->level->getTile($pos);
+					if($tile instanceof ItemFrame and $tile->getItem()->getId() !== Item::AIR) {
```
dktapps, line 2930:
```suggestion
					if($tile instanceof ItemFrame and $tile->getItem()->getId() !== Item::AIR){
```

src/pocketmine/Player.php
```diff
@@ -2924,6 +2924,18 @@ public function handlePlayerAction(PlayerActionPacket $packet) : bool{
...
+						if (lcg_value() <= $tile->getItemDropChance()) {
```
dktapps, line 2931:
```suggestion
						if (lcg_value() <= $tile->getItemDropChance()){
```

dktapps (approved):

PR #4127, opened 2021-03-27
Update RFC instructions

SOF3:
since we have migrated RFCs to Discussions

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -114,13 +114,11 @@ class ExampleClass{
 ```
 
 <!-- TODO: RFC and voting on the forums instead -->
-### RFC and Voting
-* These are big Pull Requests or contributions that change important behavior.
-* RFCs will be tagged with the *Type: Request For Comments* label
-* A vote will be held once the RFC is ready. All users can vote commenting on the Pull Request
-* Comments MUST use "Yes" or "No" on the FIRST sentence to signify the vote, except when they don't want it to be counted.
+### Major API changes
+* Major API changes, no matter API-breaking or not, SHOULD be proposed by creating a topic in [Discussions](https://github.com/pmmp/PocketMine-MP/discussions).
+* The change proposal enters the "Voting Period" of 7 days after the team determines that sufficient discussion has been made.
+* During the Voting Period, comments on the proposal with "YES" or "NO" in the *first* sentence indicate the user's vote.
 * If your comment is a voting comment, specify the reason of your vote or it won't be counted.
-* After voting has been closed, no further votes will be counted.
-* An RFC will be rejected if less than 50% + 1 (simple majority) has voted Yes.
-* If the RFC is approved, Team Members have the final word on its implementation or rejection.
+* The team will *consider* the voting results after the Voting Period and decide whether the discussion can be implemented.
+* Contributors may implement the proposed change *after* the Voting Period.
 * RFCs with complex voting options will specify the vote percentage or other details.


dktapps:
Do we ever actually follow a voting phase?

SOF3:
No. I doubt we ever even passed any RFC at all.

CONTRIBUTING.md
```diff
@@ -114,13 +114,11 @@ class ExampleClass{
...
 <!-- TODO: RFC and voting on the forums instead -->
-### RFC and Voting
-* These are big Pull Requests or contributions that change important behavior.
-* RFCs will be tagged with the *Type: Request For Comments* label
-* A vote will be held once the RFC is ready. All users can vote commenting on the Pull Request
-* Comments MUST use "Yes" or "No" on the FIRST sentence to signify the vote, except when they don't want it to be counted.
+### Major API changes
+* Major API changes, no matter API-breaking or not, SHOULD be proposed by creating a topic in [Discussions](https://github.com/pmmp/PocketMine-MP/discussions).
+* The change proposal enters the "Voting Period" of 7 days after the team determines that sufficient discussion has been made.
+* During the Voting Period, comments on the proposal with "YES" or "NO" in the *first* sentence indicate the user's vote.
```
dktapps, line 120: I think this can be replaced by discussion votes and/or reactions.

CONTRIBUTING.md
```diff
@@ -114,13 +114,11 @@ class ExampleClass{
...
 * If your comment is a voting comment, specify the reason of your vote or it won't be counted.
-* After voting has been closed, no further votes will be counted.
-* An RFC will be rejected if less than 50% + 1 (simple majority) has voted Yes.
-* If the RFC is approved, Team Members have the final word on its implementation or rejection.
+* The team will *consider* the voting results after the Voting Period and decide whether the discussion can be implemented.
+* Contributors may implement the proposed change *after* the Voting Period.
```
dktapps, line 123: I think this should be rephrased to something like "the change may be implemented after the voting period passes" so that we don't suggest that only specific people can implement an RFC.

CONTRIBUTING.md
```diff
@@ -114,13 +114,11 @@ class ExampleClass{
...
 
 <!-- TODO: RFC and voting on the forums instead -->
-### RFC and Voting
-* These are big Pull Requests or contributions that change important behavior.
-* RFCs will be tagged with the *Type: Request For Comments* label
-* A vote will be held once the RFC is ready. All users can vote commenting on the Pull Request
-* Comments MUST use "Yes" or "No" on the FIRST sentence to signify the vote, except when they don't want it to be counted.
+### Major API changes
+* Major API changes, no matter API-breaking or not, SHOULD be proposed by creating a topic in [Discussions](https://github.com/pmmp/PocketMine-MP/discussions).
+* The change proposal enters the "Voting Period" of 7 days after the team determines that sufficient discussion has been made.
```
dktapps, line 119: is 7 days really long enough? Most popular change proposals and enhancements don't even get 20 votes after being open for a month.

SOF3: Or do we just dismiss the voting period and just make 7 days the "final comment period"?

dktapps: I don't think 7 days is in any way practical.

SOF3: What I mean is that there shouldn't be votes at all, and the last 7 days are just a way to notify people that the RFC is finalizing soon. They could always have commented in the months before the finalization period, and 7 days is just a reminder.
Otherwise, just don't have a reminder and finalize the RFC straight away when the time comes.

dktapps:
Closing due to lack of activity. I'm also not sold on discussions (they kinda suck to be honest).

Issue #4130, opened 2021-03-29
call InventoryCloseEvent whenever Player->removeWindow() is called


UnknownOre:
## Description
call InventoryCloseEvent when a window is removed and don't depend on the client packet

## Justification
a the time being InventoryCloseEvent only gets called when the client send ContainerClosePacket which might break some plugins logic depending on InventoryCloseEvent

dktapps:
That's only because the client sends ContainerClose back to the server. That shouldn't be relied on.

Dapro718:
It works with fake inventories being sent as well.

PR #4133, opened 2021-03-31
Call InventoryCloseEvent when Player->removeWindow() is called. (#4130)

Dapro718:
## Introduction
When the method Player->removeWindow() is called, InventoryCloseEvent is not called as explained in issue #4130. Technically, it should be because the player is still closing the inventory even thought its being closed by the server/plugin.

### Relevant issues
There is no way to track if an inventory is closed with Player->removeWindow(), so plugins like ``InvMenu`` break. InvMenu depends on InventoryCloseEvent to remove players from a window session. Without this event being called, players become stuck in this session and cannot view any other windows.
[Example](https://user-images.githubusercontent.com/60823817/111855257-33f7a000-88fa-11eb-8dc8-99aad349d61c.mp4) (short video)

### API changes
Yes, there are additional changes when method Player->removeWindow() is called: InventoryCloseEvent gets called.

### Behavioural changes
none

## Backwards compatibility
Some developers have caught onto this, and have been calling the event themselves. They will need to remove this call to prevent the event from triggering twice.

## Follow-up
n/a

## Tests
Installed Inventory Menu and ran the same script used in the video above. After being send the first window and having it removed, the player could receive more windows.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -4140,6 +4140,10 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 
 		if($id !== null){
 			$inventory->close($this);
+			
+			$ev = new InventoryCloseEvent($inventory, $this);
+			$ev->call();
+			
 			unset($this->windows[$hash], $this->windowIndex[$id], $this->permanentWindows[$id]);
 		}
 	}


src/pocketmine/Player.php
```diff
@@ -4140,6 +4140,10 @@ public function removeWindow(Inventory $inventory, bool $force = false){
 
 		if($id !== null){
 			$inventory->close($this);
+			
+			$ev = new InventoryCloseEvent($inventory, $this);
+			$ev->call();
```
T14Raptor, line 4145: This will cause InventoryCloseEvent to get called twice if a player closes a container.

dktapps:
Why is this closed?

Issue #4134, opened 2021-03-31
Watchdog Timer (Force shutdown when freeze)


AkmalFairuz:
## Description
Force shutdown server when main thread freeze within a certain time.
When the server freeze type /stop on console and CTRL+C not working, must kill PocketMine process id manually.

https://en.wikipedia.org/wiki/Watchdog_timer

## Justification
- Make server keep online.

dktapps:
On Linux, you can use ctrl+z and then `kill -9 $(cat server.lock)`. On Windows you can just close the terminal window or kill it from task manager.

dktapps:
I'm going to close this, because the way you did it in your PR can easily be implemented as a plugin.

This is addressing the symptoms of a problem and not its cause. Most of the time, this will be actively harmful, because it would get in the way of debugging (the server might kill itself before anyone got a chance to attach a debugger, denying any possibility of identifying the cause of the freeze). In addition, a freeze might occur for other reasons, such as the server running out of memory and thrashing swap, something else hogging the CPU, etc; in which cases a watchdog would be useless.

PR #4135, opened 2021-03-31
Kill server when freeze within a certain time.

AkmalFairuz:
## Introduction

Kill server when main thread freeze within a certain time.

### Relevant issues
* Fixes https://github.com/pmmp/PocketMine-MP/issues/4134

## Tests

sleep main thread for 10000 seconds.
Screenshot (504)


Mohagames205:
What about plugins that manipulate the world and take a long time to respond? Will there be an option to disable the watchdog?

AkmalFairuz:
> In the current code, the lastRespond variable is not updated, so the server will be killed unconditionally after 180 seconds even if the server is not frozen.
> 
> open image

Fixed, i hate php-cs-fixer.

dktapps:
You should really follow the proper developer flow with git to contribute. Editing on GitHub is just making your job harder and more messy.

PR #4139, opened 2021-04-01
instead of kicking player for toggling flight cancel the event

UnknownOre:
## Introduction
a lot of people have complained before that they get kicked for flight while the movement cheats is enabled

### Behavioural changes
now it will cancel the event instead of kicking the player


dktapps:
This kick should probably be removed, since it can false-positive fire in case of player toggling flight at the same time as their gamemode is being changed server-side. It would be better to have PlayerToggleFlightEvent set to cancelled before calling it.

dktapps:
Build failures are unrelated.

PR #4141, opened 2021-04-02
small inventory bug fix

### LINKED ISSUE #4140: problem in inventory while being inside another entity or player
### Issue description

- Expected result: The inventory action being executed normally
- Actual result: The action gets reverted
1. hold an item
2. get inside another entity
3. open your inventory and try moving the item you are holding

### OS and versions

* PocketMine-MP: 3.18
* PHP: 7.4
* Server OS: Windows 10

UnknownOre:
## Introduction

#4140 

### Behavioural changes

now InteractPacket will only run doCloseInventory if the player doesn't have his inventory open or the action isn't InteractPacket::ACTION_MOUSEOVER

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -2843,9 +2843,9 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 		}
 
 		$target = $this->level->getEntity($packet->target);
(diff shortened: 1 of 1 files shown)


dktapps:
I opted for a simpler fix of my own, since there's already a workaround in place for the same bug in a more specific case. Thank you for the PR.

PR #4142, opened 2021-04-02
call InventoryCloseEvent on removeWindow()

### LINKED ISSUE #4130: call InventoryCloseEvent whenever Player->removeWindow() is called

UnknownOre:
## Introduction

as explained in #4130 the server depend on ContainerClosePacket to call InventoryCloseEvent, this is problematic cause it will break plugins depending on that event if the client didn't send ContainerClosePacket.

* Fixes #4130 

-->

### Behavioural changes
now InventoryCloseEvent will be called whenever the inventory is closed from the server side

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -4134,15 +4134,12 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 	public function removeWindow(Inventory $inventory, bool $force = false){
 		$id = $this->windows[$hash = spl_object_hash($inventory)] ?? null;
 
-		if($id !== null){
-			(new InventoryCloseEvent($inventory, $this))->call();
-		}
-
 		if($id !== null and !$force and isset($this->permanentWindows[$id])){
 			throw new \InvalidArgumentException("Cannot remove fixed window $id (" . get_class($inventory) . ") from " . $this->getName());
 		}
 
 		if($id !== null){
+			(new InventoryCloseEvent($inventory, $this))->call();
 			$inventory->close($this);
 			unset($this->windows[$hash], $this->windowIndex[$id], $this->permanentWindows[$id]);
 		}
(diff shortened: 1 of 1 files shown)


src/pocketmine/Player.php
```diff
@@ -4135,6 +4134,8 @@ public function addWindow(Inventory $inventory, int $forceId = null, bool $isPer
 	public function removeWindow(Inventory $inventory, bool $force = false){
 		$id = $this->windows[$hash = spl_object_hash($inventory)] ?? null;
 
+		(new InventoryCloseEvent($inventory, $this))->call();
```
dktapps, line 4137: This is going to fire when using removeWindow() with an inventory that wasn't opened to begin with.

dktapps (changes requested):
I think this is OK, minus one or two things.

My verdict on permanent windows is that since InventoryOpenEvent is already fired for these (though perhaps not intentionally), it's probably wise to have it fire for their closing too.

However, on PM4, the behaviour of both these events changes. InventoryOpen is only fired for non-permanent inventories, and InventoryClose is the same. So, I'm not 100% sure if changing InventoryClose behaviour at this stage of PM3 lifetime is a good idea.

src/pocketmine/Player.php
```diff
@@ -4135,6 +4134,10 @@ public function addWindow(Inventory $inventory, int $forceId = null, bool $isPer
 	public function removeWindow(Inventory $inventory, bool $force = false){
 		$id = $this->windows[$hash = spl_object_hash($inventory)] ?? null;
 
+		if($id !== null){
+			(new InventoryCloseEvent($inventory, $this))->call();
+		}
+
```
dktapps, line 4140: I think it's sufficient to just move it inside the bottom if statement.

PR #4170, opened 2021-04-13
Fixed bug that never lets mountainous terrain generate

TheShermanTanker:
## Introduction

As of now the BiomeSelector class has a bug that prevents mountains from generating in the world, left unfixed for several concerns about a possible fix being inaccurate. This pull request attempts to create a decent fix for this issue

This PR message is still currently a WIP 

### Relevant issues

No relevant issues

### API changes

None
### Behavioural changes

Mountains and Small Mountains now generate in the world naturally
## Backwards compatibility

No backwards incompatibility, this merely changes the terrain generator, which is entirely server side and currently untouchable by plugins
## Follow-up

Unlikely to require any follow up after merge

## Tests

Generated mountains:
image

### DIFF
--- a/src/pocketmine/level/generator/normal/Normal.php
+++ b/src/pocketmine/level/generator/normal/Normal.php
@@ -147,7 +147,7 @@ class Normal extends Generator{
 					//According to the wiki, mountains have a temperature of 0.2-0.3,
 					//so substituting rainfall for temperature was actually pretty accurate
 					//These are all the best estimates for mountain temperature
-				    if($temperature < 0.20){
+					if($temperature < 0.20){
 						return Biome::MOUNTAINS;
 					}elseif($temperature < 0.40){
 						return Biome::SMALL_MOUNTAINS;
--- a/src/pocketmine/level/generator/normal/Normal.php
+++ b/src/pocketmine/level/generator/normal/Normal.php
@@ -144,12 +144,12 @@ class Normal extends Generator{
 						return Biome::BIRCH_FOREST;
 					}
 				}else{
-					//FIXME: This will always cause River to be used since the rainfall is always greater than 0.8 if we
-					//reached this branch. However I don't think that substituting temperature for rainfall is correct given
-					//that mountain biomes are supposed to be pretty cold.
-					if($rainfall < 0.25){
+					//According to the wiki, mountains have a temperature of 0.2-0.3,
+					//so substituting rainfall for temperature was actually pretty accurate
+					//These are all the best estimates for mountain temperature
+				    if($temperature < 0.20){
 						return Biome::MOUNTAINS;
-					}elseif($rainfall < 0.70){
+					}elseif($temperature < 0.40){
 						return Biome::SMALL_MOUNTAINS;
 					}else{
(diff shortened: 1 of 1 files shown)


src/pocketmine/level/generator/normal/Normal.php
```diff
@@ -144,12 +144,12 @@ protected function lookup(float $temperature, float $rainfall) : int{
 						return Biome::BIRCH_FOREST;
 					}
 				}else{
-					//FIXME: This will always cause River to be used since the rainfall is always greater than 0.8 if we
-					//reached this branch. However I don't think that substituting temperature for rainfall is correct given
-					//that mountain biomes are supposed to be pretty cold.
-					if($rainfall < 0.25){
+					//According to the wiki, mountains have a temperature of 0.2-0.3,
+					//so substituting rainfall for temperature was actually pretty accurate
+					//These are all the best estimates for mountain temperature
```
dktapps, line 149: This comment is going to make no sense in retrospect, because nobody will know that this was a problem after the bug is fixed.

dktapps:
Interesting, in your screenshot I see there's a river that actually looks like a river instead of a giant blob. Is this caused by the PR changes or just a coincidence?

TheShermanTanker:
> 
> 
> Interesting, in your screenshot I see there's a river that actually looks like a river instead of a giant blob. Is this caused by the PR changes or just a coincidence?

I believe this is actually because rivers were originally intended to be generated near or next to mountains, and is definitely caused by the changes, if the world was generated without the PR, all the mountains you see in the picture would also be a part of the river in the screenshot; IE a giant blob, just like you mentioned, but now instead of generating more river, mountains generate instead, hence why it now actually looks like a proper one

dktapps:
I guess the river is now a big ring instead of an oval.

src/pocketmine/level/generator/normal/Normal.php
TheShermanTanker: Should I delete it then?

dktapps: That would seem to be the obvious action ..

TheShermanTanker:
> 
> 
> I guess the river is now a big ring instead of a blob.

It doesn't actually stretch around in a ring, it is really just a strip of river next to the mountain, unfortunately I didn't have the hindsight to take a screenshot from an angle that can show this

src/pocketmine/level/generator/normal/Normal.php
TheShermanTanker: Ight ignore my 2IQ question I'll remove it

thebigsmileXD:
Just a quick comment about temperatures:
You can find a .json file with biome definitions and temperatures in the game's files, this may lead to a more accurate result than trusting the wiki for facts

TheShermanTanker:
> 
> 
> Just a quick comment about temperatures:
> You can find a .json file with biome definitions and temperatures in the game's files, this may lead to a more accurate result than trusting the wiki for facts

Got it, will refer to the json if I do a PR in the future

PR #4171, opened 2021-04-13
API method to clear block cache per chunk

VixikHD:
## Introduction

Plugins whose use SubChunkIteratorManager for filling needs to send chunks to player afterwards.  It is also needed to clear chunk caches to avoid sending outdated blocks. Currently there is method to clear `Level->chunkCache`, whilst there is no way to clear `Level->blockCache` (per chunk). Although there is method `Level->clearCache(true)`, developers could not use it to clear cache in certain chunks, whose were changed.

### Relevant issues
- Nothing

### API changes
- Added method `Level->clearBlockCache(int $chunkX, int $chunkZ)`

### Behavioural changes
- Nothing

## Backwards compatibility
- Any backwards incompatible changes

## Follow-up
- Nothing

## Tests

- Tested with [BuilderTools plugin](https://github.com/CzechPMDevs/BuilderTools/tree/feature/pm-fork)

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -1047,6 +1047,15 @@ class Level implements ChunkManager, Metadatable{
 	/**
 	 * @return void
 	 */
+	public function clearBlockCache(int $chunkX, int $chunkZ){
+		if(isset($this->blockCache[$hash = Level::chunkHash($chunkX, $chunkZ)])) {
+			unset($this->blockCache[$hash]);
+		}
+	}
+
+	/**
+	 * @return void
+	 */
 	public function clearChunkCache(int $chunkX, int $chunkZ){
 		unset($this->chunkCache[Level::chunkHash($chunkX, $chunkZ)]);
 	}


src/pocketmine/level/Level.php
```diff
@@ -1044,6 +1044,15 @@ public function clearCache(bool $force = false){
 		}
 	}
 
+	/**
+	 * @return void
+	 */
+	public function clearBlockCache(int $chunkX, int $chunkZ){
+		if(isset($this->blockCache[$hash = Level::chunkHash($chunkX, $chunkZ)])) {
```
dktapps, line 1051: This check is useless.

src/pocketmine/level/Level.php
```diff
@@ -1044,6 +1044,15 @@ public function clearCache(bool $force = false){
 		}
 	}
 
+	/**
+	 * @return void
+	 */
+	public function clearBlockCache(int $chunkX, int $chunkZ){
```
dktapps, line 1050: Native `void` typehint should be used here. The only reason it's doc-commented elsewhere is for backwards compatibility. New API methods should use native types.

dktapps (changes requested):
This should target `master`.

PR #4172, opened 2021-04-14
API method for clearing block cache per chunk (#4171 for master)

DaisukeDaisuke:
## Introduction

This is a pull request that rewrites #4171 for the master.
Please see #4171.

> Plugins whose use SubChunkIteratorManager for filling needs to send chunks to player afterwards. It is also needed to clear chunk caches to avoid sending outdated blocks. Currently there is method to clear Level->chunkCache, whilst there is no way to clear Level->blockCache (per chunk). Although there is method Level->clearCache(true), developers could not use it to clear cache in certain chunks, whose were changed.
### Relevant issues

> Nothing

### API changes

> Added method Level->clearBlockCache(int $chunkX, int $chunkZ)
### Behavioural changes

> Nothing
## Backwards compatibility

> Any backwards incompatible changes
## Follow-up

> Nothing

## Tests

I ran the following code and confirmed that "worked" was displayed normally.
```php
<?php

namespace test;

use pocketmine\player\Player;

class main extends PluginBase{
	public function onCommand(CommandSender $sender, Command $command, string $label, array $args): bool{
		if(!$sender instanceof Player){
			return true;
		}
		if($label === "test"){
			$pos = $sender->getPosition();
			$sender->getWorld()->clearBlockCache($pos->x >> 4, $pos->z >> 4);
			$sender->sendMessage("worked");
		}
		return true;
	}
}
```

### DIFF
--- a/src/world/World.php
+++ b/src/world/World.php
@@ -948,7 +948,13 @@ class World implements ChunkManager{
 	}
 
 	public function clearBlockCache(int $chunkX, int $chunkZ) : void{
+		$chunk = $this->getChunk($chunkX, $chunkZ);
 		unset($this->blockCache[World::chunkHash($chunkX, $chunkZ)]);
+		if($chunk !== null){
+			foreach($this->getChunkPlayers($chunkX, $chunkZ) as $p){
+				$p->onChunkChanged($chunkX, $chunkZ, $chunk);
+			}
+		}
 	}
 
 	/**
--- a/src/world/World.php
+++ b/src/world/World.php
@@ -947,6 +947,10 @@ class World implements ChunkManager{
 		}
 	}
 
+	public function clearBlockCache(int $chunkX, int $chunkZ) : void{
+		unset($this->blockCache[World::chunkHash($chunkX, $chunkZ)]);
+	}
+
 	/**
 	 * @return bool[] fullID => bool
 	 */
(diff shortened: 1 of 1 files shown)


dktapps:
Technically speaking it's possible to just use World->setChunk() to destroy the cache for a chunk.

dktapps:
In addition, clearing block cache is not sufficient if a chunk was modified directly. Chunk packet caches should also be cleared.

src/world/World.php
```diff
@@ -947,6 +947,16 @@ public function clearCache(bool $force = false) : void{
 		}
 	}
 
+	public function clearBlockCache(int $chunkX, int $chunkZ) : void{
+		$chunk = $this->getChunk($chunkX, $chunkZ);
+		unset($this->blockCache[World::chunkHash($chunkX, $chunkZ)]);
+		if($chunk !== null){
+			foreach($this->getChunkPlayers($chunkX, $chunkZ) as $p){
+				$p->onChunkChanged($chunkX, $chunkZ, $chunk);
+			}
```
DaisukeDaisuke, line 956: Since I don't know how to clear the cache of ChunkCache.php directly, I decided to call the onChunkChanged function of the player based on the surrounding code. (If my code reading comprehension is correct).

VixikHD: This is already possible to do with plugin api. Previously I wanted add that method just for to avoid doing stuff like [this](https://github.com/CzechPMDevs/BuilderTools/blob/default/src/czechpmdevs/buildertools/editors/object/FillSession.php#L187-L216). Method to clear chunk cache is removed in 4.0.0, that was also reason why I wanted make it for 3.x

dktapps: It's much easier to just use setChunk() instead of that big mess ...

dktapps:
I'm going to close this, since `World->setChunk()` is already the better option for this (there are various other things that need to be cleaned up, and you'd end up just duplicating most of the function anyway).

PR #4176, opened 2021-04-16
I suggest adding the ability to remove the notification that the play…

celishere:
This will allow in some cases for developers to remove spam in the console when using this feature.


SOF3:
The `close` function is an internal function if I remember correctly. It doesn't make sense to add a plugin-only parameter to an internal function.

dktapps:
For auditability this is a bad idea. I doubt any live server would actually use this.

SOF3:
What we really need is to separate logs into a higher variety of channels so that they are easier to filter.

PR #4178, opened 2021-04-16
Possibilities to disable xp orb collecting

HimmelKreis4865:
## Introduction
Any human (except for Players that are in spectator gamemode) collects xp orbs, which can't be disabled.
So there's no possibility to turn it off in case it's a custom human that basically represents a NPC or something like that

### Relevant issues
Any human collects XP orbs
* This can be turned off via code

-->

### API changes
`Human::canCollectXp(): bool`

### Behavioural changes
-

## Backwards compatibility
-

## Follow-up
-

Requires translations:
-

-->

## Tests
Tested, XP orbs are not trying to get collected by the human with disabled collecting, they just lay where they are, players can still collect the orbs.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -4207,7 +4207,7 @@ class Player extends Human implements CommandSender, ChunkLoader, IPlayer{
 
 	}
 	
-	public function canCollectXp(): bool {
+	public function canCollectXp() : bool{
 		return !$this->isSpectator();
 	}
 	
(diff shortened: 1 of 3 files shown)


dktapps:
This should be targeted at `master`. `stable` is in feature freeze.

Otherwise, concept is OK (formatting needs some work though).

HimmelKreis4865:
oof - I went brainafk, I'm going to recode this for api4 and reopen an issue soon.

PR #4180, opened 2021-04-17
Fix dragon egg teleporting in creative

### LINKED ISSUE #4179: Dragon Egg is not destroyed
### Issue description

- Expected result: What were you expecting to happen? Dragon Egg is destroyed
- Actual result: What actually happened? Dragon Egg is not destroyed and teleported somewhere

Vanilla behavior

https://user-images.githubusercontent.com/32565818/115112786-26e5d380-9fc2-11eb-94e1-76fe58269c62.mp4

PMMP behavior

https://user-images.githubusercontent.com/32565818/115112807-32d19580-9fc2-11eb-96dc-ddd7e057ea50.mp4

1. Place Dragon Egg
2. Try to destroy Dragon Egg

### OS and versions

* PocketMine-MP: a32eb4ebc3e82d52cd44d4dc03dcd59a965ed75c
* PHP: 7.4
* Server OS: Windows
* Game version: PE/Win10

### Plugins

Mohagames205:
## Introduction
This PR aims to solve https://github.com/pmmp/PocketMine-MP/issues/4179
Dragon eggs should be broken if the player is in gamemode c

### Relevant issues
* Fixes #4179 

### API changes
None.

### Behavioural changes
Dragon eggs don't teleport when they're attacked when the player is in creative

## Backwards compatibility
All the changes are BC

## Follow-up
Nope

## Tests
https://streamable.com/c5kj1v (note that I am switching between right and left click in the video)

### DIFF
--- a/src/block/DragonEgg.php
+++ b/src/block/DragonEgg.php
@@ -60,8 +60,9 @@ class DragonEgg extends Transparent implements Fallable{
 	public function onAttack(Item $item, int $face, ?Player $player = null) : bool{
 		if($player !== null && !$player->getGamemode()->equals(GameMode::CREATIVE())){
 			$this->teleport();
+			return true;
 		}
-		return true;
+		return false;
 	}
 
 	public function teleport() : void{
--- a/src/block/DragonEgg.php
+++ b/src/block/DragonEgg.php
@@ -29,6 +29,7 @@ use pocketmine\event\block\BlockTeleportEvent;
 use pocketmine\item\Item;
 use pocketmine\item\ToolTier;
 use pocketmine\math\Vector3;
+use pocketmine\player\GameMode;
 use pocketmine\player\Player;
 use pocketmine\world\particle\DragonEggTeleportParticle;
 use pocketmine\world\World;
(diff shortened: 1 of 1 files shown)


dktapps:
There are some problems flagged by phpstan, please address them.

src/block/DragonEgg.php
```diff
@@ -57,7 +58,9 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
 	}
 
 	public function onAttack(Item $item, int $face, ?Player $player = null) : bool{
-		$this->teleport();
+		if($player !== null && $player->getGamemode() != GameMode::CREATIVE()){
```
dktapps, line 61:
```suggestion
		if($player !== null && !$player->getGamemode()->equals(GameMode::CREATIVE())){
```

Mohagames205: Is there a reason why !== is not reliable here?

src/block/DragonEgg.php
```diff
@@ -57,7 +58,9 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
 	}
 
 	public function onAttack(Item $item, int $face, ?Player $player = null) : bool{
-		$this->teleport();
+		if($player !== null && !$player->getGamemode()->equals(GameMode::CREATIVE())){
+			$this->teleport();
+		}
 		return true;
```
dktapps, line 64: This should return false if nothing was done, otherwise it may have side effects.

src/block/DragonEgg.php
dktapps: Threads make things complicated. The enum members are able to be cloned and serialized.

PR #4182, opened 2021-04-17
Fixed action stopping when needing to build PHP

fuyutsuki:
## Introduction

Fixed action stopping when needing to build PHP with insufficient permissions.

### Relevant issues

none

## Changes

- You will be able to build PHP


dktapps:
it never has insufficient permissions because it's set in the file attributes ...

PR #4184, opened 2021-04-18
Remove "ench" tag when all enchantments are removed from an item (Fixes #4144)

### LINKED ISSUE #4144: Item->removeEnchantment() does not remove the "ench" NBT tag if there are no enchantments left
### Issue description

- Expected result: The ``ench`` NBT tag would be removed from the item, removing the enchanted texture
- Actual result: The tag remained and the item had the enchanted texture without having any enchantments

1. Enchant an item
2. Remove the enchantment from the item

### OS and versions

* PocketMine-MP: 3.18.1
* PHP: 7.4.16
* Server OS: Windows
* Game version: PE/Win10

Code to reproduce:
```php
<?php
declare(strict_types=1);

namespace test;

use pocketmine\item\enchantment\Enchantment;
use pocketmine\item\enchantment\EnchantmentInstance;
use pocketmine\item\Item;

class Test extends PluginBase implements Listener {

  public function onEnable(): void {
  }

  public function onPlayerJoin(PlayerJoinEvent $event): void {
    $item = Item::get(Item::DIAMOND_SWORD);
    $item->addEnchantment(new EnchantmentInstance(Enchantment::getEnchantment(Enchantment::SHARPNESS)));
    $item->removeEnchantment(Enchantment::SHARPNESS);
    $event->getPlayer()->getInventory()->addItem($item);
  }
}
```

TheNewHEROBRINEX:
## Introduction
As reported in #4144, Item->removeEnchantment() doesn't remove the "ench" tag when there are no more enchantments left on the item.
This PR fixes this issue by checking if, after the enchantment has been removed, there are no other ones left and acting accordingly.

### Relevant issues

* Fixes #4144 

### API changes

None

### Behavioural changes

Now the shining effect won't be shown on items which don't have any enchantments but previously did.

## Backwards compatibility

Yes

## Follow-up

None

## Tests

I added a phpunit test to verify that the "ench" tag is being correctly removed when there is the need to do so.
I also confirmed that the reproducing code provided in the issue does not reproduce anymore.

### DIFF
--- a/tests/phpunit/item/ItemTest.php
+++ b/tests/phpunit/item/ItemTest.php
@@ -72,7 +72,7 @@ class ItemTest extends TestCase{
 		$item = ItemFactory::get(Item::DIAMOND_SWORD);
 		$item->addEnchantment(new EnchantmentInstance(Enchantment::getEnchantment(Enchantment::SHARPNESS)));
 		$item->removeEnchantment(Enchantment::SHARPNESS);
-		self::assertFalse($item->getNamedTag()->hasTag(Item::TAG_ENCH, ListTag::class));
+		self::assertNull($item->getNamedTag()->getTag(Item::TAG_ENCH));
 	}
 
 	/**
--- a/tests/phpunit/item/ItemTest.php
+++ b/tests/phpunit/item/ItemTest.php
@@ -25,12 +25,16 @@ namespace pocketmine\item;
 
 use PHPUnit\Framework\TestCase;
 use pocketmine\block\BlockFactory;
+use pocketmine\item\enchantment\Enchantment;
+use pocketmine\item\enchantment\EnchantmentInstance;
+use pocketmine\nbt\tag\ListTag;
 
 class ItemTest extends TestCase{
 
 	public function setUp() : void{
 		BlockFactory::init();
 		ItemFactory::init();
+		Enchantment::init();
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


tests/phpunit/item/ItemTest.php
```diff
@@ -61,6 +65,19 @@ public function testItemBlockRegistered() : void{
...
+	 * Tests that when all enchantments are removed from an item, the "ench" tag is removed as well
+	 */
+	public function testEnchantmentRemoval() : void{
+		$item = ItemFactory::get(Item::DIAMOND_SWORD);
+		$item->addEnchantment(new EnchantmentInstance(Enchantment::getEnchantment(Enchantment::SHARPNESS)));
+		$item->addEnchantment(new EnchantmentInstance(Enchantment::getEnchantment(Enchantment::UNBREAKING)));
+		$item->removeEnchantment(Enchantment::SHARPNESS);
+		self::assertTrue($item->getNamedTag()->hasTag(Item::TAG_ENCH, ListTag::class));
+		$item->removeEnchantment(Enchantment::UNBREAKING);
+		self::assertFalse($item->getNamedTag()->hasTag(Item::TAG_ENCH, ListTag::class));
```
dktapps, line 78: This test can be simplified. It only needs to add and remove 1 enchantment.

TheNewHEROBRINEX: My intention was to test both that the ench is not removed when there are still other enchantments and that it is removed when there are no more enchantments.

dktapps: You don't need to test that. Stick to testing the thing you're trying to fix.

tests/phpunit/item/ItemTest.php
```diff
@@ -61,6 +65,16 @@ public function testItemBlockRegistered() : void{
...
 	}
 
+	/**
+	 * Tests that when all enchantments are removed from an item, the "ench" tag is removed as well
+	 */
+	public function testEnchantmentRemoval() : void{
+		$item = ItemFactory::get(Item::DIAMOND_SWORD);
+		$item->addEnchantment(new EnchantmentInstance(Enchantment::getEnchantment(Enchantment::SHARPNESS)));
+		$item->removeEnchantment(Enchantment::SHARPNESS);
+		self::assertFalse($item->getNamedTag()->hasTag(Item::TAG_ENCH, ListTag::class));
```
dktapps, line 75:
```suggestion
		self::assertNull($item->getNamedTag()->getTag(Item::TAG_ENCH));
```
PM4's NBT doesn't have `hasTag()`, and this version is better suited to both versions.

PR #4185, opened 2021-04-18
Fix inventories not opening after having inventory open and dying

### LINKED ISSUE #4177: Bug with inventories after death
### Issue description

- Expected result: Inventories can be opened after death
- Actual result: Inventories cannot be opened after death if...

If at the time of death a player has an open inventory, he will no longer be able to open inventories until he reconnects to the server

### OS and versions
* PocketMine-MP: 4.0.0
* PHP: 7.4
* Server OS: Ubuntu 20.04
* Game version: PE/Win10

### Plugins
NO

Mohagames205:
## Introduction
This PR aims to solve inventories not opening when you die with an open inventory. 

### Relevant issues
* Fixes #4177

### API changes
None

### Behavioural changes
Inventories now open even when you die with an open inventory

## Backwards compatibility
Nope

## Follow-up
I don't know if these changes are sufficient. During my tests, everything worked fine, but there may be some edge cases?

## Tests
https://streamable.com/7t0s2k


src/network/mcpe/handler/DeathPacketHandler.php
```diff
@@ -57,6 +58,11 @@ public function handlePlayerAction(PlayerActionPacket $packet) : bool{
 		return false;
 	}
 
+	public function handleContainerClose(ContainerClosePacket $packet): bool
```
TheNewHEROBRINEX, line 61: In accord with PocketMine's code formatting, here you should put a space between ")" and ":" and put the "{" on the same line as the function declaration.

TheNewHEROBRINEX (changes requested):

src/network/mcpe/handler/DeathPacketHandler.php
Mohagames205: Fixed! thanks for letting me know

dktapps:
My gut feeling is that this problem breaks more than one packet will fix.

dktapps:
This problem arises because `openHardcodedWindows` is not cleaned up when the player dies. Normal windows are removed via `removeAllWindows()`, but because of workarounds I added for 1.16, the player's own inventory and crafting grid are not closed by the server.

dktapps:
Thanks for your contribution, but a more complete fix has now been introduced.

Mohagames205:
Oh cool, what was the actual fix? I'm curious as to how it should have been fixed.

dktapps:
Your fix was actually pretty close. But it needed to ensure that the window actually got closed and not just send a packet. Otherwise we could have ended up with it stuck in openHardcodedWindows and it wouldn't have worked correctly anyway.

PR #4186, opened 2021-04-18
Added CancelTaskException class

JavierLeon9966:
## Introduction

Adds CancelTaskException class which cancels the current task by throwing this exception.
### Relevant issues

None
### API changes

Added CancelTaskException class, by throwing this exception the current task gets canceled.
### Behavioural changes

None
## Backwards compatibility

None
## Follow-up

None
## Tests

711a05c


SOF3 (approved):
Looks good, with reasonable use case.

src/scheduler/CancelTaskException.php
```diff
@@ -0,0 +1,28 @@
...
+ * @link http://www.pocketmine.net/
+ *
+ *
+*/
+
+declare(strict_types=1);
+
+namespace pocketmine\scheduler;
+
+final class CancelTaskException extends \Exception{
```
dktapps, line 26: Please add a doc comment with a brief explanation of what this exception is used for.

tests/phpunit/scheduler/TaskSchedulerTest.php
```diff
@@ -0,0 +1,51 @@
...
+
+	public function testCancel() : void{
+		$task = $this->scheduler->scheduleTask(new CancelTask());
+		$cancelled = false;
+		try{
+			$task->run();
+		}catch(CancelTaskException $e){
+			$cancelled = true;
+		}
+		self::assertTrue($cancelled, "Task was not cancelled");
```
dktapps, line 49: You should call `this->scheduler->mainThreadHeartbeat()` in here instead, and then assert that `task->isCancelled()` is true.

dktapps (commented):
Looks good, minus one or two minor nits.

dktapps:
`TaskScheduler->run()` and `Task->onRun()` should have `@throws CancelTaskException` annotations added.

JavierLeon9966:
Ok, ready

PR #4188, opened 2021-04-19
Allow plugins to use PSR-4 namespace mapping

dktapps:
this is a reduced implementation of PSR-4 which serves the 99% use case without being horribly breakable.

Plugins may now specify a `src-namespace-prefix`, which should be set to the namespace of the classes in `src`.
If the old system is used, `src-namespace-prefix` can be omitted, or set to an empty string.

Examples:
- If `src-namespace-prefix` is `dktapps\test`, `dktapps\test\Main` will be searched for in `src/Main.php`, instead of `src/dktapps/test/Main.php`.

This implements the RFC described in #4115.
### API changes
- Added `PluginDescription->getSrcNamespacePrefix()`
- A newer version of `pocketmine/classloader` is needed, which breaks backwards compatibility (but probably not in any way that plugins will care about).
- `plugin.yml` now accepts a `src-namespace-prefix` directive, which influences where the autoloader will look for classes.

### Behavioural changes
Plugins are 
## Backwards compatibility
This is fully backwards compatible (minus DynamicClassLoader API changes).

## Follow-up
- Consider allowing `main` to be specified relative to `src-namespace-prefix`, so that the namespace doesn't need to be specified twice.

## Tests
This has been tested locally with a plugin locally, structured as follows:
```
test
    |___ plugin.yml
    |___ src
        |___ Main.php
```
`plugin.yml`:
```yml
---
name: test
version: 0.0.1
main: dktapps\test\Main
src-namespace-prefix: dktapps\test
api: 4.0.0
load: STARTUP
...
```
`Main.php`:
```php
<?php

declare(strict_types=1);

namespace dktapps\test;

		$this->getLogger()->info("IT WORKS");
	}
}
```


dktapps:
TesterPlugin has been migrated to PSR-4 to demonstrate the feature in action.

dktapps:
Since the stability of this feature primarily depends on upstream ClassLoader (which has now had unit tests added), I think this is good to merge.

PR #4202, opened 2021-04-29
First look at RakLib per-session channels

dktapps:


dktapps:
This PR is a PM-side implementation of pmmp/RakLib#52.

The performance change caused by this PR has not yet been assessed in detail, however I note the following points:
- Per-session channels eliminate the need to transmit session IDs over channels, which saves 4 bytes of memcopy for every packet sent and received.
- Lock contention on the channels themselves is substantially reduced, since now each session has its own I/O channels, instead of using a pair of global channels.
- Overhead when terminating malicious sessions is significantly reduced, since the whole channel can be immediately discarded. With global channels, we still have to remove the packets from that session from the packet buffer.

However, this comes at a price:
- Currently, all channels are still signalling to a single Snooze notifier. This means that the Snooze handler has to scan every channel to check for updates when any channel sends a message. This is a decidedly more expensive operation than the single-channel approach, with an O(nPlayers) scaling factor instead of O(1).
- The above problem could be changed by using a notifier per session, but this merely moves the problem of bruteforce checking for updates to the Snooze handler instead, which has to scan every notifier to check which ones caused notifications. An alternative would be to have a small set of shared notifiers, which would achieve the best of both worlds (less lock contention while still keeping notification source count small).

Therefore, it's not currently clear whether this implementation will be beneficial for performance or not.

dktapps:
https://github.com/pmmp/Snooze/commit/b93dfe9d4c62a4eb8411885abd0bd46ce09cfc67 significantly improves the performance of having many notifiers; however, the problem remains how we get those notifiers installed in the first place.

dktapps:
Closing this as stale, to be revisited at a later date. Due to ext-pmmpthread design changes, the way we implement this probably needs to be rethought anyway.

PR #4221, opened 2021-05-18
Generation overhaul (For Pocketmine 3.20)

TheShermanTanker:
## Introduction
This PR is a redo of https://github.com/pmmp/PocketMine-MP/pull/4170 but with correct rainfall and temperature values now inputted for all Biomes instead of just Mountains. The result is vastly improved terrain generation, as seen below

Note: This PR is meant for 3.20 (I've seen the dropped PR due to concerns about generation changes on patch so I know xd)

Edit: No idea why the render distance is so small, I set it to 40 in my client

### API changes
None

### Behavioural changes
Vastly improve terrain generation

## Backwards compatibility
None

## Follow-up
Unlikely to require any

Requires translations:

None

## Tests
image
_Larger biomes with better connections with surrounding biomes_

image
_Valley generated in between biomes_

image
_Islands can now generate_

image
_Expansive mountain generation_

image
_A river that now actually looks like one!_


dktapps:
PM3 is on feature freeze. This would have to go in PM4 at best, and it's kinda 11th hour for PM4 right now. Not to mention the fact that this will completely disrupt world generation for existing worlds, while the benefits are not that clear - I'm only seeing a few screenshots, we need large scale imagery to get a real idea of how this changes things.

TheShermanTanker:
> 
> 
> PM3 is on feature freeze. This would have to go in PM4 at best, and it's kinda 11th hour for PM4 right now. Not to mention the fact that this will completely disrupt world generation for existing worlds, while the benefits are not that clear - I'm only seeing a few screenshots, we need large scale imagery to get a real idea of how this changes things.

Would a video be better?

TheShermanTanker:
Update: I merged my private commits into the repo forgetting this was being used for a PR. Oops...

TheShermanTanker:
Closed PR because I accidentally merged something that wasn't supposed to be there :/

lukeeey:
> painstakingly changed all the classes named Level to World

For the vast vast majority of it it’s probably just a simple find and replace which literally takes a few seconds

> I accidentally merged something

You can delete commits, you know?

benda95280:
I'm sad to read again and again that we are not going to have anything else than protocol / security and bug fix 😞 
Why don't have a branch for that 😢 _It's a sad sad world ..._

dktapps:
We have a branch for that. It's called `master`.

benda95280:
Yep, but after 2 years working on PmmP for a server for my childs, there is too much works to be able to update plugins used to new API 😢

dktapps:
That's not my problem. My task is to ensure that PM3 remains stable with minimum maintenance while the real stuff gets done on PM4. The reason for the feature freeze is because PM3 and PM4 are wildly different, and adding anything to PM3 causes disruption for users as well as a large amount of additional work for me to maintain a dead version (PM4 is far superior to PM3, and frankly I can't wait to get rid of PM3).

PR #4224, opened 2021-05-22
Add eclipse files to gitignore

TheShermanTanker:


SOF3:
Do it in your own .git/info/exclude. Eclipse is not a recommended IDE for this project, and we do not want a slippery slope to add ignore files for every IDE in existence.

buchwasa:
I disagree with the closure of this, Eclipse should be included as it is a common IDE, I mean PM ignores files from Netbeans and when was the last time you heard of someone using Netbeans?

SOF3:
Nothing stops you from adding your ignore files to .git/info/exclude. Why does it have to be in the repo for everyone?

buchwasa:
Isn't that defeating the purpose of a gitignore? If me and 100 others use Eclipse on this project, then it makes more sense to provide it in the gitignore rather than tell every single person to deal with it themselves

SOF3:
There are 100 editors out there. Do we have to add 100 ignore templates? I fail to see evidence that eclipse PHP is a popular editor out there. Each editor, operating system, file management tool and other tools creates their own files. Do we have to add ignore files for Dropbox temp files just because someone clones their repo in Dropbox directory?

Nevertheless, if there are sufficiently many users who reply to this issue and agree that eclipse is popular, we might reopen it.

buchwasa:
What you're saying is nonsense, you're tunneling on only part of what I'm saying. There are hundreds of editors out there, but I'm arguing the case for common ones, as I explained in my first and second responses. Eclipse may not be as popular as it once was, as VS Code and PhpStorm have taken over the market, but it is still a common IDE as would be Netbeans, it's just one of those IDEs that most people have heard of.

SOF3:
Please add future suggestions to this discussion: https://github.com/pmmp/PocketMine-MP/discussions/4228

PR #4225, opened 2021-05-22
add populator api

TheShermanTanker:
## Introduction
Currently, plugins calling Biome->addPopulator will not actually have their populators registered and hence will be unable to (Conveniently) generate their own structures. This PR now allows generation threads to use the Biome data on the main thread, which in turn enables the proper use of addPopulator() by plugins

### Relevant issues
Not a performance issue

### API changes
Biome->addPopulator called by plugins on the main thread will now genuinely have their populators registered to the generator

### Behavioural changes
Plugins calling Biome->addPopulator for custom populators will now be able to populate levels, but other than this no changes in behaviour

## Backwards compatibility
Not backwards compatible. The addPopulator function will still be accessible by plugins in previous versions but will be entirely useless. There aren't any compatibility issues that cause hard crashes or severe issues though

## Follow-up
NIL

Requires translations:
None

## Tests
image
To spare you the gory details I made a quick plugin with a startup load sequence (Postworld arrives too late) and added a custom populator that simply added glass all over the world without touching any of the threads from the plugin, and the result was this monstrosity (Only meant as proof of concept so the populator isn't too complex)


dktapps:
This doesn't really make any sense, because you can much more easily just register biomes during your Generator's __construct() call.

dktapps:
This is also targeted at the wrong branch. PM3 is in feature freeze.

TheShermanTanker:
> 
> 
> This doesn't really make any sense, because you can much more easily just register biomes during your Generator's __construct() call.

I did  sort of try that, but it was quite a drag to have to do that for each Generator instance rather than just immediately in the Task they were registered in, out of convenience



> 
> 
> This is also targeted at the wrong branch. PM3 is in feature freeze.

Oops

dktapps:
I'm closing this because it doesn't make sense - you can much more easily register the needed biomes in your custom generator's `__construct`. In addition, this is targeted at a feature-frozen branch.

PR #4234, opened 2021-05-28
Apply damage items even if it is broken

korado531m7:
## Introduction
Broken items that is created by give command or plugins will be taken damage when called ```Item -> applyDamage()```.
Currently, broken items will not be removed.

### Relevant issues
nothing

### API changes
nothing

### Behavioural changes
Broken items cannot be used for a number of times.

## Backwards compatibility
nothing

## Follow-up
nothing

## Tests
* I have confirmed that players can be logged in normally.
* I have confirmed that durable items can use normally.
* I have confirmed that broken durable items was removed when use without any errors

### DIFF
--- a/src/pocketmine/item/Durable.php
+++ b/src/pocketmine/item/Durable.php
@@ -52,7 +52,7 @@ abstract class Durable extends Item{
 	 * @return bool if any damage was applied to the item
 	 */
 	public function applyDamage(int $amount) : bool{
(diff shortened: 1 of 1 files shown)


dktapps:
The current behaviour was designed with elytra in mind. This would break them to alter the behaviour of a case that's unsupported anyway (not achievable without game modifications). I don't think this makes sense.

PR #4237, opened 2021-06-03
Better sorting of items in InventoryHelpersTrait::addItem()

### LINKED ISSUE #1412: BaseInventory->addItem() with a list of separate, stackable items puts each stack in its own slot
### Issue description
```
$inventory->addItem(
    Item::get(Item::PLANKS, 0, 16),
    Item::get(Item::PLANKS, 0, 16),
    Item::get(Item::PLANKS, 0, 16),
    Item::get(Item::PLANKS, 0, 16)
);
```
will put each of the stacks in a separate slot instead of combining them as expected.

1. use the above code
2. profit

jasonw4331:
## Introduction

Dylan posted an issue on it, so its getting changed.

### Relevant issues

Fixes #1412

### API changes

`InventoryHelpersTrait::addItem()` will now stack items together even when provided separately as parameters.

### Behavioral changes

N/A

## Backwards compatibility

N/A

## Follow-up

N/A

## Tests

```php
$inventory->addItem(
	VanillaBlocks::OAK_PLANKS()->asItem()->setCount(16),
	VanillaBlocks::OAK_PLANKS()->asItem()->setCount(16),
	VanillaBlocks::OAK_PLANKS()->asItem()->setCount(16),
	VanillaBlocks::OAK_PLANKS()->asItem()->setCount(16)
);
```

image

### DIFF
--- a/src/inventory/InventoryHelpersTrait.php
+++ b/src/inventory/InventoryHelpersTrait.php
@@ -144,6 +144,22 @@ trait InventoryHelpersTrait{
 			}
 		}
 
+		foreach($itemSlots as $key1 => $slot1){
+			foreach($itemSlots as $key2 => $slot2){
+				if($key1 !== $key2 and $slot1->equals($slot2) and $slot1->getCount() < $slot1->getMaxStackSize()){
+					$amount = min($slot1->getMaxStackSize() - $slot1->getCount(), $slot2->getCount(), $this->getMaxStackSize());
+					if($amount > 0){
+						$slot2->setCount($slot2->getCount() - $amount);
+						$slot1->setCount($slot1->getCount() + $amount);
+						$itemSlots[$key1] = $slot1;
+						if($slot2->getCount() <= 0){
+							unset($itemSlots[$key2]);
+						}
+					}
+				}
+			}
+		}
+
 		$emptySlots = [];
 
 		for($i = 0, $size = $this->getSize(); $i < $size; ++$i){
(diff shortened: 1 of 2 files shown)


dktapps:
To be honest, I think a better solution is to just separate the logic of addItem() into an internal method, and then call the internal method once for each argument. That would guarantee consistency with minimal changes.

jasonwynn10:
Is that closer to what you're talking about?

src/inventory/InventoryHelpersTrait.php
```diff
@@ -168,43 +163,32 @@ public function addItem(Item ...$slots) : array{
...
-
-			if(count($itemSlots) === 0){
-				break;
-			}
 		}
 
-		if(count($itemSlots) > 0 and count($emptySlots) > 0){
+		if(count($emptySlots) > 0){
 			foreach($emptySlots as $slotIndex){
 				//This loop only gets the first item, then goes to the next empty slot
```
dktapps, line 181: The loop that this comment describes no longer exists.

dktapps:
Please write a PHPUnit test for this.

src/inventory/InventoryHelpersTrait.php
```diff
@@ -144,6 +144,17 @@ public function addItem(Item ...$slots) : array{
 			}
 		}
 
+		/** @var Item[] $returnSlots */
+		$returnSlots = [];
+
+		foreach($itemSlots as $item){
+			$returnSlots[] = $this->internalAddItem($item);
```
dktapps, line 151: This should check if the returned item `isNull()`, and don't add it to the result if so.

dktapps (changes requested):
This PR should include unit tests to show that the behaviour is as expected (the behaviour of addItem(1, 2, 3, 4) should be the same as 4 separate addItem() calls).

dktapps (approved):

PR #4238, opened 2021-06-03
Check Enderchest tile for viewer count

### LINKED ISSUE #4021: Ender chest lid closes if 2 players are viewing and one of them closes their ender chest GUI
### Issue description
Ender chests don't track their own viewer count; instead they allow the inventory to do it, like a regular chest does.

This is problematic because an ender chest inventory is not actually a chest inventory - an ender chest is merely a gateway to a part of the player's own inventory. This means that (under normal circumstances) an `EnderChestInventory` never has more than 1 viewer.

1. Have 2 players open an ender chest
2. Have one of them close their ender chest GUI.
3. On the other client, you can see that the ender chest lid is now closed.

### OS and versions

* PocketMine-MP: 3.17.4 / 4.0.0 @ d4290837f33655fd0de629bbac1228cedd680526
* PHP: 7.4.14

jasonw4331:
### Relevant issues

Fixes #4021

### API changes

N/A

### Behavioral changes

In-game sounds and block animations for enderchests now match vanilla.

## Backwards compatibility

N/A

## Follow-up

N/A

## Tests

In-game testing shows the behavior matches vanilla.

### DIFF
--- a/src/block/inventory/AnimatedBlockInventoryTrait.php
+++ b/src/block/inventory/AnimatedBlockInventoryTrait.php
@@ -34,6 +34,12 @@ trait AnimatedBlockInventoryTrait{
 		return count($this->getViewers());
 	}
 
+	/**
+	 * @return Player[]
+	 * @phpstan-return array<int, Player>
+	 */
+	abstract public function getViewers() : array;
+
 	abstract protected function getOpenSound() : Sound;
 
 	abstract protected function getCloseSound() : Sound;
--- a/src/block/inventory/AnimatedBlockInventoryTrait.php
+++ b/src/block/inventory/AnimatedBlockInventoryTrait.php
@@ -41,7 +41,7 @@ trait AnimatedBlockInventoryTrait{
 	public function onOpen(Player $who) : void{
 		parent::onOpen($who);
 
-		if($this->getHolder()->isValid() and $this->getViewerCount() < 1){
+		if($this->getHolder()->isValid() and $this->getViewerCount() === 1){
 			//TODO: this crap really shouldn't be managed by the inventory
 			$this->animateBlock(true);
 			$this->getHolder()->getWorld()->addSound($this->getHolder()->add(0.5, 0.5, 0.5), $this->getOpenSound());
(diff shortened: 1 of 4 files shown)


src/block/tile/EnderChest.php
```diff
@@ -27,6 +27,19 @@
...
+
+	public function getViewerCount() : int{
+		return $this->viewerCount;
+	}
+
+	public function setViewerCount(int $viewerCount) : self{
+		if($viewerCount < 0)
+			throw new \UnexpectedValueException($viewerCount . ' is too small of a value');
+		$this->viewerCount = $viewerCount;
+		return $this;
```
dktapps, line 40: Tiles don't usually have fluent APIs (particularly since plugins are supposed to avoid them in PM4).

dktapps:
A cleaner way to address this would be to replace `AnimatedBlockInventoryTrait->getViewers()` with `getViewerCount()`, so that it can be implemented differently for different types of inventories.

src/block/inventory/AnimatedBlockInventoryTrait.php
```diff
@@ -30,11 +30,9 @@
 trait AnimatedBlockInventoryTrait{
 	use BlockInventoryTrait;
 
-	/**
-	 * @return Player[]
-	 * @phpstan-return array<int, Player>
-	 */
-	abstract public function getViewers() : array;
```
dktapps, line 37: This should be retained, since other members of the trait use it.

src/block/inventory/EnderChestInventory.php
```diff
@@ -85,6 +86,14 @@ protected function internalSetContents(array $items) : void{
 		$this->inventory->setContents($items);
 	}
 
+	public function getViewerCount() : int{
+		$enderChest = $this->getHolder()->getWorld()->getTile($this->getHolder());
+		if(!$enderChest instanceof EnderChest){
+			throw new \UnexpectedValueException('Tile is not an Enderchest');
```
dktapps, line 92: I think it's OK to just return 0 in this case. It's possible for the tile to be missing or incorrect due to plugin interference, older worlds using setBlock(), the block being destroyed, etc.

Throwing here will make the server crash if the tile doesn't exist. It's not really a bug in the server, so it shouldn't do that.

src/block/tile/EnderChest.php
```diff
@@ -27,6 +27,19 @@
...
 
+	protected int $viewerCount = 0;
+
+	public function getViewerCount() : int{
+		return $this->viewerCount;
+	}
+
+	public function setViewerCount(int $viewerCount) : void{
+		if($viewerCount < 0){
+			throw new \InvalidArgumentException($viewerCount . ' is too small of a value');
```
dktapps, line 38: You don't need to include the `viewerCount` in the message - it'll be visible in the stack trace.
However, this message should specify that the value can't be negative.

dktapps (commented):
This looks mostly OK, but I've commented on a couple of minor nits.

PR #4239, opened 2021-06-04
Make Liquids check if they are flowing into unloaded chunks

### LINKED ISSUE #3990: Liquids may attempt to flow into locked terrain
Link to crashdump: https://crash.pmmp.io/view/4711635

jasonw4331:
## Introduction

Crashdump https://crash.pmmp.io/view/4711635

### Relevant issues

Fixes #3990

### API changes

N/A

### Behavioral changes

Liquids will now detect if the block they are attempting to flow into is in a loaded chunk or not

## Backwards compatibility

N/A

## Follow-up

N/A

## Tests

Tests pending

### DIFF
--- a/src/block/Liquid.php
+++ b/src/block/Liquid.php
@@ -500,6 +500,6 @@ abstract class Liquid extends Transparent{
 	}
 
 	protected function canFlowInto(Block $block) : bool{
-		return $this->pos->getWorld()->isInWorld($block->pos->x, $block->pos->y, $block->pos->z) and $block->canBeFlowedInto() and !($block instanceof Liquid and $block->isSource()); //TODO: I think this should only be liquids of the same type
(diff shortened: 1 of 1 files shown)


dktapps:
This doesn't fix the crashdump you referenced.

src/block/Liquid.php
```diff
@@ -500,6 +500,6 @@ protected function liquidCollide(Block $cause, Block $result) : bool{
 	}
 
 	protected function canFlowInto(Block $block) : bool{
-		return $this->pos->getWorld()->isInWorld($block->pos->x, $block->pos->y, $block->pos->z) and $block->canBeFlowedInto() and !($block instanceof Liquid and $block->isSource()); //TODO: I think this should only be liquids of the same type
+		return $this->pos->getWorld()->isInWorld($block->pos->x, $block->pos->y, $block->pos->z) and $block->canBeFlowedInto() and !($block instanceof Liquid and $block->isSource()) and !($block instanceof Air and $block->pos->getWorld()->getChunk($block->pos->x >> 4, $block->pos->z >> 4) === null); //TODO: I think this should only be liquids of the same type
```
dktapps, line 503: This check is confusing. It should be refactored to `(!block instanceof Air || isChunkLoaded)`.

(That said, please refer to my earlier comment: this doesn't fix the crash you're referring to).

jasonwynn10: I've refactored as requested. How would this not fix the mentioned crash?

dktapps:
I'm going to close this, since it's just playing whack-a-mole with the actual underlying problem at best.

PR #4242, opened 2021-06-04
Added Type NBT data for illager banners

### LINKED ISSUE #2951: Illager banner in creative inventory places a normal white banner
### Issue description
Illager banner appears to use a non-standard method to define its pattern in NBT, which is not recognized by PM. This leads to the placement of a regular white banner when placing an illager banner.

This is considered a bug because the item appears in the creative inventory, but doesn't work as designed.

1. Get an illager banner from the creative inventory.
2. Place it and notice it turns white.

### OS and versions

* PocketMine-MP: 3.8.3, master
* PHP: 7.3.5
* Server OS: linux
* Game version: all

jasonw4331:
## Introduction

Banners can now be set and used as the illager type.

### Relevant issues

Fixes #2951

### API changes

6 new functions:
`item\Banner::isIllagerPattern()`
`item\Banner::setIllagerPattern(bool)`
`tile\Banner::isIllagerPattern()`
`tile\Banner::setIllagerPattern(bool)`
`block\BaseBanner::isIllagerPattern()`
`block\BaseBanner::setIllagerPattern(bool)`

### Behavioral changes

N/A

## Backwards compatibility

N/A

## Follow-up

WIP

## Tests

Pending Tests

### DIFF
--- a/src/block/tile/Banner.php
+++ b/src/block/tile/Banner.php
@@ -40,6 +40,7 @@ use pocketmine\world\World;
 class Banner extends Spawnable{
 
 	public const TAG_BASE = "Base";
+	public const TAG_TYPE = "Type"; // IntTag
 	public const TAG_PATTERNS = "Patterns";
 	public const TAG_PATTERN_COLOR = "Color";
 	public const TAG_PATTERN_NAME = "Pattern";
@@ -52,6 +53,8 @@ class Banner extends Spawnable{
 	 * @phpstan-var list<BannerPatternLayer>
 	 */
 	private $patterns = [];
+
+	private bool $illagerPattern = false;
 
 	public function __construct(World $world, Vector3 $pos){
 		$this->baseColor = DyeColor::BLACK();
@@ -83,6 +86,8 @@ class Banner extends Spawnable{
 				$this->patterns[] = new BannerPatternLayer($patternType, $patternColor);
 			}
 		}
+
+		$this->illagerPattern = (bool) $nbt->getInt(self::TAG_TYPE, 0);
 	}
 
 	protected function writeSaveData(CompoundTag $nbt) : void{
@@ -97,6 +102,7 @@ class Banner extends Spawnable{
 			);
 		}
 		$nbt->setTag(self::TAG_PATTERNS, $patterns);
+		$nbt->setInt(self::TAG_TYPE, (int) $this->illagerPattern);
 	}
 
 	protected function addAdditionalSpawnData(CompoundTag $nbt) : void{
@@ -111,6 +117,7 @@ class Banner extends Spawnable{
 			);
 		}
 		$nbt->setTag(self::TAG_PATTERNS, $patterns);
+		$nbt->setInt(self::TAG_TYPE, (int) $this->illagerPattern);
 	}
 
 	/**
@@ -142,6 +149,14 @@ class Banner extends Spawnable{
 	 */
 	public function setPatterns(array $patterns) : void{
 		$this->patterns = $patterns;
+	}
+
+	public function isIllagerPattern() : bool{
+		return $this->illagerPattern;
+	}
+
+	public function setIllagerPattern(bool $illagerPattern) : void{
+		$this->illagerPattern = $illagerPattern;
 	}
 
 	public function getDefaultName() : string{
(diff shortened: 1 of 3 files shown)


jasonwynn10:
I’d like to fix the banners which are stored in the creative inventory next. What’s the best place for the server recognize the difference in nbt sent from the client?

src/block/tile/Banner.php
```diff
@@ -83,6 +86,8 @@ public function readSaveData(CompoundTag $nbt) : void{
 				$this->patterns[] = new BannerPatternLayer($patternType, $patternColor);
 			}
 		}
+
+		$this->illagerPattern = (bool) $nbt->getInt(self::TAG_TYPE, 0);
```
dktapps, line 90: `getInt() !== 0` is the preferred way for this.

src/block/tile/Banner.php
```diff
@@ -97,6 +102,7 @@ protected function writeSaveData(CompoundTag $nbt) : void{
 			);
 		}
 		$nbt->setTag(self::TAG_PATTERNS, $patterns);
+		$nbt->setInt(self::TAG_TYPE, (int) $this->illagerPattern);
```
dktapps, line 105: `illagerPattern ? 1 : 0` is the preferred way for this.

src/block/tile/Banner.php
```diff
@@ -111,6 +117,7 @@ protected function addAdditionalSpawnData(CompoundTag $nbt) : void{
...
```
dktapps, line 120: Same again here.

src/item/Banner.php
```diff
@@ -113,6 +125,7 @@ protected function deserializeCompoundTag(CompoundTag $tag) : void{
 				$this->patterns[] = new BannerPatternLayer($patternType, $patternColor);
 			}
 		}
+		$this->illagerPattern = (bool) $tag->getInt(self::TAG_TYPE, 0);
```
dktapps, line 128: See above.

src/item/Banner.php
```diff
@@ -133,5 +146,6 @@ protected function serializeCompoundTag(CompoundTag $tag) : void{
 		}else{
 			$tag->removeTag(self::TAG_PATTERNS);
 		}
+		$tag->setInt(self::TAG_TYPE, (int) $this->illagerPattern);
```
dktapps, line 149: See above.

dktapps (commented):
The reason I never implemented this is because I wasn't sure how the illager banner type should be handled. I'm not sure if an illager banner can have additional patterns.

dktapps:
These APIs should also be provided in `block\BaseBanner`.

dktapps:
It would be nice if you would test what happens when an illager banner receives extra patterns. I don't think there's a way to do that in vanilla.

jasonwynn10:
When the illager banner type is set, any pattern layers turn into this:

image

src/block/tile/Banner.php
```diff
@@ -120,6 +121,12 @@ protected function addAdditionalSpawnData(CompoundTag $nbt) : void{
 		$nbt->setInt(self::TAG_TYPE, $this->illagerPattern ? 1 : 0);
 	}
 
+	public function copyDataFromItem(Item $item) : void{
+		parent::copyDataFromItem($item);
+		/** @var \pocketmine\item\Banner $item */
```
dktapps, line 126: This isn't guaranteed.

src/item/Banner.php
```diff
@@ -90,6 +93,16 @@ public function setPatterns(array $patterns) : self{
 		return $this;
 	}
 
+	public function isIllagerPattern() : bool{
+		return $this->illagerPattern;
+	}
+
+	public function setIllagerPattern(bool $illagerPattern) : self{
+		$this->illagerPattern = $illagerPattern;
+		$this->color = DyeColor::WHITE(); // meta 15 matches vanilla
```
dktapps, line 102: This is unexpected behaviour, particularly since the colour could later be changed back to something else.

src/item/Banner.php
```diff
@@ -66,7 +66,7 @@ public function setColor(DyeColor $color) : self{
 	}
 
 	public function getMeta() : int{
-		return DyeColorIdMap::getInstance()->toInvertedId($this->color);
+		return $this->illagerPattern ? 15 : DyeColorIdMap::getInstance()->toInvertedId($this->color);
```
dktapps, line 69: That's even worse. Now it will always show as white regardless of the server-side colour, even if the client starts to support this in the future.

jasonwynn10: What's the expected behavior?

dktapps:
Since this creates conditionally useless APIs (and is also 2 years obsolete), I'm going to close this. The new serializer system in PM5 should facilitate dealing with illager banners in a better way.

PR #4243, opened 2021-06-04
The top of blocks are now collided with correctly

### LINKED ISSUE #2041: Magma blocks don't work any more
### Issue description
Since the bug fix in https://github.com/pmmp/PocketMine-MP/commit/cb90e30bcf3c8012146bc37441bc59cb79ff2281, magma blocks no longer deal damage when an entity walks over the top of them. This is because the hack that was used to trigger onEntityCollide() is now no longer in effect.

1. walk on magma
2. notice it doesn't damage you

### OS and versions
* PocketMine-MP: anything since https://github.com/pmmp/PocketMine-MP/commit/cb90e30bcf3c8012146bc37441bc59cb79ff2281

### LINKED ISSUE #2731: No damage when standing on top of a cactus
### Issue description
The client seems to think cacti have a full bounding box. This causes the same problem we see in #2041 with magma.

- Expected result: Standing on top of a cactus should deal damage.
- Actual result: Happily walking round on top of cacti immune.

1. Stand on top of a cactus
2. Notice you're not getting damaged by it.

### OS and versions

* PocketMine-MP: 3.5.9
* PHP: 7.2.14
* Server OS: linux

jasonwynn10:
## Introduction

Block collisions for the blocks directly below the player will now take effect when the player is within 3/16 of the block.

### Relevant issues

* Fixes #2731
* Fixes #2041

### API changes

N/A

### Behavioral changes

N/A

## Backwards compatibility

N/A

## Follow-up

N/A

## Tests

In-game testing shows standing on cactus and magma blocks damages the player.
Standing on trapdoors and half slabs does not damage the player.
Standing on carpets will damage the player, matching vanilla.

### DIFF
--- a/src/entity/Entity.php
+++ b/src/entity/Entity.php
@@ -1247,7 +1247,7 @@ abstract class Entity{
 			$inset = 0.001; //Offset against floating-point errors
 
 			$minX = (int) floor($this->boundingBox->minX + $inset);
-			$minY = (int) floor($this->boundingBox->minY + $inset);
+			$minY = (int) floor($this->boundingBox->minY + $inset - (($this->boundingBox->minY - ((int) $this->boundingBox->minY)) > 2 / 16 ? 0 : 1));
 			$minZ = (int) floor($this->boundingBox->minZ + $inset);
 			$maxX = (int) floor($this->boundingBox->maxX - $inset);
 			$maxY = (int) floor($this->boundingBox->maxY - $inset);


dktapps:
please don't put special characters in your branch name; this makes it a pain in the ass to checkout in a bash shell.

This change causes other unexpected behaviour, such as getting set on fire when standing on a block adjacent to the fire.

PR #4246, opened 2021-06-06
Remove movement workaround from 1.14.30

buchwasa:
## Introduction
This pull request removes the movement workaround from 1.14.30 where the packets MoveActorAbsolute and MoveActorDelta no longer worked on players. I do not know when this was fixed but this was tested on 1.16.221 and it works fine. It is said that the MoveActor packets are smoother over the MovePlayer counterparts and I do believe this myself, but it could also be a case of the placebo effect in my own testing.

## Changes
Human->broadcastMovement() was removed, using Entity->broadcastMovement() instead.

## Tests
https://user-images.githubusercontent.com/17461354/120941860-2381e500-c6f3-11eb-9125-ecdfa1df8632.mp4


dktapps:
This looks OK, but I'm going to hold off on merging this until I have a chance to inspect this for myself. I don't want to merge this just to revert it again.

PR #4283, opened 2021-06-24
InventoryHelpersTrait::AddItem() cannot add items with a count greater than maxstack

aieuo:
## Introduction
Fixed a bug that prevented adding more than 64 items to the inventory even if there was enough space when the following code was executed.

```php
$player->getInventory()->addItem(VanillaItems::APPLE()->setCount(100));
```

### DIFF
--- a/tests/phpunit/inventory/BaseInventoryTest.php
+++ b/tests/phpunit/inventory/BaseInventoryTest.php
@@ -85,4 +85,18 @@ class BaseInventoryTest extends TestCase{
 		}
 		self::assertSame(20, $leftoverCount);
 	}
+
+	public function testAddItemWithLargeQuantity() : void{
+		$inventory = new class(10) extends SimpleInventory{
+
+		};
+		$inventory->addItem(VanillaItems::APPLE()->setCount(100));
+
+		$count = 0;
+		foreach($inventory->getContents() as $item){
+			self::assertTrue($item->equals(VanillaItems::APPLE()));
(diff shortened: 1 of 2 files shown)


dktapps:
Please add a unit test for this.

tests/phpunit/inventory/BaseInventoryTest.php
```diff
@@ -85,4 +85,18 @@ public function testAddMultipleItemsInOneCallWithLeftover() : void{
 		}
 		self::assertSame(20, $leftoverCount);
 	}
+
+	public function testAddItemWithLargeQuantity() : void{
+		$inventory = new class(10) extends SimpleInventory{
+
+		};
+		$inventory->addItem(VanillaItems::APPLE()->setCount(100));
```
dktapps, line 93: You should assert that the number of items returned from `addItem()` is zero as well.

tests/phpunit/inventory/BaseInventoryTest.php
```diff
@@ -85,4 +85,18 @@ public function testAddMultipleItemsInOneCallWithLeftover() : void{
 		}
 		self::assertSame(20, $leftoverCount);
 	}
+
+	public function testAddItemWithLargeQuantity() : void{
```
dktapps, line 89: testAddItemWithOversizedCount would be a better name.

dktapps (approved):
Build failures are unrelated to the PR.

PR #4293, opened 2021-07-03
A few changes to get rid of the extra loop

PechenkaPM:
## Introduction
I noticed a useless loop that iterates over all entities and can be removed

### Relevant issues
N/A

### API changes
N/A

### Behavioural changes
N/A

## Backwards compatibility
N/A

## Follow-up
N/A

-->

## Tests
Pending tests


PechenkaPM:
I tested this, everything works

dktapps:
We don't want to update every single entity every tick. In particular, inanimate / non-moving entities that are not within a ticking chunk won't get updated (e.g. stationary dropped items, paintings). This change will cause performance degradation.

PechenkaPM:
You remove entities from the list of ticking (updateEntities), after which you call the "tickChunks" function, and it returns the entities back by calling "scheduleUpdate" for all "entities"

dktapps:
Actually, it doesn't. As I said, it only updates entities which are within ticking chunks.

PR #4324, opened 2021-07-20
Caves

Pinesolz:
## Introduction
Introducing Caves to PocketMine-MP
Will generate caves throughout worlds
Caves will be built up of various blocks based on biomes/regions

I'm sure theres stuff missing, but its a start...
-->

## Changes
Added Caves and CaveNodes files ( pocketmine \ level \ generator \ populator )

### Behavioural changes
Will help add caves to worlds

## Follow-up

-->

### DIFF
--- a/src/pocketmine/level/generator/populator/Caves.php
+++ b/src/pocketmine/level/generator/populator/Caves.php
@@ -1,0 +1,135 @@
+<?php
+
+namespace pocketmine\level\generator\populator;
+
+use pocketmine\level\ChunkManager;
+use pocketmine\level\generator\populator\Populator;
+use pocketmine\math\Vector3;
+use pocketmine\utils\Random;
+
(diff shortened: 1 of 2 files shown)


dktapps:
stable is in feature freeze, and even if it wasn't, this can't be added on a patch release.

dktapps:
This code is a 1:1 copy and paste from an old spoon. https://github.com/search?q=generateCaveBranch&type=code

Aside from the licensing issues (the code comes from Genisys, which is a rat's nest of copyright violations), it's also plainly obvious that you added this code with no understanding of what it does - you haven't even registered the populators, so this code will do nothing whatsoever. Therefore, even if I didn't have other reasons to close this, it's obvious that you wouldn't be able to amend this if I were to take issue with its various technical faults due to blindly copy-pasting code you don't understand.

PR #4328, opened 2021-07-22
Use OpenSSL for ECDH during client login, drop mdanter/ecc

dktapps:
While this code is a bit more cancerous to read, it's a lot more performant, and costs us one less dependency (mdanter/ecc is no longer required).

This was salvaged from a stash I created several months ago, so I don't remember for sure how functional this code is; however, it appears OK, and everything is working as intended as far as I can tell.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
       "ext-zlib": ">=1.2.11",
       "composer-runtime-api": "^2.0",
       "adhocore/json-comment": "^1.1",
-      "mdanter/ecc": "^1.0",
+      "fgrosse/phpasn1": "^2.3",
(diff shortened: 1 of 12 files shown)


dktapps:
One issue I've just become aware of is that the key resources will leak on <8.0, because they are resources and not expressly cleaned up - resources are only automatically destroyed at request end, which means that over the lifetime of a server, every async worker that runs one of these tasks will leak.

dktapps:
After cleanup and some tests, some performance numbers:

`ProcessLoginTask`: before change ~40ms, after change ~15ms
`PrepareEncryptionTask`: before change ~180ms, after change ~10ms

As an added bonus, ~50KB is also slashed off the phar size.

PR #4332, opened 2021-07-24
Use a subprocess for reading lines from STDIN

dktapps:
## Introduction
		 * This pile of shit exists because PHP on Windows is broken, and can't handle stream_select() on stdin or pipes
		 * properly - stdin native triggers stream_select() when a key is pressed, causing it to get stuck in fgets()
		 * waiting for a line that might never come (and Windows doesn't support character-based reading either), and
		 * pipes just constantly trigger stream_select() instead of only when data is returned, rendering it useless.
		 *
		 * This results in whichever process reads stdin getting stuck on shutdown, which previously forced us to kill
		 * the entire server process to make it go away.
		 *
		 * To get around this problem, we delegate the responsibility of reading stdin to a subprocess, which we can
		 * then brutally murder when the server shuts down, without killing the entire server process.
		 * Thankfully, stream_select() actually works properly on sockets, so we can use them for inter-process
		 * communication.

I recently noticed that my PHP debug build wasn't reporting memory leaks because of the server process force-kill on shutdown.
This PR minimizes the impact of that and various other problems (e.g. destructors not getting run).

### API changes
None.

### Behavioural changes
- A subprocess is now used for reading lines from STDIN.
- Server no longer gets force-killed on Windows during shutdown when a custom console is used, when you happened to press a key without pressing enter, or when a player ran /stop in game.

## Backwards compatibility

## Follow-up
I wanted to use stream_socket_pair() and bind a socket directly to the subprocess' stdout, to remove the need for a temporary socket server for the subprocess to connect to. However, this led to a completely non-functioning implementation for some reason (the parent process never received any data), so I went with this instead.

## Tests
This has been extensively tested on Windows. *nix platforms have typically been much less of a pain in the ass than Windows, so I expect it will work fine there too - we'll see once the CI passes.

### DIFF
--- a/src/console/ConsoleReaderChildProcess.php
+++ b/src/console/ConsoleReaderChildProcess.php
@@ -1,0 +1,49 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
+ *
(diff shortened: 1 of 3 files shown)


dktapps:
This has now been tested and found working on Windows, macOS, Linux and Android. I think we're good to merge.

PR #4338, opened 2021-07-31
the function stringToLegacy gives already null useless to put a defau…

SenseiTarzan:
**Small change RuntimeBlockMapping **
I changed the RuntimeBlockMapping because there is a useless thing at line 90 because stringToLegacy returns an int or a null

`$id = $legacyIdMap->stringToLegacy($pair->getId()) ?? null; to $id = $legacyIdMap->stringToLegacy($pair->getId()); `


dktapps:
Thanks. This looks like leftover code from the days when it was handling raw arrays.

PR #4343, opened 2021-08-01
 add Slow Falling Effect in PMMP 4.0

SenseiTarzan:
add Slow Falling and remove fall damage entity if has Slow Falling

### DIFF
--- a/src/entity/effect/VanillaEffects.php
+++ b/src/entity/effect/VanillaEffects.php
@@ -58,6 +58,7 @@ use function assert;
  * @method static Effect WATER_BREATHING()
  * @method static Effect WEAKNESS()
  * @method static WitherEffect WITHER()
+ * @method static SlowFallingEffect Slow_Falling()
  */
 final class VanillaEffects{
 	use RegistryTrait;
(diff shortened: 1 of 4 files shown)


src/entity/effect/VanillaEffects.php
```diff
@@ -58,6 +58,7 @@
  * @method static Effect WATER_BREATHING()
  * @method static Effect WEAKNESS()
  * @method static WitherEffect WITHER()
+ * @method static SlowFallingEffect Slow_Falling()
```
buchwasa, line 61: You shouldn't manually write this, it's autogenerated

SenseiTarzan: How does it

SenseiTarzan: How does it work?

dktapps: https://github.com/pmmp/PocketMine-MP/blob/master/build/generate-registry-annotations.php

SenseiTarzan: it's good I just did with generate-registry-annotations for the annatotion of the sloaw falling in the vanillaEffect and the change of the names after the generation of the method

dktapps:
This also needs to be registered in PotionType enum: https://github.com/pmmp/PocketMine-MP/blob/07d97bbdeb891a019f11272b0a10a23f135180de/src/item/PotionType.php#L198
https://github.com/pmmp/PocketMine-MP/blob/07d97bbdeb891a019f11272b0a10a23f135180de/src/item/PotionType.php#L201

SenseiTarzan:
ok I show that there is no problem

dktapps:

Issue #4352, opened 2021-08-07
BlockGrowEvent not called when growing Bamboo 


ColinHDev:
### Issue description
The BlockGrowEvent should be called when a crop or plant is growing through randomTick or bone meal.
However, it is missing in both the BambooSapling and Bamboo classes.
The event should be called if the grow function in one of these classes is executed like it is called in the grow functions of the Sugarcane and Cactus class (two crops, which are the most similar to Bamboo).

### Steps to reproduce the issue
- register a listener for the BlockGrowEvent
- let Bamboo grow through randomTick or Bonemeal
-> Event won't be called

### OS and versions
* PocketMine-MP: 4.0.0 (Phar build on 1st August, but with a look at the recent commits, it can be easily seen that the problem hasn't been addressed yet)
https://github.com/pmmp/PocketMine-MP/commit/eb23d27004967c7e08ecdedb3747ce3aab662451
* PHP: 8.0.9
* Using JIT: no
* Server OS:  Debian GNU/Linux 10 (buster
* Game version: Win10

### Plugins
- not caused by a Plugin, it would also occur on a server without any plugins

### Crashdump, backtrace or other files
---

ColinHDev:
Just out of curiosity, may I ask, why the growing of Saplings never called an event?
I know, BlockGrowEvent wouldn't suit, because it just provides the original Block and its new state, but couldn't there just be a SaplingGrowEvent that would provide the sapling block as well as the BlockTransaction of the tree?

dktapps:
BlockTransaction didn't exist prior to PM4.

ColinHDev:
Yes, I know, but I am talking about pm4 (and upwards)

dktapps:
The same reason why there are no rockets to Mars - nobody has done it yet.

ColinHDev:
If I make a PR, would these additions (BlockGrowEvent on Bamboo and SaplingGrowEvent) be accepted to the current master branch or currently not because of feature freeze?

Issue #2645, opened 2019-01-05
No tree growth related events


SalmonDE:
### Issue description
Trees don't fire events instead they are instantly generated. This makes it nearly impossible to e.g. stop the tree from growing.

dktapps:
Updated your comment to remove issue template noise.

PR #4354, opened 2021-08-08
StructureGrowEvent

ColinHDev:
## Introduction
So far, no event has been called when growing Bamboo (or Bamboo Saplings) or Saplings.
Both Bamboo (or Bamboo Saplings) or Saplings use BlockTransactions which makes them unsuitable for the BlockGrowEvent.
With the addition of the BlockSproutEvent, there is now an event that will be called in these situations where a plant grows more than one block at a time and provides that BlockTransaction to the plugin as well the block which grows and its new state.

### Relevant issues
* Solves https://github.com/pmmp/PocketMine-MP/issues/4352
* Solves https://github.com/pmmp/PocketMine-MP/issues/2645 (thanks to @SalmonDE for mentioning)

### API changes
Upon using the Tree::growTree() function, there would be an optional seventh parameter, which can be used to enable the calling of the new BlockSproutEvent.
Upon using the Tree::placeObject() function, there would be an optional sixth parameter, which can be used to enable the calling of the new BlockSproutEvent.

### Behavioural changes
- none

## Backwards compatibility
The Tree class is not only called when a sapling grows but also during chunk population. Because we only want to call an event when a sapling grows, a $callEvent (bool) variable has been added to the functions Tree::growTree() and Tree::placeObject().
This causes breaks in every plugin which has classes that extend the Tree class. These plugins would need to add the $callEvent variable to their Tree child classes.
Due to the default value of the $callEvent variable (= false) there shouldn't be any breaks in existing generators.

## Follow-up
- - -

Requires translations:
-> no

## Tests
the tree generation works without any changes to the (default) generator itself:
(proof that there shouldn't be any breaks in generator plugins unless they use their own Tree child classes)
tree generation works
Screenshot (496)
Screenshot (497)

tested with https://github.com/ColinHDev/BlockSproutEventTest
(plugin registers listener which prints the name of the growing block and the name of its new state to the console and cancels the event when Bamboo (not Bamboo Saplings) or Birch Saplings try to grow.)

before using bone meal (once per block):
before using bone meal

after using bone meal:
after using bone meal

old block name and name of the new State logged in the console during the event:
block names logged in console


SalmonDE:
Also related to this issue: #2645

dktapps:
Please reformat this according to the code style in the repository.

ColinHDev:
it should match now according to https://github.com/pmmp/PocketMine-MP/blob/stable/CONTRIBUTING.md

src/event/block/BlockSproutEvent.php
```diff
@@ -0,0 +1,26 @@
...
+namespace pocketmine\event\block;
+
+use pocketmine\block\Block;
+use pocketmine\world\BlockTransaction;
+
+/**
+ * Called when Saplings or Bamboo grow.
+ * These types of plants tend to change multiple blocks at once upon growing, that's why this event returns the BlockTransaction.
+ */
+class BlockSproutEvent extends BaseBlockChangeEvent{
```
dktapps, line 14: This is misleading. It's not a single block that's being changed, while this extension (and thereby the presence of `::getBlock()` and `::getNewState()`) implies that it is.

ColinHDev:
I was unsure whether to use BaseBlockChangeEvent or only BlockEvent.
I decided to use BaseBlockChangeEvent to give plugins also directly what happens to the block that is growing.
But yeah, you could do that by getting the block from the transaction.
I will switch it to BlockEvent

ColinHDev:
changed it and updated the test plugin (https://github.com/ColinHDev/BlockSproutEventTest) to work with the changed functions

src/event/block/BlockSproutEvent.php
```diff
@@ -0,0 +1,29 @@
...
+namespace pocketmine\event\block;
+
+use pocketmine\block\Block;
+use pocketmine\event\Cancellable;
+use pocketmine\event\CancellableTrait;
+use pocketmine\world\BlockTransaction;
+
+/**
+ * Called when Saplings or Bamboo grow.
+ * These types of plants tend to change multiple blocks at once upon growing, that's why this event returns the BlockTransaction.
```
dktapps, line 14:
```suggestion
 * These types of plants tend to change multiple blocks at once upon growing.
```
New users aren't going to question the fact that it returns a BlockTransaction.

src/event/block/BlockSproutEvent.php
```diff
@@ -0,0 +1,29 @@
...
+
+namespace pocketmine\event\block;
+
+use pocketmine\block\Block;
+use pocketmine\event\Cancellable;
+use pocketmine\event\CancellableTrait;
+use pocketmine\world\BlockTransaction;
+
+/**
+ * Called when Saplings or Bamboo grow.
```
dktapps, line 13:
```suggestion
 * Called when objects such as Saplings or Bamboo grow.
```
This might be applied to other things in the future (e.g. kelp, chorus plants, etc).

src/world/generator/object/Tree.php
```diff
@@ -100,11 +102,22 @@ public function canPlaceObject(ChunkManager $world, int $x, int $y, int $z, Rand
...
 
-	public function placeObject(ChunkManager $world, int $x, int $y, int $z, Random $random) : void{
+	public function placeObject(ChunkManager $world, int $x, int $y, int $z, Random $random, bool $callEvent = false) : void{
 		$transaction = new BlockTransaction($world);
 		$this->placeTrunk($x, $y, $z, $random, $this->generateChunkHeight($random), $transaction);
 		$this->placeCanopy($x, $y, $z, $random, $transaction);
 
+		if($callEvent){
+			$ev = new BlockSproutEvent(
+				$world->getBlockAt($x, $y, $z), // returns the sapling (-> block which grows)
```
dktapps, line 112: This comment seems redundant.

celishere:
👍🏼

ColinHDev:
I updated the comments, I hope everything is fine now.

src/world/generator/object/Tree.php
```diff
@@ -100,11 +102,22 @@ public function canPlaceObject(ChunkManager $world, int $x, int $y, int $z, Rand
...
+	public function placeObject(ChunkManager $world, int $x, int $y, int $z, Random $random, bool $callEvent = false) : void{
 		$transaction = new BlockTransaction($world);
 		$this->placeTrunk($x, $y, $z, $random, $this->generateChunkHeight($random), $transaction);
 		$this->placeCanopy($x, $y, $z, $random, $transaction);
 
+		if($callEvent){
+			$ev = new BlockSproutEvent(
+				$world->getBlockAt($x, $y, $z),
+				$transaction
+			);
```
dktapps, line 114: Seems unnecessary to spread this over multiple lines now, no?

ColinHDev: It felt a little bit crowded for me if everything was in one line, but I will change it if you are ok with it.

ColinHDev:
Are there still problems with this PR or is it not planned to be merged to 4.0 or why is it not merged yet?

dktapps:
My only issue with it is that I'm not convinced `BlockSproutEvent` is a good name for the event. It doesn't really describe what the event does. I can easily see people getting confused wtf is the difference to `BlockGrowEvent` and not being sure which event they need.

ColinHDev:
And using extra comments for an explanation?
At the moment, I don't have a good idea for a better event name. Do you have one?

ColinHDev:
My mother tongue is not English, so it is quite hard for me to come up with a good name for that specific case.
Currently, the event would be called on growing Bamboo and Saplings but later on, possibly growing kelp if waterlogging would become possible. So it will be called if plants grow over multiple blocks.
So maybe something like BlockEnlargeEvent or PlantEnlargeEvent or BlockGrowEnlargedEvent?

ColinHDev:
retry to post:
My mother tongue is not English, so it is quite hard for me to come up with a good name for that specific case.
Currently, the event would be called on growing Bamboo and Saplings but later on, possibly growing kelp if waterlogging would become possible. So it will be called if plants grow over multiple blocks.
So maybe something like BlockEnlargeEvent or PlantEnlargeEvent or BlockGrowEnlargedEvent?

ColinHDev:
I think I never asked. Do you think, @dktapps, that StructureGrowEvent is good enough as a name?

ColinHDev:
yikes, that did not work as intended. I am using Git way too little.
sorry :/

dktapps:
You can remove the botched commits:
```
git reset HEAD~2 --hard
git push -f
```

ColinHDev:
Thank you very much for the help, @dktapps 
That was frightening. For a minute, I thought, I fucked everything up ._.

PR #4356, opened 2021-08-09
Bows can now use offhand arrows

rasu3n:
## Introduction

Bows currently do not use offhand arrows
## Tests

https://user-images.githubusercontent.com/76860328/128688083-ee1081de-3755-4c85-8b3e-6de5d21c6318.mp4

### DIFF
--- a/src/item/Bow.php
+++ b/src/item/Bow.php
@@ -46,8 +46,16 @@ class Bow extends Tool implements Releasable{
 
 	public function onReleaseUsing(Player $player) : ItemUseResult{
 		$arrow = VanillaItems::ARROW();
-		if($player->hasFiniteResources() and !$player->getInventory()->contains($arrow)){
-			return ItemUseResult::FAIL();
+
+		$inventory = null;
+		if($player->hasFiniteResources()){
+			if($player->getOffHandInventory()->contains($arrow)){ // Receives a Mismatch Transaction. why???
+				$inventory = $player->getOffHandInventory();
+			}elseif($player->getInventory()->contains($arrow)){
+				$inventory = $player->getInventory();
+			}else{
+				return ItemUseResult::FAIL();
+			}
 		}
 
 		$location = $player->getLocation();
@@ -110,7 +118,7 @@ class Bow extends Tool implements Releasable{
 
 		if($player->hasFiniteResources()){
 			if(!$infinity){ //TODO: tipped arrows are still consumed when Infinity is applied
-				$player->getInventory()->removeItem($arrow);
+				$inventory?->removeItem($arrow);
 			}
 			$this->applyDamage(1);
 		}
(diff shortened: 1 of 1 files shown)


src/item/Bow.php
```diff
@@ -110,7 +112,7 @@ public function onReleaseUsing(Player $player) : ItemUseResult{
 
 		if($player->hasFiniteResources()){
 			if(!$infinity){ //TODO: tipped arrows are still consumed when Infinity is applied
-				$player->getInventory()->removeItem($arrow);
+				$inventory?->removeItem($arrow);
```
dktapps, line 115: I don't see why a nullsafe operator is needed here?

src/item/Bow.php
```diff
@@ -46,7 +46,9 @@ public function getMaxDurability() : int{
 
 	public function onReleaseUsing(Player $player) : ItemUseResult{
 		$arrow = VanillaItems::ARROW();
-		if($player->hasFiniteResources() and !$player->getInventory()->contains($arrow)){
+
+		$inventory = $player->getOffHandInventory();
+		if($player->hasFiniteResources() and !($inventory->contains($arrow) or ($inventory = $player->getInventory())->contains($arrow))){// Receives a Mismatch Transaction when using offhand arrows. why???
```
dktapps, line 51: This condition is quite confusing.

src/item/Bow.php
Rush2929: This is because if a player who was in creative mode at the time of 4d1f18e changes the game mode with EntityShootBowEvent, the inventory object may access the $inventory variable to which it has not been assigned. I intended to remove this as it is no longer needed in the 9524d8e but I forgot it

Rush2929:
Hmmm... When a player in creative mode uses offhand arrows and changes to survival mode with EntityShootBowEvent, the slots are not synchronized on the client side. I believe this has to do with # 4231. However, this is a bug that only occurs when using the plugin, so you probably don't need to worry about it.

src/item/Bow.php
```diff
@@ -46,7 +46,13 @@ public function getMaxDurability() : int{
...
 	public function onReleaseUsing(Player $player) : ItemUseResult{
 		$arrow = VanillaItems::ARROW();
-		if($player->hasFiniteResources() and !$player->getInventory()->contains($arrow)){
+		$inventory = match(true){
+			$player->getOffHandInventory()->contains($arrow) => $player->getOffHandInventory(),
+			$player->getInventory()->contains($arrow) => $player->getInventory(),
+			default => null
+		};
+
+		if($player->hasFiniteResources() and $inventory === null){// Receives a Mismatch Transaction when using offhand arrows. why???
```
dktapps, line 55: This comment can be removed. It's not something that can be fixed by this part of the code anyway.

dktapps:
Merged, thanks.

PR #4366, opened 2021-08-13
Terminal: fix some shell command execution

### LINKED ISSUE #4365: Terminal will crash in some special enviroment
# Enviroment
OS: MacOS(Catalina)
Terminal: bash(embed)
PocketMine-MP: 2782af7e22b2c25b79cca15587233b6eadd854ab

Since 2782af7e22b2c25b79cca15587233b6eadd854ab use typed property, `tput xxx`  may return string|null but not only string, and it will crash the server, and `tput sitm` return null in my system(terminal).

Blackjack200:
## Introduction
sometimes tput return null back, and because PM4 use typed property in Terminal now, so it will leading a crash.
### Relevant issues
* Fixes #4365 
### API changes
no
### Behavioural changes
very small impact(can ignore)
## Backwards compatibility
no

## Tests
use this pr, start up a server to reproduce crash, the crash disappeared, i think that it just like php behavior before Terminal use typed property (null will convert null to string(0) "" in string concat)

### DIFF
--- a/src/utils/Terminal.php
+++ b/src/utils/Terminal.php
@@ -109,7 +109,7 @@ abstract class Terminal{
 	}
 
 	protected static function executeShell(string $command) : string{
-		return (string) shell_exec($command);	
+		return (string) shell_exec($command);
 	}
 
 	protected static function getEscapeCodes() : void{
--- a/src/utils/Terminal.php
+++ b/src/utils/Terminal.php
@@ -108,42 +108,46 @@ abstract class Terminal{
 		self::$COLOR_WHITE = "\x1b[38;5;231m";
 	}
 
+	protected static function executeShell(string $command) : string{
+		return (string) shell_exec($command);	
+	}
+
 	protected static function getEscapeCodes() : void{
-		self::$FORMAT_BOLD = `tput bold`;
-		self::$FORMAT_OBFUSCATED = `tput smacs`;
-		self::$FORMAT_ITALIC = `tput sitm`;
-		self::$FORMAT_UNDERLINE = `tput smul`;
+		self::$FORMAT_BOLD = self::executeShell('tput bold');
+		self::$FORMAT_OBFUSCATED = self::executeShell('tput smacs');
+		self::$FORMAT_ITALIC = self::executeShell('tput sitm');
+		self::$FORMAT_UNDERLINE = self::executeShell('tput smul');
 		self::$FORMAT_STRIKETHROUGH = "\x1b[9m"; //`tput `;
 
-		self::$FORMAT_RESET = `tput sgr0`;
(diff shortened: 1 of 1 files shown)


dktapps:
That's interesting. I wonder why PHPStan doesn't detect this.

src/utils/Terminal.php
```diff
@@ -108,42 +108,46 @@ protected static function getFallbackEscapeCodes() : void{
...
-			self::$COLOR_GRAY = $colors >= 256 ? `tput setaf 145` : `tput setaf 7`;
-			self::$COLOR_DARK_GRAY = $colors >= 256 ? `tput setaf 59` : `tput setaf 8`;
-			self::$COLOR_BLUE = $colors >= 256 ? `tput setaf 63` : `tput setaf 12`;
-			self::$COLOR_GREEN = $colors >= 256 ? `tput setaf 83` : `tput setaf 10`;
-			self::$COLOR_AQUA = $colors >= 256 ? `tput setaf 87` : `tput setaf 14`;
-			self::$COLOR_RED = $colors >= 256 ? `tput setaf 203` : `tput setaf 9`;
-			self::$COLOR_LIGHT_PURPLE = $colors >= 256 ? `tput setaf 207` : `tput setaf 13`;
-			self::$COLOR_YELLOW = $colors >= 256 ? `tput setaf 227` : `tput setaf 11`;
-			self::$COLOR_WHITE = $colors >= 256 ? `tput setaf 231` : `tput setaf 15`;
+			self::$COLOR_BLACK = (string) ($colors >= 256 ? `tput setaf 16` : `tput setaf 0`);
```
dktapps, line 126: why not using executeShell for this too?

Blackjack200: because this is complex expression, we don't need to force cast them to string twice

dktapps:
I'm actually wondering why we have this `tput` mess at all. It's doing exactly the same thing that `getFallbackEscapeCodes` is doing, but in a more complicated way. In addition, it sticks with the colour scheme of the first terminal to open the CLI, which can cause colour breakage (e.g. ssh from xterm vs xterm256 shell still uses 8-color terminal).

Blackjack200:
> I'm actually wondering why we have this `tput` mess at all. It's doing exactly the same thing that `getFallbackEscapeCodes` is doing, but in a more complicated way. In addition, it sticks with the colour scheme of the first terminal to open the CLI, which can cause colour breakage (e.g. ssh from xterm vs xterm256 shell still uses 8-color terminal).

time to use a php ansi library like https://github.com/bramus/ansi-php ?

dktapps:
No, that's overkill. But I am thinking we could probably just remove the `getEscapeCodes()` function and rely exclusively on `getFallbackEscapeCodes()`.

PR #4368, opened 2021-08-13
Effect: implement slowfall

Blackjack200:
## Introduction

Implement vanilla snowfall effect(according to Minecraft Java Edition 1.16.2 source)

### API changes

[+] Entity::getGravity() : float

This method return the gravity that need in caculate (default $this->gravity for compactibility)

Reason:
 https://github.com/Blackjack200/minecraft_client_1_16_2/blob/c7f87b96efaeb477d9604354aa23ada0eb637ec6/net/minecraft/world/entity/LivingEntity.java#L1795-L1796

### Behavioural changes

The living entity will fall slowly when get slowfall effect

## Backwards compatibility

the plugins that override tryChangeMovement or manipluate gravity.

## Follow-up

I think we don't have a system to implement this very well

## Requires translations:

```ini
potion.slowFalling=Slow Falling
```

(according to Minecraft Bedrock Edition 1.17.11 language file en_US.lang)

## Tests
1. Player get slow fall like vanilla did.
2. Living Entities get slow fall like vanilla did(i test with Zombie)
3. The potions let entites get effect correctly
4. Test video: https://youtu.be/g-Ysi97rRsQ

### DIFF
--- a/src/entity/Living.php
+++ b/src/entity/Living.php
@@ -188,6 +188,13 @@ abstract class Living extends Entity{
 		}
 	}
 
+	public function getGravity() : float{
+		if($this->effectManager->has(VanillaEffects::SLOW_FALLING())){
+			return $this->effectManager->get(VanillaEffects::SLOW_FALLING())->getType()->getGravity();
+		}
+		return $this->gravity;
+	}
+
 	public function getMaxHealth() : int{
 		return (int) $this->healthAttr->getMaxValue();
 	}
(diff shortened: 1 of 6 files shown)


src/entity/Living.php
```diff
@@ -188,6 +188,13 @@ public function setHealth(float $amount) : void{
 		}
 	}
 
+	public function getGravity() : float{
+		if($this->effectManager->has(VanillaEffects::SLOW_FALLING())){
+			return $this->effectManager->get(VanillaEffects::SLOW_FALLING())->getType()->getGravity();
```
dktapps, line 193: This can be reduced to `VanillaEffects::SLOW_FALLING()->getGravity()`.

Blackjack200: ok

Blackjack200:
hmmm

PR #4370, opened 2021-08-14
Item: implement record_pigstep

Blackjack200:
## Introduction

as the title

### Behavioural changes

the pigstep record appear in the creative contents. and it can be played correctly

## Backwards compatibility

I think it just not support the plugin which implement this record

## Tests

1. /give player record_pigstep

2. put it in record box (works)


ipad54:
Screenshot_20210814_160700_com android chrome

PR #4373, opened 2021-08-16
BaseInventory->removeItem() now alway check item NamedTag

### LINKED ISSUE #4233: removeItem without a named tag can remove same item with named tag
### Issue description

- Expected result: `Inventory->removeItem()` to remove the item that has no named tag
- Actual result: It removed an enchanted version of the item, not comparing named tag value since the item to remove has none

This is caused by https://github.com/pmmp/PocketMine-MP/blob/f68b9e79e1ebadf6b80641292a14208f6d2919a0/src/inventory/InventoryHelpersTrait.php#L211 only checking if the first item has named tag value, though I'm not sure how to approach it, hence this issue exists.

1. Get two of the same item
2. Enchant one of them (or add anything that results in `hasNamedTag()` to return true)
3. `Inventory->removeItem()` the first item without named tag 

### OS and versions

* PocketMine-MP: 4.0 - pmmp/PocketMine-MP@f68b9e79e1ebadf6b80641292a14208f6d2919a0

NgLamVN:
## Introduction
* Due to original item dont have NamedTag, so removeItem() can remove item that have same type but have different NamedTag. 
### Relevant issues
* Fixes #4233 
## Tests
*  Try to remove item like #4233, now it only remove original item.

### DIFF
--- a/src/inventory/BaseInventory.php
+++ b/src/inventory/BaseInventory.php
@@ -266,7 +266,7 @@ abstract class BaseInventory implements Inventory{
 			}
 
 			foreach($itemSlots as $index => $slot){
-				if($slot->equals($item, !$slot->hasAnyDamageValue(), $slot->hasNamedTag())){
+				if($slot->equals($item, !$slot->hasAnyDamageValue(), true)){
 					$amount = min($item->getCount(), $slot->getCount());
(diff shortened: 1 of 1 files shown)


dktapps:
This behaviour is intended.

PR #4374, opened 2021-08-16
Implement Player -> Entity interaction (revival)

CortexPE:
## Introduction
This PR implements Player to Entity interaction... Which makes ArmorStands, Villager Trading UIs, Operations on Player interact (like on Mineplex) possible without overusing EntityDamageEvent or using a separate Packet Handler for plugins.

Stale Pull Request: #2124

### Relevant issues
Currently, No issues address this but it is a 'TODO' in https://github.com/pmmp/PocketMine-MP/blob/5da90b9530c4100cd7c3971675faea74f27b5438/src/player/Player.php#L1666-L1672

### API changes
This adds Entity::onInteract() and PlayerEntityInteractEvent

### Behavioural changes
None

## Backwards compatibility
Purely API additions, no BC Breaks.

## Follow-up
None

## Tests
This can be tested by registering an entity that uses Entity::onInteract() or by using a plugin and listening for PlayerEntityInteractEvent


src/event/player/PlayerEntityInteractEvent.php
```diff
@@ -0,0 +1,74 @@
...
+	/** @var Item */
+	protected $item;
+
+	/** @var Vector3 */
+	protected $clickPos;
+
+	/** @var int */
+	protected $slot;
+
+	public function __construct(Player $player, Entity $entity, Item $item, Vector3 $clickPos, int $slot){
```
dktapps, line 51: This can use promoted properties since we're targeting PHP 8.0 now.

src/entity/Entity.php
```diff
@@ -1080,6 +1081,13 @@ public function onCollideWithPlayer(Player $player) : void{
 
 	}
 
+	/**
+	 * Called when interacted or tapped by a Player
+	 */
+	public function onInteract(Player $player, Item $item, Vector3 $clickPos, int $slot) : bool{
```
dktapps, line 1087: It's already capable of fetching this information (slot is always getItemInHandIndex()).

CortexPE: perhaps I could just ditch passing item-related info altogether to `Entity::onInteract()` and `PlayerEntityInteractEvent`?

src/event/player/PlayerEntityInteractEvent.php
CortexPE: I'm using a quite outdated version of PHPStorm that doesn't fully cover PHP8, but I assume this is what you're referring to?

```php
class PlayerEntityInteractEvent extends PlayerEvent implements Cancellable{
	use CancellableTrait;

	public function __construct(
		protected Player $player,
		protected Entity $entity,
		protected Vector3 $clickPos
	){}

	public function getEntity() : Entity{
		return $this->entity;
	}

	public function getClickPosition() : Vector3{
		return $this->clickPos;
	}
}
```

dktapps: Yes, except it should be private, not protected.

CortexPE: the `player` property is inherited from the `PlayerEvent`, so I'm once again assuming that it should work like this:
```php
	public function __construct(
		protected Player $player,
		private Entity $entity,
		private Vector3 $clickPos
	){}
```

dktapps: No, the Player parameter shouldn't be promoted. Pass it to the parent constructor.

CortexPE: Like this?

```php
	public function __construct(
		Player $player,
		private Entity $entity,
		private Vector3 $clickPos
	){
		$this->player = $player;
	}
```

ColinHDev: PlayerEvent has no constructor so it should be right

src/entity/Entity.php
```diff
@@ -1080,6 +1080,13 @@ public function onCollideWithPlayer(Player $player) : void{
 
 	}
 
+	/**
+	 * Called when interacted or tapped by a Player
```
dktapps, line 1084: Please also document the return value (an opaque true/false return has very unclear usage).

dktapps (commented):
This looks OK, although slightly lacking in documentation.

dktapps:
Some clarification on interactEntity()'s return value: It's actually used to decide whether or not the held item slot should be synced with the client, which actually doesn't have anything to do with the success of the interaction at all. This is, however, a hamfisted way of handling inventory sync and will change in the future.

For now, I think this is OK.

dktapps:
Is the click vector absolute or relative? This should be documented so that we don't get another confusing mess like we got with PlayerInteractEvent.

SOF3:
time to rename PlayerInteractEvent too? it is now misleading that PlayerInteractEvent generalizes both of them (although it indeed does, but as a separate event instead of a superclass)

PR #4376, opened 2021-08-16
Introduce GitHub Actions to automatically build and upload release artifacts, and to auto-update update.pmmp.io

### LINKED ISSUE #2628: Automate the release flow
Currently the release flow is cumbersome and takes time. The only parts of releasing that should be manual are writing the changelog and pressing the release button.

dktapps:
## Introduction
This is the first step towards phasing out jenkins.pmmp.io. 

### Creating releases
Right now, the release process is very cumbersome:
1. Run `build/make-release.php`
2. Go to jenkins.pmmp.io and manually trigger a PocketMine-MP job build
3. Download PocketMine-MP.phar and friends from Jenkins build results
4. Upload PocketMine-MP.phar to draft release
5. Manually write / copy-paste-and-edit changelog stub (subject to frequent mistakes)
6. Publish release.

Most of the steps in this process are error-prone, and for our efforts we get a split distribution system: some users get their phars from Jenkins, while others get them from GitHub. This also fragments our download statistics.

This PR does the following:
1. When a new tag is created, build optimised PocketMine-MP.phar
2. Create draft release
3. Attach PocketMine-MP.phar and start scripts to draft release

This reduces the amount of user interaction to create a release significantly, while also unifying distribution platforms, making release production less fragile, and also portable (this whole system will work with only minimal changes needed for forks).

#### Existing clusterfuck
Both https://get.pmmp.io and the PocketMine-MP "Auto Updater" (not actually an auto updater,but that's a topic for another time) depend on https://update.pmmp.io, which in turn depends on this API: https://jenkins.pmmp.io/job/PocketMine-MP/Stable/api/json?pretty=true&tree=timestamp,artifacts[fileName],url.

Because this API doesn't provide all the information needed by update.pmmp.io, the Jenkins build also generates this file, which, once we know where to get the build info from, can replace the Jenkins API almost entirely: https://jenkins.pmmp.io/job/PocketMine-MP/Stable/artifact/build_info.json/*view*/

https://update.pmmp.io/api then takes the Jenkins API info, combines it with build_info.json (https://gist.github.com/brandon15811/719e44c22f53f759a929ee5d8387a1f7), and then spits it out for public consumption.

#### With this PR
`release-artifacts.yml` generates a `build_info.json` using this script: https://github.com/pmmp/PocketMine-MP/blob/gh-actions-release-artifacts/build/generate-build-info-json.php
This file is then attached to the draft release created.

When the release is published, the `update-updater-api.yml` workflow runs, simply grabs the `build_info.json` from the release artifacts, and commits it to https://github.com/pmmp/update.pmmp.io.

https://update.pmmp.io/api serve script then becomes significantly simpler, since it just needs to grab the correct file from GitHub (perhaps we can even just have Cloudflare rewrite the URLs directly, but /api is kept for compatibility for now).

### Relevant issues
Fixes #2628

### API changes
Added `pocketmine\BUILD_CHANNEL` constant. This isn't expected to be used by plugins, and is currently used to pass information about the release to the updater API. There might be a better way to do this, so don't expect it to stick.

### Behavioural changes
None.

## Backwards compatibility
Fully backwards compatible.
Special care was taken to ensure that the build_info.json produced is compatible with both https://get.pmmp.io and the PM AutoUpdater, so that https://update.pmmp.io/api can be seamlessly transitioned to the new distribution system.

## Follow-up
Currently, the action will attempt to build a release for every tag created. This might be problematic for cases where tags are being made purely for historical purposes, but since the releases anyway require human intervention to publish (only drafts are created), I don't consider this a major issue.

https://get.pmmp.io should be amended to use the `source_url` field of `build_info.json` when fetching artifacts such as `README.md`, `CONTRIBUTING.md`, and `LICENSE`.

## Tests
This has been thoroughly tested on my fork: https://github.com/dktapps/PocketMine-MP.
An example auto-update of the updater API can be seen here: https://github.com/pmmp/update.pmmp.io/commit/e16d1924806cd2bbe231fb2ab2066295bc237da5

### DIFF
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -1,12 +1,12 @@
-name: Create and upload release artifacts
+name: Draft release
 
 on:
   push:
     tags: "*"
 
 jobs:
-  artifacts:
-    name: Create release artifacts
+  draft:
+    name: Create GitHub draft release
     if: "startsWith(github.event.head_commit.message, 'Release ')"
     runs-on: ubuntu-20.04
     strategy:
--- a/.github/workflows/release-artifacts.yml
+++ b/.github/workflows/release-artifacts.yml
@@ -81,7 +81,7 @@ jobs:
           echo ::set-output name=PM_VERSION_MD::$(php -r 'require "vendor/autoload.php"; echo str_replace(".", "", \pocketmine\BASE_VERSION);')
 
       - name: Generate build info
-        run: php build/generate-build-info-json.php ${{ github.sha }} ${{ steps.get-pm-version.outputs.PM_VERSION }} > build_info.json
+        run: php build/generate-build-info-json.php ${{ github.sha }} ${{ steps.get-pm-version.outputs.PM_VERSION }} ${{ github.repository }} > build_info.json
 
       - name: Upload release artifacts
         uses: actions/upload-artifact@v2
--- a/.github/workflows/release-artifacts.yml
+++ b/.github/workflows/release-artifacts.yml
@@ -128,9 +128,9 @@ jobs:
         with:
           name: release_artifacts
           path: |
-          ${{ github.workspace }}/PocketMine-MP.phar
-          ${{ github.workspace }}/start.*
-          ${{ github.workspace }}/build_info.json
+            ${{ github.workspace }}/PocketMine-MP.phar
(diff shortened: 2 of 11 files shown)


dktapps:
This seems to be ready to merge. Everything seems to be working correctly in fork testing.

PR #4381, opened 2021-08-18
Update bow sound

Pab450:
## Introduction
When the durability is exhausted, the bow does not emit Item Break Sound. 

## Follow-up
Make sure the breakable objects in the game make a sound when they break.

## Tests
Use a bow until it breaks.


dktapps:
I think a more general solution for this would be better (the same problem will occur with crossbows when they get implemented, and other things with release actions).

dktapps:
Superseded by https://github.com/pmmp/PocketMine-MP/commit/bf8b5905e4a02c6267d06ec26d3b1db26f96c680. Thanks for bringing this to light.

PR #4384, opened 2021-08-19
remove InventoryPickup*Event, add EntityItemPickupEvent

rasu3n:
## Introduction

https://github.com/pmmp/PocketMine-MP/pull/4384#issuecomment-901952181
### API changes

InventoryPickupArrowEvent has been removed. Use the EntityItemPickupEvent instead.
InventoryPickupItemEvent has been removed. Use the EntityItemPickupEvent instead.
## Tests

```php
public function onItemPickup(EntityItemPickupEvent $event) : void {
  $entity = $event->getEntity();
  if (!($entity instanceof Player)) {
    return;
  }

  $entity->sendMessage(get_debug_type($event->getPickedUp()));
  if ($entity->isSneaking()) {
    $event->cancel();
  }
}
```


dktapps:
This is no less confusing, since now a null inventory is picking up an item, instead of a full one.

Rush2929:
What am I supposed to do?

dktapps:
If you ask me, the whole event should be scrapped. It doesn't make any sense.
It would be better replaced by `EntityItemPickupEvent` and `BlockItemPickupEvent`.

dktapps:
I think it would be a good idea to differentiate between the picked up item and the entity it comes from. If we handle both dropped items and arrow entities in the same event, it's not immediately clear what item will end up being added to the inventory. Then you end up needing boilerplate code with `instanceof` to get the resulting item.

dktapps:
The event should also include a "destination inventory", which can optionally be null.

src/event/entity/EntityPickupItemEvent.php
```diff
@@ -36,13 +38,43 @@ class EntityPickupItemEvent extends EntityEvent implements Cancellable{
...
-		return $this->pickedUp;
+	public function getOrigin() : Entity{
+		return $this->origin;
+	}
+
+	/**
+	 * Items to be received
+	 */
+	public function getItem() : Item{
+		return $this->item;
```
dktapps, line 56: This should return a clone.

src/event/entity/EntityPickupItemEvent.php
```diff
@@ -36,13 +38,43 @@ class EntityPickupItemEvent extends EntityEvent implements Cancellable{
...
+	}
+
+	/**
+	 * Change the items to receive.
+	 */
+	public function setItem(Item $item) : void{
+		$this->item = $item;
```
dktapps, line 63: This should clone the parameter.

src/entity/projectile/Arrow.php
```diff
@@ -194,7 +194,7 @@ public function onCollideWithPlayer(Player $player) : void{
 			$viewer->getNetworkSession()->onPlayerPickUpItem($player, $this);
 		}
 
-		$playerInventory->addItem(clone $item);
+		$ev->getInventory()->addItem(clone $ev->getItem());
```
dktapps: This should be cloned at the call site.

Also, the inventory might be null, so this needs to be checked before trying to do anything with it.

src/entity/projectile/Arrow.php
```diff
@@ -174,13 +174,15 @@ public function onCollideWithPlayer(Player $player) : void{
 		}
 
 		$item = VanillaItems::ARROW();
-
 		$playerInventory = $player->getInventory();
-		if($player->hasFiniteResources() and !$playerInventory->canAddItem($item)){
-			return;
+		if(!$playerInventory->canAddItem($item)){
+			if($player->hasFiniteResources()){
+				return;
```
dktapps, line 180: I think it would be better to pre-cancel the event instead of hardcoding this behaviour. This would allow plugins to customise what happens when a player tries to pickup an item with a full inventory (e.g. sending it directly to their ender chest inventory instead).

src/entity/object/ItemEntity.php
```diff
@@ -227,12 +227,14 @@ public function onCollideWithPlayer(Player $player) : void{
 
 		$item = $this->getItem();
 		$playerInventory = $player->getInventory();
-
-		if($player->hasFiniteResources() and !$playerInventory->canAddItem($item)){
-			return;
+		if(!$playerInventory->canAddItem($item)){
+			if($player->hasFiniteResources()){
+				return;
```
dktapps, line 232: Same again here.

dktapps (approved):

PR #4385, opened 2021-08-19
Implement proper 32x64 skin workaround

CortexPE:
## Introduction
This pull request introduces a workaround to fix the broken rendering of 32x64 skins in MC:Bedrock 1.17 clients formerly without applying a [proper conversion process](https://imgur.com/a/hfaqL) ([source](https://www.reddit.com/r/Minecraft/comments/1vfugk/guide_a_better_guide_to_updating_your_skin_to_the/)), the skins may look glitched or may be [missing the right arm and the right leg](https://twitter.com/dktapps/status/1404067959253909506?s=20).

### Behavioural changes
Skins will now be correctly converted with parts mirrored for the 64x64 skin format, and the randomly colored placeholder skin for "persona" skins will now also be rendered properly on 1.17 clients

## Tests
With:
image

Without:
image

### DIFF
--- a/src/network/mcpe/convert/LegacySkinAdapter.php
+++ b/src/network/mcpe/convert/LegacySkinAdapter.php
@@ -32,8 +32,10 @@ use function is_string;
 use function json_decode;
 use function json_encode;
 use function json_last_error_msg;
+use function str_pad;
 use function random_bytes;
 use function str_repeat;
+use function strlen;
 
 class LegacySkinAdapter implements SkinAdapter{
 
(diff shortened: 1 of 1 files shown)


dktapps:
It occurs to me that this translation really ought to be done when sending the skin, not when receiving it. It's the client that has a problem, not the server.

It would likely also be useful to have this as an API method outside of the network layer in case plugins want to convert their own skins (or for writing unit tests).

dktapps:

PR #4386, opened 2021-08-19
implement temporary bans

cooldogedev:
## Introduction

This PR aims to implement the missing temporary bans.

### API changes
```
public function addBan(string $target, ?string $source, ?string $reason = null, ?string $creationDate = null, ?string $expirationDate = null, bool $save = true) : BanEntry;
```

## Backwards compatibility

This PR is BC-Breaking as the addBan parameters have been changed and the ban formats have been changed, meaning that any previous bans/ip bans will not load up and might cause a ceash upon loading the ban list. 

### Requires translations:

It does not need new translations, however a new translation key can be added to indicate the duration of the ban.

## Tests

`/ban <player> [reason] [date: 1day]`
`/ban-ip <ip> [reason] [date: 1day]`

### DIFF
--- a/src/command/defaults/BanCommand.php
+++ b/src/command/defaults/BanCommand.php
@@ -23,16 +23,17 @@ declare(strict_types=1);
 
 namespace pocketmine\command\defaults;
 
+use Exception;
 use pocketmine\command\Command;
 use pocketmine\command\CommandSender;
 use pocketmine\command\utils\InvalidCommandSyntaxException;
 use pocketmine\lang\KnownTranslationFactory;
 use pocketmine\lang\KnownTranslationKeys;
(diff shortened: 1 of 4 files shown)


dktapps:
This PR includes way too many superfluous changes that aren't necessary for what the PR wants to do.

jasonwynn10:
Can you make it so ban reasons aren't restricted to 1-word entries?

cooldogedev:
> Can you make it so ban reasons aren't restricted to 1-word entries?

That's not possible without swapping the reason with the ban duration, I'll probably look into rewriting the ban commands later.

dktapps:
A separate command for temporary bans is probably a better idea.

PR #4387, opened 2021-08-20
New name for MCBE

JuqnDev:
This I no longer know can be called Pocket Edition


dktapps:
Thanks for the PR. Since there were a few other places this needed to be changed, and this PR targets the wrong branch, I made the changes directly and committed them myself.

JuqnDev:
> Thanks for the PR. Since there were a few other places this needed to be changed, and this PR targets the wrong branch, I made the changes directly and committed them myself.

In both branches it contains "Pocket Edition" and I think we should change it.

dktapps:
The changes will be merged from stable into master.

PR #4393, opened 2021-08-21
Misleading Chunk methods in World class

ColinHDev:
## Introduction
As pointed out by @dktapps via. discord (https://discord.com/channels/373199722573201408/373214753147060235/878764690526830592) the method getChunks() in the World class which returns all loaded chunks from that world is misleading.

### Relevant issues
https://discord.com/channels/373199722573201408/373214753147060235/878764690526830592

### API changes
The method name was changed from getChunks() to getLoadedChunks() so it is less misleading and indicates that not every existing but just every currently loaded chunk will be returned.

## Backwards compatibility
Any plugins that use this method would need to rename it.

## Tests
The method is only used in the GarbageCollectorCommand class and the StatusCommand class as can be seen through PhpStorm's and GitHub's usage search and should therefore not require any additional tests because just the name was changed and not the method itself
image
image

### DIFF
--- a/src/command/defaults/GarbageCollectorCommand.php
+++ b/src/command/defaults/GarbageCollectorCommand.php
@@ -54,10 +54,10 @@ class GarbageCollectorCommand extends VanillaCommand{
 		$memory = memory_get_usage();
 
 		foreach($sender->getServer()->getWorldManager()->getWorlds() as $world){
-			$diff = [count($world->getChunks()), count($world->getEntities())];
+			$diff = [count($world->getLoadedChunks()), count($world->getEntities())];
 			$world->doChunkGarbageCollection();
 			$world->unloadChunks(true);
-			$chunksCollected += $diff[0] - count($world->getChunks());
+			$chunksCollected += $diff[0] - count($world->getLoadedChunks());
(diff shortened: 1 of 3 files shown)


ColinHDev:
There also exists the getChunk() method which is misleading as the getChunks() method.
But it is defined in the ChunkManager interface and more often used than the getChunks() method so changing it would result in a big PR
image
image

VixikHD:
This may not be related, but [World->getChunkPlayers()](https://github.com/pmmp/PocketMine-MP/blob/master/src/world/World.php#L579-L589) method has also misleading name.

dktapps:
You can also just right click -> refactor -> rename

dktapps:
I think a rename of the other methods should be a topic for a separate PR, since that's a much more involved change thanks to ChunkManager.

ColinHDev:
so renaming getChunk to getLoadedChunk and leaving setChunk how it is?

dktapps:
No, this PR is enough by itself. Any further changes should be discussed independently, since they have much more far-reaching effects.

ColinHDev:
I know I was referring to a new PR

PR #4395, opened 2021-08-22
Refactor Block & Tile: getPos() to getPosition()

SalmonDE:
## Introduction

The purpose of this pull request is to establish more consistency in the naming scheme of position getters in PM4 by adjusting them in the Block and Tile classes, as well as their usages within the codebase, in order to correspond to the getPosition() method in the Entity class as well as the name of the Position class.

Furthermore, the class property 'pos' has been renamed to 'position' in hope of reducing additional guesswork for developers and as stated already, for the sake of consistency. However, the scope of this pull request hasn't been broadened up to local variables, hence they remain unchanged.

### Relevant issues

I am not aware of any relevant issues.

### API changes

Changed method signatures:
- `Block->getPos()` -> `Block->getPosition()`
- `Block->getPosOffset()` -> `Block->getPositionOffset()`
- `Tile->getPos()` -> `Tile->getPosition()`

Changed class properties:
- `Block->pos` -> `Block->position`
- `Tile->pos` -> `Tile->position`

### Behavioural changes

No behavioural changes are intended nor expected.

## Backwards compatibility

All API changes as stated above are heavily backwards-incompatible and will cause many plugins to break. I'd argue the impact on PM3 plugins is acceptable since they would have faced breaking changes anyway, making the additional cost to maintain consistency negligible.
On the other hand, it might not be worth it to break existing plugins for PM4 over this change.
For this reason, I'm also still ambivalent about this pull request and kindly ask for more opinions on this matter.

## Follow-up

No follow-ups planned so far.

## Tests

Besides manually going through the changes via git and testing various gameplay mechanics within the game no further intensive testing has been done. I'd be glad for more pairs of eyes to take a look at the changes and point out any flaws to me.

### DIFF
--- a/src/block/Anvil.php
+++ b/src/block/Anvil.php
@@ -78,7 +78,7 @@ class Anvil extends Transparent implements Fallable{
 
 	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
 		if($player instanceof Player){
-			$player->setCurrentWindow(new AnvilInventory($this->pos));
+			$player->setCurrentWindow(new AnvilInventory($this->position));
 		}
 
 		return true;
--- a/src/block/Bamboo.php
+++ b/src/block/Bamboo.php
@@ -114,8 +114,8 @@ class Bamboo extends Transparent{
 		return 12 + (self::getOffsetSeed($x, 0, $z) % 5);
 	}
 
-	public function getPosOffset() : ?Vector3{
-		$seed = self::getOffsetSeed($this->pos->getFloorX(), 0, $this->pos->getFloorZ());
+	public function getPositionOffset() : ?Vector3{
+		$seed = self::getOffsetSeed($this->position->getFloorX(), 0, $this->position->getFloorZ());
 		$retX = (($seed % 12) + 1) / 16;
 		$retZ = ((($seed >> 8) % 12) + 1) / 16;
 		return new Vector3($retX, 0, $retZ);
@@ -133,9 +133,9 @@ class Bamboo extends Transparent{
 	}
 
 	private function seekToTop() : Bamboo{
-		$world = $this->pos->getWorld();
+		$world = $this->position->getWorld();
 		$top = $this;
-		while(($next = $world->getBlock($top->pos->up())) instanceof Bamboo && $next->isSameType($this)){
+		while(($next = $world->getBlock($top->position->up())) instanceof Bamboo && $next->isSameType($this)){
 			$top = $next;
 		}
 		return $top;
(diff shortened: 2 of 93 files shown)


ColinHDev:
It is just the name that is changed. So to make your plugin compatible, you would only need to change the method name 乁( •_• )ㄏ

ColinHDev:
But I think that change should be done in favour of consistency.
It is weird that to get the position of a block we use Block::getPos() but to get the position of an entity we use Entity::getPosition()

dktapps:
It might be less impactful (actually, a lot less) to change `Entity::getPosition()` instead.

ColinHDev:
Are Abbreviations the way to go? I mean we never use them (apart from the AABB class but it would be way to long to write its full name):
getVector3() instead of getVec3() 
getLocation() instead of getLoc()
Why do it than for the Position class if not for Vector, Location, or other math related class?

dktapps:
You can already see from the diffs that this change makes a lot of hard-to-read code even harder as it is, without conveying any new information.

dktapps:
Everything in Liquid.php speaks for itself.

SalmonDE:
I had considered changing the signature of `Entity::getPosition()`, but for similar reasons as @ColinHDev pointed out I was hesitant and instead chose the other route. A potential change to `Entity::getPos()` would make it stand out from these other signatures:

- `Entity::getOffsetPosition()`
- `Entity::setPosition()`
- `Entity::setPositionAndRotation()`

Of course, these can be changed just as easily, however, there are a few other cases that undermine the point of this adjustment:

- `Position::asPosition()`
- `Block::position()`

And although this might be far-fetched, there's also `PlayerRespawnEvent::getRespawnPosition()` as well as the setter method.

I personally think changing these methods to a shorter version is overkill. At the same time though, this pull request initially tried to generate more consistency.
My dilemma with adopting `Entity::getPos()` is to sacrifice some consistency and ignore other method signatures or to change quite a few of them and their usages, probably generating a moderate diff in the process, too. And this excludes the consideration of other classes as well as their setters/getters throughout the codebase, such as `Location` and `Vector3` (and I guess technically the `Position` class itself, as its name isn't `Pos`), which @ColinHDev noted already.

Having that in mind, I do think the current changes made by the pull request are justifiable, by taking into account, that while longer code might worsen it, the readability in most parts remains on a comparable level as it was previously.

ColinHDev:
> Everything in Liquid.php speaks for itself.

The change from pos to position is a diff of +5 characters. If the code is bad, we won't make it significantly worse by extending some variables by five characters, at least in my opinion

dktapps:
That's fair. Originally when this was added it wasn't supposed to be permanent (blocks won't contain positions in future), but temporary things end up being permanent anyway.

PR #4396, opened 2021-08-23
Add Swim, Glide & BoundingBox correction

Crasher508:
## Introduction
This PR can add Swimming, Gliding and corrected the boundingBox. Also it's exhaust on swim and non damage on glide. 

## Tests
Minecraft 23 08 2021 07_29_31
Minecraft 23 08 2021 07_29_54

### DIFF
--- a/src/entity/Entity.php
+++ b/src/entity/Entity.php
@@ -328,6 +328,17 @@ abstract class Entity{
 		);
 	}
 
+	/**
+	 * Update entity's height and width
+	 *
+	 * @param float $height
+	 * @param float $width
+	 */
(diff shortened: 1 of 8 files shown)


dktapps:
This PR is doing too many things at once. As a rule of thumb, if you find yourself needing "and" or "also" to describe what a PR/commit is doing, it's doing too much and should be split up.

In addition, this doesn't solve the problem that originally arose with these features, which is how to deal with custom bounding box sizes. Not all entities are 1.8x0.6.

inxomnyaa (changes requested):

dktapps:

PR #4398, opened 2021-08-23
Workaround for #4291 fall damage bug

### LINKED ISSUE #4291: PM4 Fall damage bug
### PM4 Fall damage bug.

- Expected result: Damage taken on contact with surface.
- 
- Actual result: Damage taken when player hits wall while falling, after the damage has been dealt player can finish falling down without taking any damage when hitting the ground.

1. Build a wall high enough for damage to be dealt. 
2. While falling force yourself onto the wall you are falling next to. On mcbe win10 use “W” key to push yourself onto the wall. On pe use forward button.

### OS and versions
* PocketMine-MP: master 444d902990e00125afb6f1182d95cc0543b134c3

* PHP: 8.0
* Server OS: Ubuntu 18.04
* Game version: PE and Win10

Plugins:
The issue is reproducible without plugins.

Video demonstration: https://www.youtube.com/watch?v=7pZ5DPNK92k&feature=youtu.be

CortexPE:
## Introduction
This Pull Request simply allows for a `0.0001` margin of error on the sides of the player's bounding box when the server checks for the blocks colliding with the player

### Relevant issues
* Fixes #4291

### Behavioural changes
When the server checks for the player's ground state, if the player is not in spectator mode, it will now trim the sides of the bounding box by `0.0001` blocks, in order to account for possible floating point precision errors

## Tests
Without Margin:
https://user-images.githubusercontent.com/17762324/130431606-dfb4d32b-70a6-4293-a161-6fa7ddbd6860.mp4

With margin;
https://user-images.githubusercontent.com/17762324/130431594-8c0036bb-60d5-4435-9ce4-2d9e03255bd9.mp4

### DIFF
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -1064,6 +1064,7 @@ class Player extends Human implements CommandSender, ChunkListener, IPlayer{
 			$bb = clone $this->boundingBox;
 			$bb->minY = $this->location->y - 0.2;
 			$bb->maxY = $this->location->y + 0.2;
+			$bb->contract(0.00001, 0, 0.00001); // account for floating point precision errors
 
 			$this->onGround = $this->isCollided = count($this->getWorld()->getCollisionBlocks($bb, true)) > 0;
 		}
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -1064,7 +1064,7 @@ class Player extends Human implements CommandSender, ChunkListener, IPlayer{
 			$bb = clone $this->boundingBox;
 			$bb->minY = $this->location->y - 0.2;
 			$bb->maxY = $this->location->y + 0.2;
-			$bb->contract(0.00001, 0, 0.00001); // account for floating point precision errors
+			$bb->contract(0.0001, 0, 0.0001); // account for floating point precision errors
 
 			$this->onGround = $this->isCollided = count($this->getWorld()->getCollisionBlocks($bb, true)) > 0;
 		}


dktapps:
What I don't understand is why this bug only appears on PM4, if this is the solution.

TheNewHEROBRINEX:
Shouldn't this be done by using the `$epsilon` parameter of [`AxisAlignedBB::intersectWith`](https://github.com/pmmp/Math/blob/0.3.0/src/AxisAlignedBB.php#L351)? It already defaults to `0.00001` but apparently it's not enough in this case.

dktapps:
^ this is why I'm not convinced that this workaround actually fixes anything.

CortexPE:
> 
> 
> Shouldn't this be done by using the `$epsilon` parameter of [`AxisAlignedBB::intersectWith`](https://github.com/pmmp/Math/blob/0.3.0/src/AxisAlignedBB.php#L351)? It already defaults to `0.00001` but apparently it's not enough in this case.

Perhaps a less invasive workaround is to simply adjust the epsilon default value to `0.0001`? Which is changing the default margin from one hundred-thousandths of 1 block to one ten-thousandths of 1 block... It's a 10x difference

dktapps:
Testing shows that what's causing this can be a rounding error in the order of 1/10,000. On PM3 this isn't a problem because the incoming client positions are rounded to 4dp anyway, but that rounding disappeared in this commit: https://github.com/pmmp/PocketMine-MP/commit/82c8fa696a2e8609d8ae086c52f96abe7c7705b6

dktapps:
Fixed by 7701e1ff988af54eee167bc3e2281762b6c017d7.

Issue #4399, opened 2021-08-23
BlockItemPickupEvent


ColinHDev:
## Description
The BlockItemPickupEvent would be an event under the pocketmine\event\block namespace and would extend the BlockEvent and implement the Cancellable interface. The contructor could be nearly the same as the constructor of EntityItemPickupEvent except its first parameter, $collector, which needs to be of type Block instead of Entity.
It would be called when a block picks up an item, arrow, etc.

## Justification
With the remove of the InventoryPickupEvent and the addition of the EntityItemPickupEvent through eb9188c30958c8efa4ad8acf1126712a4013d4ca it became no longer possible to implement that blocks which hold inventories can pickup items, for example in case of the Hopper.
To allow this addition in a future PR a BlockItemPickupEvent would be needed to implement.

ColinHDev:
I know, if this event would be implemented, it would be dead code because at the moment, it would never be called.
But since PRs shouldn't be filled with too many changes, the event would be needed to make a PR for Hoppers. (Otherwise we would have Hoppers and the Event in the PR and that could be considered as overfilled.)

ColinHDev:
Just for clarification:
- I am planning to make a PR to add the event but I first wanted to make sure that this idea is good enough
- I know that a PR would basically just add dead code but I am also planning to make a PR to fully implement Hoppers (right now they are half implemented: basically they are just containers right now without any ability to transport items) and to do so this event would be needed.

dktapps:
To be honest, it would have taken a lot less time and effort to just make the PR directly.

ColinHDev:
yeah.. 
I thought I would save time with creating an issue instead of a PR if you don't like the idea

dktapps:
Implemented in #4402.

PR #4402, opened 2021-08-24
BlockItemPickupEvent

ColinHDev:
## Introduction
With the removal of the InventoryPickupEvent and the addition of the EntityItemPickupEvent through eb9188c30958c8efa4ad8acf1126712a4013d4ca, it became no longer possible to implement event listeners for blocks (that hold inventories) who pick up items, for example in the case of the Hopper.
To allow this addition in a future PR a BlockItemPickupEvent would be needed to implement.

### Relevant issues
* Solves https://github.com/pmmp/PocketMine-MP/issues/4399

### API changes
- Adds the BlockItemPickupEvent which is currently never called. It would be called in the future if blocks pick up items, arrows, etc.

## Follow-up
To avoid leaving this event unused for long a PR that would fully implement Hoppers should be also merged to the master branch.

## Tests
Currently, the BlockItemPickupEvent is never called and is basically just dead code.

### DIFF
--- a/src/event/block/BlockItemPickupEvent.php
+++ b/src/event/block/BlockItemPickupEvent.php
@@ -1,0 +1,60 @@
+<?php
+
+declare(strict_types=1);
+
+namespace pocketmine\event\block;
+
+use pocketmine\block\Block;
+use pocketmine\entity\Entity;
+use pocketmine\event\Cancellable;
+use pocketmine\event\CancellableTrait;
+use pocketmine\inventory\Inventory;
+use pocketmine\item\Item;
+
+/**
+ * Called when a block picks up an item, arrow, etc.
+ */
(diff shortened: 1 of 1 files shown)


dktapps:
> it became no longer possible to implement that blocks that hold inventories can pick up items

Slight nitpick: it's not possible to do if you want to have events. It's not impossible period.

ColinHDev:
oops

ColinHDev:
Rejected because of unprecise pr description lmao 
Should be better now

dktapps:
No, I'm just making sure the record is accurate in case some random person happens across this PR in 5 years time for some reason, so they don't get misled.

ColinHDev:
Yea, it's understandable so no problem.

PR #4407, opened 2021-08-25
Refactor Tree classes

ColinHDev:
## Introduction
The PR https://github.com/pmmp/PocketMine-MP/pull/4354 introduced the StructureGrowEvent which will be called when structures grow multiple blocks at once. Because of how tree growth is implemented (Sapling class and chunk populator share the same tree placement files) the calling of the event when Sapling grows while it won't be called during chunk population was rather hacky implemented with an $callEvent variable in the Tree classes. But because the PR would have become too big, this issue in coding style wasn't addressed there and it was discussed with @dktapps via. discord that a PR would follow which would refactor the Tree classes (https://discord.com/channels/373199722573201408/551141879123542020/877987305195970581).
This PR aims to refactor the Tree classes to fix this issue in the coding style. 

### API changes
- Removed the method  `Tree::growTree(ChunkManager, int, int, int, Random, ?TreeType = null, bool = false) : void`
- Removed the method `Tree::placeObject(ChunkManager, int, int, int, Random, bool= false) : void`
- Added the method `Tree::get(Random, ?TreeType = null) : ?Tree` which returns the corresponding tree to the tree type or null if a tree class was not found for that tree type
- Added the method `Tree::getBlockTransaction(ChunkManager, int, int, int, Random) : ?BlockTransaction` which returns either BlockTransaction containing all the blocks the tree would change upon growing at the given coordinates or null if the tree can't be grown
- Added the method `Sapling::grow() : void` which handles the growth of saplings when using bone meal on them or if they grow during a random tick

## Backwards compatibility
- Any plugins (e.g. world generators) that use PocketMine-MP's implementation of the tree placement will break. To solve breaks in chunk populators following could be changed:
 ```php
// Before the PR 
Tree::growTree($world, $x, $y, $z, $random, $treeType) : void;
```
```php
// After the PR
$tree = Tree::get($random, $this->type);
$transaction = $tree?->getBlockTransaction($world, $x, $y, $z, $random);
$transaction?->apply();
```
## Tests
- This PR was first tested by generating a new world with PocketMine-MP's NORMAL generator and checking for the three tree types this generator populates throughout the world.
  - Oak:
  Screenshot (530)
  - Spruce:
  Screenshot (531)
  - Birch: (both normal birch and superbirch work (right most tree))
  Screenshot (532)
- Afterwards, this PR was tested by placing the saplings of the four implemented tree types (oak, spruce, birch, jungle) and using bone meal on them to make them grow: (Because using bone meal and growing during a random tick now use the same method to place the tree, it isn't necessary to explicitly check if these trees also grow during a random tick)
image

### DIFF
--- a/src/block/Sapling.php
+++ b/src/block/Sapling.php
@@ -110,7 +110,6 @@ class Sapling extends Flowable{
 
 	private function grow() : void{
 		$random = new Random(mt_rand());
-
 		$tree = Tree::get($random, $this->treeType);
 		$transaction = $tree?->getBlockTransaction($this->position->getWorld(), $this->position->getFloorX(), $this->position->getFloorY(), $this->position->getFloorZ(), $random);
 		if($transaction === null){
--- a/src/block/Sapling.php
+++ b/src/block/Sapling.php
@@ -24,6 +24,7 @@ declare(strict_types=1);
 namespace pocketmine\block;
 
 use pocketmine\block\utils\TreeType;
+use pocketmine\event\block\StructureGrowEvent;
 use pocketmine\item\Fertilizer;
 use pocketmine\item\Item;
 use pocketmine\math\Facing;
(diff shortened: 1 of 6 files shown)


dktapps (changes requested):
There are a lot of superfluous changes in this PR that bloat the diff and obscure what's really going on. You should present a diff with the minimum possible amount of changes.

dktapps (approved):

PR #4412, opened 2021-08-29
Fix offhand pickup

Rush2929:
## Introduction

Currently, the offhand check in the PlayerItemPickupEvent is not strictly enforced, so there is a possibility that items will disappear.
## Tests

before:

https://user-images.githubusercontent.com/76860328/131239826-cbc21e26-c15c-420f-870c-239903eda1aa.mp4

after:

https://user-images.githubusercontent.com/76860328/131239828-ab1f2515-32c2-4d29-ada6-7dc609ffee98.mp4

### DIFF
--- a/src/entity/object/ItemEntity.php
+++ b/src/entity/object/ItemEntity.php
@@ -227,7 +227,7 @@ class ItemEntity extends Entity{
 
 		$item = $this->getItem();
 		$playerInventory = match(true){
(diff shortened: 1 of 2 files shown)


dktapps:
Looks OK, although the second commit is just polluting the diff. I couldn't see what was going on in the diff at first. Don't mix unrelated changes together.

PR #4413, opened 2021-08-29
Implement Brewing

VixikHD:
## Introduction

This pull request implements basic Brewing stand functionality.

### Relevant issues

- There is no issue related to this pull request

### API changes

- New `pocketmine\event\inventory\BrewingFuelUseEvent` and `pocketmine\event\inventory\BrewItemEvent` events
- Added methods `BrewingStandInventory->getIngredient(): Item`, `BrewingStandInventory->setIngredient(Item): void`, `BrewingStandInventory->getFuel(): Item` and `BrewingStandInventory->setFuel(Item): void`

### Behavioural changes

- Brewing stand is now implemented

## Backwards compatibility

- None

## Tests
https://streamable.com/4ma0lh


src/block/tile/BrewingStand.php
```diff
@@ -45,11 +53,11 @@ class BrewingStand extends Spawnable implements Container, Nameable{
...
 
 	/** @var int */
-	private $brewTime = 0;
+	private $brewTime;
 	/** @var int */
-	private $maxFuelTime = 0;
+	private $maxFuelTime;
 	/** @var int */
-	private $remainingFuelTime = 0;
+	private $remainingFuelTime;
```
dktapps, line 60: These defaults need to stay. They might not be initialized if the tile wasn't loaded from disk.

src/block/tile/BrewingStand.php
```diff
@@ -83,6 +91,14 @@ protected function writeSaveData(CompoundTag $nbt) : void{
...
 	}
 
+	protected function addAdditionalSpawnData(CompoundTag $nbt) : void{
+		$this->addNameSpawnData($nbt);
+
+		$nbt->setShort(self::TAG_BREW_TIME_PE, $this->brewTime);
+		$nbt->setShort(self::TAG_MAX_FUEL_TIME, $this->maxFuelTime);
+		$nbt->setShort(self::TAG_REMAINING_FUEL_TIME_PE, $this->remainingFuelTime);
+	}
+
```
dktapps, line 101: Are these really necessary?

src/block/tile/BrewingStand.php
```diff
@@ -108,4 +124,125 @@ public function getInventory(){
...
+		$canBrew = $this->canBrew();
+
+		if($this->remainingFuelTime <= 0 and $canBrew and $fuel->equals(VanillaItems::BLAZE_POWDER(), true, false)) {
+			$this->checkFuel($fuel);
+		}
+
+		if($this->remainingFuelTime > 0){
+			if($canBrew){
+				if($this->brewTime == 0){
+					$this->brewTime = 400;
```
dktapps, line 186: This shouldn't be hardcoded. Make a constant.

src/event/inventory/BrewingFuelUseEvent.php
```diff
@@ -0,0 +1,55 @@
...
+
+	public function getBrewingStand() : BrewingStand{
+		return $this->brewingStand;
+	}
+
+	public function getFuelTime() : int{
+		return $this->fuelTime;
+	}
+
+	public function setFuelTime(int $fuelTime) : void{
```
dktapps, line 52: This is undocumented, so nobody will know whether the time is in ticks, or seconds.

dktapps (commented):
I haven't reviewed this in-depth yet, but it's looking interesting.

src/block/tile/BrewingStand.php
VixikHD: Brew time is not really necessary, but others are in order to inform client about state of the fuel when update is not running.

src/event/inventory/BrewingFuelUseEvent.php
dktapps: The new documentation doesn't solve this problem.

src/block/inventory/BrewingStandInventory.php
```diff
@@ -33,4 +34,20 @@ public function __construct(Position $holder, int $size = 5){
 		$this->holder = $holder;
 		parent::__construct($size);
 	}
+
+	public function getIngredient() : Item{
+		return $this->getItem(0);
+	}
+
+	public function setIngredient(Item $item) : void{
+		$this->setItem(0, $item);
```
dktapps, line 43: ugh, of course this is different to PC... thanks for nothing mojang

src/block/tile/BrewingStand.php
```diff
@@ -108,4 +126,125 @@ public function getInventory(){
...
+			if($canBrew){
+				if($this->brewTime == 0){
+					$this->brewTime = self::BREW_TIME;
+					--$this->remainingFuelTime;
+				}
+
+				--$this->brewTime;
+
+				if($this->brewTime <= 0){
+					for($i = 1; $i <= 3; ++$i){
```
dktapps, line 195: This shouldn't be hardcoded. Please update this to use the constants added on the master branch.

src/block/tile/BrewingStand.php
```diff
@@ -108,4 +126,125 @@ public function getInventory(){
...
+
+		$canBrew = $this->canBrew();
+
+		if($this->remainingFuelTime <= 0 and $canBrew and $fuel->equals(VanillaItems::BLAZE_POWDER(), true, false)) {
+			$this->checkFuel($fuel);
+		}
+
+		if($this->remainingFuelTime > 0){
+			if($canBrew){
+				if($this->brewTime == 0){
```
dktapps, line 187:
```suggestion
				if($this->brewTime === 0){
```

src/crafting/PotionContainerChangeRecipe.php
```diff
@@ -0,0 +1,59 @@
...
+	/** @var Item */
+	private $ingredient;
+	/** @var int */
+	private $outputItemId;
+
+	public function __construct(int $inputItemId, Item $ingredient, int $outputItemId) {
+		$this->inputItemId = $inputItemId;
+		$this->ingredient = clone $ingredient;
+		$this->outputItemId = $outputItemId;
+	}
```
dktapps, line 42: This constructor can use promoted properties.

src/crafting/PotionContainerChangeRecipe.php
```diff
@@ -0,0 +1,59 @@
...
+	public function getIngredient() : Item{
+		return clone $this->ingredient;
+	}
+
+	public function getOutputItemId() : int{
+		return $this->outputItemId;
+	}
+
+	public function getOutputFor(Item $input) : ?Item{
+		return $input->getId() == $this->getInputItemId() ? ItemFactory::getInstance()->get($this->getOutputItemId(), $input->getMeta()) : null;
```
dktapps, line 57:
```suggestion
		return $input->getId() === $this->getInputItemId() ? ItemFactory::getInstance()->get($this->getOutputItemId(), $input->getMeta()) : null;
```

src/crafting/PotionTypeRecipe.php
```diff
@@ -0,0 +1,58 @@
...
+	/** @var Item */
+	private $ingredient;
+	/** @var Item */
+	private $output;
+
+	public function __construct(Item $input, Item $ingredient, Item $output) {
+		$this->input = clone $input;
+		$this->ingredient = clone $ingredient;
+		$this->output = clone $output;
+	}
```
dktapps, line 41: This constructor can use promoted properties.

src/crafting/PotionTypeRecipe.php
```diff
@@ -0,0 +1,58 @@
...
+	public function getIngredient() : Item{
+		return clone $this->ingredient;
+	}
+
+	public function getOutput() : Item{
+		return clone $this->output;
+	}
+
+	public function getOutputFor(Item $input) : ?Item{
+		return $input->equals($this->getInput(), true, false) ? $this->getOutput() : null;
```
dktapps, line 56: `getInput()` here is causing an unnecessary clone. Since you're not returning it, you can bypass the getter.

src/event/inventory/BrewItemEvent.php
```diff
@@ -0,0 +1,74 @@
...
+
+	/** @var BrewingRecipe */
+	private $recipe;
+
+	public function __construct(BrewingStand $brewingStand, Item $input, Item $output, BrewingRecipe $recipe){
+		parent::__construct($brewingStand->getBlock());
+		$this->brewingStand = $brewingStand;
+		$this->input = $input;
+		$this->output = $output;
+		$this->recipe = $recipe;
```
dktapps, line 52: This constructor can use promoted properties.

dktapps: This event should include the slot ID of the slot being brewed, so that plugins know which slot the event is occurring on.

src/block/tile/BrewingStand.php
```diff
@@ -108,4 +126,125 @@ public function getInventory(){
...
+
+		$this->maxFuelTime = $this->remainingFuelTime = $ev->getFuelTime();
+	}
+
+	private function canBrew() : bool{
+		if($this->inventory->getIngredient()->isNull()){
+			return false;
+		}
+
+		for($i = 1; $i <= 3; ++$i){
```
dktapps, line 149: This shouldn't be hardcoded.

src/block/tile/BrewingStand.php
```diff
@@ -108,4 +126,125 @@ public function getInventory(){
...
+				continue;
+			}
+
+			if($this->position->getWorld()->getServer()->getCraftingManager()->matchBrewingRecipe($input, $this->inventory->getIngredient()) !== null){
+				return true;
+			}
+		}
+
+		return false;
+	}
```
dktapps, line 161: I think this should return an array of the brewable slots. When I first saw this function, I mistook it for a general "can anything be brewed" method, and that could easily be accidentally used incorrectly in future refactors.

src/block/tile/BrewingStand.php
```diff
@@ -108,4 +126,125 @@ public function getInventory(){
...
+		$prevMaxFuelTime = $this->maxFuelTime;
+
+		$ret = false;
+
+		$fuel = $this->inventory->getFuel();
+		$ingredient = $this->inventory->getIngredient();
+
+		$canBrew = $this->canBrew();
+
+		if($this->remainingFuelTime <= 0 and $canBrew and $fuel->equals(VanillaItems::BLAZE_POWDER(), true, false)) {
```
dktapps, line 181: This blaze powder check should be done inside `checkFuel()` and cause a pre-cancellation of `BrewingFuelUseEvent`, so that plugins can allow different items to be used as fuel.

src/event/inventory/BrewingFuelUseEvent.php
```diff
@@ -0,0 +1,61 @@
...
+
+	/** @var BrewingStand */
+	private $brewingStand;
+	/** @var int */
+	private $fuelTime = 20;
+
+	public function __construct(BrewingStand $brewingStand){
+		parent::__construct($brewingStand->getBlock());
+		$this->brewingStand = $brewingStand;
+	}
```
dktapps, line 42: This constructor can use promoted properties.

dktapps (changes requested):
This looks good so far, but there are a few things that should be addressed.
Also, the formatting is inconsistent in some places - if you use PhpStorm, please reformat your code using the provided `Project` code style.

src/block/tile/BrewingStand.php
VixikHD: Other items could not have been used as fuel, it is blocked by client.

src/crafting/PotionTypeRecipe.php
VixikHD: Even if those items are being cloned?

dktapps: Yes. You can keep the assignments in the constructor body for the cloning, and still get to use promoted properties to avoid the property declarations.

src/block/tile/BrewingStand.php
dktapps: That's beside the point. We don't make API decisions based on what the Mojang client does at a set point in time. The possibility to do this should be there on the server regardless.

src/block/tile/BrewingStand.php
```diff
@@ -108,4 +124,134 @@ public function getInventory(){
...
+			if($canBrew){
+				if($this->brewTime === 0){
+					$this->brewTime = self::BREW_TIME;
+					--$this->remainingFuelTime;
+				}
+
+				--$this->brewTime;
+
+				if($this->brewTime <= 0){
+					foreach([BrewingStandInventory::SLOT_BOTTLE_LEFT, BrewingStandInventory::SLOT_BOTTLE_MIDDLE, BrewingStandInventory::SLOT_BOTTLE_RIGHT] as $slot){
```
dktapps, line 202: This can just foreach over `$recipes` provided above.

src/block/tile/BrewingStand.php
```diff
@@ -108,4 +124,134 @@ public function getInventory(){
...
+						$input = $this->inventory->getItem($slot);
+						if($input->isNull()){
+							continue;
+						}
+
+						$recipe = $this->position->getWorld()->getServer()->getCraftingManager()->matchBrewingRecipe($input, $ingredient);
```
dktapps, line 208: This isn't necessary - you already did it in `getBrewableRecipes()`.

src/crafting/PotionContainerChangeRecipe.php
dktapps: The property can still be declared using the promoted style. You just need to keep the clone assignment in the constructor.

src/block/inventory/BrewingStandInventory.php
```diff
@@ -39,4 +40,20 @@ public function __construct(Position $holder, int $size = 5){
...
+		$this->setItem(self::SLOT_INGREDIENT, $item);
+	}
+
+	public function getFuel() : Item{
+		return $this->getItem(self::SLOT_FUEL);
+	}
+
+	public function setFuel(Item $item) : void{
+		$this->setItem(self::SLOT_FUEL, $item);
+	}
```
dktapps, line 58: These methods are redundant. It's sufficient to just use the constants with setItem().

dktapps (commented):
Please reformat the code using the provided `Project` code style.

dktapps:
I think this is now more or less ready to merge.

I will note that the one thing I'm unhappy about is the exposure of legacy item IDs to the API in `PotionContainerChangeRecipe`. However, this is a hard problem to solve given currently available API, so I don't consider it a showstopper.

PR #4414, opened 2021-08-29
implement dynamic inventory management

HimmelKreis4865:
## Introduction
This PR implements a dynamic inventory management, that simplifies sending and handling of new inventories, that are not yet implemented in pm4.

### Relevant issues
 - There is no issue related to this pull request

### API changes
 - new class `pocketmine\inventory\InventoryFactory`
 - `InventoryFactory::getInstance()->register()` to register a new inventory

## Backwards compatibility
Backward compatible as long as there are no inventories that were previously implemented, which would now need to be registered.

## Tests
Implemented Grindstone successfully:
```php
InventoryFactory::getInstance()->register(GrindstoneInventory::class, WindowTypes::GRINDSTONE, UIInventorySlotOffset::GRINDSTONE);
```
GrindstoneInventory.php
```php
class GrindstoneInventory extends SimpleInventory implements BlockInventory {
	use BlockInventoryTrait;
	
	public function __construct(Position $position, int $size = 3) {
		$this->holder = $position;
		parent::__construct($size);
	}
}
```
In addition, I was still able to open all other inventories

### DIFF
--- a/src/inventory/InventoryFactory.php
+++ b/src/inventory/InventoryFactory.php
@@ -1,0 +1,117 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * @author PocketMine Team
+ * @link http://www.pocketmine.net/
+ *
+ *
+*/
+
+declare(strict_types=1);
+
+namespace pocketmine\inventory;
+
+use Closure;
+use pocketmine\block\inventory\AnvilInventory;
+use pocketmine\block\inventory\BrewingStandInventory;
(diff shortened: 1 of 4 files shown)


dktapps:
This is mixing protocol into the API.

HimmelKreis4865:
But isn't that still better than leaving that hardcoded, where no other inventories can be added? 
Or do you have a better idea for it?

src/inventory/InventoryFactory.php
```diff
@@ -0,0 +1,117 @@
...
+use pocketmine\block\inventory\LoomInventory;
+use pocketmine\crafting\FurnaceType;
+use pocketmine\network\mcpe\protocol\types\inventory\UIInventorySlotOffset;
+use pocketmine\network\mcpe\protocol\types\inventory\WindowTypes;
+use pocketmine\utils\AssumptionFailedError;
+use pocketmine\utils\SingletonTrait;
+use pocketmine\utils\Utils;
+use function array_key_exists;
+
+final class InventoryFactory{
```
dktapps, line 41: This doesn't belong in the `inventory` namespace. `network\mcpe` or `network\mcpe\convert` would be more appropriate.

src/network/mcpe/convert/InventoryFactory.php
```diff
@@ -0,0 +1,119 @@
...
+	public function getCoreOffset(int $netOffset): ?int{
+		if (!array_key_exists($netOffset, $this->netToCoreOffsets)) return null;
+		return $this->netToCoreOffsets[$netOffset];
+	}
+	
+	public function getWindowType(Inventory $inventory): int {
+		if (!array_key_exists($inventory::class, $this->windowTypes)){
+			throw new \UnexpectedValueException("Window type of Inventory " . $inventory::class . " was not registered!");
+		}
+		if (($type = $this->windowTypes[$inventory::class]) instanceof Closure) $type = $type($inventory);
```
dktapps, line 116: This will break subclassed inventories.

dktapps (changes requested):
Overall the concept looks OK. But the formatting needs work, naming is a confusing mess, and this isn't a perfect replacement for the old code.

dktapps:

Issue #4405, opened 2021-08-25
InventoryTransaction's source can't be anything except a player


ColinHDev:
## Description
Currently, the source of an InventoryTransaction and therefore the only reason an InventoryTransactionEvent is called is a Player.
But how seen in the code, the comment at the beginning of the InventoryTransactionEvent class suggests that it could also be called on inventory transactions by non-player entities or hoppers.
https://github.com/pmmp/PocketMine-MP/blob/master/src/event/inventory/InventoryTransactionEvent.php#L33
But how seen in the constructor of the InventoryTransaction class it only accepts players.
https://github.com/pmmp/PocketMine-MP/blob/master/src/inventory/transaction/InventoryTransaction.php#L72

## Justification
A change would be needed to implement that the InventoryTransactionEvent could also be called if no player is involved, for example on an inventory transaction by a hopper.

## Possible Solution
- As the source could both be an entity or a block we could use an in php8 introduced union type to change the data type from `Player $source` to `Entity | Block $source`. However also the data type of most other classes who are related to InventoryTransaction would be needed to change, for example in the InventoryAction classes.

dktapps:
Right now, Inventory transactions only originate from players anyway (they are just a way to make sure players aren't cheating). I think the doc comment should be updated, rather than changing the source information.

ColinHDev:
I am planning to make a PR which fully implements hoppers if this PR would be merged https://github.com/pmmp/PocketMine-MP/pull/4402. 
In preparation I started to look into it and found that comment in the event so I thought that this event should be used for that case. Also because the InventoryTransaction handels the interaction between two inventories which would be the case for hoppers

ColinHDev:
yikes, the comment I was referencing was made in 2014 in 5e97da2e11270d87c950e3c35709b8bc10a00447 when the event was added
sorry, my bad

dktapps:
I've never much liked InventoryTransactionEvent (or transactions in general) and find them overcomplicated. But changing source type to allow non-players would break a whole bunch of assumptions, so it's not the best idea.

ColinHDev:
If you don't like the idea anyway, I won't try to use it
But I still would like to use events on hoppers but no event would suit at the moment. What do you think about the addition of an ItemChangeInventoryEvent or something for that case?

ColinHDev:
The problem could be: Not every block a hopper can push into is a container.
[...]

PR #4416, opened 2021-08-29
Implemented Hopper pushing, pulling and picking up items logic

ColinHDev:
## Introduction
In pm4 hopper blocks were implemented to have an inventory. But the logic for pushing, pulling and picking up items was missing nonetheless.
This PR aims to add this logic to the hopper.
**This PR contains many comments because the hopper logic is very complex. To prevent future developers from getting confused about how certain things were done, most parts were commented to explain what was done and why.**

### Relevant issues
- Fixes https://github.com/pmmp/PocketMine-MP/blob/master/src/block/Hopper.php#L98-L102

### Sources
- https://minecraft.fandom.com/wiki/Hopper
  - https://static.wikia.nocookie.net/minecraft_gamepedia/images/3/38/Hopper_logic_flowchart.png/revision/latest?cb=20210518094503

### API changes
- Added private function in block class `Hopper::push(HopperInventory) : bool` which handles pushing items from the hopper to a tile in the direction the hopper is facing.
- Added private function in block class `Hopper::pull(HopperInventory, Container) : bool` which handles pulling items by the hopper from a container above.
- Added private function in block class `Hopper::pickup(HopperInventory) : bool` which handles picking up items by the hopper.
- Added getter in tile class `Hopper::getTransferCooldown() : int` to get the transfer cooldown of that hopper.
- Added setter in tile class `Hopper::setTransferCooldown(int)` to set the transfer cooldown of that hopper.

### Behavioural changes
- Because hoppers need to be updated every tick, it could impact the server's performance if a large amount of hoppers is loaded.

### TODOs
- Pushing to specific blocks couldn't be implemented because these blocks aren't implemented at the moment. This includes the Composter, the Brewing Stand and the Jukebox (not working as in vanilla).
- Pulling from specific blocks couldn't be implemented because these blocks aren't implemented at the moment. This includes the Composter, the Brewing Stand and the Jukebox.
- Calling an event when a hopper moves an item between its own and another inventory.
- Calling an event when a hopper picks up an item.

## Backwards compatibility
- This PR is fully backwards compatible. Already placed hoppers should also work without any problems when updating to a version of pm4 which merged these changes.

## Tests
- Functionality tests
  - Every block a hopper can push items into was tested if it works as explained in the sources.
  https://www.youtube.com/watch?v=4gSyuViaPaU
  - Every block a hopper can pull items from was tested if it works as explained in the sources.
  https://www.youtube.com/watch?v=6NWvr6Kv88E
  - The collision box of the area where the hopper should pick up items was checked and works as explained in the sources.
  https://www.youtube.com/watch?v=hVEPiK9KWkA

- "Performance" tests
  - 128 hoppers pushing 27 * 64 dirt from one chest to another:
    https://timings.pmmp.io/?id=158627
    Screenshot (543)

### DIFF
--- a/src/block/Hopper.php
+++ b/src/block/Hopper.php
@@ -163,7 +163,7 @@ class Hopper extends Transparent{
 			//TODO: Brewing Stand
 			//TODO: Jukebox (improve)
 			if($destination instanceof TileFurnace){
-				// If the hopper is facing down, it will push every item to the furnace's smelting slot, even items that aren't smeltable.
+				// If the hopper is facing down, it will push every item to the furnace's input slot, even items that aren't smeltable.
 				// If the hopper is facing in any other direction, it will only push items that can be used as fuel to the furnace's fuel slot.
 				if($this->facing === Facing::DOWN){
 					$slotInFurnace = FurnaceInventory::SLOT_INPUT;
--- a/src/block/Hopper.php
+++ b/src/block/Hopper.php
@@ -129,6 +129,7 @@ class Hopper extends Transparent{
 		}
 
 		$success = $this->push($tile);
+		// Hoppers that have a container above them, won't try to pick up items.
 		$origin = $this->position->getWorld()->getTile($this->position->getSide(Facing::UP));
 		if($origin instanceof Container){
 			$success |= $this->pull($tile, $origin);
--- a/src/block/Hopper.php
+++ b/src/block/Hopper.php
@@ -116,13 +116,7 @@ class Hopper extends Transparent{
 			$tile->setTransferCooldown($transferCooldown);
 		}
 
-		if($this->isPowered()){
-			// If a hopper is powered it is deactivated and won't push, pull or pick up items.
-			return;
-		}
-
-		if($transferCooldown > 0){
-			// Skipping this tick because the hopper is still on cooldown of 8 ticks.
+		if($this->isPowered() || $transferCooldown > 0){
 			return;
 		}
 
--- a/src/block/Hopper.php
+++ b/src/block/Hopper.php
@@ -27,7 +27,6 @@ use pocketmine\block\tile\Container;
 use pocketmine\block\tile\Furnace as TileFurnace;
 use pocketmine\block\tile\Hopper as TileHopper;
 use pocketmine\block\tile\Jukebox as TileJukebox;
-use pocketmine\block\tile\ShulkerBox as TileShulkerBox;
 use pocketmine\block\utils\BlockDataSerializer;
 use pocketmine\block\utils\InvalidBlockStateException;
 use pocketmine\block\utils\PoweredByRedstoneTrait;
@@ -229,16 +228,6 @@ class Hopper extends Transparent{
 					$tile->getInventory()->setItem($slot, $item);
 				}
 				return true;
-
-			}elseif($destination instanceof TileShulkerBox){
-				// Hoppers can't push a shulkerbox into another shulkerbox.
-				if($item->getBlock() instanceof ShulkerBox){
-					continue;
-				}
-				$itemToPush = $item->pop();
-				if(!$destination->getInventory()->canAddItem($itemToPush)){
-					continue;
-				}
 
 			}elseif($destination instanceof Container){
 				$itemToPush = $item->pop();
--- a/src/block/Hopper.php
+++ b/src/block/Hopper.php
@@ -227,6 +227,7 @@ class Hopper extends Transparent{
 				if($jukeboxBlock instanceof Jukebox){
 					$jukeboxBlock->insertRecord($item->pop());
 					$jukeboxBlock->getPosition()->getWorld()->setBlock($jukeboxBlock->getPosition(), $jukeboxBlock);
+					$tile->getInventory()->setItem($slot, $item->isNull() ? ItemFactory::air() : $item);
 				}
 				return true;
 
--- a/src/block/Hopper.php
+++ b/src/block/Hopper.php
@@ -170,14 +170,14 @@ class Hopper extends Transparent{
 				// If the hopper is facing in any other direction, it will only push items that can be used as fuel to the furnace's fuel slot.
 				if($this->facing === Facing::DOWN){
 					// ID of the smelting slot
-					$slot = 0;
+					$slotInFurnace = 0;
 					$itemInFurnace = $destination->getInventory()->getSmelting();
 				}else{
 					if($item->getFuelTime() === 0){
 						continue;
 					}
 					// ID of the fuel slot
-					$slot = 1;
+					$slotInFurnace = 1;
 					$itemInFurnace = $destination->getInventory()->getFuel();
 				}
 				if(!$itemInFurnace->isNull()){
@@ -195,7 +195,7 @@ class Hopper extends Transparent{
 
 				// TODO event on item inventory switch
 
-				$destination->getInventory()->setItem($slot, $itemInFurnace);
+				$destination->getInventory()->setItem($slotInFurnace, $itemInFurnace);
 				$tile->getInventory()->setItem($slot, $item->isNull() ? ItemFactory::air() : $item);
 				return true;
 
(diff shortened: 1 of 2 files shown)


dktapps:
I don't have the time to do an in-depth review right now, but one quick note: setting an `isNull` item into an inventory does the same thing as setting an air item. You don't need to check for it.

ColinHDev:
Oh, yeah, I see that now.
Thanks, I changed it

src/block/Hopper.php
```diff
@@ -96,8 +103,238 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+
+		if($transferCooldown > 0){
+			// Skipping this tick because the hopper is still on cooldown of 8 ticks.
+			return;
+		}
+
+		$success = $this->push($tile);
+		// Hoppers that have a container above them, won't try to pick up items.
+		$origin = $this->position->getWorld()->getTile($this->position->getSide(Facing::UP));
+		if($origin instanceof Container){
```
sylvrs, line 132: This could be simplified to:
`$success |= $origin instanceof Container ? $this->pull($tile, $origin) : $this->pickup($tile);`

src/block/Hopper.php
```diff
@@ -96,8 +103,238 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+			$success |= $this->pull($tile, $origin);
+		}else{
+			$success |= $this->pickup($tile);
+		}
+		// The cooldown is only set back to 8 ticks if the hopper has done anything.
+		if((bool) $success){
```
sylvrs, line 138: This cast seems unnecessary given the fact that Hopper::push(), Hopper::pull(), and Hopper::pickup() all return booleans.

src/block/Hopper.php
```diff
@@ -96,8 +103,238 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+			$tile->setTransferCooldown(8);
```
sylvrs, line 139: This should be moved to a constant. Hardcoding the value is not good practice.

src/block/Hopper.php
```diff
@@ -96,8 +103,238 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+				return true;
+
+			}elseif($destination instanceof TileHopper){
+				$itemToPush = $item->pop();
+				if(!$destination->getInventory()->canAddItem($itemToPush)){
+					continue;
+				}
+				// Hoppers pushing into empty hoppers set the empty hoppers transfer cooldown back to 8 ticks.
+				if(count($destination->getInventory()->getContents()) === 0){
+					$destination->setTransferCooldown(8);
```
sylvrs, line 207: See my above comment about using hardcoded values.

src/block/Hopper.php
```diff
@@ -96,8 +103,238 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+					if(!$itemInFurnace->canStackWith($item)){
+						continue;
+					}
+					$item->pop();
+					$itemInFurnace->setCount($itemInFurnace->getCount() + 1);
+				}else{
+					$itemInFurnace = $item->pop();
+				}
+
+				// TODO event on item inventory switch
```
sylvrs, line 194: I saw this in your main PR TODOs, but I believe that events should be included before this PR is merged.

sylvrs (changes requested):
The PR overall seems good, but there are some nitpicks that I thought I'd point out 😄

src/block/Hopper.php
ColinHDev: I especially didn't choose the ternary operator here because in the future, it can't be used without making everything crowded. 
Right now, it is easy to just use the Container interface as data type. But if Composters and Jukebox (Tiles that aren't Containers but still affected by hoppers) are fully implemented in the future, this if statement needs to change

src/block/Hopper.php
ColinHDev: The bitwise operations cast $success at this point to an integer of 0 (false) or 1 (true).
I know that it is automatically cast to a boolean in the if statement but I made this to ensure that this definitely is a boolean and won't break in the future (idk.. if `declare(strict_types=1)` would stop auto casting in if statements in a future php version)

src/block/Hopper.php
ColinHDev: Dylan already pointed out that this PR should only contain the logic itself and no events.
Also, the event for block picking up an item only exists since this commit https://github.com/pmmp/PocketMine-MP/commit/2139171a5525a89061cbadfa30f3f3f1243973b4 and we don't have an event for inventory interaction (InventoryTransactionEvent is not possible for this case, explained in https://github.com/pmmp/PocketMine-MP/issues/4405)

src/block/Hopper.php
ColinHDev: Hm
I am unsure about a clear name for that constant.
TRANSFER_COOLDOWN doesn't suit in my opinion, because the variable that stores the current cooldown of the hopper already has just a slightly different name which could confuse.
But I don't think that MAX_TRANSFER_COOLDOWN is so perfect either.

src/block/Hopper.php
```diff
@@ -96,8 +103,238 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+				if(!$destination->getInventory()->canAddItem($itemToPush)){
+					continue;
+				}
+				// Hoppers pushing into empty hoppers set the empty hoppers transfer cooldown back to 8 ticks.
+				if(count($destination->getInventory()->getContents()) === 0){
+					$destination->setTransferCooldown(8);
+				}
+
+			}elseif($destination instanceof TileJukebox){
+				if(!$item instanceof Record){
```
dktapps, line 211: Please surround this with braces like so:
```suggestion
				if(!($item instanceof Record)){
```
I know it's not functionally necessary, but it makes it easier to read.

dktapps (changes requested):
Not a depth review, but I have quite a few PRs to go over. Overall doesn't look bad, but I do find a lot of the comments unnecessary and there are a few other things I'm not too stoked about. I'll come back to this.

src/block/Hopper.php
ColinHDev: The bitwise operators were nice because they helped me implement the hopper logic (only reset cooldown if done anything) with just these few lines.
And I choose to cast it back to a boolean instead of comparing its value because it looked more clear to me.

ColinHDev:
Changed everything except the bitwise operator for now.
If I removed it, the code would get a bit bigger from (for example)
```php
$success |= $this->pickup($tile);
```
to
```php
if($success){
    $this->pickup($tile);
}else{
    $success = $this->pickup($tile);
}
```

dktapps:
It can be written as `$success = $success && doSomething(...)` to stay purely within the bounds of boolean behaviour.

ColinHDev:
> It can be written as `$success = $success && doSomething(...)` to stay purely within the bounds of boolean behaviour.
It shouldn't at least according to both my knowledge and the wiki (I looked it up for that https://www.php.net/manual/en/language.operators.logical.php).
If pushing failed and `$success` would be false, the pull or pickup functions wouldn't be called

dktapps:
Oops. I meant `||`, not `&&`.

ColinHDev:
But then the functions wouldn't be called if the pushing succeeded and `$success` would be true

dktapps:
You're right. I guess the condition order would have to be flipped.

ColinHDev:
Yes
```php
$success = $this->pickup($inventory) || $success;
```
should do it.
It looks a little bit awkward in my opinion, that's why I wanted to use bitwise operators. But if it works and you think it suits better, I will change it

ColinHDev:
Changed everything, also bitwise operators. I would leave this PR open until https://github.com/pmmp/PocketMine-MP/pull/4413 is merged (or rejected) because its implementations of the BrewingStand and the improvements of the BrewingStandInventory would help m4e to also implement pushing and pulling logic for BrewingStands

ColinHDev:
I still don't have a good idea how to ditch dependency on Container files in the pickup logic.. It works for now, but how stated in the TODO in the future, not all blocks a hopper can pull from are containers

PR #4419, opened 2021-08-30
Implement ItemEntity merging.

rasu3n:
### API changes

Added ItemMergeEvent
Added ItemEntity->setStackSize(int)
### Behavioural changes

ItemEntity will attempt to merge with nearby mergeable items.

## Tests

https://user-images.githubusercontent.com/76860328/131283479-fa651d60-f941-472d-9650-4bd08cdc7fbb.mp4


dktapps:
Don't mix unrelated changes.

Rush2929:
> Don't mix unrelated changes.

There are no unrelated changes.

dktapps:
There are two distinctly separate sets of changes in here which have no dependency on each other.

Rush2929:
I'm really sorry, can you please tell me where the problem is...

dktapps:
This PR contains 2 unrelated changes:
1) Implementing item entity coalescing
2) Allowing players to pick up some (but not all) items from a dropped item entity with a full inventory.

These two changes don't relate to each other.

Rush2929:
> This PR contains 2 unrelated changes:
> 
> 1. Implementing item entity coalescing
> 2. Allowing players to pick up some (but not all) items from a dropped item entity with a full inventory.
> 
> These two changes don't relate to each other.

Thank you very much. Now, for this pull request, I will implement only the merge one, and separate the others into a separate pull request.

dktapps:
A brief look over the code looks OK, but I don't believe it's necessary (or a good idea for performance) for non-moving dropped items to check for nearby dropped items. I think it would be better if only moving dropped items did these checks.

dktapps:
This is going in the opposite direction than what I asked for.

Rush2929:
I'm sorry about that. I've been busy with real life lately, so I'll resume work after tomorrow at the earliest.

dktapps (changes requested):
Right now, this PR is just going to degrade overall performance, ironically, for a feature that's intended to improve performance. This is why I never bothered to implement it in the first place.

The way this PR is implementing merge checks is particularly suboptimal, since it requires non-moving items to continuously check for other nearby items. This is nonsensical, since a movement of an item entity near another item entity is required for merge conditions to be reached anyway - if no item entity recently entered your "nearby" area, you can't do any merging anyway, so you're just wasting CPU time.

In a nutshell, this means that only **moving** items need to check for other nearby items. Since this should be the 1% case (most items will sit still on the ground for most of their lifetime), this would be the way to implement it with minimal performance degradation.

PR #4429, opened 2021-09-03
Farmland changes to dirt when jumped on.

aderoian:
## Introduction
In vanilla Minecraft, if an entity jumps on farmland, that farmland gets 'trampled' and turns to dirt. This pull request adds this functionality to PocketMine.

### Relevant issues
https://github.com/pmmp/PocketMine-MP/projects/14#card-16053523

## Changes
none

### Behavioural changes
Server replaces the farmland with dirt when 'trampled'.

## Backwards compatibility
none

## Follow-up
none

Requires translations: none

## Tests

Placed farmland block and fell on it from different heights (including jumping), all cases the farmland changed to dirt. Walked across farmland and fell from less than 1 block, all cases the farmland did not change to dirt.

No plugin were installed.

### DIFF
--- a/src/block/Block.php
+++ b/src/block/Block.php
@@ -236,8 +236,7 @@ class Block{
 	/**
 	 * Called when an Entity falls on this block.
 	 */
-	public function onEntityFall(Entity $entity, float $fallDistance, Vector3 $fallPos): void
-	{
+	public function onEntityFall(Entity $entity, float $fallDistance, float $jumpVelocity): void{
 
 	}
 
--- a/src/block/Block.php
+++ b/src/block/Block.php
@@ -234,6 +234,14 @@ class Block{
 	}
 
 	/**
+	 * Called when an Entity falls on this block.
+	 */
+	public function onEntityFall(Entity $entity, float $fallDistance, Vector3 $fallPos): void
+	{
+
+	}
+
+	/**
 	 * Returns whether random block updates will be done on this block.
 	 */
 	public function ticksRandomly() : bool{
--- a/src/block/Block.php
+++ b/src/block/Block.php
@@ -236,7 +236,7 @@ class Block{
 	/**
 	 * Called when an Entity falls on this block.
 	 */
-	public function onEntityFall(Entity $entity, float $fallDistance, float $jumpVelocity): void{
+	public function onEntityFall(Entity $entity, float $fallDistance): void{
 
 	}
 
--- a/src/block/Crops.php
+++ b/src/block/Crops.php
@@ -24,7 +24,6 @@ declare(strict_types=1);
 namespace pocketmine\block;
 
 use pocketmine\block\utils\BlockDataSerializer;
-use pocketmine\entity\Entity;
 use pocketmine\event\block\BlockGrowEvent;
 use pocketmine\item\Fertilizer;
 use pocketmine\item\Item;
--- a/src/block/Crops.php
+++ b/src/block/Crops.php
@@ -24,6 +24,7 @@ declare(strict_types=1);
 namespace pocketmine\block;
 
 use pocketmine\block\utils\BlockDataSerializer;
+use pocketmine\entity\Entity;
 use pocketmine\event\block\BlockGrowEvent;
 use pocketmine\item\Fertilizer;
 use pocketmine\item\Item;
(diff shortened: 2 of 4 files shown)


src/entity/Living.php
```diff
@@ -326,6 +327,9 @@ public function fall(float $fallDistance) : void{
 				$this->broadcastSound(new EntityLandSound($this, $fallBlock));
 			}
 		}
+		if($fallDistance > 0.99 && $fallBlock->getId() === BlockLegacyIds::FARMLAND) {
+			$this->getWorld()->setBlock($fallBlockPos, BlockFactory::getInstance()->get(BlockLegacyIds::DIRT, 0));
```
alvin0319, line 331:
```php
VanillaBlocks::DIRT()
```
would be better on here

Govdim:
You need to add an event

dktapps:
Farmland isn't important enough to get its own special space in Entity just for this. This code doesn't belong here.

src/entity/Living.php
dktapps: There's no guarantee this will actually work at all, since farmland has a 1x1x1 collision box. This means that the fall position collected above will actually be the block above the farmland in every case (unless a rounding error occurs).

dktapps (changes requested):
There's no guarantee this will actually work at all, since farmland has a 1x1x1 collision box. This means that the fall position collected above will actually be the block above the farmland in every case (unless a rounding error occurs).

In addition, farmland isn't important enough to merit this kind of specialization. An `onEntityFall()` hook in `Block` would be better.

Cosmic5173:
There is an if/else branch that checks to see if the fall block is air, in the cases where position = block above farmland. If so, it gets the block below that air block.

src/block/Block.php
```diff
@@ -233,6 +233,14 @@ public function onNearbyBlockChange() : void{
 
 	}
 
+	/**
+	 * Called when an Entity falls on this block.
+	 */
+	public function onEntityFall(Entity $entity, float $fallDistance, Vector3 $fallPos): void
+	{
```
dktapps, line 240: This should be formatted according to the PM code style.

dktapps: It would also be useful to pass the entity's velocity in here (for stuff like slime blocks).

dktapps: Passing `fallPos` here seems pointless, since it's always going to be the block's own position.

src/block/Crops.php
```diff
@@ -111,4 +112,11 @@ public function onRandomTick() : void{
 			}
 		}
 	}
-}
+
+	public function onEntityFall(Entity $entity, float $fallDistance, Vector3 $fallPos) : void{
+		$blockBelow = $this->getSide(Facing::DOWN);
+		if($blockBelow->getId() === BlockLegacyIds::FARMLAND){
+			$blockBelow->onEntityFall($entity, $fallDistance, $fallPos);
```
dktapps, line 119: I don't like this. It feels too much like a hack.

src/block/Farmland.php
```diff
@@ -112,4 +114,10 @@ public function getDropsForCompatibleTool(Item $item) : array{
 	public function getPickedItem(bool $addUserData = false) : Item{
 		return VanillaBlocks::DIRT()->asItem();
 	}
+
+	public function onEntityFall(Entity $entity, float $fallDistance, Vector3 $fallPos) : void{
+		if($fallDistance > 0.5) {
```
dktapps, line 119: Formatting issue here.

src/block/Crops.php
Cosmic5173: wdym? fall() in Living returns the block the player falls onto, including crops

src/block/Farmland.php
```diff
@@ -112,4 +114,10 @@ public function getDropsForCompatibleTool(Item $item) : array{
 	public function getPickedItem(bool $addUserData = false) : Item{
 		return VanillaBlocks::DIRT()->asItem();
 	}
+
+	public function onEntityFall(Entity $entity, float $fallDistance, Vector3 $fallPos) : void{
+		if($fallDistance > 0.99) {
```
TheNewHEROBRINEX: Why not `$fallDistance >= 1`?

src/entity/Living.php
```diff
@@ -329,7 +329,11 @@ public function fall(float $fallDistance) : void{
 			}
 		}
 
-		$fallBlock->onEntityFall($this, $fallDistance, $fallBlockPos);
+		if ($fallBlock instanceof Flowable ){
+			$fallBlockPos = $fallBlockPos->subtract(0, 1, 0);
+			$fallBlock = $this->getWorld()->getBlock($fallBlockPos);
+		}
+		$fallBlock->onEntityFall($this, $fallDistance, $this->getJumpVelocity());
```
dktapps, line 336: This isn't useful. The actual velocity of the impact is what I was talking about.

Cosmic5173: Is that the damage?

dktapps: No, it's the velocity. (or motion, since PM likes to incorrectly name things.)

dktapps: In fact, since it's unrelated to the PR, you can omit it entirely.

dktapps:
Obsoleted by #4434.

PR #4433, opened 2021-09-05
Implement a proposed EntityRegistrar for PM4

CortexPE:
## Introduction
PM4 currently lacks a way to overwrite and/or "register" entities, this is due to the new improved design principles that avoids type-unsafe variadic argument passing to entity constructors and for the better separation of the network layer from the plugin API

### Build
Implemented `Closure` signature generation for `build/generate-registry-annotations.php`

### API changes
Added a `VanillaEntities` class that works similarly to other `Vanilla<type>` classes

## Backwards compatibility
Since there is no actual entity registration in PM4 and developers needed to construct the entities themselves, backwards-compatibility is not an issue.

## Follow-up
**This pull request is not done, this is just the first stage. However I am asking for comments for further development as the next stages require more changes in the internal PocketMine codebase to migrate from direct entity construction, to the usage of the `VanillaEntities` class**

`EntityFactory` should be renamed appropriately to `EntityDataParser` (or a similar name) as its actual purpose is to construct the entities from the NBT data read from disk.

## Tests
```php
// allowed:
VanillaEntities::register("human", function(Location $location, Skin $skin, ?CompoundTag $nbt = null):CustomHuman{
	return new CustomHuman($location, $skin, $nbt);
});

// not allowed:
VanillaEntities::register("human", function(Location $location, Skin $skin, Player $linked, ?CompoundTag $nbt = null):EnderPearl{
	return new EnderPearl($location, null, $nbt);
});
```


dktapps:
> `EntityFactory` should be renamed appropriately to `EntityDataParser` (or a similar name) as its actual purpose is to construct the entities from the NBT data read from disk.

It should also be moved to the `pocketmine/data` package.

src/entity/VanillaEntities.php
```diff
@@ -0,0 +1,154 @@
...
+		});
+		self::_registryRegister("zombie", function(Location $location, ?CompoundTag $nbt = null) : Zombie{
+			return new Zombie($location, $nbt);
+		});
+		self::_registryRegister("human", function(Location $location, Skin $skin, ?CompoundTag $nbt = null) : Human{
+			return new Human($location, $skin, $nbt);
+		});
+	}
+
+	public static function register(string $name, \Closure $callback) : void{
```
dktapps, line 120: I don't like this. This is bringing back the exact same problems we had in the first place with inability to statically analyse code.

src/entity/VanillaEntities.php
```diff
@@ -0,0 +1,154 @@
...
+	public static function __callStatic($name, $arguments){
+		try{
+			return self::_registryFromString($name)(...$arguments);
+		}catch(\InvalidArgumentException $e){
+			throw new \Error($e->getMessage(), 0, $e);
+		}
+	}
+
+	/**
+	 * @return callable[]
```
dktapps, line 146: We don't use `callable` type. Use `\Closure` instead.

src/entity/VanillaEntities.php
```diff
@@ -0,0 +1,154 @@
...
+	 */
+	public static function getAll() : array{
+		//phpstan doesn't support generic traits yet :(
+		/** @var callable[] $result */
```
dktapps, line 150: Same again here.

build/generate-registry-annotations.php
```diff
@@ -42,6 +47,55 @@
 	die("Provide a path to process");
 }
 
+function value2str($value){
+	if(is_string($value)) return "\"$value\"";
+	elseif(is_bool($value)) return $value ? "true" : "false";
+	elseif(is_null($value)) return "null";
+	elseif(is_array($value)) return "[]"; // todo: array values
+	return (string) $value;
+}
```
dktapps, line 56: `match(true)` would better suit for this.

dktapps (changes requested):
Concept is OK, but I'm not a big fan of some of the implementation details.

src/entity/VanillaEntities.php
CortexPE: I'm not sure how it could be declared in PHPStan that this only allows a compatible signature when compared to the entity registered before it

dktapps: It can't. That's exactly what I've been saying. This has the same problems that made the old systen unsuitable.

ColinHDev:
Any updates on that, since it's quite a pain to overwrite existing entities.

dktapps:
I'm currently leaning towards an `EntityCreationEvent`. I don't like this solution; it's too much of a hack and it's too inflexible. If we're going to do a hack for a poor API, we can at least do the hack well.

dktapps:

PR #4434, opened 2021-09-05
Farmland turns to dirt when jumped on.

aderoian:
## Introduction
In vanilla Minecraft, if an entity jumps on farmland, that farmland gets 'trampled' and turns to dirt. This pull request adds this functionality to PocketMine.

### Relevant issues
https://github.com/pmmp/PocketMine-MP/projects/14#card-16053523

## Changes
none

### Behavioural changes
Server replaces the farmland with dirt when 'trampled'.

## Backwards compatibility
none

## Follow-up
none

Requires translations: none

## Tests

Placed farmland block and fell on it from different heights (including jumping), all cases the farmland changed to dirt. Walked across farmland and fell from less than 1 block, all cases the farmland did not change to dirt.

No plugin were installed.

### DIFF
--- a/src/block/Farmland.php
+++ b/src/block/Farmland.php
@@ -30,6 +30,7 @@ use pocketmine\event\entity\EntityTrampleFarmlandEvent;
 use pocketmine\item\Item;
 use pocketmine\math\AxisAlignedBB;
 use pocketmine\math\Facing;
+use function lcg_value;
 
 class Farmland extends Transparent{
 
@@ -89,9 +90,9 @@ class Farmland extends Transparent{
 		}
 	}
 
-	public function onEntityLand(Entity $entity): ?float{
+	public function onEntityLand(Entity $entity) : ?float{
 		if($entity instanceof Living && lcg_value() < $entity->getFallDistance() - 0.5){
-			$ev = new EntityTrampleFarmlandEvent($entity, $this->getPosition()->getWorld()->getBlock($entity->getPosition()->down()));
+			$ev = new EntityTrampleFarmlandEvent($entity, $this);
 			$ev->call();
 			if(!$ev->isCancelled()){
 				$this->getPosition()->getWorld()->setBlock($this->getPosition(), VanillaBlocks::DIRT());
--- a/src/block/Farmland.php
+++ b/src/block/Farmland.php
@@ -24,6 +24,8 @@ declare(strict_types=1);
 namespace pocketmine\block;
 
 use pocketmine\block\utils\BlockDataSerializer;
+use pocketmine\entity\Entity;
+use pocketmine\entity\Living;
 use pocketmine\item\Item;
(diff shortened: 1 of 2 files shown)


src/block/Farmland.php
```diff
@@ -86,6 +88,14 @@ public function onRandomTick() : void{
 		}
 	}
 
+	public function onEntityLand(Entity $entity): ?float
+	{
+		if($entity instanceof Living && lcg_value() < $entity->getFallDistance()) {
```
dktapps, line 93: In vanilla an additional 0.5 is subtracted from the fall distance.

src/block/Farmland.php
```diff
@@ -86,6 +88,14 @@ public function onRandomTick() : void{
...
+			$this->getPosition()->getWorld()->setBlock($this->getPosition(), VanillaBlocks::DIRT(), true);
```
dktapps, line 94:
```suggestion
			$this->getPosition()->getWorld()->setBlock($this->getPosition(), VanillaBlocks::DIRT());
```

dktapps (commented):
Please reformat this according to the PM-provided `Project` code style.

I'd also like to see an event added for this.

src/block/Farmland.php
```diff
@@ -86,6 +89,17 @@ public function onRandomTick() : void{
 		}
 	}
 
+	public function onEntityLand(Entity $entity): ?float{
+		if($entity instanceof Living && lcg_value() < $entity->getFallDistance() - 0.5){
+			$ev = new EntityTrampleFarmlandEvent($entity, $this->getPosition()->getWorld()->getBlock($entity->getPosition()->down()));
```
dktapps, line 94: This is unnecessarily complicated. You can just pass `$this`.

src/block/Farmland.php
```diff
@@ -86,6 +89,17 @@ public function onRandomTick() : void{
 		}
 	}
 
+	public function onEntityLand(Entity $entity): ?float{
```
dktapps, line 92:
```suggestion
	public function onEntityLand(Entity $entity) : ?float{
```

dktapps:
It seems like you're missing a new file.

dktapps (approved):

src/block/Farmland.php
```diff
@@ -86,6 +90,17 @@ public function onRandomTick() : void{
 		}
 	}
 
+	public function onEntityLand(Entity $entity) : ?float{
+		if($entity instanceof Living && lcg_value() < $entity->getFallDistance() - 0.5){
+			$ev = new EntityTrampleFarmlandEvent($entity, $this);
```
lukeeey, line 95: `EntityTrambleBlockEvent` to match the abstraction of `EntityTeleportEvent`

lukeeey: oops, i meant `BlockTeleportEvent`

PR #4444, opened 2021-09-09
Fixed coarse dirt becoming dirt instead of farmland.

matcracker:
## Introduction

In the BE vanilla when you use the hoe in a coarse dirt block it becomes a farmland block. Instead in PM, it has been implemented like JE vanilla which becomes a dirt block.

### Relevant issues

* Fixed coarse dirt becoming dirt instead of farmland.

### API changes

No.

### Behavioural changes

No.

## Backwards compatibility

It's compatible.

## Follow-up

No.

## Tests

Is it really necessary in this case?


dktapps:
This is a bug in Bedrock. https://bugs.mojang.com/browse/MCPE-139477

matcracker:
I tried to search too but I didn't find the bug so I thought it was a disparity between BE and JE. Forget it, I close the PR.

PR #4445, opened 2021-09-09
Added Item and Player to StructureGrowEvent

matcracker:
## Introduction

I find useful to know, when a StructureGrowEvent is fired, if a structure has been grown by a player or not.

### Relevant issues

No issues.

### API changes

* Changed `StructureGrowEvent` constructor from `__construct(Block $block, BlockTransaction $transaction)` to `__construct(Block $block, BlockTransaction $transaction, ?Player $player)`
* Implemented `StructureGrowEvent::getPlayer() : ?Player`

### Behavioural changes

No.

## Backwards compatibility

It breaks the compatibility for who is calling the event. As we are in beta this change may be allowed. 

## Follow-up

No.

## Tests

Source code:
```php
public function onStructureGrow(StructureGrowEvent $event) : void{
    $blockName = $event->getBlock()->getName();
    $playerName = $event->getPlayer()?->getName();

    var_dump($blockName);
    var_dump($playerName);
}
```

Results (without fertilizer):
```
string(6) "Bamboo Sapling"
NULL
```
```
string(6) "Bamboo"
NULL
```
```
string(11) "Oak Sapling"
NULL
```
Results (with fertilizer):
```
string(14) "Bamboo Sapling"
string(10) "matcracker"
```
```
string(14) "Bamboo Sapling"
string(10) "matcracker"
```
```
string(6) "Bamboo"
string(10) "matcracker"
```
```
string(11) "Oak Sapling"
string(10) "matcracker"
```


src/event/block/StructureGrowEvent.php
```diff
@@ -17,13 +19,33 @@ class StructureGrowEvent extends BlockEvent implements Cancellable{
...
 
 	public function getTransaction() : BlockTransaction{
 		return $this->transaction;
 	}
+
+	/**
+	 * It returns the item used from the player to grow the structure.
+	 * It returns null when the structure grows by itself.
+	 */
+	public function getItem() : ?Item{
```
matcracker, line 40: Maybe `getGrowingItem()` is better?

src/block/Sapling.php
```diff
@@ -100,23 +100,23 @@ public function ticksRandomly() : bool{
...
+				$this->grow(null, null);
 			}else{
 				$this->ready = true;
 				$this->position->getWorld()->setBlock($this->position, $this);
 			}
 		}
 	}
 
-	private function grow() : void{
+	private function grow(?Item $item, ?Player $player) : void{
```
sylvrs, line 111: Any reason for not defaulting these parameters to null? Personally, passing null in the call can be a bit unsightly.

matcracker: I had read once that it was better to avoid the default parameters because they could be forgotten. Since the function is private I might as well use them. I await a response from @dktapps.

dktapps: Yes, optional parameters should be avoided in almost all cases.

dktapps:
I don't think this is necessary. You can use PlayerInteractEvent for this.

matcracker:
> I don't think this is necessary. You can use PlayerInteractEvent for this.

Yes, but I find it like a workaround that involves useless complexity to the plugin. Example, if I want to get the structure blocks and also the player/item, I need to use 2 events, save the info in an external variable and then use it. While, with 1 event I can do all the necessary without going out of the context.

src/event/block/StructureGrowEvent.php
thebigsmileXD: I think getItem is good since it can be anything. getFertilizer might also work but be confusing in cases like bamboo on bamboo

dktapps:
The main issue I have with this is that you have to do multiple null checks, i.e. checking if the Player is null tells you that the structure was grown by a player, but the item used could still be null as far as a static analyser is concerned. I don't like this very much, but I'm not really sure if there's a better solution.
That said, I wonder why the Item is needed at all, since you can get it from the inventory anyway...

matcracker:
> The main issue I have with this is that you have to do multiple null checks, i.e. checking if the Player is null tells you that the structure was grown by a player, but the item used could still be null as far as a static analyser is concerned. I don't like this very much, but I'm not really sure if there's a better solution.
> That said, I wonder why the Item is needed at all, since you can get it from the inventory anyway...

I agree with you. The item is not necessary, it can be obtained from the player's inventory. I make the changes.

thebigsmileXD:
Something came to my mind too late, that is:
Player should probably been replaced by InventoryHolder or InventoryTransaction to allow stuff like Dispenser bonemealing a structure also call this event

Issue #2715, opened 2019-01-29
Players are not where they appear to be when sleeping in bed


dktapps:
### Issue description
While a player appears to be in a bed, PocketMine-MP does not consider them to be in the bed. Instead, they are still thought to be standing where they were when they entered the bed.

This can be observed by trying to place blocks in the player's location after they appear to enter the bed. Shooting arrows through the apparently empty space will also cause the sleeping player to take damage.

### Steps to reproduce the issue
1. Have player1 enter a bed.
2. Have player2 attempt to place blocks where player1 was before entering the bed.

### OS and versions

* PocketMine-MP: 3.5.7
* PHP: 7.2.13
* Server OS: linux

jasonwynn10:
Does this make it so players cannot be attacked in bed?

dktapps:
If you read the issue hard enough, yes.

dadodasyra:
This seems fixed

dktapps:
Nope

dadodasyra:
> Nope

I tried to place blocks where the player was and it worked.

dktapps:
Well nothing has been done to address the issue, so that sounds like it may well be a new bug

PR #4446, opened 2021-09-09
Implement swimming/gliding including AABB recalculation

inxomnyaa:
## Introduction
Replaces #4396 since the repo was removed
This PR aims to implement swimming & gliding, including exhaust handling and bounding box resizing

### API changes
- Added `add` and `multiply` methods to `EntitySizeInfo`
- Added `PlayerToggleGlideEvent`
- Added `PlayerToggleSwimEvent`
- Added `toggleGlide()` & `toggleSwim()` to `Player`
- Entity bounding box methods were added: `setSize()`, `resetSize()`

### Behavioral changes
- Sneaking now changes the height and eyeheight of a player
- Swimming changes food exhaustion

## Backwards compatibility

No

## Follow-up
- [x] Consider moving `setGliding()` and `setSwimming()` to `Living` class
- [ ] Change size when sleeping https://github.com/pmmp/PocketMine-MP/issues/2715
- [ ] Add elytra checks for glide toggle and inWater checks for swim toggle

## Tests

- Checked the bounding box rescaling / setting, especially with scale - delivered correct results.
- [x] Todo: unlike #4396 this PR does not add Elytra (/give still makes it possible to use and test gliding)
- [x] Todo: Fix conflicts when player is both gliding or swimming + sneaking


src/entity/Entity.php
```diff
@@ -829,10 +837,10 @@ protected function checkObstruction(float $x, float $y, float $z) : bool{
 		$diffZ = $z - $floorZ;
 
 		if($world->getBlockAt($floorX, $floorY, $floorZ)->isSolid()){
-			$westNonSolid  = !$world->getBlockAt($floorX - 1, $floorY, $floorZ)->isSolid();
-			$eastNonSolid  = !$world->getBlockAt($floorX + 1, $floorY, $floorZ)->isSolid();
-			$downNonSolid  = !$world->getBlockAt($floorX, $floorY - 1, $floorZ)->isSolid();
-			$upNonSolid    = !$world->getBlockAt($floorX, $floorY + 1, $floorZ)->isSolid();
+			$westNonSolid = !$world->getBlockAt($floorX - 1, $floorY, $floorZ)->isSolid();
```
buchwasa, line 840: Unnecessary formatting changes

src/entity/Entity.php
```diff
@@ -1545,8 +1553,8 @@ protected function destroyCycles() : void{
 	}
 
 	/**
-	 * @param Player[]|null      $targets
-	 * @param MetadataProperty[] $data Properly formatted entity data, defaults to everything
+	 * @param Player[]|null                        $targets
```
buchwasa, line 1556: likewise

src/entity/Entity.php
```diff
@@ -1615,6 +1623,7 @@ public function broadcastAnimation(Animation $animation, ?array $targets = null)
 
 	/**
 	 * Broadcasts a sound caused by the entity. If the entity is considered "silent", the sound will be dropped.
+	 *
```
buchwasa, line 1626: here too

src/player/Player.php
```diff
@@ -946,6 +949,36 @@ public function stopSleep() : void{
...
+		}else{
+			$this->resetSize();
+		}
+	}
+
+	public function setSwimming(bool $value = true) : void{
+		parent::setSwimming($value);
+
+		if($value){
+			$this->setSize((new EntitySizeInfo($this->size->getWidth(),$this->size->getWidth(),$this->size->getWidth()-($this->size->getHeight() - $this->size->getEyeHeight())))->scale($this->getScale()));
```
buchwasa, line 976: I don't think this logic should be in Player, as it doesn't apply to just players, for example, Drowned. They may not be implemented right now, but if something like that is implemented, they will require the same size logic

src/entity/Entity.php
```diff
@@ -874,7 +882,7 @@ protected function checkObstruction(float $x, float $y, float $z) : bool{
 
 			$force = lcg_value() * 0.2 + 0.1;
 
-			$this->motion = match($direction){
+			$this->motion = match ($direction) {
```
buchwasa, line 885: here too

buchwasa (changes requested):
There's a lot going on with this PR, your title is misleading because it contains changes that aren't related to it, like sneaking

src/entity/Entity.php
thebigsmileXD: I have set phpstorm to use the PM config that dylan sent me, no idea why this editorConfig does this. Maybe PM now does it this way?

src/entity/EntitySizeInfo.php
```diff
@@ -52,4 +52,20 @@ public function scale(float $newScale) : self{
 			$this->eyeHeight * $newScale
 		);
 	}
+
+	public function add(float $height = 0, float $width = 0, float $eyeHeight = 0) : self{
```
dktapps, line 56: Avoid optional parameters.

src/entity/EntitySizeInfo.php
```diff
@@ -52,4 +52,20 @@ public function scale(float $newScale) : self{
...
+		return new self(
+			$this->height + $height,
+			$this->width + $width,
+			$this->eyeHeight + $eyeHeight
+		);
+	}
+
+	public function multiply(float $height = 1, float $width = 1, float $eyeHeight = 1) : self{
```
dktapps, line 64: Avoid optional parameters.

src/player/Player.php
```diff
@@ -946,6 +949,36 @@ public function stopSleep() : void{
...
+		}else{
+			$this->resetSize();
+		}
+	}
+
+	public function setGliding(bool $value = true) : void{
+		parent::setGliding($value);
+
+		if($value){
+			$this->setSize((new EntitySizeInfo($this->size->getWidth(),$this->size->getWidth(),$this->size->getWidth()-($this->size->getHeight() - $this->size->getEyeHeight())))->scale($this->getScale()));
```
dktapps, line 966: This is really difficult to read due to lack of spacing.

src/player/Player.php
dktapps: `Living` would be better.

src/entity/Entity.php
dktapps: There's a code style provided in the repo itself called `Project`, you can use that. But even so, the styling isn't perfect.

src/entity/Entity.php
```diff
@@ -328,6 +326,16 @@ protected function recalculateBoundingBox() : void{
 		);
 	}
 
+	public function setSize(EntitySizeInfo $size) : void{
```
dktapps, line 329: Plugins can't make effective use of this because there is no getter.

src/entity/Entity.php
```diff
@@ -24,6 +24,7 @@
 /**
  * All the entity classes
  */
+
```
dktapps, line 27:
```suggestion
```

src/entity/Entity.php
thebigsmileXD: I am using Project + editorconfig, it still appears to reformat phpdocs though. Might be a missing setting.

src/player/Player.php
thebigsmileXD: Already resolved locally

src/entity/Entity.php
thebigsmileXD: True, yet with the PM code style settings this will happen on every PR. I will try to manually patch this change out, yet still PM should fix this.

thebigsmileXD:
It might be useful to add elytra checks when toggling gliding similar to https://github.com/pmmp/PocketMine-MP/pull/1934/files#diff-cf181dff0292664e4aa13c8a1731dc62131489fa404f4ac1357493d320264cceR2726-R2729

dktapps (commented):
This looks more or less OK, minus a few nits.

thebigsmileXD:
Merge? I want to do the next thing

src/entity/Living.php
```diff
@@ -225,6 +226,38 @@ public function setSprinting(bool $value = true) : void{
...
+		$this->networkPropertiesDirty = true;
+		$this->recalculateSize();
+	}
+
+	public function recalculateSize() : void{
+		$size = $this->getInitialSizeInfo();
+		if($this->isSwimming() || $this->isGliding()){
+			$this->setSize((new EntitySizeInfo($size->getWidth(), $size->getWidth(), $size->getHeight() - $size->getEyeHeight()))->scale($this->getScale()));
+		}elseif($this->isSneaking()){
+			$this->setSize($size->multiply(11 / 12, 1, 11 / 12)->scale($this->getScale()));
```
dktapps, line 257: The eye height change here feels wrong. I would expect the distance between eye and top of head to remain the same when crouching, but this would reduce that distance to 11/12.

dktapps (commented):
Overall this looks OK as far as I can tell. There's just one thing that's bugging me.

src/entity/Living.php
dktapps: According to [this issue](https://bugs.mojang.com/browse/MC-90598), the correct height for a player would be 1.54 (original height - 0.08).

(Bizarrely, this reduces the distance to top of head to 0.11, while it's 0.15 when not crouching. Still 11/12 is incorrect for eye height - would give 1.485 instead of the expected 1.54.)

thebigsmileXD: bedrock is different though

dktapps: Well to my knowledge this is one thing that's the same across both versions (not counting the recent change for slabs).

src/entity/Living.php
```diff
@@ -116,6 +116,10 @@ abstract class Living extends Entity{
 	protected $sprinting = false;
 	/** @var bool */
 	protected $sneaking = false;
+	/** @var bool */
+	protected $gliding = false;
```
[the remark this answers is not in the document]
thebigsmileXD: Why did you revert this then

src/entity/Living.php
```diff
@@ -249,14 +251,13 @@ public function setSwimming(bool $value = true) : void{
...
 		$size = $this->getInitialSizeInfo();
 		if($this->isSwimming() || $this->isGliding()){
-			$this->setSize((new EntitySizeInfo($size->getWidth(), $size->getWidth(), $size->getHeight() - $size->getEyeHeight()))->scale($this->getScale()));
-		}elseif($this->isSneaking()){
-			$this->setSize($size->multiply(11 / 12, 1, 11 / 12)->scale($this->getScale()));
+			$width = $size->getWidth();
+			$this->setSize((new EntitySizeInfo($width, $width, $width / 2))->scale($this->getScale()));
 		}else{
-			$this->resetSize();
+			$this->setSize($this->getInitialSizeInfo()->scale($this->getScale()));
```
thebigsmileXD, line 260: unnecessary getInitialSizeInfo call

src/entity/Living.php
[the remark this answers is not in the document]
dktapps: Because it shouldn't change unrelated code, and it looks weird to have some of them native and others not.

thebigsmileXD:
I consider the eye height code too vague to be used in PM. It may cause unintended glitches and side effects with projectile positions and anticheat.

dktapps:
Swimming as a whole is likely to cause problems with anti-cheat. I've focused on ensuring that arrows shot while swimming launch from a reasonable position and take a reasonable path (unlike vanilla).

thebigsmileXD:
> unlike vanilla

which sounds wrong - PM should mirror the behavior of Vanilla as much as it can imo

dktapps:
I refuse to mirror obvious bugs.

Issue #4437, opened 2021-09-07
[API4] Anvil does not deal fall damage and does not sound when it hits the ground


MrBlasyMSK:
### Issue description
If you try to throw an anvil at the head to kill or harm them, it does not do so, nor does the sound of hitting the ground sound

- Expected result: What were you expecting to happen? Hurts the player and makes the sound when it hits the ground
- Actual result: What actually happened? Nothing

### Steps to reproduce the issue
1. Throw an anvil from a height of more than 3 or 4 blocks
2. Nothing happens when it falls, that should not happen

### OS and versions

* PocketMine-MP: 4.0.0-BETA1 (git hash: 210b9c7b75fe3183600e85f9f4a24ef125fcb1cf)
* PHP: 8.0.9
* Using JIT: not supported
* Server OS: linux
* Game version: Win10

### Plugins

Plugins (2): BuilderTools v1.3.0-beta1, DevTools v1.15.0+dev

- If you remove all plugins, does the issue still occur? Yes
- If the issue is **not** reproducible without plugins:
  - Have you asked for help on our forums before creating an issue?
  - Can you provide sample, *minimal* reproducing code for the issue? If so, paste it in the bottom section

### Crashdump, backtrace or other files

Expected result video: https://streamable.com/l0ohoh
Actual result video: https://streamable.com/n2g9fy

dktapps:
This is not currently implemented in any version of PM.

MrBlasyMSK:
Renamed issue

IvanCraft623:
Implemented by #5312, #5345, #5348

PR #4460, opened 2021-09-17
Anvil Damage Implementation

nathanswanson:
## Introduction

When a player or an entity has an anvil drop on them no damage is applied to same entity.

### Relevant issues

[Issues 4437](https://github.com/pmmp/PocketMine-MP/issues/4437)

### API changes

No API Changes.

### Behavioural changes

- Anvils now create falling damage at a max of 40HP.  The damage is linked to the tick rate so odd behavior is limited.  
- EntityDamageEvent has been updated to contain "CAUSE_FALLING_ANVIL" 
- PlayerDeathEvent translation factory now has anvil death message.

_Nathan was squashed by an anvil_

## Backwards compatibility

Unlikely to have any issues.

## Follow-up

- Falling Block types could benefit to having an event to handle collisions for plugins.
- Sound when anvil finishes falling.
- Anvil functionality.
- Anvil piston behavior.

Requires translations:

| Name | Value in eng.ini |
| :--: | :---: |
| `death.attack.anvil` | `{%a} was squashed by an anvil` |

## Tests

Game Implementation so no PHPUnit test used. I tested it by dropping anvils at various heights on me and all mobs implemented.

### DIFF
--- a/src/block/Anvil.php
+++ b/src/block/Anvil.php
@@ -38,6 +38,7 @@ use pocketmine\math\Facing;
 use pocketmine\math\Vector3;
 use pocketmine\player\Player;
 use pocketmine\world\BlockTransaction;
+use function min;
 
 class Anvil extends Transparent implements Fallable{
 	use FallableTrait;
--- a/src/block/Anvil.php
+++ b/src/block/Anvil.php
@@ -29,6 +29,9 @@ use pocketmine\block\utils\BlockDataSerializer;
 use pocketmine\block\utils\Fallable;
 use pocketmine\block\utils\FallableTrait;
 use pocketmine\block\utils\HorizontalFacingTrait;
+use pocketmine\entity\Living;
+use pocketmine\event\entity\EntityDamageByBlockEvent;
+use pocketmine\event\entity\EntityDamageEvent;
 use pocketmine\item\Item;
 use pocketmine\math\AxisAlignedBB;
 use pocketmine\math\Facing;
@@ -94,5 +97,14 @@ class Anvil extends Transparent implements Fallable{
 
 	public function tickFalling() : ?Block{
 		return null;
+	}
+
+	public function onHitGround(int $fallDistance, int $fallTime, AxisAlignedBB $boundsOnDespawn) : void{
+		foreach($this->position->getWorld()->getCollidingEntities($boundsOnDespawn) as $ent){
+			if($ent instanceof Living){
+				$damageSource = new EntityDamageByBlockEvent($this, $ent, EntityDamageEvent::CAUSE_FALLING_BLOCK, min($fallDistance * 2, 40));
+				$ent->attack($damageSource);
+			}
+		}
 	}
 }
--- a/src/block/Anvil.php
+++ b/src/block/Anvil.php
@@ -102,11 +102,22 @@ class Anvil extends Transparent implements Fallable{
 	}
 
 	public function onHitGround(FallingBlock $blockEntity) : void{
-		foreach($this->position->getWorld()->getCollidingEntities($blockEntity->getBoundingBox()) as $ent){
-			if($ent instanceof Living){
-				if($blockEntity->fallDistance > 1) {
-					$damageSource = new EntityDamageByEntityEvent($blockEntity, $ent, EntityDamageEvent::CAUSE_FALLING_BLOCK, min($blockEntity->fallDistance * 2, 40));
-					$ent->attack($damageSource);
+		if($this->position->isValid()){
+			foreach($this->position->getWorld()->getCollidingEntities($blockEntity->getBoundingBox()) as $ent){
+				if($ent instanceof Living){
+					if($blockEntity->fallDistance > 1){
+						//If player has helmet do 1/4 the normal damage
+						$helmet = $ent->getArmorInventory()->getHelmet();
+						if($helmet !== null && !$helmet->isNull()){
+							$damageDone = $blockEntity->fallDistance * 0.5;
+							$ent->damageItem($helmet, (int) $damageDone);
+						}else{
+							$damageDone = $blockEntity->fallDistance * 2;
+						}
+
+						$damageSource = new EntityDamageByEntityEvent($blockEntity, $ent, EntityDamageEvent::CAUSE_FALLING_BLOCK, min($damageDone, 40));
+						$ent->attack($damageSource);
+					}
 				}
 			}
 		}
--- a/src/block/Anvil.php
+++ b/src/block/Anvil.php
@@ -104,8 +104,10 @@ class Anvil extends Transparent implements Fallable{
 	public function onHitGround(FallingBlock $blockEntity) : void{
 		foreach($this->position->getWorld()->getCollidingEntities($blockEntity->getBoundingBox()) as $ent){
 			if($ent instanceof Living){
-				$damageSource = new EntityDamageByEntityEvent($blockEntity, $ent, EntityDamageEvent::CAUSE_FALLING_BLOCK, min($blockEntity->fallDistance * 2, 40));
-				$ent->attack($damageSource);
+				if($blockEntity->fallDistance > 1) {
+					$damageSource = new EntityDamageByEntityEvent($blockEntity, $ent, EntityDamageEvent::CAUSE_FALLING_BLOCK, min($blockEntity->fallDistance * 2, 40));
+					$ent->attack($damageSource);
+				}
 			}
 		}
 	}
--- a/src/block/Anvil.php
+++ b/src/block/Anvil.php
@@ -30,7 +30,8 @@ use pocketmine\block\utils\Fallable;
 use pocketmine\block\utils\FallableTrait;
 use pocketmine\block\utils\HorizontalFacingTrait;
 use pocketmine\entity\Living;
-use pocketmine\event\entity\EntityDamageByBlockEvent;
+use pocketmine\entity\object\FallingBlock;
+use pocketmine\event\entity\EntityDamageByEntityEvent;
 use pocketmine\event\entity\EntityDamageEvent;
 use pocketmine\item\Item;
 use pocketmine\math\AxisAlignedBB;
@@ -100,10 +101,10 @@ class Anvil extends Transparent implements Fallable{
 		return null;
 	}
 
-	public function onHitGround(int $fallDistance, int $fallTime, AxisAlignedBB $boundsOnDespawn) : void{
-		foreach($this->position->getWorld()->getCollidingEntities($boundsOnDespawn) as $ent){
+	public function onHitGround(FallingBlock $blockEntity) : void{
+		foreach($this->position->getWorld()->getCollidingEntities($blockEntity->getBoundingBox()) as $ent){
 			if($ent instanceof Living){
-				$damageSource = new EntityDamageByBlockEvent($this, $ent, EntityDamageEvent::CAUSE_FALLING_BLOCK, min($fallDistance * 2, 40));
+				$damageSource = new EntityDamageByEntityEvent($blockEntity, $ent, EntityDamageEvent::CAUSE_FALLING_BLOCK, min($blockEntity->fallDistance * 2, 40));
 				$ent->attack($damageSource);
 			}
 		}
--- a/src/block/Anvil.php
+++ b/src/block/Anvil.php
@@ -33,6 +33,7 @@ use pocketmine\entity\Living;
 use pocketmine\entity\object\FallingBlock;
 use pocketmine\event\entity\EntityDamageByEntityEvent;
 use pocketmine\event\entity\EntityDamageEvent;
+use pocketmine\event\Durable;
 use pocketmine\item\Item;
 use pocketmine\math\AxisAlignedBB;
 use pocketmine\math\Facing;
@@ -108,7 +109,7 @@ class Anvil extends Transparent implements Fallable{
 					if($blockEntity->fallDistance > 1){
 						//If player has helmet do 1/4 the normal damage
 						$helmet = $ent->getArmorInventory()->getHelmet();
-						if($helmet !== null && !$helmet->isNull()){
+						if($helmet != null && !$helmet->isNull() && $helmet instanceof Durable){
 							$damageDone = $blockEntity->fallDistance * 0.5;
 							$ent->damageItem($helmet, (int) $damageDone);
 						}else{
--- a/src/block/Anvil.php
+++ b/src/block/Anvil.php
@@ -109,15 +109,19 @@ class Anvil extends Transparent implements Fallable{
 					if($blockEntity->fallDistance > 1){
 						//If player has helmet do 1/4 the normal damage
 						$helmet = $ent->getArmorInventory()->getHelmet();
+
 						if(!$helmet->isNull() && $helmet instanceof Durable){
-							$damageDone = $blockEntity->fallDistance * 0.5;
-							$ent->damageItem($helmet, (int) $damageDone);
+							$damageDone = $blockEntity->fallDistance * 1.5;
 						}else{
 							$damageDone = $blockEntity->fallDistance * 2;
 						}
 
 						$damageSource = new EntityDamageByEntityEvent($blockEntity, $ent, EntityDamageEvent::CAUSE_FALLING_BLOCK, min($damageDone, 40));
 						$ent->attack($damageSource);
+
+						if(!$damageSource->isCancelled() && !$helmet->isNull() && $helmet instanceof Durable) {
+							$ent->damageItem($helmet, (int) $damageDone);
+						}
 					}
 				}
 			}
--- a/src/block/Anvil.php
+++ b/src/block/Anvil.php
@@ -33,7 +33,7 @@ use pocketmine\entity\Living;
 use pocketmine\entity\object\FallingBlock;
 use pocketmine\event\entity\EntityDamageByEntityEvent;
 use pocketmine\event\entity\EntityDamageEvent;
-use pocketmine\event\Durable;
+use pocketmine\item\Durable;
 use pocketmine\item\Item;
 use pocketmine\math\AxisAlignedBB;
 use pocketmine\math\Facing;
@@ -109,7 +109,7 @@ class Anvil extends Transparent implements Fallable{
 					if($blockEntity->fallDistance > 1){
 						//If player has helmet do 1/4 the normal damage
 						$helmet = $ent->getArmorInventory()->getHelmet();
-						if($helmet != null && !$helmet->isNull() && $helmet instanceof Durable){
+						if(!$helmet->isNull() && $helmet instanceof Durable){
 							$damageDone = $blockEntity->fallDistance * 0.5;
 							$ent->damageItem($helmet, (int) $damageDone);
 						}else{
--- a/src/block/Anvil.php
+++ b/src/block/Anvil.php
@@ -23,6 +23,7 @@ declare(strict_types=1);
 
 namespace pocketmine\block;
 
+use InvalidArgumentException;
 use pocketmine\block\inventory\AnvilInventory;
 use pocketmine\block\utils\BlockDataSerializer;
 use pocketmine\block\utils\Fallable;
@@ -63,7 +64,7 @@ class Anvil extends Transparent implements Fallable{
 	/** @return $this */
 	public function setDamage(int $damage) : self{
 		if($damage < 0 || $damage > 2){
-			throw new \InvalidArgumentException("Damage must be in range 0-2");
+			throw new InvalidArgumentException("Damage must be in range 0-2");
 		}
 		$this->damage = $damage;
 		return $this;
--- a/src/block/utils/Fallable.php
+++ b/src/block/utils/Fallable.php
@@ -37,7 +37,6 @@ interface Fallable{
 
 	/**
 	 * Called just before entity destroys itself.
-	 *
 	 */
 	public function onHitGround(int $fallDistance, int $fallTime, AxisAlignedBB $boundsOnDespawn) : void;
 }
--- a/src/block/utils/Fallable.php
+++ b/src/block/utils/Fallable.php
@@ -24,7 +24,7 @@ declare(strict_types=1);
 namespace pocketmine\block\utils;
 
 use pocketmine\block\Block;
-use pocketmine\math\AxisAlignedBB;
+use pocketmine\entity\object\FallingBlock;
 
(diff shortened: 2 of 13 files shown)


src/entity/object/FallingBlock.php
```diff
@@ -97,13 +101,29 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
...
+		$this->fallTime+=1;
 
 		$hasUpdate = parent::entityBaseTick($tickDiff);
 
 		if(!$this->isFlaggedForDespawn()){
 			$world = $this->getWorld();
 			$pos = $this->location->add(-$this->size->getWidth() / 2, $this->size->getHeight(), -$this->size->getWidth() / 2)->floor();
 
+
+			if($this->getBlock()->getId() == VanillaBlocks::ANVIL()->GetId()) {
```
thebigsmileXD, line 113: `getId()` and `===`

src/entity/object/FallingBlock.php
```diff
@@ -97,13 +101,29 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
 		if($this->closed){
 			return false;
 		}
+		$this->fallTime+=1;
```
thebigsmileXD, line 104: `$this->fallTime += 1`

src/event/player/PlayerDeathEvent.php
```diff
@@ -129,6 +129,9 @@ public static function deriveMessage(string $name, ?EntityDamageEvent $deathCaus
 			case EntityDamageEvent::CAUSE_DROWNING:
 				return KnownTranslationFactory::death_attack_drown($name);
 
+			Case EntityDamageEvent::CAUSE_FALLING_ANVIL:
```
thebigsmileXD, line 132: `case`

inxomnyaa (changes requested):
Styling

src/lang/KnownTranslationFactory.php
```diff
@@ -569,10 +569,18 @@ public static function commands_whitelist_remove_usage() : Translatable{
...
 	public static function commands_whitelist_usage() : Translatable{
 		return new Translatable(KnownTranslationKeys::COMMANDS_WHITELIST_USAGE, []);
 	}
 
+
+	public static function death_attack_anvil(Translatable|string $param0) : Translatable{
+		return new Translatable(KnownTranslationKeys::DEATH_ATTACK_ANVIL, [
+			0 => $param0,
+		]);
+	}
```
dktapps, line 582: This file is auto generated. Don't change it by hand.

src/lang/KnownTranslationKeys.php
```diff
@@ -128,6 +128,7 @@ final class KnownTranslationKeys{
 	public const COMMANDS_WHITELIST_REMOVE_SUCCESS = "commands.whitelist.remove.success";
 	public const COMMANDS_WHITELIST_REMOVE_USAGE = "commands.whitelist.remove.usage";
 	public const COMMANDS_WHITELIST_USAGE = "commands.whitelist.usage";
+	public const DEATH_ATTACK_ANVIL = "death.attack.anvil";
```
dktapps, line 131: This file is auto generated. Don't change it by hand.

src/event/entity/EntityDamageEvent.php
```diff
@@ -63,6 +63,7 @@ class EntityDamageEvent extends EntityEvent implements Cancellable{
 	public const CAUSE_MAGIC = 13;
 	public const CAUSE_CUSTOM = 14;
 	public const CAUSE_STARVATION = 15;
+	public const CAUSE_FALLING_ANVIL = 16;
```
dktapps, line 66: FALLING_BLOCK would be better (in case more blocks like this are added in the future).

src/entity/object/FallingBlock.php
```diff
@@ -153,4 +173,5 @@ protected function syncNetworkData(EntityMetadataCollection $properties) : void{
 	public function getOffsetPosition(Vector3 $vector3) : Vector3{
 		return $vector3->add(0, 0.49, 0); //TODO: check if height affects this
 	}
+
```
dktapps, line 176: Unnecessary formatting change.

src/entity/object/FallingBlock.php
```diff
@@ -97,13 +101,29 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
...
 			$pos = $this->location->add(-$this->size->getWidth() / 2, $this->size->getHeight(), -$this->size->getWidth() / 2)->floor();
 
+
+			if($this->getBlock()->getId() == VanillaBlocks::ANVIL()->GetId()) {
+
+				$collidedEntities = $world->getCollidingEntities($this->getBoundingBox());
+				foreach($collidedEntities as $ent)
+					if($ent instanceof Living) {
+						$lastCause = $ent->lastDamageCause;
+						if(!($lastCause instanceof EntityDamageByBlockEvent) || $lastCause->getDamager() !== $this->getBlock()) {
```
dktapps, line 119: I'm not very clear what the point of this check is. This is just going to prevent entities from receiving a second damage from falling anvil if their last damage was from an anvil, irrespective of which anvil caused it.

src/entity/object/FallingBlock.php
dktapps: `fallTime++` is better.

src/entity/object/FallingBlock.php
```diff
@@ -47,6 +50,7 @@ public static function getNetworkTypeId() : string{ return EntityIds::FALLING_BL
 
 	protected $gravity = 0.04;
 	protected $drag = 0.02;
+	private $fallTime = 0;
```
dktapps, line 53: This should have a type.

dktapps (changes requested):

src/lang/KnownTranslationKeys.php
nathanswanson: How would I generate another translation key then?

src/event/entity/EntityDamageEvent.php
nathanswanson: Yeah you bring up a good point I will change this.

src/lang/KnownTranslationKeys.php
alvin0319: https://github.com/pmmp/PocketMine-MP/blob/master/build/generate-known-translation-apis.php

src/entity/object/FallingBlock.php
nathanswanson: Yeah now that you mention it It is unexpected behavior. Its goal was to make sure an anvil only damages player once while falling otherwise it damages every tick. I'll fix this.

nathanswanson: I retrace my last statement it does damage entity RESPECTIVE to which anvil.

I tested this by dropping multiple anvils on me consecutively each one damaging every time.

src/lang/KnownTranslationKeys.php
dktapps: It has to be added to the English translation file first, and then the code generated with this script ^

nathanswanson:
PHP CS Fixer is not working on my PHP Storm so for now I will try to fix manually.

src/event/entity/EntityDamageEvent.php
thebigsmileXD: Dripstone i.e.

nathanswanson:
Why does code style fail? on the actual php file its one line, on the log it shows + then -.

Also I made it so the autotranslation key is generated for the anvil death event. But now I get a generated code consistency error.

dktapps:
> Why does code style fail? on the actual php file its one line, on the log it shows + then -.

You've left trailing whitespace on the line. Blank lines shouldn't have any indentation.

> Also I made it so the autotranslation key is generated for the anvil death event. But now I get a generated code consistency error.

You haven't added the string to the language file.

dktapps:
You need to submit a pull request to the master branch of this repo to add the string to `eng.ini`: https://github.com/pmmp/Language

Once that's done, the `resources/locale` submodule should be updated and the translation APIs regenerated.

nathanswanson:
I added the PR. This PR should be good to go after that translation is merged.

src/entity/object/FallingBlock.php
```diff
@@ -97,13 +104,27 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
...
 		}
+		$this->fallTime++;
 
 		$hasUpdate = parent::entityBaseTick($tickDiff);
 
 		if(!$this->isFlaggedForDespawn()){
 			$world = $this->getWorld();
 			$pos = $this->location->add(-$this->size->getWidth() / 2, $this->size->getHeight(), -$this->size->getWidth() / 2)->floor();
 
+			if($this->getBlock()->getId() === VanillaBlocks::ANVIL()->getId()) {
```
dktapps, line 115: This would be better in `Anvil->tickFalling()`.

dktapps: On second look, this is in completely the wrong place. Anvil damage should only apply when the anvil entity actually turns back into a block.

src/entity/object/FallingBlock.php
dktapps: Checking block identity is not reliable. It depends on the World's block cache, which might have been cleared since the damage occurred.

src/entity/object/FallingBlock.php
```diff
@@ -97,13 +104,27 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
...
+
+				$collidedEntities = $world->getCollidingEntities($this->getBoundingBox());
+				foreach($collidedEntities as $ent)
```
dktapps, line 118: Braces must be used.

dktapps: This variable is also redundant. The above statement could be moved directly into the foreach.

src/entity/object/FallingBlock.php
thebigsmileXD: That is not the case for dripstone, it damages everything in it's path and i am pretty sure anvils also work this way.

nathanswanson: I will switch it but I feel like this is a case where the vanilla implementation is worse

thebigsmileXD: On a side note, standing on slabs does not prevent you from getting the damage

nathanswanson:
Ideally I wanted to implement the damage code into the anvil class but having a falling block makes is a little more tricky. I put the code in the FallingBlock->onGround() method.

nathanswanson:
After debugging workspace I FINALLY got my checkers to all work now.

nathanswanson:
Strange, my coding checker didn't detect /** var */ as a problem and ill fix, regardless this PR cannot be merged until language/pmmp gets pulled so I recommend this be marked as on hold or whatever equivalent

dktapps:
You can pull it yourself, if you want to.

```
cd resources/locale
git checkout master
git pull
cd ../..
git add resources/locale
```
and then commit as you would normally.

src/block/Anvil.php
```diff
@@ -63,7 +64,7 @@ public function getDamage() : int{ return $this->damage; }
 	/** @return $this */
 	public function setDamage(int $damage) : self{
 		if($damage < 0 || $damage > 2){
-			throw new \InvalidArgumentException("Damage must be in range 0-2");
+			throw new InvalidArgumentException("Damage must be in range 0-2");
```
dktapps, line 67: This change should be removed (it's not relevant to the goal of the PR).

src/entity/object/FallingBlock.php
```diff
@@ -153,4 +160,19 @@ protected function syncNetworkData(EntityMetadataCollection $properties) : void{
 	public function getOffsetPosition(Vector3 $vector3) : Vector3{
 		return $vector3->add(0, 0.49, 0); //TODO: check if height affects this
 	}
+
+	protected function onHitGround() : ?float{
+		if($this->getBlock()->getId() === VanillaBlocks::ANVIL()->getId()) {
```
dktapps, line 165: I still think this should be delegated to the block, similarly to how `tickFalling()` is handled.

src/entity/object/FallingBlock.php
```diff
@@ -153,4 +160,19 @@ protected function syncNetworkData(EntityMetadataCollection $properties) : void{
...
+			foreach($this->getWorld()->getCollidingEntities($this->getBoundingBox()) as $ent){
+				if($ent instanceof Living){
+					$lastCause = $ent->lastDamageCause;
+					if(!($lastCause instanceof EntityDamageByBlockEvent) || $lastCause->getDamager() !== $this->getBlock()){
+						$damageSource = new EntityDamageByBlockEvent($this->getBlock(), $ent, EntityDamageEvent::CAUSE_FALLING_BLOCK, min($this->fallTime / 2, 40));
```
dktapps, line 170: I'm not completely sure `EntityDamageByBlockEvent` suits here, since it's actually the entity doing the damage.

src/entity/object/FallingBlock.php
```diff
@@ -97,14 +103,15 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
...
+		$this->fallTime++;
 
 		$hasUpdate = parent::entityBaseTick($tickDiff);
 
 		if(!$this->isFlaggedForDespawn()){
 			$world = $this->getWorld();
 			$pos = $this->location->add(-$this->size->getWidth() / 2, $this->size->getHeight(), -$this->size->getWidth() / 2)->floor();
 
-			$this->block->position($world, $pos->x, $pos->y, $pos->z);
+			$this->block->position($world, $pos->getFloorX(), $pos->getFloorY(), $pos->getFloorZ());
```
dktapps, line 114: This change has no relevance to the PR objective.

src/entity/object/FallingBlock.php
```diff
@@ -71,7 +77,7 @@ public static function parseBlockNBT(BlockFactory $factory, CompoundTag $nbt) :
 		}
 
 		if($blockId === 0){
-			throw new \UnexpectedValueException("Missing block info from NBT");
+			throw new UnexpectedValueException("Missing block info from NBT");
```
dktapps, line 80: This change has no relevance to the PR objective.

src/entity/object/FallingBlock.php
```diff
@@ -48,8 +53,9 @@ public static function getNetworkTypeId() : string{ return EntityIds::FALLING_BL
 	protected $gravity = 0.04;
 	protected $drag = 0.02;
 
-	/** @var Block */
-	protected $block;
+	private int $fallTime = 0;
+
+	protected Block $block;
```
dktapps, line 58: This change has no relevance to the PR objective.

nathanswanson:
Alright what about something like this then, fallable now has a method that is called when a falling entity hits the ground.

src/block/utils/Fallable.php
```diff
@@ -33,4 +34,10 @@ interface Fallable{
 	 * Return null if you don't want to change the usual behaviour.
 	 */
 	public function tickFalling() : ?Block;
+
+	/**
+	 * Called just before entity destroys itself.
+	 *
+	 */
+	public function onHitGround(int $fallDistance, int $fallTime, AxisAlignedBB $boundsOnDespawn) : void;
```
thebigsmileXD, line 41: I would not pass the bounding box here. I would instead replace all parameters with FallingBlock, since you can just get the parameters from there PLUS see which block hit the ground. This should probably also call an event of some sorts. PHPDoc comment too big.

inxomnyaa (changes requested):

src/block/Anvil.php
```diff
@@ -100,10 +101,10 @@ public function tickFalling() : ?Block{
 		return null;
 	}
 
-	public function onHitGround(int $fallDistance, int $fallTime, AxisAlignedBB $boundsOnDespawn) : void{
-		foreach($this->position->getWorld()->getCollidingEntities($boundsOnDespawn) as $ent){
+	public function onHitGround(FallingBlock $blockEntity) : void{
+		foreach($this->position->getWorld()->getCollidingEntities($blockEntity->getBoundingBox()) as $ent){
 			if($ent instanceof Living){
-				$damageSource = new EntityDamageByBlockEvent($this, $ent, EntityDamageEvent::CAUSE_FALLING_BLOCK, min($fallDistance * 2, 40));
+				$damageSource = new EntityDamageByEntityEvent($blockEntity, $ent, EntityDamageEvent::CAUSE_FALLING_BLOCK, min($blockEntity->fallDistance * 2, 40));
```
thebigsmileXD, line 107: I have been thinking about the damage calculation of this. It appears the minimum fall distance (from survival tests) to do damage is >1 block (~1.3 damages you, tested by standing on an enchanting table). So instead of the min() there should also be a check for this.

inxomnyaa (changes requested):

src/block/Anvil.php
nathanswanson: Added a check for making sure drop is >1 block

src/block/Anvil.php
```diff
@@ -104,8 +104,10 @@ public function tickFalling() : ?Block{
 	public function onHitGround(FallingBlock $blockEntity) : void{
 		foreach($this->position->getWorld()->getCollidingEntities($blockEntity->getBoundingBox()) as $ent){
 			if($ent instanceof Living){
-				$damageSource = new EntityDamageByEntityEvent($blockEntity, $ent, EntityDamageEvent::CAUSE_FALLING_BLOCK, min($blockEntity->fallDistance * 2, 40));
-				$ent->attack($damageSource);
+				if($blockEntity->fallDistance > 1) {
+					$damageSource = new EntityDamageByEntityEvent($blockEntity, $ent, EntityDamageEvent::CAUSE_FALLING_BLOCK, min($blockEntity->fallDistance * 2, 40));
```
thebigsmileXD, line 108: Should the damage be casted to int? Or is passing floats valid / correct in the case of falling blocks?

thebigsmileXD:
> Wearing a helmet reduces the damage by 1⁄4, but this costs durability on the helmet

Idk if there is a BlockSolidifyEvent of some sorts, it would be useful to invoke it with the falling block's block and the resulting block being the damaged block, so plugins can cancel the breaking.

Plus anvil destruction https://minecraft.fandom.com/wiki/Anvil#Becoming_damaged

nathanswanson:
> > Wearing a helmet reduces the damage by 1⁄4, but this costs durability on the helmet
> 
> 
> 
> Idk if there is a BlockSolidifyEvent of some sorts, it would be useful to invoke it with the falling block's block and the resulting block being the damaged block, so plugins can cancel the breaking.
> 
> 
> 
> Plus anvil destruction https://minecraft.fandom.com/wiki/Anvil#Becoming_damaged 

While adding the helmet damage makes sense I feel like the rest of the listed implementations here  are unrelated to the PR.

src/block/Anvil.php
nathanswanson: Does vanilla implement the damage as integers? I assumed they also do floating point damage.

nathanswanson:
Note: Latest change exposes Living::damageItem()

src/block/Anvil.php
```diff
@@ -102,11 +102,22 @@ public function tickFalling() : ?Block{
...
+						if($helmet !== null && !$helmet->isNull()){
+							$damageDone = $blockEntity->fallDistance * 0.5;
+							$ent->damageItem($helmet, (int) $damageDone);
+						}else{
+							$damageDone = $blockEntity->fallDistance * 2;
+						}
+
+						$damageSource = new EntityDamageByEntityEvent($blockEntity, $ent, EntityDamageEvent::CAUSE_FALLING_BLOCK, min($damageDone, 40));
+						$ent->attack($damageSource);
+					}
```
thebigsmileXD, line 120: else flagForDespawn()?

nathanswanson: flagForDespawn is controlled in Fallingblock.php already.

src/block/Anvil.php
```diff
@@ -108,7 +109,7 @@ public function onHitGround(FallingBlock $blockEntity) : void{
 					if($blockEntity->fallDistance > 1){
 						//If player has helmet do 1/4 the normal damage
 						$helmet = $ent->getArmorInventory()->getHelmet();
-						if($helmet !== null && !$helmet->isNull()){
+						if($helmet != null && !$helmet->isNull() && $helmet instanceof Durable){
```
thebigsmileXD, line 112: What happened here, reverted strict check?

inxomnyaa (changes requested):

src/block/Anvil.php
nathanswanson: Because I got this PHPStan error.

`Error: Strict comparison using !== between pocketmine\item\Item and null will always evaluate to true.`

Rush2929: The return value of `getHelmet()` is always Item.

nathanswanson: Yeah thats why I decided to just remove the check.

src/block/Anvil.php
```diff
@@ -94,4 +101,26 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
 	public function tickFalling() : ?Block{
 		return null;
 	}
+
+	public function onHitGround(FallingBlock $blockEntity) : void{
+		if($this->position->isValid()){
+			foreach($this->position->getWorld()->getCollidingEntities($blockEntity->getBoundingBox()) as $ent){
+				if($ent instanceof Living){
+					if($blockEntity->fallDistance > 1){
+						//If player has helmet do 1/4 the normal damage
```
dktapps, line 110: The damage is reduced **by** 1/4, not **to** 1/4 (at least, according to the wiki).

src/block/Anvil.php
```diff
@@ -94,4 +101,26 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
...
+						$helmet = $ent->getArmorInventory()->getHelmet();
+						if(!$helmet->isNull() && $helmet instanceof Durable){
+							$damageDone = $blockEntity->fallDistance * 0.5;
+							$ent->damageItem($helmet, (int) $damageDone);
```
dktapps, line 114: I don't like this. At this point the damage has not yet been done and might be cancelled. This denies plugins the opportunity to prevent the damage to health and armour via EntityDamageEvent cancelling.

src/block/Anvil.php
nathanswanson: Yeah I tested it you are correct.

dktapps:

PR #4462, opened 2021-09-18
Item/elytra

inxomnyaa:
## Introduction

Adds the elytra item

### Relevant issues

Followup PR to #4446

## Backwards compatibility

- `minecraft:elytra` is registered, thus plugins have to use `$override = true` when registering their own Elytra now.

## Follow-up

- [ ] Add elytra item check before Player gliding action
- [ ] Implement rocket boost
- [ ] Implement in-flight durability reduction
- [ ] Add firework flight boost

## Tests

- Elytra can be put into the armor slot without transaction mismatch
- Right-clicking the item also equips it
- Damage calculation does not crash the server
- [ ] The elytra should not take damage when armor is damaged.

### DIFF
--- a/src/item/Elytra.php
+++ b/src/item/Elytra.php
@@ -1,0 +1,31 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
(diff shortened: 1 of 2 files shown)


redmcteam:
Elytra duration less?

thebigsmileXD:
> Elytra duration less?

If by that you mean the durability being reduced during flight, this will be done in a followup PR. This PR focuses solely on implementing the item itself, not the mechanics.

dktapps:
I'm not sure what the point of that is.

thebigsmileXD:
> 
> 
> I'm not sure what the point of that is.

The point is to not have a massive PR that changes all kinds of stuff *again*, since small PRs seems to be a targeted theme with pull requests. The elytra would still work in the game. Without the implementation of fireworks, firework-based flight boost is also not an option.

Without phantoms, you have no way to (naturally) get any phantom membranes to repair elytra anyways, so for now it would be counterproductive to give them durability reduction.

thebigsmileXD:
>  The PR is useless by itself

PR #4467, opened 2021-09-19
Lectern Implementation

nathanswanson:
## Introduction

Adds Lectern to game, able to hold book and be saved on world file.

### API changes

BlockState access to lectern block

- hasBook: bool
- page: int
- totalPage: int
- book: item

### Behavioural changes

N/A

## Backwards compatibility

Unlikely to be backwards compatible.

## Follow-up

Current bug which I'm dealing with is on servers if a written book has more than one page, you can't navigate to the new page and the arrows don't work.

~~Make book drop on top not inside when removed from lectern.~~

## Tests

Tested it on IOS and Win10: 

- placed written book / book&quill
- removed written book / book&quill
- made sure unable to place multiple books
- checked to make sure if lectern was indestructible when book is present
- checked save state and visual state when updating book


src/block/Lectern.php
```diff
@@ -0,0 +1,90 @@
...
+use pocketmine\item\ItemFactory;
+use pocketmine\item\WritableBookBase;
+use pocketmine\math\Vector3;
+use pocketmine\player\Player;
+
+class Lectern extends Transparent {
+	use FacesOppositePlacingPlayerTrait;
+	use HorizontalFacingTrait;
+
+	protected bool $hasBook = false;
```
dktapps, line 23: This seems pointless, since the `book` field being null would represent this information anyway.

nathanswanson: I have $hasBook because of the lectern having it in its client blockState.

dktapps: It's not in the blockstate, it's in the tile.

nathanswanson: You are right I got them mixed up I removed it from the block

src/block/Lectern.php
```diff
@@ -0,0 +1,88 @@
...
+use pocketmine\item\ItemFactory;
+use pocketmine\item\WritableBookBase;
+use pocketmine\math\Vector3;
+use pocketmine\player\Player;
+
+class Lectern extends Transparent {
+	use FacesOppositePlacingPlayerTrait;
+	use HorizontalFacingTrait;
+
+	protected int $page = 0;
```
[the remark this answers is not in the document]
dktapps: This should also be exposed via API getters and setters.

src/block/Lectern.php
```diff
@@ -0,0 +1,88 @@
...
+
+	public function getStateBitmask() : int{
+		return 0b11;
+	}
+
+	public function readStateFromWorld() : void{
+		parent::readStateFromWorld();
+		$tile = $this->position->getWorld()->getTile($this->position);
+
+		if($tile instanceof TileLectern) {
```
dktapps, line 43: Formatting issue here.

src/block/Lectern.php
```diff
@@ -0,0 +1,88 @@
...
+		$tile = $this->position->getWorld()->getTile($this->position);
+		if($tile instanceof TileLectern){
+			$tile->setBook($this->book);
+			$tile->setPage($this->page);
+			$tile->setTotalPages($this->totalPages);
+		}
+	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
+
```
dktapps, line 63:
```suggestion
```

src/block/Lectern.php
```diff
@@ -0,0 +1,88 @@
...
+		if($item instanceof WritableBookBase) {
```
dktapps, line 64: Formatting issue here.

src/block/Lectern.php
```diff
@@ -0,0 +1,88 @@
...
+		}
+		return false;
+	}
+
+	public function onAttack(Item $item, int $face, ?Player $player = null) : bool{
+		$tile = $this->position->getWorld()->getTile($this->position);
+		if($this->book !== null && !$this->book->isNull() && $tile instanceof TileLectern)
+		{
+			$droppedBook = new ItemEntity(Location::fromObject($this->position->up(), $this->position->getWorld(), 0, 0), $this->book);
+			$droppedBook->spawnToAll();
```
dktapps, line 77: This should use `World::dropItem()` instead of creating the entity directly.

src/block/tile/Lectern.php
```diff
@@ -0,0 +1,90 @@
...
+	protected function addAdditionalSpawnData(CompoundTag $nbt) : void{
+		$nbt->setByte(self::TAG_HAS_BOOK, $this->hasBook ? 1 : 0);
+		$nbt->setInt(self::TAG_PAGE, $this->page);
+		$nbt->setInt(self::TAG_TOTAL_PAGES, $this->totalPages);
+		$nbt->setTag(self::TAG_BOOK, $this->book->nbtSerialize());
+	}
+
+	public function readSaveData(CompoundTag $nbt) : void{
+		if(($itemTag = $nbt->getCompoundTag(self::TAG_BOOK)) !== null){
+
```
dktapps, line 45:
```suggestion
```

src/block/tile/Lectern.php
```diff
@@ -0,0 +1,90 @@
...
+	private $hasBook = false;
+
+	/** @var int */
+	private $page = 0;
+
+	/** @var int */
+	private $totalPages = 0;
+
+	/** @var Item */
+	private $book;
```
dktapps, line 29: These fields can use native types instead of phpdocs.

src/block/Lectern.php
```diff
@@ -0,0 +1,88 @@
...
+
+			$tile->setBook(ItemFactory::air());
+			$tile->setPage(0);
+			$tile->setTotalPages(0);
+			$this->readStateFromWorld();
```
dktapps, line 82: This should not be here. Any interactions with tile data should be done in `writeStateToWorld()` only. Update the block's properties and then use `setBlock()` to update the world.

dktapps (changes requested):
This PR still seems incomplete, and the formatting still needs work.

I don't see any code to deal with turning pages in the book. Everyone will see page zero.
totalPages is also not updated when the book is provided.

src/block/Lectern.php
```diff
@@ -0,0 +1,97 @@
...
+			$this->book = ItemFactory::air();
+			$this->viewedPage = 0;
+
+			$this->position->getWorld()->setBlock($this->position, $this);
+			return true;
+		}
+		return false;
+	}
+
+	public function setViewedPage(int $pageNumber) : void{
```
dktapps, line 84: This should be fluent like other block getters and setters.

src/block/Lectern.php
```diff
@@ -0,0 +1,97 @@
...
+		$this->viewedPage = $pageNumber;
+
+		$this->position->getWorld()->setBlock($this->position, $this);
```
dktapps, line 87: This shouldn't be here. This method might be called on a block which has just been created and doesn't have a position yet.

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -869,6 +872,20 @@ public function handleLabTable(LabTablePacket $packet) : bool{
 		return false; //TODO
 	}
 
+	public function handleLecternUpdate(LecternUpdatePacket $packet) : bool{
+		//TODO: Cancel page turn event
+		//Drop book is handled on server side with an interact event.
+		//Total pages is gathered from book.
+
+		$lectern = $this->player->getWorld()->getBlock(new Position($packet->x, $packet->y, $packet->z, $this->player->getWorld()));
```
dktapps, line 880: This can use `getBlockAt()`.

src/block/Lectern.php
```diff
@@ -0,0 +1,95 @@
...
+			$this->book = ItemFactory::air();
+			$this->viewedPage = 0;
+
+			$this->position->getWorld()->setBlock($this->position, $this);
+			return true;
+		}
+		return false;
+	}
+
+	public function setViewedPage(int $viewedPage) : void{
```
dktapps, line 84: This should be fluent, like other block setters.

src/block/Lectern.php
```diff
@@ -0,0 +1,95 @@
...
+		$this->viewedPage = $viewedPage;
+	}
+
+	public function getViewedPage() : int{
+		return $this->viewedPage;
+	}
+
+	public function setBook(Item $book) : void{
+		$this->book = $book;
```
dktapps, line 93: There's no getter corresponding to this setter.

src/block/Lectern.php
```diff
@@ -0,0 +1,95 @@
...
+
+	public function setViewedPage(int $viewedPage) : void{
+		$this->viewedPage = $viewedPage;
+	}
+
+	public function getViewedPage() : int{
+		return $this->viewedPage;
+	}
+
+	public function setBook(Item $book) : void{
```
dktapps, line 92: This should be fluent, like other block setters.

src/block/tile/Lectern.php
```diff
@@ -0,0 +1,77 @@
...
+		parent::__construct($world, $pos);
+	}
+
+	protected function addAdditionalSpawnData(CompoundTag $nbt) : void{
+		$nbt->setByte(self::TAG_HAS_BOOK, !$this->book->isNull() ? 1 : 0);
+		$nbt->setInt(self::TAG_PAGE, $this->viewedPage);
+		$nbt->setTag(self::TAG_BOOK, $this->book->nbtSerialize());
+
+		if($this->book instanceof WritableBookBase){
+			$nbt->setInt(self::TAG_TOTAL_PAGES, 0);
```
dktapps, line 36: ????

src/block/Lectern.php
dktapps: This should clone the given item.

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -869,6 +872,21 @@ public function handleLabTable(LabTablePacket $packet) : bool{
...
 	}
 
+	public function handleLecternUpdate(LecternUpdatePacket $packet) : bool{
+		//TODO: Cancel page turn event
+		//Drop book is handled on server side with an interact event.
+		//Total pages is gathered from book.
+
+		$lectern = $this->player->getWorld()->getBlockAt($packet->x, $packet->y, $packet->z);
+		if($lectern instanceof Lectern){
+			$lectern->setViewedPage($packet->page);
```
dktapps, line 882: There's no validation here. This could contain anything.

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -869,6 +872,21 @@ public function handleLabTable(LabTablePacket $packet) : bool{
 		return false; //TODO
 	}
 
+	public function handleLecternUpdate(LecternUpdatePacket $packet) : bool{
+		//TODO: Cancel page turn event
+		//Drop book is handled on server side with an interact event.
+		//Total pages is gathered from book.
+
+		$lectern = $this->player->getWorld()->getBlockAt($packet->x, $packet->y, $packet->z);
```
dktapps, line 880: You should check that the target terrain is loaded, and also the distance (a player shouldn't be able to update a lectern a hundred blocks away).

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -869,6 +872,21 @@ public function handleLabTable(LabTablePacket $packet) : bool{
 		return false; //TODO
 	}
 
+	public function handleLecternUpdate(LecternUpdatePacket $packet) : bool{
+		//TODO: Cancel page turn event
+		//Drop book is handled on server side with an interact event.
+		//Total pages is gathered from book.
```
dktapps, line 878: You should check that the total pages match. If it doesn't, it means there's a desync between client/server (maybe the book was changed and the client didn't know about it before it sent this packet). In such case, the action should be ignored.

src/block/Lectern.php
```diff
@@ -0,0 +1,95 @@
...
+			$this->book = $item->pop();
+			$this->position->getWorld()->setBlock($this->position, $this);
+			return true;
+		}
+		return false;
+	}
+
+	public function onAttack(Item $item, int $face, ?Player $player = null) : bool{
+		$tile = $this->position->getWorld()->getTile($this->position);
+		if($this->book !== null && !$this->book->isNull() && $tile instanceof TileLectern){
```
dktapps, line 71: This tile check is unnecessary. You're not interacting with it in this method.

src/block/Lectern.php
dktapps: Bump.

dktapps:

PR #4469, opened 2021-09-20
Edit PHPDoc for ArmorInventory and EffectManager properties

Mohagames205:
## Introduction
ArmorInventory and EffectManager can also be null thus the PHPDoc should reflect that

https://github.com/pmmp/PocketMine-MP/blob/master/src/entity/Living.php#L813-L814

### API changes
Nope

### Behavioural changes
Nope

## Backwards compatibility
Yes it's BC

### DIFF
--- a/src/entity/Living.php
+++ b/src/entity/Living.php
@@ -93,7 +93,7 @@ abstract class Living extends Entity{
 	/** @var EffectManager */
 	protected $effectManager;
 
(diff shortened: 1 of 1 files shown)


dktapps:
This causes more pain than gain. Those fields are only nulled to destroy cyclic references when the entity is no longer used.

PR #4476, opened 2021-09-24
Fixed incorrect parameter checking in BlockFactory::get()

leolee3914:
## Introduction
Fixed incorrect parameter checking in BlockFactory::get()

https://github.com/pmmp/PocketMine-MP/commit/61c59be299e369878aae00976a970904d7ed8d6b#diff-5ee6ec119d7ba65f7894e25fff09a74933c09607be0e9307011ccbd4cbeffd28R946


dktapps:
I think changing the original code to use >= would be more appropriate.

PR #4479, opened 2021-09-25
Add armor slot transaction item validation

inxomnyaa:
## Introduction

This PR adds item validation to armor slots to prevent clients from putting incorrect items into them

### API changes

- Added Item::isValidSlot() 

### Behavioural changes

- Only skulls, pumpkins and helmets can be placed in the head armor slot anymore
- The other slots only accept armor items of the correct type

## Backwards compatibility

- No BC break
- Plugins can still put any item in there

## Follow-up

- Add Elytra slot validation to #4462
- Add checks for offhand

## Tests

- Put items into & out of the slots
- Tried putting an invalid item || chestplates in the slots using gophertunnel - transaction failed
- Put incorrect armor pieces using plugins - still works

### DIFF
--- a/src/inventory/transaction/action/SlotChangeAction.php
+++ b/src/inventory/transaction/action/SlotChangeAction.php
@@ -73,7 +73,7 @@ class SlotChangeAction extends InventoryAction{
 			throw new TransactionValidationException("Slot does not contain expected original item");
 		}
 		if(!$this->sourceItem->isValidSlot($this->inventory, $this->inventorySlot)){
-			throw new TransactionValidationException("Slot {$this->inventorySlot} in inventory of type ".get_class($this->inventory)." does not accept this item");
+			throw new TransactionValidationException("Slot {$this->inventorySlot} in inventory of type " . get_class($this->inventory) . " does not accept this item");
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


dktapps:
I'd prefer if the inventory did these checks, instead of the item.

thebigsmileXD:
> 
> 
> I'd prefer if the inventory did these checks, instead of the item.

Actually had it this way before. I do not like this approach because you'd pollute the inventory with multiple item-specific checks and disallow plugins from adding their own checks (unless they try overriding an inventory class). Also, it is easier to remember adding this check to the item class instead of Inventory (one modified file vs multiple)

buchwasa:
> > I'd prefer if the inventory did these checks, instead of the item.
> 
> Actually had it this way before. I do not like this approach because you'd pollute the inventory with multiple item-specific checks and disallow plugins from adding their own checks (unless they try overriding an inventory class). Also, it is easier to remember adding this check to the item class instead of Inventory (one modified file vs multiple)

It doesn’t pollute inventory as much as you’re polluting item classes with inventory specific checks

thebigsmileXD:
I am sorry that this PR is too polluting.

dktapps:
why is this closed?

thebigsmileXD:
> 
> 
> why is this closed?

Because the approach was not appreciated.

Issue #2635, opened 2019-01-04
Config/manifest schemas & validation


dktapps:
Currently validation of Config formats is barely present. Where it is present, it's inconsistent, requires a relatively large amount of code, and really isn't nice to look at.

It's currently easy to crash the server with a bad config, or produce undefined behaviour.

Possible solutions:
- PHP config validation library like [symfony/config](https://symfony.com/doc/current/components/config.html)
 - Pros: Allows using PHP code to validate configs
 - Cons: IDEs find it difficult to understand (the library is not well designed), verbose, non-portable, hard to use
- A library implementing [JSON schemas](http://json-schema.org/).

TBC...

dktapps:
Well, I really don't want to reinvent the wheel if a standard solution can be found. I agree the JSON schemas are not very readable.

PR #4493, opened 2021-10-06
Implemented full validation for plugin.yml

dktapps:
## Introduction
This PR uses [wol-soft/php-json-schema-model-generator](https://github.com/wol-soft/php-json-schema-model-generator) to generate code from a JSON Schema spec to handle validation of plugin.yml, without any runtime reflection.

This fixes quite a lot of PHPStan errors with PluginDescription, as well as fixing many crashes and improper behaviour that could result from having incorrect data in plugin.yml.

Now, the server will gracefully fail to load the plugin a la other reasons for failure, instead of crashing when the wrong data types are given.

### Relevant issues
This is an experimental first step towards solving https://github.com/pmmp/PocketMine-MP/issues/2635.

### API changes
- `PermissionParser::loadPermissions()` is removed, since it's no longer useful for anything.
- `PermissionParserException` is removed (only recently added anyway, so I doubt anyone will care).

### Behavioural changes
- The server no longer crashes due to incorrectly structured plugin.yml, and instead provides proper readable errors.
- The `gen` directory must now be included when building a server phar.

## Backwards compatibility
This reduces compatibility slightly:

- `permissions.k.default` is now mandatory for every permission.
- `commands.name.permission` is now mandatory for every command.

However, these are changes that needed to happen anyway.

## Follow-up
Similar schema generation can be done for the following, to improve the robustness of the server and make it easier for users to find mistakes:

- server.properties
- pocketmine.yml
- LoginPacket data (this is really an issue for BedrockProtocol)
- various JSON items in BedrockData
- plugin_list.yml
- resource_packs.yml
- Resource pack manifest.json
- Item JSON format
- recipes.json
- creativeitems.json
- skin geometry?

## Tests
This has not yet been thoroughly tested, however it correctly finds some intentionally-introduced errors in my local plugins, as well as being happy with the ones that are fine.


dktapps:
Due to lack of upstream activity on blocking bugs, this PR can't proceed, so I'm going to close it.

PR #4500, opened 2021-10-10
Fix behavior of the redstone wire when placing it or breaking the block underneath.

IvanCraft623:
## Introduction
This pull request limits the places where the redstone wire block can be placed and now when the block is broken under the redstone wire, it breaks, also dropping the respective item.

### API changes
N/A

### Behavioural changes
- The redstone wire block can now only be placed on non-transparent blocks, with the exceptions: Beacon, Glass, Farmland, Glowstone, GrassPath, HardenedGlass, Hopper, Ice, Melon, SeaLantern and Slime.
- As the block underneath the redstone wire breaks, it will break as well.

## Backwards compatibility
N/A

## Tests
Tested it on Android:
- It was tested in vanilla that transparent blocks are capable of supporting redstone wire and it was ensured that these had the same behavior in pocketmine.
- Breaking the block below also breaks the redstone wire.
- It was ensured that redstone wire can be placed in the slabs if it is type Double or Top.
- The behavior was compared with vanilla and the result was identical.

### DIFF
--- a/src/block/RedstoneWire.php
+++ b/src/block/RedstoneWire.php
@@ -23,11 +23,30 @@ declare(strict_types=1);
 
 namespace pocketmine\block;
 
+use pocketmine\item\Item;
+use pocketmine\math\Facing;
+use pocketmine\math\Vector3;
 use pocketmine\block\utils\AnalogRedstoneSignalEmitterTrait;
 use pocketmine\block\utils\BlockDataSerializer;
+use pocketmine\world\BlockTransaction;
 
 class RedstoneWire extends Flowable{
 	use AnalogRedstoneSignalEmitterTrait;
+
+	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
+		$down = $this->getSide(Facing::DOWN);
+		if($down->isSolid()){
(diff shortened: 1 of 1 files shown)


dktapps:
There has to be some common property of these blocks that we're not seeing. Redstone dust isn't the only block with such behaviour.

src/block/RedstoneWire.php
```diff
@@ -23,12 +23,31 @@
...
 use pocketmine\block\utils\AnalogRedstoneSignalEmitterTrait;
 use pocketmine\block\utils\BlockDataSerializer;
+use pocketmine\world\BlockTransaction;
 
 class RedstoneWire extends Flowable{
 	use AnalogRedstoneSignalEmitterTrait;
 
+	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
+		$down = $this->getSide(Facing::DOWN);
+		if($down->isSolid()){
```
NTT1906: also, check if the block is not a transparent block aka opacity block.

NTT1906: https://minecraft.fandom.com/wiki/Opacity
Java blocks won't have any differences right?

IvanCraft623:
Surpassed by #4886

PR #4501, opened 2021-10-11
Add fletching table

IceCruelStuff:
## Introduction

Registers fletching table. It can be used as fuel in furnace.

### Relevant issues

N/A
### API changes

N/A
### Behavioural changes

Adds fletching tables
## Backwards compatibility

N/A
## Follow-up

N/A
## Tests

https://streamable.com/xlzrkz


dktapps:
Do they not have any functionality?

dktapps (commented):
This also needs to be registered in [`StringToItemParser`](https://github.com/pmmp/PocketMine-MP/blob/master/src/item/StringToItemParser.php) so that the `/give` command can see it.

PR #4507, opened 2021-10-13
Prevent ServerCrash when given ItemId is out of Range

HighTecDev:
Catch InvalidArgumentException too

## Introduction
It leads to a server crash due to an InvalidArgumentException (thrown by Itemfactory::getListOffset() on Line 498) if an ItemID outside the permitted range is entered with the GiveCommand.

### Relevant issues
No issue existing for this

### API changes
No

### Behavioural changes
No

## Backwards compatibility
No

## Follow-up
/

## Tests
Before Changes:
Triggering the crash by executing command "/give <playername> 999999999 <amount>
After Changes:
"Item not Found"-Message will be send to CommandSender


HighTecDev:
Forgot to mention in the description that this is about the GiveCommand

PR #4511, opened 2021-10-17
Allow picking up part of dropped items

### LINKED ISSUE #4499: Implement support for picking up part of an itemstack from a dropped item
## Description
Right now, it's not possible to pick up, say, a dropped item of 64x stone, even if you actually have space for 63 of those stone (e.g. a slot of stone with 1 item in it). This is not much of an issue right now, because the only usual ways to get such a large single dropped item are:

- dropping a slot from your inventory
- breaking a chest or other container

However, it's still problematic.

## Problems
It's inconvenient to implement right now because `Inventory::canAddItem()` doesn't tell us if we could add _some_ of the items in the stack, only if the entire stack can be added. Changing this would require an API break.

## Possible workarounds
On failure to add a stack of some items to an inventory, try again with half the amount. Then try to add half of the items remaining (quarter of the original). Rinse and repeat until you reach a stack size of 1 and can't add 1 more item.

While this sounds clever, it's a workaround that shouldn't really be necessary.

aderoian:
## Introduction
Because of the method `Inventory->canAddItem()` not checking whether *part* of an item can be added to an inventory, players could never pickup part of an item stack. The method `Inventory->getAddableItemQuantity()` which was added not too long ago returns the amount of the item stack that can be added.

`ItemEntity->onCollideWithPlayer()` now uses  `Inventory->getAddableItemQuantity()` to check if an item can be added, and adds that amount, leaving the rest of the items in the item stack. One thing to note: there is a client-sided bug where the item stack texture does not update on the ground to show the new item count. [Example](https://youtu.be/35QzsbO9GWY) of this bug.

### Relevant issues
Fixes #4499

### API changes
none

### Behavioural changes
none

## Backwards compatibility
n/a

## Follow-up
n/a

Requires translations:

n/a

## Tests

Spent some time dropping different amounts of items, and picking them up. Dropped more items than what could be picked up by the player to ensure only part of the item stack was picked up.

Test Video: https://youtu.be/p1ZyEQTgps8

### DIFF
--- a/src/entity/object/ItemEntity.php
+++ b/src/entity/object/ItemEntity.php
@@ -33,7 +33,9 @@ use pocketmine\item\Item;
 use pocketmine\math\Vector3;
 use pocketmine\nbt\tag\CompoundTag;
 use pocketmine\network\mcpe\convert\TypeConverter;
+use pocketmine\network\mcpe\protocol\ActorEventPacket;
 use pocketmine\network\mcpe\protocol\AddItemActorPacket;
+use pocketmine\network\mcpe\protocol\AnimateEntityPacket;
 use pocketmine\network\mcpe\protocol\types\entity\EntityIds;
 use pocketmine\network\mcpe\protocol\types\inventory\ItemStackWrapper;
 use pocketmine\player\Player;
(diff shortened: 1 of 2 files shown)


dktapps:
respawnToAll() doesn't do what you think it does.

Cosmic5173:
> The way vanilla did this back in the day was to despawn the old item entity and create a new one. That's the most portable way to do it.
> 
> You could also try using this (I think it sets the stack count): https://github.com/pmmp/BedrockProtocol/blob/master/src/ActorEventPacket.php#L84

Give me a few moments Ill try these

Cosmic5173:
> This is rather overcomplicated and still lies to `EntityItemPickupEvent` about how many items will be picked up.

Which part(s) are over complicated?

As for "lies to `EntityItemPickupEvent`", I honestly dont have any solution without re-working the event.

Cosmic5173:
> To do it this way wouldn't have required getAddableItemQuantity() at all. canAddItem((clone item)->setCount(1)) would have been enough. But this way doesn't give plugins accurate information.

What if the item was bigger than one?

How do you feel with letting EntityItemPickupEvent determine the left over item if there is one?
When the event sets the item and (item count) > (added item count), drop that left over item even if its different than the original?

Cosmic5173:
Removed the ability to setItem() in EntityPickupItemEvent because it causes unnecessary problems and is not needed. If a plugin *really* wants to chance the item, it can simply despawn the item and add it manually.

dktapps:
You're really overcomplicating this ...

PR #4532, opened 2021-10-31
Check to see if the player can start using the Releasable item.

Rush2929:
## Introduction

Currently, using an item that implements Releasable will start using the item unconditionally. (Server side)
This will cause the following problems

https://user-images.githubusercontent.com/76860328/139593349-ec141c04-5efe-4e0b-8b6c-6d9f9e74f891.mp4

https://user-images.githubusercontent.com/76860328/139593403-887d534d-39b1-45c4-81ef-5c087a613fa4.mp4

This pull request allows Releasable to check if the player is really ready to start using the item.

### API changes

Add `Releasable->canStartUsingItem(Player) : bool`

## Backwards compatibility

Items that implement Releasable need to implement canStartUsingItem.
## Tests

https://user-images.githubusercontent.com/76860328/139594117-68c6a692-079d-4fcc-bcc3-b556501b5262.mp4

(Code used in videos)
```php
$this->getServer()->getPluginManager()->registerEvent(PlayerToggleSneakEvent::class, function(PlayerToggleSneakEvent $ev) : void {
    if ($ev->isSneaking()) {
        $ev->getPlayer()->getHungerManager()->setFood(0.0);
    }
$this->getScheduler()->scheduleRepeatingTask(new ClosureTask(function() : void {
    foreach ($this->getServer()->getOnlinePlayers() as $player) {
        $player->sendTip("IsUsingItem: " . ($player->isUsingItem() ? "true\nDuration: " . $player->getItemUseDuration() : "false"));
    }
}), 1);
```


src/item/Bow.php
```diff
@@ -123,4 +123,9 @@ public function onReleaseUsing(Player $player) : ItemUseResult{
 
 		return ItemUseResult::SUCCESS();
 	}
+
+	public function canStartUsingItem(Player $player) : bool{
+		$arrow = VanillaItems::ARROW();
+		return !$player->isSpectator() && !$player->hasFiniteResources() || $player->getOffHandInventory()->contains($arrow) || $player->getInventory()->contains($arrow);
```
dktapps, line 129:
```suggestion
		return !$player->isSpectator() && (!$player->hasFiniteResources() || $player->getOffHandInventory()->contains($arrow) || $player->getInventory()->contains($arrow));
```

dktapps (changes requested):
Concept ACK, but I don't like littering isSpectator() calls everywhere. Spectator is already checked before calling PlayerItemUseEvent, so these checks are just going to get in the way.

src/item/Food.php
```diff
@@ -41,4 +42,8 @@ public function getAdditionalEffects() : array{
 	public function onConsume(Living $consumer) : void{
 
 	}
+
+	public function canStartUsingItem(Player $player) : bool{
+		return !$player->hasFiniteResources() || !$this->requiresHunger() || $player->getHungerManager()->isHungry();
```
dktapps, line 47: We don't currently support eating in creative. This condition should be consistent with this one: https://github.com/pmmp/PocketMine-MP/blob/c77829f4adee4d5fe61829823f4dbad545c100e9/src/entity/Human.php#L168

PR #4538, opened 2021-11-02
Fixed an issue that could cause food to be eaten at the wrong time.

Rush2929:
## Introduction

Currently, if change the counts of items, etc. while eating food, it will be consumed incorrectly.

https://user-images.githubusercontent.com/76860328/139905346-11ad94f9-aa85-4d93-acb0-2de92347c8a2.mp4

This is solved by suspending the use of the item when a MobEquipmentPacket is received.
## Tests

https://user-images.githubusercontent.com/76860328/139916524-40f3d83c-0ae9-495b-8904-4001be370dbf.mp4


Rush2929:
https://github.com/Rush2929/PocketMine-MP/blob/20d3743ae1d95bda176bc24cdb58d83be35543b0/src/network/mcpe/handler/InGamePacketHandler.php#L493

https://user-images.githubusercontent.com/76860328/139917125-7cdd498b-e5dc-43e1-824d-f22d40379333.mp4


https://user-images.githubusercontent.com/76860328/139917176-d8dba420-203c-45f2-8061-f121239a1df0.mp4

dktapps:
I really don't like this. Whatever the issue is exactly, there has to be a better way than an ugly hack like this.

PR #4544, opened 2021-11-04
Update ActorEventPacket.php

ambiennt:


ambiennt:
+ Added 1.17 actor flags to this PR

dktapps:
what's your source for these?

ambiennt:
> what's your source for these?

BDS

dktapps:
I've added these to BedrockProtocol.

PR #4547, opened 2021-11-05
Implement Trident

IvanCraft623:
## Introduction

Implements item and trident projectile.

### API changes

`Projectile::onHit()` now returns `Vector3` representing the new projectile motion.

### Behavioural changes
`Projectile::onHitEntity()` no longer flags the projectile for despawn, child classes decide whether the entity should be despawned.

## Follow-up

Implement enchantments
Implement custom death attack message for thrown tridents

## Requires translations:
TODO

| Name | Value in eng.ini |
| :--: | :---: |
| `death.attack.thrownTrident` | `{%0} was impaled to death by {%1}` |

## Tests
https://youtu.be/y__tfdHA3FE


redmcteam:
Trident enchantments?

arfush:
@dktapps When you merged it? It's really good implementation.

I thought that it must be in PocketMine-MP

IvanCraft623:
oh no!

dktapps:
That's a github bug, don't worry about it.

IvanCraft623:
I will take up this PR when I have time.

IvanCraft623:
> Trident enchantments?

Enchantments will be implemented in a separate PR:
- [Loyalty](https://github.com/IvanCraft623/PocketMine-MP/tree/loyalty)
- [Impaling](https://github.com/IvanCraft623/PocketMine-MP/tree/impaling)

NhanAZ (dismissed):

dktapps:
I don't love that we're breaking BC to implement this. Is there a way to do it while retaining BC?

Gewinum:
> I don't love that we're breaking BC to implement this. Is there a way to do it while retaining BC?

sometimes sacrifices are to be made

Gewinum:
> I don't love that we're breaking BC to implement this. Is there a way to do it while retaining BC?

the change he made(getnetworktypeid) is done in pm6, isn't it tho? or what are you talking about

IvanCraft623:
> the change he made(getnetworktypeid) is done in pm6, isn't it tho? or what are you talking about

@Gewinum dylan means about  `Projectile::onHit()` return value change

IvanCraft623:
> I don't love that we're breaking BC to implement this. Is there a way to do it while retaining BC?

The BC Break is to be able to modify the motion that the projectile will have when impacting a target, previously it was simply set to zero, but the trident has the fature of "bouncing" when hitting an entity. I thought of an alternative of something like getMotionOnHit(ProjectileHitEvent), but I disliked it and we already had onHit which does not have a return value.

IvanCraft623:
In its beginnings this PR did not contain BCBreak, this was done by creating a new trident inside onHitEntity() with the desired motion. But I discarded that because it felt hacky.

dktapps:
Seems to me like we should enable the use of setMotion in onHit, as that's the conventional way to set motion.

There's also a behavioural BC break wrt. the despawn behaviour when hitting objects. I'm not sure how we could make that non-BC-breaking.

dktapps:
I think we can avoid BC breaks:

- For the post-hit velocity, either add a `getPostHitVelocity(ProjectileHitEvent) : ?Vector3`, or permit the use of `setMotion()` inside the onHit* handlers
- For post-hit despawning, add `despawnsOnHit() : bool` or perhaps an `onPostHit() : void` that can be overridden by subclasses.

However, the branch will need to be rebased if we retarget this back to minor-next, since there's been stuff merged into it from major-next.

src/entity/EntityFactory.php
```diff
@@ -171,6 +173,19 @@ public function __construct(){
...
 		}, ['ThrownPotion', 'minecraft:potion', 'thrownpotion']);
 
+		$this->register(Trident::class, function(World $world, CompoundTag $nbt) : Trident{
+			$itemTag = $nbt->getCompoundTag(Trident::TAG_ITEM);
+			if($itemTag === null){
+				throw new SavedDataLoadingException("Expected \"" . Trident::TAG_ITEM . "\" NBT tag not found");
+			}
+
+			$item = Item::nbtDeserialize($itemTag);
+			if($item->isNull() || !$item instanceof TridentItem){
```
dktapps: Is there a reason to require this to be a TridentItem? It doesn't look like it needs any specific methods.

src/entity/projectile/Projectile.php
```diff
@@ -245,7 +246,9 @@ protected function move(float $dx, float $dy, float $dz) : void{
 			}
 
 			$this->isCollided = $this->onGround = true;
-			$this->motion = Vector3::zero();
+			if ($motionBeforeOnHit->equals($this->motion)) {
```
dktapps:
```suggestion
			if($motionBeforeOnHit->equals($this->motion)){
```

src/entity/projectile/Projectile.php
```diff
@@ -307,7 +310,9 @@ protected function onHitEntity(Entity $entityHit, RayTraceResult $hitResult) : v
 			}
 		}
 
-		$this->flagForDespawn();
+		if ($this->despawnsOnEntityHit()) {
```
dktapps:
```suggestion
		if($this->despawnsOnEntityHit()){
```

src/item/Trident.php
```diff
@@ -0,0 +1,95 @@
...
+		$ev->call();
+		if($ev->isCancelled()){
+			$ev->getEntity()->flagForDespawn();
+			return ItemUseResult::FAIL;
+		}
+		$ev->getEntity()->spawnToAll();
+		$location->getWorld()->addSound($location, new TridentThrowSound());
+
+		if($player->hasFiniteResources()){
+			$item = $entity->getItem();
```
dktapps: Shouldn't this be `$entity->getItem()->pop()`? If there are multiple tridents in a stack, we don't want to give the player back a whole stack of tridents when they pick it up.

dktapps: Should also not be directly passing `$this` in `TridentEntity` for the same reason

src/world/sound/TridentHitSound.php
```diff
@@ -0,0 +1,35 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\world\sound;
+
+use pocketmine\math\Vector3;
+use pocketmine\network\mcpe\protocol\LevelSoundEventPacket;
+use pocketmine\network\mcpe\protocol\types\LevelSoundEvent;
+
+class TridentHitSound implements Sound{
```
dktapps: What's the difference between this and the hit ground sound? These names are obscure.

dktapps (changes requested):

src/entity/EntityFactory.php
IvanCraft623: Consistency, if I do `TridentEntity->getItem()` I would expect a `TridentItem`.

src/item/Trident.php
IvanCraft623: > Shouldn't this be `$entity->getItem()->pop()`?

I assumed Durable items cannot be stacked, but aparently they can on PM :P

src/item/Trident.php
```diff
@@ -69,7 +69,6 @@ public function onReleaseUsing(Player $player, array &$returnedItems) : ItemUseR
 			$item = $entity->getItem();
 			$item->applyDamage(1);
```
dktapps: I feel like it would make more sense to damage the item before passing it to TridentEntity constructor instead of having to get/set it here

src/entity/EntityFactory.php
dktapps: It just seems like an unnecessary constraint to me, considering there's no special methods on `TridentItem`.

src/entity/projectile/Trident.php
```diff
@@ -0,0 +1,180 @@
...
+		$shouldDespawn = false;
+
+		$playerInventory = $player->getInventory();
+		$ev = new EntityItemPickupEvent($player, $this, $this->getItem(), $playerInventory);
+		if($player->hasFiniteResources() && !$playerInventory->canAddItem($ev->getItem())){
+			$ev->cancel();
+		}
+		if($this->spawnedInCreative){
+			$ev->cancel();
+			$shouldDespawn = true;
```
dktapps, line 159: This is going to despawn tridents thrown by creative players if a survival player tries to pick them up, even though no item will be received. Doesn't seem right to me.

dktapps (changes requested):

dktapps:
This PR seems mostly done, just a couple things left to deal with.

src/entity/projectile/Trident.php
IvanCraft623: That's the expected behavior

dktapps: Expected by who?

dktapps: I'm pretty sure this is a bug in Bedrock. No other items behave like this.
Arrows fired in creative aren't expected to despawn when a survival player steps on them.

Also, from the wiki:

> Only the owner of a trident thrown in Creative mode or enchanted with Loyalty can pick it up; other players in Creative and players in Survival mode including its own owner can't pick up the trident thrown in Creative.

I would expect this to work the same way arrows do, with modes NONE/ANY/CREATIVE pickup modes.

dktapps: For posterity, seems this really is intentional, and it's also this way for arrows in Bedrock, which raises the question how to implement this in PM, considering that we already have the Java-style behaviour for arrows.

dktapps (approved):

smartcmd:
wow

b1zeyofficial:
> > Trident enchantments?
> 
> Enchantments will be implemented in a separate PR:
> 
> * [Loyalty](https://github.com/IvanCraft623/PocketMine-MP/tree/loyalty)
> * [Impaling](https://github.com/IvanCraft623/PocketMine-MP/tree/impaling)

are you still planning to do PR on these topics?

PR #4562, opened 2021-11-09
Added CustomSound and CustomParticle

Govdim:
## Introduction
Added the ability to send custom sounds and particles to players using internal classes

### API changes
Added CustomSound class
Added CustomParticle class


dktapps:
These implicitly expose protocol to the API. If you want to use these, you should stick to packets and stop kidding yourself that you're not using the protocol.

Issue #3130, opened 2019-09-30
Entity->setImmobile() does not immobilize entities


dktapps:
The API method `Entity->setImmobile()` was introduced in 1.6.1 or so. This API method is misleading because it only sets a `NO_AI` flag on the network layer, and does not actually prevent the entity from moving.

dktapps:
These functions have been renamed by edafe9d21fd2b72b6baa6cb2c879abd6a263a446 to more accurately reflect their purpose. If there is demand, we can also add the ability to server-side immobilize an entity too, but this isn't going to happen for a 5.0 release.

PR #4563, opened 2021-11-10
Player: allow controlling client-sided block collisions irrespective of Spectator Mode

buchwasa:
## Introduction
Presently, PocketMine does not provide a simple way to set if players have the ability to ignore block collision or not

## Changes
To maintain compatibility with spectators, in internalSetGameMode, the block collision flag is toggled

### API changes
The following variables have been implemented:
- protected Player->blockCollision

The following methods have been implemented:
- Player->setHasBlockCollision(bool) : void
- Player->hasBlockCollision() : bool

### DIFF
--- a/src/network/mcpe/NetworkSession.php
+++ b/src/network/mcpe/NetworkSession.php
@@ -793,8 +793,8 @@ class NetworkSession{
 			$for->getId()
 		);
 
-		$pk->setFlag(AdventureSettingsPacket::WORLD_IMMUTABLE, $for->isWorldImmutable());
-		$pk->setFlag(AdventureSettingsPacket::NO_PVP, $for->getNoPvP());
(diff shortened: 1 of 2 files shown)


dktapps:
Hi, apart from `setNoClip()` I don't think these make much sense in their current form. As it stands, they would end up being another `setImmobile()` problem (see #3130 ) because they only influence client behaviour. Plugin devs would expect these to, for example, prevent the player from PvPing server-side (it doesn't) and prevent the player from building (it doesn't). I think these would get used inappropriately by developers who would misunderstand their intended purpose.

buchwasa:
Makes sense, I'll go ahead and drop everything but no clip. My original intent was to only do no clip, but I wasn't sure how leaving the other 2 alone would go over, now I know.

dktapps:
I think it should probably get a better name if we want to put it on the API also. "no clip" is hard even for English speakers to make sense of.

PR #4574, opened 2021-11-14
Fix position of setworldspawn command

Nerahikada:
## Introduction
~~`setworldspawn` command is now consistent with vanilla behavior.~~
Now, the world's spawn position is set to the floored value instead of the rounded value.

### Behavioural changes
World spawn position is no longer using rounded values.


dktapps:
floor != truncated. There is a difference in the negatives.

Nerahikada:
Oops, you are exactly right.
My explanation is bad, it uses the floor value and not the truncated value.

Nerahikada:
Sorry for my lack of research, but it seems that the input value must also be a floor value.
Is it better to change the `getInteger` function? Or should I use another method?

dktapps:
floor(getDouble()) would be fine.

PR #4575, opened 2021-11-14
Allow input of relative coordinates to setworldspawn command

Nerahikada:
## Introduction
This is actually a continuation of #4574.
The input value is now converted to a floor value, and implemented the input of relative coordinates.

### Behavioural changes
- Relative coordinates can now be used.
- Target world is now always the world of the player who executed the command.
- If the command is executed from the console, the default world will be targeted.
  - If relative coordinates are entered, the coordinates (0,0,0) will be set as the base point. (This behavior is based on BDS.)

### DIFF
--- a/src/command/defaults/SetWorldSpawnCommand.php
+++ b/src/command/defaults/SetWorldSpawnCommand.php
@@ -31,6 +31,7 @@ use pocketmine\math\Vector3;
 use pocketmine\permission\DefaultPermissionNames;
 use pocketmine\player\Player;
 use pocketmine\utils\TextFormat;
+use pocketmine\world\World;
 use function count;
 
 class SetWorldSpawnCommand extends VanillaCommand{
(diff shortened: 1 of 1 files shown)


src/command/defaults/SetWorldSpawnCommand.php
```diff
@@ -60,8 +60,18 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
...
+			if($sender instanceof Player){
+				$base = $sender->getPosition();
+				$world = $base->getWorld();
+			}else{
+				$base = new Vector3(0.0, 0.0, 0.0);
+				$world = $sender->getServer()->getWorldManager()->getDefaultWorld();
+			}
+			$pos = (new Vector3(
+				$this->getRelativeDouble($base->x, $sender, $args[0]),
+				$this->getRelativeDouble($base->y, $sender, $args[1], 0, 256),
```
dktapps, line 72: Do not hardcode the bounds. Use World::Y_MIN/MAX.

PR #4576, opened 2021-11-15
Fixed saplings cannot be placed on farmland

### LINKED ISSUE #4528: Saplings cannot be placed on farmland
### Issue description
If the saplings are placed on farmland, they will be destroyed.

1. Place any saplings on farmland

### OS and versions
test.pmmp.io:19132
* PocketMine-MP: 63dfcc6

Nerahikada:
## Introduction
Fixed an issue that caused saplings to be destroyed only when they were placed on farmland.

### Relevant issues
* Fixes #4528

### Behavioural changes
- Saplings can now be placed correctly on farmland and podzol.

### DIFF
--- a/src/block/Sapling.php
+++ b/src/block/Sapling.php
@@ -71,7 +71,8 @@ class Sapling extends Flowable{
 			$block instanceof Dirt ||
 			$block instanceof Grass ||
 			$block instanceof Podzol ||
-			$block instanceof Farmland;
+			$block instanceof Farmland ||
+			$block instanceof Mycelium;
 	}
 
 	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
--- a/src/block/Sapling.php
+++ b/src/block/Sapling.php
@@ -66,9 +66,16 @@ class Sapling extends Flowable{
 		return $this;
 	}
 
+	private function canBeSupportedBy(Block $block) : bool{
+		return
+			$block instanceof Dirt ||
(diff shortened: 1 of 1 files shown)


dktapps:
I'd prefer if the logic for this was placed in a separate method, instead of being duplicated.

src/block/Sapling.php
```diff
@@ -66,9 +66,16 @@ public function setReady(bool $ready) : self{
 		return $this;
 	}
 
+	private function canBeSupportedBy(Block $block) : bool{
+		return
+			$block instanceof Dirt ||
+			$block instanceof Grass ||
+			$block instanceof Podzol ||
+			$block instanceof Farmland;
```
NTT1906, line 74: add mycelium

NTT1906: and some new dirt variants...

Nerahikada: It seems that mycelium can be placed from 1.18

Nerahikada: This should be enough for now for the blocks implemented on pmmp

src/block/Sapling.php
```diff
@@ -66,9 +66,17 @@ public function setReady(bool $ready) : self{
 		return $this;
 	}
 
+	private function canBeSupportedBy(Block $block) : bool{
+		return
+			$block instanceof Dirt ||
+			$block instanceof Grass ||
+			$block instanceof Podzol ||
+			$block instanceof Farmland ||
+			$block instanceof Mycelium;
```
dktapps, line 75: instanceof should not be used for this

dktapps:

PR #4578, opened 2021-11-17
Add posibility of CommandParameters

Crasher508:
## Introduction
Add posibility of CommandParameters

### Relevant issues
Can't auto complete commands

### API changes
add Command->getData()


dktapps:
This serves no practical purpose and mixes protocol into the API without actually doing anything useful.

PR #4601, opened 2021-11-27
Add constants for deviceos data received from getExtraData()

Xd-pro:
## Introduction
You should be able to check this data more easily.

### API changes
Added constants to PlayerInfo for device operating systems.

## Follow-up
Add constants for input methods.

## Tests
Tests are not necessary, these changes can't break anything, but I have checked the values of the device operating systems by joining with different devices. Syntax has been checked in PhpStorm.


pooooooon:
here
https://github.com/pmmp/BedrockProtocol/blob/master/src/types/DeviceOS.php

PR #4610, opened 2021-12-01
Humans can now do emotes

IvanCraft623:
## Introduction

It will allow plugin-developers to make their Human entities emote using the PM api.

### API changes

`Human::emote()`


arfush:
It's a good idea, but it can add a special interface/enumeration/registry in which to describe the available emotions.

IvanCraft623:
Merge?

dktapps:
NB: This is technically a BC break, but since the NetworkSession is not considered part of the API, this is OK for next-minor.

PR #4613, opened 2021-12-03
getType() of SplashPotion

xxAROX:


xDidntPot:
format...

xxAROX:
@DidntPot its same format like Potion.php, so its valid

PR #4617, opened 2021-12-05
Implemented fire spread (take 2)

### LINKED ISSUE #1424: Fire doesn't check the correct block to attach.
### Issue description

The fire can be ignited on all solid blocks. In this way the fire is consired as a ground, so, it doesn't permit the passing of falling blocks.

Already tested with @dktapps.

1. Ignite a fire into a block side
2. The fire is invisible.

### LINKED ISSUE #1779: Fire burns an incredibly long time
### Issue description
- Expected result: Fire must be extinguished within 30-40 seconds
- Actual result: Fire is burning more than a few minutes

1. join to server
2. Click flint and steel on any block

### OS and versions
* PocketMine-MP: https://github.com/pmmp/PocketMine-MP/commit/51cec525eeca1d57f525cbe75f7cb4d8a4bdff4f

dktapps:
## Introduction
This is a rehash of the remaining parts of #2107.

### Relevant issues
- fixes #1424
- fixes #1779
- fixes a bunch of other issues with fire misbehaving or being weak

### API changes
None, but `BlockSpreadEvent` will be called in more cases (every time fire spreads to a new block).

### Behavioural changes
- Fire now spreads more like vanilla.
- `BlockSpreadEvent` is now called in more cases.

## Backwards compatibility
None.

## Follow-up
- Lava should set nearby things on fire.

## Tests
This has been tested many times over the last 3 years and its issues (those that I know about) have now all been resolved.


src/block/Fire.php
```diff
@@ -186,6 +192,53 @@ private function burnBlock(Block $block, int $chanceBound) : void{
...
+			//Higher blocks have a lower chance of catching fire
+			$randomBound = 100 + ($y > 1 ? ($y - 1) * 100 : 0);
+
+			for($z = -1; $z <= 1; ++$z){
+				for($x = -1; $x <= 1; ++$x){
+					if($x === 0 and $y === 0 and $z === 0){
+						continue;
+					}
+
+					if(!$world->isChunkLoaded(($this->position->x + $x) >> Chunk::COORD_BIT_SIZE, ($this->position->z + $z) >> Chunk::COORD_BIT_SIZE)){
```
dktapps, line 210: This should probably include an `isInWorld()` check too.

src/block/Fire.php
```diff
@@ -186,6 +192,53 @@ private function burnBlock(Block $block, int $chanceBound) : void{
...
 	}
 
+	private function spreadFire() : void{
+		$world = $this->position->getWorld();
+		$difficulty7 = $world->getDifficulty() * 7;
+		$age30 = $this->age + 30;
+
+		for($y = -1; $y <= 4; ++$y){
+			//Higher blocks have a lower chance of catching fire
+			$randomBound = 100 + ($y > 1 ? ($y - 1) * 100 : 0);
```
dktapps, line 202: This should use `World::Y_MIN`, or fire may not work correctly after the caves expansion is implemented.

PR #4623, opened 2021-12-07
Add ExperienceOrb attracts lock feature

### LINKED ISSUE #4589: ExperienceOrb: no possibility of interfering in search for target entity
## Description
Currently, you have the possibility by listening for `PlayerExperienceChangeEvent`, stop a player, or human entity to be more specific, from receiving experience.
But by the way the experience orb entity searches for its target, there is no way for plugin developers to interfere and for example, block experience orbs to be attracted to certain players / human entities, apart from extending the whole ExperienceOrb class and overwriting the `ExperienceOrb::entityBaseTick()` method.
https://github.com/pmmp/PocketMine-MP/blob/master/src/entity/object/ExperienceOrb.php#L179#L175

## Justification
There is a variety of situations, where you wouldn't want experience orbs being attracted to specific entities, for example:
- Human entities are an easy way to create non-player entities with player skins, for example in the Slapper plugin or player skull plugins (e.g.: https://github.com/ColinHDev/CSkull/blob/main/src/ColinHDev/CSkull/entities/SkullEntity.php#L21). In these usages, you would not want experience orbs to be attracted to your entities and rather ignore them entirely.
- When writing a plugin that allows your staff members to go completely invisible, without them going into spectator mode and pretending they left the server, you would not want experience orbs flying to your hidden staff members, which would expose them.

## Possible solutions
- One suggestion was a `Human::canPickupExperience()` method which could be overwritten by plugins in their entity classes. But this suggestion has two flaws in my opinion: First, the method could be misleading with `ExperienceManager::canPickupXp()` and second, this would work for those kinds of plugins listed in the first example. But for that method to be useful for plugins listed in the second example, they would need to overwrite the Player class, which is also rather impractical, even more than overwriting the ExperienceOrb class. (The second point can be ruled out if such a method would be implemented with a class attribute, a setter and getter like `Entity::hasGravity()` and `Entity::setHasGravity()`.)
- Another suggestion was an `ExperienceOrbAttractEvent`, which could be called before the following line https://github.com/pmmp/PocketMine-MP/blob/master/src/entity/object/ExperienceOrb.php#L178. This event would be very useful for both given plugin examples, as they could just cancel the event or maybe also provide a replacement entity with an inbuilt `ExperienceOrbAttractEvent::setEntity()` method, so that these plugins could not only remove experience orbs from being attracted to their entities but also give plugin developers a way to implement their own system of searching.

## Alternative methods
- It should be possible to create your own ExperienceOrb class which overwrites the `ExperienceOrb::entityBaseTick()` method so that you would be able to make a custom search, but that approach is rather impractical.
- In the case of the first example, it would also be possible for plugins to make their entities an instance of `Living` or only `Entity` but not `Human` because experience orbs only search for human entities. But this is also rather impractical because you would end up just copying many methods like skin and inventory handling from the Human to your own class.

ShockedPlot7560:
## Introduction
As explained in the issue #4589 , there is a need to allow developers to remove players from the attraction of xp orbs. 

### Relevant issues
* Fixes #4589

### API changes
The following methods have been added to the ``ExperienceManager`` :
* ``isAttractsLocked`` : Returns true if the Human no longer accepts xp orbs attracts, false otherwise
* ``setAttractsLocked`` : The setter to change the acceptance status of xp orbs

### Behavioural changes
In ``ExperienceOrb``:
* The ``getTargetPlayer`` method will no longer return an instance of ``Human`` if ``isAttractsLocked`` return true for the instance
* The ``setTargetPlayer`` method will no longer choose a ``Human`` that returns true to ``isAttractsLocked``
* In ``entityBaseTick`` the orb will no longer choose a target that returns true to ``isAttractsLocked``

## Follow-up
No specific action is required

## Tests
* To test the normal system, mine a block of diamond ore in survival and the orb will indeed follow you.
* To test the locked system, create a plugin that disables orbs by default for any given player and it will look like this:
image

### DIFF
--- a/src/entity/object/ExperienceOrb.php
+++ b/src/entity/object/ExperienceOrb.php
@@ -145,7 +145,7 @@ class ExperienceOrb extends Entity{
 		}
 
 		$entity = $this->getWorld()->getEntity($this->targetPlayerRuntimeId);
-		if($entity instanceof Human && !$entity->getXpManager()->isAttractsLocked()){
+		if($entity instanceof Human){
 			return $entity;
 		}
 
@@ -153,11 +153,7 @@ class ExperienceOrb extends Entity{
 	}
 
 	public function setTargetPlayer(?Human $player) : void{
-		if(($player instanceof Human && !$player->getXpManager()->isAttractsLocked()) or $player === null){
-			$this->targetPlayerRuntimeId = $player !== null ? $player->getId() : null;
-		}else{
-			$this->targetPlayerRuntimeId = null;
-		}
+		$this->targetPlayerRuntimeId = $player !== null ? $player->getId() : null;
 	}
 
 	protected function entityBaseTick(int $tickDiff = 1) : bool{
@@ -170,7 +166,7 @@ class ExperienceOrb extends Entity{
 		}
 
 		$currentTarget = $this->getTargetPlayer();
-		if($currentTarget !== null and (!$currentTarget->isAlive() or $currentTarget->location->distanceSquared($this->location) > self::MAX_TARGET_DISTANCE ** 2)){
+		if($currentTarget !== null and (!$currentTarget->isAlive() or $currentTarget->getXpManager()->isAttractsLocked() or $currentTarget->location->distanceSquared($this->location) > self::MAX_TARGET_DISTANCE ** 2)){
 			$currentTarget = null;
 		}
 
--- a/src/entity/object/ExperienceOrb.php
+++ b/src/entity/object/ExperienceOrb.php
@@ -145,7 +145,7 @@ class ExperienceOrb extends Entity{
 		}
 
 		$entity = $this->getWorld()->getEntity($this->targetPlayerRuntimeId);
-		if($entity instanceof Human){
+		if($entity instanceof Human && !$entity->getXpManager()->isAttractsLocked()){
 			return $entity;
 		}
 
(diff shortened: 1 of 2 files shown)


ColinHDev:
Unexpected, but an interesting-looking solution.

src/entity/object/ExperienceOrb.php
```diff
@@ -145,15 +145,19 @@ public function getTargetPlayer() : ?Human{
...
 		return null;
 	}
 
 	public function setTargetPlayer(?Human $player) : void{
-		$this->targetPlayerRuntimeId = $player !== null ? $player->getId() : null;
+		if(($player instanceof Human && !$player->getXpManager()->isAttractsLocked()) or $player === null){
+			$this->targetPlayerRuntimeId = $player !== null ? $player->getId() : null;
+		}else{
+			$this->targetPlayerRuntimeId = null;
+		}
```
ColinHDev, line 160: This looks really overcomplicated. This would be much simpler
```php
if($player !== null && !$player->getXpManager()->isAttractsLocked()){
			$this->targetPlayerRuntimeId = $player->getId();
		}else{
			$this->targetPlayerRuntimeId = null;
		}
```
or
```php
$this->targetPlayerRuntimeId = $player !== null && !$player->getXpManager()->isAttractsLocked() ? $player->getId() : null;
```

But does this check even belong there? In every case, where pm would set the target player itself, it does the checks alone. So shouldn't plugin devs be required to do their own checks? Possibly someone wants to bypass the lock of another plugin (whyever).

dktapps: No, it doesn't belong here. If the player can't be tracked, the orb should try to find a different target.

dktapps: This whole change is redundant.

dktapps (commented):
Code looks OK (as in it will work) but `attractsLocked` is a really abysmal name.

ShockedPlot7560:
Wouldn't it be better to call it :
* attractsExperienceOrb
or 
* orbSeverity
> It must be made clear in the name that this property must be able to be used to cut the gravitation of the orbs around the player and remove his ability to recover them

I otherwise agree that the changes in ``setTargetPlayer`` or ``getTargetPlayer`` its rather redundant and deserves to be removed to allow the developers more freedom.

dktapps:
`canAttractsXpOrbs` and `setCanAttractXpOrbs` would be better (for consistency with other methods).

ShockedPlot7560:
Indeed, it makes a little more sense

dktapps (approved):
Tested and found working.

Issue #458, opened 2017-03-24
Torches cannot be placed on top of some blocks


dktapps:
### Issue description
Torches cannot be placed on any transparent block other than fences or cobblestone walls.
http://minecraft.gamepedia.com/Torch#Placement

### Steps to reproduce the issue

1. place glass block
2. place a torch on top of it

### OS and versions

* PocketMine-MP: #454
* PHP: 7.0.16
* Server OS: w10
* Game version: Win10

flaxues:
Beds also can't be placed on slabs

thebigsmileXD:
Effects:

- [ ] Flowers
- [ ] Mushrooms
- [ ] Torches
- [ ] Redstone-Torches
- [ ] Signs
- [ ] Pressure plates
- [ ] Flower Pots

jasonwynn10:
I don't think beds can be placed on beds in vanilla anyway @Matthww

dktapps:
Relates to #4856, fixed by #4886

PR #4634, opened 2021-12-09
Flower pot

ShockedPlot7560:
## Introduction
The current problem is that the flowerpots have been implemented but at their functional limit. For example, we cannot currently put them on anything that is stairs in a high position / slab etc. But also, some plants weren't allowed to go there even though the vanilla version of minecraft allowed them.

### Relevant issues
This pull request is slightly related to #458 due to the fact that it patches this part for flower pots

### API changes
Several functions have been added to the FlowerPot class to check the contents of the flowerpot and the support underneath it.

### Behavioural changes
Apart from the fact that the flower pots can be placed on a larger number of blocks, no

## Follow-up
This pull request is entirely related to the flowerpot but many other blocks are in the same case like the torches.
It can be converted into a whole change and a fix of all these blocks to fix this problem that has been present for quite some time.

## Tests
The tests are quite simple, use the flowerpots as if you were using them in vanilla minecraft, on the slabs, stairs, and other transparent blocks that support it

### DIFF
--- a/src/block/FlowerPot.php
+++ b/src/block/FlowerPot.php
@@ -114,6 +114,10 @@ class FlowerPot extends Flowable{
 		}elseif(!$this->canAddPlant($plant)){
 			return false;
 		}else{
+			if($plant instanceof BambooSapling){
+				//Hack to convert bamboosapling to bamboo block (different id)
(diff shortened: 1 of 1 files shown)


dktapps:
Obsoleted by #4886.

PR #4638, opened 2021-12-11
Add gamemode strings as valid values for gamemode property

### LINKED ISSUE #2692: Make `gamemode` key in server.properties accept gamemode names instead of magic numbers

sylvrs:
## Introduction
At the moment, users must rely on the magic numbers surrounding gamemodes (0 => Survival, 1 => Creative, etc.) in the configuration. This has been an issue for a long time. This pull request hopes to resolve that by migrating to the string forms of the gamemodes.

### Relevant issues
Resolves #2692

### API changes
N/A

### Behavioural changes
N/A

## Backwards compatibility
The largest issue with this change would be migration from the magic numbers to the strings. By default, if the value can't be resolved to a valid gamemode, it will return `GameMode::SURVIVAL()`. This is undesired behavior, especially for any servers that have the default gamemode set to something other than survival. Two solutions to this issue would be:
1. The server sends a warning message if the gamemode string is out of date.
2. The server automatically migrates the magic numbers to the string values.

The downside of both of these solutions is the temporary code behind it. A version would have to be determined to remove that from the source.

Update: 
GameMode::fromString() accepts magic numbers as aliases so this pull request is fully backwards compatible with previous versions.

## Follow-up
~~While not directly related, it may be worth looking at updating SetupWizard to use the gamemode strings as well. With this pull request, the SetupWizard is configured to parse the ID back into a gamemode and then into a name.~~

No follow up is needed for this pull request.

Requires translations:
Added translation `gamemode_options` (pmmp/Language#99)

## Tests
To test, I set up a new server environment and run the wizard. Here is the output of the server.properties from the wizard:
```properties
#Properties Config file
#Sat Dec 11 02:09:16 CST 2021
language=eng
motd=Testing Server
server-name=Testing Server
server-port=19132
gamemode=creative
max-players=300
white-list=off
enable-query=on
server-portv6=19133
enable-ipv6=on
force-gamemode=off
hardcore=off
pvp=on
difficulty=2
generator-settings=
level-name=world
level-seed=
level-type=DEFAULT
auto-save=on
view-distance=8
xbox-auth=on
```

Changing the `gamemode` property to `spectator` results in this successful message from the console:
`[02:11:48.452] [Server thread/INFO]: Default game type: Spectator Mode`

Inputting a number into the `gamemode` property automatically resolves to survival mode (as any other invalid value would).

The setup wizard displays and parses the gamemodes correctly. Here is the output from the setup:
```
[*] You are going to set up your server now.
[*] If you don't want to change the default value, just press Enter.
[*] You can edit them later on the server.properties file.
[?] Give a name to your server (PocketMine-MP Server):
[*] Do not change the default port value if this is your first server.
[?] Server port (19132):
[?] Select one of the following game modes: [survival, creative, adventure, spectator] (survival): adv
[?] Select one of the following game modes: [survival, creative, adventure, spectator] (survival): spe
[?] Select one of the following game modes: [survival, creative, adventure, spectator] (survival): creative
[?] Max. online players (20):
[*] An OP is the player admin of the server. OPs can run more commands than normal players
[?] OP player name (example, your game name):
[*] The white-list only allows players in it to join.
[?] Do you want to enable the white-list? (y/N):
[!] Query is a protocol used by different tools to get information of your server and players logged in.
[!] If you disable it, you won't be able to use server lists.
[?] Do you want to disable Query? (y/N):
[*] Getting your external IP and internal IP
[!] Your external IP is xxx.xxx.xxx.xxx. You may have to port-forward to your internal IP xxx.xxx.xxx.xxx
[!] Be sure to check it, if you have to forward and you skip that, no external players will be able to join. [Press Enter]
```

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -36,7 +36,6 @@ use pocketmine\crafting\CraftingManager;
 use pocketmine\crafting\CraftingManagerFromDataHelper;
 use pocketmine\crash\CrashDump;
 use pocketmine\crash\CrashDumpRenderer;
-use pocketmine\data\java\GameModeIdMap;
 use pocketmine\entity\EntityDataHelper;
 use pocketmine\entity\Location;
 use pocketmine\event\HandlerListManager;
--- a/src/Server.php
+++ b/src/Server.php
@@ -357,7 +357,7 @@ class Server{
 	}
 
 	public function getGamemode() : GameMode{
-		return GameModeIdMap::getInstance()->fromId($this->configGroup->getConfigInt("gamemode", 0)) ?? GameMode::SURVIVAL();
+		return GameMode::fromString($this->configGroup->getConfigString("gamemode", GameMode::SURVIVAL()->name())) ?? GameMode::SURVIVAL();
 	}
 
 	public function getForceGamemode() : bool{
@@ -804,7 +804,7 @@ class Server{
 					"enable-ipv6" => true,
 					"white-list" => false,
 					"max-players" => 20,
-					"gamemode" => 0,
+					"gamemode" => GameMode::SURVIVAL()->name(),
 					"force-gamemode" => false,
 					"hardcore" => false,
 					"pvp" => true,
--- a/src/wizard/SetupWizard.php
+++ b/src/wizard/SetupWizard.php
@@ -159,8 +159,6 @@ LICENSE;
 		}while(true);
 		$config->set("server-port", $port);
 
-		$this->message($this->lang->translate(KnownTranslationFactory::gamemode_info()));
-
 		$selection = "[" . implode(", ", array_map(static fn(GameMode $gameMode): string => $gameMode->name(), GameMode::getAll())) . "]";
 		do{
 			$gamemode = GameMode::fromString($this->getInput($this->lang->translate(KnownTranslationFactory::gamemode_options($selection)), GameMode::SURVIVAL()->name()));
--- a/src/wizard/SetupWizard.php
+++ b/src/wizard/SetupWizard.php
@@ -164,7 +164,7 @@ LICENSE;
 		do{
 			$gamemode = (int) $this->getInput($this->lang->translate(KnownTranslationFactory::default_gamemode()), (string) GameModeIdMap::getInstance()->toId(GameMode::SURVIVAL()));
 		}while($gamemode < 0 or $gamemode > 3);
-		$config->set("gamemode", $gamemode);
+		$config->set("gamemode", GameModeIdMap::getInstance()->fromId($gamemode)->name());
 
 		$config->set("max-players", (int) $this->getInput($this->lang->translate(KnownTranslationFactory::max_players()), (string) self::DEFAULT_PLAYERS));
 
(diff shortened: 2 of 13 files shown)


src/Server.php
```diff
@@ -357,7 +356,7 @@ public function getServerUniqueId() : UuidInterface{
 	}
 
 	public function getGamemode() : GameMode{
-		return GameModeIdMap::getInstance()->fromId($this->configGroup->getConfigInt("gamemode", 0)) ?? GameMode::SURVIVAL();
+		return GameMode::fromString($this->configGroup->getConfigString("gamemode", GameMode::SURVIVAL()->name())) ?? GameMode::SURVIVAL();
```
dktapps, line 359: This will break existing configs.

src/wizard/SetupWizard.php
```diff
@@ -162,9 +162,10 @@ private function generateBaseConfig() : void{
 		$this->message($this->lang->translate(KnownTranslationFactory::gamemode_info()));
 
 		do{
-			$gamemode = (int) $this->getInput($this->lang->translate(KnownTranslationFactory::default_gamemode()), (string) GameModeIdMap::getInstance()->toId(GameMode::SURVIVAL()));
-		}while($gamemode < 0 or $gamemode > 3);
-		$config->set("gamemode", $gamemode);
+			$gamemodeInput = (int) $this->getInput($this->lang->translate(KnownTranslationFactory::default_gamemode()), (string) GameModeIdMap::getInstance()->toId(GameMode::SURVIVAL()));
```
dktapps, line 165: I'd rather just ditch the IDs and use GameMode::fromString() exclusively, so that the aliases used can be the same used by the /gamemode command. Less confusing for new users that way.

src/Server.php
sylvrs: I mentioned this in my original post. Do you want to accept both ints and strings, or have some sort of migration system?

src/wizard/SetupWizard.php
sylvrs: I also mentioned this in my original post. I definitely agree that it would be better to just use the string version, but we would need to update language files to match.

src/Server.php
dktapps: I'm not sure. But having numeric gamemodes resolve to survival is definitely unexpected behaviour.

sylvrs: I believe that something like this would work:
`[11:48:01.017] [Server thread/WARNING]: Outdated value "1" for "gamemode" property was found in server.properties. It has been migrated to "creative".`

The only thing holding that back would also be implementing it into the language files.

dktapps: Actually, strike everything I said; GameMode registers numeric aliases.

src/wizard/SetupWizard.php
sylvrs: Would it be worth having `gamemode_info` in the language files accept arguments of all of the gamemode names?

dktapps: A single parameter which accepts `implode(", ", gameModeNames)` would be enough, I think.

dktapps:
Base branch has been updated to include new translations.

PR #4648, opened 2021-12-13
Add a way to explicitly blame a plugin for crashing the server

dktapps:
## Introduction
We often see cryptic crashes like these: https://crash.pmmp.io/view/5617460

These are actually caused by bogus plugin code, but due to a PHP bug, they are reported in PluginManager instead of at their actual location.
This is not likely to be addressed in the near future (it's been a bug for a very long time), and in the meantime it's adding pollution to the crash archive.

A new crashdump plugin_involvement 'blame' status is added, which the crash archive will need to support. I think for now it's important to distinguish between these at the CA level, so that any bugs in the call proxy can be identified correctly.

### API changes
No changes to public API.

A new class `BlamePluginCallProxy` is added for internals use.

### Behavioural changes
Errors that occur while constructing a plugin's main class will now be correctly blamed on the plugin.

## Follow-up
We may also want to extend this to the following areas:

- plugin.yml parsing (debatable; we should be validating it properly anyway)
- command permission setting (this is the source of a lot of PM4 "core" crashes right now)

## Tests
```php
	public const TEST = \pocketmine\item\Item::IRON_INGOT;
}
```

Before the patch: 
[Mon_Dec_13-19.52.31-GMT_2021.log](https://github.com/pmmp/PocketMine-MP/files/7706672/Mon_Dec_13-19.52.31-GMT_2021.log)

After the patch:
[Mon_Dec_13-19.53.06-GMT_2021.log](https://github.com/pmmp/PocketMine-MP/files/7706675/Mon_Dec_13-19.53.06-GMT_2021.log)

### DIFF
--- a/src/crash/BlamePluginCallProxy.php
+++ b/src/crash/BlamePluginCallProxy.php
@@ -1,0 +1,60 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
+ *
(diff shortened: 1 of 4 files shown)


dktapps:
Upstream issue reported as https://github.com/php/php-src/issues/7771

dktapps:
Upstream issue was fixed in 8.2.0.

dktapps:
Obsolete now that primary PHP version is 8.2 (with 8.1 to be dropped soon™️)

PR #4666, opened 2021-12-19
Implement BlockMeltEvent

dries-c:
## Introduction
Currently, there's no way to stop Ice from melting using events since the BlockBreakEvent isn't called due to it not being Player related.

### API changes
Add `BlockMeltEvent` extending `BaseBlockChangeEvent`

### DIFF
--- a/src/block/Ice.php
+++ b/src/block/Ice.php
@@ -55,7 +55,7 @@ class Ice extends Transparent{
 			$ev = new BlockMeltEvent($this, VanillaBlocks::WATER());
 			$ev->call();
 			if(!$ev->isCancelled()){
-				$this->position->getWorld()->setBlock($this->position, $ev->getNewState(), false);
+				$this->position->getWorld()->setBlock($this->position, $ev->getNewState());
 			}
 		}
 	}
--- a/src/block/Ice.php
+++ b/src/block/Ice.php
@@ -23,7 +23,7 @@ declare(strict_types=1);
 
 namespace pocketmine\block;
 
-use pocketmine\event\block\IceMeltEvent;
+use pocketmine\event\block\BlockMeltEvent;
 use pocketmine\item\enchantment\VanillaEnchantments;
 use pocketmine\item\Item;
 use pocketmine\player\Player;
@@ -52,10 +52,10 @@ class Ice extends Transparent{
 
 	public function onRandomTick() : void{
 		if($this->position->getWorld()->getHighestAdjacentBlockLight($this->position->x, $this->position->y, $this->position->z) >= 12){
-			$ev = new IceMeltEvent($this);
+			$ev = new BlockMeltEvent($this, VanillaBlocks::WATER());
 			$ev->call();
 			if(!$ev->isCancelled()){
-				$this->position->getWorld()->useBreakOn($this->position);
+				$this->position->getWorld()->setBlock($this->position, $ev->getNewState(), false);
 			}
 		}
 	}
--- a/src/block/Ice.php
+++ b/src/block/Ice.php
@@ -23,6 +23,7 @@ declare(strict_types=1);
 
 namespace pocketmine\block;
 
+use pocketmine\event\block\IceMeltEvent;
 use pocketmine\item\enchantment\VanillaEnchantments;
 use pocketmine\item\Item;
 use pocketmine\player\Player;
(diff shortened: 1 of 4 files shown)


thebigsmileXD:
I'd rename it to something more generic like BlockMeltEvent in case Mojang adds any other block with similar mechanic

ColinHDev:
It's also called EntityTrample**Farmland**Event: https://github.com/pmmp/PocketMine-MP/blob/stable/src/event/entity/EntityTrampleFarmlandEvent.php

thebigsmileXD:
> It's also called EntityTrample**Farmland**Event: [`stable`/src/event/entity/EntityTrampleFarmlandEvent.php](https://github.com/pmmp/PocketMine-MP/blob/stable/src/event/entity/EntityTrampleFarmlandEvent.php?rgh-link-date=2021-12-19T18%3A24%3A26Z)

https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Fwww.reactiongifs.us%2Fwp-content%2Fuploads%2F2018%2F01%2Ftenor.gif&f=1&nofb=1

Mistakes were made.

ColinHDev:
I just don't align with your opinion that a more general name should be used here and provide an example for my point.
The melting behaviour is unique to ice blocks, as well as being trampled is to farmland.
Or should we also rename `PlayerToggleSneakEvent` to `EntityToggleSneakEvent` because one day, a new entity might come, which has the ability to sneak?

src/event/block/IceMeltEvent.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace pocketmine\event\block;
+
+use pocketmine\event\Cancellable;
+use pocketmine\event\CancellableTrait;
+
+/**
+ * Called when ice melts due to the light level being too high
+ */
+class IceMeltEvent extends BlockEvent implements Cancellable{
```
ColinHDev, line 32: Maybe this should extend `BaseBlockChangeEvent` so that the block's new state could be provided, as it would be different whether ice or snow (if that would be implemented later) would melt. (https://github.com/pmmp/PocketMine-MP/blob/stable/src/block/SnowLayer.php#L103)

ColinHDev:
> I'd rename it to something more generic like BlockMeltEvent in case Mojang adds any other block with similar mechanic

Never mind, snow can also melt.

Driesboy:
Seems like I missed that, will add

thebigsmileXD:
> I just don't align with your opinion that a more general name should be used here and provide an example for my point. The melting behaviour is unique to ice blocks, as well as being trampled is to farmland. Or should we also rename `PlayerToggleSneakEvent` to `EntityToggleSneakEvent` because one day, a new entity might come, which has the ability to sneak?

As said, mistakes were made. There are several events named Player*Event, even though they apply to many more entities than just players

In this case, did not even think of snow but it instantly proven my point

ColinHDev:
I just did not like your explanation "It should be changed because someday it might address more blocks".
I immediately stepped down from my point, as it was pointed out on discord that snow can also melt.

src/block/Ice.php
```diff
@@ -51,7 +52,11 @@ public function ticksRandomly() : bool{
 
 	public function onRandomTick() : void{
 		if($this->position->getWorld()->getHighestAdjacentBlockLight($this->position->x, $this->position->y, $this->position->z) >= 12){
-			$this->position->getWorld()->useBreakOn($this->position);
+			$ev = new BlockMeltEvent($this, VanillaBlocks::WATER());
+			$ev->call();
+			if(!$ev->isCancelled()){
+				$this->position->getWorld()->setBlock($this->position, $ev->getNewState(), false);
```
dktapps, line 58: This shouldn't disable block updating (it's needed to recalculate light level and notify nearby entities).

src/block/SnowLayer.php
```diff
@@ -100,7 +101,11 @@ public function ticksRandomly() : bool{
 
 	public function onRandomTick() : void{
 		if($this->position->getWorld()->getBlockLightAt($this->position->x, $this->position->y, $this->position->z) >= 12){
-			$this->position->getWorld()->setBlock($this->position, VanillaBlocks::AIR(), false);
+			$ev = new BlockMeltEvent($this, VanillaBlocks::AIR());
+			$ev->call();
+			if(!$ev->isCancelled()){
+				$this->position->getWorld()->setBlock($this->position, $ev->getNewState(), false);
```
dktapps, line 107: Same again here.

dktapps (dismissed):

PR #4670, opened 2021-12-22
Added gold(§g) color support

xxAROX:
## Introduction
Adds support for `§g` for console

### API changes
```diff
+ TextFormat::MINECOIN_GOLD
+ Terminal::$COLOR_MINECOIN_GOLD
```

### DIFF
--- a/src/utils/Terminal.php
+++ b/src/utils/Terminal.php
@@ -132,7 +132,7 @@ abstract class Terminal{
 			self::$COLOR_DARK_AQUA = $colors >= 256 ? $setaf(37) : $setaf(6);
 			self::$COLOR_DARK_RED = $colors >= 256 ? $setaf(124) : $setaf(1);
 			self::$COLOR_PURPLE = $colors >= 256 ? $setaf(127) : $setaf(5);
-			self::$COLOR_ORANGE = $colors >= 256 ? $setaf(214) : $setaf(3);
+			self::$COLOR_GOLD = $colors >= 256 ? $setaf(214) : $setaf(3);
 			self::$COLOR_GRAY = $colors >= 256 ? $setaf(145) : $setaf(7);
 			self::$COLOR_DARK_GRAY = $colors >= 256 ? $setaf(59) : $setaf(8);
 			self::$COLOR_BLUE = $colors >= 256 ? $setaf(63) : $setaf(12);
--- a/src/utils/Terminal.php
+++ b/src/utils/Terminal.php
@@ -48,7 +48,7 @@ abstract class Terminal{
 	public static string $COLOR_DARK_AQUA = "";
 	public static string $COLOR_DARK_RED = "";
 	public static string $COLOR_PURPLE = "";
-	public static string $COLOR_ORANGE = "";
+	public static string $COLOR_GOLD = "";
 	public static string $COLOR_GRAY = "";
 	public static string $COLOR_DARK_GRAY = "";
 	public static string $COLOR_BLUE = "";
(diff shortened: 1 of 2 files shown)


ColinHDev:
I don't think that introducing a bc break is the best approach.
`TextFormat::GOLD` should keep its name, while we search for an alternative name for §g. Maybe `MINECOIN_GOLD` as it is called in the [wiki](https://minecraft.fandom.com/wiki/Formatting_codes#Color_codes) or something like `DARK_YELLOW`.

xxAROX:
okay wait

fuyutsuki:
Duplicate of #3347

xxAROX:
> Duplicate of #3347

other one was aborted

src/utils/Terminal.php
```diff
@@ -210,6 +210,7 @@ public static function toANSI(string $string) : string{
 				TextFormat::LIGHT_PURPLE => Terminal::$COLOR_LIGHT_PURPLE,
 				TextFormat::YELLOW => Terminal::$COLOR_YELLOW,
 				TextFormat::WHITE => Terminal::$COLOR_WHITE,
+				TextFormat::MINECOIN_GOLD => Terminal::$COLOR_YELLOW,
```
dktapps, line 213: This shouldn't reuse the YELLOW property. If plugins want to make it display as a separate colour, they should be able to do so.

xxAROX:
@dktapps is there a reason, why TextFormat::toHTML is a 3-digit code and not 6?

dktapps:
> @dktapps is there a reason, why TextFormat::toHTML is a 3-digit code format and not 6?

None that I'm aware of. That code predates me.

dktapps:
The regex for matching colour codes needs to be updated in two additional places:
https://github.com/pmmp/PocketMine-MP/blob/3677d004bfd8bbd16bef24f2b74469802157017e/src/utils/TextFormat.php#L133
https://github.com/pmmp/PocketMine-MP/blob/3677d004bfd8bbd16bef24f2b74469802157017e/src/utils/TextFormat.php#L147

src/utils/TextFormat.php
```diff
@@ -130,7 +130,7 @@ private static function preg_replace(string $pattern, string $replacement, strin
 	 * @return string[]
 	 */
 	public static function tokenize(string $string) : array{
-		$result = preg_split("/(" . TextFormat::ESCAPE . "[0-9a-fk-or])/u", $string, -1, PREG_SPLIT_NO_EMPTY | PREG_SPLIT_DELIM_CAPTURE);
+		$result = preg_split("/(" . TextFormat::ESCAPE . "[0-9a-fg-or])/u", $string, -1, PREG_SPLIT_NO_EMPTY | PREG_SPLIT_DELIM_CAPTURE);
```
dktapps: Now you broke it. It should be `a-gk-o`.

xxAROX: xd, sorry im tierd af

xxAROX: 2days up

KygekDev: @xxAROX You forgot to fix this line. You only fixed line 147.

xxAROX: > 

thanks for notice, i didn't saw that one

KygekDev (commented):
Good

dktapps:
I think 184 would be a better 256color approximation for this, it more closely matches the actual colour.

226 is indistinguishable from actual yellow.

image

src/utils/Terminal.php
```diff
@@ -142,11 +144,12 @@ protected static function getEscapeCodes() : void{
 			self::$COLOR_LIGHT_PURPLE = $colors >= 256 ? $setaf(207) : $setaf(13);
 			self::$COLOR_YELLOW = $colors >= 256 ? $setaf(227) : $setaf(11);
 			self::$COLOR_WHITE = $colors >= 256 ? $setaf(231) : $setaf(15);
+			self::$COLOR_MINECOIN_GOLD = $colors >= 256 ? $setaf(226) : $setaf(11);
```
dktapps:
```suggestion
			self::$COLOR_MINECOIN_GOLD = $colors >= 256 ? $setaf(184) : $setaf(11);
```

Issue #4668, opened 2021-12-20
Swimming Exhaustion


thebigsmileXD:
### Issue description

With #4446 swimming got introduced. Despite exhaustion being mentioned several times, it was not included in the commit to prevent doing too many things at once - causing me to forget to add a follow-up PR. Since the wiki can not be trusted with values, and my IDA being broken at the moment, i can not figure out the correct values. Nukkit and similar do not seem to have such values either (searched saturation / exhaustion)

dktapps:
Technical details on this:

- while swimming underwater (head below water), the exhaustion is `0.015 * oldPosition->distance(newPosition)`
- while swimming on the surface of water (head above water), the exhaustion is `0.015 * sqrt((oldPosition->x - newPosition->x) ** 2 + (oldPosition->z - newPosition->z) ** 2)` (y axis distance is NOT considered)

PR #4677, opened 2021-12-27
Gameplay/swimming exhaust

inxomnyaa:
## Introduction

Implements correct swimming exhaust as mentioned in #4668

### Relevant issues

#4668

### API changes

Added Entity::isInsideOfLiquid()

### Behavioural changes

Swimming now uses the correct exhaust

## Follow-up

It is to mention that `isUnderwater` has no way to consider if it should check for the boundingbox being fully submerged or partially in water, which might cause problems with things like fishing bobber (which in vanilla resets the RNG when it "bobs" out of water - it only checks for the head being submerged, as it is needed for breathing code.

~~Adding a isSubmergedInLiquid(Liquid, bool fully) might be a good idea for the future.~~ EDIT: isInsideOfLiquid() has been added to address this

## Tests

Since exhaust values are a bit wonky in PMMP anyways, the tests might be a bit flawed. It appears swimming is working fine though with the exhaust now, using up far less hunger than it used to.

~~One thing i am not entirely sure of yet is if the isSwimming and "!isUnderwater" works properly together, still testing~~
EDIT: after excessive tests i can conclude that exhaust is applied correctly now, both with "bobbing" swimming and sprint-swimming

### DIFF
--- a/src/entity/Entity.php
+++ b/src/entity/Entity.php
@@ -1130,15 +1130,15 @@ abstract class Entity{
 	}
 
 	public function isInsideOfSolid() : bool{
-		$block = $this->getWorld()->getBlockAt((int) floor($this->location->x), (int) floor($y = ($this->location->y + $this->getEyeHeight())), (int) floor($this->location->z));
+		$block = $this->getWorld()->getBlockAt((int) floor($this->location->x), (int) floor($this->location->y + $this->getEyeHeight()), (int) floor($this->location->z));
 
 		return $block->isSolid() && !$block->isTransparent() && $block->collidesWithBB($this->getBoundingBox());
 	}
 
 	/**
 	 * Checks if this entity is inside of a specific liquid
-	 * @param bool   $checkEyePos If true, only the block at the eye position is checked for, otherwise all colliding liquids
-	 * @param bool   $submerged If true, all blocks in the bounding box must be the liquid
+	 * @param bool $checkEyePos If true, only the block at the eye position is checked for, otherwise all colliding liquids
+	 * @param bool $submerged   If true, all blocks in the bounding box must be the liquid
 	 */
 	public function isInsideOfLiquid(Liquid $liquid, bool $checkEyePos, bool $submerged = false) : bool{
 		$y = $this->location->y + $this->getEyeHeight();
@@ -1146,7 +1146,7 @@ abstract class Entity{
 		if($checkEyePos){
 			return $this->getWorld()->getBlockAt((int) floor($this->location->x), (int) $blockY, (int) floor($this->location->z)) instanceof $liquid;
 		}
-		$collisionBlocks = $this->getWorld()->getCollisionBlocks($this->boundingBox);
+		$collisionBlocks = $this->getBlocksAroundWithEntityInsideActions();
 		$blocks = array_filter($collisionBlocks, static function($block) use ($liquid, $y, $blockY, $submerged) : bool{
 			if($submerged && $block instanceof $liquid){
 				$f = ($blockY + 1) - ($block->getFluidHeightPercent() - 0.1111111);
--- a/src/entity/Entity.php
+++ b/src/entity/Entity.php
@@ -1157,9 +1157,9 @@ abstract class Entity{
 		$countFiltered = count($blocks);
 		if($countFiltered === 0){
 			return false;
-		}elseif(!$submerged && $countFiltered > 0){
+		}elseif(!$submerged){
 			return true;
-		}elseif($submerged && $countFiltered === count($collisionBlocks)){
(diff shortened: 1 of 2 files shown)


src/player/Player.php
```diff
@@ -1213,9 +1213,14 @@ protected function processMostRecentMovements() : void{
 
 			$horizontalDistanceTravelled = sqrt((($from->x - $to->x) ** 2) + (($from->z - $to->z) ** 2));
 			if($horizontalDistanceTravelled > 0){
-				//TODO: check swimming (adds 0.015 exhaustion in MCPE)
 				if($this->isSprinting()){
 					$this->hungerManager->exhaust(0.1 * $horizontalDistanceTravelled, PlayerExhaustEvent::CAUSE_SPRINTING);
+				}elseif($this->isSwimming()){
```
dktapps: This won't be set if the player is just bobbing up and down on the surface of water.

dktapps:
`isUnderwater()` is not the same as `isInWater()`, which is what we would need for the surface swimming.

src/entity/Entity.php
```diff
@@ -1128,11 +1130,41 @@ public function isUnderwater() : bool{
 	}
 
 	public function isInsideOfSolid() : bool{
-		$block = $this->getWorld()->getBlockAt((int) floor($this->location->x), (int) floor($y = ($this->location->y + $this->getEyeHeight())), (int) floor($this->location->z));
+		$block = $this->getWorld()->getBlockAt((int) floor($this->location->x), (int) floor($this->location->y + $this->getEyeHeight()), (int) floor($this->location->z));
```
ShockedPlot7560, line 1133: unrelated change

src/entity/Entity.php
```diff
@@ -1128,11 +1130,41 @@ public function isUnderwater() : bool{
...
+	public function isInsideOfLiquid(Liquid $liquid, bool $checkEyePos, bool $submerged = false) : bool{
+		$y = $this->location->y + $this->getEyeHeight();
+		$blockY = floor($y);
+		if($checkEyePos){
+			return $this->getWorld()->getBlockAt((int) floor($this->location->x), (int) $blockY, (int) floor($this->location->z)) instanceof $liquid;
+		}
+		$collisionBlocks = $this->getBlocksAroundWithEntityInsideActions();
+		$blocks = array_filter($collisionBlocks, static function($block) use ($liquid, $y, $blockY, $submerged) : bool{
+			if($submerged && $block instanceof $liquid){
+				$f = ($blockY + 1) - ($block->getFluidHeightPercent() - 0.1111111);
```
ShockedPlot7560, line 1152: What's this value 0.1111 ?

src/entity/Entity.php
inxomnyaa: Should be a follow-up, agreed

src/entity/Entity.php
inxomnyaa: https://github.com/pmmp/PocketMine-MP/blob/77db7a8941d3d91c7e452cb3ca469f59fbd8f4df/src/entity/Entity.php#L1123

src/entity/Entity.php
```diff
@@ -1128,11 +1130,41 @@ public function isUnderwater() : bool{
...
+	/**
+	 * Checks if this entity is inside of a specific liquid
+	 * @param bool $checkEyePos If true, only the block at the eye position is checked for, otherwise all colliding liquids
+	 * @param bool $submerged   If true, all blocks in the bounding box must be the liquid
+	 */
+	public function isInsideOfLiquid(Liquid $liquid, bool $checkEyePos, bool $submerged = false) : bool{
+		$y = $this->location->y + $this->getEyeHeight();
+		$blockY = floor($y);
+		if($checkEyePos){
+			return $this->getWorld()->getBlockAt((int) floor($this->location->x), (int) $blockY, (int) floor($this->location->z)) instanceof $liquid;
```
ShockedPlot7560, line 1147: Why not using floorX, floorZ methods from vector ?

src/entity/Entity.php
```diff
@@ -1128,11 +1130,41 @@ public function isUnderwater() : bool{
...
+		});
+		$countFiltered = count($blocks);
+		if($countFiltered === 0){
+			return false;
+		}elseif(!$submerged){
+			return true;
+		}elseif($countFiltered === count($collisionBlocks)){
+			return true;
+		}
+		return false;
```
ShockedPlot7560, line 1165: this conditions can be reduced maybe with only one.

inxomnyaa:
I don't really touch PHP anymore these days (can't be bothered to loose my mind for minor edits), so i'll leave it here for anyone to claim.

PR #4683, opened 2021-12-30
PlayerInteractEvent: added APIs to independently control reaction of item and block

### LINKED ISSUE #3267: unable to cancel item activation without cancelling block activation
## Description
We need some way to fine-tune the actions that result when when right clicking on a block using an item.

## Justification
It's not possible to prevent the use of an item by clicking on a block without also preventing the clicked block reacting to click.

This is problematic in arenas where we need to prevent the use of items like spawn eggs and paintings, but we still want to allow the use of doors and trapdoors.

## Alternative methods
Providing the functionality that I mentioned, disabling painting and spawn egg, without disabling doors and trapdoors, can currently only be done by hardcoding a list of blacklisted items or whitelisted blocks.

dktapps:
## Introduction

this allows, for example, banning the usage of spawn eggs, without preventing opening of doors, without the need for item ID whitelists.

It also allows customizing the behaviour of item and block interactions when sneaking - it's now possible to force spawn eggs to work when sneaking, or force containers to open.

Finally, this also allows preventing any interaction at all without preventing block placement (by setting both to false). Since cancelling the event will typically prevent placement too (which might not be desired).

The naming of the functions is up for discussion, but the feature itself is stolen from Spigot and is much needed.

### Relevant issues
- Fixes #3267 

### API changes
- Added the following new API methods:
  - `PlayerInteractEvent->useBlock()` - returns whether the clicked block's `onInteract()` will be called
  - `PlayerInteractEvent->useItem()` - returns whether the held item's `onInteractBlock()` will be called
  - `PlayerInteractEvent->setUseBlock()`
  - `PlayerInteractEvent->setUseItem()`
Note that setting these both to false does NOT cancel the event; rather, it makes the action behave as if the player was sneaking.

## Backwards compatibility
Backwards compatible, but could cause unexpected results for plugins not aware of these features - e.g. a plugin may assume that chests are never opened if the player was sneaking, which this change would violate.

## Tests
This has not yet been tested.


dktapps:
This causes some buggy behaviour when using `setUseItem(false)` and `setUseBlock(true)` when using a lava bucket on a door: a ghost lava block is placed (unexpected) and the door is opened (expected). The lava doesn't exist on the server side, so it doesn't burn or spread.

ColinHDev:
Would it help if I also test with this so the PR can get merged, or are you currently busy with the work on `next-major`?

ShockedPlot7560:
I like the idea, I'd dig it up from the depths, soon

dktapps:
I think the main issue with this is that `World->useItemOn()` and `Player->onInteractBlock()` can only return `true` (succeeded, don't resync anything) or `false` (failed, resync held item & clicked block). This PR introduces a situation where the interaction succeeds, but resync of held item and/or clicked block is still necessary, which is not currently possible to represent with this API.

I think fixing this will probably require BC breaks, unless deprecated wrappers are used.

dktapps:
Turns out we can just always sync blocks during a block interaction. Since the client doesn't tell us what it expects, this is probably the best course of action anyway. Might even fix some issues in other areas.

ShockedPlot7560 (approved):

PR #4687, opened 2022-01-01
Implement beacon

ShockedPlot7560:
## Introduction
Beacons are currently available but completely useless as the functionality has not been added.

### API changes
* Added BeaconInventory and BeaconAction classes
* Completion of the beacon classe

## Tests
* Build beacon structure as vanilla
* Activate it with the ore of your choice
image


dktapps:
a

ghost:
I'm not sure that dktapps will merged it.

ShockedPlot7560:
I am in the process of making changes to improve performance

dktapps:
> I'm not sure that dktapps will merged it.

I haven't had time to review it yet.

ghost:
> > I'm not sure that dktapps will merged it.
> 
> I haven't had time to review it yet.

I've meant that it's useless for most users of PocketMine-MP. This functionality can be move to plugin.

I thought pmmp was following the way of minimalism. No?

dktapps:
If that were the case, we could strip almost all the code in the name of "minimalism"...

src/block/Beacon.php
```diff
@@ -27,14 +27,87 @@
...
+		BlockLegacyIds::EMERALD_BLOCK => true
+		//TODO netherite block
+	];
+
+	private const ALLOWED_PRIMARY_EFFECTS = [
+		EffectIds::HASTE => true,
+		EffectIds::JUMP_BOOST => true,
+		EffectIds::RESISTANCE => true,
+		EffectIds::SPEED => true,
+		EffectIds::STRENGTH => true
```
dktapps: Effect IDs should not be used in core code.

src/block/Beacon.php
```diff
@@ -27,14 +27,87 @@
...
+	];
+
+	private const ALLOWED_SECONDARY_EFFECTS = [
+		EffectIds::REGENERATION => true
```
dktapps: Same again here.

src/block/Beacon.php
```diff
@@ -27,14 +27,87 @@
...
+	];
+
+	private int $primaryEffect;
```
dktapps: These should be `Effect` instances, not effect IDs.

src/block/Beacon.php
```diff
@@ -27,14 +27,87 @@
...
+		return $this->primaryEffect;
+	}
+
+	/**
+	 * @return $this
+	 * @throws \InvalidArgumentException
+	 */
+	public function setPrimaryEffect(int $primaryEffect) : self{
+		if(!isset(self::ALLOWED_PRIMARY_EFFECTS[$primaryEffect])){
+			throw new \InvalidArgumentException("Effect ID \"$primaryEffect\" is not allowed in the primary effect");
```
dktapps: This is too restrictive. While players shouldn't be able to do this, plugins should be allowed to if they want.

src/block/Beacon.php
```diff
@@ -49,16 +122,16 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
 
-		$primaryE = $tile->getPrimaryEffect();
-		$secondaryE = $tile->getSecondaryEffect();
+		if(!$this->viewSky()){
+			return;
+		}
 
-		if($primaryE === 0 && $secondaryE === 0 || !$this->viewSky()){
+		$effectIdMap = EffectIdMap::getInstance();
+		$primaryE = $effectIdMap->fromId($this->primaryEffect);
```
dktapps: Please name the variables properly.

src/block/Beacon.php
```diff
@@ -49,16 +122,16 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
 		$this->position->getWorld()->scheduleDelayedBlockUpdate($this->position, 20 * 3);
 
-		$primaryE = $tile->getPrimaryEffect();
-		$secondaryE = $tile->getSecondaryEffect();
+		if(!$this->viewSky()){
+			return;
+		}
 
-		if($primaryE === 0 && $secondaryE === 0 || !$this->viewSky()){
+		$effectIdMap = EffectIdMap::getInstance();
```
dktapps: EffectIdMap should never be needed except for storing data, which is done by the tile.

src/block/Beacon.php
```diff
@@ -23,11 +23,176 @@
...
+		return $this;
+	}
+
+	public function getSecondaryEffect() : int{
+		return $this->secondaryEffect;
+	}
+
+	/** @return $this */
+	public function setSecondaryEffect(int $secondaryEffect) : self{
+		if(!isset(self::ALLOWED_PRIMARY_EFFECTS[$secondaryEffect]) || !isset(self::ALLOWED_SECONDARY_EFFECTS[$secondaryEffect])){
```
dktapps: Again, too restrictive.

src/block/Beacon.php
```diff
@@ -23,11 +23,176 @@
...
+		$effectIdMap = EffectIdMap::getInstance();
+		$primaryE = $effectIdMap->fromId($this->primaryEffect);
+		$secondaryE = $effectIdMap->fromId($this->secondaryEffect);
+		if($primaryE === null && $secondaryE === null){
+			return;
+		}
+
+		$beaconLevel = 0;
+		for($i = 1; $i <= 4; $i++){
+			if(!$this->isBeaconLevelValid($i)){
```
dktapps: This is going to be abysmal for performance.

src/block/Beacon.php
```diff
@@ -23,11 +23,176 @@
...
+				break;
+			}
+			$beaconLevel++;
+		}
+		if($beaconLevel > 0){
+			$radius = (10 * $beaconLevel) + 10;
+			$effectDuration = 9 + (2 * $beaconLevel);
+
+			$world = $this->position->getWorld();
+			$aabb = (new AxisAlignedBB(0, 0, 0, 1, $world->getMaxY(), 1))->offset($this->position->x, 0, $this->position->z)->expand($radius, 0, $radius);
```
dktapps: This seems suspect. I'm pretty sure only the players _above_ the beacon should receive effects.

src/block/Beacon.php
```diff
@@ -23,11 +23,176 @@
...
+			if($primaryE === $secondaryE){
```
dktapps: This looks like premature optimisation.

src/block/Beacon.php
```diff
@@ -23,11 +23,176 @@
...
+			$player->setCurrentWindow(new BeaconInventory($this->position));
+		}
+
+		return true;
+	}
+
+	public function onScheduledUpdate() : void{
+		$this->position->getWorld()->scheduleDelayedBlockUpdate($this->position, 20 * 3);
+
+		if(!$this->viewSky()){
```
dktapps: This is going to be abysmal for performance.

src/block/inventory/Beaconinventory.php
```diff
@@ -0,0 +1,44 @@
+<?php
```
dktapps: Incorrect file name casing.

src/network/mcpe/convert/TypeConverter.php
```diff
@@ -272,6 +273,10 @@ public function createInventoryAction(NetworkInventoryAction $action, Player $pl
 					}
 					$pSlot = $action->inventorySlot;
 
+					if($pSlot === UIInventorySlotOffset::BEACON_PAYMENT){
+						return new BeaconAction($old, $new);
```
dktapps: This is incorrectly implemented. The item is not directly consumed when it's placed into the beacon window, only when the effects have been selected. Therefore, the inventory should work the same as any other temporary inventory such as crafting tables.

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -657,6 +658,17 @@ public function handleBlockActorData(BlockActorDataPacket $packet) : bool{
 			}
 
 			$this->session->getLogger()->debug("Invalid sign update data: " . base64_encode($packet->nbt->getEncodedNbt()));
+		}elseif($block instanceof Beacon){
+			try{
+				$block->setPrimaryEffect($nbt->getInt("primary", 0));
+				$block->setSecondaryEffect($nbt->getInt("secondary", 0));
```
dktapps: The following things should be checked before activating the beacon:

- Whether or not the height of the pyramid allows the selected effects
- Whether the beacon inventory window contains a valid type of payment (ingots, diamonds, emeralds)

There should also be an event to allow controlling this from a plugin.

dktapps (changes requested):

src/block/Beacon.php
JavierLeon9966: According to the wiki:
> The beacon affects an area in the shape of a square column, which reaches downward and out to each side at a range determined by the size of the pyramid base (see table below), and upward a distance of that range + 384 blocks.

https://minecraft.fandom.com/wiki/Beacon#Range

src/block/Beacon.php
JavierLeon9966: If the primary and secondary effects are the same then we increase one level, that’s how we’re checking it.

dktapps: I see, I didn't see that.

src/block/Beacon.php
```diff
@@ -23,11 +23,176 @@
...
+			$beaconLevel++;
+		}
+		if($beaconLevel > 0){
+			$radius = (10 * $beaconLevel) + 10;
+			$effectDuration = 9 + (2 * $beaconLevel);
+
+			$world = $this->position->getWorld();
+			$aabb = (new AxisAlignedBB(0, 0, 0, 1, $world->getMaxY(), 1))->offset($this->position->x, 0, $this->position->z)->expand($radius, 0, $radius);
+			if($primaryE === $secondaryE){
+				if($secondaryE === null){
```
dktapps: This isn't necessary. You already checked if it was null above.

JavierLeon9966: For some reason PHPStan doesn’t like it

src/block/Beacon.php
dktapps: It still shouldn't be reaching all the way down to y=0.

src/block/Beacon.php
```diff
@@ -23,11 +23,176 @@
...
+
+		$beaconLevel = 0;
+		for($i = 1; $i <= 4; $i++){
+			if(!$this->isBeaconLevelValid($i)){
+				break;
+			}
+			$beaconLevel++;
+		}
+		if($beaconLevel > 0){
+			$radius = (10 * $beaconLevel) + 10;
```
dktapps, line 141: This could just be 11 x radius.

src/block/Beacon.php
JavierLeon9966: I tested in vanilla and it reaches until the void, so this should have used `World->getMinY()`

src/block/Beacon.php
dktapps: That's because this condition is borked: 
image

This should only check the primary effect. A beacon may have a primary effect without a secondary effect.

src/block/Beacon.php
ShockedPlot7560: If we put just 11x radius, for level 1, it will be 11 of range but 20 needed

JavierLeon9966: Defining the variable `$beaconLevel` as: `$beaconLevel = 1` could simplify this expression to: `$radius = 10 * $beaconLevel;`. Also don’t forget to also check `if($beaconLevel > 1)`

dktapps: My bad. My brain isn't working very well right now.

src/block/Beacon.php
dktapps: This sounds sus. According to the wiki quote you just pulled, it's supposed to reach down `$radius` blocks below the beacon block.

ShockedPlot7560: > The beacon affects an area in the shape of a square column

so, it means from layer 0 to the highest one normally following the vertical axis

dktapps: `which reaches downward and out to each side at a range determined by the size of the pyramid base (see table below), and upward a distance of that range + 384 blocks.`

JavierLeon9966: Well, it also says: "…and upward a distance of that range + 384 blocks.", which means from -64 up to 320

JavierLeon9966: That message was delayed btw. And also my bad, that seems to be the way is working in vanilla now that I tested again

dktapps: No, it doesn't. It means the beacon Y coordinate + 384.

dktapps: Corresponding Java code: 
image

JavierLeon9966: Further testing in vanilla suggests that there’s no max upward distance

dktapps: Well there needs to be a limit because we can't check AABB intersections into infinity. Stopping at the world maxY is fine.

src/block/Beacon.php
ShockedPlot7560: however, we must check the beacon level at each schedule update as the view with the sky

src/block/Beacon.php
ShockedPlot7560: we have to check the sky view every time

src/block/Beacon.php
```diff
@@ -44,61 +51,52 @@ final class Beacon extends Transparent{
...
 		parent::readStateFromWorld();
 		$tile = $this->position->getWorld()->getTile($this->position);
 		if($tile instanceof TileBeacon){
-			$this->primaryEffect = $tile->getPrimaryEffect();
-			$this->secondaryEffect = $tile->getSecondaryEffect();
+			$effectIdMap = EffectIdMap::getInstance();
+			$primaryEffectId = $tile->getPrimaryEffect();
+			$primaryEffect = $effectIdMap->fromId($primaryEffectId);
+			if($primaryEffect === null){
+				throw new InvalidBlockStateException("Invalid primary effect $primaryEffectId");
```
dktapps: The beacon may not yet have been activated.

ShockedPlot7560: Indeed, stupid me

src/block/Beacon.php
```diff
@@ -23,11 +23,209 @@
...
+		if($beaconLevel > 0){
+			$radius = (10 * $beaconLevel) + 10;
+			$effectDuration = 9 + (2 * $beaconLevel);
+
+			$world = $this->position->getWorld();
+			$aabb = $this->getCollisionBoxes()[0]->expandedCopy($radius, $radius, $radius)->addCoord(0, $world->getMaxY(), 0);
+			if($this->primaryEffect === $this->secondaryEffect){
+				foreach($world->getNearbyEntities($aabb) as $entity){
+					if($entity instanceof Player){
+						if($this->primaryEffect !== null){
```
JavierLeon9966: You should check this before iterating

src/block/Beacon.php
```diff
@@ -23,11 +23,209 @@
...
+							$entity->getEffects()->add(new EffectInstance($this->primaryEffect, $effectDuration * 20, 1));
+						}
+					}
+				}
+			}else{
+				$effects = [];
```
JavierLeon9966: Check if the effects are not empty before iterating

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -657,6 +661,37 @@ public function handleBlockActorData(BlockActorDataPacket $packet) : bool{
...
+					$event = new BeaconActivateEvent($block, $primaryEffect);
+					$secondaryEffect = EffectIdMap::getInstance()->fromId($nbt->getInt("secondary", 0));
+					if($secondaryEffect !== null && in_array($secondaryEffect, $allowedEffects, true)){
+						$event->setSecondaryEffect($secondaryEffect);
+					}
+					$event->call();
+					if(!$event->isCancelled()){
+						$event->setPrimaryEffect($event->getPrimaryEffect());
+						$event->setSecondaryEffect($event->getSecondaryEffect());
+						$inventory->getInput()->pop();
```
JavierLeon9966: You are not setting the new item back into the inventory.

JavierLeon9966 (changes requested):

src/block/Beacon.php
ShockedPlot7560: We already check it, if all effects are null, the iteration will be done on an empty array

JavierLeon9966: But you would be unnecessary iterating over something with no value

ShockedPlot7560:
When the item is consumed, when the beacon is activated, the item is simply not removed from the inventory.
When retrieving the packet from the TypeConverter, it doesn't want to either. Am I missing something or have I done something wrong?

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -657,6 +661,37 @@ public function handleBlockActorData(BlockActorDataPacket $packet) : bool{
...
+					$event = new BeaconActivateEvent($block, $primaryEffect);
+					$secondaryEffect = EffectIdMap::getInstance()->fromId($nbt->getInt("secondary", 0));
+					if($secondaryEffect !== null && in_array($secondaryEffect, $allowedEffects, true)){
+						$event->setSecondaryEffect($secondaryEffect);
+					}
+					$event->call();
+					if(!$event->isCancelled()){
+						$event->setPrimaryEffect($event->getPrimaryEffect());
+						$event->setSecondaryEffect($event->getSecondaryEffect());
+						$inventory->setInput($inventory->getInput()->pop());
```
dktapps: This is just going to set the item to itself. pop() returns the item removed from the stack.

ShockedPlot7560: how am i supposed to remove the item from the beacon inventory? the item pop produces an error when we activate the beacon because it does not know the window with the id -10.

JavierLeon9966:
```suggestion
						$item = $inventory->getInput();
						$item->pop();
						$inventory->setInput($item);
```

ShockedPlot7560:
A persistent problem that needs to be discussed is the verification of the beacon level.
Currently, if you build a level 4 beacon and do not activate it with a side effect, it will remain permanent even if you break the base of the beacon to reduce its level.
The vanilla gameplay prohibits this, and then removes the secondary effect. However, if we put this check, it implies that plugins won't be able to put second effects if the beacon level is below 4. 

The question almost does not arise but, should we check this or let the fact that if the beacon reduces its level, the active effects remain the same

src/network/mcpe/handler/InGamePacketHandler.php
dktapps: > how am i supposed to remove the item from the beacon inventory? the item pop produces an error when we activate the beacon because it does not know the window with the id -10.

Too vague for me to have any idea what the problem is.

ShockedPlot7560: Finally, the problem was solved, I was not using the pop() correctly

dktapps:
If the beacon level reduces the secondary effect should be removed. This is an unlikely case so it doesn't bear much thinking about.

ShockedPlot7560:
> If the beacon level reduces the secondary effect should be removed. This is an unlikely case so it doesn't bear much thinking about.

good

src/block/Beacon.php
```diff
@@ -23,11 +23,210 @@
...
+	private const ALLOWED_BLOCK_IDS = [
+		BlockLegacyIds::IRON_BLOCK => true,
+		BlockLegacyIds::GOLD_BLOCK => true,
+		BlockLegacyIds::DIAMOND_BLOCK => true,
+		BlockLegacyIds::EMERALD_BLOCK => true
+		//TODO netherite block
+	];
+
+	/** @var null|Effect */
+	private $primaryEffect;
```
Crasher508:
```suggestion
	private ?Effect $primaryEffect = null;
```

src/block/Beacon.php
```diff
@@ -23,11 +23,210 @@
...
+
+	private const ALLOWED_BLOCK_IDS = [
+		BlockLegacyIds::IRON_BLOCK => true,
+		BlockLegacyIds::GOLD_BLOCK => true,
+		BlockLegacyIds::DIAMOND_BLOCK => true,
+		BlockLegacyIds::EMERALD_BLOCK => true
+		//TODO netherite block
+	];
+
+	/** @var null|Effect */
```
Crasher508: I think PHPDoc is very old, better with initialization
```suggestion
```

src/block/Beacon.php
```diff
@@ -23,11 +23,210 @@
...
+	private $primaryEffect;
+	/** @var null|Effect */
+	private $secondaryEffect;
```
Crasher508:
```suggestion
	private ?Effect $secondaryEffect = null;
```

src/block/Beacon.php
```diff
@@ -23,11 +23,210 @@
...
+		BlockLegacyIds::IRON_BLOCK => true,
+		BlockLegacyIds::GOLD_BLOCK => true,
+		BlockLegacyIds::DIAMOND_BLOCK => true,
+		BlockLegacyIds::EMERALD_BLOCK => true
+		//TODO netherite block
+	];
+
+	/** @var null|Effect */
+	private $primaryEffect;
+	/** @var null|Effect */
```
Crasher508:
```suggestion
```

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -657,6 +661,39 @@ public function handleBlockActorData(BlockActorDataPacket $packet) : bool{
...
+			$itemId = $inventory->getInput()->getId();
+			if(!isset(Beacon::ALLOWED_ITEM_IDS[$itemId])){
+				throw new PacketHandlingException("Invalid input $itemId");
+			}
+
+			if(($beaconLevel = $block->getBeaconLevel()) > 0 && $block->viewSky()){
+				$allowedEffects = $block->getAllowedEffect($block->getBeaconLevel());
+				$primaryEffect = EffectIdMap::getInstance()->fromId($nbt->getInt("primary", 0));
+				if($primaryEffect !== null && in_array($primaryEffect, $allowedEffects, true)){
+					$event = new BeaconActivateEvent($block, $primaryEffect);
```
Crasher508: Does we really need a event?

Crasher508 (changes requested):
Works nice, tested
idk. who need a event/PHPDoc seems a little bit old

src/network/mcpe/handler/InGamePacketHandler.php
SOF3: I think the main reason why not everything has an event right now is that an event file has too much boilerplate (create a new file, add license headers, generate getters/setters). Looks fine to me if it is already created.

ShockedPlot7560: I think that an event, in this case, is interesting, especially for the plugins that would like to modify the actions done.

jasonw4331 (approved):

src/block/Beacon.php
```diff
@@ -23,11 +23,209 @@
...
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if($player instanceof Player){
+			$player->setCurrentWindow(new BeaconInventory($this->position));
+		}
+
+		return true;
+	}
+
+	public function onScheduledUpdate() : void{
+		$this->position->getWorld()->scheduleDelayedBlockUpdate($this->position, 20 * 3);
```
IvanCraft623, line 131: According to the wiki, updates are every 4 seconds

src/block/Beacon.php
```diff
@@ -23,11 +23,209 @@
...
+use pocketmine\item\Item;
+use pocketmine\item\ItemTypeIds;
+use pocketmine\math\Vector3;
+use pocketmine\player\Player;
+use function array_merge;
+
 final class Beacon extends Transparent{
 
+	public const MIN_LEVEL_BEACON = 1;
+	public const MAX_LEVEL_BEACON = 4;
```
IvanCraft623, line 42: Inclunding `BEACON` in the constant name seems redundant since it is on the `Beacon` class

src/block/Beacon.php
```diff
@@ -23,11 +23,209 @@
...
+
+	public const ALLOWED_ITEM_IDS = [
+		ItemTypeIds::IRON_INGOT => true,
+		ItemTypeIds::GOLD_INGOT => true,
+		ItemTypeIds::DIAMOND => true,
+		ItemTypeIds::EMERALD => true,
+		ItemTypeIds::NETHERITE_INGOT => true
+	];
+
+	private const ALLOWED_BLOCK_IDS = [
```
IvanCraft623, line 52: It would be nice to allow plugins to be able to modify this

src/block/Beacon.php
```diff
@@ -23,11 +23,209 @@
...
+		$this->position->getWorld()->scheduleDelayedBlockUpdate($this->position, 20 * 3);
+
+		if($this->primaryEffect === null){
+			return;
+		}
+		if(($beaconLevel = $this->getBeaconLevel()) >= self::MIN_LEVEL_BEACON){
+			if(!$this->viewSky()){
+				return;
+			}
+
```
IvanCraft623, line 140: We need some validation here, it is possible that players might destroy blocks on any level and some effects will be no longer avaible

src/block/Beacon.php
```diff
@@ -23,11 +23,209 @@
...
+						foreach($effects as $effect){
+							$entity->getEffects()->add(new EffectInstance($effect, $effectDuration * 20, 0));
+						}
+					}
+				}
+			}
+		}
+	}
+
+	public function isBeaconLevelValid(int $level) : bool{
```
IvanCraft623, line 168: Please document this function, i feel that the name isn't clear enough

src/block/Beacon.php
```diff
@@ -23,11 +23,209 @@
...
+				$block = $world->getBlock($pos->add($x, 0, $z));
+				if(!isset(self::ALLOWED_BLOCK_IDS[$block->getTypeId()])){
+					return false;
+				}
+			}
+		}
+		return true;
+	}
+
+	public function viewSky() : bool{
```
IvanCraft623, line 186: The function name is not the most appropriate.

src/block/Beacon.php
```diff
@@ -23,11 +23,209 @@
...
+				VanillaEffects::STRENGTH()
+			],
+			self::MIN_LEVEL_BEACON + 3 => [
+				VanillaEffects::REGENERATION()
+			]
+		];
+	}
+
+	/** @return Effect[] */
+	public function getAllowedEffect(int $beaconLevel) : array {
```
IvanCraft623, line 220:
```suggestion
	public function getAllowedEffects(int $beaconLevel) : array {
```

src/inventory/transaction/action/BeaconPaymentAction.php
```diff
@@ -0,0 +1,122 @@
...
+		$block = $position->getWorld()->getBlock($position);
+		if (!$block instanceof Beacon) {
+			throw new TransactionValidationException("Target block is not a beacon");
+		}
+
+		$allowedEffects = $block->getAllowedEffect($block->getBeaconLevel());
+		if(!in_array($this->primaryEffect, $allowedEffects, true)){
+			throw new TransactionValidationException("Primary effect provided is not allowed");
+		}
+		if($this->secondaryEffect !== null && !in_array($this->secondaryEffect, $allowedEffects, true)){
```
IvanCraft623, line 72: This validation is incomplete. Secondary effect can only be Regeneration or the same as primary effect. Current code allows that a hacked-cleint to send a primary effect as secondary.

src/inventory/transaction/action/BeaconPaymentAction.php
```diff
@@ -0,0 +1,122 @@
...
+		}
+
+		$position = $this->inventory->getHolder();
+		$block = $position->getWorld()->getBlock($position);
+		if (!$block instanceof Beacon) {
+			throw new TransactionValidationException("Target block is not a beacon");
+		}
+
+		$allowedEffects = $block->getAllowedEffect($block->getBeaconLevel());
+		if(!in_array($this->primaryEffect, $allowedEffects, true)){
```
IvanCraft623, line 69: Avoid using in_array when possible

src/block/Beacon.php
```diff
@@ -23,11 +23,209 @@
...
+			throw new InvalidArgumentException("Beacon level must be in range " . self::MIN_LEVEL_BEACON . "-" . self::MAX_LEVEL_BEACON . ", $level given");
+		}
+
+		$world = $this->position->getWorld();
+		$pos = $this->position->subtract(0, $level, 0);
+		for($x = -$level; $x <= $level; $x++){
+			for($z = -$level; $z <= $level; $z++){
+				$block = $world->getBlock($pos->add($x, 0, $z));
+				if(!isset(self::ALLOWED_BLOCK_IDS[$block->getTypeId()])){
+					return false;
```
dktapps, line 179: I really don't like this. I know that's the most obvious way to do it, but it's going to lag like shit.

src/block/Beacon.php
dktapps: Type tags would be the best thing

dktapps (changes requested):
I stalled to review this because the performance will suck, but there currently isn't a great alternative way to detect changes in the pyramid.

dktapps:
As discussed internally, it would be best to move forward with a ChunkListener-based system for triggering pyramid rechecks as per @Muqsit's [beacons plugin](https://github.com/Muqsit/Beacon).

ShockedPlot7560:
I've no more time to maintain this one as i'm focusing on more internal system, if someone want to take it back, free to you.

PR #4689, opened 2022-01-02
TimeCommand: changed behavior to affect a single world

Nerahikada:
## Introduction
Currently, the time command affects all loaded worlds.
This PR restricts the affect of the time command to a single world.

### Behavioural changes
- The time command no longer affects the time of all worlds.
  - The player executes the command, the time of the world the player is in will be changed.
  - The console executed the command, the default world time will be changed.

### DIFF
--- a/src/command/defaults/TimeCommand.php
+++ b/src/command/defaults/TimeCommand.php
@@ -29,6 +29,7 @@ use pocketmine\command\utils\InvalidCommandSyntaxException;
 use pocketmine\lang\KnownTranslationFactory;
 use pocketmine\permission\DefaultPermissionNames;
 use pocketmine\player\Player;
+use pocketmine\utils\AssumptionFailedError;
 use pocketmine\world\World;
 use function count;
 use function implode;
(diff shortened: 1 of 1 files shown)


src/command/defaults/TimeCommand.php
```diff
@@ -98,48 +95,30 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
...
-			}
+			$value = match($args[1]){
+				"day" => World::TIME_DAY,
+				"noon" => World::TIME_NOON,
+				"sunset" => World::TIME_SUNSET,
+				"night" => World::TIME_NIGHT,
+				"midnight" => World::TIME_MIDNIGHT,
+				"sunrise" => World::TIME_SUNRISE,
+				default => $this->getInteger($sender, $args[1], 0),
+			};
```
dktapps: This change is unrelated.

dktapps (commented):
I'm doubtful that this is desirable.

xDidntPot:
Tbh this is useful (atleast for me).

src/command/defaults/TimeCommand.php
Nerahikada: I didn't think it was necessary to keep the old code.

dktapps: It's also not necessary to change it, and doesn't contribute to the goal of the PR. Don't change unrelated things.

Nerahikada: I don't understand why you would keep the old code if there is no purpose for it, such as backward compatibility. LOL

dktapps: As I said, it's not adding to the goal of the PR. It's just polluting your diff.

dktapps:
Closing due to lack of activity and lack of compelling reasons to make the change.

PR #4692, opened 2022-01-03
Performance improvements for Server->getPlayerExact()

KingOfTurkey38:
## Introduction
Server->getPlayerExact() iterates over all the players to find the player that matches the given username.
The CPU usage caused by this function can be significantly reduced by implementing a Map<username, Player>
You can find detailed performance tests here https://github.com/KingOfTurkey38/PeformanceTests

### API changes
No API changes

### Behavioural changes
No behavioural changes

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -259,6 +259,8 @@ class Server{
 
 	/** @var Player[] */
 	private array $playerList = [];
+	/** @var Player[] */
+	private array $playerListByUsername = [];
 
 	private SignalHandler $signalHandler;
 
@@ -636,14 +638,7 @@ class Server{
 	 * Returns an online player with the given name (case insensitive), or null if not found.
 	 */
 	public function getPlayerExact(string $name) : ?Player{
-		$name = strtolower($name);
-		foreach($this->getOnlinePlayers() as $player){
-			if(strtolower($player->getName()) === $name){
-				return $player;
-			}
-		}
-
-		return null;
+		return $this->playerListByUsername[$name] ?? null;
 	}
 
 	/**
@@ -1685,6 +1680,7 @@ class Server{
 		}
 		$rawUUID = $player->getUniqueId()->getBytes();
 		$this->playerList[$rawUUID] = $player;
+		$this->playerListByUsername[strtolower($player->getName())] = $player;
 
 		if($this->sendUsageTicker > 0){
 			$this->uniquePlayers[$rawUUID] = $rawUUID;
@@ -1696,6 +1692,7 @@ class Server{
 	public function removeOnlinePlayer(Player $player) : void{
 		if(isset($this->playerList[$rawUUID = $player->getUniqueId()->getBytes()])){
 			unset($this->playerList[$rawUUID]);
+			unset($this->playerListByUsername[strtolower($player->getName())]);
 			foreach($this->playerList as $p){
 				$p->getNetworkSession()->onPlayerRemoved($player);
 			}
--- a/src/Server.php
+++ b/src/Server.php
@@ -638,7 +638,7 @@ class Server{
 	 * Returns an online player with the given name (case insensitive), or null if not found.
 	 */
 	public function getPlayerExact(string $name) : ?Player{
-		return $this->playerListByUsername[$name] ?? null;
+		return $this->playerListByUsername[strtolower($name)] ?? null;
 	}
 
 	/**


SOF3:
Plugins should not have used getPlayerExact on a loop in the first place. If you just want to identify a player with a scalar, use the UUID.

KingOfTurkey38:
> Plugins should not have used getPlayerExact on a loop in the first place. If you just want to identify a player with a scalar, use the UUID.

I don't think that plugin developers even thought that getPlayerExact uses a loop. There's no reason to make it a loop instead of a Map<username, Player>.

src/Server.php
```diff
@@ -1696,6 +1692,7 @@ public function addOnlinePlayer(Player $player) : bool{
 	public function removeOnlinePlayer(Player $player) : void{
 		if(isset($this->playerList[$rawUUID = $player->getUniqueId()->getBytes()])){
 			unset($this->playerList[$rawUUID]);
+			unset($this->playerListByUsername[strtolower($player->getName())]);
```
dktapps, line 1695: This code is not future proof. This could result in a memory leak if the playerList was altered by a plugin (e.g. to hide the player). Future API changes may enable that.

SOF3:
> > Plugins should not have used getPlayerExact on a loop in the first place. If you just want to identify a player with a scalar, use the UUID.
> 
> I don't think that plugin developers even thought that getPlayerExact uses a loop. There's no reason to make it a loop instead of a Map<username, Player>.

Yes, I agree that using O(n) lookup is not a good idea. Just wanted to point out that it's still incorrect for plugins to use the username as an identifier while it isn't.

SOF3:
On a related note, methods that are computationally expensive (maybe more than average O(log n)) should not use the `get` verb anyway. Any thoughts on renaming `getPlayerByPrefix` to `findPlayerByPrefix`?

dktapps:
Poggit suggests there is relatively little usage of this function anyway: https://poggit.pmmp.io/grepPlugins/2022-01-03-2f0f143b15f412ec.html

dktapps:
Moreover, I don't see any place where this is used in a hot path (typically commands, forms, or AsyncTask::onCompletion()), so I doubt this change provides any meaningful benefit.

KingOfTurkey38:
> Moreover, I don't see any place where this is used in a hot path (typically commands, forms, or AsyncTask::onCompletion()), so I doubt this change provides any meaningful benefit.

It currently might not be used a lot by the plugins on poggit but that doesn't mean that it won't be used in the future a lot by poggit plugins. I agree with Sofe. Methods that are computationally expensive (maybe more than average O(log n)) should not use the "get" verb but something like "find" instead. So plugin developers know that it'll be computationally more expensive than just a Map<key, value>.

dktapps:
As I said earlier, we're dealing with a community that will write `if(in_array(thing, array)){ thingIdx = array_search(thing, array); }` and then blame PocketMine for their performance issues. I don't think "proper" naming is going to change that.

That a thing may be used a certain way in the future is not an argument to change it.

SOF3:
> > Moreover, I don't see any place where this is used in a hot path (typically commands, forms, or AsyncTask::onCompletion()), so I doubt this change provides any meaningful benefit.
> 
> It currently might not be used a lot by the plugins on poggit but that doesn't mean that it won't be used in the future a lot by poggit plugins. I agree with Sofe. Methods that are computationally expensive (maybe more than average O(log n)) should not use the "get" verb but something like "find" instead. So plugin developers know that it'll be computationally more expensive than just a Map<key, value>.

Fun fact: Map<key, value> index get is worst case O(n) (or worse?) if you're unlucky and bumped into hash collision.

SOF3:
> As I said earlier, we're dealing with a community that will write `if(in_array(thing, array)){ thingIdx = array_search(thing, array); }` and then blame PocketMine for their performance issues. I don't think "proper" naming is going to change that.
> 
> That a thing may be used a certain way in the future is not an argument to change it.

Having a community that generally doesn't understand about computational performance doesn't change the fact that we should name our methods properly. It is about proper code styles. At least that's more meaningful than adding `: void` to `PluginBase::onEnable`.

As for this PR, my main concern is that player names should never have been used as identifiers. Is there any chance in the future that players are allowed to rename themselves during a session? This was not reasonable in the past because names were used to identify accounts anyway, but we are moving away from names to UUID gradually (although that's still a long journey). But I don't think adding more value to username-as-identifier is a good idea.


> > Moreover, I don't see any place where this is used in a hot path (typically commands, forms, or AsyncTask::onCompletion()), so I doubt this change provides any meaningful benefit.
> 
> It currently might not be used a lot by the plugins on poggit but that doesn't mean that it won't be used in the future a lot by poggit plugins.

I don't see how it would become the case in the future if it isn't the case in the past. Plugins doing that should switch to get-by-uuid instead. That's why I suggested renaming the verb, because that at least reminds developers who care about time complexity (even if they are a smaller community) that there are better ways to do the same thing.

This method is only problematic when it is used in a loop, causing O(n²) performance in your benchmarks. But for the typical case, usernames are searched using `getPlayerByPrefix` anyway, and that is O(number of players) anyway. (Of course we could optimize that into a trie and make it O(length of username) instead, but the implementation of a trie in PHP is most likely going to cause even more constant-time overhead)

dktapps:
Closing due to lack of sufficient justification, and inactivity on review comments. As previously mentioned, this is a cold path anyway.

PR #4696, opened 2022-01-04
Implemented campfire

ipad54:
## Introduction
This PR implements campfire

### API changes
N/A

### Behavioural changes
Added Campfire, registered 'Campfire' furnace type

## Follow-up
N/A

## Tests
https://youtu.be/7aHexDz3LMU


ipad54:
Oops...

src/block/Campfire.php
```diff
@@ -0,0 +1,193 @@
...
+		if($this->getSide(Facing::DOWN)->isTransparent()){
+			return false;
+		}
+		if($player !== null){
+			$faces = [
+				4 => 3,
+				3 => 2,
+				2 => 0,
+				5 => 1
+			];
```
dktapps: That's not how this is done. The facing field should contain `Facing::` values only, not magic metadata garbage.

src/block/Campfire.php
```diff
@@ -0,0 +1,193 @@
...
+use pocketmine\world\sound\ItemFrameAddItemSound;
+use function count;
+
+class Campfire extends Transparent{
+	use HorizontalFacingTrait;
+
+	protected bool $extinguished = false;
+
+	public function writeStateToMeta() : int{
+		return $this->facing | ($this->extinguished ? BlockLegacyMetadata::CAMPFIRE_FLAG_EXTINGUISHED : 0);
```
dktapps: Same again here.

src/block/Campfire.php
```diff
@@ -0,0 +1,193 @@
...
+				return true;
+			}
+			if($item instanceof Shovel && !$this->extinguished){
+				$item->applyDamage(1);
+				$this->extinguish();
+				return true;
+			}
+
+			$tile = $this->position->getWorld()->getTile($this->position);
+			if($tile instanceof TileCampfire && $tile->addItem(clone $item)){
```
dktapps: You should not interact with the tile except in `readStateFromWorld` and `writeStateFromWorld`. See, for example, `ItemFrame`.

src/block/Campfire.php
```diff
@@ -0,0 +1,193 @@
...
+	use HorizontalFacingTrait;
+
+	protected bool $extinguished = false;
+
+	public function writeStateToMeta() : int{
+		return $this->facing | ($this->extinguished ? BlockLegacyMetadata::CAMPFIRE_FLAG_EXTINGUISHED : 0);
+	}
+
+	public function readStateFromData(int $id, int $stateMeta) : void{
+		$this->facing = $stateMeta & 0x03;
```
dktapps: This is wrong. Use `BlockDataSerializer` to convert the metadata (see other blocks for examples).

src/block/Campfire.php
dktapps: There should also be APIs on the Block class to change the items in the campfire without touching the tile.

src/block/Campfire.php
```diff
@@ -0,0 +1,193 @@
...
+				$this->extinguish();
+				return true;
+			}
+
+			$tile = $this->position->getWorld()->getTile($this->position);
+			if($tile instanceof TileCampfire && $tile->addItem(clone $item)){
+				$item->pop();
+				$this->position->getWorld()->addSound($this->position->add(0.5, 0.5, 0.5), new ItemFrameAddItemSound());
+				$this->position->getWorld()->setBlock($this->position, $this);
+				if(count($tile->getContents()) == 1) $this->position->getWorld()->scheduleDelayedBlockUpdate($this->position, 1);
```
dktapps:
```suggestion
				if(count($tile->getContents()) === 1){
					$this->position->getWorld()->scheduleDelayedBlockUpdate($this->position, 1);
				}
```

src/block/VanillaBlocks.php
```diff
@@ -665,6 +666,7 @@ protected static function setup() : void{
 		self::register("carved_pumpkin", $factory->get(410, 0));
 		self::register("chemical_heat", $factory->get(192, 0));
 		self::register("chest", $factory->get(54, 2));
+		self::register("campfire", $factory->get(464, 0));
```
dktapps: This is in the wrong place. The list must be ordered alphabetically.

src/block/tile/Campfire.php
```diff
@@ -0,0 +1,147 @@
...
+use pocketmine\crafting\FurnaceRecipe;
+use pocketmine\crafting\FurnaceType;
+use pocketmine\item\Item;
+use pocketmine\nbt\NBT;
+use pocketmine\nbt\tag\CompoundTag;
+use pocketmine\nbt\tag\IntTag;
+use pocketmine\nbt\tag\ListTag;
+use function count;
+
+class Campfire extends Spawnable{
```
dktapps: Why is this not using `ContainerTrait`?

src/world/sound/ItemFrameAddItemSound.php
```diff
@@ -0,0 +1,35 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\world\sound;
+
+use pocketmine\math\Vector3;
+use pocketmine\network\mcpe\protocol\LevelSoundEventPacket;
+use pocketmine\network\mcpe\protocol\types\LevelEvent;
+
+class ItemFrameAddItemSound implements Sound{
```
dktapps: This is clearly a bad name

src/block/tile/Campfire.php
```diff
@@ -0,0 +1,147 @@
...
+		foreach($this->itemTime as $time){
+			$times[] = new IntTag($time);
+		}
+		$nbt->setTag(self::TAG_ITEM_TIME, new ListTag($times));
+	}
+
+	protected function addAdditionalSpawnData(CompoundTag $nbt) : void{
+		foreach($this->items as $slot => $item){
+			$nbt->setTag("Item" . $slot, $item->nbtSerialize());
+			$nbt->setInt("ItemTime" . $slot, $this->getItemTime($slot));
```
dktapps: seriously?

src/block/Campfire.php
```diff
@@ -0,0 +1,193 @@
...
+
+	public function onEntityInside(Entity $entity) : bool{
+		if($this->extinguished){
+			if($entity->isOnFire()){
+				$this->fire();
+				return true;
+			}
+			return false;
+		}
+		if($entity instanceof SplashPotion && $entity->getPotionType()->getDisplayName() == PotionType::WATER()->getDisplayName()){
```
dktapps:
```suggestion
		if($entity instanceof SplashPotion && $entity->getPotionType()->getDisplayName() === PotionType::WATER()->getDisplayName()){
```

src/block/Campfire.php
```diff
@@ -0,0 +1,193 @@
...
+
+	public function onScheduledUpdate() : void{
+		$tile = $this->position->getWorld()->getTile($this->position);
+
+		if($tile instanceof TileCampfire && !$tile->closed){
+			if(!$this->extinguished){
+				foreach($tile->getContents() as $slot => $item){
+					$tile->increaseSlotTime($slot);
+					if($tile->getItemTime($slot) >= $tile->getFurnaceType()->getCookDurationTicks()){
+						$tile->setItem(ItemFactory::air(), $slot);
```
dktapps:
```suggestion
						$tile->setItem(VanillaItems::AIR(), $slot);
```

src/block/Campfire.php
```diff
@@ -0,0 +1,193 @@
...
+						$tile->setSlotTime($slot, 0);
+						$this->position->world->setBlock($this->position, $this);
+						$result = ($item = $this->position->getWorld()->getServer()->getCraftingManager()->getFurnaceRecipeManager($tile->getFurnaceType())->match($item)) instanceof FurnaceRecipe ? $item->getResult() : ItemFactory::air();
```
dktapps:
```suggestion
						$result = ($item = $this->position->getWorld()->getServer()->getCraftingManager()->getFurnaceRecipeManager($tile->getFurnaceType())->match($item)) instanceof FurnaceRecipe ? $item->getResult() : VanillaItems::AIR();
```

src/block/tile/Campfire.php
ipad54: Because ContainerTrait have getRealInventory() method, but Campfire doesn't have inventory

dktapps: Inventory is item storage. It doesn't matter if the player can see it or not.

JavierLeon9966:
~~Also keep in mind that when a projectile that is on fire lands on the campfire it should lit on fire.~~ My bad, didn’t notice this code:
https://github.com/pmmp/PocketMine-MP/blob/e4c4c5c66595fbe799a74c8cbdcef05150409b3d/src/block/Campfire.php#L157-L163

src/world/sound/ItemFrameAddItemSound.php
ipad54: How would you name this sound?

src/block/VanillaBlocks.php
```diff
@@ -694,6 +693,7 @@ protected static function setup() : void{
 		self::register("cut_sandstone", $factory->get(24, 2));
 		self::register("cut_sandstone_slab", $factory->get(421, 3));
 		self::register("cyan_glazed_terracotta", $factory->get(229, 2));
+		self::register("campfire", $factory->get(464, 0));
```
dktapps: This is still in the wrong place. It should be between `cake` and `carpet`.

If you're using PhpStorm, select the lines, go Edit -> Sort Lines if you don't know how to sort alphabetically.

src/block/Campfire.php
```diff
@@ -0,0 +1,264 @@
...
+				$this->increaseSlotTime($slot);
+				if($this->getSlotTime($slot) >= $this->getFurnaceType()->getCookDurationTicks()){
+					$this->setItem(VanillaItems::AIR(), $slot);
+					$this->setSlotTime($slot, 0);
+					$result = ($item = $this->position->getWorld()->getServer()->getCraftingManager()->getFurnaceRecipeManager($this->getFurnaceType())->match($item)) instanceof FurnaceRecipe ? $item->getResult() : VanillaItems::AIR();
+					$this->position->getWorld()->dropItem($this->position->add(0, 1, 0), $result);
+				}
+			}
+			$this->position->getWorld()->setBlock($this->position, $this);
+			if(!empty($this->items)){
```
jasonw4331: Following phpstan's analysis, this needs to be a more strict comparison. Don't to use empty() when you can check the count of items

src/block/Campfire.php
```diff
@@ -0,0 +1,264 @@
...
+			$this->extinguish();
+			return true;
+		}elseif($entity instanceof Living){
+			$entity->attack(new EntityDamageByBlockEvent($this, $entity, EntityDamageEvent::CAUSE_FIRE, 1));
+			$entity->setOnFire(8);
+		}
+		return false;
+	}
+
+	public function onScheduledUpdate() : void{
```
jasonw4331, line 232: Campfire sounds will also need implemented here similar to #4755

jasonw4331 (changes requested):
Mainly just the campfire sounds and phpstan fixes need implemented

src/block/Campfire.php
```diff
@@ -0,0 +1,231 @@
...
+	/** @phpstan-var array<int, int> */
+	protected array $cookingTimes = [];
+
+	public function writeStateToMeta() : int{
+		return BlockDataSerializer::writeHorizontalFacing($this->facing) | (!$this->lit ? BlockLegacyMetadata::CAMPFIRE_FLAG_EXTINGUISHED : 0);
+	}
+
+	public function readStateFromData(int $id, int $stateMeta) : void{
+		$this->facing = BlockDataSerializer::readHorizontalFacing($stateMeta & 0x03);
+		$this->lit = ($stateMeta & BlockLegacyMetadata::CAMPFIRE_FLAG_EXTINGUISHED) !== BlockLegacyMetadata::CAMPFIRE_FLAG_EXTINGUISHED;
```
dktapps:
```suggestion
		$this->lit = ($stateMeta & BlockLegacyMetadata::CAMPFIRE_FLAG_EXTINGUISHED) === 0;
```

src/block/Campfire.php
```diff
@@ -0,0 +1,231 @@
...
+		return $this->lit;
+	}
+
+	/** @return $this */
+	public function setLit(bool $lit = true) : self{
+		$this->lit = $lit;
+		return $this;
+	}
+
+	public function getFurnaceType() : FurnaceType{
```
dktapps: This function is redundant. It's not required by any interface or abstraction, and if we're dealing with an instance of `Campfire`, we already know it's a campfire.

src/block/Campfire.php
```diff
@@ -0,0 +1,231 @@
...
+	public function getInventory() : CampfireInventory{
+		return $this->inventory;
+	}
+
+	public function isLit() : bool{
+		return $this->lit;
+	}
+
+	/** @return $this */
+	public function setLit(bool $lit = true) : self{
```
dktapps: Defaults should not be used here.

src/block/Campfire.php
```diff
@@ -0,0 +1,231 @@
...
+		if($tile instanceof TileCampfire){
+			$tile->setCookingTimes($this->cookingTimes);
+		}
+	}
+
+	public function getStateBitmask() : int{
+		return 0b111;
+	}
+
+	public function hasEntityCollision() : bool{
```
dktapps, line 106: This function is redundant

src/block/Campfire.php
```diff
@@ -0,0 +1,231 @@
...
+use function count;
+use function mt_rand;
+
+class Campfire extends Transparent{
+	use FacesOppositePlacingPlayerTrait;
+	use HorizontalFacingTrait;
+
+	protected bool $lit = true;
+	protected CampfireInventory $inventory;
+	/** @phpstan-var array<int, int> */
```
dktapps: This should be accompanied by a `@var` as well

src/block/Campfire.php
```diff
@@ -0,0 +1,231 @@
...
+	public function setLit(bool $lit = true) : self{
+		$this->lit = $lit;
+		return $this;
+	}
+
+	public function getFurnaceType() : FurnaceType{
+		return FurnaceType::CAMPFIRE();
+	}
+
+	public function setCookingTime(int $slot, int $time) : void{
```
dktapps, line 147: This needs validation of the parameters.

src/block/Campfire.php
```diff
@@ -0,0 +1,231 @@
...
+			}
+			if($item instanceof Shovel && $this->lit){
+				$item->applyDamage(1);
+				$this->extinguish();
+				return true;
+			}
+
+			$ingredient = clone $item;
+			$ingredient->setCount(1);
+			if($this->inventory->canAddItem($ingredient)){
```
dktapps:
```suggestion
			if(count($this->inventory->addItem($ingredient)) === 0){
```
This wastes less time checking if the item can be added when you're going to add it anyway.

src/block/Campfire.php
```diff
@@ -0,0 +1,231 @@
...
+				$this->inventory->addItem($ingredient);
```
dktapps: don't forget to remove this after changing it

src/block/Campfire.php
```diff
@@ -0,0 +1,231 @@
...
+				return true;
+			}
+			return false;
+		}
+		if($entity instanceof SplashPotion && $entity->getPotionType()->equals(PotionType::WATER())){
+			$this->extinguish();
+			return true;
+		}elseif($entity instanceof Living){
+			$entity->attack(new EntityDamageByBlockEvent($this, $entity, EntityDamageEvent::CAUSE_FIRE, 1));
+			$entity->setOnFire(8);
```
dktapps: Missing return here (unclear if intentional or not)

src/block/inventory/CampfireInventory.php
```diff
@@ -0,0 +1,57 @@
...
+
+	public function canAddItem(Item $item) : bool{
+		return $this->holder->getWorld()->getServer()->getCraftingManager()->getFurnaceRecipeManager(FurnaceType::CAMPFIRE())->match($item) instanceof FurnaceRecipe && parent::canAddItem($item);
+	}
+
+	public function onSlotChange(int $index, Item $before) : void{
+		parent::onSlotChange($index, $before);
+
+		$block = $this->holder->getWorld()->getBlock($this->holder);
+		if($block instanceof Campfire){
```
dktapps: An `InventoryChangeListener` should be used here, instead of baking this directly into the inventory.

src/block/inventory/CampfireInventory.php
```diff
@@ -0,0 +1,57 @@
...
+		$this->holder = $holder;
+		parent::__construct(4);
+	}
+
+	public function getMaxStackSize() : int{
+		return 1;
+	}
+
+	public function canAddItem(Item $item) : bool{
+		return $this->holder->getWorld()->getServer()->getCraftingManager()->getFurnaceRecipeManager(FurnaceType::CAMPFIRE())->match($item) instanceof FurnaceRecipe && parent::canAddItem($item);
```
dktapps: This check should not be here. It's misleading.

src/block/tile/Campfire.php
```diff
@@ -0,0 +1,105 @@
...
+use pocketmine\nbt\tag\ListTag;
+use pocketmine\world\World;
+
+class Campfire extends Spawnable implements Container{
+	use ContainerTrait;
+
+	public const TAG_COOKING_TIMES = "CookingTimes";
+
+	protected CampfireInventory $inventory;
+	/** @phpstan-var array<int, int> */
```
dktapps: This should be accompanied by a traditional `@var`

src/block/tile/Campfire.php
```diff
@@ -0,0 +1,105 @@
...
+
+	/**
+	 * @return CampfireInventory
+	 */
+	public function getRealInventory(){
+		return $this->inventory;
+	}
+
+	/**
+	 * @phpstan-return array<int, int>
```
dktapps, line 77: This should be accompanied by a traditional `@return`

src/block/tile/Campfire.php
```diff
@@ -0,0 +1,105 @@
...
+	 */
+	public function getCookingTimes() : array{
+		return $this->cookingTimes;
+	}
+
+	/**
+	 * @phpstan-param array<int, int> $cookingTimes
```
dktapps, line 85: This should be accompanied by a traditional `@param`

src/block/tile/Campfire.php
```diff
@@ -0,0 +1,105 @@
...
+		$this->cookingTimes = $cookingTimes;
+	}
+
+	public function readSaveData(CompoundTag $nbt) : void{
+		$this->loadItems($nbt);
+
+		if(($tag = $nbt->getTag(self::TAG_COOKING_TIMES)) instanceof ListTag){
+			/**
+			 * @var int $slot
+			 * @var IntTag $time
```
dktapps: This is not guaranteed to be an `IntTag`. You haven't checked the type of tag that the list contains.

dktapps (dismissed):

src/block/Campfire.php
ipad54: Why? Without this function, onEntityInside() will not be called

dktapps: My mistake, I mistook it for something else

IvanCraft623:
The code for this PR looks pretty good! From a little glance I can tell it needs updating and a little documentation, I'll be picking up this PR.

src/block/Campfire.php
```diff
@@ -0,0 +1,234 @@
...
+use pocketmine\player\Player;
+use pocketmine\utils\Limits;
+use pocketmine\world\BlockTransaction;
+use pocketmine\world\sound\CampfireSound;
+use pocketmine\world\sound\FireExtinguishSound;
+use pocketmine\world\sound\FlintSteelSound;
+use function count;
+use function mt_rand;
+
+class Campfire extends Transparent{
```
IvanCraft623, line 62: This could use `LightableTrait`

src/block/Campfire.php
```diff
@@ -0,0 +1,234 @@
...
+	use HorizontalFacingTrait{
+		describeBlockOnlyState as encodeFacingState;
+	}
+
+	protected bool $lit = true;
+	protected CampfireInventory $inventory;
+	/** @var array<int, int> */
+	protected array $cookingTimes = [];
```
IvanCraft623, line 78: Please document this, what unit of time?

src/block/Campfire.php
```diff
@@ -0,0 +1,234 @@
...
+		$w->bool($this->lit);
+	}
+
+	public function readStateFromWorld() : Block{
+		parent::readStateFromWorld();
+		$tile = $this->position->getWorld()->getTile($this->position);
+		if($tile instanceof TileCampfire){
+			$this->inventory = $tile->getInventory();
+			$this->cookingTimes = $tile->getCookingTimes();
+		}
```
IvanCraft623: You should also initialize inventory on else case, otherwise it will generate an error when trying to cook something. Or throw an exception

src/block/Campfire.php
```diff
@@ -0,0 +1,234 @@
...
+		return $this->lit;
+	}
+
+	/** @return $this */
+	public function setLit(bool $lit) : self{
+		$this->lit = $lit;
+		return $this;
+	}
+
+	public function setCookingTime(int $slot, int $time) : void{
```
IvanCraft623, line 147: Document the units

IvanCraft623 (changes requested):

src/block/Campfire.php
```diff
@@ -0,0 +1,240 @@
...
+
+	private function ignite() : void{
+		$this->position->getWorld()->addSound($this->position, new FlintSteelSound());
+		$this->position->getWorld()->setBlock($this->position, $this->setLit(true));
+	}
+
+	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
+		if(!$this->getSide(Facing::DOWN)->getSupportType(Facing::UP)->hasCenterSupport()){
+			return false;
+		}
```
IvanCraft623: According to my tests in vanilla campfires doesn't have any placement restrictions

src/block/tile/Campfire.php
```diff
@@ -0,0 +1,122 @@
...
+				$nbt->setInt(self::TAG_COOKING_TIME . $slot, $cookingTime);
+			}
+		}
+	}
+
+	protected function addAdditionalSpawnData(CompoundTag $nbt) : void{
+		for($slot = 1; $slot <= 4; $slot++){
+			$item = $this->inventory->getItem($slot - 1);
+			if(!$item->isNull()){
+				$nbt->setTag(self::TAG_ITEM . $slot, $item->nbtSerialize());
```
IvanCraft623: Please don't read/save data like this, instead you should add the four constants, and map them to each slot. Something like:
```php
foreach ([
	0 => self::TAG_FIRST_COOKING_ITEM,
	...
] as $slot => $tag) {
	// code...
}
```

src/block/Campfire.php
```diff
@@ -0,0 +1,240 @@
...
+		if($player !== null){
+			$this->facing = $player->getHorizontalFacing();
+		}
+		$this->lit = true;
+		return parent::place($tx, $item, $blockReplace, $blockClicked, $face, $clickVector, $player);
+	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
```
IvanCraft623, line 175: Campfires can also be ignited by using a fire charge or an item enchanted with fire aspect

IvanCraft623 (changes requested):

src/block/Campfire.php
ipad54: I definitely remember it had more restrictions, but now there is only one: campfire cannot be placed on another campfire

src/block/Campfire.php
```diff
@@ -0,0 +1,251 @@
...
```
IvanCraft623, line 175: Current implementarion doesn't allow to add items when the campfire is extingished, in vanilla you can do it

src/block/tile/Campfire.php
```diff
@@ -0,0 +1,154 @@
...
+		foreach([
+			0 => self::TAG_FIRST_COOKING_TIME,
+			1 => self::TAG_SECOND_COOKING_TIME,
+			2 => self::TAG_THIRD_COOKING_TIME,
+			3 => self::TAG_FOURTH_COOKING_TIME
+		] as $slot => $tag){
+			if(($tag = $nbt->getTag($tag)) instanceof IntTag){
+				$this->cookingTimes[$slot] = $tag->getValue();
+			}
+		}
```
IvanCraft623: You can reduce all this with something like:

```php
foreach ([
	[0, self::TAG_FIRST_ITEM, self::TAG_FIRST_COOKING_TIME]
	...
] as [$slot, $itemTag, $cookingTimeTag]) {
	// code...
}
```

src/block/tile/Campfire.php
```diff
@@ -0,0 +1,154 @@
...
+use pocketmine\nbt\tag\CompoundTag;
+use pocketmine\nbt\tag\IntTag;
+use pocketmine\world\World;
+
+class Campfire extends Spawnable implements Container{
+	use ContainerTrait;
+	private const TAG_FIRST_ITEM = "Item1";
+	private const TAG_SECOND_ITEM = "Item2";
+	private const TAG_THIRD_ITEM = "Item3";
+	private const TAG_FOURTH_ITEM = "Item4";
```
IvanCraft623: I'm not sure if this is the more accurate name, `ITEM` is not very specific about what is.

I think that perhaps one of the following options would be a better fit: `TAG_N_COOKING_ITEM`, `TAG_N_INPUT_ITEM`, `TAG_N_INGREDIENT`.

IvanCraft623: I would also like others to suggest which one they prefer or to propose their options

src/block/Campfire.php
```diff
@@ -0,0 +1,251 @@
...
+
+	protected function recalculateCollisionBoxes() : array{
+		return [AxisAlignedBB::one()->trim(Facing::UP, 9 / 16)];
+	}
+
+	public function getInventory() : CampfireInventory{
+		return $this->inventory;
+	}
+
+	/** Sets the number of ticks left to cook the item in the given slot */
```
IvanCraft623: Please follow the function documentation style of the project

IvanCraft623: This documentation seems to be wrong, it should be: `Sets the number of ticks during the item in the given slot has been cooked`

Since it does not return the left ticks to finish the cooking process

src/block/Campfire.php
```diff
@@ -0,0 +1,251 @@
...
+	public function setCookingTime(int $slot, int $time) : void{
+		if($slot < 0 || $slot > 3){
+			throw new \InvalidArgumentException("Slot must be in range 0-3");
+		}
+		if($time < 0 || $time > Limits::INT32_MAX){
```
IvanCraft623: This max limit doesn't make sense it should be actually `FurnaceType::CAMPFIRE->getCookDurationTicks()`

src/block/Campfire.php
```diff
@@ -0,0 +1,251 @@
...
+		$this->position->getWorld()->addSound($this->position, new FlintSteelSound());
+		$this->position->getWorld()->setBlock($this->position, $this->setLit(true));
+	}
+
+	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
+		if($this->getSide(Facing::DOWN)->hasSameTypeId($this)){
+			return false;
+		}
+		if($player !== null){
+			$this->facing = $player->getHorizontalFacing();
```
IvanCraft623, line 169: This was changed by me. But it seems that facing should actually be `Facing::opposite($player->getHorizontalFacing())`. Although the actual code matches vanilla, it is likely to be a bug. Similar to #6375

IvanCraft623 (changes requested):

src/block/Campfire.php
IvanCraft623: Reported as: https://bugs.mojang.com/browse/MCPE-183478

src/block/tile/Campfire.php
ipad54: I think ```TAG_N_INPUT_ITEM``` would be the best option because ```INGREDIENT``` is too generic concept and  ```COOKING_ITEM``` can be easily misread as ```COOKING_TIME```

src/block/Campfire.php
```diff
@@ -0,0 +1,256 @@
...
+					$this->setCookingTime($slot, 0);
+					$result = ($item = $this->position->getWorld()->getServer()->getCraftingManager()->getFurnaceRecipeManager(FurnaceType::CAMPFIRE)->match($item)) instanceof FurnaceRecipe ? $item->getResult() : VanillaItems::AIR();
+					$this->position->getWorld()->dropItem($this->position->add(0, 1, 0), $result);
+				}
+			}
+			if(count($items) > 0){
+				$this->position->getWorld()->setBlock($this->position, $this);
+			}
+			if(mt_rand(1, 10) === 1){
+				$this->position->getWorld()->addSound($this->position, new CampfireSound());
```
IvanCraft623: I'm pretty sure this should be more frecuent, would have to get the intervals from the BDS observation

IvanCraft623:
Soul campfire variant is missing :P

src/block/Campfire.php
ipad54: It plays between 1 to 5 seconds according to https://github.com/pmmp/PocketMine-MP/blob/stable/src/block/Furnace.php#L75 and my personal tests

src/block/Campfire.php
```diff
@@ -0,0 +1,256 @@
...
+			}
+
+			$this->position->getWorld()->scheduleDelayedBlockUpdate($this->position, 20);
```
ipad54: It's a little bit incorrect: in vanilla cooking time increases every tick. I made 20 here because block starts to "twitch" when destroyed due to very frequent ```setBlock()``` calls. The only solution is to interact with the tile directly (how it's made in furnaces), but Dylan wouldn't like that.

IvanCraft623: There is no need to reduce the update intervals, it won't make much difference to the player and reducing them only means more work for the server. Vanilla parity is not worth it in this case.

src/block/Campfire.php
```diff
@@ -0,0 +1,267 @@
...
+				$this->setCookingTime($slot, min($maxCookDuration, $this->getCookingTime($slot) + 20));
+				if($this->getCookingTime($slot) >= $maxCookDuration){
+					$this->inventory->setItem($slot, VanillaItems::AIR());
+					$this->setCookingTime($slot, 0);
+					$result = ($item = $this->position->getWorld()->getServer()->getCraftingManager()->getFurnaceRecipeManager($furnaceType)->match($item)) instanceof FurnaceRecipe ? $item->getResult() : VanillaItems::AIR();
+					$this->position->getWorld()->dropItem($this->position->add(0, 1, 0), $result);
+				}
+			}
+			if(count($items) > 0){
+				$this->position->getWorld()->setBlock($this->position, $this);
```
IvanCraft623, line 258: What is the purpose of this `setBlock()`? Isn't it already updating in case of any change, thanks to `CallbackInventoryListener::onAnyChange()` ?

ipad54: Cooking time increasing isn't covered by it

src/block/Campfire.php
```diff
@@ -0,0 +1,269 @@
...
+			if($item->getTypeId() === ItemTypeIds::FIRE_CHARGE){
+				$item->pop();
+				$this->ignite();
+				$this->position->getWorld()->addSound($this->position, new BlazeShootSound());
+				return true;
+			}elseif($item->getTypeId() === ItemTypeIds::FLINT_AND_STEEL || $item->hasEnchantment(VanillaEnchantments::FIRE_ASPECT())){
+				if($item instanceof Durable){
+					$item->applyDamage(1);
+				}
+				$this->ignite();
```
ShockedPlot7560, line 186: no sound for flint and steel and fire aspect ?

src/block/Campfire.php
```diff
@@ -0,0 +1,269 @@
...
+		}
+	}
+
+	public function onEntityInside(Entity $entity) : bool{
+		if(!$this->lit){
+			if($entity->isOnFire()){
+				$this->ignite();
+				return false;
+			}
+			return true;
```
ShockedPlot7560: return is not needed here

src/block/Campfire.php
```diff
@@ -0,0 +1,269 @@
...
+		if($this->lit){
+			$items = $this->inventory->getContents();
+			$furnaceType = $this->getFurnaceType();
+			$maxCookDuration = $furnaceType->getCookDurationTicks();
+			foreach($items as $slot => $item){
+				$this->setCookingTime($slot, min($maxCookDuration, $this->getCookingTime($slot) + self::UPDATE_INTERVAL_TICKS));
+				if($this->getCookingTime($slot) >= $maxCookDuration){
+					$this->inventory->setItem($slot, VanillaItems::AIR());
+					$this->setCookingTime($slot, 0);
+					$result = ($item = $this->position->getWorld()->getServer()->getCraftingManager()->getFurnaceRecipeManager($furnaceType)->match($item)) instanceof FurnaceRecipe ? $item->getResult() : VanillaItems::AIR();
```
ShockedPlot7560: this one line is difficult to read

src/block/Campfire.php
ipad54: The sound is played in ```ignite()```

src/block/Campfire.php
```diff
@@ -0,0 +1,269 @@
...
+		$this->lit = true;
+		return parent::place($tx, $item, $blockReplace, $blockClicked, $face, $clickVector, $player);
+	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if(!$this->lit){
+			if($item->getTypeId() === ItemTypeIds::FIRE_CHARGE){
+				$item->pop();
+				$this->ignite();
+				$this->position->getWorld()->addSound($this->position, new BlazeShootSound());
```
ShockedPlot7560, line 180: So two sound will be played ?

ipad54: Yes, same happens in vanilla.

IvanCraft623 (dismissed):
Looks good enough for me

src/block/Campfire.php
```diff
@@ -0,0 +1,280 @@
...
+
+	private function extinguish() : void{
+		$this->position->getWorld()->addSound($this->position, new FireExtinguishSound());
+		$this->position->getWorld()->setBlock($this->position, $this->setLit(false));
+	}
+
+	private function ignite() : void{
+		$this->position->getWorld()->addSound($this->position, new FlintSteelSound());
+		$this->position->getWorld()->setBlock($this->position, $this->setLit(true));
+		$this->position->getWorld()->scheduleDelayedBlockUpdate($this->position, 20);
```
ipad54: @IvanCraft623 you have probably missed this place when adding the update interval ticks constant

ipad54 (commented):
Looks finished for me now

src/block/Campfire.php
```diff
@@ -0,0 +1,278 @@
...
+					$this->inventory->setItem($slot, VanillaItems::AIR());
+					$this->setCookingTime($slot, 0);
+					$this->position->getWorld()->dropItem($this->position->add(0.5, 1, 0.5), $ev->getResult());
+				}
+			}
+			if(count($items) > 0){
+				$this->position->getWorld()->setBlock($this->position, $this);
+			}
+			if(mt_rand(1, 6) === 1){
+				$this->position->getWorld()->addSound($this->position, new CampfireSound());
```
nicholass003: would it be better if uses 
```php
$this->getFurnaceType()->getCookSound();
```
 instead of declaring a new Sound class?

IvanCraft623 (dismissed):

nicholass003 (approved):

src/block/Campfire.php
```diff
@@ -0,0 +1,277 @@
...
+	private function extinguish() : void{
+		$this->position->getWorld()->addSound($this->position, new FireExtinguishSound());
+		$this->position->getWorld()->setBlock($this->position, $this->setLit(false));
+	}
+
+	private function ignite() : void{
+		$this->position->getWorld()->addSound($this->position, new FlintSteelSound());
+		$this->position->getWorld()->setBlock($this->position, $this->setLit(true));
+		$this->position->getWorld()->scheduleDelayedBlockUpdate($this->position, self::UPDATE_INTERVAL_TICKS);
+	}
```
ShockedPlot7560: By convention, private methods are often placed at the end?

ShockedPlot7560 (approved):

IvanCraft623 (approved):

tests/phpstan/configs/actual-problems.neon
```diff
@@ -580,6 +580,11 @@ parameters:
 			count: 1
 			path: ../../../src/network/mcpe/ChunkRequestTask.php
 
+		-
+			message: "#^Match expression does not handle remaining values\\: pocketmine\\\\crafting\\\\FurnaceType\\:\\:CAMPFIRE\\|pocketmine\\\\crafting\\\\FurnaceType\\:\\:SOUL_CAMPFIRE$#"
+			count: 1
+			path: ../../../src/network/mcpe/InventoryManager.php
+
```
dktapps, line 587: This is not the way. A `LogicException` or `AssumptionFailedError` in the undesired cases would be more appropriate.
I've fixed it but please don't do this again.

PR #4710, opened 2022-01-06
Implemented kelp

ShockedPlot7560:
## Introduction
Implemented Kelp

### API changes
* Add kelp in registry
### Behavioural changes
N/A

## Tests
[https://youtu.be/LbO1iiyFZX8](https://youtu.be/LbO1iiyFZX8)


src/item/VanillaItems.php
```diff
@@ -133,6 +133,7 @@
  * @method static Item DRAGON_BREATH()
  * @method static Skull DRAGON_HEAD()
  * @method static DriedKelp DRIED_KELP()
+ * @method static ItemBlock KELP()
```
dktapps: We don't put blocks in VanillaItems. `VanillaBlocks::KELP()->asItem()` would be used instead.

src/item/StringToItemParser.php
```diff
@@ -615,6 +615,7 @@ private static function make() : self{
 		$result->registerBlock("jungle_wood", fn() => VanillaBlocks::JUNGLE_WOOD());
 		$result->registerBlock("jungle_wood_stairs", fn() => VanillaBlocks::JUNGLE_STAIRS());
 		$result->registerBlock("jungle_wooden_stairs", fn() => VanillaBlocks::JUNGLE_STAIRS());
+		$result->registerBlock("kelp_block", fn() => VanillaBlocks::KELP());
```
dktapps: StringToItemParser should not include blocks that can't be obtained in item form.

src/block/Kelp.php
```diff
@@ -128,4 +130,11 @@ private function grow(Kelp $block) : bool{
 		}
 		return false;
 	}
+
+	public function asItem(): Item{
+		return ItemFactory::getInstance()->get(
+			ItemIds::KELP,
```
dktapps: This should be specified in the `BlockIdentifier` constructor. It shouldn't be needed here.

ShockedPlot7560: Indeed, oops

src/block/VanillaBlocks.php
```diff
@@ -163,6 +163,7 @@
  * @method static DoubleTallGrass DOUBLE_TALLGRASS()
  * @method static DragonEgg DRAGON_EGG()
  * @method static DriedKelp DRIED_KELP()
+ * @method static Kelp KELP()
```
JavierLeon9966: This should be below `JUNGLE_WOOD()`

src/item/ItemFactory.php
```diff
@@ -96,6 +96,7 @@ public function __construct(){
 		$this->register(new CookedSalmon(new ItemIdentifier(ItemIds::COOKED_SALMON, 0), "Cooked Salmon"));
 		$this->register(new Cookie(new ItemIdentifier(ItemIds::COOKIE, 0), "Cookie"));
 		$this->register(new DriedKelp(new ItemIdentifier(ItemIds::DRIED_KELP, 0), "Dried Kelp"));
+		$this->register(new ItemBlock(new ItemIdentifier(ItemIds::KELP, 0), VanillaBlocks::KELP()));
```
JavierLeon9966: This should be below `ItemIds::IRON_NUGGET`

src/item/StringToItemParser.php
```diff
@@ -1124,6 +1124,7 @@ private static function make() : self{
 		$result->register("iron_shovel", fn() => VanillaItems::IRON_SHOVEL());
 		$result->register("iron_sword", fn() => VanillaItems::IRON_SWORD());
 		$result->register("jungle_boat", fn() => VanillaItems::JUNGLE_BOAT());
+		$result->register("kelp", fn() => VanillaBlocks::KELP()->asItem());
```
JavierLeon9966:
```suggestion
		$result->registerBlock("kelp", fn() => VanillaBlocks::KELP());
```
And also it should be below this:
```php
$result->registerBlock("jungle_wooden_stairs", fn() => VanillaBlocks::JUNGLE_STAIRS());
```

JavierLeon9966 (changes requested):

src/block/VanillaBlocks.php
ShockedPlot7560: These comments are not sorted alphabetically, are they?

JavierLeon9966: They are as you can see: `DOUBLE_TALLGRASS()`, `DRAGON_EGG()`, `DRIED_KELP()`, etc.

dktapps: You can generate the comment using `build/generate-registry-annotations.php`.

ipad54:
After rejoining happens this, and you can restore your breathing in kelp, because PM not support waterlogging
Screenshot_20220107_134742_com mojang minecraftpe
Screenshot_20220107_134738_com mojang minecraftpe

ShockedPlot7560:
True but its like sea pickles or coral, when waterlogged will be available, it will works normally.

ipad54:
In vanilla kelps breaks if there isn't water nearby

src/block/Kelp.php
```diff
@@ -0,0 +1,131 @@
...
+use pocketmine\math\Vector3;
+use pocketmine\player\Player;
+use pocketmine\world\BlockTransaction;
+use function mt_rand;
+
+class Kelp extends Transparent{
+
+	protected int $age = 0;
+
+	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
```
dktapps: This should be further down with the rest of the on* functions.

dktapps (commented):
This implementation is not currently compatible with PM due to the limited size of metadata used. 25 age states require 5 bits to store; PM's metadata storage only allows 4.

dktapps:
Please retarget this to `next-major`

PR #4713, opened 2022-01-07
Replaced `strtolower` and `strtoupper` functions to `mb_strtolower` and `mb_strtoupper`

### LINKED ISSUE #3585: strtoupper() and strtolower() have locale-dependent behaviour which breaks stuff
### Issue description
credits go to @Eren5960 for discovering this problem, but it's much more problematic than just RegistryTrait (#3580).

`strtoupper()` and `strtolower()` behaviour varies depending on locale. As seen in #3580, `strtoupper("inner") !== "INNER"` when the system locale is set to `tr_TR.UTF-8`.

This is a problem for a wide variety of things, such as:
- [x] `ItemFactory::fromString()`
- [x] `Enchantment::getEnchantmentByName()`
- [x] `EventPriority::fromString()`
- [x] `Effect::getEffectByName()`
- [x] `load:` directive in `plugin.yml` if not written in uppercase (potentially)
- [x] `@priority` annotations for event handlers (e.g. `@priority highest` instead of `@priority HIGHEST`)
- [ ] `@ignoreCancelled` annotations for event handlers (potentially)
- [x] API version checks (potentially, not confirmed)
- [ ] `level-type` in `server.properties` (aliases are stored lowercase, but typically written in uppercase)
- [ ] Banlist, whitelist, op list - change in system locale could result in players magically not being whitelisted, not being banned or not being op anymore
- [ ] Resource packs: change of UUID case when loaded might not work correctly depending on locale

the list goes on and on and on...

### OS and versions

* PocketMine-MP: 3.13.0
* PHP: 7.3
* Server OS: linux (probably macOS too, couldn't reproduce on Windows but that doesn't mean it isn't a problem there too)

### Solution
Eliminate and ban usages of `strtolower()` and `strtoupper()` in favour of `mb_strtolower()` and `mb_strtoupper()`. However, this isn't a trivial task because of how many places these functions are used.

JavierLeon9966:
## Introduction

Currently `strtolower` and `strtoupper` changes result according to system locale. This could lead to some bugs listed by the issue below. This pull request replaces every `strtolower` and `strtoupper` to `mb_strtolower` and `mb_strtoupper`.
### Relevant issues

* Fixes #3585
### API changes

None
### Behavioural changes

Any code that uses lowercase or uppercase letters now don’t have locale-dependent behavior.
## Backwards compatibility

None
## Follow-up

None
## Tests

- [ ]


dktapps:
This has lots of side effects and potential ramifications that haven't been explored. There is a reason I haven't already done this myself.

This will cause behavioural changes in many areas, particularly where non-ASCII encodings are involved, since they'll start getting transformed in ways that wasn't anticipated when the code was written. A more conservative approach would be to use `US-ASCII` for the `encoding` parameters.

JavierLeon9966:
Done

dktapps:
This is now obsoleted by PHP 8.2, which made the standard library string manipulation functions locale-unaware (manipulating ASCII only).

PR #4714, opened 2022-01-07
Improve client side performance of FloatingTextParticle

inxomnyaa:
## Introduction

Currently, Floating Text Particles use fake players, which does not only need more packets to be sent to the client, but also degrades client-side performance. Featured servers like Hive are using either custom entities or falling sand.
Tests on NetherGames have also shown some client performance increase. There have been multiple discussions about the performance increase of this feature already in the past, especially on Discord.
### Behavioral changes

Technically yes, since less packets have to be encoded. The floating text particle itself does not behave any different on the client.

## Backwards compatibility

No breaks

## Follow-up
Change to an entity in the future to allow proper visibility toggling and allow for updating the text

## Tests

First image is falling sand, second is Player. works just the same. I have ensured it is just as messed up as before 😜
image
image


dktapps:
As I wrote previously I'm hesitant to make this change simply because of how unreliable floating text has been in the past. The reason it currently uses player entities is because player nametags are guaranteed to work (if they don't there are much bigger problems to worry about).

If this were to be changed, I would recommend using an entity that can normally accept a nametag in vanilla gameplay (e.g. a mob), rather than falling sand, since falling sand entities are not typically able to have nametags. This means that if it breaks Mojang will not notice or care and we will end up fixing it yet again.

dktapps:
After performing some of my own tests and being unable to find an adequate alternative solution, I've decided to merge this as-is. 

My own performance testing with 1000 FTPs in a 10x10x10 area showed an increase from an average 1.5 FPS to 8.0 FPS with this change, so I've decided it's worth the risk. I suppose we can always switch back to players if Mojang breaks something again...

PR #4715, opened 2022-01-08
Implemented End Crystal

ipad54:
## Introduction
Implemented End Crystal

### Behavioural changes
N/A

## Tests
https://www.youtube.com/watch?v=qEakbZGoN9M


ipad54:
setShowBase() method works fine
Screenshot_97

src/entity/object/EnderCrystal.php
```diff
@@ -0,0 +1,111 @@
...
+use pocketmine\network\mcpe\protocol\types\entity\EntityMetadataCollection;
+use pocketmine\network\mcpe\protocol\types\entity\EntityMetadataFlags;
+use pocketmine\world\Explosion;
+use pocketmine\world\Position;
+
+class EnderCrystal extends Entity implements Explosive{
+
+	public static function getNetworkTypeId() : string{ return EntityIds::ENDER_CRYSTAL; }
+
+	public const TAG_SHOWBASE = "ShowBottom"; //TAG_Byte
```
dktapps: This should not be public.

src/entity/object/EnderCrystal.php
```diff
@@ -0,0 +1,111 @@
...
+
+	protected $gravity = 0.0;
+	protected $drag = 0.0;
+
+	protected bool $showBase = false;
+
+	public function __construct(Location $location, ?CompoundTag $nbt = null, bool $showBase = false){
```
dktapps: `showBase` should come before the `nbt` parameter, and should not be optional.

src/entity/object/EnderCrystal.php
```diff
@@ -0,0 +1,111 @@
...
+			}
+			$explosion->explodeB();
+		}
+	}
+
+	protected function syncNetworkData(EntityMetadataCollection $properties) : void{
+		parent::syncNetworkData($properties);
+
+		$properties->setGenericFlag(EntityMetadataFlags::SHOWBASE, $this->showBase);
+		$properties->setGenericFlag(EntityMetadataFlags::ONFIRE, false); //TODO: Hack to prevent fire animation
```
dktapps: `isFireProof()` should be overridden instead of this hack.

src/entity/object/EnderCrystal.php
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	public function updateMovement(bool $teleport = false) : void{
+		//NOOP
+	}
+
+	protected function initEntity(CompoundTag $nbt) : void{
+		parent::initEntity($nbt);
+
+		$this->showBase = $nbt->getByte(self::TAG_SHOWBASE, 0) === 1;
```
dktapps: Make up your mind where you're reading this. It's either in EntityFactory or here, not both.

src/entity/object/EnderCrystal.php
dktapps: We typically don't add constructor parameters unless they are mandatory.

src/entity/object/EnderCrystal.php
```diff
@@ -0,0 +1,111 @@
...
+		if($source->getCause() !== EntityDamageEvent::CAUSE_FIRE && $source->getCause() !== EntityDamageEvent::CAUSE_FIRE_TICK){
+			parent::attack($source);
+			if(!$this->isFlaggedForDespawn() && !$source->isCancelled()){
+				$this->flagForDespawn();
+				$this->explode();
+			}
+		}
+	}
+
+	public function updateMovement(bool $teleport = false) : void{
```
dktapps: wtf is this?

src/item/ItemFactory.php
```diff
@@ -97,6 +97,7 @@ public function __construct(){
 		$this->register(new Cookie(new ItemIdentifier(ItemIds::COOKIE, 0), "Cookie"));
 		$this->register(new DriedKelp(new ItemIdentifier(ItemIds::DRIED_KELP, 0), "Dried Kelp"));
 		$this->register(new Egg(new ItemIdentifier(ItemIds::EGG, 0), "Egg"));
+		$this->register(new EndCrystal(new ItemIdentifier(ItemIds::END_CRYSTAL, 0), "End Crystal"));
```
dktapps: This needs to be added to `StringToItemParser`.

dktapps (dismissed):

src/entity/object/EnderCrystal.php
ipad54: IsFireProof() method does not prevent fire animation
Screenshot_101

JavierLeon9966: Override `isOnFire()`. Apparently that's a bug since in vanilla you aren't on fire with fire resistance or in creative mode

dktapps: Don't override it. If you add a workaround it'll be a pain to get rid of it later once the bug is fixed.

ipad54: This bug finally has been fixed

dktapps:
image
 Not sure this is supposed to work?

ipad54:
Same happens in vanilla

dktapps:
The explosion behaviour seems to differ from vanilla.

In vanilla, no blocks at or below the same Y as the obsidian are destroyed. In this PR, the blocks are destroyed as if a TNT blast went off.

dktapps:
Vanilla EC:
image
Vanilla TNT:
image

src/entity/object/EnderCrystal.php
```diff
@@ -0,0 +1,109 @@
...
+
+	public function setShowBase(bool $showBase) : void{
+		$this->showBase = $showBase;
+		$this->networkPropertiesDirty = true;
+	}
+
+	public function attack(EntityDamageEvent $source) : void{
+		parent::attack($source);
+		if(
+			$source->getCause() !== EntityDamageEvent::CAUSE_VOID &&
```
dktapps: Java code suggests that explosion causes should be added to this (end crystals don't explode when destroyed by other end crystals).

dktapps:
This PR also doesn't include beam targeting (`EntityMetadataProperties::BLOCK_TARGET` property).

src/entity/object/EnderCrystal.php
```diff
@@ -0,0 +1,109 @@
...
+
+		$nbt->setByte(self::TAG_SHOWBASE, $this->showBase ? 1 : 0);
+		return $nbt;
+	}
+
+	public function explode() : void{
+		$ev = new EntityPreExplodeEvent($this, 6);
+		$ev->call();
+		if(!$ev->isCancelled()){
+			$explosion = new Explosion(Position::fromObject($this->location->add(0, $this->size->getHeight() / 2, 0), $this->getWorld()), $ev->getRadius(), $this);
```
dktapps: The vanilla behaviour suggests that the following should be used:
```suggestion
			$explosion = new Explosion(Position::fromObject($this->getPosition(), $ev->getRadius(), $this);
```
This fixes blocks below the support getting destroyed.

src/entity/object/EnderCrystal.php
ipad54: > (end crystals don't explode when destroyed by other end crystals).

They do it in bedrock

ipad54:
beam targeting now works
Screenshot_321

src/entity/object/EnderCrystal.php
```diff
@@ -80,20 +96,33 @@ protected function initEntity(CompoundTag $nbt) : void{
 		$this->setHealth(1);
 
 		$this->setShowBase($nbt->getByte(self::TAG_SHOWBASE, 0) === 1);
+
+		$beamX = $nbt->getInt(self::TAG_BLOCKTARGET_X, 0);
+		$beamY = $nbt->getInt(self::TAG_BLOCKTARGET_Y, 0);
+		$beamZ = $nbt->getInt(self::TAG_BLOCKTARGET_Z, 0);
+
+		if($beamX !== 0 || $beamY !== 0 || $beamZ !== 0){
```
dktapps: This is probably not a good idea. The beam could actually target 0,0,0.

src/entity/object/EnderCrystal.php
```diff
@@ -61,6 +68,15 @@ public function setShowBase(bool $showBase) : void{
 		$this->networkPropertiesDirty = true;
 	}
 
+	public function getBeamPosition() : ?Vector3{
+		return $this->beamPosition;
+	}
+
+	public function setBeamPosition(?Vector3 $beamPosition) : void{
```
dktapps: "beam target" would be a better name.

src/entity/object/EnderCrystal.php
ipad54: When beam position 0, 0, 0 is sent to client, the client doesn't draw beam

dktapps: That's not the point. This should be checked by checking the presence of the tags. Do not fill in defaults.

src/entity/object/EnderCrystal.php
```diff
@@ -0,0 +1,142 @@
...
+		}
+	}
+
+	protected function syncNetworkData(EntityMetadataCollection $properties) : void{
+		parent::syncNetworkData($properties);
+
+		$properties->setGenericFlag(EntityMetadataFlags::SHOWBASE, $this->showBase);
+		if($this->beamTarget !== null){
+			$properties->setBlockPos(EntityMetadataProperties::BLOCK_TARGET, BlockPosition::fromVector3($this->beamTarget));
+		}
```
dktapps: This needs to be filled with something if the beam target is set to NULL, otherwise the old value will continue to apply

src/entity/object/EnderCrystal.php
```diff
@@ -0,0 +1,142 @@
...
+use pocketmine\nbt\tag\CompoundTag;
+use pocketmine\nbt\tag\IntTag;
+use pocketmine\network\mcpe\protocol\types\BlockPosition;
+use pocketmine\network\mcpe\protocol\types\entity\EntityIds;
+use pocketmine\network\mcpe\protocol\types\entity\EntityMetadataCollection;
+use pocketmine\network\mcpe\protocol\types\entity\EntityMetadataFlags;
+use pocketmine\network\mcpe\protocol\types\entity\EntityMetadataProperties;
+use pocketmine\world\Explosion;
+
+class EnderCrystal extends Entity implements Explosive{
```
dktapps: Why the naming inconsistency?

ipad54: In Bedrock Vanilla itself entity is called 'EnderCrystal' and item is called 'EndCrystal'. But i can change the naming if you want

ShockedPlot7560: If we had to stick to all the vanilla quirks, it wouldn't be possible. I think keeping the End crystal name everywhere would be better.

src/item/EndCrystal.php
```diff
@@ -0,0 +1,59 @@
...
+use pocketmine\entity\object\EnderCrystal;
+use pocketmine\math\AxisAlignedBB;
+use pocketmine\math\Facing;
+use pocketmine\math\Vector3;
+use pocketmine\player\Player;
+use function count;
+
+class EndCrystal extends Item{
+
+	public function onInteractBlock(Player $player, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, array &$returnedItems) : ItemUseResult{
```
ShockedPlot7560, line 38: Is it possible to place a crystal by clicking on one side and it would then spawn on top? If not, this code allows it by not checking the clicked side.

src/item/EndCrystal.php
```diff
@@ -0,0 +1,59 @@
...
+		if($blockClicked->getTypeId() === BlockTypeIds::OBSIDIAN || $blockClicked->getTypeId() === BlockTypeIds::BEDROCK){
+			$pos = $blockClicked->getPosition();
+			$world = $pos->getWorld();
+			$bb = AxisAlignedBB::one()
+				->offset($pos->getX(), $pos->getY(), $pos->getZ())
+				->extend(Facing::UP, 1);
+			if(
+				count($world->getNearbyEntities($bb)) === 0 &&
```
ShockedPlot7560, line 46: We can't put a crystal on a block where there are already items, so? Or is this code just to check that there are no crystals already there? Because if it checks for the presence of crystal, it checks more generally for the presence of any entity

src/item/EndCrystal.php
ipad54: In bedrock vanilla you can click on any side and the crystal will spawn on top

src/item/EndCrystal.php
ipad54: In bedrock vanilla you can't spawn crystal if there are any entities on top (such as players, projectiles, items, etc)

src/entity/object/EnderCrystal.php
dktapps: Consistency within PM is more important.

Nexus9w9 (approved):

src/item/EndCrystal.php
```diff
@@ -0,0 +1,59 @@
...
+				$world->getBlock($pos->up())->getTypeId() === BlockTypeIds::AIR &&
+				$world->getBlock($pos->up(2))->getTypeId() === BlockTypeIds::AIR
```
ShockedPlot7560: getSide would be better or not ?

ipad54: Probably yes

ShockedPlot7560 (dismissed):
Look ok to me

ipad54:
The only thing is that I need to fix CS. I’ll do it a little bit later, when I come home

src/item/EndCrystal.php
```diff
@@ -0,0 +1,59 @@
...
+use pocketmine\player\Player;
+use function count;
+
+class EndCrystal extends Item{
+
+	public function onInteractBlock(Player $player, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, array &$returnedItems) : ItemUseResult{
+		if($blockClicked->getTypeId() === BlockTypeIds::OBSIDIAN || $blockClicked->getTypeId() === BlockTypeIds::BEDROCK){
+			$pos = $blockClicked->getPosition();
+			$world = $pos->getWorld();
+			$bb = AxisAlignedBB::one()
```
JavierLeon9966, line 42: This should be a 2x2x2 BB because of recent vanilla changes. You can no longer place an ender crystal next to one. Don't forget to offset every axis by 0.5 to center it.

ShockedPlot7560: Really ? ... :/

ipad54: If you mean that there need to be a distance in 1 block between 2 crystals then the current code works fine since the crystal spawns at the centered position

JavierLeon9966: 1. End crystal
2. Air
3. End crystal
That's how it should be (imagine it horizontally). The end crystal does spawn centered but that BB isn't centered.

JavierLeon9966: After further testing, it seems I got confused and the BB isn't centered in the block corner, it *starts* from the corner and the 1x1x2 BB that starts inside the block is somehow the correct behavior in vanilla, which I thought it was a bug here, but seems like a bug in bedrock's side

dktapps (dismissed):
This should be retested with latest code & game version, but the code looks OK

dktapps (dismissed):

ipad54:
After thorough testing it seems that it's fully replicates the vanilla behaviour except the explosion damage & knockback part. In vanilla, if you equip a netherite armour set, place a crystal and stand on the top of obsidian, you will die immediately. In PM you will survive with 1 health point remaining. However, if a crystal is placed on 1-2 blocks higher than your position, you will receive much less damage in vanilla than in PM. Knockback also differs a little. 

This is not only the end crystal problem, same happens with TNT. If you go directly into the Primed TNT entity in vanilla with netherite set equipped, you will die immediately too, while in PM you will get only the half of that damage.

dktapps:
Yeah that doesn't surprise me. I'm happy to merge as is given that these features weren't implemented by the PR.

dktapps (approved):

dktapps:
Actually, one more thing: Can someone check if middle-click on an end crystal is supposed to give its item? Like #5397

Gaprix:
I checked on vanilla, and after a middle-click, the crystal is given to the player (or selected in the hotbar)

PR #4723, opened 2022-01-11
Incorrect BlockBreakInfo

leolee3914:
## Tests

test with plugin
```
	var $t = 0;
	
	function onEnable () : void {
		$this->getServer()->getPluginManager()->registerEvents($this,$this);
	}
	
	function tap ( PlayerInteractEvent $e ) {
		if ( $e->getAction() === PlayerInteractEvent::LEFT_CLICK_BLOCK ) {
			$this->t = hrtime(true);
		}
	}
	
	function bb ( BlockBreakEvent $e ) {
		$e->getPlayer()->sendMessage('T = ' . round((hrtime(true) - $this->t) / 1_000_000_000, 2) . 's');
		$this->t = 0;
	}
```
Result (PE, Android):

Bamboo: https://minecraft.fandom.com/wiki/Bamboo
- wooden sword: 0, 0, 0.21, 0.2, 0.21, 0.2 => 0
- wooden sword: 0, 0, 0, 0, 0, 0.2, 0, 0, 0, 0 => 0
- sometimes this will be 0.05 or 0.2.  i think this should be considered instant break because it seems the unexplained delay occurs

Nether wart blocks: https://minecraft.fandom.com/wiki/Nether_Wart_Block
- no tool: 1.46, 1.51, 1.5, 1.51 => 1.5
- wooden hoe: 0.71, 0.75, 0.76 => 0.75
- diamond hoe: 0.15, 0.2, 0.2 => 0.2

Infested Stone: https://minecraft.fandom.com/wiki/Infested_Block
- no tool: 1.1, 1.15, 1.16, 1.15 => 1.15 (97:0 - 97:5 have the same result)
- diamond pickaxe: 0.1, 0.16, 0.15, 0.15 => 0.15 (97:0 - 97:5 have the same result)

Lily Pad: https://minecraft.fandom.com/wiki/Lily_Pad
- no tool: 0

Leaves: https://minecraft.fandom.com/wiki/Leaves
- no tool: 0.25, 0.31, 0.3, 0.3 => 0.3
- wooden sword: 0.15, 0.2, 0.2, 0.21 => 0.2
- stone hoe: 0.05, 0.1, 0.1, 0.1 => 0.1
- iron hoe: 0.0, 0.05, 0.05, 0.05 => 0 (instant break)
- shears: 0.0, 0.05, 0.05, 0.05 => 0 (instant break)

I have not tested the return of BlockBreakInfo->getBreakTime() and BlockBreakInfo->isToolCompatible(), pls tell me if it is necessary

### DIFF
--- a/src/block/BlockFactory.php
+++ b/src/block/BlockFactory.php
@@ -458,11 +458,10 @@ class BlockFactory{
 		$planksBreakInfo = new BlockBreakInfo(2.0, BlockToolType::AXE, 0, 15.0);
 		$leavesBreakInfo = new class(0.2, BlockToolType::HOE) extends BlockBreakInfo{
 				public function getBreakTime(Item $item) : float{
-					return match($item->getBlockToolType()){
-						BlockToolType::SWORD => 0.2,
-						BlockToolType::SHEARS => 0.0,
-						default => parent::getBreakTime($item),
-					};
+					if($item->getBlockToolType() === BlockToolType::SHEARS){
+						return 0.0;
+					}
+					return parent::getBreakTime($item);
 				}
 			};
 		$signBreakInfo = new BlockBreakInfo(1.0, BlockToolType::AXE);
--- a/src/block/BlockFactory.php
+++ b/src/block/BlockFactory.php
@@ -115,14 +115,7 @@ class BlockFactory{
 		$this->registerAllMeta(new ActivatorRail(new BID(Ids::ACTIVATOR_RAIL, 0), "Activator Rail", $railBreakInfo));
 		$this->registerAllMeta(new Air(new BID(Ids::AIR, 0), "Air", BlockBreakInfo::indestructible(-1.0)));
 		$this->registerAllMeta(new Anvil(new BID(Ids::ANVIL, 0), "Anvil", new BlockBreakInfo(5.0, BlockToolType::PICKAXE, ToolTier::WOOD()->getHarvestLevel(), 6000.0)));
-		$this->registerAllMeta(new Bamboo(new BID(Ids::BAMBOO, 0), "Bamboo", new class(2.0 /* 1.0 in PC */, BlockToolType::AXE) extends BlockBreakInfo{
-			public function getBreakTime(Item $item) : float{
-				if($item->getBlockToolType()){
-					return 0.0;
-				}
-				return parent::getBreakTime($item);
-			}
-		}));
+		$this->registerAllMeta(new Bamboo(new BID(Ids::BAMBOO, 0), "Bamboo", new BlockBreakInfo(2.0 /* 1.0 in PC */, BlockToolType::AXE)));
 		$this->registerAllMeta(new BambooSapling(new BID(Ids::BAMBOO_SAPLING, 0), "Bamboo Sapling", BlockBreakInfo::instant()));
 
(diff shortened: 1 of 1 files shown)


src/block/BlockFactory.php
```diff
@@ -365,7 +372,7 @@ public function isAffectedBySilkTouch() : bool{
 			$crackedStoneBrick = new Opaque(new BID(Ids::STONEBRICK, Meta::STONE_BRICK_CRACKED), "Cracked Stone Bricks", $stoneBreakInfo),
 			$chiseledStoneBrick = new Opaque(new BID(Ids::STONEBRICK, Meta::STONE_BRICK_CHISELED), "Chiseled Stone Bricks", $stoneBreakInfo)
 		);
-		$infestedStoneBreakInfo = new BlockBreakInfo(0.75);
+		$infestedStoneBreakInfo = new BlockBreakInfo(0.75, BlockToolType::PICKAXE);
```
jasonw4331, line 375: All infested blocks are actually breakable with the fist, not any specific tool in bedrock edition

jasonw4331 (changes requested):

src/block/BlockFactory.php
leolee3914: so BlockBreakInfo->toolHarvestLevel is default (0) and BlockBreakInfo->isToolCompatible() should always return true
is that correct?

dktapps: > All infested blocks are actually breakable with the fist, not any specific tool in bedrock edition

This change doesn't affect that, it only sets the preferred tool type.

src/block/BlockFactory.php
```diff
@@ -441,15 +448,23 @@ public function isAffectedBySilkTouch() : bool{
...
 		$this->registerAllMeta(new WeightedPressurePlateHeavy(new BID(Ids::HEAVY_WEIGHTED_PRESSURE_PLATE, 0), "Weighted Pressure Plate Heavy", $weightedPressurePlateBreakInfo));
 		$this->registerAllMeta(new WeightedPressurePlateLight(new BID(Ids::LIGHT_WEIGHTED_PRESSURE_PLATE, 0), "Weighted Pressure Plate Light", $weightedPressurePlateBreakInfo));
 		$this->registerAllMeta(new Wheat(new BID(Ids::WHEAT_BLOCK, 0), "Wheat Block", BlockBreakInfo::instant()));
 
 		$planksBreakInfo = new BlockBreakInfo(2.0, BlockToolType::AXE, 0, 15.0);
-		$leavesBreakInfo = new BlockBreakInfo(0.2, BlockToolType::SHEARS);
+		$leavesBreakInfo = new class(0.2, BlockToolType::HOE) extends BlockBreakInfo{
+				public function getBreakTime(Item $item) : float{
+					return match($item->getBlockToolType()){
+						BlockToolType::SWORD => 0.2,
```
dktapps: This is actually the same as what's already returned for swords.

src/block/BlockFactory.php
leolee3914: oic
`//swords break any block 1.5x faster than hand`

src/block/BlockFactory.php
leolee3914: oic
`//swords break any block 1.5x faster than hand`

PR #4731, opened 2022-01-16
Implemented Composter

NTT1906:
## Introduction
 - Implemented Composter
### API changes
 - Add composter in registry

### Behavioural changes
N/A

## Backwards compatibility
N/A

## Tests
[Link](https://youtu.be/E79IYBNDpx8)


NTT1906:
finally bruh :'C

src/block/Composter.php
```diff
@@ -0,0 +1,190 @@
...
+	public const CROP_GROWTH_EMITTER_PARTICLE = "minecraft:crop_growth_emitter";
+
+	protected int $composter_fill_level = 0;
+
+	protected function writeStateToMeta() : int{
+		return $this->composter_fill_level;
+	}
+
+	public function writeStateToItemMeta() : int{
+		return $this->composter_fill_level;
```
dktapps, line 56: This function is redundant. The item is not supposed to retain the fill level.

src/block/Composter.php
```diff
@@ -0,0 +1,190 @@
...
+use pocketmine\world\sound\ComposterFillSound;
+use pocketmine\world\sound\ComposterFillSuccessSound;
+use pocketmine\world\sound\ComposterReadySound;
+use function abs;
+use function mt_rand;
+
+class Composter extends Transparent{
+	public const CROP_GROWTH_EMITTER_PARTICLE = "minecraft:crop_growth_emitter";
+
+	protected int $composter_fill_level = 0;
```
dktapps, line 49: This doesn't comply with the project coding style.

Also, `composter` is redundant; it's in the `Composter` class context, the fact that it's a "composter" fill level is already implied.

src/block/Composter.php
```diff
@@ -0,0 +1,190 @@
...
+	protected function recalculateCollisionBoxes() : array{
+		$boxes = [$this->getSideCollisionBox(Facing::DOWN)];
+		foreach (Facing::HORIZONTAL as $side) {
+			$boxes[] = $this->getSideCollisionBox($side);
+		}
+		return $boxes;
+	}
+
+	protected function getSideCollisionBox(int $face = Facing::NORTH) : AxisAlignedBB{
+		$empty = abs(15 - 2 * $this->composter_fill_level) - (int) ($this->composter_fill_level === 0);
```
dktapps, line 76: I have no idea what's going on here.

src/block/Composter.php
```diff
@@ -0,0 +1,190 @@
...
+		return $this->composter_fill_level === 8;
+	}
+
+	public function getComposterFillLevel() : int{
+		return $this->composter_fill_level;
+	}
+
+	/**
+	 * @throws \Exception
+	 */
```
dktapps, line 94: This docblock is redundant.

src/block/Composter.php
```diff
@@ -0,0 +1,190 @@
...
+			) as $entity) {
+			if ($entity instanceof Player || $entity instanceof Projectile) continue;
+			$motion = $entity->getMotion();
+			$motion->y = 0.2; //Lower can make entity clip through the block, the game cannot catch up the speed of the entity? :/
+			$entity->setMotion($motion);
+		}
+	}
+
+	protected function spawnParticleEffect() : void{
+		$packet = SpawnParticleEffectPacket::create(
```
dktapps, line 122: That's not how this is done. We don't use packets in core code.

src/block/Composter.php
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	/**
+	 * @throws \Exception
+	 */
+	public function compost(Block|Player $origin, ?Item $item = null) : bool{
+		if ($this->composter_fill_level >= 8) {
+			$this->position->getWorld()->dropItem(
+				$this->position->add(0.5, 0.85, 0.5),
+				new Fertilizer(new ItemIdentifier(ItemIds::DYE, 15), "Bone Meal"),
```
dktapps, line 140:
```suggestion
				VanillaItems::BONE_MEAL(),
```

src/block/Composter.php
```diff
@@ -0,0 +1,190 @@
...
+				new Vector3(0, 0, 0)
+			);
+			$this->position->getWorld()->addSound($this->position, new ComposterEmptySound());
+			$this->spawnParticleEffect();
+			$this->composter_fill_level = 0;
+			$this->position->getWorld()->setBlock($this->position, $this);
+			return false;
+		} else {
+			if ($item === null || !ComposterUtils::isCompostable($item)) return false;
```
dktapps, line 149: Unbraced conditions are not allowed.

src/block/Composter.php
```diff
@@ -0,0 +1,190 @@
...
+				$this->position->add(0.5, 0.85, 0.5),
+				new Fertilizer(new ItemIdentifier(ItemIds::DYE, 15), "Bone Meal"),
+				new Vector3(0, 0, 0)
+			);
+			$this->position->getWorld()->addSound($this->position, new ComposterEmptySound());
+			$this->spawnParticleEffect();
+			$this->composter_fill_level = 0;
+			$this->position->getWorld()->setBlock($this->position, $this);
+			return false;
+		} else {
```
dktapps, line 148: This `else` is redundant. You're returning in the previous case.

src/block/Composter.php
```diff
@@ -0,0 +1,190 @@
...
+			if ($item === null || !ComposterUtils::isCompostable($item)) return false;
+			$this->spawnParticleEffect();
+
+			$percent = ComposterUtils::getPercentage($item);
+			if (mt_rand(1, 100) <= $percent) {
+				$this->pushCollidedEntities();
```
dktapps, line 154: This is not something that should be done at a block-specific level (it's generally not implemented yet).

src/block/Composter.php
```diff
@@ -0,0 +1,190 @@
...
+			}
+			$this->position->getWorld()->setBlock($this->position, $this);
+			return true;
+		}
+	}
+
+	public function getDrops(Item $item) : array{
+		return $this->composter_fill_level === 8 ? [
+			VanillaBlocks::COMPOSTER()->asItem(),
+			new Fertilizer(new ItemIdentifier(ItemIds::DYE, 15), "Bone Meal")
```
dktapps, line 173:
```suggestion
			VanillaItems::BONE_MEAL()
```

src/block/Composter.php
```diff
@@ -0,0 +1,190 @@
...
+		);
+		foreach ($this->position->getWorld()->getViewersForPosition($this->position) as $player) {
+			$player->getNetworkSession()->sendDataPacket($packet);
+		}
+	}
+
+	/**
+	 * @throws \Exception
+	 */
+	public function compost(Block|Player $origin, ?Item $item = null) : bool{
```
dktapps, line 136: The `$origin` parameter appears to be redundant. Also, the item should not be nullable.

src/block/Composter.php
```diff
@@ -0,0 +1,190 @@
...
+
+	protected function recalculateCollisionBoxes() : array{
+		$boxes = [$this->getSideCollisionBox(Facing::DOWN)];
+		foreach (Facing::HORIZONTAL as $side) {
+			$boxes[] = $this->getSideCollisionBox($side);
+		}
+		return $boxes;
+	}
+
+	protected function getSideCollisionBox(int $face = Facing::NORTH) : AxisAlignedBB{
```
dktapps, line 75: Default parameter values should be avoided. They will only enable mistakes in this case.

src/block/Composter.php
```diff
@@ -0,0 +1,190 @@
...
+		$empty = abs(15 - 2 * $this->composter_fill_level) - (int) ($this->composter_fill_level === 0);
+		return ($face === Facing::DOWN || $face === Facing::UP) ? AxisAlignedBB::one()->contract(2 / 16, 0, 2 / 16)->trim(Facing::UP, $empty / 16) : AxisAlignedBB::one()->trim(Facing::opposite($face), 14 / 16);
```
dktapps, line 77: This would benefit from newlines. It's difficult to read.

src/block/Composter.php
```diff
@@ -0,0 +1,190 @@
...
+
+	public function isEmpty() : bool{
+		return $this->composter_fill_level === 0;
+	}
+
+	public function isReady() : bool{
+		return $this->composter_fill_level === 8;
+	}
+
+	public function getComposterFillLevel() : int{
```
dktapps, line 88: This should be accompanied by a fluent setter similar to other blocks.

src/block/utils/ComposterUtils.php
```diff
@@ -0,0 +1,261 @@
...
+		self::register(new ItemBlock(new ItemIdentifier(ItemIds::RED_MUSHROOM_BLOCK, 12), VanillaBlocks::RED_MUSHROOM_BLOCK()), 85);
+		self::register(new ItemBlock(new ItemIdentifier(ItemIds::RED_MUSHROOM_BLOCK, 13), VanillaBlocks::RED_MUSHROOM_BLOCK()), 85);
+		self::register(new ItemBlock(new ItemIdentifier(ItemIds::RED_MUSHROOM_BLOCK, 14), VanillaBlocks::RED_MUSHROOM_BLOCK()), 85);
+
+		//TODO: add Nether Wart Block, Warped Wart Blocks
+
+		// region 100% percentage compost
+		self::register(new Item(new ItemIdentifier(ItemIds::CAKE, 0), "Cake"), 100);
+		self::register(new Item(new ItemIdentifier(ItemIds::PUMPKIN_PIE, 0), "Pumpkin Pie"), 100);
+	}
```
dktapps, line 231: You're not supposed to use `new Item`. Use `VanillaBlocks` and `VanillaItems`.

src/block/utils/ComposterUtils.php
```diff
@@ -0,0 +1,261 @@
...
+
+	public static function register(Item $item, int $percentage = 30, bool $overwrite = false) : bool{
```
dktapps, line 233: Don't provide percentage default. This is just creating an avenue for bugs.

src/block/utils/ComposterUtils.php
```diff
@@ -0,0 +1,261 @@
...
+use pocketmine\item\MelonSeeds;
+use pocketmine\item\Potato;
+use pocketmine\item\PumpkinSeeds;
+use pocketmine\item\SweetBerries;
+use pocketmine\item\WheatSeeds;
+use function count;
+
+class ComposterUtils {
+	/**
+	 * @var mixed[]
```
dktapps, line 49: This should not be `mixed[]`.

src/block/utils/ComposterUtils.php
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	public static function getPercentage(Item $item) : int{
+		if (count(self::$list) === 0) {
+			self::registryDefault();
+		}
+		return self::$list[self::getListOffset($item->getId(), $item->getMeta())] ?? 0;
+	}
+
+	public static function getListOffset(int $id, int $variant = 0) : int{
```
dktapps, line 255: This should not be public.

src/block/utils/ComposterUtils.php
```diff
@@ -0,0 +1,261 @@
...
+use pocketmine\item\ItemIds;
+use pocketmine\item\Melon;
+use pocketmine\item\MelonSeeds;
+use pocketmine\item\Potato;
+use pocketmine\item\PumpkinSeeds;
+use pocketmine\item\SweetBerries;
+use pocketmine\item\WheatSeeds;
+use function count;
+
+class ComposterUtils {
```
dktapps, line 47: `ComposterUtils` is too vague. The name should be more specific to what the class actually does.

In addition, this should use `SingletonTrait` instead of being all static.

dktapps (changes requested):
This PR needs a lot of work. I've left some initial review comments but there will be more.

Govdim:
What is the next-minor branch for?

src/block/Composter.php
NTT1906: basically it calculates each side collision boxes, so i was thinking of making 4 walls each side and one layer at the bottom which will rise up when the composter get composted

src/block/Composter.php
NTT1906: > That's not how this is done. We don't use packets in core code.

You can't do it without using packet. Spawn particle separately is okay but it's not the same in vanilla

src/block/utils/ComposterUtils.php
NTT1906: > You're not supposed to use `new Item`. Use `VanillaBlocks` and `VanillaItems`.

yeah, i'm sorry

ghost:
> What is the next-minor branch for?

Branch for next minor version 🤷‍♂️.
For example 4.1.0 is next minor version.

Read https://en.m.wikipedia.org/wiki/Software_versioning (Sequence-based identifiers)

src/block/Composter.php
JavierLeon9966: Changing the `abs` to `max(0, …)` would be a better way

src/block/Composter.php
```diff
@@ -0,0 +1,190 @@
...
+		return $this->composter_fill_level;
+	}
+
+	/**
+	 * @throws \Exception
+	 */
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
+		if ($player instanceof Player && !$player->isSneaking()) {
```
JavierLeon9966: Checking if the player is sneaking is redundant 
```suggestion
		if ($player instanceof Player) {
```

JavierLeon9966 (changes requested):

src/block/Composter.php
NTT1906: so i should remove it?

NTT1906: you mean that `pushCollidedEntities `

NTT1906:
Should I make it into a JSON file and save it at [Bedrock Data](https://github.com/pmmp/BedrockData), when the server is loaded it will load the file the same way as [make](https://github.com/pmmp/PocketMine-MP/blob/9f4fcfafdb11141a8bd063c716df120e33ac8ee0/src/crafting/CraftingManagerFromDataHelper.php#L35)?

src/block/Composter.php
dktapps: > > That's not how this is done. We don't use packets in core code.
> 
> You can't do it without using packet. Spawn particle separately is okay but it's not the same in vanilla

Obviously, packet is required. My point is that core code is not supposed to use them directly. This should be done via a Particle class.

src/block/Composter.php
dktapps: > basically it calculates each side collision boxes, so i was thinking of making 4 walls each side and one layer at the bottom which will rise up when the composter get composted

You missed my point entirely. The code should be self explanatory and it's not because you tried too hard to be clever.

src/block/Composter.php
dktapps: Yes

NTT1906:
I messed up everything...

NTT1906:
I messed up this so can you mark this as invalid?

dktapps:
Why is this closed ...

NTT1906:
Is the first part of [4740](https://github.com/pmmp/PocketMine-MP/pull/4740)

NTT1906:
> Why is this closed ...

i messed up because the base branch changed

PR #4732, opened 2022-01-16
Implemented StoneCutter Block

ipad54:
## Introduction
Implemented StoneCutter Block

### API changes
Added $blockType property to ShapelessRecipe
Added getBlockType() method to ShapelessRecipe

### Behavioural changes
N/A

## Follow-up
N/A

## Tests
https://youtu.be/HNuWtTHMf3s


ghost:
If some player breaks stonecutter?
Items will disappear forever. Yes? I'm sure of it

dktapps:
Actually it won't: https://github.com/pmmp/PocketMine-MP/pull/4732/files#diff-adf08eca6b5de3fb8b72b57f0c8223b70b29f5b1fd8314619282f7e9f8c7b5f5R30

ghost:
> Actually it won't: https://github.com/pmmp/PocketMine-MP/pull/4732/files#diff-adf08eca6b5de3fb8b72b57f0c8223b70b29f5b1fd8314619282f7e9f8c7b5f5R30

What this means?

src/block/StoneCutterBlock.php
```diff
@@ -0,0 +1,62 @@
...
+use pocketmine\block\utils\BlockDataSerializer;
+use pocketmine\block\utils\FacesOppositePlacingPlayerTrait;
+use pocketmine\block\utils\HorizontalFacingTrait;
+use pocketmine\item\Item;
+use pocketmine\math\AxisAlignedBB;
+use pocketmine\math\Facing;
+use pocketmine\math\Vector3;
+use pocketmine\player\Player;
+
+class StoneCutterBlock extends Transparent{
```
dktapps: This should simply be `Stonecutter`.

src/block/inventory/StoneCutterInventory.php
```diff
@@ -0,0 +1,40 @@
...
+
+use pocketmine\inventory\SimpleInventory;
+use pocketmine\inventory\TemporaryInventory;
+use pocketmine\world\Position;
+
+class StoneCutterInventory extends SimpleInventory implements BlockInventory, TemporaryInventory{
+	use BlockInventoryTrait;
+
+	public const SLOT_INPUT = 0;
+	public const SLOT_OUTPUT = 1;
```
dktapps: The output isn't a real slot, same as for crafting.

src/crafting/CraftingManagerFromDataHelper.php
```diff
@@ -43,13 +44,18 @@ public static function make(string $filePath) : CraftingManager{
 		$itemDeserializerFunc = \Closure::fromCallable([Item::class, 'jsonDeserialize']);
 
 		foreach($recipes["shapeless"] as $recipe){
-			if($recipe["block"] !== "crafting_table"){ //TODO: filter others out for now to avoid breaking economics
+			$blockType = match ($recipe["block"]){
```
dktapps: This leaks protocol info into the API. An enum should be added for this similar to FurnaceType.

src/block/inventory/StoneCutterInventory.php
```diff
@@ -0,0 +1,40 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\block\inventory;
+
+use pocketmine\inventory\SimpleInventory;
+use pocketmine\inventory\TemporaryInventory;
+use pocketmine\world\Position;
+
+class StoneCutterInventory extends SimpleInventory implements BlockInventory, TemporaryInventory{
```
dktapps: This should be `StonecutterInventory`.

src/block/VanillaBlocks.php
```diff
@@ -1098,6 +1099,7 @@ protected static function setup() : void{
 		self::register("stone_pressure_plate", $factory->get(70, 0));
 		self::register("stone_slab", $factory->get(421, 2));
 		self::register("stone_stairs", $factory->get(435, 0));
+		self::register("stonecutter_block", $factory->get(452, 0));
```
dktapps: `_block` is redundant. We know it's a block already or we wouldn't be using `VanillaBlocks`.

src/block/BlockFactory.php
```diff
@@ -388,6 +388,7 @@ public function isAffectedBySilkTouch() : bool{
 		$this->registerAllMeta(new Stair(new BID(Ids::MOSSY_STONE_BRICK_STAIRS, 0), "Mossy Stone Brick Stairs", $stoneBreakInfo));
 		$this->registerAllMeta(new StoneButton(new BID(Ids::STONE_BUTTON, 0), "Stone Button", new BlockBreakInfo(0.5, BlockToolType::PICKAXE)));
 		$this->registerAllMeta(new StonePressurePlate(new BID(Ids::STONE_PRESSURE_PLATE, 0), "Stone Pressure Plate", new BlockBreakInfo(0.5, BlockToolType::PICKAXE, ToolTier::WOOD()->getHarvestLevel())));
+		$this->registerAllMeta(new StoneCutterBlock(new BID(Ids::STONECUTTER_BLOCK, 0, ItemIds::STONECUTTER_BLOCK), "Stonecutter Block", new BlockBreakInfo(3.5, BlockToolType::PICKAXE)));
```
dktapps: `Block` suffix is unnecessary.

dktapps (changes requested):

src/block/BlockFactory.php
dktapps: This is also in the wrong place. Please learn how alphabetical order works.

dktapps:
This is also missing from StringToItemParser.

src/network/mcpe/cache/CraftingDataCache.php
```diff
@@ -88,7 +88,7 @@ private function buildCraftingDataCache(CraftingManager $manager) : CraftingData
 						return $converter->coreItemStackToNet($item);
 					}, $recipe->getResults()),
 					$nullUUID,
-					$recipe->getBlockType(),
+					$recipe->getBlockType()->name(),
```
dktapps: This is not guaranteed to match the network name.

dktapps:
~IIRC, there is a sound when crafting stuff using a stonecutter. It's missing from the PR.~ Never mind, seems like it's client-side.

dktapps:
The block doesn't face in the correct direction when placed facing east or west. You need to use `readHorizontalFacing` instead of `readLegacyHorizontalFacing`, and a state bitmask of `0b111` instead of `0b11`.

dktapps:
> > Actually it won't: https://github.com/pmmp/PocketMine-MP/pull/4732/files#diff-adf08eca6b5de3fb8b72b57f0c8223b70b29f5b1fd8314619282f7e9f8c7b5f5R30
> 
> What this means?

It means that [this person](https://github.com/pmmp/PocketMine-MP/pull/4732#issuecomment-1013970137) has no idea what they were talking about.

dktapps:
Apart from a couple of bugs and minor issues, this PR basically "just works", so it's in pretty good shape already. There's some discussion to be had about API design, but the PR is overall looking OK.

src/block/Stonecutter.php
```diff
@@ -0,0 +1,62 @@
...
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
+		if($player !== null){
+			$player->setCurrentWindow(new StonecutterInventory($this->position));
+		}
+		return true;
+	}
+
+	protected function recalculateCollisionBoxes() : array{
+		return [AxisAlignedBB::one()->trim(Facing::UP, 0.5)];
```
dktapps: This is incorrect. The height is actually 9/16, so it should be trimmed by 7/16.

dktapps:
CS checks are failing because there should be an empty line at the end of every file. You can either fix it manually or run php-cs-fixer locally to fix it.

src/item/StringToItemParser.php
```diff
@@ -870,6 +870,7 @@ private static function make() : self{
 		$result->registerBlock("stone_wall", fn() => VanillaBlocks::COBBLESTONE_WALL());
 		$result->registerBlock("stonebrick", fn() => VanillaBlocks::STONE_BRICKS());
 		$result->registerBlock("stonecutter", fn() => VanillaBlocks::LEGACY_STONECUTTER());
```
dktapps: I think we can change this to use the new stonecutter as well. Might be worth adding a `legacy_stonecutter` alias for the old one, though.

src/crafting/CraftingBlockType.php
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * This doc-block is generated automatically, do not modify it manually.
+ * This must be regenerated whenever registry members are added, removed or changed.
+ * @see build/generate-registry-annotations.php
+ * @generate-registry-docblock
+ *
+ * @method static CraftingBlockType CRAFTING_TABLE()
+ * @method static CraftingBlockType STONECUTTER()
+ */
+final class CraftingBlockType{
```
dktapps: I don't really think `block type` is a suitable expression for this. Most of these recipes work on the 2x2 crafting grid, which isn't a block at all.

dktapps: I think `ShapelessRecipeType` would probably suit better. Its members would be `CRAFTING` and `STONECUTTER`.

src/crafting/ShapelessRecipe.php
```diff
@@ -32,12 +32,14 @@ class ShapelessRecipe implements CraftingRecipe{
...
 	/** @var Item[] */
 	private $results;
+	private ShapelessRecipeType $recipeType;
 
 	/**
 	 * @param Item[] $ingredients No more than 9 total. This applies to sum of item stack counts, not count of array.
 	 * @param Item[] $results List of result items created by this recipe.
 	 */
-	public function __construct(array $ingredients, array $results){
+	public function __construct(array $ingredients, array $results, ?ShapelessRecipeType $recipeType = null){
```
dktapps: Please put a TODO in the doc comment here -  we'll want to make the `recipeType` parameter mandatory in PM5.

Also, I think we can refer to it as just `type`, given the context.

src/crafting/ShapelessRecipe.php
```diff
@@ -63,6 +65,10 @@ public function getResultsFor(CraftingGrid $grid) : array{
 		return $this->getResults();
 	}
 
+	public function getRecipeType() : ShapelessRecipeType{
+		return $this->recipeType;
```
dktapps: This could simply be `type` rather than `recipeType`. Similar goes for the method name.

dktapps:
For posterity I'll mention it here: there's currently no verification for whether the crafted recipe was actually crafted using a stonecutter. In addition, this allows non-stonecutter recipes to be used on a stonecutter.

That said, this problem also exists elsewhere with big/small crafting recipes lacking verification, since it only cares if you have the right number of ingredients.

This isn't a terribly harmful or useful exploit, but it may allow you to obtain some materials that you otherwise wouldn't be able to in early game.

src/crafting/CraftingManagerFromDataHelper.php
```diff
@@ -43,12 +43,19 @@ public static function make(string $filePath) : CraftingManager{
 		$itemDeserializerFunc = \Closure::fromCallable([Item::class, 'jsonDeserialize']);
 
 		foreach($recipes["shapeless"] as $recipe){
-			if($recipe["block"] !== "crafting_table"){ //TODO: filter others out for now to avoid breaking economics
+			$recipeType = match ($recipe["block"]){
```
dktapps:
```suggestion
			$recipeType = match($recipe["block"]){
```

PR #4733, opened 2022-01-17
do we need this again?

mukeenanyafiq:
do we need this file back again?


mukeenanyafiq:
if not needed, please tell the same file like this

ColinHDev:
`PluginIdentifiableCommand` was replaced with https://github.com/pmmp/PocketMine-MP/blob/stable/src/plugin/PluginOwned.php

PR #4740, opened 2022-01-18
Implemented composter

NTT1906:
## Introduction
Add composter. Is a copy of [4731](https://github.com/pmmp/PocketMine-MP/pull/4731) 

### Relevant issues
N/A

## Changes
- Add composter recipes and some minor changes to Crafting systems.

### API changes
- [CraftingManagerFromDataHelper::make()](https://github.com/pmmp/PocketMine-MP/blob/59de045ecbacfb9acc297e75170819727d68ae09/src/crafting/CraftingManagerFromDataHelper.php#L35)' first parameter is now an array.

## Backwards compatibility
No

## Follow-up
- Add a new "composts.json" file to [Bedrock Data](https://github.com/pmmp/BedrockData)

## Tests
None

Attach scripts or actions to test this pull request, as well as the result
Soon

PR #4742, opened 2022-01-19
Implemented Composter

NTT1906:
## Introduction
Add composter, again... This is a copy of [4740](https://github.com/pmmp/PocketMine-MP/pull/4740)

### Relevant issues
N/A

## Changes
- Add composter
- ?
## Tests
[Video](https://youtu.be/p0NnhyktQH0)


NTT1906:
It seems to have been overshadowed after 3 pull requests on the same topic...

dktapps:
Stop being needy. There is not enough reviewer manpower to go around and I have much more important things to worry about right now. I'll review it when I have the available time.

src/block/Composter.php
```diff
@@ -0,0 +1,162 @@
...
+use function max;
+use function mt_rand;
+
+class Composter extends Transparent{
+	public const READY = 8;
+
+	public const MIN_LEVEL = 0;
+	public const MAX_LEVEL = 7;
+
+	protected int $fill_level = self::MIN_LEVEL;
```
ShockedPlot7560: Use camel case instead

src/block/Composter.php
```diff
@@ -0,0 +1,162 @@
...
+use pocketmine\world\sound\ComposterFillSuccessSound;
+use pocketmine\world\sound\ComposterReadySound;
+use function max;
+use function mt_rand;
+
+class Composter extends Transparent{
+	public const READY = 8;
+
+	public const MIN_LEVEL = 0;
+	public const MAX_LEVEL = 7;
```
ShockedPlot7560: This isnt the max level

src/block/utils/CompostFactory.php
```diff
@@ -0,0 +1,173 @@
...
+declare(strict_types=1);
+
+namespace pocketmine\block\utils;
+
+use pocketmine\block\VanillaBlocks;
+use pocketmine\item\Item;
+use pocketmine\item\VanillaItems;
+use pocketmine\utils\SingletonTrait;
+
+final class CompostFactory{
```
ShockedPlot7560, line 31: It would be better if it were placed directly in the items as for the burning time.

src/block/VanillaBlocks.php
```diff
@@ -828,8 +829,10 @@ public function getBreakTime(Item $item) : float{
 		self::register("mossy_cobblestone_stairs", new Stair(new BID(Ids::MOSSY_COBBLESTONE_STAIRS), "Mossy Cobblestone Stairs", $cobblestoneBreakInfo));
 
 		self::register("cobweb", new Cobweb(new BID(Ids::COBWEB), "Cobweb", new Info(new BreakInfo(4.0, ToolType::SWORD | ToolType::SHEARS, 1))));
+		self::register("composter", new Composter(new BID(Ids::COMPOSTER), "Composter", new Info(BreakInfo::axe(2))));
 		self::register("cocoa_pod", new CocoaBlock(new BID(Ids::COCOA_POD), "Cocoa Block", new Info(BreakInfo::axe(0.2, null, 15.0))));
 		self::register("coral_block", new CoralBlock(new BID(Ids::CORAL_BLOCK), "Coral Block", new Info(BreakInfo::pickaxe(7.0, ToolTier::WOOD()))));
+
```
ShockedPlot7560: Unrelated change

src/block/Composter.php
```diff
@@ -0,0 +1,162 @@
...
+	public function setFillLevel(int $fill_level) : self{
+		if($fill_level < 0 || $fill_level > self::READY){
+			throw new \InvalidArgumentException("Fill level  must be in range 0 ... " . self::READY);
+		}
+		$this->fill_level = $fill_level;
+		return $this;
+	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if (($player instanceof Player) && $this->compost(clone $item)) {
```
ShockedPlot7560: Why cloning the item ?

ShockedPlot7560: .+ prefer checking if player is null instead of using instance checking

src/block/Composter.php
```diff
@@ -0,0 +1,162 @@
...
+			$item->pop();
+		}
+		return true;
+	}
+
+	public function onScheduledUpdate() : void{
+		if ($this->fill_level === self::MAX_LEVEL) {
```
ShockedPlot7560: Dont use another constant for that

src/block/Composter.php
```diff
@@ -0,0 +1,162 @@
...
+			$this->position->getWorld()->addSound($this->position, new ComposterFillSuccessSound());
+			++$this->fill_level;
+			if ($this->fill_level === self::MAX_LEVEL) {
+				$this->position->getWorld()->scheduleDelayedBlockUpdate($this->position, 20);
+			}
+		} else {
+			$this->position->getWorld()->addSound($this->position, new ComposterFillSound());
+		}
+
+		$this->position->getWorld()->setBlock($this->position, $this);
```
ShockedPlot7560: You can moove it into the if because there is not new changes for the else.

src/block/Composter.php
```diff
@@ -0,0 +1,162 @@
...
+
+	public function onScheduledUpdate() : void{
+		if ($this->fill_level === self::MAX_LEVEL) {
+			$this->position->getWorld()->addSound($this->position, new ComposterReadySound());
+			$this->fill_level = self::READY;
+			$this->position->getWorld()->setBlock($this->position, $this);
+		}
+	}
+
+	public function compost(Item $item) : bool{
```
ShockedPlot7560: I think that this function need to contain only the compost logic and no all the conditions which can test elsewhere

src/block/Composter.php
```diff
@@ -0,0 +1,162 @@
...
+			}
+		} else {
+			$this->position->getWorld()->addSound($this->position, new ComposterFillSound());
+		}
+
+		$this->position->getWorld()->setBlock($this->position, $this);
+		return true;
+	}
+
+	public function empty(bool $hasDrop = false) : void{
```
ShockedPlot7560:
```suggestion
	public function empty(bool $withDrop = false) : void{
```

ShockedPlot7560 (changes requested):

src/block/Composter.php
NTT1906: > This isnt the max level

I should rename this to layer

src/block/utils/CompostFactory.php
NTT1906: > It would be better if it were placed directly in the items as for the burning time.

That would take a lot of time to do. Personally, I think it would be really hard to manage with

src/block/Composter.php
NTT1906: > I should rename this to layer

nvm...

src/block/Composter.php
NTT1906: > Dont use another constant for that
Can I know the reason behind this?

src/block/Composter.php
NTT1906: > This isnt the max level
It was based on Java edition... couldn't come up with a better idea :l....

src/block/Composter.php
```diff
@@ -0,0 +1,164 @@
...
+	public function setFillLevel(int $fillLevel) : self{
+		if($fillLevel < 0 || $fillLevel > self::MAX_LEVEL){
+			throw new \InvalidArgumentException("Layers must be in range " . self::MIN_LEVEL . " ... " . self::MAX_LEVEL);
+		}
+		$this->fillLevel = $fillLevel;
+		return $this;
+	}
+
+	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
+		if ($this->fillLevel === 7) {
```
ShockedPlot7560: I don't think hardcoding this value looks good. A function or constant can be better.

src/block/Composter.php
```diff
@@ -0,0 +1,164 @@
...
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if ($player !== null && $this->addItem($item)) {
+			$item->pop();
+		}
+		return true;
+	}
+
+	public function onScheduledUpdate() : void{
+		if ($this->fillLevel === 7) {
```
ShockedPlot7560: Same here

src/block/Composter.php
```diff
@@ -0,0 +1,164 @@
...
+			$this->position->getWorld()->addSound($this->position, new ComposterReadySound());
+			$this->fillLevel = self::MAX_LEVEL;
+			$this->position->getWorld()->setBlock($this->position, $this);
+		}
+	}
+
+	public function addItem(Item $item) : bool{
+		if ($this->fillLevel >= self::MAX_LEVEL) {
```
ShockedPlot7560:
```suggestion
		if ($this->isReady()) {
```

src/block/Composter.php
```diff
@@ -0,0 +1,164 @@
...
+			$this->empty(true);
+			return false;
+		}
+		if ($this->fillLevel === 7 || !CompostFactory::getInstance()->isCompostable($item)) {
```
ShockedPlot7560: Same here for the hardcoded value.

src/block/Composter.php
```diff
@@ -0,0 +1,164 @@
...
+			return false;
+		}
+
+		$this->position->getWorld()->addParticle($this->position->add(0.5, 0.5, 0.5), new CropGrowthEmitterParticle());
+		if (mt_rand(1, 100) <= CompostFactory::getInstance()->getPercentage($item)) {
+			$this->position->getWorld()->addSound($this->position, new ComposterFillSuccessSound());
+			if (++$this->fillLevel === 7) {
```
ShockedPlot7560: Same here

src/block/Composter.php
```diff
@@ -0,0 +1,164 @@
...
+				$this->position->getWorld()->scheduleDelayedBlockUpdate($this->position, 20);
+			}
+			$this->position->getWorld()->setBlock($this->position, $this);
```
ShockedPlot7560: Seems better to set and update after

src/block/Composter.php
```diff
@@ -0,0 +1,164 @@
...
+		} else {
+			$this->position->getWorld()->addSound($this->position, new ComposterFillSound());
+		}
+		return true;
+	}
+
+	public function empty(bool $withDrop = false) : void{
+		if ($withDrop && $this->fillLevel === self::MAX_LEVEL) {
```
ShockedPlot7560:
```suggestion
		if ($withDrop && $this->isReady()) {
```

src/block/Composter.php
```diff
@@ -0,0 +1,164 @@
...
+			$this->position->getWorld()->addParticle($this->position, new CropGrowthEmitterParticle());
+		}
+
+		$this->position->getWorld()->addSound($this->position, new ComposterEmptySound());
+		$this->fillLevel = self::MIN_LEVEL;
+		$this->position->getWorld()->setBlock($this->position, $this);
+	}
+
+	public function getDropsForCompatibleTool(Item $item) : array{
+		return $this->fillLevel === self::MAX_LEVEL ? [
```
ShockedPlot7560:
```suggestion
		return $this->isReady() ? [
```

src/block/Composter.php
NTT1906: > I don't think hardcoding this value looks good. A function or constant can be better.

I couldn't come up with any good names...
const START_CHANGE_LEVEL?

NTT1906: > I don't think hardcoding this value looks good. A function or constant can be better.

```
Mojang:
Eventually, the materials will start to break down into a dark, crumbly substance. This is the "mature" stage, and it's full of nutrients that your plants will love! 
```
The final state is `mature` so PRE_MATURE?

NTT1906: some other names are `immature`, `intermediate`, `unfinished `, `young compost`...

NTT1906: `immature` seems good to me...

NTT1906:
@ShockedPlot7560 Should I add events?

src/block/Composter.php
ShockedPlot7560: i don't really know, but hardcoded value need to be change for sure.

ShockedPlot7560:
> @ShockedPlot7560 Should I add events?

Probably yes

src/block/Composter.php
```diff
@@ -0,0 +1,202 @@
...
+use pocketmine\world\sound\ComposterFillSound;
+use pocketmine\world\sound\ComposterFillSuccessSound;
+use pocketmine\world\sound\ComposterReadySound;
+use function max;
+use function mt_rand;
+
+class Composter extends Transparent{
+
+	public const MIN_LEVEL = 0;
+	public const IMMATURE_LEVEL = 7;
```
ShockedPlot7560: I think, FULL_LEVEL can be used and for MAX_LEVEL since it only indicates the possibility of the composter dropping a bone meal I think COLLECTING_LEVEL might be better?

src/block/Composter.php
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	public function isEmpty() : bool{
+		return $this->fillLevel === self::MIN_LEVEL;
+	}
+
+	public function isImmature() : bool{
+		return $this->fillLevel === self::IMMATURE_LEVEL;
+	}
+	public function isReady() : bool{
```
ShockedPlot7560: canCollect can be better

src/block/Composter.php
```diff
@@ -0,0 +1,202 @@
...
+			$boxes[] = AxisAlignedBB::one()->trim(Facing::opposite($side), 14 / 16);
+		}
+		return $boxes;
+	}
+
+	public function isEmpty() : bool{
+		return $this->fillLevel === self::MIN_LEVEL;
+	}
+
+	public function isImmature() : bool{
```
ShockedPlot7560: isFull can be better

src/block/Composter.php
ShockedPlot7560: Need space between each function

src/block/Composter.php
```diff
@@ -0,0 +1,202 @@
...
+		}
+		return true;
+	}
+
+	public function empty() : void{
+		$drops = [VanillaItems::BONE_MEAL()];
+		if(ComposterEmptyEvent::hasHandlers()){
+			$ev = new ComposterEmptyEvent($this, $drops);
+			$ev->call();
+			$drops = $ev->getDrops();
```
ShockedPlot7560: Can be done after checking if the event has been cancelled.

src/block/utils/CompostFactory.php
```diff
@@ -0,0 +1,187 @@
...
+	use SingletonTrait;
+
+	protected const MIN_PERCENTAGE = 0;
+	protected const MAX_PERCENTAGE = 100;
```
ShockedPlot7560: percentage is by definition necessarily from 0 to 100

src/event/block/ComposterFillEvent.php
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	public function getFillLevel() : int{
+		return $this->fillLevel;
+	}
+
+	public function isSuccess() : bool{
+		return $this->success;
+	}
+	public function setSuccess(bool $success) : void{
```
ShockedPlot7560, line 55: Need space beatween each function

src/block/utils/CompostFactory.php
NTT1906: > percentage is by definition necessarily from 0 to 100

wdym, adding some sort of validation to input percentage?

src/block/Composter.php
NTT1906: > I think, FULL_LEVEL can be used and for MAX_LEVEL since it only indicates the possibility of the composter dropping a bone meal I think COLLECTING_LEVEL might be better?

so FULL_LEVEL for 7 and COLLECTING_LEVEL for 8?

NTT1906: > > I think, FULL_LEVEL can be used and for MAX_LEVEL since it only indicates the possibility of the composter dropping a bone meal I think COLLECTING_LEVEL might be better?
> 
> so FULL_LEVEL for 7 and COLLECTING_LEVEL for 8?

if so, MIN_LEVEL should be renamed to EMPTY_LEVEL?

src/block/Composter.php
NTT1906: > canCollect can be better

isCollectable?

src/block/Composter.php
ShockedPlot7560: Can be better probably yes.

src/block/Composter.php
ShockedPlot7560: Maybe yes

src/block/utils/CompostFactory.php
ShockedPlot7560: The input percentage must be validated, but I don't think it's necessary to have a constant to delimit the percentage, which is by definition from 0 to 100.

src/event/block/ComposterMatureEvent.php
```diff
@@ -0,0 +1,31 @@
...
+ *
+ *
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\event\block;
+
+/**
+ * Called when composter start to mature, e.g. if the fill lever is 7
```
ShockedPlot7560: This may be clearer I think
```suggestion
 * Called when the state of the composter allows recovery of the bone meal. e.g. the state changes from FULL_LEVEL to COLLECTING_LEVEL
```

src/block/utils/CompostFactory.php
```diff
@@ -0,0 +1,187 @@
...
+			return true;
+		}
+		return false;
+	}
+
+	public function isCompostable(Item $item) : bool{
+		return !$item->isNull() && isset($this->list[$item->getTypeId()]);
+	}
+
+	public function getPercentage(Item $item) : int{
```
ShockedPlot7560: Might be better to return null instead of 0 when the percentage doesn't exists.

src/block/Composter.php
```diff
@@ -0,0 +1,203 @@
...
+		if(ComposterFillEvent::hasHandlers()){
+			$ev = new ComposterFillEvent($this, $item, $this->fillLevel, $success);
+			$ev->call();
+			if($ev->isCancelled()){
+				return false;
+			}
+			$success = $ev->isSuccess();
+		}
+		if($success){
+			++$this->fillLevel;
```
ShockedPlot7560:
```suggestion
			$this->fillLevel++;
```

src/block/Composter.php
```diff
@@ -0,0 +1,203 @@
...
+	public function setFillLevel(int $fillLevel) : self{
+		if($fillLevel < 0 || $fillLevel > self::COLLECTING_LEVEL){
+			throw new \InvalidArgumentException("Layers must be in range " . self::EMPTY_LEVEL . " ... " . self::COLLECTING_LEVEL);
+		}
+		$this->fillLevel = $fillLevel;
+		return $this;
+	}
+
+	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
+		if($this->isFull()){
```
ShockedPlot7560, line 94: How can it be full when it's about to be placed?

src/block/utils/CompostFactory.php
NTT1906: okay

src/block/Composter.php
NTT1906: > How can it be full when it's about to be placed?

In vanilla, if you place a composter with a level of 7 (Full)... It then got scheduled to make collectable composter

NTT1906:
@ShockedPlot7560 I don't know what I was thinking before deciding to use SingletonTrait, should CompostFactory be shifted to static instead of Singleton?

src/block/utils/CompostFactory.php
NTT1906: > Might be better to return null instead of 0 when the percentage doesn't exists.
 
I think, returning 0 can also mean that the item does not have the chance to raise the level of compost.

The role of checking percentage doesn't exists can be similar to the `isCompostable` function with a few modifications.

ShockedPlot7560:
Singleton is ok

src/block/utils/CompostFactory.php
ShockedPlot7560: Three choices: 
- We return null because the item is not present, so I wouldn't be shocked if the value didn't exist?
- Return 0, but then it would have to be documented?
- Keep it like that because the logic is ok?

src/block/Composter.php
ShockedPlot7560: I was originally wondering how to place a block that already had a level filled, but then I remembered the click wheel.

NTT1906: > I was originally wondering how to place a block that already had a level filled, but then I remembered the click wheel.

might I ask what is this "click wheel"?

ShockedPlot7560: I called it that, but it's the mouse wheel, when you click on it.

NTT1906 (commented):
Done?
<sub> This comment was for closing the review :p </sub>

<sub><sup>Edit: it's not working :c</sup></sub>

dktapps:
Internals team discussion had a consensus that we weren't happy with the code in this PR, mainly around the mess of CompostFactory. It's poorly named & the data could be associated better in other ways. This was previously mentioned in review comments but the discussion was incorrectly marked as resolved.

PR #4747, opened 2022-01-20
Update furnace

### LINKED ISSUE #4363: Furnaces don't make any sound
### Issue description
In vanilla, furnaces (including blast furnaces and smokers) make faint crackling sounds which are completely absent in PM.

1. Smelt something in a furnace
2. Turn your volume up to 100%
3. Notice there's no sound from the furnace.

### OS and versions
[18:54:23.675] [Server thread/INFO]: This server is running PocketMine-MP
[18:54:23.675] [Server thread/INFO]: Server version: 4.0.0+dev (git hash: 9cc798f87283ed8d0376cd6eaf12d5b4f6e7fe6d-dirty)
[18:54:23.675] [Server thread/INFO]: Compatible Minecraft version: 1.17.10 (protocol version: 448)
[18:54:23.676] [Server thread/INFO]: PHP version: 8.0.9
[18:54:23.676] [Server thread/INFO]: PHP JIT: disabled
[18:54:23.676] [Server thread/INFO]: Operating system: win

* Game version: irrelevant

ShockedPlot7560:
## Introduction
A todo had been present for a long time in the block furnace, requiring to move the behavior of the tile to the block.
This pull request is also intended to make the furnace more vanilla

### Relevant issues
* Fixes #4363 
### API changes
* Add sound classes
* remove all unused function in Furnace tile

### Behavioural changes
* furnace work now in block and not tile

## Follow-up
Rename in LevelSoundEvent
* BLOCK_FURNACE_LIT to BLOCK_FURNACE_FIRE_CRACKLE

## Tests
Youtube test: https://youtu.be/r22mtA7DBxg

### DIFF
--- a/src/block/Furnace.php
+++ b/src/block/Furnace.php
@@ -173,9 +173,9 @@ class Furnace extends Opaque{
 	}
 
 	public function onScheduledUpdate() : void{
-		$furnace = $this->position->getWorld()->getTile($this->position);
-		if($furnace instanceof TileFurnace && !$furnace->isClosed()){
-			$inventory = $furnace->getInventory();
+		$tile = $this->position->getWorld()->getTile($this->position);
+		if($tile instanceof TileFurnace && !$tile->isClosed()){
+			$inventory = $tile->getInventory();
 
 			$prevCookTime = $this->cookTime;
 			$prevRemainingFuelTime = $this->remainingFuelTime;
@@ -185,7 +185,7 @@ class Furnace extends Opaque{
 			$raw = $inventory->getSmelting();
 			$product = $inventory->getResult();
 
-			$furnaceType = $furnace->getFurnaceType();
+			$furnaceType = $tile->getFurnaceType();
 			$smelt = $this->position->getWorld()->getServer()->getCraftingManager()->getFurnaceRecipeManager($furnaceType)->match($raw);
 			$canSmelt = ($smelt instanceof FurnaceRecipe and $raw->getCount() > 0 and (($smelt->getResult()->equals($product) and $product->getCount() < $product->getMaxStackSize()) or $product->isNull()));
 
@@ -202,7 +202,7 @@ class Furnace extends Opaque{
 					if($this->cookTime >= $furnaceType->getCookDurationTicks()){
 						$product = $smelt->getResult()->setCount($product->getCount() + 1);
 
-						$ev = new FurnaceSmeltEvent($furnace, $raw, $product);
+						$ev = new FurnaceSmeltEvent($tile, $raw, $product);
 						$ev->call();
 
 						if(!$ev->isCancelled()){
--- a/src/block/Furnace.php
+++ b/src/block/Furnace.php
@@ -96,7 +96,9 @@ class Furnace extends Opaque{
 		return $this->lit;
 	}
 
-	/** @return $this */
+	/**
+	 * @return $this
+	 */
 	public function setLit(bool $lit = true) : self{
 		$this->lit = $lit;
 		return $this;
--- a/src/block/Furnace.php
+++ b/src/block/Furnace.php
@@ -23,7 +23,9 @@ declare(strict_types=1);
 
 namespace pocketmine\block;
 
+use pocketmine\block\tile\BlastFurnace;
 use pocketmine\block\tile\Furnace as TileFurnace;
+use pocketmine\block\tile\Smoker;
 use pocketmine\block\utils\FacesOppositePlacingPlayerTrait;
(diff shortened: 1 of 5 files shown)


src/block/Furnace.php
```diff
@@ -39,6 +51,12 @@ class Furnace extends Opaque{
 	protected BlockIdentifierFlattened $idInfoFlattened;
 
 	protected bool $lit = false; //this is set based on the blockID
+	/** @var int */
+	private $remainingFuelTime = 0;
+	/** @var int */
+	private $cookTime = 0;
+	/** @var int */
+	private $maxFuelTime = 0;
```
ColinHDev: Couldn't those be assigned a type directly instead of using doc comments?

src/block/Furnace.php
```diff
@@ -62,14 +99,42 @@ public function isLit() : bool{
 		return $this->lit;
 	}
 
-	/**
-	 * @return $this
-	 */
+	/** @return $this */
```
ColinHDev: I see the reason for the PHPDoc, to indicate that it's a fluent setter, but why are these comments shrunk down to one line?

src/block/Furnace.php
```diff
@@ -81,10 +146,121 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+				}
+			}
+		}
+	}
+
+	public function fireCrackle() : void{
+		$tile = $this->position->getWorld()->getTile($this->position);
+		if($tile instanceof BlastFurnace){
+			$this->position->getWorld()->addSound($this->position, new BlastFurnaceFireCrackleSound);
+		}elseif ($tile instanceof Smoker) {
```
ColinHDev: Doesn't this violate PMMP's code style?
```suggestion
		}elseif($tile instanceof Smoker){
```

src/block/Furnace.php
```diff
@@ -81,10 +146,121 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+						$this->cookTime -= $furnaceType->getCookDurationTicks();
+					}
+					$this->position->getWorld()->setBlock($this->position, $this);
+				}elseif($this->remainingFuelTime <= 0){
+					$this->onStopSmelting();
+				}else{
+					$this->cookTime = 0;
+					$this->position->getWorld()->setBlock($this->position, $this);
+				}
+				if(mt_rand(1, 100) < 5){
```
ColinHDev, line 229: Where did you get these values from?
I currently haven't looked into IDA, but there is no information about the probability in the wiki. 
(I did look into Java and it is a probability of 10% there instead of 4%.)
```java
if (random.nextDouble() < 0.1D) {...}
```

ColinHDev: Also, in Java the position of the sound would be:
```php
new Vector3(
    $this->position->x + 0.5,
    $this->position->y,
    $this->position->z + 0.5
);
```
But that also would need to be checked with IDA. Just wanted to mention, that those values in the PR could not be true.

src/block/Furnace.php
ShockedPlot7560: To follow the return format of the other setter in the class

src/block/Furnace.php
ShockedPlot7560: > Where did you get these values from? I currently haven't looked into IDA, but there is no information about the probability in the wiki. (I did look into Java and it is a probability of 10% there instead of 4%.)
> 
> ```java
> if (random.nextDouble() < 0.1D) {...}
> ```

I have actually set 4% for the random sound because after many tests, it is the one that seemed closer to vanilla. No precise official value is put on the wiki, we only note "Plays randomly when active". 10% seems a lot, the sound is present all the time, while it is not necessarily present all the time in vanilla

src/block/Furnace.php
ShockedPlot7560: Indeed

src/block/Furnace.php
ColinHDev: Yes, but why shortened after all?
I don't think that's ever done in pm

dktapps: Not true, but this shouldn't be changing code that doesn't have anything to do with the PR objective.

dktapps:
> Fixes Furnaces don't make any sound #4363

This should have been done separately. Now I can't take the fix without reviewing the rest of the PR.

dktapps:
This PR is doing too many things at once and the diff is huge so I can't see wtf is going on. Please split it up and submit separate changes.

ShockedPlot7560:
ok good

PR #4749, opened 2022-01-21
Implement PlayerViewDistanceChangeEvent

### LINKED ISSUE #4550: Add PlayerViewDistanceChangeEvent
## Description
Add an event that's fired when a player requests a different view distance.

## Justification
I recently implemented a cool tool for simulating chunk sending to players, as seen here: https://twitter.com/dktapps/status/1456397007946461190?s=20

which allowed me to debug and test a fix for a teleporting bug in PM4.

I'd like to publish the code that generates the logs for this as a plugin.

However, it's not possible to implement this right now without hacks in the PM core, since the simulation has to be redone if the player's view distance changes (e.g. if it's larger than previous, a larger frame must be used).

## Alternative methods
I can listen to packets (sucks) or I can .. i don't know ... spam Player->getViewDistance() ?

ShockedPlot7560:
## Introduction
There is currently no simple way to recover a player's view distance change

### Relevant issues
* Fixes #4550 

### API changes
* Add PlayerViewDistanceChangeEvent

### Behavioural changes
When a player requests a new distance, if it is different from the player's current distance, the event will be called.
The event will not be called when the player connecting because no view distance is established

## Tests
Make a simple plugin of this type: 
```php

	}

	public function onViewDistance(PlayerViewDistanceChangeEvent $event){
		$this->getServer()->broadcastMessage($event->getPlayer()->getName() . " request a different radius, " . $event->getNewDistance() . " instead of " . $event->getOldDistance());
	}
}
```
Go to your settings and change your rendering distance

### DIFF
--- a/src/event/player/PlayerViewDistanceChangeEvent.php
+++ b/src/event/player/PlayerViewDistanceChangeEvent.php
@@ -1,0 +1,50 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * @author PocketMine Team
+ * @link http://www.pocketmine.net/
+ *
+ *
+*/
+
+declare(strict_types=1);
+
+namespace pocketmine\event\player;
+
+use pocketmine\event\Cancellable;
+use pocketmine\event\CancellableTrait;
+use pocketmine\player\Player;
+
+/**
+ * Called when a player requests a different viewing distance than the current one.
+ */
+class PlayerViewDistanceChangeEvent extends PlayerEvent implements Cancellable{
+	use CancellableTrait;
+
+	protected int $radius;
+
+	public function __construct(Player $player, int $radius){
+		$this->player = $player;
+		$this->radius = $radius;
+	}
+
+	public function setRadius(int $radius) : void{
+		$this->radius = $radius;
+	}
+
+	public function getRadius() : int{
+		return $this->radius;
+	}
+}
--- a/src/event/player/PlayerViewDistanceChangeEvent.php
+++ b/src/event/player/PlayerViewDistanceChangeEvent.php
@@ -29,17 +29,17 @@ use pocketmine\player\Player;
  * Called when a player requests a different viewing distance than the current one.
  */
 class PlayerViewDistanceChangeEvent extends PlayerEvent{
-	protected int $radius;
+	protected int $newRadius;
 	protected int $oldRadius;
 
-	public function __construct(Player $player, int $radius, int $oldRadius){
+	public function __construct(Player $player, int $newRadius, int $oldRadius){
 		$this->player = $player;
-		$this->radius = $radius;
+		$this->newRadius = $newRadius;
 		$this->oldRadius = $oldRadius;
 	}
 
-	public function getRadius() : int{
-		return $this->radius;
+	public function getNewRadius() : int{
+		return $this->newRadius;
 	}
 
 	public function getOldRadius() : int{
--- a/src/event/player/PlayerViewDistanceChangeEvent.php
+++ b/src/event/player/PlayerViewDistanceChangeEvent.php
@@ -32,10 +32,10 @@ class PlayerViewDistanceChangeEvent extends PlayerEvent{
 	protected int $newRadius;
 	protected int $oldRadius;
 
-	public function __construct(Player $player, int $newRadius, int $oldRadius){
+	public function __construct(Player $player, int $oldRadius, int $newRadius){
 		$this->player = $player;
-		$this->newRadius = $newRadius;
 		$this->oldRadius = $oldRadius;
+		$this->newRadius = $newRadius;
 	}
 
 	public function getNewRadius() : int{
--- a/src/event/player/PlayerViewDistanceChangeEvent.php
+++ b/src/event/player/PlayerViewDistanceChangeEvent.php
@@ -39,15 +39,15 @@ class PlayerViewDistanceChangeEvent extends PlayerEvent{
 	}
 
 	/**
-	 * Returns an int corresponding to the number of chunks forming the new view distance
+	 * Returns the new view radius, measured in chunks.
 	 */
 	public function getNewDistance() : int{
 		return $this->newDistance;
 	}
 
 	/**
-	 * Returns an int corresponding to the number of chunks forming the previous view distance, before change
-	 * If the player connects to the server, it will return -1
+	 * Returns the old view radius, measured in chunks.
+	 * A value of -1 means that the player has just connected and did not have a view distance before this event.
 	 */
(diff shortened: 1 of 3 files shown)


src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -685,6 +686,13 @@ public function handleMapInfoRequest(MapInfoRequestPacket $packet) : bool{
 	}
 
 	public function handleRequestChunkRadius(RequestChunkRadiusPacket $packet) : bool{
+		if($this->player->getViewDistance() !== $packet->radius){
+			$ev = new PlayerViewDistanceChangeEvent($this->player, $packet->radius);
```
dktapps: This is the wrong place. It won't be called if it's changed for reasons other than packet.

src/event/player/PlayerViewDistanceChangeEvent.php
```diff
@@ -0,0 +1,50 @@
...
+
+	protected int $radius;
+
+	public function __construct(Player $player, int $radius){
+		$this->player = $player;
+		$this->radius = $radius;
+	}
+
+	public function setRadius(int $radius) : void{
+		$this->radius = $radius;
```
dktapps: This should have bounds checks.

dktapps: I'm also doubtful this is a good idea.

src/event/player/PlayerViewDistanceChangeEvent.php
```diff
@@ -0,0 +1,50 @@
...
+
+use pocketmine\event\Cancellable;
+use pocketmine\event\CancellableTrait;
+use pocketmine\player\Player;
+
+/**
+ * Called when a player requests a different viewing distance than the current one.
+ */
+class PlayerViewDistanceChangeEvent extends PlayerEvent implements Cancellable{
+	use CancellableTrait;
```
dktapps: I'm doubtful that this should be cancellable. It will cause bugs on the client unless synced back.

dktapps (changes requested):

src/network/mcpe/handler/InGamePacketHandler.php
ShockedPlot7560: So in the Player class directly? This will group together all the way to change its view distance

src/event/player/PlayerViewDistanceChangeEvent.php
ShockedPlot7560: Indeed, when the player connects, if a plugin cancels the event, it will be blocked

JavierLeon9966: What if the player changes the view distance while in-game?

ShockedPlot7560: If the player is already logged in, this was not going to take their request. However, making this event cancellable is not the right solution because it can block some behavior.

src/event/player/PlayerViewDistanceChangeEvent.php
```diff
@@ -0,0 +1,48 @@
...
+	protected int $oldRadius;
+
+	public function __construct(Player $player, int $radius, int $oldRadius){
+		$this->player = $player;
+		$this->radius = $radius;
+		$this->oldRadius = $oldRadius;
+	}
+
+	public function getRadius() : int{
+		return $this->radius;
```
dktapps: This should probably be renamed to `getNewRadius()` since we're having a `getOldRadius()` here.

ShockedPlot7560: For sure, it's better indeed

src/player/Player.php
```diff
@@ -485,6 +487,11 @@ public function setViewDistance(int $distance) : void{
 
 		$this->getNetworkSession()->syncViewAreaRadius($this->viewDistance);
 
+		if($this->viewDistance !== $oldViewDistance){
+			$ev = new PlayerViewDistanceChangeEvent($this, $this->viewDistance, $oldViewDistance);
+			$ev->call();
+		}
```
dktapps: I think this should be called before the view distance change, rather than after. If we decide to make it cancellable later on, we don't want to subtly break plugin assumptions about `Player->getViewDistance()` during the event.

src/event/player/PlayerViewDistanceChangeEvent.php
```diff
@@ -0,0 +1,48 @@
...
+use pocketmine\player\Player;
+
+/**
+ * Called when a player requests a different viewing distance than the current one.
+ */
+class PlayerViewDistanceChangeEvent extends PlayerEvent{
+	protected int $newRadius;
+	protected int $oldRadius;
+
+	public function __construct(Player $player, int $newRadius, int $oldRadius){
```
dktapps: Weird parameter order?

dktapps:
I think this should use the `distance` terminology consistently, rather than `radius`.

src/event/player/PlayerViewDistanceChangeEvent.php
```diff
@@ -0,0 +1,48 @@
...
+	protected int $newRadius;
+	protected int $oldRadius;
+
+	public function __construct(Player $player, int $oldRadius, int $newRadius){
+		$this->player = $player;
+		$this->oldRadius = $oldRadius;
+		$this->newRadius = $newRadius;
+	}
+
+	public function getNewRadius() : int{
```
dktapps: Documentation should be added to indicate that this number is in chunks.

src/event/player/PlayerViewDistanceChangeEvent.php
```diff
@@ -0,0 +1,48 @@
...
+		return $this->newRadius;
+	}
+
+	public function getOldRadius() : int{
```
dktapps: Same again here.

src/event/player/PlayerViewDistanceChangeEvent.php
```diff
@@ -29,20 +29,27 @@
...
-		$this->oldRadius = $oldRadius;
-		$this->newRadius = $newRadius;
+		$this->oldDistance = $oldDistance;
+		$this->newDistance = $newDistance;
 	}
 
-	public function getNewRadius() : int{
-		return $this->newRadius;
+	/**
+	 * Returns an int corresponding to the number of chunks forming the new view distance
```
dktapps: This is super misleading.

src/event/player/PlayerViewDistanceChangeEvent.php
```diff
@@ -29,20 +29,27 @@
...
+	 */
+	public function getNewDistance() : int{
+		return $this->newDistance;
 	}
 
-	public function getOldRadius() : int{
-		return $this->oldRadius;
+	/**
+	 * Returns an int corresponding to the number of chunks forming the previous view distance, before change
+	 * If the player connects to the server, it will return -1
```
dktapps:
```suggestion
	 * A value of -1 means that the player has just connected and did not have a view distance before this event.
```

src/event/player/PlayerViewDistanceChangeEvent.php
dktapps:
```suggestion
	 * Returns the new view radius, measured in chunks.
```

src/event/player/PlayerViewDistanceChangeEvent.php
```diff
@@ -29,20 +29,27 @@
...
+	 * Returns an int corresponding to the number of chunks forming the new view distance
+	 */
+	public function getNewDistance() : int{
+		return $this->newDistance;
 	}
 
-	public function getOldRadius() : int{
-		return $this->oldRadius;
+	/**
+	 * Returns an int corresponding to the number of chunks forming the previous view distance, before change
```
dktapps:
```suggestion
	 * Returns the old view radius, measured in chunks.
```

src/event/player/PlayerViewDistanceChangeEvent.php
ShockedPlot7560: As this is very clear, I have not been able to express it more clearly.

dktapps: "if the player connects" is unclear. Every player connects to the server ... does that mean it will always return -1? No.

dktapps (approved):

PR #4755, opened 2022-01-23
Implement furnace sound

ShockedPlot7560:
## Introduction
Furnace simply do not produce sound

### Relevant issues
* Fixes: #4363 

### API changes
* Add classes related to the 3 sounds of the 3 types of furnace

## Follow-up
* Rename ``BLOCK_FURNACE_LIT`` to ``BLOCK_FURNACE_FIRE_CRACKLE`` in ``LevelSoundEvent``

## Tests
https://youtu.be/G5K2JezTzFA

### DIFF
--- a/src/block/tile/Furnace.php
+++ b/src/block/tile/Furnace.php
@@ -36,9 +36,13 @@ use pocketmine\math\Vector3;
 use pocketmine\nbt\tag\CompoundTag;
 use pocketmine\network\mcpe\protocol\ContainerSetDataPacket;
 use pocketmine\player\Player;
+use pocketmine\world\sound\BlastFurnaceFireCrackleSound;
+use pocketmine\world\sound\FurnaceLitSound;
+use pocketmine\world\sound\SmokerSmokeSound;
 use pocketmine\world\World;
 use function array_map;
 use function max;
+use function mt_rand;
 
 abstract class Furnace extends Spawnable implements Container, Nameable{
 	use NameableTrait;
@@ -207,6 +211,9 @@ abstract class Furnace extends Spawnable implements Container, Nameable{
 				$this->cookTime = 0;
 			}
 			$ret = true;
+			if(mt_rand(1, 100) < 5){
+				$this->fireCrackle();
+			}
 		}else{
 			$this->onStopSmelting();
 			$this->remainingFuelTime = $this->cookTime = $this->maxFuelTime = 0;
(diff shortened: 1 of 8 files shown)


src/block/tile/Furnace.php
```diff
@@ -232,4 +239,15 @@ public function onUpdate() : bool{
 
 		return $ret;
 	}
+
+	public function fireCrackle() : void{
+		$tile = $this->position->getWorld()->getTile($this->position);
```
dktapps: This is ridiculous.

dktapps (changes requested):
These names are absurdly inconsistent.

src/block/tile/Furnace.php
```diff
@@ -212,7 +212,13 @@ public function onUpdate() : bool{
 			}
 			$ret = true;
 			if(mt_rand(1, 100) < 5){
-				$this->fireCrackle();
+				if($this instanceof BlastFurnace){
```
dktapps: If you ever find yourself doing `this instanceof`, you're doing something crazy that defeats the point of OOP.

also, why is this in the tile...

ShockedPlot7560: Indeed ...
However, the fact that this is in the tile is that all the behavior of the current furnaces are in it. Should all instances of furnace be required to specify the sound they play when in action ?

dktapps: Putting it in the tile just means it will have to be moved again later, which wastes everyone's time. It should be done in Furnace->onScheduledUpdate() instead.

Also, you should be able to use FurnaceType instead of looking at the tile class.

src/world/sound/BlastFurnaceFireCrackleSound.php
```diff
@@ -0,0 +1,35 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\world\sound;
+
+use pocketmine\math\Vector3;
+use pocketmine\network\mcpe\protocol\LevelSoundEventPacket;
+use pocketmine\network\mcpe\protocol\types\LevelSoundEvent;
+
+final class BlastFurnaceFireCrackleSound implements Sound{
```
dktapps: Just because the Mojang names are absurdly inconsistent doesn't mean PM should be too. This can simply be named `BlastFurnaceLitSound` or even just `BlastFurnaceSound`. Same for the others.

src/world/sound/SmokerSmokeSound.php
```diff
@@ -0,0 +1,35 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\world\sound;
+
+use pocketmine\math\Vector3;
+use pocketmine\network\mcpe\protocol\LevelSoundEventPacket;
+use pocketmine\network\mcpe\protocol\types\LevelSoundEvent;
+
+final class SmokerSmokeSound implements Sound{
```
dktapps: See above.

dktapps:
> Rename BLOCK_FURNACE_LIT to BLOCK_FURNACE_FIRE_CRACKLE in LevelSoundEvent

The names are auto-generated.

dktapps (changes requested):

src/world/sound/BlastFurnaceFireCrackleSound.php
ShockedPlot7560: ``BlastFurnaceSound`` may be clearer than ``BlastFurnaceLitSound``, the ``FurnaceLitSound`` should be renamed to ``FurnaceSound``. There is, for the moment, only one sound produced by the furnace, not requiring to specify the fact that they are active.

PR #4759, opened 2022-01-23
Fixed incorrect drops for Cobweb

rasu3n:
## Introduction

https://minecraft.fandom.com/wiki/Cobweb#Breaking

### DIFF
--- a/src/block/Cobweb.php
+++ b/src/block/Cobweb.php
@@ -39,7 +39,7 @@ class Cobweb extends Flowable{
 	}
 
 	public function getDropsForCompatibleTool(Item $item) : array{
-		if($item->getBlockToolType() === BlockToolType::SHEARS){
+		if(($item->getBlockToolType() & BlockToolType::SHEARS) !== 0){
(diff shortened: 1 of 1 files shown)


src/block/Cobweb.php
```diff
@@ -39,6 +39,9 @@ public function onEntityInside(Entity $entity) : bool{
 	}
 
 	public function getDropsForCompatibleTool(Item $item) : array{
+		if($item->getBlockToolType() === BlockToolType::SHEARS){
```
dktapps:
```suggestion
		if(($item->getBlockToolType() & BlockToolType::SHEARS) !== 0){
```
Items may be multiple block tool types at the same time (e.g. plugins may implement supertools which can break any block).

rasu3n:
https://github.com/pmmp/PocketMine-MP/pull/4759/commits/39d518d31d22eaa38415fffd76182912c1a50b18
I'm sorry I couldn't commit to the proposal. Thank you very much!

PR #4762, opened 2022-01-23
Fix torch placement

ShockedPlot7560:
## Introduction
The current version of PocketMine-MP does not allow to place the torches like Vanilla, limiting the placement only on solid blocks or barriers / walls. However, minecraft vanilla allows placement on slabs in high position / stair on top position, etc etc

### Relevant issues
* Linked #458 

### Behavioural changes
A more complete check is made on the torch support

## Follow-up
Keep up to date with the implementation of new blocks such as the composter or other unimplemented vanilla blocks

## Tests
https://user-images.githubusercontent.com/66992287/150692172-736af462-cff5-451e-b991-4bb7cb820030.mp4

### DIFF
--- a/src/block/Torch.php
+++ b/src/block/Torch.php
@@ -99,35 +99,40 @@ class Torch extends Flowable{
 
 	private function isValidSupport(Block $block, int $face) : bool{
 		//TODO: side of composter
-		if($block instanceof Stair){
-			if($face === Facing::UP && $block->isUpsideDown()){
+		if($block instanceof Stair && (
+			$face === Facing::UP && $block->isUpsideDown() ||
+			$face === $block->getFacing()
+		)){
 				return true;
-			}elseif($face === $block->getFacing()){
-				return true;
-			}
-		}elseif($block instanceof Slab){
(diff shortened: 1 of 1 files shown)


Govdim:
Torches cannot be placed on trapdoors. Check it out in single player.

ShockedPlot7560:
> Torches cannot be placed on trapdoors. Check it out in single player.

indeed, I thought I had succeeded, so much for me

dktapps:
I'm not a fan of this solution. It feels too much like a special-case hack.

What we really need is some way for blocks to indicate which of their faces are solid and/or can provide support to other blocks.

ShockedPlot7560:
> I'm not a fan of this solution. It feels too much like a special-case hack.
> 
> What we really need is some way for blocks to indicate which of their faces are solid and/or can provide support to other blocks.

I totally agree on this point, I have thought of many other solutions. However, there are blocks that will have a solid face in one case, but in other cases, they will not accept the block.

dktapps:
Closing in favour of #4886

PR #4768, opened 2022-01-24
Items that have contents now drops them on destroy

JavierLeon9966:
## Introduction

Currently when you destroy an item with contents (ex: shulker box) it doesn’t drop them.
### API changes

Added the following methods in `Item`:
- `Item->getContainedItems()`: Returns the contents of such item, ex: shulker boxes, chests, brewing stands, hoppers, etc.
- `Item->setContainedItems()`: Sets the contents of the item.

Added the following method in `ContainerTrait`:
- `ContainerTrait->copyContentsFromItem()`: Copied the citem contents into the block entity's inventory.

Added the following helper class:
- `pocketmine\block\tile\util\ContainerHelper`

Added the following trait:
- `pocketmine\block\util\ContainerTrait`
### Behavioural changes

Any item that has contents in it will now drop them once destroyed. This can include bundles once implemented.
## Tests

https://github.com/pmmp/PocketMine-MP/assets/58715544/6d547d91-aad7-4bcd-b8e7-d6cbed46249b


ghost:
Items should also scatter in different directions.  Is it provided?

dktapps:
That's a topic for a different issue. Player deaths should also have the same behaviour but don't.

ghost:
> That's a topic for a different issue. Player deaths should also have the same behaviour but don't.

Will it be in the next next-minor?

NTT1906:
I thought that the bundle was canceled?

NTT1906:
oh, it's not😆

dktapps (changes requested):
I think it would be better to have some kind of API like `Item::getContainedItems()` instead of making bogus assumptions about NBT.

JavierLeon9966:
> I think it would be better to have some kind of API like `Item::getContainedItems()` instead of making bogus assumptions about NBT.

Yeah, especially when the bundle gets implemented. It could’ve a different NBT compared to chests/shulker boxes.
Now this would require item classes for both `Chest` and `ShulkerBox` items or some trait like `ContainerItemTrait` that has this NBT deserializer, and a serializer to let plugins add items with `Item::setContainedItems()`

dktapps:
> > I think it would be better to have some kind of API like `Item::getContainedItems()` instead of making bogus assumptions about NBT.
> 
> Yeah, especially when the bundle gets implemented. It could’ve a different NBT compared to chests/shulker boxes. Now this would require item classes for both `Chest` and `ShulkerBox` items or some trait like `ContainerItemTrait` that has this NBT deserializer, and a serializer to let plugins add items with `Item::setContainedItems()`

I think a general implementation in `Item` would be sufficient. I'm just wary of assuming internal item NBT structure from outside and not being able to override it.

ShockedPlot7560:
Need to be update with the upstream

src/item/Item.php
```diff
@@ -338,6 +357,15 @@ protected function deserializeCompoundTag(CompoundTag $tag) : void{
...
 		$this->keepOnDeath = $tag->getByte(self::TAG_KEEP_ON_DEATH, 0) !== 0;
+
+		$this->containedItems = [];
+		$containedItems = $tag->getListTag(Container::TAG_ITEMS);
+		if($containedItems !== null && $containedItems->getTagType() === NBT::TAG_Compound){
+			/** @var CompoundTag $itemNBT */
+			foreach($containedItems as $itemNBT){
+				$this->containedItems[] = Item::nbtDeserialize($itemNBT);
+			}
+		}
```
ShockedPlot7560: I'm not a big fan of this, as its a unique code from container. Wouldnt be possible to move into a child class or somewhere ?

JavierLeon9966: Moved the duplicate code to a new `ContainerHelper` class

src/block/tile/util/ContainerHelper.php
```diff
@@ -0,0 +1,89 @@
...
+		if(($inventoryTag = $tag->getTag(Container::TAG_ITEMS)) instanceof ListTag && $inventoryTag->getTagType() === NBT::TAG_Compound){
+			$contents = [];
+			/** @var CompoundTag $itemNBT */
+			foreach($inventoryTag as $slot => $itemNBT){
+				try{
+					$count = $itemNBT->getByte(SavedItemStackData::TAG_COUNT);
+					if($count === 0){
+						continue;
+					}
+					$contents[$slot] = Item::nbtDeserialize($itemNBT);
```
ipad54: IIRC, this will break loading container data from vanilla worlds because the save format in Chiseled Bookshelves and generic containers is a little bit different. Chiseled Bookshelves have their slot stored as a key in inventory compound tag (how it’s implemented here), but generic containers save their slot in `SavedItemStackData::TAG_SLOT` byte tag. See https://github.com/pmmp/PocketMine-MP/blob/stable/src/block/tile/ContainerTrait.php#L55

src/block/tile/util/ContainerHelper.php
```diff
@@ -0,0 +1,89 @@
...
+		$items = [];
+		foreach($contents as $slot => $item){
+			if($includeEmpty && $item->isNull()){
+				$items[$slot] = CompoundTag::create()
+					->setByte(SavedItemStackData::TAG_COUNT, 0)
+					->setShort(SavedItemData::TAG_DAMAGE, 0)
+					->setString(SavedItemData::TAG_NAME, "")
+					->setByte(SavedItemStackData::TAG_WAS_PICKED_UP, 0);
+			}else{
+				$items[$slot] = $item->nbtSerialize();
```
ipad54: Same here. See https://github.com/pmmp/PocketMine-MP/blob/stable/src/block/tile/ContainerTrait.php#L75

src/block/tile/util/ContainerHelper.php
JavierLeon9966: I didn't notice that detail. Thanks for pointing it out. I'll just make chiseled bookshelves have their own code

JavierLeon9966:
Found 3 issues that are outside the scope of this PR:
- When placing chiseled bookshelves with books in it, it doesn't render on the block
- When placing a brewing stand with blaze powder in it, the fuel time renders full until rendering the block again
- `NamingTrait::copyDataFromItem` may need to be renamed to `NamingTrait::copyNameFromItem` to reduce duplicate code

src/item/Item.php
```diff
@@ -238,6 +246,23 @@ public function setKeepOnDeath(bool $keepOnDeath) : void{
...
+		return $this->containedItems;
+	}
+
+	/**
+	 * @param Item[] $items
+	 * @phpstan-param array<int, Item> $items
+	 */
+	public function setContainedItems(array $items) : void{
+		Utils::validateArrayValueType($items, static function(Item $_) : void{});
+		$this->containedItems = $items;
```
dktapps: This should probably clone the items to prevent surprises

dktapps (commented):
Looks OK apart from a couple of minor nits

dktapps:
I don't like that `Item[]` is used inside `Item`, since it is essentially an inventory at that point, but without the ability to manipulate contents. I'm still on the fence about the best way to implement this.

PR #4774, opened 2022-01-26
Whitelist command now kicks non-whitelisted online players

### LINKED ISSUE #3868: /whitelist on should kick non-whitelisted online players
When the server whitelist is enabled, people who got into the server in time should be kicked. It does not seem reasonable to want to enable whitelist without kicking everyone already online not in the whitelist.

This changes behaviour and may be appropriate in major.next.

JavierLeon9966:
## Introduction

Currently when you remove someone from whitelist or enable whitelist it doesn’t kick all the non-whitelisted online players.
### Relevant issues

* Fixes #3868 
### Behavioural changes

The server now kicks non-whitelisted online players once whitelist has been enabled.
## Tests

https://user-images.githubusercontent.com/58715544/151087681-dfd922c9-613b-4fea-9c37-d1a9c34a09f4.mov

### DIFF
--- a/src/command/defaults/WhitelistCommand.php
+++ b/src/command/defaults/WhitelistCommand.php
@@ -140,7 +140,7 @@ class WhitelistCommand extends VanillaCommand{
 
 	private function kickNonWhitelistedPlayers(Server $server) : void{
 		foreach($server->getOnlinePlayers() as $player){
-			if(!$server->isWhitelisted($player)){
+			if(!$server->isWhitelisted($player->getName())){
 				$player->kick("Server whitelisted.");
 			}
 		}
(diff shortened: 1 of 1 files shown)


src/command/defaults/WhitelistCommand.php
```diff
@@ -116,7 +123,11 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
 					return true;
 				case "remove":
 					if($this->testPermission($sender, DefaultPermissionNames::COMMAND_WHITELIST_REMOVE)){
-						$sender->getServer()->removeWhitelist($args[1]);
+						$server = $sender->getServer();
+						$server->removeWhitelist($args[1]);
+						if($server->hasWhitelist()){
```
dktapps: `!isWhitelisted()` would be more appropriate here, in case some plugin alters the whitelist behaviour. We don't currently support this, but might in the future.

JavierLeon9966: Done

PR #4780, opened 2022-01-27
DisablePluginException now can be used to disable plugins

### LINKED ISSUE #2671: Introduce an explicitly supported method to disable a plugin from onEnable()
## Description
An explicit method to `throw` runtime errors from plugin enable should be introduced to allow plugins to easily disable themselves on runtime errors. This could involve allowing explicitly thrown `PluginException`s from `onEnable()`.

## Justification

Plugins are currently required to write code something like:
```php
if(!$something){
    $this->getLogger()->error("something broke");
    $this->getServer()->getPluginManager()->disablePlugin($this);
    return;
}
```

On the 3.x line, it's possible to circumvent this boilerplate by throwing an exception of any kind. This behaviour is not documented and not explicitly supported. This will not work on 4.0 (it'll crash the server). However, the exception throwing is convenient, and as such is abused by many plugins (PurePerms does this in many areas, without any care for what catches the exception).

## Alternative methods
Described above. These alternative methods are inferior to the proposed method.

JavierLeon9966:
## Introduction

Currently there is no way to explicitly disable a plugin, and the only way just causes boilerplate.
### Relevant issues

* Fixes #2671
### API changes

* `DisablePluginException` can be used to disable a plugin when thrown inside `Plugin->onEnable()`
## Tests

```php
<?php

declare(strict_types = 1);

namespace JavierLeon9966\PluginExceptionTest;

use pocketmine\plugin\DisablePluginException;

    public function onEnable(): void{
        throw new DisablePluginException("Testing!");
    }
}
```


dktapps:
PluginException is too generic. It's thrown for error cases which should crash the server, such as attempting to register an event handler with an invalid priority.

JavierLeon9966:
What about DisablePluginException? If so then what exception should it extend?

dktapps:
It doesn't really matter what it extends. RuntimeException would be fine.

NB: I already attempted to implement this once and there were issues that I don't remember. It's not as simple as it looks.

SOF3:
What is the point of *disabling* the plugin instead of crashing the server?

dktapps:
Crashing indicates a bug. A plugin may want to disable itself for non-bug reasons, such as failure to connect to a database, invalid configuration, or other reasons. None of these things are bugs in the code, so they shouldn't generate crashdumps.

SOF3:
In those cases they should still stop the server. While not generating a crashdump, it does not make sense to continue running the server with a possibly critical plugin disabled automatically.

dktapps:
> In those cases they should still stop the server. While not generating a crashdump, it does not make sense to continue running the server with a possibly critical plugin disabled automatically.

This has already been discussed multiple times and it's outside the scope of this issue.

src/plugin/PluginManager.php
```diff
@@ -434,7 +434,17 @@ public function enablePlugin(Plugin $plugin) : void{
...
 
 			$plugin->getScheduler()->setEnabled(true);
-			$plugin->onEnableStateChange(true);
+			try{
+				$plugin->onEnableStateChange(true);
+			}catch(DisablePluginException $e){
+				$errorMessage = $e->getMessage();
+				if($errorMessage !== ""){
+					$plugin->getLogger()->error($errorMessage);
+				}
```
dktapps: I think it would be better to omit this, and let the plugin itself do any error logging, so that it can localize any messages as needed. Passing a message to DisablePluginException doesn't make much sense since you can't translate.

SOF3: What prevents the plugin from providing a translated message to the exception?

dktapps: They aren't able to inject language strings into any language context that the server knows about. It's better to just let them log the way they normally would.

SOF3: I said *translated* not *translatable*

dktapps: My bad, did not see that.

PR #4801, opened 2022-02-05
Fix for #4680 - /kill not actually killing

### LINKED ISSUE #4680: /kill command doesn’t kill the player directly
### Issue description

- Expected result: The player should directly die.
- Actual result: The player takes 500 hearts of damage.

1. Give yourself max level (255) health boost or absorption with /effect
2. Execute /kill
3. Notice that you only took damage

### OS and versions

* PocketMine-MP: 4.0.0 (git hash: 468faa464b2bc5c97f23fafbb71ea61035f6f218)
* PHP: 8.0.12
* Using JIT: no 
* Server OS: linux
* Game version: iOS

### Plugins
This issue can be reproduced without plugins.

https://user-images.githubusercontent.com/58715544/147520799-5e7c24a9-c2f0-4337-88c3-31653abbd7ad.mp4

jasonw4331:
## Introduction

Fixes /kill not functioning when player had more than 1000 health

### Relevant issues

* Fixes #4680

### API changes

N/A

### Behavioral changes

/kill will now *always* kill the player

## Backwards compatibility

N/A

## Follow-up

N/A

## Tests

In-game testing with absorption 255 + regen 255 + health boost 255 + resistance 255 results in player death

### DIFF
--- a/src/command/defaults/KillCommand.php
+++ b/src/command/defaults/KillCommand.php
@@ -58,7 +58,7 @@ class KillCommand extends VanillaCommand{
 			return true;
 		}
 
-		$player->attack(new EntityDamageEvent($player, EntityDamageEvent::CAUSE_SUICIDE, 1000));
+		$player->attack(new EntityDamageEvent($player, EntityDamageEvent::CAUSE_SUICIDE, PHP_INT_MAX));
 		if($player === $sender){
 			$sender->sendMessage(KnownTranslationFactory::commands_kill_successful($sender->getName()));
 		}else{
--- a/src/command/defaults/KillCommand.php
+++ b/src/command/defaults/KillCommand.php
@@ -31,6 +31,7 @@ use pocketmine\lang\KnownTranslationFactory;
 use pocketmine\permission\DefaultPermissionNames;
 use function count;
 use function implode;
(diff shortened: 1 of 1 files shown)


NTT1906:
I don't know if pm4 is developing towards bedrock or java, or mixed... but as far as I remember, bedrock' /kill only deals massive damage and doesn't remove the entity.

jasonw4331:
I think it's best to go with what the user is expecting rather than matching either Bedrock or Java in particular. Killing the player is the expected result of using /kill, so by that logic it should be implemented this way.

dktapps:
> I think it's best to go with what the user is expecting rather than matching either Bedrock or Java in particular. Killing the player is the expected result of using /kill, so by that logic it should be implemented this way.

+1 to this. Intuitively making sense is better than bug-for-bug parity with either version.

dktapps (commented):
This is problematic since it doesn't allow plugins to intervene or intercept the event. This is a behavioural break.
That said, whether or not plugins should be able to cancel this kind of thing is up for debate.

ghost:
$player->attack(new EntityDamageEvent($player, EntityDamageEvent::CAUSE_SUICIDE, PHP_INT_MAX));

PR #4802, opened 2022-02-05
Added missing drops for Bell

NTT1906:
## Introduction
Add missing drops for Bell
### Relevant issues
None

### API changes
None
### Behavioural changes
N/A
## Backwards compatibility
N/A
## Follow-up
N/A
## Tests
None

Attach scripts or actions to test this pull request, as well as the result
N/A


jasonw4331:
Please ~~rebase onto next-minor instead of stable~~ merge into stable instead of next-minor

dktapps:
This is a bug fix, it should target stable.

dktapps:
This is actually pointless anyway, since blocks already drop themselves by default if the correct tool was used. This does nothing useful.

NTT1906:
> This is actually pointless anyway, since blocks already drop themselves by default if the correct tool was used. This does nothing useful.

but what about being destroyed by nearby update?

dktapps:
That's a general problem not specific to bells. This change causes an additional bug, which is that it drops when any tool is used (or none at all).

NTT1906:
> That's a general problem not specific to bells. This change causes an additional bug, which is that it drops when any tool is used (or none at all).

Bell got dropped by any tools and none too.

NTT1906:
https://user-images.githubusercontent.com/54394881/152657479-655694e6-8c17-4da7-a450-ad9ddc49de0f.mp4

dktapps:
Even if that were true, overriding getDrops() is not the way to deal with it - that breaks a bunch of other things.

NTT1906:
what is this whole weird-looking stuff...

jasonw4331 (approved):

ShockedPlot7560 (approved):

ShockedPlot7560:
Can be merged I think

TobiasGrether (approved):

PR #4804, opened 2022-02-06
Stop throwing exceptions on extra properties.

TheBlackPlague:
## Introduction
This code currently throws errors when properties other than the base required ones are added. This can be from resource packs created by "bridge.", where the IDE adds a "generated_with" property to the manifest. Ultimately, this leads to PM flagging it as an undefined property. Should this be set true again, it is recommended that extra work be put into ensuring that the Undefined Property flag doesn't detect extra defined properties.

### Relevant issues
- Won't load packs made via Bridge. - https://bridge-core.app/ (No issue made for this).

### API changes

There are no API changes with this PR.

### Behavioural changes

The server will behave the exact same way. The PR just removes whether the server checks for typos in extra manifest properties as mentioned by @dktapps .

## Backwards compatibility

There should be no backwards compatibility changes other than resource packs not loading (even when they're safe) before, being able to load now because they're safe to be loaded.

## Follow-up

No actions need to be done.

## Tests
This PR just changes a boolean value for exceptions. No behavior/API changes, it can be safely merged after respective reviews.

### DIFF
--- a/src/resourcepacks/ZippedResourcePack.php
+++ b/src/resourcepacks/ZippedResourcePack.php
@@ -105,7 +105,6 @@ class ZippedResourcePack implements ResourcePack{
 		}
 
 		$mapper = new \JsonMapper();
-		// $mapper->bExceptionOnUndefinedProperty = true;
 		$mapper->bExceptionOnMissingData = true;
 
 		try{
--- a/src/resourcepacks/ZippedResourcePack.php
+++ b/src/resourcepacks/ZippedResourcePack.php
@@ -105,7 +105,7 @@ class ZippedResourcePack implements ResourcePack{
 		}
 
 		$mapper = new \JsonMapper();
-		$mapper->bExceptionOnUndefinedProperty = true;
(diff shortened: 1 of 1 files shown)


src/resourcepacks/ZippedResourcePack.php
```diff
@@ -105,7 +105,7 @@ public function __construct(string $zipPath){
 		}
 
 		$mapper = new \JsonMapper();
-		$mapper->bExceptionOnUndefinedProperty = true;
+		// $mapper->bExceptionOnUndefinedProperty = true;
```
dktapps: Don't comment code. If you're going to remove it, just remove it.

TheBlackPlague: This has been solved in commit [5f495f4](https://github.com/pmmp/PocketMine-MP/pull/4804/commits/5f495f42c99b26a32a743fe3ee503f88e639c3b4).

dktapps (approved):

PR #4806, opened 2022-02-07
subchunk version 9

encyptAES:
update subchunk version

## Introduction

subchunk version 9
## Tests

tested


dktapps:
It's pointless. I intentionally didn't bother with it.

dktapps:
To clarify, the height byte is entirely ignored by the client - you can fill anything you want in there and it doesn't care, it will just use the order of the packet, exactly the same as v8 does. Therefore there is no point to using it.

encyptAES:
okay np

PR #4807, opened 2022-02-07
Allows to add arbitrary BlockInventory callbacks to ContainerOpenCallbacks in InventoryManager

PJZ9n:
## Introduction

Currently, if you create a custom inventory that extends `BlockInventory`, a ContainerOpenPacket with WindowType of　` WindowTypes::CONTAINER` will be generated unconditionally and the added callback will not be called.
The workaround is to add a custom callback at the beginning of the ObjectSet.
https://github.com/pmmp/PocketMine-MP/blob/38e34093cf63f2a70f8a76a2c6ad6f3cd13d9b53/src/network/mcpe/InventoryManager.php#L190

I'm not quite sure how this default value is used, so I may need to make further changes.

### API changes

InventoryManager: Unknown BlockInventory returns null

### Behavioural changes

[Unknown BlockInventory returns null](https://github.com/pmmp/PocketMine-MP/commit/36a434892af657abc76c69aa7300f53b73a36a28)

## Backwards compatibility

Maybe it doesn't break compatibility, but you can get and call callbacks with `InventoryManager::getContainerOpenCallbacks()`.
However, since this is a network change, it does not affect API compatibility in the first place.

## Tests

This change is minor and I don't know how to test it.

### DIFF
--- a/src/network/mcpe/InventoryManager.php
+++ b/src/network/mcpe/InventoryManager.php
@@ -187,8 +187,11 @@ class InventoryManager{
 				$inv instanceof AnvilInventory => WindowTypes::ANVIL,
 				$inv instanceof HopperInventory => WindowTypes::HOPPER,
 				$inv instanceof CraftingTableInventory => WindowTypes::WORKBENCH,
-				default => WindowTypes::CONTAINER
+				default => null
 			};
+			if($windowType === null){
+				return null;
+			}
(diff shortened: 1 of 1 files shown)


dktapps:
This breaks a ton of stuff including chests.

dktapps:
Since this would require enumerating every standard container type (chest, shulkerbox, barrel, etc) I don't think this change makes a whole lot of sense.

PR #4814, opened 2022-02-09
Fixed incorrect drops for BrewingStand

leolee3914:
## Tests

Vanilla MCBE 1.12.1: https://youtu.be/0hkaHn4ejII
Vanilla MCBE 1.18.10: https://youtu.be/uEJ3nTqo8Ws

### DIFF
--- a/src/block/BrewingStand.php
+++ b/src/block/BrewingStand.php
@@ -145,4 +145,8 @@ class BrewingStand extends Transparent{
 			}
 		}
 	}
+
+	public function getDropsForIncompatibleTool(Item $item) : array{
+		return [$this->asItem()];//always drops in BE
+	}
 }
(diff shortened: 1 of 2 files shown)


src/block/BrewingStand.php
```diff
@@ -145,4 +145,8 @@ public function onScheduledUpdate() : void{
 			}
 		}
 	}
+
+	public function getDropsForIncompatibleTool(Item $item) : array{
```
ShockedPlot7560, line 149: Can be replaced by ``getDrops`` because it will drop exactly the same no matter how it's broken

dktapps:
To be honest, I think this behaviour is a bug in Bedrock. With all the parity changes they've been making lately I won't be surprised if this behaviour quietly falls in line with Java in the future.

ShockedPlot7560:
I quite agree, given the way they're trying to align the two versions, it's likely to change in the future. This would mean that this PR and many others supposed to correct this "bug" are no longer relevant.

ShockedPlot7560:
Given that there are differences between Java and Bedrock, we can assume that this is a bug in the game. Once the two versions are in agreement, we may or may not consider making a change.

PR #4816, opened 2022-02-09
Fixed mycelium spreading

leolee3914:
## Introduction

mycelium does not spread to coarse dirt
## Tests

Vanilla (100x random tick speed, left=dirt, right=coarse dirt): https://youtu.be/7RIYRoX5xA8


ghost:
PR, which solves all the problems of mankind!  Panacea for all diseases!

dktapps:
This should actually be using the same logic as grass, but it hasn't yet been separated out for them both to use.

PR #4819, opened 2022-02-11
Implement Head Yaw.

TheBlackPlague:
## Introduction

Currently, PM doesn't really consider head-yaw to be a property. This can be tedious/nearly impossible when you want to change head-yaw without messing directly with the protocol. 

### Relevant issues

- There isn't really an issue for this.

### API changes

Location now supports a property `headYaw`. The property is strictly of type `float`. The constructor accepts this as an optional argument. All methods used to change the rotation of an entity also use the `headYaw` property if set. Furthermore, all methods mentioned above accept this as an optional argument. 

### Behavioural changes

The server should continue to behave the same way. If no `headYaw` is provided, then by default, the player's `yaw` is substituted in.

## Backwards compatibility

To preserve BC, the argument has been added in as optional to all methods. Meaning plugins or internal functions not supplying a `headYaw` argument to either the constructor or changed methods will continue to work the same way with no behavior change. If they provide the optional field, they'll be able to use it. 

A possible issue is the methods extending the `Location` class. They might need to update the relevant signature for their constructor. A possible fix for this issue is by removing it from a constructor and adding it as a function-based setter.

## Tests

This adds a property that doesn't break any internal BC. Henceforth, it doesn't necessarily need to be tested as it behaves the same way.

### DIFF
--- a/src/entity/Entity.php
+++ b/src/entity/Entity.php
@@ -1331,15 +1331,24 @@ abstract class Entity{
 		return true;
 	}
 
-	public function setRotation(float $yaw, float $pitch) : void{
+	public function setRotation(float $yaw, float $pitch, float $headYaw = -1) : void{
 		$this->location->yaw = $yaw;
 		$this->location->pitch = $pitch;
+
+		if ($headYaw < 0) {
+			$headYaw = $yaw;
+		}
+		$this->location->headYaw = $headYaw;
 		$this->scheduleUpdate();
 	}
 
-	protected function setPositionAndRotation(Vector3 $pos, float $yaw, float $pitch) : bool{
+	protected function setPositionAndRotation(Vector3 $pos, float $yaw, float $pitch, float $headYaw = -1) : bool{
 		if($this->setPosition($pos)){
-			$this->setRotation($yaw, $pitch);
+			if ($headYaw < 0) {
+				$headYaw = $yaw;
+			}
+
+			$this->setRotation($yaw, $pitch, $headYaw);
 
 			return true;
 		}
@@ -1388,9 +1397,12 @@ abstract class Entity{
 	/**
 	 * @param Vector3|Position|Location $pos
 	 */
-	public function teleport(Vector3 $pos, ?float $yaw = null, ?float $pitch = null) : bool{
+	public function teleport(Vector3 $pos, ?float $yaw = null, ?float $pitch = null, ?float $headYaw = null) : bool{
 		if($pos instanceof Location){
 			$yaw = $yaw ?? $pos->yaw;
+			/** @noinspection CallableParameterUseCaseInTypeContextInspection - This will never happen. */
+			// Just here for BC.
+			$headYaw = $headYaw ?? $pos->headYaw;
 			$pitch = $pitch ?? $pos->pitch;
 		}
 		$from = $this->location->asPosition();
@@ -1404,7 +1416,7 @@ abstract class Entity{
 		$pos = $ev->getTo();
 
 		$this->setMotion(new Vector3(0, 0, 0));
-		if($this->setPositionAndRotation($pos, $yaw ?? $this->location->yaw, $pitch ?? $this->location->pitch)){
+		if($this->setPositionAndRotation($pos, $yaw ?? $this->location->yaw, $pitch ?? $this->location->pitch, $headYaw ?? $this->location->headYaw)){
 			$this->resetFallDistance();
 			$this->setForceMovementUpdate();
 
@@ -1441,7 +1453,7 @@ abstract class Entity{
 			$this->getMotion(),
 			$this->location->pitch,
 			$this->location->yaw,
-			$this->location->yaw, //TODO: head yaw
+			$this->location->headYaw,
 			array_map(function(Attribute $attr) : NetworkAttribute{
 				return new NetworkAttribute($attr->getId(), $attr->getMinValue(), $attr->getMaxValue(), $attr->getValue(), $attr->getDefaultValue());
 			}, $this->attributeMap->getAll()),
--- a/src/entity/Location.php
+++ b/src/entity/Location.php
@@ -32,30 +32,43 @@ class Location extends Position{
 	/** @var float */
 	public $yaw;
 	/** @var float */
+	public float $headYaw;
+	/** @var float */
 	public $pitch;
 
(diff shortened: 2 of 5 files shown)


src/player/Player.php
```diff
@@ -2341,7 +2341,7 @@ public function broadcastSound(Sound $sound, ?array $targets = null) : void{
 	/**
 	 * TODO: remove this
 	 */
-	protected function sendPosition(Vector3 $pos, ?float $yaw = null, ?float $pitch = null, int $mode = MovePlayerPacket::MODE_NORMAL) : void{
+	protected function sendPosition(Vector3 $pos, ?float $yaw = null, ?float $pitch = null, int $mode = MovePlayerPacket::MODE_NORMAL, ?float $headYaw = null) : void{
 		$this->getNetworkSession()->syncMovement($pos, $yaw, $pitch, $mode);
```
JavierLeon9966, line 2345:
```suggestion
		$this->getNetworkSession()->syncMovement($pos, $yaw, $pitch, $mode, $headYaw);
```

src/player/Player.php
```diff
@@ -2350,7 +2350,7 @@ protected function sendPosition(Vector3 $pos, ?float $yaw = null, ?float $pitch
 	/**
 	 * {@inheritdoc}
 	 */
-	public function teleport(Vector3 $pos, ?float $yaw = null, ?float $pitch = null) : bool{
+	public function teleport(Vector3 $pos, ?float $yaw = null, ?float $pitch = null, ?float $headYaw = null) : bool{
 		if(parent::teleport($pos, $yaw, $pitch)){
```
JavierLeon9966, line 2354:
```suggestion
		if(parent::teleport($pos, $yaw, $pitch, $headYaw)){
```

dktapps:
I'm doubtful this should be in Location, since Location is generic for all entities, but objects such as arrows don't have head yaw.

JavierLeon9966 (changes requested):
You forgot to pass `$headYaw` in some functions so here’s a list including this one: https://github.com/pmmp/PocketMine-MP/blob/00717e15d4762ca93ad4a991062b343e2e75a14d/src/player/Player.php#L2358

TheBlackPlague:
> I'm doubtful this should be in Location, since Location is generic for all entities, but objects such as arrows don't have head yaw.

Where do you suggest it should be then? In Entity, I presume then? Or in the Human?

dktapps:
All entities descend from Entity, that's no better than Location.

I would suggest Living, but it's kinda awkwardly placed in there and there's a lot more needs to be done to properly support it (e.g. lookAt() needs to be aware of it).

TheBlackPlague:
> All entities descend from Entity, that's no better than Location.
> 
> I would suggest Living, but it's kinda awkwardly placed in there and there's a lot more needs to be done to properly support it (e.g. lookAt() needs to be aware of it).

I can implement such in Living, that shouldn't be an issue, along with making `lookAt()` aware of it. Would you be able to give me a list of methods that use the yaw in a way that headYaw would be a better use? That would speed up my development time a lot.

JavierLeon9966:
Also `getDirectionVector()`,`getHorizontalFacing()`, `sendSpawnPacket()`, `setPosition()`, `setRotation()`, `setPositionAndRotation()`, ` broadcastMovement()`, `move()`

TheBlackPlague:
Furthermore, I am also unsure why all the variables in Location/Vector3 remain public— there should be Getters and Setters for those properties. Ensuring BC becomes 10x harder if properties are publicized unnecessarily. 

A way to counter this is using the `Magic Methods` (`__set()` and `__get()`). Your thoughts @dktapps ?

dktapps:
I think it's pointless. This whole yarn about BC is nonsense. __get() and __set() exist exactly for the purpose of avoiding BC breaks. Adding them at this point serves no advantage. In addition, the fields of Location and Vector3 should not be mutable, so we don't want setters anyway. But that's outside the scope of this PR.

dktapps:
I think it's also worth mentioning that one of the reasons why Vector3 in particular continues to have public members is because it's used in hot pathways in the core code. I don't like having them public (or, more accurately, I don't like having them _mutable_), but it's currently a tradeoff that I feel ok about making.

TheBlackPlague:
After further discussion with @dktapps , it was decided that this cannot be reasonably implemented in a way that's good for the code-base & is backward compatible. What this means is that there is a good chance this will be in `5.0.0` (the next major release), but it will not be in any `4.x.x` (minor releases).

src/entity/Entity.php
```diff
@@ -1331,15 +1331,24 @@ protected function setPosition(Vector3 $pos) : bool{
...
 		$this->scheduleUpdate();
 	}
 
-	protected function setPositionAndRotation(Vector3 $pos, float $yaw, float $pitch) : bool{
+	protected function setPositionAndRotation(Vector3 $pos, float $yaw, float $pitch, float $headYaw = -1) : bool{
 		if($this->setPosition($pos)){
-			$this->setRotation($yaw, $pitch);
+			if ($headYaw < 0) {
+				$headYaw = $yaw;
+			}
```
ShockedPlot7560, line 1349: Already done in setRotation, no need to write again

src/entity/Entity.php
```diff
@@ -1388,9 +1397,12 @@ public function isOnGround() : bool{
 	/**
 	 * @param Vector3|Position|Location $pos
 	 */
-	public function teleport(Vector3 $pos, ?float $yaw = null, ?float $pitch = null) : bool{
+	public function teleport(Vector3 $pos, ?float $yaw = null, ?float $pitch = null, ?float $headYaw = null) : bool{
 		if($pos instanceof Location){
 			$yaw = $yaw ?? $pos->yaw;
+			/** @noinspection CallableParameterUseCaseInTypeContextInspection - This will never happen. */
```
ShockedPlot7560, line 1403: this noinspection comment has no place to be here

src/entity/Location.php
```diff
@@ -32,32 +32,45 @@ class Location extends Position{
 	/** @var float */
 	public $yaw;
 	/** @var float */
+	public $headYaw;
```
ShockedPlot7560, line 35: HeadYaw has no place to be in Location.

ShockedPlot7560:
Need to target next-minor

ShockedPlot7560:
Oops, missclick on close

dktapps:
The PR is very stale, might as well close it and open a tracking issue instead

PR #4833, opened 2022-02-17
definition of the default meta 0 in BlockFactory

Virvolta:
In order to allow the definition of a block without putting the meta

## Introduction
Ihis allows you to put 0 without having to put it every time

no api change

### DIFF
--- a/src/block/BlockFactory.php
+++ b/src/block/BlockFactory.php
@@ -1042,7 +1042,7 @@ class BlockFactory{
 	 *
 	 * Deserializes a block from the provided legacy ID and legacy meta.
 	 */
-	public function get(int $id, int $meta) : Block{
+	public function get(int $id, int $meta = 0) : Block{
 		if($meta < 0 || $meta >= (1 << Block::INTERNAL_METADATA_BITS)){
 			throw new \InvalidArgumentException("Block meta value $meta is out of bounds");
 		}


dktapps:
This is intentionally mandatory.

dktapps:
I'll point out the fact that you don't actually need BlockFactory in 99% of cases, so this affects no one that is using the API correctly. This function should only be used to deserialize stored data, and you can't store a block without its metadata. Therefore allowing it to be optional makes no sense.

Virvolta:
I was thinking that you have done with the items and therefore not with the blocks

dktapps:
Actually it's the other way round. It was previously optional for BlockFactory::get(), and it was changed to mandatory to prevent bugs. As I said, there's no good reason to make it optional. In PM4 you should be using VanillaBlocks anyway.

Virvolta:
Personally I use VanillaBlock, it but to make a config where I ask for an id and a meta I have to use what because I can't do in a VanillaBlock::COAL config for example

dktapps:
You should be using StringToItemParser for that.

Virvolta:
and for the blocks ?

dktapps:
Item->getBlock()

Virvolta:
ok ok, and for methods like $item->getId() === BlockLegacyIds::COAL are to be avoided too I guess ?

dktapps:
Those things currently can't be avoided.

Virvolta:
Have you thought about how to avoid this in the future?

dktapps:
This discussion is getting off topic.

Virvolta:
it's true excuse me

PR #4851, opened 2022-02-20
Change setPageText of WritableBook

### LINKED ISSUE #4694: Improve or remove WritableBook->addPage()
## Description
`WritableBook->addPage()` is identical to `WritableBook->setPageText(pageID, "")`.

Since a plugin will rarely (if ever) want to create a book page without putting some text on it, this method is pointless and is just taking up space.

Two possible improvements come to mind:

- Make `WritableBook->addPage()` accept a `pageText` parameter. Since `add` specifically suggests that new pages will be created if they don't exist, this is likely a good idea to replace the functionality of `WritableBook->setPageText()`.
- Remove `WritableBook->addPage()`, and use `WritableBook->setPageText()` for everything.

ShockedPlot7560:
## Introduction
As stated in the issue that is linked to this PR, the ``addPage`` function is identical to ``setPageText``, both will add empty pages.

One of the given proposals is more suitable. Deleting the ``setPageText`` function was not the right way to go about it, as it would remove the ability for the plugin to directly modify the content of a page. 
The second one was to remove ``addPage``, which was content with simply adding empty pages if necessary. This mechanism not being used elsewhere could be moved to ``setPageText`` to keep this part of control to the plugin.

At the moment, there is no other solution that seems to me to be better than the second one.

### Relevant issues
* Fixes #4694 

### API changes
* Remove ``addPage(pageID)`` from ``WritableBookBase``

### Behavioural changes
N/A

## Backwards compatibility
N/A

## Tests
* Get a Book & Quill
* Create any pages of you want

### DIFF
--- a/src/item/WritableBookBase.php
+++ b/src/item/WritableBookBase.php
@@ -69,29 +69,18 @@ abstract class WritableBookBase extends Item{
 	 *
 	 * @return $this
 	 */
-	public function setPageText(int $pageId, string $pageText) : self{
+	public function setPageText(int $pageId, string $pageText = "") : self{
 		if(!$this->pageExists($pageId)){
-			$this->addPage($pageId);
+			if($pageId < 0){
+				throw new \InvalidArgumentException("Page number \"$pageId\" is out of range");
+			}
+
+			for($current = count($this->pages); $current < $pageId; $current++){
+				$this->pages[] = new WritableBookPage("");
+			}
 		}
 
 		$this->pages[$pageId] = new WritableBookPage($pageText);
-		return $this;
-	}
-
-	/**
-	 * Adds a new page with the given page ID.
-	 * Creates a new page for every page between the given ID and existing pages that doesn't yet exist.
-	 *
-	 * @return $this
-	 */
-	public function addPage(int $pageId) : self{
-		if($pageId < 0){
-			throw new \InvalidArgumentException("Page number \"$pageId\" is out of range");
-		}
-
-		for($current = count($this->pages); $current <= $pageId; $current++){
-			$this->pages[] = new WritableBookPage("");
-		}
 		return $this;
 	}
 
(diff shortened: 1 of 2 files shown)


dktapps:
This is a BC break. It can't target next-minor.

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -825,7 +825,7 @@ public function handleBookEdit(BookEditPacket $packet) : bool{
 			case BookEditPacket::TYPE_SWAP_PAGES:
 				if(!$newBook->pageExists($packet->pageNumber) || !$newBook->pageExists($packet->secondaryPageNumber)){
 					//the client will create pages on its own without telling us until it tries to switch them
-					$newBook->addPage(max($packet->pageNumber, $packet->secondaryPageNumber));
+					$newBook->setPageText(max($packet->pageNumber, $packet->secondaryPageNumber));
```
dktapps: This is no longer idiomatic. It's not obvious what this code does anymore.

src/item/WritableBookBase.php
```diff
@@ -69,29 +69,18 @@ public function getPageText(int $pageId) : string{
 	 *
 	 * @return $this
 	 */
-	public function setPageText(int $pageId, string $pageText) : self{
+	public function setPageText(int $pageId, string $pageText = "") : self{
```
dktapps: This makes no sense. Why would you call something called `setPageText()` and not provide any text ...

dktapps (changes requested):

src/network/mcpe/handler/InGamePacketHandler.php
ShockedPlot7560: I did not see that this was already checked in setPageText. 
Unnecessary repetition, might as well delete this part of the code

dktapps:

PR #4870, opened 2022-02-28
Entity: workaround teleport client bug #4394 by despawning and respawning the …

### LINKED ISSUE #4394: Players freezes after teleportations randomly
### Issue description

- Expected result: Players movement should continue being broadcasted to others
- Actual result: Players movement is temporarily frozen

1. Throw an ender pearl while running, make sure to stop moving before it lands. Now until you move you're stuck on everyone elses screen.
### OS and versions

* PocketMine-MP: 4.0.0
* PHP: 8.0.7
* Using JIT: no
* Server OS: Ubuntu
* Game version: Win10

### Plugins
Plugins (2): Core, Anticheat

  Yes
  This fix was given by ZodiaX
  ```php
  // after teleporting a player do this
  $player->broadcastMovement(true);
  $player->broadcastMotion();
    ```

https://youtu.be/_JZ71hPMbtk

dktapps:
…entity

this is a sucky solution but it works well enough for this scenario.

From my research it appears that while the client-side MovementInterpolator is active on the client and hasn't yet reached its target, any teleport sent during the lerp will get overridden by the lerp.
This appears to last 3-4 ticks (about 150-200 ms) which explains why this can be reproduced by stopping movement just before the ender pearl lands (human reaction times would make the reproduction unreliable otherwise).

Things to note:
- All entities are affected by this bug.
- MovePlayerPacket appears not to have the problem, but we can't use it because it doesn't work for non-players.
- MoveActorAbsolute and MoveActorDelta are both handled by the same code and will have the same lerping bug.

This is the cleanest solution I could come up with.

### Relevant issues
fixes #4394 

### Behavioural changes
A different method is now used to send teleport changes to the client.

## Backwards compatibility
None. This affects internal protocol usage only.

## Follow-up
This hack should be removed if the bug in MCPE is ever fixed.

## Tests
This can be tested before and after using the following code:
```php
	public function onEntityInteract(PlayerEntityInteractEvent $event) : void{
		$entity = $event->getEntity();
		$location = $entity->getLocation();
		$this->getServer()->broadcastPackets($entity->getViewers(), [MoveActorAbsolutePacket::create(
			$entity->getId(),
			$entity->getOffsetPosition($location),
			$location->pitch,
			$location->yaw,
			$location->yaw,
			(
				($entity->onGround ? MoveActorAbsolutePacket::FLAG_GROUND : 0)
			)
		)]);
		$event->getEntity()->teleport($event->getEntity()->getPosition()->add(10, 0, 0));
	}
```

It can also be tested by two players where one is using an ender pearl as described in #4394.


dktapps:
@Prim69 as the original reporter of this issue I'd like to get your feedback on this to see if there's any problems with it.

dktapps:
Merging as-is due to lack of comment; assuming this works fine. No doubt someone will complain after the fact ...

Issue #4856, opened 2022-02-22
Verification of the placement in the worlds


ShockedPlot7560:
## Introduction 

Many blocks (most as Flowable) cannot be placed on certain types of blocks. We don't currently have a way for each block to give the surface type for each side.  
With this in mind, I would like to know your position after mine to refine this system and potentially move on to a PR. As you can see on [the block placement wiki](https://minecraft.fandom.com/wiki/Opacity/Placement) (not necessarily up to date but gives an overview) some blocks only allow some blocks on one side, while they accept others on others. This has the disadvantage that you can't really find a generic placement solution for each of them. 

## The concept

The system I imagine would be based on 2 functions in particular which would allow each one to provide the size of the surface of the given side. There are two types of notion to understand for the continuation: 
- The placement, represents the face that is clicked when you want to place a block, in other words, refers to the support block for the placement.
- The attachability, will represent the surface of the block to be placed that will be in contact with the support block. For example, the banner oriented horizontally will be attached to its support block.

With these notions in mind, there would be a function handling each case. Each one would return a float representing the (abstract) size of its face, where 0 represents a non-placable/attachable face and 1 the maximum obtainable. First, ``getSurfaceSize(int $face)`` will provide the size of its surfaces with the desired face as parameters. Second, ``getAttachSize(int $face, ?int $tempFacing = null)`` will reference the minimum size needed for this block to be placed. Let's imagine a torch with an attachment size of 0.5, a block of earth (1 in surface), can be placed but cannot be placed on a block with 0 or 0.1 area.

So, when we place a block, a check will be made looking at the size of the clicked surface, and the attachment size of the opposite of the current block and will compare them. If the surface is larger than or equal to the attachment, the placement is ok, otherwise the placement is impossible. A dedicated function is most likely to be considered.

## Abstract sizes

The difficulty is going to reside there in the definition of constant growing represents a hierarchy of placement. Several manage to distinguish themselves:
> - The signs / banners are placed on the same types of blocks and appear to be the minimum for each side.
> - The torches / lanterns / pressure plates will also behave similarly.  
> - ...
> 
> All blocks can be grouped according to different criteria

The best would be to start from 0 (zero sizes) and go to 1 (full sizes). With this in mind, torches and other blocks with similar behavior would have a size of 0.5, this size would allow better flexibility.

[...]

Issue #4529, opened 2021-10-31
Lanterns cannot be placed on all transparent blocks


Nerahikada:
### Issue description
Lanterns can be placed on some permeable blocks.
For example, it can place on certain surfaces of stairs or slabs.

### Steps to reproduce the issue
1. Place lanterns on glass

### OS and versions
* PocketMine-MP: 63dfcc6
* reproduced at test.pmmp.io:19132

Nerahikada:
This problem is not only seen in the lantern but also in some other blocks such as the bed.

alvin0319:
The problem is caused by this code (in case of bed): https://github.com/pmmp/PocketMine-MP/blob/faad2365e2529d4bb6e86fcb26d6f46dc7ee93c9/src/block/Bed.php#L185

dktapps:
That makes no sense. That would allow lanterns to be placed on blocks like tallgrass.

dktapps:
Relates to #4856, fixed by #4886

Issue #3299, opened 2020-02-08
Button,Presure plates etc can be placed on another button


provsalt:
### Issue description

- Expected result: Button does not place
- Actual result: Button places snd look like this 
![Screenshot_2020-02-08-23-20-03-53_5c8300b655012b1930f2e0a7b81bf6a9](https://user-images.githubusercontent.com/39023494/74087787-89bb1100-4a87-11ea-87ac-b51f4d13f22c.png)

### Steps to reproduce the issue
1. Sneak
2. and then place a button on the button.
3. you will realize that you can place a button on that button

### OS and versions

* PocketMine-MP: 3.11.5
* PHP: 7.2
* Server OS: Linux, Ubuntu 18.04 LTS
* Game version: PE

### Plugins

Plugins (0):

### Crashdump, backtrace or other files

None

dktapps:
yes, there's a TODO in the code.

dktapps:
Relates to #4856, fixed by #4886

PR #4886, opened 2022-03-05
Introduce support types logic

ShockedPlot7560:
## Introduction
Many block cannot be placed on certain types of blocks. In addition, there is currently no way for blocks to specify the type of their surface for block placement

## The concept 
It all comes down to the new feature that has been added: ``getSupportType(Facing) : SupportType``. It will then return 4 different types of surfaces according to this table: 
Type | Description
----- | -----
Full | The support face doesn't have any restrictions on support it can provide.
Edge | Support any element which reach the eadges of the block : repeaters,comparator on top of hoppers, cauldron)
Center | Support any elements which can placed in the center (torch on top of fence or wall)
None | No placement is allowed

The discussion and further explanation of this concept can be found here : https://github.com/pmmp/PocketMine-MP/issues/4856

### Relevant issues
* #4856 
* https://github.com/pmmp/PocketMine-MP/issues/458
* https://github.com/pmmp/PocketMine-MP/issues/4529
* https://github.com/pmmp/PocketMine-MP/issues/3299

Some blocks have also had their placement corrected such as ``RedstoneWire`` which was not checking any support

### API changes
* Add ``getSupportType()`` in ``Block``
* Add ``SupportType`` classes

### Behavioural changes
Any block that inherits from the parent class : ``Block`` will have a ``None`` surface by default. The default value for ``Opaque`` is: ``Full`` 

## Backwards compatibility
Some block placements have been reviewed and fixed but some may have been impacted due to the change in behavior of the block placement

## Tests
Take the blocks that have been modified and test their placement one by one in all possible positions and situations. 
These were done during development and would be too long to reproduce on video.


src/block/Stair.php
```diff
@@ -128,4 +129,11 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
 
 		return parent::place($tx, $item, $blockReplace, $blockClicked, $face, $clickVector, $player);
 	}
+
+	public function getSupportType(int $facing) : SupportType{
+		if($facing === Facing::UP && $this->isUpsideDown() || $facing === $this->facing || $facing === Facing::DOWN && !$this->isUpsideDown()){
+			return SupportType::FULL();
```
dktapps: This needs to include logic for corner stairs.

dktapps: This is also potentially subject to precedence issues (needs more ().)

ShockedPlot7560: in fact, the corner stairs are in this case problematic

ShockedPlot7560: Do we really implement as vanilla or do we add this feature: 
```
Take a corner stair type of INNER_RIGHT facing north.
In vanilla, you will be able to put torch on the north face and not on the east.
Knowing that east face are full, should we allowed the torch placement or not ?
```

dktapps:
Hi, I haven't had time to review this yet. It's a big PR and I've been busy with other things the last couple of weeks. I'll get to it soon.

ShockedPlot7560:
Ok no problem.

dktapps:
A couple of suggestions come to mind (I haven't reviewed the whole thing yet):

`SupportType->hasEdgeSupport()` and `SupportType->hasCenterSupport()` would probably be useful, because:

| Support type | Supports edges | Supports center |
|:--------------:|:------------------:|:------------------:|
| None | | |
| Center | | YES |
| Edge | YES | |
| Full | YES | YES |

Typically this means that we're usually going to be checking for two types of support: `FULL` + one of `CENTER` or `EDGE`.

src/block/Block.php
```diff
@@ -642,4 +643,8 @@ public function calculateIntercept(Vector3 $pos1, Vector3 $pos2) : ?RayTraceResu
 
 		return $currentHit;
 	}
+
+	public function getSupportType(int $facing) : SupportType{
+		return SupportType::NONE();
```
dktapps: I'm not sure this really makes sense. `Block` assumes the presence of a 1x1x1 collision cube, which normally equates to full support.

dktapps (commented):
Concept is OK, although there's definitely ways to make the code clearer, and some of the choices for support types are a bit weird. Needs more work.

dktapps:
Things noticed so far:
- Hopper's upper face seems to be FULL, rather than EDGE. This is counterintuitive, but it's the only way I can explain the behaviour I see (torch can be placed on top of hopper, but not on top of stuff like cauldron or composter, while rails can)
- Spawner is weird. It accepts placement of torches, but not stuff like buttons, levers, pressure plates etc. I don't know what to make of it.
- Buttons should require CENTER support type
- Flower pot should require CENTER support type
- Lanterns should require CENTER support type
- Lever should require CENTER support type
- Torch requires CENTER support when placed on the top face of a block. It can't be placed on top of a cauldron or composter.

... will update via edits while testing

src/block/Block.php
ShockedPlot7560: Indeed, this part should be as ``Full``, and the ``Transparent`` should be modified for a none support

src/block/DeadBush.php
```diff
@@ -34,15 +35,15 @@
 class DeadBush extends Flowable{
 
 	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
-		if(!$this->getSide(Facing::DOWN)->isTransparent()){
+		if($this->getSide(Facing::DOWN)->getSupportType(Facing::UP)->equals(SupportType::FULL())){
```
dktapps: I don't think this needs to be changed

src/block/Lantern.php
```diff
@@ -77,7 +78,10 @@ protected function canAttachTo(Block $b) : bool{
 	}
 
 	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
-		if(!$this->canAttachTo($this->position->getWorld()->getBlock($blockReplace->getPosition()->up())) && !$this->canAttachTo($this->position->getWorld()->getBlock($blockReplace->getPosition()->down()))){
+		if($face === Facing::DOWN && $blockReplace->getSide(Facing::UP)->getSupportType(Facing::DOWN)->equals(SupportType::NONE())){
+			return false;
+		}
+		if($face === Facing::UP && $blockReplace->getSide(Facing::DOWN)->getSupportType(Facing::UP)->equals(SupportType::NONE())){
```
dktapps: This can be simplified to check for `Axis::Y` and use `Facing::opposite()` to get the appropriate facings.

src/block/Hopper.php
```diff
@@ -99,5 +100,12 @@ public function onScheduledUpdate() : void{
 		//TODO
 	}
 
+	public function getSupportType(int $facing) : SupportType{
+		if($facing === Facing::UP){
+			return SupportType::EDGE();
```
dktapps: It's unclear what the correct behaviour of this should be. Many blocks (e.g. torches, lanterns) can be placed on top of a hopper, but can't be placed on top of other EDGE blocks (such as composters); however, rails can be placed on both.

I think the correct thing for hopper would be to treat the top face as FULL. I don't like it, but it seems to be how the game itself treats it.

src/block/RedMushroom.php
```diff
@@ -43,7 +44,7 @@ public function onNearbyBlockChange() : void{
 
 	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
 		$down = $this->getSide(Facing::DOWN);
-		if(!$down->isTransparent()){
+		if(!$down->getSupportType(Facing::UP)->equals(SupportType::FULL())){
```
dktapps: I don't think this should be changed

src/block/Vine.php
```diff
@@ -116,7 +117,7 @@ protected function recalculateCollisionBoxes() : array{
 	}
 
 	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
-		if(!$blockClicked->isSolid() || Facing::axis($face) === Axis::Y){
+		if(!$blockClicked->getSupportType($face)->equals(SupportType::FULL()) || Facing::axis($face) === Axis::Y){
```
dktapps: I don't think this is necessary

dktapps:
The following blocks need to be changed:
- Doors (any FULL/EDGE block is sufficient; they should be placeable on slabs, composters and cauldrons)
- Beds (CENTER/EDGE/FULL all seem to be OK)
- Dead bush (these require certain specific blocks, not support types)
- Lantern (the placement logic has not been completely updated)
- Mushroom (the logic has been changed incorrectly; the original logic is correct aside from the missing lightlevel check)

dktapps:
- Bell (any support seems to be OK; they can be placed on full blocks, composters, and fences, but not on beds)
- Snow layer (appears to require CENTER support, as best I can tell)
- Coral (appears to require CENTER support)
- Coral fan (appears to require CENTER support)

ShockedPlot7560:
> Spawner is weird. It accepts placement of torches, but not stuff like buttons, levers, pressure plates etc. I don't know what to make of it.

Looking at the wiki and practicing my vanilla tests, we can see that the spawner accepts torches but not other blocks. This can be explained by the fact that it is transparent, and thus everything that is lever, button, pressure plate is useless (does not transmit any current)

dktapps:
That's not enough to explain it. Glass has all those same properties, and yet still accepts the placement of all of these blocks.

ShockedPlot7560:
It is true, in this case, it is necessary to leave in a logic external to minecraft vanilla and allow the placement of all

ShockedPlot7560:
> Lantern (the placement logic has not been completely updated)

Are you talking about the check to define if the lantern is hanging?
```php
$this->hanging = ($face === Facing::DOWN || !$this->canAttachTo($this->position->getWorld()->getBlock($blockReplace->getPosition()->down())));
```

ShockedPlot7560:
All requested changes have been made. The only things pending, waiting for an answer, for debate are: 
- Spawner: Should we give them a ``FULL`` support for all surfaces, or should we try to stick to minecraft vanilla as currently.
- Lantern: Which investment is currently unimplemented, missing?

src/block/Transparent.php
```diff
@@ -23,9 +23,15 @@
...
+use pocketmine\block\utils\SupportType;
+
 class Transparent extends Block{
 
 	public function isTransparent() : bool{
 		return true;
 	}
+
+	public function getSupportType(int $facing) : SupportType{
+		return SupportType::NONE();
```
dktapps: This is incorrect. Many transparent blocks are still full cubes.

dktapps:
> Spawner: Should we give them a FULL support for all surfaces, or should we try to stick to minecraft vanilla as currently.

Spawner shouldn't have any support, since that's how it appears to behave.

src/block/RedMushroom.php
```diff
@@ -42,11 +41,18 @@ public function onNearbyBlockChange() : void{
...
 
 		return false;
 	}
+
+	protected function canBeSupportedBy(Block $block) : bool{
+		$position = $block->getPosition();
+		$lightLevel = $position->getWorld()->getFullLightAt($position->x, $position->y, $position->z);
+		$id = $block->getId();
+		// TODO: nylium support
+		return ($lightLevel <= 12 && !$block->isTransparent()) || $id === BlockLegacyIds::MYCELIUM || $id === BlockLegacyIds::PODZOL;
```
dktapps: The light level here will always be zero if the block is opaque.

dktapps:
> Lantern: Which investment is currently unimplemented, missing?

You've left a usage of `canAttachTo()` which has inconsistent behaviour with the new code.

src/block/Transparent.php
ShockedPlot7560: I agree on this point, however, some blocks are not real cube and do not require special support that ``NONE``.  
Instead of defining the supports of the transparent blocks one by one in ``NONE``, it is better to do the case by case of transparent blocks not having a null support ?

dktapps: No, I don't agree. For `Flowable` this would make sense, but not for `Transparent`.

src/block/Lantern.php
```diff
@@ -81,7 +81,7 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
 			return false;
 		}
 
-		$this->hanging = ($face === Facing::DOWN || !$this->canAttachTo($this->position->getWorld()->getBlock($blockReplace->getPosition()->down())));
+		$this->hanging = ($face === Facing::DOWN || !$this->canBeSupportedBy($this->position->getWorld()->getBlock($blockReplace->getPosition()->down()), Facing::UP));
```
dktapps, line 85: canAttachTo() is now dead code

src/block/MonsterSpawner.php
```diff
@@ -42,11 +40,4 @@ protected function getXpDropAmount() : int{
 	public function onScheduledUpdate() : void{
 		//TODO
 	}
-
-	public function getSupportType(int $facing) : SupportType{
-		if($facing === Facing::UP || $facing === Facing::DOWN){
-			return SupportType::CENTER();
-		}
-		return SupportType::FULL();
-	}
```
dktapps: This should have NONE support type

ShockedPlot7560: Indeed, I thought ``Block`` was still setting ``NONE`` by default, so much for me

ShockedPlot7560:
Is the PhpStan problem important? It's a typing problem, but casting the value as an integer doesn't seem like a good idea.

Looking at the uses of ``getFullLightAt``, I could notice that the use is the same, but no typing error came out.

dktapps:
No, it's currently ignored via `actual-problems.neon` baseline

src/block/Fence.php
```diff
@@ -93,4 +94,11 @@ protected function recalculateCollisionBoxes() : array{
 
 		return $bbs;
 	}
+
+	public function getSupportType(int $facing) : SupportType{
+		if(Facing::axis($facing) === Axis::Y){
+			return SupportType::CENTER();
+		}
+		return SupportType::NONE();
```
dktapps: This can be reduced to a ternary

src/block/Hopper.php
```diff
@@ -99,5 +100,12 @@ public function onScheduledUpdate() : void{
 		//TODO
 	}
 
+	public function getSupportType(int $facing) : SupportType{
+		if($facing === Facing::UP){
+			return SupportType::FULL();
+		}
+		return SupportType::NONE();
+	}
```
dktapps: This can be reduced to a ternary

src/block/SnowLayer.php
```diff
@@ -81,7 +82,7 @@ protected function recalculateCollisionBoxes() : array{
 	}
 
 	private function canBeSupportedBy(Block $b) : bool{
-		return $b->isSolid() || ($b instanceof SnowLayer && $b->isSameType($this) && $b->layers === self::MAX_LAYERS);
+		return $b->getSupportType(Facing::UP)->hasCenterSupport() || ($b instanceof SnowLayer && $b->isSameType($this) && $b->layers === self::MAX_LAYERS);
```
dktapps: I think the second condition can be removed with this change

src/block/Torch.php
```diff
@@ -87,12 +89,21 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
...
+				if($this->canBeSupportBy($block, Facing::opposite($side))){
 					$this->facing = Facing::opposite($side);
 					return parent::place($tx, $item, $blockReplace, $blockClicked, $face, $clickVector, $player);
 				}
 			}
 		}
 		return false;
 	}
+
+	protected function canBeSupportBy(Block $support, int $face) : bool{
```
dktapps: Inconsistent function name

src/block/Wall.php
```diff
@@ -76,4 +77,11 @@ protected function recalculateCollisionBoxes() : array{
 				->trim(Facing::EAST, $east ? 0 : $inset)
 		];
 	}
+
+	public function getSupportType(int $facing) : SupportType{
+		if($facing === Facing::UP || $facing === Facing::DOWN){
```
dktapps: Facing::axis can be used here

src/block/Wall.php
```diff
@@ -76,4 +77,11 @@ protected function recalculateCollisionBoxes() : array{
...
+			return SupportType::CENTER();
+		}
+		return SupportType::NONE();
+	}
```
dktapps, line 84: This can be reduced to a ternary

dktapps:
To clarify the reasoning for the inversion of default support types:

Support type differences generally go hand-in-hand with collision box shape differences (with few exceptions). Therefore, since the default behaviour is to assume that blocks have a full collision box, they should also be assumed to have full support on all faces unless otherwise specified. This is why Transparent can't declare NONE support.

I'm aware that this results in more changed files than it would otherwise have done, but this way makes more sense - you'll notice that usually the changes sit right next to `recalculateCollisionBoxes()` or something similar.

ShockedPlot7560:
That was the behavior I understood in the end. It was not my logic at first, but now it seems more coherent with the rest

src/block/BaseRail.php
```diff
@@ -220,11 +222,13 @@ private function setConnections(array $connections) : void{
...
 	public function onNearbyBlockChange() : void{
-		if($this->getSide(Facing::DOWN)->isTransparent()){
+		$supportType = $this->getSide(Facing::DOWN)->getSupportType(Facing::UP);
+		if($supportType->equals(SupportType::NONE()) || $supportType->equals(SupportType::CENTER())){
 			$this->position->getWorld()->useBreakOn($this->position);
 		}else{
 			foreach($this->getCurrentShapeConnections() as $connection){
-				if(($connection & RailConnectionInfo::FLAG_ASCEND) !== 0 && $this->getSide($connection & ~RailConnectionInfo::FLAG_ASCEND)->isTransparent()){
+				$supportType = $this->getSide($connection & ~RailConnectionInfo::FLAG_ASCEND)->getSupportType(Facing::UP);
+				if(($connection & RailConnectionInfo::FLAG_ASCEND) !== 0 && ($supportType->equals(SupportType::NONE()) || $supportType->equals(SupportType::CENTER()))){
```
dktapps: I'm not convinced this is correct. Slope support should depend on the face of the block it leans on, not the top face.

src/block/BaseRail.php
```diff
@@ -38,7 +39,8 @@
 abstract class BaseRail extends Flowable{
 
 	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
-		if(!$blockReplace->getSide(Facing::DOWN)->isTransparent()){
+		$supportType = $blockReplace->getSide(Facing::DOWN)->getSupportType(Facing::UP);
+		if(!$supportType->equals(SupportType::NONE()) && !$supportType->equals(SupportType::CENTER())){
```
dktapps: `supportType->hasEdgeSupport()` can be used here

src/block/BaseRail.php
```diff
@@ -220,11 +222,13 @@ private function setConnections(array $connections) : void{
 	}
 
 	public function onNearbyBlockChange() : void{
-		if($this->getSide(Facing::DOWN)->isTransparent()){
+		$supportType = $this->getSide(Facing::DOWN)->getSupportType(Facing::UP);
+		if($supportType->equals(SupportType::NONE()) || $supportType->equals(SupportType::CENTER())){
```
dktapps: `!supportType->hasEdgeSupport()` can be used here

src/block/BaseRail.php
dktapps: `!supportType->hasEdgeSupport()` can be used here

src/block/Slab.php
```diff
@@ -128,4 +129,13 @@ protected function recalculateCollisionBoxes() : array{
 	public function getDropsForCompatibleTool(Item $item) : array{
 		return [$this->asItem()->setCount($this->slabType->equals(SlabType::DOUBLE()) ? 2 : 1)];
 	}
+
+	public function getSupportType(int $facing) : SupportType{
+		if($this->getSlabType()->equals(SlabType::DOUBLE())){
```
dktapps: Getter usage unnecessary here

src/block/BaseRail.php
ShockedPlot7560: It is not even either of the two logics. As can be tested in vanilla, by putting a hopper and rails, the rail seems to be based on the block below it all the time.

dktapps: Never mind, you are correct: 
image

src/block/BaseRail.php
```diff
@@ -38,7 +38,8 @@
 abstract class BaseRail extends Flowable{
 
 	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
-		if(!$blockReplace->getSide(Facing::DOWN)->isTransparent()){
+		$supportType = $blockReplace->getSide(Facing::DOWN)->getSupportType(Facing::UP);
```
dktapps: Extra variable can be removed

src/block/BaseRail.php
```diff
@@ -220,11 +221,13 @@ private function setConnections(array $connections) : void{
 	}
 
 	public function onNearbyBlockChange() : void{
-		if($this->getSide(Facing::DOWN)->isTransparent()){
+		$supportType = $this->getSide(Facing::DOWN)->getSupportType(Facing::UP);
```
dktapps: same here

src/block/BaseRail.php
```diff
@@ -220,11 +221,13 @@ private function setConnections(array $connections) : void{
...
+		if(!$supportType->hasEdgeSupport()){
 			$this->position->getWorld()->useBreakOn($this->position);
 		}else{
 			foreach($this->getCurrentShapeConnections() as $connection){
-				if(($connection & RailConnectionInfo::FLAG_ASCEND) !== 0 && $this->getSide($connection & ~RailConnectionInfo::FLAG_ASCEND)->isTransparent()){
+				$supportType = $this->getSide($connection & ~RailConnectionInfo::FLAG_ASCEND)->getSupportType(Facing::UP);
```
dktapps: and here

src/block/FlowerPot.php
```diff
@@ -104,19 +104,23 @@ protected function recalculateCollisionBoxes() : array{
 	}
 
 	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
-		if($this->getSide(Facing::DOWN)->isTransparent()){
+		if($this->canBeSupportedBy($this->getSide(Facing::DOWN), Facing::UP)){
```
dktapps:
```suggestion
		if(!$this->canBeSupportedBy($this->getSide(Facing::DOWN), Facing::UP)){
```

src/block/Lantern.php
```diff
@@ -72,22 +73,27 @@ protected function recalculateCollisionBoxes() : array{
 		];
 	}
 
-	protected function canAttachTo(Block $b) : bool{
-		return !$b->isTransparent();
-	}
-
 	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
-		if(!$this->canAttachTo($this->position->getWorld()->getBlock($blockReplace->getPosition()->up())) && !$this->canAttachTo($this->position->getWorld()->getBlock($blockReplace->getPosition()->down()))){
+		if(Facing::axis($face) === Axis::Y && !$this->canBeSupportedBy($blockReplace->getSide(Facing::opposite($face)), $face)){
```
dktapps: This is incorrect. The block should look for up/down supports regardless of which face was clicked.

You can see the problem by right-clicking with a lantern on a wall when there are no available attachment points - the lantern item will be consumed, but no lantern is placed.

src/block/RedMushroom.php
dktapps: I'm not actually sure this logic belongs in this PR at all. It's unrelated to the objective, which is to introduce support types logic.

src/block/Sapling.php
```diff
@@ -125,4 +124,12 @@ private function grow(?Player $player) : bool{
 	public function getFuelTime() : int{
 		return 100;
 	}
+
+	protected function hasValidSupport() : bool{
+		$down = $this->getSide(Facing::DOWN);
+		if($down->getId() === BlockLegacyIds::GRASS || $down->getId() === BlockLegacyIds::DIRT || $down->getId() === BlockLegacyIds::FARMLAND){
```
dktapps: Not sure this change belongs in this PR.

src/block/Sugarcane.php
```diff
@@ -118,11 +118,24 @@ public function onRandomTick() : void{
...
+	 * checks only the solid support of the sugar cane and not the proximity with water
+	 */
+	protected function canBeSupportedBy(Block $block) : bool{
+		$id = $block->getId();
+		//TODO: rooted dirt, moss block
+		return $id === BlockLegacyIds::GRASS
+			|| $id === BlockLegacyIds::DIRT
+			|| $id === BlockLegacyIds::PODZOL
+			|| $id === BlockLegacyIds::MYCELIUM
+			|| $id === BlockLegacyIds::SAND;
```
dktapps: Not sure this change belongs in this PR.

src/block/RedMushroom.php
ShockedPlot7560: I'm going to move all this to other pull requests soon if not done before. (Light condition)

dktapps:
Please don't mark comments as resolved until they've been addressed by commits on the branch

src/block/DeadBush.php
```diff
@@ -64,4 +64,12 @@ public function getFlameEncouragement() : int{
...
 		return 100;
 	}
+
+	protected function canBeSupportedBy(Block $block) : bool{
+		$id = $block->getId();
+		return $id === BlockLegacyIds::SAND
+			|| $id === BlockLegacyIds::PODZOL
+			|| $id === BlockLegacyIds::DIRT
+			|| $id === BlockLegacyIds::TERRACOTTA;
+	}
```
dktapps: This should also be moved to a separate PR

dktapps:
Updated description:

- this does not fix vines being placeable on the side of cacti - that requires further research
- it does not fix painting interaction with signs
- it does not fix saplings on farmland (that should be done separately)

From what I can see in my testing, this looks pretty much good to go.

ShockedPlot7560:
Thanks for updating, I hadn't noticed that it was obsolete
Once this PR is done, I will make separate placements that have not been fixed by it

src/block/RedstoneWire.php
```diff
@@ -45,4 +57,14 @@ public function readStateFromWorld() : void{
...
 	}
+
+	public function onNearbyBlockChange() : void{
+		if(!$this->canBeSupportedBy($this->getSide(Facing::DOWN))){
+			$this->position->getWorld()->useBreakOn($this->position);
+		}
+	}
+
+	private function canBeSupportedBy(Block $block) : bool{
+		return $block->getSupportType(Facing::UP)->hasEdgeSupport();
```
dktapps:
```suggestion
		return $block->getSupportType(Facing::UP)->hasCenterSupport();
```

src/block/Slab.php
```diff
@@ -125,6 +126,15 @@ protected function recalculateCollisionBoxes() : array{
 		return [AxisAlignedBB::one()->trim($this->slabType->equals(SlabType::TOP()) ? Facing::DOWN : Facing::UP, 0.5)];
 	}
 
+	public function getSupportType(int $facing) : SupportType{
+		if($this->slabType->equals(SlabType::DOUBLE())){
+			return SupportType::FULL();
+		}elseif($facing === Facing::UP && $this->slabType->equals(SlabType::TOP()) || $facing === Facing::DOWN && $this->slabType->equals(SlabType::BOTTOM())){
```
dktapps:
```suggestion
		}elseif(($facing === Facing::UP && $this->slabType->equals(SlabType::TOP())) || ($facing === Facing::DOWN && $this->slabType->equals(SlabType::BOTTOM()))){
```

src/block/SnowLayer.php
```diff
@@ -80,8 +81,15 @@ protected function recalculateCollisionBoxes() : array{
...
+	public function getSupportType(int $facing) : SupportType{
+		if(!$this->canBeReplaced()){
+			return SupportType::FULL();
+		}
+		return SupportType::NONE();
+	}
+
 	private function canBeSupportedBy(Block $b) : bool{
-		return $b->isSolid() || ($b instanceof SnowLayer && $b->isSameType($this) && $b->layers === self::MAX_LAYERS);
+		return $b->getSupportType(Facing::UP)->hasCenterSupport();
```
dktapps: This requires full support, as far as I can tell from testing

PR #4901, opened 2022-03-16
EnumTrait: Name can be set to one character

PJZ9n:
The cause of this problem is that `\D` is interpreted as a single character.

## Introduction
This fixes an issue where the Enum name could not be set to a single character.

### Relevant issues
No

### API changes
It's not an API change, but it's a bug fix.

### Behavioural changes
The enum name can be one character!

## Backwards compatibility
No break, this is bug fix.
Code that sets the enum name to one character and expects an exception to break is broken, but this is rare.

## Follow-up
None

## Tests
This is a simple regular expression change

7ca0ed39482dc3760d7a15fae020bd3a
69d9ace1e5645b56d9dee68744e5197d

### DIFF
--- a/src/utils/EnumTrait.php
+++ b/src/utils/EnumTrait.php
@@ -70,7 +70,7 @@ trait EnumTrait{
 	 * @throws \InvalidArgumentException
 	 */
 	private function __construct(string $enumName){
-		if(preg_match('/^\D[A-Za-z\d_]+$/u', $enumName, $matches) === 0){
+		if(preg_match('/^(?!\d)[A-Za-z\d_]+$/u', $enumName, $matches) === 0){
 			throw new \InvalidArgumentException("Invalid enum member name \"$enumName\", should only contain letters, numbers and underscores, and must not start with a number");
 		}
 		$this->enumName = $enumName;


dktapps:
This is an incorrect fix. The better solution would be to change the `+` modifier to a `*`.

PJZ9n:
~~I thought so too, but then it would match a one-letter underscore.~~
This was the expected behavior

PJZ9n:
I thought so too, but then it would matches one character other than numbers. (e.g. `@`)

dktapps:
That's a topic for another time. Stick to one problem at a time.

PJZ9n:
Why is `(?!\d)` not optimal?

Does this change mean that it doesn't meet this specification?
`should only contain letters, numbers and underscores, and must not start with a number`

I think changing `+` to `*` is just causing a new bug.

dktapps:
You're overcomplicating this by trying to fix two unrelated things at the same time.

dktapps:
Never mind, I see the problem. Excuse my mistake, I haven't slept for 2 days.

dktapps:
Just so I understand fully, the regex asserts that the string does not begin with a number, and then the [] match matches from position 0, correct? Previously I suppose it must have matched from 1.

PJZ9n:
Yes, it's correct.
Previously, `\D` itself matched as a "non-numeric one character".
Therefore, when a one-letter enum name was input, it did not match the following `[A-Za-z\d_]+`.

If you don't like this commit fixing the bug, you can do the following:
First, fix the inability to create a single character EnumTrait. (`^\D[A-Za-z\d_]+$` to `^\D[A-Za-z\d_]*$`), and make a commit that fixes it because it has a bug. (`^\D[A-Za-z\d_]*$` to ` ^(?!\d)[A-Za-z\d_]+$`)
Or you could put it together in one commit, "EnumTrait: Accurate name validation".

What do you think?

dktapps:
No, I think this is sufficient. My only concern with it was because I misunderstood what the regex was doing.

PJZ9n:
ok, you can merge this.

PR #4912, opened 2022-03-19
Throw exception if an event handler is a generator

SOF3:
## Introduction

Plugins that copy code from await-generator examples may forget to wrap their code in `Await::f2c`, and end up having a function that merely returns a `Generator`, executing nothing semantically without triggering any error.

It would be nicer to require them to declare `: void`, but that might break a lot of plugins. Meanwhile, handler functions that are actually generator functions are doubtlessly bugs.

By the way, `ReflectionFunctionAbstract::isGenerator()` only reports functions with a literal `yield`, and does not report functions that declare `Generator` as a return type (although that is equally pointless in the case of event handlers).

### API changes

Throw exception if a Listener class contains handler methods that are generator methods (have a literal `yield` in the function body, such that calling the function without using the return value executes absolutely nothing other than allocating a generator object).

### Behavioural changes

nil

## Backwards compatibility

Any plugins affected by this were most likely (if not always) written incorrectly to start with. I do not consider this as a BC break.

## Tests

TODO


dktapps:
ACK, but the build needs to pass.

dktapps:
Actually I'm not sure that throwing an exception here makes sense. In line with other checks it might make more sense to just skip it instead.

dktapps:
It's more likely that a plugin declares a generator function accepting an Event class and doesn't consider that it might be treated as a handler.

Since the current behaviour in this case would be to just do nothing anyway (calling a generator function does nothing until the generator object is iterated on), it would already be doing nothing in such cases anyway. I don't think it's reasonable to start randomly throwing exceptions for this case, since there wouldn't actually have been any unexpected behaviour to begin with.

dktapps:
For posterity, my latest from Discord:

can i get an example of a way #4912 is a problem?
i'm having a hard time seeing why making PM yell about this is the right solution
the harder i look at it, the less sense it makes
the problem you described: people may call a generator function and forget to wrap it in f2c/g2c
the solution: ban event handlers from being generators ???
i don't see how this solves anything
also, this change doesn't cover registerEvent(), which accepts a closure directly

SOF3:
the problem with generators in particular is that it's easy for people to copy code (e.g. from some other plugins) without being aware that it only works in an await-generator context. The dangerous thing is that the handler will simply do nothing and it is difficult for them or others to debug this issue until they realized they copied yield statements into the function, and then they still need to be aware that yield statements will make a function simply do nothing other than returning an object that does nothing without getting called.

dktapps:
If people are blindly copying code, they are going to have problems in any case. I don't really consider this a dangerous issue. It's not really any different than forgetting to apply @handleCancelled for dealing with bans in PlayerPreLoginEvent.

dktapps:
I'm going to merge this, since I think the benefit probably outweighs the negatives. If anyone has unintentional handlers that bail because of this, they can simply use `@notHandler` (which is what they should have done anyway).

dktapps:
Implemented by 4fab518384b1525f2f126643c57f1df6446557df.

PR #4913, opened 2022-03-19
Fixed falling damage caused by changing the player's position during a PlayerToggleSneakEvent.

rasu3n:
## Introduction

https://user-images.githubusercontent.com/76860328/159113616-3370d5d9-ae7a-4039-8a2c-76f5430e8229.mp4

## Tests

https://user-images.githubusercontent.com/76860328/159114401-62e3f1cc-7cdb-49cb-b9b3-dab887bc1e1b.mp4

Code to reproduce the problem
```php
/** @var PluginBase $plugin */
$server = $plugin->getServer();
$server->getPluginManager()->registerEvent(PlayerToggleSneakEvent::class, static function(PlayerToggleSneakEvent $ev) : void{
	if(!$ev->isSneaking()){
		return;
	}
	$player = $ev->getPlayer();
	$position = $player->getPosition();
	$block = $position->getWorld()->getBlock($position->down());
	$newY = match(true){
		$block->isSameType(VanillaBlocks::IRON()) => 4,
		$block->isSameType(VanillaBlocks::GOLD()) => -4,
		default => null
	};
	if($newY === null){
		return;
	}
	$player->teleport($position->add(0, $newY, 0));
}, EventPriority::NORMAL, $plugin);
$plugin->getScheduler()->scheduleRepeatingTask(new ClosureTask(static function() use ($server) : void{
	foreach($server->getOnlinePlayers() as $player){
		$position = $player->getPosition();
		$position->getWorld()->addParticle($position, new RedstoneParticle());
	}
}), 1);
```

### DIFF
--- a/src/network/mcpe/handler/InGamePacketHandler.php
+++ b/src/network/mcpe/handler/InGamePacketHandler.php
@@ -228,7 +228,9 @@ class InGamePacketHandler extends PacketHandler{
 			$this->player->jump();
 		}
 
-		if(!$this->forceMoveSync){
+		/** @phpstan-var bool $forceMoveSync */
+		$forceMoveSync = $this->forceMoveSync;
+		if(!$forceMoveSync){
 			//TODO: this packet has WAYYYYY more useful information that we're not using
 			$this->player->handleMovement($newPos);
 		}
--- a/src/network/mcpe/handler/InGamePacketHandler.php
+++ b/src/network/mcpe/handler/InGamePacketHandler.php
@@ -228,7 +228,7 @@ class InGamePacketHandler extends PacketHandler{
 			$this->player->jump();
 		}
 
-		if($curPos->equals($this->player->getPosition())){
+		if(!$this->forceMoveSync){
 			//TODO: HACK Allow plugins to update position with PlayerToggleSneakEvent etc.
 
 			//TODO: this packet has WAYYYYY more useful information that we're not using
--- a/src/network/mcpe/handler/InGamePacketHandler.php
+++ b/src/network/mcpe/handler/InGamePacketHandler.php
@@ -228,9 +228,7 @@ class InGamePacketHandler extends PacketHandler{
 			$this->player->jump();
 		}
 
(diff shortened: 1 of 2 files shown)


dktapps:
I don't understand what this is attempting to fix.

rasu3n:
> I don't understand what this is attempting to fix.

Fixes possible falling damage and other problems that can occur when using `Player->teleport()` in PlayerToggleSneakEvent.

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -228,8 +228,12 @@ public function handlePlayerAuthInput(PlayerAuthInputPacket $packet) : bool{
 			$this->player->jump();
 		}
 
-		//TODO: this packet has WAYYYYY more useful information that we're not using
-		$this->player->handleMovement($newPos);
+		if(!$this->forceMoveSync){
+			//TODO: HACK Allow plugins to update position with PlayerToggleSneakEvent etc.
```
dktapps: I don't think this comment is necessary. It's not a hack.

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -228,7 +228,9 @@ public function handlePlayerAuthInput(PlayerAuthInputPacket $packet) : bool{
 			$this->player->jump();
 		}
 
-		if(!$this->forceMoveSync){
+		/** @phpstan-var bool $forceMoveSync */
+		$forceMoveSync = $this->forceMoveSync;
+		if(!$forceMoveSync){
```
dktapps: wtf?

rasu3n: > wtf?

This is because phpstan infers that the forceMoveSync property is false.
Would it be better to use baseline?

dktapps: Yes. Add it to `phpstan-bugs.neon`.

This is caused by https://github.com/phpstan/phpstan/issues/5727.

Issue #4894, opened 2022-03-12
Using projectiles on entities revert the item


CortexPE:
### Issue description

- Expected result: The item count should be reduced
- Actual result: The item count is retained client-side (rolled back to previous state)

### Steps to reproduce the issue

1. Put a SplashPotion or EnderPearl onto your hotbar
2. Interact with an entity (or throw it onto an entity)

### OS and versions
* PocketMine-MP: 4.2.2 (possibly occurring on older versions)
* PHP: 8.0.13
* Using JIT: no
* Server OS: Win11/Ubuntu 20.04
* Game version: Win10

### Plugins
- If you remove all plugins, does the issue still occur? Yes

### Crashdump, backtrace or other files
https://user-images.githubusercontent.com/17762324/158015004-7db5f6b6-4cab-4271-ae2f-b19d85819423.mp4

PR #4925, opened 2022-03-24
Fix ghost projectile items (#4894)

kenzeve:
## Introduction

This PR fixes ghost projectile items when using item on entity or block

### Relevant issues
#4894

## Tests

Before: https://youtu.be/2gEbOprIR0g
After: https://youtu.be/YyBmGodRhz4


dktapps:
This breaks a range of things, including (but not limited to) clicking on random blocks (e.g. signs).

PR #4930, opened 2022-03-28
Fix sugarcane behaviour on fertilizers

### LINKED ISSUE #4549: Sugarcane: non-vanilla behaviour on fertilizers
## Description
While working on https://github.com/pmmp/PocketMine-MP/issues/4545, I stumbled upon a non-vanilla behaviour of pmmp's implementation of sugarcane reacting on fertilizers.
In this video, you can see the vanilla behaviour of sugarcane: https://youtu.be/WZJLEkgbNUM
By the following line (https://github.com/pmmp/PocketMine-MP/blob/master/src/block/Sugarcane.php#L85) you can tell, that fertilizers only work on the lower, but not on the middle part. 

## Possible Solution
The Bamboo class has the method `Bamboo::seekToTop() : Bamboo` (https://github.com/pmmp/PocketMine-MP/blob/master/src/block/Bamboo.php#L136#L143), which returns the highest Bamboo block of that single plant.
The Sugarcane class could implement a similar method `Sugarcane::seekToBottom() : Sugarcane`, which would return the lowest Sugarcane block of that plant. 
That way
```php
if(!$this->getSide(Facing::DOWN)->isSameType($this)){
    $this->grow();
}
```
(https://github.com/pmmp/PocketMine-MP/blob/master/src/block/Sugarcane.php#L85#L87) could be changed to
```php
$this->seekToBottom()->grow();
```

ShockedPlot7560:
## Introduction
As explained in #4549, sugarcane, when using a fertilizer, does not behave in the same way as vanilla.  
If we use on a block of sugar cane two blocks high, it will not reach the 3 blocks as desired. This behavior only works when we use at the base.

### Relevant issues
* Fixes #4549

### Behavioural changes
* The ``grow()`` method will no longer return false when a fertilizer is used at the base and a sugarcane block is encountered.

## Tests
* Place a sugarcane
* Grow it by clicking with bone meal
* destroy the top block to get a 2 high sugarcane
* grow it by clicking on the middle or on the bottom


src/block/Sugarcane.php
```diff
@@ -49,6 +49,15 @@ public function getStateBitmask() : int{
 		return 0b1111;
 	}
 
+	private function seekToBottom() : Sugarcane{
+		$world = $this->position->getWorld();
+		$bottom = $this;
+		while(($next = $world->getBlock($bottom->position->down())) instanceof Sugarcane && $next->isSameType($this)){
```
dktapps, line 55:
```suggestion
		while($world->getBlock($bottom->position->down())->isSameType($this)){
```

src/block/Sugarcane.php
```diff
@@ -64,7 +73,7 @@ private function grow() : bool{
 				}
 				$this->position->getWorld()->setBlock($b->position, $ev->getNewState());
 				$grew = true;
-			}else{
+			}elseif(!$b instanceof Sugarcane){
```
dktapps:
```suggestion
			}elseif(!$b->isSameType($this)){
```

ShockedPlot7560: Just out of curiosity, what is the difference between the two?

dktapps: The class may be used for more than one type of block.

src/block/Sugarcane.php
```diff
@@ -49,6 +49,15 @@ public function getStateBitmask() : int{
 		return 0b1111;
 	}
 
+	private function seekToBottom() : Sugarcane{
```
dktapps: Instead of forcing this to be a `Sugarcane` instance, it would be better to return a `Vector3` or `Position`, and allow `grow()` to accept it as a parameter.

src/block/Sugarcane.php
```diff
@@ -49,36 +50,36 @@ public function getStateBitmask() : int{
...
 	}
 
-	private function seekToBottom() : Sugarcane{
+	private function seekToBottom() : Position{
 		$world = $this->position->getWorld();
-		$bottom = $this;
-		while(($next = $world->getBlock($bottom->position->down())) instanceof Sugarcane && $next->isSameType($this)){
-			$bottom = $next;
+		$bottom = $this->position;
+		while(($next = $world->getBlock($bottom->down())) instanceof Sugarcane && $next->isSameType($this)){
```
dktapps: instanceof Sugarcane is no longer necessary here

jasonw4331 (approved):

PR #4949, opened 2022-04-07
add Anti Air Jump and (Anti High Jump not work with toolbox)

SenseiTarzan:
the anticheat works 100% without problem you can test it by yourself as I said the anti high jump does not work on toolbox because it does not use the same packet system for high jump


dktapps:
We don't want anti cheat in the core. This is a job for a plugin.

SenseiTarzan:
and why not, it's easier to make it work than with a damn plugin because there are some things that are protected that it's impossible to apple like the revertMovement

dktapps:
Because it makes the experience worse for everyone unless absolutely perfect, and perfecting anti cheat involves much extra work for maintainers. That's why the preexisting movement anti-cheat and anti-fly were removed.

SenseiTarzan:
> Because it makes the experience worse for everyone unless absolutely perfect, and perfecting anti cheat involves much extra work for maintainers. That's why the preexisting movement anti-cheat and anti-fly were removed.

https://youtu.be/oF5BB5X7Udc

PR #4950, opened 2022-04-07
syntax php

SenseiTarzan:
it's more logical this way


SenseiTarzan:
I think it would be better like that because it is more logical and it uses the php standards

PR #4976, opened 2022-04-20
Add dispenser and dropper

ShockedPlot7560:
## Introduction
As redstone is a very complex concept to set up and as it is not yet implemented, some components are still unavailable in game.

Dropper and dispenser are key components in the redstone circuits, but they are also key components in the build. It is therefore necessary, even if they have no logical behavior, to implement them to be used in builds.  
This pull request, therefore, simply includes the two blocks with their storage logic and placement logic. 

### API changes
* Add Dispenser class of bloc, and his tile.
* Add DispenserInventory
* Add Dropper class of bloc and his tile
* Add DropperInventory

## Follow-up
* Implements redstone logic
* Add sound for dropper and dispenser

## Tests

https://youtu.be/GepCehAanGw


dktapps:
Not a big fan of all the code duplication here...

jasonw4331:
This PR should be refactored to be in API 5 I believe

dktapps:
There's no requirement for that, PM4 hasn't entered any feature freeze.

ShockedPlot7560:
I'm closing this PR because it no longer points to the same version and changes are needed.

PR #5003, opened 2022-04-29
Furnaces produce xp

### LINKED ISSUE #4418: Furnaces don't produce xp
### Description
Furnaces should produce xp and store that amount in them until a player collects an item from their result slot or the furnace is broken. How pointed out in https://github.com/pmmp/PocketMine-MP/issues/4417, this is currently not implemented.

### Sources
- https://minecraft.fandom.com/wiki/Furnace
- https://minecraft.fandom.com/wiki/Smelting

### Possible solutions
- The FurnaceRecipe class (https://github.com/pmmp/PocketMine-MP/blob/b4e23a57d261e9bf254af2c3a8dc1831894ee8bf/src/crafting/FurnaceRecipe.php) could get a `private float $droppedXP` variable which would contain the amount of xp smelting one of this item would return.
When the item was successfully smelted, that amount would be added to the furnace tiles `private float $storedXP` variable.
When an item is collected from the result slot or the furnace is broken, it would drop the stored amount of xp how it is explained in the sources.

IvanCraft623:
## Introduction

As it is well explained in the related issue, Furnaces produce xp and store that amount in them until a player picks up an item from their result slot or the furnace breaks.

Source: https://minecraft.fandom.com/wiki/Experience

### Relevant issues

* Fixes #4418 

### API changes

The following methods have been added to the `Item` class:
- `getSmeltingXp`

### Behavioural changes

When a player or another breaks or picks up the items in the result slot, the stored experience will be dropped.

## Tests

- Cook something and collect the result
- Break the oven without collecting the result


JustStevee:
Dktapps: Hi, I don't have time to review this PR right now because LBSG is paying me $5000 per second for support 1.13+ worlds.

Please wait, your PR will be reviewed in 150 years!

dktapps:
I don't want to add a bunch of classes just to have functions that return fixed values. That's the hell we're trying to get away from.

src/block/tile/Furnace.php
```diff
@@ -47,12 +49,24 @@ abstract class Furnace extends Spawnable implements Container, Nameable{
 	public const TAG_BURN_TIME = "BurnTime";
 	public const TAG_COOK_TIME = "CookTime";
 	public const TAG_MAX_TIME = "MaxTime";
+	public const TAG_STORED_XP = "StoredXPInt";
+
+	public static function getXpDropForItem(Item $raw, int $count) : int{
```
ShockedPlot7560, line 54: For a question of logic, this function should not be placed between constants and properties, no ?

ShockedPlot7560 (commented):
As @dktapps says, adding a lot of functions to just return a float is a bit of hell.  
The easiest way would be to put the amount of xp in the recipe for ovens, however that is not in itself a part of the recipe but a part of the item itself.  

A plausible solution would be to generate a json file represents a map id / value in BedrockData that could then be used in the item's function to retrieve its xp count.  
It is a solution that I propose, not the best nor the only one

ShockedPlot7560:
I think the best way to implement this is to add the amount of xp directly to the furnace recipes. However, this has nothing to do with crafting as such, so I'm not sure.

dktapps:
> I think the best way to implement this is to add the amount of xp directly to the furnace recipes. However, this has nothing to do with crafting as such, so I'm not sure.

I previously considered this, but there are a couple of problems and obstacles:
- The amount of XP dropped is tied to the item, not the recipe
- The data that the BedrockProtocol recipe data is generated from doesn't include any XP information, so it would have to be generated with a different method or be patched in afterwards

pandaaaBE:
Why is this pr not merged? is something wrong?

ShockedPlot7560:
The past comments is clear about the current situation of this PR

IvanCraft623:
Closing due to loss of interest and I am unable to find a way to auto-generate expereince drops values

pandaaaBE:
Why are you not creating a class (e.g. Smelting.php) and you store Fuel and XP. Fuel has maybe 25 items+ and XP has maybe 40 items+.

PR #5014, opened 2022-05-05
Added dark mode variant of the logo

inxomnyaa:
## Introduction

Currently the text in the logo is unreadable in dark mode, due to black on black colors.
I added an inverted version that only shows with dark mode using the [picture](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture) tag.
The tag is supported in all browsers except IE since ~2015
image

### Behavioural changes
*Light mode*
image
*Dark mode*
image

## Backwards compatibility

I added an exception block for old IE versions, new versions dropped support for `[if IE]` though.


inxomnyaa:
pocketmine-dark
The file, in case you want to add it to the CDN

dktapps:
What happens in IE with this change?

inxomnyaa:
> What happens in IE with this change?

The original image is displayed (the `` tag) as it will only recognize that one

SOF3 (approved):

dktapps:
We should probably just stick the logo files directly in the repo. No sense to keep bombarding cdn.pocketmine.net when we can make GitHub do the work for us ...

PR #5030, opened 2022-05-12
Fix button dropping when supporting block is broken

DuoIncure:
## Introduction
Currently buttons do not drop when the supporting block is broken, this PR is to fix that.

## Tests
Before: https://streamable.com/odyp8i

After: https://streamable.com/va7k0n


dktapps:
This will most likely be obsoleted by #4886, which is a much more comprehensive solution to this problem.

dktapps:
This will most likely be obsoleted by #4886, which is a much more comprehensive solution to this problem.

PR #5054, opened 2022-05-20
Mushroom: checking the light for placement

ShockedPlot7560:
## Introduction
I noticed that while working on the placement of the blocks and this is a part that has been moved from the original, which fix the logic for the placement of mushrooms.  
Indeed, before, mushrooms could be placed on all non-transparent blocks, but as we can see in the wiki:
> Mushrooms can be planted and grown on blocks in [light](https://minecraft.fandom.com/wiki/Light) levels below 13 and not directly underneath the sky. If planted on [mycelium](https://minecraft.fandom.com/wiki/Mycelium), [podzol](https://minecraft.fandom.com/wiki/Podzol) or [nylium](https://minecraft.fandom.com/wiki/Nylium), they can stay planted in any light level, including be planted and remain under light from sky above.

## Changes

Red/Brown mushrooms can:
- be placed in caves, on any support
- be placed with a light level lower than 13, on any support
- on the mycelium, podzol in any situation

## Tests

https://user-images.githubusercontent.com/66992287/169575333-b74d074d-389b-461c-8869-eea861dc99d8.mp4


src/block/RedMushroom.php
```diff
@@ -43,7 +43,12 @@ public function onNearbyBlockChange() : void{
 
 	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
 		$down = $this->getSide(Facing::DOWN);
-		if(!$down->isTransparent()){
+		$position = $this->getPosition();
+		$lightLevel = $position->getWorld()->getFullLightAt($position->x, $position->y, $position->z);
+		$highestBlock = $position->getWorld()->getHighestBlockAt($position->x, $position->z);
```
dktapps: This is redundant. If the light level is < 13, the block cannot be under direct sky light. In addition, placement of blocks such as glass don't bypass this limit.

src/block/RedMushroom.php
```diff
@@ -43,10 +43,12 @@ public function onNearbyBlockChange() : void{
 
 	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
 		$down = $this->getSide(Facing::DOWN);
-		if(!$down->isTransparent()){
+		$position = $this->getPosition();
+		$lightLevel = $position->getWorld()->getFullLightAt($position->x, $position->y, $position->z);
+		//TODO: nylium support
+		if(($lightLevel <= 12 && !$down->isTransparent()) || $down->isSameType(VanillaBlocks::MYCELIUM()) || $down->isSameType(VanillaBlocks::PODZOL())){
```
dktapps: ID comparisons should be used here. VanillaBlocks creates a new object for every call.

You may want to retarget this to next-major.

ShockedPlot7560: For the retarget, I don't see much use in the sense that it was in the logical continuation of #4886 what concerned the placement of the blocks.

dktapps: Fine. Nonetheless, IDs should be used for comparison, not VanillaBlocks.

ShockedPlot7560: Alright

PR #5055, opened 2022-05-20
Limit valid supporting blocks of dead bush

ShockedPlot7560:
## Introduction
I noticed that while working on the placement of the blocks and this is a part that has been moved from the original, which fix the logic for the placement of DeadBush, and add specific support.
As we can see in the wiki:
> They can be placed on any kind of [sand](https://minecraft.fandom.com/wiki/Sand), [terracotta](https://minecraft.fandom.com/wiki/Terracotta), [dirt](https://minecraft.fandom.com/wiki/Dirt), [podzol](https://minecraft.fandom.com/wiki/Podzol)

Moreover, after tests, it turns out that mycelium is also added to the list of block support for dead bush

## Changes

DeadBush can now be placed on all kind of Sand, Dirt, Podzol, Mycelium, terracotta (Hardened clay for one block)

## Tests

https://user-images.githubusercontent.com/66992287/169579723-b4c08d42-569c-454b-8764-ce1d5549c250.mp4

### DIFF
--- a/src/block/DeadBush.php
+++ b/src/block/DeadBush.php
@@ -66,10 +66,12 @@ class DeadBush extends Flowable{
 	}
 
 	private function isValidSupport(Block $block) : bool{
-		return $block->isSameType(VanillaBlocks::SAND())
-			|| $block->isSameType(VanillaBlocks::PODZOL())
-			|| $block->isSameType(VanillaBlocks::MYCELIUM())
-			|| $block->isSameType(VanillaBlocks::DIRT())
-			|| $block->isSameType(VanillaBlocks::HARDENED_CLAY());
+		$blockId = $block->getId();
+		return $blockId === BlockLegacyIds::SAND
+			|| $blockId === BlockLegacyIds::PODZOL
+			|| $blockId === BlockLegacyIds::MYCELIUM
+			|| $blockId === BlockLegacyIds::DIRT
+			|| $blockId === BlockLegacyIds::HARDENED_CLAY
+			|| $blockId === BlockLegacyIds::STAINED_HARDENED_CLAY;
 	}
 }
--- a/src/block/DeadBush.php
+++ b/src/block/DeadBush.php
@@ -66,12 +66,10 @@ class DeadBush extends Flowable{
 	}
 
 	private function isValidSupport(Block $block) : bool{
-		$id = $block->getId();
-		return $id === BlockLegacyIds::SAND
-			|| $id === BlockLegacyIds::PODZOL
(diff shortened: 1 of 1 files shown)


src/block/DeadBush.php
```diff
@@ -64,4 +64,14 @@ public function getFlameEncouragement() : int{
...
 	}
+
+	private function isValidSupport(Block $block) : bool{
+		$id = $block->getId();
+		return $id === BlockLegacyIds::SAND
+			|| $id === BlockLegacyIds::PODZOL
+			|| $id === BlockLegacyIds::MYCELIUM
+			|| $id === BlockLegacyIds::DIRT
+			|| $id === BlockLegacyIds::HARDENED_CLAY
+			|| $id === BlockLegacyIds::TERRACOTTA;
```
dktapps: terracotta and hardened clay are the same thing

src/block/DeadBush.php
```diff
@@ -64,4 +64,12 @@ public function getFlameEncouragement() : int{
 	public function getFlammability() : int{
 		return 100;
 	}
+
+	private function isValidSupport(Block $block) : bool{
+		return $block instanceof Sand
+			|| $block instanceof Podzol
+			|| $block instanceof Mycelium
+			|| $block instanceof Dirt
+			|| $block instanceof HardenedClay;
```
dktapps: instanceof should not be used for this.

src/block/DeadBush.php
```diff
@@ -64,4 +64,12 @@ public function getFlameEncouragement() : int{
 	public function getFlammability() : int{
 		return 100;
 	}
+
+	private function isValidSupport(Block $block) : bool{
+		return $block->isSameType(VanillaBlocks::SAND())
+			|| $block->isSameType(VanillaBlocks::PODZOL())
+			|| $block->isSameType(VanillaBlocks::MYCELIUM())
+			|| $block->isSameType(VanillaBlocks::DIRT())
+			|| $block->isSameType(VanillaBlocks::HARDENED_CLAY());
```
dktapps: isSameType should not be used in this way. VanillaBlocks creates a new object for every call.

In addition, this doesn't account for red sand or stained clay.

src/block/DeadBush.php
```diff
@@ -64,4 +64,14 @@ public function getFlameEncouragement() : int{
 	public function getFlammability() : int{
 		return 100;
 	}
+
+	private function isValidSupport(Block $block) : bool{
```
dktapps: This would be better named `canBeSupportedBy` for consistency with other blocks.

PR #5059, opened 2022-05-21
Update ItemSpawnEvent doc

IvanCraft623:
## Introduction
Provide more details about the event

### DIFF
--- a/src/event/entity/ItemSpawnEvent.php
+++ b/src/event/entity/ItemSpawnEvent.php
@@ -29,6 +29,7 @@ use pocketmine\entity\object\ItemEntity;
  * Called when an item is spawned or loaded.
  *
  * Some possible reasons include:
+ * - item is loaded from disk
  * - player dropping an item
  * - block drops
  * - loot of a player or entity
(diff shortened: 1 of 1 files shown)


dktapps:
I think this documentation is still misleading.

Contrary to what people might expect, this is fired when an item is loaded from disk as well as when it's spawned for the first time. Adding reasons that point towards it being created is adding to that confusion IMO.

src/event/entity/ItemSpawnEvent.php
```diff
@@ -26,6 +26,14 @@
 use pocketmine\entity\object\ItemEntity;
 
 /**
+ * Called when an item is spawned or loaded.
```
IvanCraft623, line 29: Here I have already mentioned that also when it is loaded

dktapps: Yes, but it's just two words, easily overlooked. The entire rest of the doc comment is detailing ways the item might have been created, which creates the misleading impression that that's the only time this event is called.

(Jury is out on whether it should be getting called for items loaded from disk ...)

Issue #4770, opened 2022-01-25
Add LiquidDisappearEvent


AkmalFairuz:
## Description
Add event when liquid block disappear. In my case it was useful for bedwars plugin to remove blocks placed in memory.
https://github.com/pmmp/PocketMine-MP/blob/32b07e094055c42859e6ac9d937b4b055063cf3a/src/block/Liquid.php#L301

## Alternative methods
Register blocks that extends to custom Liquid class.

dktapps:
I don't see the use case. This happens because of a source block being deleted. I don't understand why you would want to prevent the flow from going away.

dktapps:
This is a weird enough use case (which I don't see the point of) that I think it makes more sense if you stick to registering custom liquids.

PR #5078, opened 2022-05-31
Add LiquidDecayEvent and BlockDecayEvent

AkmalFairuz:
## Introduction
Event when liquid decay like LeavesDecayEvent.
I have a Replay System, but the replay system doesn't record liquid when it decays

### API changes
- Added LeavesDecayEvent
- Added BlockDecayEvent
- LeavesDecayEvent is now `extends` to BlockDecayEvent class


Matze997:
https://github.com/pmmp/PocketMine-MP/blob/stable/src/block/Liquid.php#L342

Just use BlockSpreadEvent?

AkmalFairuz:
> https://github.com/pmmp/PocketMine-MP/blob/stable/src/block/Liquid.php#L342
> 
> Just use BlockSpreadEvent?

BlockSpreadEvent not called when liquid decay

dktapps:
This implementation will only be called when the final patch of liquid decays to nothing. It won't be called for the decay steps in between.

In addition, liquid disappearing and leaf decay are two entirely different mechanics, which are only related by poor choice of naming. They should not share an event.

dktapps:
My previous comments on this topic were also never addressed: https://github.com/pmmp/PocketMine-MP/issues/4770

jasonw4331:
The decay event itself is useful for block replacements without being cancellable

PR #5089, opened 2022-06-02
PlayerExhaustEvent: popularize to EntityExhaustEvent

### LINKED ISSUE #5086: PlayerExhaustEvent: misunderstood of getPlayer method
### Issue description

PlayerExhaustEvent may misunderstood to player's hunger exhaust, and the method getPlayer is confused because it may return a non-player object if using a custom human.

https://github.com/pmmp/PocketMine-MP/blob/c87a3b054cf29282464f50b6234638b2bbe665ab/src/event/player/PlayerExhaustEvent.php#L32

Blackjack200:
## Introduction

PlayerExhaustEvent is located in \pocketmine\event\player but it not only called by player but also called in custom human entity, it isn't a suitable name.

This commit popularize PlayerExhaustEvent.

### Relevant issues

* Fixes #5086 

### API changes

- PlayerExhaustEvent -> EntityExhaustEvent

- PlayerExhastEvent::getPlayer():Human => EntityExhaustEvent::getEntity()

- PlayerExhaustEvent->player(private property) removed.

### Behavioural changes

server call EntityExhaustEvent instead of PlayerExhaustEvent now.

## Backwards compatibility

There is backwards incompatible changes.

Solution:
- Replace PlayerExhaustEvent to EntityExhaustEvent
- Call EntityExhaustEvent::getEntity() instead of PlayerExhaustEvent::getPlayer()
- Any reflection call of PlayerExhaustEvent::player should replace to EntityExhaustEvent::entity

## Tests

I write a plugin to test the event, just like

```php
public function o(EntityExhaustEvent $e) : void {
     var_dump("T");
}
```

### DIFF
--- a/src/event/entity/EntityExhaustEvent.php
+++ b/src/event/entity/EntityExhaustEvent.php
@@ -21,17 +21,16 @@
 
 declare(strict_types=1);
 
-namespace pocketmine\event\player;
+namespace pocketmine\event\entity;
 
 use pocketmine\entity\Human;
 use pocketmine\event\Cancellable;
 use pocketmine\event\CancellableTrait;
-use pocketmine\event\entity\EntityEvent;
 
 /**
  * @phpstan-extends EntityEvent<Human>
  */
-class PlayerExhaustEvent extends EntityEvent implements Cancellable{
+class EntityExhaustEvent extends EntityEvent implements Cancellable{
 	use CancellableTrait;
 
 	public const CAUSE_ATTACK = 1;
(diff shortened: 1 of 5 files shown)


dktapps:
As I recall, there's a similar problem with PlayerExperienceChangeEvent.

src/event/entity/EntityExhaustEvent.php
```diff
@@ -46,23 +45,12 @@ class PlayerExhaustEvent extends EntityEvent implements Cancellable{
 	public const CAUSE_SPRINT_JUMPING = 10;
 	public const CAUSE_CUSTOM = 11;
 
-	/** @var Human */
-	protected $player;
-
 	public function __construct(
-		Human $human,
+		protected Human $human,
```
dktapps, line 49: It doesn't make any sense to promote this. The promoted property is never used.

dktapps (commented):
Concept is OK, but there's a few minor nits.

dktapps:
Closing due to lack of response to review comments.

PR #5095, opened 2022-06-05
[Fix] ItemEntity bounces on Slime Block

### LINKED ISSUE #4830: ItemEntity doesn't bounce on slime block
### Issue description
Dropping some items on slime block
- Expected result: Items bounce back
- Actual result: Items does not bounce back

1. Dropping some items on slime block
2. Watching its movement

### OS and versions

* PocketMine-MP: 4.0.2
* PHP: 8.0.15
* Using JIT: no 
* Server OS: Windows
* Game version: 1.18.10 Android/Win10

### Plugins

None
- If you remove all plugins, does the issue still occur? N/A
- If the issue is **not** reproducible without plugins: N/A
  - Have you asked for help on our forums before creating an issue? No
Checking if the entity is Item Entity too?

Swift-Strider:
## Introduction

This PR fixes items previously not bouncing on slime blocks.

### Relevant issues

* Fixes #4830

## Changes
Items now bounce on slime blocks.
### API changes

No significant API changes, as `onHitGround()` was only overriden. It may break plugins that depend on ItemEntity's previous behavior of landing without calling `$fallBlock->onEntityLand()`.

### Behavioural changes

Items now bounce on slime blocks.

## Backwards compatibility

This is a backwards compatible change.
## Follow-up
Perhaps moving generic physics code and applying that to ItemEntity, armor_stand, etc.

## Tests
I haven't added any tests.

### DIFF
--- a/src/entity/object/ItemEntity.php
+++ b/src/entity/object/ItemEntity.php
@@ -186,6 +186,16 @@ class ItemEntity extends Entity{
 		return true;
 	}
 
+	protected function onHitGround() : ?float{
+		$fallBlockPos = $this->location->floor();
+		$fallBlock = $this->getWorld()->getBlock($fallBlockPos);
+		if(count($fallBlock->getCollisionBoxes()) === 0){
+			$fallBlockPos = $fallBlockPos->down();
+			$fallBlock = $this->getWorld()->getBlock($fallBlockPos);
+		}
+		return $fallBlock->onEntityLand($this);
+	}
+
 	public function canSaveWithChunk() : bool{
 		return !$this->item->isNull() && parent::canSaveWithChunk();
 	}
--- a/src/entity/object/ItemEntity.php
+++ b/src/entity/object/ItemEntity.php
@@ -39,6 +39,7 @@ use pocketmine\network\mcpe\protocol\AddItemActorPacket;
 use pocketmine\network\mcpe\protocol\types\entity\EntityIds;
 use pocketmine\network\mcpe\protocol\types\inventory\ItemStackWrapper;
 use pocketmine\player\Player;
+use function count;
 use function max;
 
 class ItemEntity extends Entity{


jasonw4331 (changes requested):
I think the code for calling `Block::onEntityLand()` should be moved from Living to Entity while leaving the fall damage code in Living instead of creating boilerplate for this use case.

dktapps:
> I think the code for calling `Block::onEntityLand()` should be moved from Living to Entity while leaving the fall damage code in Living instead of creating boilerplate for this use case.

You might be right about this, but it might have other unintended consequences, e.g. bouncy XP orbs, bouncy arrows, etc.

ColinHDev:
And limiting the affected entities in the [`Slime` class](https://github.com/pmmp/PocketMine-MP/blob/next-minor/src/block/Slime.php) itself?

The `Block::onEntityLand()` already has it in its name: It's supposedly called once an entity lands on it. So every block extending that method should decide for themselves to which entities they want to limit their behaviour.

src/entity/object/ItemEntity.php
```diff
@@ -186,6 +187,16 @@ protected function applyDragBeforeGravity() : bool{
 		return true;
 	}
 
+	protected function onHitGround() : ?float{
+		$fallBlockPos = $this->location->floor();
+		$fallBlock = $this->getWorld()->getBlock($fallBlockPos);
+		if(count($fallBlock->getCollisionBoxes()) === 0){
+			$fallBlockPos = $fallBlockPos->down();
+			$fallBlock = $this->getWorld()->getBlock($fallBlockPos);
```
ShockedPlot7560, line 195: So, if the down block has no collision too, it will stop here, no ?

dktapps:
> And limiting the affected entities in the [`Slime` class](https://github.com/pmmp/PocketMine-MP/blob/next-minor/src/block/Slime.php) itself?
> 
> The `Block::onEntityLand()` already has it in its name: It's supposedly called once an entity lands on it. So every block extending that method should decide for themselves to which entities they want to limit their behaviour.

Yeah, this is probably the correct fix.

However this PR is very stale so I'm going to close it.

Issue #2506, opened 2018-10-31
Namespaced plugin identifiers


SOF3:
## Introduction
Currently, plugins are uniquely identified by their names. This is not a good solution, because plugin names tend to be short and easily clash. A more detailed *unique* identifier for plugins would be desirable.

### OS level
Windows:
- User-friendly applications are listed in the Start Menu. They are grouped using folders. Folder name can be company name or software name.
- Most applications are available in the `Program Files`/`Program Files (x86)` folders. Similar grouping as Start Menu.

Android:
- Applications are uniquely identified by the package name, e.g. `net.pocketmine.server`.
- Apps are listed in the app menu, but name clashes don't matter.

### Language level
Java (maven): each artifact has a group ID and an artifact ID. Apart from the earliest Apache libraries, most libraries are in the format `package.name:artifact-name`.
JavaScript (npm): packages are identified by name, but they are moving towards namespaced identifiers too.
PHP (composer/packagist): packages are identified by GitHub username + GitHub repo name. Note that this forces the use of GitHub, and for whatever political or psychological reasons, some people do not like using services under Microsoft.

## Comparison
Most PocketMine plugins can be found on GitHub, so an approach similar to composer's could be appropriate. Since the same user is unlikely to have multiple plugins with the same name (\*cough\* @TheFixerDevelopment), author name + plugin name could be sufficient. However, it is quite common to see multiple plugins in the same repo. This might lead to some inconsistency.

<details>
<summary>Reference data from Poggit</summary>
Number of repos that have `x` plugin/virion projects in the same repo
```
+----+------+
| n  | freq |
+----+------+
|  1 | 2203 |
|  2 |   98 |
|  3 |   13 |
|  4 |   13 |
|  5 |    2 |
|  6 |    6 |
|  9 |    2 |
| 10 |    2 |
| 11 |    1 |
| 12 |    8 |
| 15 |    1 |
| 16 |    1 |
| 26 |    1 |
| 32 |    1 |
| 35 |    1 |
| 41 |    4 |
| 44 |    2 |
| 45 |    1 |
| 48 |    1 |
+----+------+
```

Number of projects that don't have plugin.yml in the repo root directory: 895/3150 = 28.41%
</details>

Since PocketMine's developer community is relatively small, there is no need to worry about author name clashes. (There aren't many people with a short name like `SOFe`) As a result, there is no need to require a domain name (e.g. `io.github.pmmp`, `com.gmail.sofe2038`, etc.).

[...]

PR #5101, opened 2022-06-07
Initial `getPlugin` updates

DuoIncure:
## Introduction
This PR implements a suggestion made previously by SOF3 and further agreed by Dylan about `getPlugin` accepting class strings instead of plugin names as the parameter.

### Relevant issues
#2506

### API changes
`getPlugin` now accepts a class string instead of plugin names

## Backwards compatibility
Any plugins that use `getPlugin` with hardcoded plugin names would have to change to use class strings instead.

## Tests
Created 2 test plugins:
1 with base plugin files (plugin.yml, Test1.php extending PluginBase but nothing inside `onEnable`)
The second has the same file structure however the following code was placed inside of `onEnable` :

```
protected function onEnable() : void {
		  $plugin = $this->getServer()->getPluginManager()->getPlugin(Test1::class);

		  $this->getLogger()->info($plugin->getDescription()->getDescription());
}
```

The result was the description of Test1 being displayed inside the console.

### DIFF
--- a/src/plugin/PluginManager.php
+++ b/src/plugin/PluginManager.php
@@ -97,6 +97,11 @@ class PluginManager{
 		}
 	}
 
+	/**
+	 * @param string $class
+	 * @phpstan-param class-string<Plugin> $class
+	 * @return Plugin|null
+	 */
 	public function getPlugin(string $class) : ?Plugin{
 		if(!is_a($class, Plugin::class, true)) return null;
 		return $this->plugins[$class];
--- a/src/plugin/PluginManager.php
+++ b/src/plugin/PluginManager.php
@@ -97,12 +97,9 @@ class PluginManager{
 		}
 	}
 
-	public function getPlugin(string $name) : ?Plugin{
-		if(isset($this->plugins[$name])){
-			return $this->plugins[$name];
-		}
-
-		return null;
+	public function getPlugin(string $class) : ?Plugin{
+		if(!is_a($class, Plugin::class, true)) return null;
+		return $this->plugins[$class];
 	}
 
 	public function registerInterface(PluginLoader $loader) : void{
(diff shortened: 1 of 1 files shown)


dktapps:
Actually my suggestion was to supplement the plugin name with the class path to allow PHPStan to infer the return type via generics, not to use the class name directly as the identifier. This is because using the class name exclusively makes it impossible for two different plugins to use the same main class, which might be desirable in some circumstances (e.g. frameworks)

SOF3:
I think this problem is better solved by an approach in the style of #4471. Main classes should not be the API entrypoint of a plugin, because that implies god objects. In my opinion, getting a `Plugin` instance should only be used when you want to know about the generic attributes of a plugin (e.g. whether it is enabled, or the name and version, etc). If you want to access the actual specific API object, it should not go through the PluginManager, but access the class directly, because API class is not necessarily (and in my opinion, should not be) the main class. This is achieved either through static variables (currently the most common approach), or through a generic type-based storage (such as #4471).

dktapps:
> I think this problem is better solved by an approach in the style of #4471. Main classes should not be the API entrypoint of a plugin, because that implies god objects. In my opinion, getting a `Plugin` instance should only be used when you want to know about the generic attributes of a plugin (e.g. whether it is enabled, or the name and version, etc). If you want to access the actual specific API object, it should not go through the PluginManager, but access the class directly, because API class is not necessarily (and in my opinion, should not be) the main class. This is achieved either through static variables (currently the most common approach), or through a generic type-based storage (such as #4471).

This isn't a solution for this problem. This is just solving this problem as a side effect of making a different, massive change which hasn't been agreed on.

SOF3:
> > I think this problem is better solved by an approach in the style of #4471. Main classes should not be the API entrypoint of a plugin, because that implies god objects. In my opinion, getting a `Plugin` instance should only be used when you want to know about the generic attributes of a plugin (e.g. whether it is enabled, or the name and version, etc). If you want to access the actual specific API object, it should not go through the PluginManager, but access the class directly, because API class is not necessarily (and in my opinion, should not be) the main class. This is achieved either through static variables (currently the most common approach), or through a generic type-based storage (such as #4471).
> 
> This isn't a solution for this problem. This is just solving this problem as a side effect of making a different, massive change which hasn't been agreed on.

The whole point of wanting to get a plugin instance by class name is because you want an instance of that class in the context of the server. That's exactly the whole problem that #4471 is trying to address.

Trying to get the main class of a plugin and trying to get an instance of a singleton-in-context are two different things. This pull request is trying to solve these two problems together with the assumption that the main class is the singleton-in-context that users want to get, and this does not make sense.

dktapps:
Closing this due to unresolved debate and lack of PR activity.

PR #5102, opened 2022-06-07
Add sendToast method, API to send toast with ToastRequestPacket

yuyaprgrm:
## Introduction

Since Bedrock 1.19.0, there is a ToastRequestPacket but there is no API to send toast.
### Relevant issues

No

### API changes

add `Player::sendToast(string $title, string $body): void`

### Behavioural changes

No

## Backwards compatibility

There is no BC break.

## Tests

```php
$player->sendToast("title", "body");
```
image

### DIFF
--- a/src/network/mcpe/NetworkSession.php
+++ b/src/network/mcpe/NetworkSession.php
@@ -1043,7 +1043,7 @@ class NetworkSession{
 		$this->sendDataPacket(EmotePacket::create($from->getId(), $emoteId, EmotePacket::FLAG_SERVER));
 	}
 
-	public function onToast(string $title, string $body){
+	public function onToast(string $title, string $body) : void{
 		$this->sendDataPacket(ToastRequestPacket::create($title, $body));
 	}
 
--- a/src/network/mcpe/NetworkSession.php
+++ b/src/network/mcpe/NetworkSession.php
@@ -1043,7 +1043,7 @@ class NetworkSession{
 		$this->sendDataPacket(EmotePacket::create($from->getId(), $emoteId, EmotePacket::FLAG_SERVER));
 	}
 
-	public function onToast(string $title, string $body) : void{
+	public function onToastNotification(string $title, string $body) : void{
 		$this->sendDataPacket(ToastRequestPacket::create($title, $body));
 	}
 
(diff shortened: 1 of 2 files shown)


ghost:
It seems to me that it would be more logical to `addToast`, because Toasts queue up if, e.g.  send more than two.

rac14:
Looks cool

SOF3:
> It seems to me that it would be more logical to `addToast`, because Toasts queue up if, e.g. send more than two.

Chat messages queue up in the chat pane too.

ghost:
> 

But the chat does not have such a significant delay as Toasts...

dktapps:
I'm not sure `sendToast()` is a clear enough name. It doesn't translate very well.

That said, there is a precedent in the Android API, so I'm unsure if it needs to be changed. Thoughts?

dktapps:
At the least, there should be a small docblock that describes what the function does, since the function's name is not very intuitive.

alvin0319:
> I'm not sure `sendToast()` is a clear enough name. It doesn't translate very well.
> 
> That said, there is a precedent in the Android API, so I'm unsure if it needs to be changed. Thoughts?

I think it would be better to change method name to `sendNotification`, since toast is used to notify users such as excessive game playing, achievements, etc.

yuyaprgrm:
I want to take a survey in this PR.

## Should `send` be used for this method name or not?
At first, I thought `send` is good naming because it was used in all sending text method like `sendMessage`, `sendTitle`, `sendTip` and `sendPopup`, but I found that toasts have different behavior than these methods above.
> But the chat does not have such a significant delay as Toasts...

Yes, it delays. If there is already a toast text shown in game screen, the new toast waits for the former toast to disappear.
I may change method name or just add a description in docblock.

## Which is better name for this method?
As Dylan pointed out, `sendToast` may unclear name to describe its functionality. 
Also, alvin gives a good name for this method, `sendNotification`.

I want to know which is better name for this method. Or add other idea.
🅰️ `sendToast`
🅱️ `sendNotification`

dktapps:
I think `sendToastNotification()` or `sendToastMessage()` makes the most sense, since both `sendToast()` and `sendNotification()` are ambiguous at first glance.

Nerahikada:
I like `sendToast()` + docblock.
As dylan said, there is precedent in the Android API and I don't think it needs to be changed.

IvanCraft623:
I think `sendToast()` + docblock or `sendToastPopup()` are ok, would be descriptive and simple.

dktapps (approved):

PR #5118, opened 2022-06-28
Improve performance of `InventoryManager::syncCreative()`

Nerahikada:
## Introduction
I found that it is taking a long time to change the game mode of the player.
This is due to the time taken to generate the CreativeContentPacket.
This PR improves the performance of `InventoryManager::syncCreative()` by caching the packet.

### Relevant issues
N/A

### API changes
N/A

## Backwards compatibility
N/A

## Tests
Reduced from about 100 milliseconds to 0.7 milliseconds.
```php
$startTime = microtime(true);
for($i = 0; $i < 10; ++$i){
	$_player->setGamemode(GameMode::SURVIVAL());
	$_player->setGamemode(GameMode::CREATIVE());
}
$time = microtime(true) - $startTime;
$_player->sendMessage(round($time * 10 ** 3, 1) . "ms");
```

### DIFF
--- a/src/inventory/CreativeInventory.php
+++ b/src/inventory/CreativeInventory.php
@@ -25,6 +25,9 @@ namespace pocketmine\inventory;
 
 use pocketmine\item\Durable;
 use pocketmine\item\Item;
+use pocketmine\network\mcpe\convert\TypeConverter;
+use pocketmine\network\mcpe\protocol\CreativeContentPacket;
+use pocketmine\network\mcpe\protocol\types\inventory\CreativeContentEntry;
 use pocketmine\utils\SingletonTrait;
 use Webmozart\PathUtil\Path;
 use function file_get_contents;
@@ -35,6 +38,8 @@ final class CreativeInventory{
 
 	/** @var Item[] */
 	private array $creative = [];
+
+	private ?CreativeContentPacket $packetCache = null;
 
 	private function __construct(){
 		$creativeItems = json_decode(file_get_contents(Path::join(\pocketmine\BEDROCK_DATA_PATH, "creativeitems.json")), true);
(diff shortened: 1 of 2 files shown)


src/inventory/CreativeInventory.php
```diff
@@ -93,10 +101,26 @@ public function remove(Item $item) : void{
...
 			unset($this->creative[$index]);
+			$this->packetCache = null;
 		}
 	}
 
 	public function contains(Item $item) : bool{
 		return $this->getItemIndex($item) !== -1;
 	}
+
+	public function getContentPacket() : CreativeContentPacket{
```
dktapps: This should not be in the API.

src/network/mcpe/InventoryManager.php
```diff
@@ -379,11 +381,21 @@ public function syncSelectedHotbarSlot() : void{
...
+			$typeConverter = TypeConverter::getInstance();
+			$nextEntryId = 1;
+
+			$this->creativeContentsCache = CreativeContentPacket::create(
+				array_map(function(Item $item) use($typeConverter, &$nextEntryId) : CreativeContentEntry{
+					return new CreativeContentEntry($nextEntryId++, $typeConverter->coreItemStackToNet($item));
+				}, $creativeInventory->getAll())
+			);
+
+			$creativeInventory->setChanged(false);
```
dktapps, line 396: If the creative inventory is modified while there are players online, this will cause the changes to be synced only for the first one.

dktapps (changes requested):
This method wastes memory and the cache invalidation doesn't work correctly.

Nerahikada:
I don't have enough time to work on this. I will open the issue for posterity.

dktapps:
> I don't have enough time to work on this. I will open the issue for posterity.

Thank you.

dktapps:
Related PR: #5703

PR #5122, opened 2022-06-30
Save and read CustomNameAlwaysVisible tag

### LINKED ISSUE #4669: Humans (or) and Entities lose attributes after rebooting the server.
### Issue description

- Expected result: After rebooting, all attributes of entities and humans should save.
- Actual result: After rebooting, all entities should have attributes...

1. Make a simple plugin for creating Entity such as Human: `$npc = new Human($sender->getServer()->getPlayerExact($sender->getName())->getLocation(), $sender->getServer()->getPlayerExact($sender->getName())->getSkin());`

2. Give for your Human nameTag: `$npc->setNameTag("7776661111");`

3. Set nameTag of your Human always visible: `$npc->setNameTagAlwaysVisible();`

4. Restart server.

5. It happened that should have been expects. nameTag is not always displayed, only when you hover on it.

### OS and versions

* PocketMine-MP: 4.0.3
* PHP: 8.0
* Using JIT: no
* Server OS: Linux Mint

IvanCraft623:
## Introduction

The CustomNameAlwaysVisible tag is not being read/saved in the entities, causing data loss.

### Relevant issues

* Fixes #4669
## Tests
```php
$human = new Human($location, $skin);
$human->setNameTag("mctestDylan");
$human->setNameTagAlwaysVisible();
```


jasonw4331 (dismissed):

ColinHDev:
Do we follow Mojang's entity save format within PM? Just asking because a save key with a comparable functionality does not exist according to the [wiki](https://minecraft.fandom.com/wiki/Entity_format).

dktapps:
> Do we follow Mojang's entity save format within PM? Just asking because a save key with a comparable functionality does not exist according to the [wiki](https://minecraft.fandom.com/wiki/Entity_format).

That's precisely why I hesitated to merge it.

ShockedPlot7560:
I don't think this should be an obstacle to merge the PR. Could it have any impact on PM at some point?

dktapps:
> I don't think this should be an obstacle to merge the PR. Could it have any impact on PM at some point?

As long as we're sure Mojang won't introduce a tag with the same name and a different value type...

dktapps:
I think we need to think about standard ways to store PM-specific data before merging this, so this doesn't become a problem in the future. (Really should've done this years ago before plugin devs were able to start injecting all kinds of crap into the NBT...)

ipad54:
> I think we need to think about standard ways to store PM-specific data before merging this, so this doesn't become a problem in the future. 

The first thing that comes to mind is to have a `PMMPExtraData` Compound tag that will be written into an entity NBT and also `loadExtraData(CompoundTag $nbt) : void`, `saveExtraData() : CompoundTag` methods. Any custom entity properties will be stored inside this tag. What do you think?

dktapps:
@ipad54 I agree with putting PM custom data inside a subtag. Not sure if it's necessary to add extra methods for it though since it won't be needed in most cases.

ipad54:
Extra methods can be pretty useful in the future if we want to store more than one custom property, especially in different entity classes.

dktapps:
I feel like it's redundant for one property. We can add more stuff later on if necessary.

PR #5130, opened 2022-07-05
Check the item that was clicked on the light block

Gaprix:
## Introduction
This Pull Request checks with what item the player is trying to change the light level of the light block. This will prevent players with a malicious client from changing the light level without having a light block in their hand.

### Behavioural changes
Light block lighting level can now only be changed if the player has a light block in hand.

## Backwards compatibility
No BC break

## Tests
```
	if($event->getBlock() instanceof Light) {
		$result1 = $event->getBlock()->onInteract(VanillaItems::AIR(), 0, new Vector3(0, 0, 0));
		$result2 = $event->getBlock()->onInteract(VanillaBlocks::LIGHT()->asItem(), 0, new Vector3(0, 0, 0));
		var_dump($result1); // bool(false)
		var_dump($result2); // bool(true)
	}
}
```


dktapps:
I don't think this is really necessary. They can also delete and replace the block if the world is mutable. If you don't want players messing with your lights, you should just use spawn protection like any other block.

Gaprix:
Some servers use protection that allows players to interact with blocks. However, I don't think they will expect the player to be able to interact with the light block.

dktapps:
Servers should be whitelisting only the blocks they want to be interactable (e.g. doors, buttons). This isn't a problem unique to light blocks.

Gaprix:
For the developers of plugins, it is unlikely to be obvious that it is necessary to protect the invisible block from the interaction of players with it.

Gaprix:
But in the case of using a whitelist, it really doesn't matter

dktapps:
If developers are blacklisting blocks one by one, they are going to get nasty surprises in many cases, whether it's for light blocks, or for any other newly added block. The correct solution to such problems is using spawn protection that whitelists blocks.

src/block/Light.php
```diff
@@ -59,12 +59,16 @@ public function setLightLevel(int $level) : self{
 	public function canBeReplaced() : bool{ return true; }
 
 	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
-		$this->level = $this->level === self::MAX_LIGHT_LEVEL ?
-			self::MIN_LIGHT_LEVEL :
-			$this->level + 1;
+		if($item->getBlock() instanceof Light){
```
dktapps: instanceof should not be used. Use `isSameType()` or compare `getTypeId()`.

dktapps:
Sorry, this got closed due to deleting `modern-world-support`, please resubmit it.

PR #5133, opened 2022-07-05
Set the correct max stack size for spyglass

Gaprix:
## Introduction
Fixes the maximum stack size of spyglass

### API changes
Spyglass now has a separate class

### Behavioural changes
Now a player cannot have more than 1 spyglass in the slot

## Tests
```
/give PlayerName spyglass
```
before:
```
[PlayerName: Given Spyglass (spyglass) * 64 to PlayerName]
```
after:
```
[PlayerName: Given Spyglass (spyglass) * 1 to PlayerName]
```

### DIFF
--- a/src/item/ItemFactory.php
+++ b/src/item/ItemFactory.php
@@ -98,7 +98,6 @@ class ItemFactory{
 		$this->register(new GoldenApple(new IID(Ids::GOLDEN_APPLE), "Golden Apple"));
 		$this->register(new GoldenAppleEnchanted(new IID(Ids::ENCHANTED_GOLDEN_APPLE), "Enchanted Golden Apple"));
 		$this->register(new GoldenCarrot(new IID(Ids::GOLDEN_CARROT), "Golden Carrot"));
-		$this->register(new Spyglass(new IID(Ids::SPYGLASS), "Spyglass"));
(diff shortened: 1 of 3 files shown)


src/item/ItemFactory.php
```diff
@@ -98,6 +98,7 @@ public function __construct(){
 		$this->register(new GoldenApple(new IID(Ids::GOLDEN_APPLE), "Golden Apple"));
 		$this->register(new GoldenAppleEnchanted(new IID(Ids::ENCHANTED_GOLDEN_APPLE), "Enchanted Golden Apple"));
 		$this->register(new GoldenCarrot(new IID(Ids::GOLDEN_CARROT), "Golden Carrot"));
+		$this->register(new Spyglass(new IID(Ids::SPYGLASS), "Spyglass"));
```
dktapps: Not clear why this is up here...

Gaprix: Because of the alphabetical order of class names

dktapps: This isn't in alphabetical order. S comes further down the alphabet.

Gaprix: Oh, I realized my mistake. I'll fix it now

PR #5136, opened 2022-07-06
Explosions can destroy unbreakable blocks like bedrock

SunsetTadpole78:
## Introduction
currently on pocketmine, explosions break unbreakable blocks like bedrock, this pull request allows to solve this problem or to force this breaking of blocks

## Changes
- patch of the bug which allows to break unbreakable blocks
- addition of a condition in the ExplosionPrimeEvent to be able to break the unbreakable blocks (allows to force the explosion)

## Tests
Before:
before
After:
after

### DIFF
--- a/src/event/entity/ExplosionPrimeEvent.php
+++ b/src/event/entity/ExplosionPrimeEvent.php
@@ -40,15 +40,15 @@ class ExplosionPrimeEvent extends EntityEvent implements Cancellable{
 
 	protected float $force;
 	private bool $blockBreaking = true;
-	private bool $canBreakUnbreakableBlocks;
+	private bool $forceToBreakAll;
 
-	public function __construct(Entity $entity, float $force, bool $canBreakUnbreakableBlocks){
+	public function __construct(Entity $entity, float $force, bool $forceToBreakAll){
 		if($force <= 0){
 			throw new \InvalidArgumentException("Explosion radius must be positive");
 		}
 		$this->entity = $entity;
 		$this->force = $force;
-		$this->canBreakUnbreakableBlocks = $canBreakUnbreakableBlocks;
+		$this->forceToBreakAll = $forceToBreakAll;
 	}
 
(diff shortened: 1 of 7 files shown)


src/event/entity/ExplosionPrimeEvent.php
```diff
@@ -60,12 +61,23 @@ public function setForce(float $force) : void{
 		}
 		$this->force = $force;
 	}
-
```
ShockedPlot7560: It is an unnecessary change

src/world/Explosion.php
```diff
@@ -129,6 +129,10 @@ public function explodeA() : bool{
 									if(!isset($this->affectedBlocks[World::blockHash($vBlockX, $vBlockY, $vBlockZ)])){
 										$_block = $this->world->getBlockAt($vBlockX, $vBlockY, $vBlockZ, true, false);
 										foreach($_block->getAffectedBlocks() as $_affectedBlock){
+
+											if(!$canBreakAll && !$_affectedBlock->getBreakInfo()->isBreakable())
```
ShockedPlot7560: Use the block form instead of the inline ?
```php
if(...){
     ...
}
```

src/event/entity/ExplosionPrimeEvent.php
```diff
@@ -38,16 +38,17 @@
 class ExplosionPrimeEvent extends EntityEvent implements Cancellable{
 	use CancellableTrait;
 
-	/** @var float */
-	protected $force;
+	protected float $force;
```
ShockedPlot7560: This is an unrelated change

dktapps:
Bedrock isn't indestructible. It just has a very (very) high blast resistance (18,000,000). The point is that small amounts of explosive can't destroy it, but a powerful enough explosive might be able to.

TNT in PM isn't able to destroy bedrock, so I don't see why this PR is necessary.

PR #5137, opened 2022-07-06
Added spyglass using animation

ipad54:
## Introduction
Added spyglass using animation

## Tests
Screenshot_24


dktapps:
of course, I don't know how I missed this ...

PR #5140, opened 2022-07-06
Check the item that was clicked on the light block

Gaprix:
## Introduction
This Pull Request checks with what item the player is trying to change the light level of the light block. This will prevent players with a malicious client from changing the light level without having a light block in their hand.

### Behavioural changes
Light block lighting level can now only be changed if the player has a light block in hand.

## Backwards compatibility
No BC break

## Tests
```
	if($event->getBlock() instanceof Light) {
		$result1 = $event->getBlock()->onInteract(VanillaItems::AIR(), 0, new Vector3(0, 0, 0));
		$result2 = $event->getBlock()->onInteract(VanillaBlocks::LIGHT()->asItem(), 0, new Vector3(0, 0, 0));
		var_dump($result1); // bool(false)
		var_dump($result2); // bool(true)
	}
}
```


dktapps:
Closing this since I don't think it really makes any sense. A world protection plugin ought to suffice, and it doesn't make sense to special-case one block.

PR #5149, opened 2022-07-10
Do not always make the coral dead immediately after placement

Gaprix:
## Introduction
At the moment, after placing the coral, it will immediately turn out to be dead. This does not match the behavior in a single player game. This PR corrects this by making the behavior of corals the same as that of coral blocks.

### API changes
The hasWater function has been added to the BaseCoral class, which checks for the existence of water around the coral.

### Behavioural changes
The coral will die only after some time out of the water. The exceptions are the coral wall and the coral floor, which die immediately.

## Backwards compatibility
No BC break

## Tests
https://youtu.be/anggFHTXkfE

### DIFF
--- a/src/block/BaseCoral.php
+++ b/src/block/BaseCoral.php
@@ -59,7 +59,7 @@ abstract class BaseCoral extends Transparent{
 
 	public function isSolid() : bool{ return false; }
 
-	public function isCoveredWithWater() : bool{
+	protected function isCoveredWithWater() : bool{
 		$world = $this->position->getWorld();
 
 		$hasWater = false;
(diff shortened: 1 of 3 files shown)


dktapps:
I don't see how this is achieving your objective ... you're checking if the coral should be dead already within place().

Gaprix:
Fan corals cannot live without water and should be placed dead immediately

Gaprix:
The rest of the corals must live for some time without water, which is implemented by my Pull Request

src/block/BaseCoral.php
```diff
@@ -65,6 +59,21 @@ public function isAffectedBySilkTouch() : bool{
 
 	public function isSolid() : bool{ return false; }
 
+	public function hasWater() : bool{
```
dktapps: This method is being exposed to the API, but I'm not sure its name is clear enough. I'm also doubtful it should be public, considering that it's intended for internal usage.

Gaprix: Would "isFilledWithWater" be a better name?

dktapps: Not really, since it's looking around the block.

PR #5153, opened 2022-07-13
remove redundant check

UnknownOre:
## Introduction
EntityEffectRemoveEvent already check for the effect duration when it gets cancelled

### DIFF
--- a/src/entity/effect/EffectManager.php
+++ b/src/entity/effect/EffectManager.php
@@ -89,7 +89,7 @@ class EffectManager{
 			$ev = new EntityEffectRemoveEvent($this->entity, $effect);
 			$ev->call();
 			if($ev->isCancelled()){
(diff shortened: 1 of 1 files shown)


src/entity/effect/EffectManager.php
```diff
@@ -89,7 +89,7 @@ public function remove(Effect $effectType) : void{
 			$ev = new EntityEffectRemoveEvent($this->entity, $effect);
 			$ev->call();
 			if($ev->isCancelled()){
-				if($hasExpired && !$ev->getEffect()->hasExpired()){ //altered duration of an expired effect to make it not get removed
+				if($hasExpired){ //altered duration of an expired effect to make it not get removed
```
dktapps: I think the entire condition can be removed.

dktapps:
the `hasExpired` variable now appears to be unused

PR #5155, opened 2022-07-13
Add API functions for Fly speed.

TheDataLioness:
## Introduction
Since the new 1.19.10 update added the Walk and Fly speed feature to the new AbilityPacket.
These values were hardcoded

### API changes
Added $player->setFlySpeed();
Added $player->getFlySpeed();
Added $player->resetFlySpeed();

Added $player->setWalkSpeed();
Added $player->getWalkSpeed();
Added $player->resetWalkSpeed();

## Tests

https://user-images.githubusercontent.com/34657342/178804701-eee1ee4c-cc99-4d72-9b1f-bf3c0a47aa89.mp4


dktapps:
I'm not clear on what the value of `setWalkSpeed` is, since we already have an attribute-controlled `setMovementSpeed` which influences this. I'm open to reasoning.

TheDataLioness:
Well I've removed that function already because of that reason.
But I forgot to push it.
I'm on vacation right now so haven't really had the time to commit the changes.

src/player/Player.php
```diff
@@ -463,7 +463,12 @@ public function getFlySpeed() : float{
 		return $this->flySpeed;
 	}
 
-	public function setWalkSpeed(float $walkSpeed = 0.1) : void{
+	public function resetFlySpeed(): void
+	{
+		$this->setFlySpeed(0.05);
```
Crasher508: I think const are better than hard code

dktapps: +1 for this. This would also remove the need for resetFlySpeed(), since you would be able to just do `setFlySpeed(DEFAULT_FLY_SPEED)` or similar.

src/network/mcpe/NetworkSession.php
```diff
@@ -806,8 +806,7 @@ public function syncAbilities(Player $for) : void{
 			$isOp ? PlayerPermissions::OPERATOR : PlayerPermissions::MEMBER,
 			$for->getId(),
 			[
-				//TODO: dynamic flying speed! FINALLY!!!!!!!!!!!!!!!!!
-				new UpdateAbilitiesPacketLayer(UpdateAbilitiesPacketLayer::LAYER_BASE, $boolAbilities, 0.05, 0.1),
+				new UpdateAbilitiesPacketLayer(UpdateAbilitiesPacketLayer::LAYER_BASE, $boolAbilities, $for->getFlySpeed(), $for->getWalkSpeed()),
```
dktapps, line 809: `getWalkSpeed()` doesn't exist. Please revert it to its original value, since we're not going to use it.

src/player/Player.php
```diff
@@ -452,6 +453,20 @@ public function isFlying() : bool{
...
+		$this->flySpeed = $flySpeed;
+		$this->getNetworkSession()->syncAbilities($this);
+	}
+
+	public function getFlySpeed() : float{
+		return $this->flySpeed;
+	}
+
+	public function resetFlySpeed() : void
+	{
```
dktapps, line 466: Braces should not be placed on a newline.

dktapps (changes requested):

AzaleeX:
ok but there won't be any chunk bug on some "big" server ?

jasonw4331:
@TheDataLioness any updates for this PR? If not it will be closed when API 5 releases

TheDataLioness:
@jasonwynn10 haven't touched PM in ages.

dktapps:

PR #5156, opened 2022-07-13
EnumTrait short for random member of enum

xxAROX:
### API changes
```diff
+ EnumTrait::getRandom()
```


dktapps:
I don't see why this needs to be in the core code.

PR #5158, opened 2022-07-13
Play burp sound when consuming a FoodSource

### LINKED ISSUE #5157: Missing sound when eating cake
### Issue description
Eating a cake slice in vanilla causes the character to make a burp sound, which is missing in PM.

1. Place a cake
2. Eat a slice
3. Notice no sound is made

### OS and versions

* PocketMine-MP: 4.6.0 / 5.0.0-ALPHA1

IvanCraft623:
## Introduction
Play the sound of burping when consuming a FoodSource

### Relevant issues

* Fixes #5157


dktapps:
On further inspection I think this is missing from more stuff than just cake.

jasonw4331 (approved):

src/block/Cake.php
```diff
@@ -99,8 +100,9 @@ public function getDropsForCompatibleTool(Item $item) : array{
 	}
 
 	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
-		if($player !== null){
-			return $player->consumeObject($this);
+		if($player !== null && $player->consumeObject($this)){
+			$this->position->getWorld()->addSound($this->position->add(0.5, 0.5, 0.5), new BurpSound());
```
dktapps: This doesn't make a lot of sense. It's not the cake that's burping, it's the player.

This should be using `Player->broadcastSound()` so that it correctly respects the `silent` flag.

dktapps (changes requested):
This sound is missing from all eating events, and should more correctly be played in `Player->consumeHeldItem()`.

src/block/Cake.php
IvanCraft623: ~~In vanilla the burp is reproduced from the center of the block, not from the eating entity.~~

IvanCraft623:
> This sound is missing from all eating events, and should more correctly be played in `Player->consumeHeldItem()`.

In the way you suggest, only if the `Consumable` is an item will the sound be reproduced.

I have decided to put it in `Living::applyConsumptionResults()` because this way all `FoodSource`s play the sound as it happens in vanilla also without limiting it only to Players.

dktapps (approved):

PR #5161, opened 2022-07-14
Correct knockback from explosions

IvanCraft623:
## Introduction
According to the tests in Vanilla, motion is added to the entity's motion, allowing features such as TNT cannons.

### Behavioural changes

Explosion knockback has changed

## Backwards compatibility

N/A

### DIFF
--- a/src/world/Explosion.php
+++ b/src/world/Explosion.php
@@ -195,7 +195,7 @@ class Explosion{
 				}
 
 				$entity->attack($ev);
-				$entity->addMotion($motion->multiply($impact));
+				$entity->addMotion($motion->x * $impact, $motion->y * $impact, $motion->z * $impact);
 			}
 		}
 
--- a/src/world/Explosion.php
+++ b/src/world/Explosion.php
@@ -195,7 +195,7 @@ class Explosion{
 				}
 
 				$entity->attack($ev);
-				$entity->addMotion($motion->x * $impact, $motion->y * $impact, $motion->z * $impact);
+				$entity->setMotion($entity->getMotion()->addVector($motion->multiply($impact)));
 			}
 		}
 
--- a/src/world/Explosion.php
+++ b/src/world/Explosion.php
@@ -195,7 +195,7 @@ class Explosion{
 				}
 
 				$entity->attack($ev);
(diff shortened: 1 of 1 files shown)


ColinHDev (approved):
I've taken a quick look into Java's source code and although the exact calculations are a bit different, the motion is also added there.

I didn't have time to check with IDA (always takes a bit of time to search for something and understand the meaning behind the obfuscated variables, etc.) but if you said making it additive made it similar and my tests suggested the same, then this seems fine to me.

src/world/Explosion.php
```diff
@@ -195,7 +195,7 @@ public function explodeB() : bool{
 				}
 
 				$entity->attack($ev);
-				$entity->setMotion($motion->multiply($impact));
+				$entity->addMotion($motion->x * $impact, $motion->y * $impact, $motion->z * $impact);
```
jasonw4331: There is no reason `setMotion()` needs to be replaced here. The math logic can also be written as `$entity->getMotion()->addVector($motion->multiply($impact))` for a clearer understanding of what is happening.

jasonw4331 (changes requested):

src/world/Explosion.php
IvanCraft623: `addMotion` requires the `x`, `y`, and `z` arguments, also that way will create an object unnecessarily.

IvanCraft623: > The math logic can also be written as `$entity->getMotion()->addVector($motion->multiply($impact))` for a clearer understanding of what is happening.

No, because `Entity::getMotion()` returns a clone of the original object.

Should be `$entity->setMotion($entity->getMotion()->addVector($motion->multiply($impact)))`

jasonw4331 (approved):

src/world/Explosion.php
dktapps: setMotion should be used here, because addMotion doesn't call EntityMotionEvent. Say what you want about EntityMotionEvent, but the behaviour shouldn't be changing if we can help it.

dktapps (changes requested):
This causes a behavioural change, because EntityMotionEvent is not fired by addMotion().

dktapps (approved):

PR #5163, opened 2022-07-16
Implement cauldrons

dktapps:
## Introduction
This pull request is a complete implementation of cauldrons, minus powder snow cauldrons.

### API changes
- The following new classes have been added:
  - `block\Cauldron` - represents an empty cauldron
  - `block\tile\Cauldron`
  - `block\LavaCauldron`
  - `block\PotionCauldron`
  - `block\WaterCauldron`
- `Block->readStateFromWorld()` is now able to return a `Block` that will replace itself, if necessary. In most cases it should just return `$this`, but in special cases (e.g. potion cauldron masquerading as a water cauldron) it may be useful.

#### Rationale for separation of `Cauldron` types
Unlike other implementations, this implementation separates each cauldron type into unique block types. This is because each different type of cauldron has different properties, and therefore different available APIs. For example:
- Only water cauldrons can be dyed, and be used to dye or wash leather armour
- Only lava cauldrons set things on fire, and has a light level of 15, unlike other variants
- Only potion cauldrons can accept potions
- A cauldron with nothing in it cannot have a level greater than 0

Separating these variants into different classes allows for a cleaner, more understandable implementation, as well as API specialization for each type (e.g. `WaterCauldron->getCustomWaterColor()`, `PotionCauldron->getPotionItem()`

## Backwards compatibility
No BC-breaking changes.

## Follow-up
Implement the following gameplay items:
- Powder Snow block
- Bucket of Powder Snow
- Powder Snow Cauldron

## Tests
This has been playtested and various known issues have yet to be addressed:
- [x] Collision box of cauldrons is not correct
- [x] Some blocks can be placed on top of cauldrons when they shouldn't be (e.g. torches)
- [ ] Particle and sound effects are currently missing
- [x] Cauldron liquid colour doesn't immediately update when dyed due to a client "bug" (similar problem as flower pots)


dktapps:
Additional missing features:
- Removing patterns from banners
- Undyeing shulker boxes

PR #5165, opened 2022-07-16
Added Entity::getLastLocation,Entity::getLastMotion

xxAROX:
## Introduction
Cannot get `lastLocation`,`lastMotion` from outside of entity class.

### API changes
```diff
+ Entity::getLastLocation()
+ Entity::getLastMotion()
```

*i don't know why its not already built in*

### DIFF
--- a/src/entity/Entity.php
+++ b/src/entity/Entity.php
@@ -1370,6 +1370,10 @@ abstract class Entity{
 		return clone $this->motion;
 	}
 
+	public function getLastMotion() : Vector3{
+		return clone $this->lastMotion;
+	}
+
 	public function setMotion(Vector3 $motion) : bool{
 		Utils::checkVector3NotInfOrNaN($motion);
 		if(!$this->justCreated){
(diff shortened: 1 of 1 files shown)


dktapps:
> i don't know why its not already built in

what would you need it for? it's only used internally for deciding whether to send movement packets over the network...

xxAROX:
detect if a entity moved: x, y, z, yaw and pitch

xxAROX:
*i need it for all kind of entities*

dktapps:
Adding an API because you need it for something specific never ends well. This code won't make it to a release for weeks at best, even if it's merged.

xxAROX:
currently i use reflection property to get the value, but thats annoying it would be easier with the changes

xxAROX:
> Adding an API because you need it for something specific never ends well. This code won't make it to a release for weeks at best, even if it's merged.

then i change it when the update comes 🤷‍♂️

dktapps:
I don't think it makes much sense to expose these. As I said previously, this is only used internally for deciding whether to send network packets or not. It's only updated when those packets are sent. Given that the conditions for sending packets don't necessarily align with the expectations of plugins, I'm doubtful that exposing these is sensible.

xxAROX:
okay

PR #5170, opened 2022-07-16
Cauldron: Fill if has water at upside

IvanCraft623:
## Introduction
If the cauldron has water in the upper part, it is now filled with water as in Vanilla. This behavior happens only with water.

https://user-images.githubusercontent.com/57236932/179370976-28f87ffd-66db-4697-9d6a-1372781c8109.mp4


dktapps:
are you sure this is any water block and not just sources? I haven't researched this mechanic so I'm unsure of the behaviour.

IvanCraft623:
> are you sure this is any water block and not just sources? I haven't researched this mechanic so I'm unsure of the behaviour.

Yes, any water blocks causes this.

IvanCraft623:
Vanilla:

https://user-images.githubusercontent.com/57236932/179380202-6e43d390-de96-4358-82d5-2fb73d8640c8.mp4

dktapps:
what about if the cauldron contains something?

IvanCraft623:
> what about if the cauldron contains something?


> what about if the cauldron contains something?

it seems that the cauldrons that mojang defines by default as water cauldrons (Cauldron, WaterCauldron, PotionCauldron) have the characteristic of filling completely with water when it is on top.

Cauldrons that have a custom color show particles when filled and the custom color disappears.

IvanCraft623:
Ready for merge!

The cauldrons are programmed to have the following behavior when having water on top as in vanilla:

`Cauldron`: It is replaced by a `WaterCauldron` filled to the maximum capacity.
`PotionCauldron`: It is replaced by a `WaterCauldron` filled to the maximum capacity.
`WaterCauldron`: Custom water color is set to null and is filled to maximum capacity.

One thing to mention is that the lava cauldrons and powder snow do not react.

NTT1906:
> One thing to mention is that the lava cauldrons and powder snow do not react.

They do but only if you continue to attack it with any items

IvanCraft623:
Result:

https://user-images.githubusercontent.com/57236932/179873101-2cc58349-8f1f-4a68-9696-186c33a0b0ac.mp4

PR #5196, opened 2022-07-27
Fixed Bamboo being not placeable on blocks it should be

ColinHDev:
## Introduction
Due to https://github.com/pmmp/PocketMine-MP/commit/d9b050fb688155ec962f574388eb48342fc8f9d1, Bamboo can no longer be placed on `sand` and `gravel` blocks.

### Relevant issues
\---

### API changes
\---

### Behavioural changes
\---

## Backwards compatibility
\---

## Follow-up
After a `gravel` tag is added, the `Bamboo` and `BambooSapling` classes need to be updated, so their `canBeSupportedBy()` method checks for that tag.

## Tests
-> Bamboo can now be placed on blocks with the `sand` tag.

### DIFF
--- a/src/block/Bamboo.php
+++ b/src/block/Bamboo.php
@@ -124,8 +124,11 @@ class Bamboo extends Transparent{
 	}
 
 	private function canBeSupportedBy(Block $block) : bool{
-		// TODO: Gravel
-		return $block->hasTypeTag(BlockTypeTags::DIRT) || $block->hasTypeTag(BlockTypeTags::MUD) || $block->hasTypeTag(BlockTypeTags::SAND);
+		return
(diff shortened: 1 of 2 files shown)


src/block/Bamboo.php
```diff
@@ -124,7 +124,8 @@ public function getModelPositionOffset() : ?Vector3{
 	}
 
 	private function canBeSupportedBy(Block $block) : bool{
-		return $block->hasTypeTag(BlockTypeTags::DIRT) || $block->hasTypeTag(BlockTypeTags::MUD);
+		// TODO: Gravel
```
dktapps: You can just check for the ID. Tags are only used when groups of similar blocks are found.

PR #5197, opened 2022-07-27
Made TallGrass placeable again

ColinHDev:
## Introduction
Due to https://github.com/pmmp/PocketMine-MP/commit/d9b050fb688155ec962f574388eb48342fc8f9d1, `TallGrass` got accidentally made unplaceable since immediately after it getting placed, the neighbour block update called on that block will break it.

### Relevant issues
\---

### API changes
\---

### Behavioural changes
\---

## Backwards compatibility
\---

## Follow-up
\---

## Tests
TallGrass won't immediately break after it gets placed. (Also fixes using bone meal on grass blocks not spawning `TallGrass` blocks, as they immediately break.)

### DIFF
--- a/src/block/TallGrass.php
+++ b/src/block/TallGrass.php
@@ -48,7 +48,7 @@ class TallGrass extends Flowable{
 
 	public function onNearbyBlockChange() : void{
 		$down = $this->getSide(Facing::DOWN);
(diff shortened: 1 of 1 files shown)


dktapps:
I'd prefer separating this logic out into a private `canBeSupportedBy` method, similar to other classes.

PR #5200, opened 2022-07-29
Cauldron: Fixed when dyeing armor the item was not given to the player.

IvanCraft623:
## Introduction

Currently when trying to dye leather armor in a cauldron, it simply does not seem to work.

## Tests
Result:
https://user-images.githubusercontent.com/57236932/181696010-ff57ac95-ee60-4b66-9359-31d701437e46.mp4

### DIFF
--- a/src/block/WaterCauldron.php
+++ b/src/block/WaterCauldron.php
@@ -137,7 +137,7 @@ final class WaterCauldron extends FillableCauldron{
 					ItemTypeIds::LEATHER_PANTS,
 					ItemTypeIds::LEATHER_BOOTS => true,
 					default => false
-				} && $item->getCustomColor()?->toRGBA() !== $this->customWaterColor->toRGBA()){
+				} && $item->getCustomColor()?->toRGBA() !== $this->customWaterColor->toRGBA() && $player !== null){
 					$newItem = (clone $item)->setCustomColor($this->customWaterColor);
 					$item->pop();
 					if($item->getCount() === 0){
--- a/src/block/WaterCauldron.php
+++ b/src/block/WaterCauldron.php
@@ -138,8 +138,8 @@ final class WaterCauldron extends FillableCauldron{
 					ItemTypeIds::LEATHER_BOOTS => true,
 					default => false
 				} && $item->getCustomColor()?->toRGBA() !== $this->customWaterColor->toRGBA()){
-					$item->pop();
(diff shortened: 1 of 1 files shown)


src/block/WaterCauldron.php
```diff
@@ -137,8 +137,14 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
 					ItemTypeIds::LEATHER_PANTS,
 					ItemTypeIds::LEATHER_BOOTS => true,
 					default => false
-				} && $item->getCustomColor()?->toRGBA() !== $this->customWaterColor->toRGBA()){
-					$item->setCustomColor($this->customWaterColor);
+				} && $item->getCustomColor()?->toRGBA() !== $this->customWaterColor->toRGBA() && $player !== null){
```
dktapps, line 140: This change is unnecessary

src/block/WaterCauldron.php
```diff
@@ -137,8 +137,14 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+					$newItem = (clone $item)->setCustomColor($this->customWaterColor);
```
dktapps: This should use the result of `$item->pop()`

src/block/WaterCauldron.php
```diff
@@ -137,8 +137,14 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+					$item->pop();
+					if($item->getCount() === 0){
+						$player->getInventory()->setItemInHand($newItem);
+					}else{ //if the stack size was bigger than 1 (usually won't happen, but might be caused by plugins
+						$player->getInventory()->addItem($newItem);
+					}
```
dktapps, line 146: Use `$returnedItems` for this.

dktapps (changes requested):

dktapps:
Actually this whole change doesn't make any sense. Any modification to `$item` should be automatically synced to the inventory.

src/block/WaterCauldron.php
IvanCraft623: The use of returnedItems will result in non-vanilla behavior, because in Vanilla the item is set to the player's current slot

dktapps:
Closing this since the solution is incorrect. The cause of the problem is described in #5220.

PR #5202, opened 2022-07-29
start.sh: Skipping the script arguments

ghost:
## Introduction
Now the `start.sh` can be used as a runner for any PHP crap! Well, there will still be no extra parameters to the PocketMine-MP

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -2,13 +2,16 @@
 DIR="$(cd -P "$( dirname "${BASH_SOURCE[0]}" )" && pwd)"
 cd "$DIR"
 
+SCRIPT_ARGUMENTS_OFFSET=1
 while getopts "p:f:l" OPTION 2> /dev/null; do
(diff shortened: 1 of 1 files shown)


dktapps:
I don't see the point.

ghost:
`./start.sh composer.phar update` instead `./bin/php7/bin/php composer.phar update` (or extra symbolic link like `ln -s ./bin/php7/bin/php php`).

Also, plugins can use the arguments passed. For example: https://github.com/poggit/devirion, use CLI arguments

ghost:
This change is just for convenience. And I doubt it will ruin anyone's life

dktapps:
If you want a change for your own convenience, you can easily create your own script.

Besides, it's simpler to just prefix the PHP binary path to your PATH environment variable instead of overengineering a script that wasn't intended for this purpose anyway.

PR #5205, opened 2022-08-01
Increases the skin size limitation of 128x128 to 512x512

xxAROX:


mukeenanyafiq:
is this even work?

xxAROX:
> is this even work?

it only removes the limitation lol

xxAROX:
this is not persona impl

dktapps:
> is this even work?

no

PR #5208, opened 2022-08-05
Items: added `Nametag`

supercrafter333:
I saw that the item `nametag` isn't registered. This PR adds nametags to PocketMine-MP 5.


dktapps:
This is pointless. It doesn't do anything.

PR #5209, opened 2022-08-06
Implemented working Name tag

IvanCraft623:
## Introduction

Implement the nametags and their function (currently only obtainable through commands because there are no anvils).

## Follow-up
- https://github.com/pmmp/Language/pull/134

## Tests

https://youtu.be/DCpjGbXHEh0
(The interactive tag for fishing rod was implemented overriding the item on a plugin for testing the api)


xerenahmed:
I think an event must be written for this feature.

IvanCraft623:
> I think an event must be written for this feature.

There is PlayerEntityInteractEvent

src/entity/Living.php
```diff
@@ -352,6 +353,19 @@ protected function onHitGround() : ?float{
 		return $newVerticalVelocity;
 	}
 
+	public function onInteract(Player $player, Vector3 $clickPos) : bool{
+		$item = $player->getInventory()->getItemInHand();
+		if($item->getTypeId() === ItemTypeIds::NAME_TAG && !$this instanceof Human && $item->hasCustomName()){
```
dktapps: `$this instanceof` defeats the purpose of OOP. Also, I'm not sure this logic belongs in `Living`.

IvanCraft623: > I'm not sure this logic belongs in `Living`

According to my tests and the wiki, all Living entities with the exception of EnderDragon and Human can be named with the nametag.

IvanCraft623: > `$this instanceof` defeats the purpose of OOP

Solved.

xRodzMC (dismissed):

xRodzMC:
Looks good, well done

dktapps:
The PR looks functionally OK.

The thing that's giving me pause is that the mode of operation here is inconsistent with every other method of item interaction. For example, weapon damage when attacking an entity is not handled by the entity itself; it's processed in `Item->onAttackEntity()`. We also already have API methods like `Item->onInteractBlock()`.

IvanCraft623:
> The PR looks functionally OK.
> 
> The thing that's giving me pause is that the mode of operation here is inconsistent with every other method of item interaction. For example, weapon damage when attacking an entity is not handled by the entity itself; it's processed in `Item->onAttackEntity()`. We also already have API methods like `Item->onInteractBlock()`.

I agree with adding a method like that, but for this specific case I am more inclined towards the idea of adding an interface for renameable entities because they have a property: `EntityMetadataProperties::INTERACTIVE_TAG`.

IvanCraft623:
The following changes have been made:
 - Logic for renaming has been moved to `NameableTrait`, renameable entities should use it and implement `Nameable`
 - Implemented API for Interactive Tags (interaction button displayed above the hotbar).
 - Interactive tag is now displayed when looking at the renameable entity.

The above changes require [this](https://github.com/pmmp/Language/pull/134) to be merged

IvanCraft623:
This PR has caught up with the latest changes and now uses the new `Item::onInteractEntity()` method.

src/entity/NameableTrait.php
```diff
@@ -0,0 +1,45 @@
...
+
+	/**
+	 * @see Entity::getInteractiveTag()
+	 */
+	public function getInteractiveTag(Player $player, Item $item) : ?string{
+		if($item->getTypeId() === ItemTypeIds::NAME_TAG && $item->hasCustomName()){
+			return KnownTranslationKeys::ACTION_INTERACT_NAME;
+		}
+		return parent::getInteractiveTag($player, $item);
+	}
```
dktapps: This seems redundant. It can be implemented in Living directly.

src/entity/Nameable.php
```diff
@@ -0,0 +1,39 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\entity;
+
+use pocketmine\item\Item;
+use pocketmine\math\Vector3;
+use pocketmine\player\Player;
+
+interface Nameable{
```
dktapps: This interface seems redundant.

src/player/Player.php
```diff
@@ -607,6 +611,36 @@ public function setDisplayName(string $name) : void{
...
+	 */
+	public function setLookingEntity(?Entity $looking) : void{
+		if($looking === null){
+			$this->lookingId = null;
+		}elseif($looking->closed){
+			throw new \InvalidArgumentException("Supplied looking entity is garbage and cannot be used");
+		}else{
+			$this->lookingId = $looking->getId();
+		}
+	}
```
dktapps, line 635: This entire block looks redundant.

src/player/Player.php
```diff
@@ -1852,6 +1887,23 @@ public function interactEntity(Entity $entity, Vector3 $clickPos) : bool{
...
+		$item = $this->inventory->getItemInHand();
+		$interactiveTag = ($looking !== null ? $looking->getInteractiveTag($this, $item) : null) ?? $item->getInteractiveTag();
+		if($interactiveTag !== $this->interactiveTag){
+			$this->interactiveTag = $interactiveTag;
+			$this->networkPropertiesDirty = true;
+			$this->sendData([$this]);
+			return true;
+		}
+		return false;
+	}
```
dktapps: This also looks redundant.

dktapps (dismissed):
The interactive tags stuff looks like a really overengineered mess and can be greatly simplified.

src/player/Player.php
IvanCraft623: What is redundant?

IvanCraft623:
I do not understand what is redundant
`Nameable` and `NameableTrait` exists because not all `Living` entities are nameable, some examples are:
- `Human`
- `EnderDragon`
and there are also non-`Living` entities that are nameable:
- `ArmorStand`

jasonw4331:
I believe the look detection is what is being considered as redundant here. The nametag can be implemented from when it gets used and shouldn't depend on whether or not an entity is being looked at.

dktapps:
The look detection isn't redundant in itself, I just think that the implementation of it is vastly overcomplicated.

PR #5218, opened 2022-08-11
 crash when joining an empty server

Likan-dep:


Superomarking:
This has been fixed in [7e6bea5](https://github.com/NetherGamesMC/PocketMine-MP/commit/7e6bea5d0de7d99d9df57c4c6020f1a90066ba02), This isn't really a pmmp but NG's multiprotocol's issue, Don't state this here.

PR #5223, opened 2022-08-12
Added Item::getIdentifier

xxAROX:
### API changes
```diff
+ Item::getIdentifier()
```

### DIFF
--- a/src/item/Item.php
+++ b/src/item/Item.php
@@ -440,7 +440,7 @@ class Item implements \JsonSerializable{
 		return VanillaBlocks::AIR();
 	}
 
-	public function getIdentifier(): ItemIdentifier{
+	public function getIdentifier() : ItemIdentifier{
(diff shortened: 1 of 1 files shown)


dktapps:
If you want to get a change merged, it should have some reason to be...

xxAROX:
> If you want to get a change merged, it should have some reason to be...

No more `new ItemIdentifier` every time when getting the identifier from an Item

dktapps:
That doesn't explain anything. I don't see why you need this at all.

xxAROX:
For classes that uses ItemIdentifier as property for identifying an Item, if i construct such a class then i only do `$item->getIdentifier()` and no longer `new ItemIdentifier($id, $meta)`

xxAROX:
if this makes no sense for you close the PR 👨

PR #5225, opened 2022-08-14
Implement Fire Charge

IvanCraft623:
## Introduction
Implements the fire charge.

## Follow-up
This item becomes a projectile with a dispenser, but the functionality of the dispensers is not yet implemented. :(

## Tests
https://user-images.githubusercontent.com/57236932/184519732-ce9dc84a-2fe6-448d-bacf-d6f7badd77bd.mp4

### DIFF
--- a/src/block/TNT.php
+++ b/src/block/TNT.php
@@ -32,6 +32,7 @@ use pocketmine\item\Durable;
 use pocketmine\item\enchantment\VanillaEnchantments;
 use pocketmine\item\FlintSteel;
 use pocketmine\item\Item;
+use pocketmine\item\ItemTypeIds;
 use pocketmine\math\RayTraceResult;
 use pocketmine\math\Vector3;
 use pocketmine\player\Player;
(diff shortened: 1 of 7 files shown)


dktapps:
Fire charges can also be used to ignite TNT.

src/block/TNT.php
```diff
@@ -82,6 +83,11 @@ public function onBreak(Item $item, ?Player $player = null, array &$returnedItem
 	}
 
 	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if($item->getTypeId() === ItemTypeIds::FIRE_CHARGE){
+			$item->pop();
+			$this->ignite();
+			return true;
+		}
```
IvanCraft623, line 90: I added a this block, because the fire charge decreases the stack count when used in tnt, but other not Durable items such as enchanted with fire aspect do not.

dktapps (approved):
LGTM

PR #5226, opened 2022-08-14
BlockFormEvent: Added getCausingBlock() method

### LINKED ISSUE #5171: BlockFormEvent: Add possibility the get the causing block for the event
## Description
Unlike for example the [`BlockBurnEvent`](https://github.com/pmmp/PocketMine-MP/blob/stable/src/event/block/BlockBurnEvent.php) or the [`BlockSpreadEvent`](https://github.com/pmmp/PocketMine-MP/blob/stable/src/event/block/BlockSpreadEvent.php), the [`BlockFormEvent`](https://github.com/pmmp/PocketMine-MP/blob/stable/src/event/block/BlockFormEvent.php) currently only contains the information about the block which will be changed and its new state.
This issue proposes that the `BlockFormEvent` gets a `causingBlock` property and a matching `getCausingBlock()` getter.
(Even the method [`liquidCollide()`](https://github.com/pmmp/PocketMine-MP/blob/stable/src/block/Liquid.php#L381-L389) requires the causing block as a parameter, which is not even used, neither in the method in the `Liquid` class nor in any subclass.)

## Justification
This proposal is most likely useless for most people, especially since the `BlockFormEvent` is not an event which might be commonly used in plugins.
But when working on a kind of land protection plugin, I ran into a problem where liquids outside of a certain area were able to interact with liquids in that area, which is not desired.

## Alternative methods
To get the causing block, you could check the block above and all horizontal blocks of the given block and check if those blocks are liquids. But this would also require checking the type of liquid of each block, since e.g. if your block is lava, you need to search for a water block, etc.
So, it would be possible, but with a lot more code than with a `getCausingBlock()` method.

ColinHDev:
## Introduction
Fixes the later-mentioned issue and allows plugins to get the block, which caused a block to form into a new state, e.g. lava to obsidian or concrete powder to concrete.

### Relevant issues
* Fixes #5171 

### API changes
- Through the new `getCausingBlock()` method within the `BlockFormEvent`, it is now possible to get the block which caused the target block to form into a new state.

### Behavioural changes
\---

## Backwards compatibility
This PR is not backwards compatible as it adds a third required parameter to the [`BlockFormEvent`'s constructor](https://github.com/ColinHDev/PocketMine-MP/blob/37724dbf85ecba030bdf67e284d0d922eaf2d8b9/src/event/block/BlockFormEvent.php#L37).

## Follow-up
\---

## Tests
```php
	public function onBlockForm(BlockFormEvent $event) : void{
		var_dump($event->getCausingBlock()->getName());
	}
```
Outputs e.g. `string(5) "Water"` when lava and water collide and form obsidian.

### DIFF
--- a/src/block/Liquid.php
+++ b/src/block/Liquid.php
@@ -366,7 +366,7 @@ abstract class Liquid extends Transparent{
 	}
 
 	protected function liquidCollide(Block $cause, Block $result) : bool{
-		$ev = new BlockFormEvent($this, $result);
+		$ev = new BlockFormEvent($this, $result, $cause);
 		$ev->call();
 		if(!$ev->isCancelled()){
 			$world = $this->position->getWorld();
--- a/src/block/ConcretePowder.php
+++ b/src/block/ConcretePowder.php
@@ -42,8 +42,8 @@ class ConcretePowder extends Opaque implements Fallable{
 	}
 
 	public function onNearbyBlockChange() : void{
-		if(($block = $this->checkAdjacentWater()) !== null){
-			$ev = new BlockFormEvent($this, $block);
+		if(($water = $this->getAdjacentWater()) !== null){
+			$ev = new BlockFormEvent($this, VanillaBlocks::CONCRETE()->setColor($this->color), $water);
 			$ev->call();
 			if(!$ev->isCancelled()){
 				$this->position->getWorld()->setBlock($this->position, $ev->getNewState());
@@ -54,16 +54,20 @@ class ConcretePowder extends Opaque implements Fallable{
 	}
 
 	public function tickFalling() : ?Block{
-		return $this->checkAdjacentWater();
+		if ($this->getAdjacentWater() === null) {
+			return null;
+		}
+		return VanillaBlocks::CONCRETE()->setColor($this->color);
 	}
 
-	private function checkAdjacentWater() : ?Block{
+	private function getAdjacentWater() : ?Water{
 		foreach(Facing::ALL as $i){
 			if($i === Facing::DOWN){
 				continue;
 			}
-			if($this->getSide($i) instanceof Water){
-				return VanillaBlocks::CONCRETE()->setColor($this->color);
+			$block = $this->getSide($i);
+			if($block instanceof Water){
+				return $block;
 			}
 		}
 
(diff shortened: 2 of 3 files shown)


src/block/Liquid.php
```diff
@@ -366,7 +366,7 @@ protected function checkForHarden() : bool{
 	}
 
 	protected function liquidCollide(Block $cause, Block $result) : bool{
-		$ev = new BlockFormEvent($this, $result);
+		$ev = new BlockFormEvent($this, $result, $cause);
```
dktapps, line 369: You might want to test and verify this actually does what you think. This parameter was never previously used, so if it wasn't passed correctly, no one will know.

ColinHDev: This method is only called within the [`Lava` block class](https://github.com/pmmp/PocketMine-MP/blob/next-major/src/block/Lava.php) in those four cases:
(I verified all four of them by using some `var_dump`s.)
- [`Lava` -> `Obsidian`](https://github.com/pmmp/PocketMine-MP/blob/next-major/src/block/Lava.php#L76)
A lava source collides with any water block and forms obsidian.
```
Target Block: 		string(4) "Lava"
New State:		string(6) "Obsidian"
Causing Block: 		string(8) "Water"
```
- [`Lava` -> `Cobblestone`](https://github.com/pmmp/PocketMine-MP/blob/next-major/src/block/Lava.php#L79)
A non-source lava block collides with any water block and forms cobblestone.
```
Target Block: 		string(4) "Lava"
New State:		string(6) "Cobblestone"
Causing Block: 		string(8) "Water"
```
- [`Lava` -> `Basalt`](https://github.com/pmmp/PocketMine-MP/blob/next-major/src/block/Lava.php#L88)
Lava (above a soul soil block) collides with blue ice and forms basalt.
```
Target Block: 		string(4) "Lava"
New State:		string(6) "Basalt"
Causing Block: 		string(8) "Blue Ice"
```
- [`Water` -> `Stone`](https://github.com/pmmp/PocketMine-MP/blob/next-major/src/block/Lava.php#L99)
A lava block flows into any water block and forces it to form into stone.
```
Target Block: 		string(4) "Water"
New State:		string(6) "Stone"
Causing Block: 		string(8) "Lava"
```

dktapps (approved):

PR #4463, opened 2021-09-18
Reverse the order of plugin disabling

Nerahikada:
## Introduction
The order in which plugins are loaded is ensured by settings such as `depend`, but the order in which they are disabled is not.
This causes the problem that dependent plugins are disabled first.
By reversing the order in which plugins are disabled, this problem is solved.

### Relevant issues
none

### API changes
none

### Behavioural changes
The order in which plugins are disabled when the server is shut down has been reversed.

## Backwards compatibility
none

## Follow-up
none

## Tests
I created two plugins and tested.
```php
/**
 * @name ParentPlugin
 * @main MyPlugin\ParentPlugin
 * @version 1.0.0
 * @api 4.0.0
 */
namespace MyPlugin{
	class ParentPlugin extends \pocketmine\plugin\PluginBase{
	}
}
```
```php
/**
 * @name ChildPlugin
 * @main MyPlugin\ChildPlugin
 * @version 1.0.0
 * @api 4.0.0
 * @depend ParentPlugin
 */
namespace MyPlugin{
	class ChildPlugin extends \pocketmine\plugin\PluginBase{
	}
}
```
server log:
```
2021-09-18 [17:50:12.786] [Server thread/INFO]: This server is running PocketMine-MP version 4.0.0-BETA3+dev
2021-09-18 [17:50:12.786] [Server thread/INFO]: PocketMine-MP is distributed under the LGPL License
2021-09-18 [17:50:17.521] [Server thread/INFO]: Loading resource packs...
2021-09-18 [17:50:18.062] [Server thread/INFO]: Loading DevTools v1.15.0+dev
2021-09-18 [17:50:18.191] [Server thread/INFO]: Loading ParentPlugin v1.0.0
2021-09-18 [17:50:18.192] [Server thread/INFO]: Loading ChildPlugin v1.0.0
2021-09-18 [17:50:18.192] [Server thread/INFO]: Enabling DevTools v1.15.0+dev
2021-09-18 [17:50:18.193] [Server thread/INFO]: [DevTools] Registered folder plugin loader
2021-09-18 [17:50:18.457] [Server thread/INFO]: Preparing world "world"
2021-09-18 [17:50:18.536] [Server thread/INFO]: Enabling ParentPlugin v1.0.0
2021-09-18 [17:50:18.536] [Server thread/INFO]: Enabling ChildPlugin v1.0.0
2021-09-18 [17:50:19.043] [Server thread/INFO]: Minecraft network interface running on 0.0.0.0:19132
[...]

Issue #4465, opened 2021-09-18
Disallow loading plugins if plugins which softdepended on them have already been loaded


dktapps:
## Description
This is essentially #4464 in reverse.

If a plugin was loaded with the assumption that softdepended plugin X is not loaded, plugin X suddenly becoming available after the fact may result in unexpected and/or undefined behaviour.

In addition, if changes like #4463 are implemented, this could lead to accidental dependency on things provided by softdepend plugins that will get torn down in the wrong order, furthering the potential for undefined behaviour + crashes.

PR #5227, opened 2022-08-14
PluginManager: Ensure dependents are disabled before dependencies in …

dktapps:
…disablePlugins()

this could later be expanded to disablePlugin() to make this disable order mandatory, to provide certainty for plugin devs.

**But**, you ask, **why not just take the load order and reverse it?**

- PluginManager is not currently expected to maintain the order of `PluginManager->plugins` or `PluginManager->enabledPlugins`, making this fragile. Plugins may be disabled and later reenabled, which implicitly changes the order of stuff like `PluginManager->enabledPlugins`.
- When randomly disabling plugins at runtime (e.g. via disablePlugin()), it's necessary to know which plugins are dependent on the plugin being disabled anyway. Tracking dependents means we can implement this in the future without doing the same work twice.

### Relevant issues
#4463 

### Behavioural changes
Plugins with dependents are now always disabled before their dependencies, regardless of enable order.

## Follow-up
Make `disablePlugin()` refuse to disable plugins which have enabled dependents.

## Tests
Locally tested with simple dependent chains (plugin1 -> plugin2).


SOF3:
Wrong base branch?

dktapps:
It's a GitHub bug. The branches are in sync.

dktapps:
Changed base branch to flush out the bug. Commit history is now clean.

src/plugin/PluginManager.php
```diff
@@ -453,6 +456,15 @@ public function enablePlugin(Plugin $plugin) : bool{
 			if($plugin->isEnabled()){ //the plugin may have disabled itself during onEnable()
 				$this->enabledPlugins[$plugin->getDescription()->getName()] = $plugin;
 
+				foreach($plugin->getDescription()->getDepend() as $dependency){
+					$this->pluginDependents[$dependency][$plugin->getDescription()->getName()] = true;
+				}
+				foreach($plugin->getDescription()->getSoftDepend() as $dependency){
+					if(isset($this->plugins[$dependency])){
```
SOF3, line 463: could the dependency get loaded after server start, causing this to work incorrectly?

dktapps: Technically yes, but this isn't a use case we actively support anyway. If the plugin reached this point, it's because we decided to ignore the missing soft dependency and load it anyway, so the plugin shouldn't really be attempting to use it anyway.

See also: #4465

SOF3: Sounds like a possible problem in the future. Isn't the fix as simple as removing this if check?

dktapps: No. That would cause plugins with unresolved dependencies to get unloaded in the wrong order.

SOF3: how would it? it is just a pseudo node to hold the place when the plugin actually gets loaded.

dktapps: Plugins won't be disabled until the number of enabled dependents reaches 0.

Having verbalized that, I can see what you mean...

dktapps: A case where your suggestion results in a problem:
- Plugin1 has a softdepend on Plugin2, but Plugin2 is not found. Softdepend is ignored, and Plugin1 is loaded anyway
- Plugin2, which circularly depends on Plugin1, is later loaded manually from outside the plugin folder
- Now, a circular dependency has been established within `pluginDependents`, and the server enter an infinite loop on shutdown

If these plugins were loaded together, they'd both fail to load due to circular dependencies. But due to the nature of softdepend, loading them separately would allow this problem to occur. That's why this check is needed.

dktapps: This issue would be removed by implementing #4465 , but that's outside the scope of this PR.

dktapps:
Any objections to merging this?

src/plugin/PluginManager.php
SOF3: I suppose we can neglect the problems with late-loaded plugins, since they already have their own problems with a tracking issue.

SOF3 (approved):

PR #5230, opened 2022-08-16
Changing the return type of the SingletonTrait getter function

SmallkingDev:
## Introduction

This change is to solve the conflict that occurs in the idea when suggesting functions related to classes that are using Singleton.


alvin0319:
#3981

SmallkingDev:
> #3981

Thanks for letting me know, I didn't know there was already a pull request related to this subject.

PR #5232, opened 2022-08-17
Implement Goat horns

IvanCraft623:
## Introduction
Implement Goat horns and its sounds

## Follow-up
- Update [LevelSoundEvent](https://github.com/pmmp/BedrockProtocol/blob/master/src/types/LevelSoundEvent.php)

## Tests

https://user-images.githubusercontent.com/57236932/185046909-a2a8421a-8e18-484b-abea-b0af2bc6b0d3.mp4


src/item/GoatHornType.php
```diff
@@ -0,0 +1,83 @@
...
+	protected static function setup() : void{
+		self::registerAll(
+			new self("ponder", fn() => new GoatHornPonderSound()),
+			new self("sing", fn() => new GoatHornSingSound()),
+			new self("seek", fn() => new GoatHornSeekSound()),
+			new self("feel", fn() => new GoatHornFeelSound()),
+			new self("admire", fn() => new GoatHornAdmireSound()),
+			new self("call", fn() => new GoatHornCallSound()),
+			new self("yearn", fn() => new GoatHornYearnSound()),
+			new self("dream", fn() => new GoatHornDreamSound())
```
dktapps: These closures seem redundant.

dktapps (commented):
The PR looks mostly fine, but I'm dubious about the way the sounds are handled.

I feel like the sounds could be done with much less code by having a single `GoatHornSound` class which accepts `GoatHornType` and `match()`es the sound ID. 

Is there some advantage to having a class per sound that I'm not seeing?

src/item/GoatHorn.php
```diff
@@ -0,0 +1,71 @@
...
+class GoatHorn extends Item implements Releasable{
+
+	private GoatHornType $goatHornType;
+
+	public function __construct(ItemIdentifier $identifier, string $name){
+		$this->goatHornType = GoatHornType::PONDER();
+		parent::__construct($identifier, $name);
+	}
+
+	protected function encodeType(RuntimeDataWriter $w) : void{
```
dktapps: This needs to be updated (encodeType no longer exists)

src/item/GoatHorn.php
```diff
@@ -0,0 +1,71 @@
...
+		$w->goatHornType($this->goatHornType);
+	}
+
+	public function getType() : GoatHornType{ return $this->goatHornType; }
```
dktapps: Prefer more specific stuff like `getSoundType` or `getHornType` for this.

We do have `getType` in other places, but it's very generic and ambiguous, so it's best avoided.

dktapps (dismissed):

src/item/GoatHornType.php
IvanCraft623: Implemented a `GoatHornType::getSoundId()`, so now works similar to records

src/item/GoatHorn.php
IvanCraft623: tbh I prefer, that the code is consistent, if it is common (as observed in several places in the PM code) I would expect by intuition that to access the variants you use `getType`, so for consistency it would be strange to change the format.

dktapps: My intent is to replace the getType() in other places with more specific APIs in the future. It currently only appears in 2-3 places, and the more specific style is much more common.

src/item/GoatHornType.php
```diff
@@ -0,0 +1,54 @@
...
+			self::PONDER => LevelSoundEvent::HORN_CALL0,
+			self::SING => LevelSoundEvent::HORN_CALL1,
+			self::SEEK => LevelSoundEvent::HORN_CALL2,
+			self::FEEL => LevelSoundEvent::HORN_CALL3,
+			self::ADMIRE => LevelSoundEvent::HORN_CALL4,
+			self::CALL => LevelSoundEvent::HORN_CALL5,
+			self::YEARN => LevelSoundEvent::HORN_CALL6,
+			self::DREAM => LevelSoundEvent::HORN_CALL7
+		};
+	}
```
dktapps: This shouldn't be exposed to the API. The only place it's used is inside GoatHornSound anyway.

IvanCraft623: Done :P

ShockedPlot7560 (dismissed):
PR looks fine

GamerMJay:
cooldown is missing :0

dktapps:
This looks OK, although the cooldown is indeed missing. That should be easily addressed with the merge of #6405.

ipad54:
I've added the missing cooldown.

Screenshot_20241111_134311

ipad54 (approved):

dktapps (approved):
Looks like this is ready to go

PR #5235, opened 2022-08-17
Implement Turtle Shell

IvanCraft623:
## Introduction
Implement turtle helmet.

## Tests

https://user-images.githubusercontent.com/57236932/185227178-db326037-6293-4eff-a439-fb7663d501ff.mp4


src/item/TurtleHelmet.php
```diff
@@ -0,0 +1,40 @@
...
+use pocketmine\entity\effect\EffectInstance;
+use pocketmine\entity\effect\VanillaEffects;
+
+class TurtleHelmet extends Armor{
+
+	/**
+	 * @return EffectInstance[]
+	 * @phpstan-return list<EffectInstance>
+	 */
+	public function getEffects() : array{
```
dktapps: This function feels pointless. You are already hardcoding the activation logic for this, so I don't see why add an extra class.

IvanCraft623: ~~I was thinking about an alternative, would be to add new parameters to ArmorTypeInfo, would be the armor effects and maybe a cooldown to apply them.~~

IvanCraft623: The suggested changes have been applied

dktapps:
I'm thinking an `Item->onTickWorn()` or something of that nature might make sense. This would be used in the future for elytra durability too.

IvanCraft623:
Done `Item->onTickWorn()`, If you have feedback, let me know.

ColinHDev:
Wouldn't it be more intuitive to only have that method in the `Armor` class?

dktapps:
Elytra aren't armour, and there's more items that aren't armour which can be worn, such as pumpkins.

src/item/TurtleHelmet.php
```diff
@@ -0,0 +1,41 @@
...
+
+use pocketmine\entity\effect\EffectInstance;
+use pocketmine\entity\effect\VanillaEffects;
+use pocketmine\entity\Human;
+use pocketmine\entity\Living;
+
+class TurtleHelmet extends Armor{
+
+	public function onTickWorn(Living $entity) : bool{
+		if($entity instanceof Human && !$entity->isUnderwater()){
```
dktapps, line 34: I'm not sure why this should be limited to humans. It seems reasonable that any entity that can equip it should be able to experience its effects, no?

src/entity/Living.php
```diff
@@ -606,6 +606,12 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
 			if($this->doAirSupplyTick($tickDiff)){
 				$hasUpdate = true;
 			}
+
+			foreach ($this->armorInventory->getContents() as $item) {
```
dktapps:
```suggestion
			foreach($this->armorInventory->getContents() as $item){
```

src/item/TurtleHelmet.php
IvanCraft623: I have already tested this, the effect is only affects to human entities, it could be a bug, I found this report on MCJava: https://bugs.mojang.com/browse/MC-195642

IvanCraft623: As can seen in the following video, the time for the zombie to drown does not increase even if it is wearing the turtle shell.
https://user-images.githubusercontent.com/57236932/186192463-e954b6fc-e285-4182-adde-efc6d430fbc8.mp4

I also tested this with pillagers

src/item/Item.php
```diff
@@ -558,6 +559,13 @@ public function onAttackEntity(Entity $victim, array &$returnedItems) : bool{
 		return false;
 	}
 
+	/**
+	 * Called when this item is being worn by an entity.
+	 */
+	public function onTickWorn(Living $entity) : bool{
+		return false;
```
dktapps, line 568: Please document the significance of the return value.

dktapps (commented):
Looks OK, minus some documentation.

src/entity/Living.php
```diff
@@ -606,6 +606,12 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
 			if($this->doAirSupplyTick($tickDiff)){
 				$hasUpdate = true;
 			}
+
+			foreach($this->armorInventory->getContents() as $item){
+				if($item->onTickWorn($this)){
```
dktapps, line 615: It's possible `onTickWorn()` may alter the item itself, e.g. elytra applying damage. In this case, the item would need to be set back into the inventory afterwards.

IvanCraft623:
Done

src/item/Item.php
dktapps: This returns true in TurtleHelmet, which isn't changing its count or durability, so this doc isn't accurate.

src/item/TurtleHelmet.php
```diff
@@ -0,0 +1,41 @@
...
+			$entity->getEffects()->add(new EffectInstance(VanillaEffects::WATER_BREATHING(), 200, 0, false));
```
xerenahmed, line 35: Why there is no `applyDamage()` for the wormed item? you were set the same item when return true

IvanCraft623: Because `TurtleHelmet` is not damaged by wearing it...

IvanCraft623:
Any changes needed to be merged?

src/item/TurtleHelmet.php
xerenahmed: so why are you returning true and re-setting the same item in inventory?

src/entity/Living.php
```diff
@@ -606,6 +606,13 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
 			if($this->doAirSupplyTick($tickDiff)){
 				$hasUpdate = true;
 			}
+
+			foreach($this->armorInventory->getContents() as $index => $item){
+				if($item->onTickWorn($this)){
+					$hasUpdate = true;
+					$this->armorInventory->setItem($index, $item);
```
xerenahmed, line 617: ..

dktapps (approved):

Issue #5204, opened 2022-07-31
Sub-MOTD (Lan-MOTD) Support


xRodzMC:
## Must be changed
- [x] Must added SubMotd support like [Nukkit](https://github.com/CloudburstMC/Nukkit)

## Screenshot
![](https://www.wikihow.com/images/thumb/7/75/Join-Servers-in-Minecraft-PE-Step-38.jpg/v4-460px-Join-Servers-in-Minecraft-PE-Step-38.jpg.webp)

## Why?
The "My World" will change as on `server.properties` provided sub-motd. We can say its useless but if we add this feature we can say people will not going to make their own pocketmine-mp for changing `VersionInfo.php` the name. So we can use function `NetworkSession->setNetwork("Server Name", "Server Sub-MOTD" = "PocketMine-MP Server");` if value is missing it will returning `PocketMine-MP` as default. People wont make their own useless Pocketmine-MP Forks to implement this feature.

dktapps:
No one is going to fork the repo unless they really have absolutely nothing better to do, and think it's worth losing official support just to change some text that no one sees anyway. It doesn't show on the main server list. This means only people developing servers will see it, and those people should have more important things to think about than a random bit of text that no player will see.

dktapps:
Besides, if someone really wants this bad enough to fork, they could easily submit a PR, which no one has done ...

xRodzMC:
Implement this on https://github.com/pmmp/PocketMine-MP/pull/5237

PR #5237, opened 2022-08-18
Implement Sub-MOTD Feature

xRodzMC:
## Introduction

This implementes the sub-motd for PocketMine-MP
### Relevant issues

* Implemented https://github.com/pmmp/PocketMine-MP/issues/5204 as **not planned**

### API changes

- **src/network/Network.php**
- **src/network/NetworkInterface.php**
- **src/network/mcpe/raklib/RaklibInterface.php**
### Behavioural changes

- You can now configure the `sub-motd` from `server.properties`

## Backwards compatibility

- You can now use `Network->setName("Motd", "Sub-MOTD")` if the submotd is not specified it would be return as `Server->getName()` method.

## Follow-up

Check the following codes before you are merging this request. Any request changes will be maintained by @xqwtxon

## Tests

- Check [this](https://github.com/xqwtxon/SubMotd-TestPlugin) repository for plugin testing.

## TODO
Well done.


xRodzMC:
I might to fix this checks.. later

But it works in the game. You might to test this at in-game. If you found a bug. You can tell me here what is it?

src/network/NetworkInterface.php
```diff
@@ -37,7 +37,7 @@ interface NetworkInterface{
 	 */
 	public function start() : void;
 
-	public function setName(string $name) : void;
+	public function setName(string $name, string $subName = "PocketMine-MP") : void;
```
dktapps: This is BC-breaking.

It would also make more sense to make this nullable and have null as the default, so that the callee can decide what to do if nothing was specified (e.g. to use previous value, or to use Server->getSubMotd())

src/network/Network.php
```diff
@@ -126,22 +127,34 @@ public function unregisterInterface(NetworkInterface $interface) : void{
...
 
 	/**
-	 * Sets the server name shown on each interface Query
+	 * Sets the server name and sub name shown on each interface Query
 	 */
-	public function setName(string $name) : void{
+	public function setName(string $name, string $subName = "PocketMine-MP") : void{
 		$this->name = $name;
+		/**
+		 * This is a hack to fix ip dummy address
```
dktapps: I have no idea what this is supposed to mean.

src/network/Network.php
```diff
@@ -126,22 +127,34 @@ public function unregisterInterface(NetworkInterface $interface) : void{
 	}
 
 	/**
-	 * Sets the server name shown on each interface Query
+	 * Sets the server name and sub name shown on each interface Query
```
dktapps: Most people won't have any idea what "sub name" means.

src/Server.php
```diff
@@ -386,6 +386,10 @@ public function getMotd() : string{
 		return $this->configGroup->getConfigString("motd", self::DEFAULT_SERVER_NAME);
 	}
 
+	public function getSubMotd() : string{
```
dktapps, line 404: A doc comment should be added that explains where this is seen (i.e. only on LAN servers, below the main server name).

dktapps (changes requested):

src/network/NetworkInterface.php
dktapps: Further, `PocketMine-MP` should not be hardcoded all over the place.

xRodzMC: Ok, i will change this to nullable.

src/network/Network.php
xRodzMC: when you are "" the motd changes to dummy address like 127.0.0.1 and second is 2001:0db8:85a3:0000:0000:8a2e:0370:7334 are prompt in the friend list

src/network/Network.php
xRodzMC: ok, i will change this what is it...

src/network/Network.php
dktapps: That's not the point. The point is that the comment is inadequate.

src/Server.php
xRodzMC: I can add

xRodzMC (commented):
done all

src/Server.php
dktapps: The doc comment you added gives no more useful information than the function name itself.

xRodzMC (commented):
is this ok the doc? or should i add `@return string`

xRodzMC (commented):
this is now ok for me, i just trying to fix this ip dummy issue because it has BC Break...

xRodzMC (commented):
this should be fixed. `>>>HEAD` and `>>>COMMITSHA` are conflicting...

xerenahmed:
@xqwtxon can you reopen the PR? It is easy than working with conflicts.

PR #5261, opened 2022-08-28
Working Living->hasLineOfSight()

### LINKED ISSUE #2959: Living->hasLineOfSight() is broken
https://github.com/pmmp/PocketMine-MP/blob/2e0dd574e0ca8e8fa81606b0124b18744ac1d6c3/src/pocketmine/entity/Living.php#L192

IvanCraft623:
## Introduction

Now returns and see whether Living is able to visualize an entity

### Relevant issues
* Fixes #2959
## Tests

https://user-images.githubusercontent.com/57236932/187098662-8f1218e9-bcbb-4143-b329-df02d0a893a8.mp4

### DIFF
--- a/src/entity/Living.php
+++ b/src/entity/Living.php
@@ -293,16 +293,22 @@ abstract class Living extends Entity{
 		return $nbt;
 	}
 
+	/**
+	 * Returns whether this entity has not block line of sight to another entity.
+	 *
+	 * @throws \InvalidArgumentException if the entities are not in the same world
+	 */
 	public function hasLineOfSight(Entity $entity) : bool{
 		$world = $this->getWorld();
 		if ($world !== $entity->getWorld()) {
-			return false;
+			throw new \InvalidArgumentException("Target entity is not in the same");
+			
 		}
 		$start = $this->getEyePos();
 		$end = $entity->getEyePos();
 		if ($end->subtractVector($start)->normalize()->lengthSquared() > 0) {
 			foreach(VoxelRayTrace::betweenPoints($start, $end) as $vector3){
-				if(!$world->getBlockAt($vector3->x, $vector3->y, $vector3->z)->isTransparent()){
+				if(!$world->getBlockAt((int) floor($vector3->x), (int) floor($vector3->y), (int) floor($vector3->z))->isTransparent()){
 					return false;
 				}
 			}
--- a/src/entity/Living.php
+++ b/src/entity/Living.php
@@ -302,7 +302,6 @@ abstract class Living extends Entity{
 		$world = $this->getWorld();
 		if ($world !== $entity->getWorld()) {
 			throw new \InvalidArgumentException("Target entity is not in the same world");
-			
 		}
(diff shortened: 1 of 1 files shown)


jasonw4331:
The code written only tests if there are no blocks in the way of the eye-to-eye line. It does not test if the entity is able to see any part of the entity passed into the argument. This function will return false even if there is a line of sight to the entity's feet

src/entity/Living.php
```diff
@@ -293,10 +293,26 @@ public function saveNBT() : CompoundTag{
...
+	 * @throws \InvalidArgumentException if the entities are not in the same world
+	 */
 	public function hasLineOfSight(Entity $entity) : bool{
-		//TODO: head height
+		$world = $this->getWorld();
+		if ($world !== $entity->getWorld()) {
+			throw new \InvalidArgumentException("Target entity is not in the same world");
+		}
+		$start = $this->getEyePos();
+		$end = $entity->getEyePos();
```
SOF3, line 307: so seeing the legs of the target entity is not considered a line of sight?

src/entity/Living.php
```diff
@@ -293,10 +293,26 @@ public function saveNBT() : CompoundTag{
...
+		if ($end->subtractVector($start)->normalize()->lengthSquared() > 0) {
```
SOF3, line 308: what is this supposed to do? the modulus of a normalized vector is always (approximately) 1. did you want to check if they are the same point instead?

dktapps:
I think having a line of sight that intersects anywhere with the target's AABB would be more suitable.

src/entity/Living.php
```diff
@@ -293,10 +293,26 @@ public function saveNBT() : CompoundTag{
...
+			foreach(VoxelRayTrace::betweenPoints($start, $end) as $vector3){
+				if(!$world->getBlockAt((int) floor($vector3->x), (int) floor($vector3->y), (int) floor($vector3->z))->isTransparent()){
```
dktapps, line 310: floor() is not necessary here, VoxelRayTrace will always return whole number vectors.

dktapps: I doubt that isTransparent is suitable here, because objects like chests are technically transparent, but aren't actually see-through. isTransparent is a property relevant only for light propagation.

dktapps:
To be honest, I doubt this function has any practical use cases. Having a direct line of sight is rarely useful. It's more likely that an entity is within another entity's _field of view_, rather than its direct sight line.

Zero usages in plugins for 3.x and 4.x: https://poggit.pmmp.io/grepPlugins/2022-09-15-9f06a1d0ed39a578.html

I'd expect someone to have at least mistakenly used it and found out it doesn't work. To further add to this hypothesis, it's been broken since inception, and nobody reported it until I noticed the code was commented in 2019.

I vote that we just get rid of the function instead.

dktapps:
This method has been deprecated with the intent of removing it in PM5, so I'm going to close this.

PR #5268, opened 2022-08-31
Introduce and use TextFormat::addBase()

dktapps:
## Introduction
This PR is an alternative, less invasive way to deal with the problems described in #4364 and #4116.

This function changes the "base" format of a string, i.e. the colours and formats that are used after a RESET (&r) formatting code.

An example of where this is needed is in the logger.

The following code:
```
$logger->notice("I'm a " . TextFormat::RED . "special" . TextFormat::RESET . " cookie");
```
causes the "cookie" part of the message to show as grey, instead of the expected aqua for NOTICE level messages.

This function operates by directly suffixing the base format to any &r formatting code found.

### Relevant issues
#4116 

### Behavioural changes
Log messages containing `&r` will now have slightly different behaviour - using `&r` in them will cause them to revert to whatever the log message's original colour was (e.g. aqua for NOTICE, yellow for WARNING, etc) instead of the terminal's default text colour.

## Backwards compatibility

## Tests
The fruits of my labour can be seen by using any of the commands affected by this PR.
Unit tests to follow.


SOF3 (approved):

dktapps:
Hesitating to merge this because I'm not sure that `rebase()` is a good enough name for the function.

dktapps:
Using addBase() instead, since it gives a better implication about the additive nature of the function (previous base formatting can't be removed).

PR #5273, opened 2022-09-02
ConsoleReader: Use proc_open()'s socket support to send commands back…

### LINKED ISSUE #4624: ConsoleReader: bogus assumption about stream_socket_accept()
### Issue description
https://github.com/pmmp/PocketMine-MP/blob/e0eeb87ea09a10bb8c5aba9324df5baba019e091/src/console/ConsoleReaderThread.php#L101

stream_socket_accept() may return false for a non-fatal reason here: the console reader subprocess may not yet have had time on the CPU to do anything yet (e.g. if the machine is overloaded). In this case, the thread should handle the problem gracefully and not crash.

### OS and versions

* PocketMine-MP: 4.0.0

dktapps:
… to the main server process

Support for this was introduced in PHP 8.0, though not mentioned in any changelog: php/php-src#5777

This simplifies the subprocess handling considerably. However, there is a potential for problems if PHP generates any E_* errors, since these get written to STDOUT as well.

This hasn't been thoroughly tested or proofed.


dktapps:
Seems like we should be able to use `ini_set('display_errors', 'stderr')` to get around the STDOUT error reporting problem, which we really ought to have it do anyway.

dktapps:
This should also fix #5561.

The STDOUT problem can be solved by passing a randomly generated token from the server process to the console reader, and requiring that any commands generated must be written to STDOUT with this prefix. Any other STDOUT should be echoed onto the parent console directly.

dktapps:
I stalled on this because, due to the use of stdout, it's possible other things might write outputs that could be considered commands. For example, error output might cause commands to be replayed. I'd prefer the chance of this to be 0 rather than "very unlikely", but I'm not sure how to make that happen.

ShockedPlot7560 (dismissed):
Looks ok too me

src/console/ConsoleReaderChildProcessDaemon.php
```diff
@@ -104,21 +101,44 @@ private function shutdownSubprocess() : void{
...
+			$line = rtrim($line, "\n");
+
+			if($line === ""){
+				//keepalive
+				return null;
+			}
+
+			$command = null;
+
+			$parts = explode(":", $line, 2);
```
dktapps: This won't work correctly if the command contains a : character. Noticed this when I flipped the params to begin with, but forgot to fix it.

PR #5276, opened 2022-09-02
Implement Lingering potion & Area Effect Cloud

IvanCraft623:
## Introduction
Implements lingering potions and area effect cloud + custom effect support, also complements #5163

### API changes

* The following new classes have been added:
  * `entity\effect\EffectContainer`
  * `pocketmine\entity\object\AreaEffectCloud`
  * `pocketmine\item\LingeringPotion`
* The following API methods have changed signatures
  * `SplashPotion->setLinger()` - now returns `self` instead of `void`

#### Rationale for `EffectContainer` addition
Unlike entities (Living) there are different types of things that can contain effects that are not necessarily applied. Such is the case of AreaEffectCloud.

Other stuff that may contain effects:
* Potion: https://github.com/pmmp/PocketMine-MP/blob/d5762d3f4487772ffb6b7c77af4cd659585f6520/src/item/Potion.php#L63
* Tipped arrows

## Backwards compatibility

No BC-breaking changes.

## Tests
https://youtu.be/cRAjwEVbb0w
This has been playtested and and there was only one issue:
* On fire animation is shown when extinguishing fire (this is an already known bug on PM).


NhanAZ (dismissed):

dktapps:
Nice, I'll review this tomorrow.

IvanCraft623:
Minecraft Bedrock implements AreaCloudEffect in a weird way, behaves differently than described in the wiki and in java.
Some examples:
* [ ] The number of times the effect is applied varies per effect (I still can't figure out the criteria).
* [x] Regardless of where the `SplashPotion` falls, the cloud will be positioned in the middle of the block.
* [x] `ReapplicationDelay` changes depending if the effect is Instantaneous.
* [x] `RadiusOnUse` is applied even the cloud has no effects.
* [x] The potion effect is stored in the `ListTag` effects, it does not make sense since it is aready storing the `PotionId`.
* [x] It is only updated every 10 ticks.

xRodzMC (dismissed):

src/entity/EntityFactory.php
```diff
@@ -81,6 +82,14 @@ public function __construct(){
 		//define legacy save IDs first - use them for saving for maximum compatibility with Minecraft PC
 		//TODO: index them by version to allow proper multi-save compatibility
 
+		$this->register(AreaEffectCloud::class, function(World $world, CompoundTag $nbt) : AreaEffectCloud{
+			$potionType = PotionTypeIdMap::getInstance()->fromId($nbt->getShort("PotionId", PotionTypeIds::WATER));
+			if($potionType === null){
+				throw new SavedDataLoadingException("No such potion type");
+			}
+			return new AreaEffectCloud(Helper::parseLocation($nbt, $world), $potionType, $nbt);
+		}, ['AreaEffectCloud', 'minecraft:area_effect_cloud'], LegacyIds::AREA_EFFECT_CLOUD);
```
dktapps: Are you sure these are correct? Don't make any assumptions - sometimes the legacy IDs in particular have seriously weird names.

src/entity/object/AreaEffectCloud.php
```diff
@@ -0,0 +1,385 @@
...
+	}
+
+	public function canBeCollidedWith() : bool{
+		return false;
+	}
+
+	/**
+	 * Recalculates the area entity's potion bubbles colour based on the active effects.
+	 */
+	protected function recalculateEffectColor() : void{
```
dktapps: Why is this logic duplicated? The same already exists in `EffectManager`.

src/entity/EntityFactory.php
IvanCraft623: Yeah, vanilla nbt tree:
image

src/entity/object/AreaEffectCloud.php
IvanCraft623: Not really, `EffectManager`'s ignores the color if the effect does not have bubbles and the default color is different.

dktapps: That's one detail. It doesn't require you to duplicate the whole section of code.

IvanCraft623: I added `setDefaultBubbleColor` and `setEffectValidatorForBubbles` api methods, I would like to know if you find them good enough to be able to mark this conversation as resolved.

src/entity/object/AreaEffectCloud.php
```diff
@@ -0,0 +1,384 @@
...
+	public function getRadiusOnUse() : float{
+		return $this->radiusOnUse;
+	}
+
+	public function setRadiusOnUse(float $radiusOnUse) : void{
+		$this->radiusOnUse = $radiusOnUse;
+		$this->networkPropertiesDirty = true;
+	}
+
+	public function getRadiusPerTick() : float{
```
dktapps: This name is unclear and doesn't describe what the function actually returns. It should either be documented or renamed.

src/entity/object/AreaEffectCloud.php
```diff
@@ -0,0 +1,384 @@
...
+		return $this->radiusPerTick;
+	}
+
+	public function setRadiusPerTick(float $radiusPerTick) : void{
+		$this->radiusPerTick = $radiusPerTick;
+		$this->networkPropertiesDirty = true;
+	}
+
+	public function getWaiting() : int{
```
dktapps: Same again here.

src/entity/object/AreaEffectCloud.php
```diff
@@ -0,0 +1,384 @@
...
+		return $this->waiting;
+	}
+
+	public function setWaiting(int $time) : void{
+		$this->waiting = $time;
+		$this->networkPropertiesDirty = true;
+	}
+
+	public function getDuration() : int{
```
dktapps: and here

src/entity/object/AreaEffectCloud.php
```diff
@@ -0,0 +1,384 @@
...
+		return $this->duration;
+	}
+
+	public function setDuration(int $duration) : void{
+		$this->duration = $duration;
+		$this->networkPropertiesDirty = true;
+	}
+
+	public function getDurationOnUse() : int{
```
dktapps: and here

src/entity/object/AreaEffectCloud.php
```diff
@@ -0,0 +1,384 @@
...
+	}
+
+	public function getInitialRadius() : float{
+		return $this->initialRadius;
+	}
+
+	/**
+	 * Sets the initial radius.
+	 *
+	 * @throws \RuntimeException if something attempted to set when it is already ticking
```
dktapps: I don't like this. If the radius needs to be set in advance, it should be passed in the constructor.

src/entity/object/AreaEffectCloud.php
```diff
@@ -0,0 +1,384 @@
...
+		}, $this->effectContainer->all());
+	}
+
+	protected function syncNetworkData(EntityMetadataCollection $properties) : void{
+		parent::syncNetworkData($properties);
+
+		$spawnTime = $this->getWorld()->getTime() - $this->age;
+		$properties->setInt(EntityMetadataProperties::AREA_EFFECT_CLOUD_SPAWN_TIME, $spawnTime);
+		$properties->setInt(EntityMetadataProperties::AREA_EFFECT_CLOUD_DURATION, $this->duration);
+		$properties->setFloat(EntityMetadataProperties::AREA_EFFECT_CLOUD_RADIUS, $this->initialRadius);
```
dktapps: Why isn't this using the current radius?

dktapps (commented):
This PR is rather bloated by the EffectContainer changes so it's difficult for me to see what's going on. I've given this only a cursory review.

src/entity/object/AreaEffectCloud.php
IvanCraft623: The client expects InitialRadius, as I mentioned in a comment the client calculates the radius on its own using SpawnTime, InitialRadius and some other stuff. If we send the current radius it will result in that on the client side the cloud will disappear much earlier than it should. See: https://github.com/df-mc/dragonfly/issues/665

IvanCraft623:
I have improved the documentation, because I am not a native speaker I would like if someone could verify that it has no grammatical errors and makes sense.

src/entity/object/AreaEffectCloud.php
```diff
@@ -0,0 +1,416 @@
...
+	public const TAG_DURATION = "Duration"; //TAG_Int
+	public const TAG_PICKUP_COUNT = "PickupCount"; //TAG_Int
+	public const TAG_DURATION_ON_USE = "DurationOnUse"; //TAG_Int
+	public const TAG_REAPPLICATION_DELAY = "ReapplicationDelay"; //TAG_Int
+	public const TAG_INITIAL_RADIUS = "InitialRadius"; //TAG_Float
+	public const TAG_RADIUS = "Radius"; //TAG_Float
+	public const TAG_RADIUS_CHANGE_ON_PICKUP = "RadiusChangeOnPickup"; //TAG_Float
+	public const TAG_RADIUS_ON_USE = "RadiusOnUse"; //TAG_Float
+	public const TAG_RADIUS_PER_TICK = "RadiusPerTick"; //TAG_Float
+	public const TAG_EFFECTS = "mobEffects"; //TAG_List
```
dktapps: These constants should not be public unless used outside the class. They are of no concern to plugin developers.

src/entity/object/AreaEffectCloud.php
```diff
@@ -0,0 +1,416 @@
...
+use function count;
+use function round;
+
+class AreaEffectCloud extends Entity{
+
+	public const DURATION = 600;
+	public const DURATION_ON_USE = 0;
+
+	public const WAIT_TIME = 10;
+	public const REAPPLICATION_DELAY = 40;
```
dktapps: These numbers are meaningless without units.

src/entity/object/AreaEffectCloud.php
```diff
@@ -0,0 +1,416 @@
...
+
+	public function isFireProof() : bool{
+		return true;
+	}
+
+	public function canBeCollidedWith() : bool{
+		return false;
+	}
+
+	public function getAge() : int{
```
dktapps, line 190: Again, meaningless without units. Please document the function.

src/entity/object/AreaEffectCloud.php
```diff
@@ -0,0 +1,416 @@
...
+	public function getPotionType() : PotionType{
+		return $this->potionType;
+	}
+
+	public function getEffects() : EffectContainer{
+		return $this->effectContainer;
+	}
+
+	/**
+	 * Returns the initial radius.
```
dktapps: Radius in what unit?

src/entity/object/AreaEffectCloud.php
dktapps: `getInitialApplicationAge()` would be a better name. This is actually supposed to be a "time until active" delay.

src/entity/object/AreaEffectCloud.php
```diff
@@ -0,0 +1,416 @@
...
+				if($this->durationOnUse !== 0){
+					$duration = $this->duration + $this->durationOnUse;
+					if($duration <= 0){
+						$this->flagForDespawn();
+						return true;
+					}
+					$this->setDuration($duration);
+				}
+			}
+			$this->setWaiting($this->age + self::WAIT_TIME);
```
dktapps: I think you're using this incorrectly. This will prevent the effect cloud from applying effects for 30 seconds at a time, which defeats the purpose of having `reapplication delay` of 2 seconds.

According to the wiki: The time before deploying the field. The Radius is ignored, meaning that any specified effects is not applied and specified particles appear only at the center of the field, until Age hits this number.

IvanCraft623: I tested for this in vanilla, instant effects have a `reapplication Delay` of 0, but due to `wait Time` they are only applied every 10 ticks. So I noticed that `wait Time` is not only used during the first ticks of the entity's life. I also saw the java code and although the `wait Time` there is 5 ticks, `wait Time` also works as a delay for applying effects.

So...

`WaitTime`: Cloud Delay to apply effects, or update.
`ReapplicationDelay`: Delay per entity to reapply the effects, capped by `WaitTime`.

src/event/entity/AreaEffectCloudApplyEvent.php
```diff
@@ -0,0 +1,69 @@
...
+	 * @return AreaEffectCloud
+	 */
+	public function getEntity(){
+		return $this->entity;
+	}
+
+	/**
+	 * Returns the effected entities.
+	 *
+	 * It is important to note that not every entity is guaranteed to be effected.
```
dktapps:
```suggestion
	 * It is important to note that not every entity is guaranteed to be affected.
```

src/event/entity/AreaEffectCloudApplyEvent.php
```diff
@@ -0,0 +1,69 @@
...
+	 * The cloud may die during the application of its effects due to the depletion
```
dktapps:
```suggestion
	 * The cloud may despawn during the application of its effects due to the depletion
```

src/event/entity/AreaEffectCloudApplyEvent.php
```diff
@@ -0,0 +1,69 @@
...
+	 * of {@link AreaEffectCloud::getDurationChangeOnUse()} or {@link AreaEffectCloud::getRadiusChangeOnUse()}.
```
dktapps:
```suggestion
	 * by {@link AreaEffectCloud::getDurationChangeOnUse()} or {@link AreaEffectCloud::getRadiusChangeOnUse()}.
```

src/event/entity/AreaEffectCloudApplyEvent.php
dktapps: I don't like this. It's not particularly useful without certainty that something is actually going to happen to all the entities as a result of the event.

src/entity/object/AreaEffectCloud.php
IvanCraft623: See: https://github.com/df-mc/dragonfly/issues/665#issuecomment-1338609858

IvanCraft623: A hack has been implemented which consists of sending incorrect values in the entity properties `AREA_EFFECT_CLOUD_RADIUS_CHANGE_ON_PICKUP` and `AREA_EFFECT_CLOUD_RADIUS_PER_TICK`, to disable client-sided radius calculation.

src/entity/projectile/SplashPotion.php
```diff
@@ -164,10 +177,13 @@ public function willLinger() : bool{
 
 	/**
 	 * Sets whether this splash potion will create an area-effect-cloud when it lands.
+	 *
+	 * @return $this
 	 */
-	public function setLinger(bool $value = true) : void{
+	public function setLinger(bool $value = true) : self{
```
dktapps: This is a BC break for inheritors of SplashPotion

src/entity/effect/EffectManager.php
```diff
@@ -91,60 +53,29 @@ public function remove(Effect $effectType) : void{
...
 	 * If a weaker effect of the same type is already applied, it will be replaced.
 	 * If a weaker or equal-strength effect is already applied but has a shorter duration, it will be replaced.
 	 *
 	 * @return bool whether the effect has been successfully applied.
 	 */
-	public function add(EffectInstance $effect) : bool{
-		$oldEffect = null;
-		$cancelled = false;
-
+	public function add(EffectInstance $effect, bool $force = false) : bool{
```
dktapps: This parameter addition is also BC breaking for inheritors

src/item/LingeringPotion.php
```diff
@@ -0,0 +1,69 @@
...
+class LingeringPotion extends ProjectileItem{
+
+	private PotionType $potionType;
+
+	public function __construct(ItemIdentifier $identifier, string $name){
+		$this->potionType = PotionType::WATER;
+		parent::__construct($identifier, $name);
+	}
+
+	protected function describeType(RuntimeDataReader|RuntimeDataWriter $w) : void{
```
dktapps: This should be `RuntimeDataDescriber`, we're not in PM5 alpha anymore

See https://github.com/pmmp/PocketMine-MP/blob/49da50659f08472ddc41b6925098006ba95de3f5/src/item/SplashPotion.php#L36-L38

src/item/LingeringPotion.php
```diff
@@ -0,0 +1,69 @@
...
+use pocketmine\entity\projectile\SplashPotion as SplashPotionEntity;
+use pocketmine\entity\projectile\Throwable;
+use pocketmine\player\Player;
+
+class LingeringPotion extends ProjectileItem{
+
+	private PotionType $potionType;
+
+	public function __construct(ItemIdentifier $identifier, string $name){
+		$this->potionType = PotionType::WATER;
```
dktapps: This can be a property default directly, it's a native enum now

src/item/StringToItemParser.php
```diff
@@ -1202,6 +1202,7 @@ private static function registerDynamicItems(self $result) : void{
 
 			$result->register($prefix("potion"), fn() => Items::POTION()->setType($potionType));
 			$result->register($prefix("splash_potion"), fn() => Items::SPLASH_POTION()->setType($potionType));
+			$result->register($prefix("lingering_potion"), fn() => Items::SPLASH_POTION()->setType($potionType));
```
dktapps:
```suggestion
			$result->register($prefix("lingering_potion"), fn() => Items::LINGERING_POTION()->setType($potionType));
```

dktapps (changes requested):

src/entity/effect/EffectContainer.php
```diff
@@ -0,0 +1,239 @@
...
+	 * Returns whether the specified effect is active.
+	 */
+	public function has(Effect $effect) : bool{
+		return isset($this->effects[spl_object_id($effect)]);
+	}
+
+	/**
+	 * In the following cases it will return true:
+	 * If a weaker effect of the same type is already applied.
+	 * If a weaker or equal-strength effect is already applied but has a shorter duration.
```
dktapps:
```suggestion
	 * - if the effect type is not already applied
	 * - if an existing effect of the same type can be replaced (due to shorter duration or lower level)
```

src/entity/effect/EffectContainer.php
```diff
@@ -0,0 +1,239 @@
...
+	 *
+	 * @return bool whether the effect can been applied.
```
dktapps: This seems redundant

src/entity/effect/EffectContainer.php
```diff
@@ -0,0 +1,239 @@
...
+			return true;
+		}
+
+		return false;
+	}
+
+	/**
+	 * @phpstan-param \Closure(EffectInstance) : bool $effectValidator
+	 */
+	public function setEffectValidatorForBubbles(\Closure $effectValidator) : void{
```
dktapps: I think this would be better named "effect filter" rather than "effect validator".

src/entity/effect/EffectContainer.php
```diff
@@ -0,0 +1,239 @@
...
+			$this->onlyAmbientEffects = $ambient;
+		}else{
+			$this->bubbleColor = null;
+			$this->onlyAmbientEffects = false;
+		}
+	}
+
+	public function getDefaultBubbleColor() : Color{
+		return $this->defaultBubbleColor;
+	}
```
dktapps: I don't like this. It's only required by AreaEffectCloud.

Really we should check if `getBubbleColor()` is null and then fill it with a default on AreaEffectCloud (or, for the sake of BC, check if the RGBA colour code is 0).

src/entity/object/AreaEffectCloud.php
```diff
@@ -0,0 +1,462 @@
...
+
+	protected function getInitialDragMultiplier() : float{ return 0.0; }
+
+	protected function getInitialGravity() : float{ return 0.0; }
+
+	protected function initEntity(CompoundTag $nbt) : void{
+		parent::initEntity($nbt);
+
+		$this->effectContainer = new EffectContainer();
+		$this->effectContainer->setDefaultBubbleColor(PotionSplashParticle::DEFAULT_COLOR());
```
dktapps: As above, don't like this. Default colour of bubbles shouldn't be the EffectContainer's responsibility, since it's only going to be used by AreaEffectCloud.

dktapps (changes requested):

src/item/LingeringPotion.php
```diff
@@ -0,0 +1,63 @@
...
+		return $this;
+	}
+
+	public function getMaxStackSize() : int{
+		return 1;
+	}
+
+	protected function createEntity(Location $location, Player $thrower) : Throwable{
+		$projectile = new SplashPotionEntity($location, $thrower, $this->potionType);
+		$projectile->setLinger();
```
dktapps: I wonder if we could just add a `bool $linger` parameter to the constructor of `SplashPotion` for this. The two classes are identical apart from this function.

IvanCraft623: Adding it on constructor since the class is not final would be a BC break for inheritors. but we could try a `setLinger()` method.

src/entity/effect/EffectContainer.php
IvanCraft623: I hate having a RGBA(0) and having to do a check on network layer to display the correct color, it would consufuse plugin-devs when doing `getBubbleColor()` on water `AreaEffectCloud`s. but is the only way for BC :c

IvanCraft623: I mean I really prefer the current solution, EffectManager/Container in the first place shouldn't be setting a RGBA(0) when there are no effects

src/item/LingeringPotion.php
dktapps: It's not BC breaking unless it's required.

dktapps: `setLinger()` isn't an option because we already have a type ID constant for lingering potion. It would be a BC break to remove it, and confusing to plugin devs not to remove it.

src/entity/effect/EffectContainer.php
dktapps: > EffectManager/Container in the first place shouldn't be setting a RGBA(0) when there are no effects

Agreed, but I don't think adding more conditionally useless stuff to its API is a good solution. We're gonna have to BC break at some point to fix this anyway.

SplashPotion already suffers from this issue anyway as seen here: https://github.com/pmmp/PocketMine-MP/blob/012894764c1e87569a893014db3ad8e8223f0817/src/entity/projectile/SplashPotion.php#L83-L95

dktapps: To be honest, looking at SplashPotion I'm not really sure why we want to do this EffectContainer refactor anyway. I'm doubtful it makes sense to allow mutating the effects given the increase in complexity it causes.

src/item/SplashPotion.php
```diff
@@ -52,10 +62,19 @@ public function getMaxStackSize() : int{
...
+		$projectile->setLinger($this->linger);
+		return $projectile;
 	}
 
 	public function getThrowForce() : float{
 		return 0.5;
 	}
+
+	/**
+	 * Returns whether this splash potion will create an area-effect cloud when it lands on it's projectile form.
```
dktapps:
```suggestion
	 * Returns whether this splash potion will create an area-effect cloud on impact.
```

src/item/VanillaItems.php
```diff
@@ -514,6 +515,7 @@ protected static function setup() : void{
 		self::register("lapis_lazuli", fn(IID $id) => new Item($id, "Lapis Lazuli"));
 		self::register("lava_bucket", fn(IID $id) => new LiquidBucket($id, "Lava Bucket", Blocks::LAVA()));
 		self::register("leather", fn(IID $id) => new Item($id, "Leather"));
+		self::register("lingering_potion", fn(IID $id) => new SplashPotion($id, "Lingering Potion", [], true));
```
dktapps:
```suggestion
		self::register("lingering_potion", fn(IID $id) => new SplashPotion($id, "Lingering Potion", linger: true));
```

src/entity/effect/EffectContainer.php
```diff
@@ -0,0 +1,239 @@
...
+
+	/**
+	 * @var \Closure[]|ObjectSet
+	 * @phpstan-var ObjectSet<\Closure(EffectInstance) : void>
+	 */
+	protected ObjectSet $effectRemoveHooks;
+
+	protected Color $defaultBubbleColor;
+
+	protected ?Color $bubbleColor = null;
```
dktapps: This is also BC breaking for inheritors (the original field wasn't nullable)

src/entity/effect/EffectContainer.php
```diff
@@ -0,0 +1,239 @@
...
+
+	protected bool $onlyAmbientEffects = false;
+
+	/**
+	 * Validates whether an effect will be used for bubbles color calculation.
+	 *
+	 * @phpstan-var \Closure(EffectInstance) : bool
+	 */
+	protected \Closure $effectValidatorForBubbles;
```
dktapps: This should be renamed to use the filter terminology

dktapps (changes requested):

src/entity/projectile/SplashPotion.php
```diff
@@ -126,10 +127,22 @@ protected function onHit(ProjectileHitEvent $event) : void{
...
+			$entity->setRadiusChangePerTick(-$entity->getInitialRadius() / $entity->getDuration());
+			if(($owner = $this->getOwningEntity()) !== null && !$owner->isClosed()){
+				$entity->setOwningEntity($owner);
+			}
+			foreach($this->getPotionEffects() as $effect){
+				$entity->getEffects()->add($effect);
+				if($effect->getType() instanceof InstantEffect){
+					$entity->setReapplicationDelay(0);
+				}
+			}
```
dktapps: I don't like this.

Currently we have an AEC entity that takes a PotionType in the constructor, but the potion type is useless because the entity doesn't use it. No effects are taken from the potion type. Good way to troll plugin devs trying to spawn AECs with a specific potion type.

src/entity/object/AreaEffectCloud.php
```diff
@@ -0,0 +1,462 @@
...
+		}
+
+		return $hasUpdate;
+	}
+
+	/**
+	 * Returns the effects the area effect cloud provides.
+	 *
+	 * Used to get COPIES to avoid accidentally modifying the same effect instance
+	 * already applied to another entity.
```
dktapps: No one is gonna read this. They'll just use `getEffects()` because that's what they're familiar with.

dktapps (changes requested):

dktapps:
This is stalled pending some verification about the usage of the NBT save data. It's not clear how `PotionId` interacts with `mobEffects`.

kostamax27:
cyclic references in:

https://github.com/IvanCraft623/PocketMine-MP/blob/d0647099a69e136ed4854cbb4bcb554789853e95/src/entity/effect/EffectContainer.php#L70

also in [AreaEffectCloud](https://github.com/IvanCraft623/PocketMine-MP/blob/d0647099a69e136ed4854cbb4bcb554789853e95/src/entity/object/AreaEffectCloud.php) is missing `unset(effectContainer)` in `destroyCycles`

dktapps (approved):

PR #5279, opened 2022-09-05
Update InstantDamage and Regeneration values

IvanCraft623:
## Introduction

While working on #5276  I realized that were outdated.

## Tests
Data were obtained from https://minecraft.fandom.com/wiki/Potion and has been play-tested to match with vanilla

### DIFF
--- a/src/entity/effect/InstantDamageEffect.php
+++ b/src/entity/effect/InstantDamageEffect.php
@@ -33,7 +33,7 @@ class InstantDamageEffect extends InstantEffect{
 
 	public function applyEffect(Living $entity, EffectInstance $instance, float $potency = 1.0, ?Entity $source = null) : void{
 		//TODO: add particles (witch spell)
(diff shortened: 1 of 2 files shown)


ColinHDev (commented):
So, I've looked into this and the changes to `Instant Damage` look good and match both the wiki as well as in-game.
But while checking the [wiki page of the regeneration effect](https://minecraft.fandom.com/wiki/Regeneration#Effect), I noticed this: `Amplifiers outside the range 0–31 (corresponding to levels 1–32) are used modulo 32, making level 33 the same as level 1, etc.`
I know, that we don't aim for bug parity here, but I am not sure whether this counts as a bug and should be left as is in the code or modified to match Mojangs weird behaviour.

dktapps:
I don't think it matters. No one will notice anyway, and if someone does try to use such a high level, they'll think the modulo behaviour itself is a bug.

PR #5289, opened 2022-09-15
fix cs and add document in Effect constructor

xerenahmed:
## Introduction
Fix cs and checks


xRodzMC (approved):

dktapps:
Hi, please keep distinct changes in separate commits

PR #5294, opened 2022-09-21
EntityFactory: Avoid code duplication on validation of creaton functions

IvanCraft623:
## Introduction
Use existing function `Utils::validateCallableSignature()` to validate callable.


dktapps:
Thanks.

For posterity, this function was unnecessary since the changes made in 40d49b88dd24c23937bc9311897f7549492143fc. I'm not sure why I didn't just remove it directly at the time.

dktapps (approved):

PR #5295, opened 2022-09-21
simply code in Permission

xerenahmed:
## Introduction

not really important but removed a TODO


dktapps:
This was left unchanged for precisely the reason that the field type doesn't match the parameter. The TODO was left there because I need to break BC in order to get rid of the useless null.

PR #5296, opened 2022-09-22
Implement snappy compression

AkmalFairuz:
## Introduction
Implement snappy compression.

You can change the compression algorithm in the `pocketmine.yml` file (default: zlib)
```yml
network:
  #Compression algorithm. Supported algorithms: snappy, zlib
  compression-algorithm: zlib
```

An snappy extension is required https://github.com/kjdev/php-ext-snappy that provide `snappy_compress` and `snappy_uncompress` function.

You can use this PM php binary fork that has snappy extension https://github.com/AkmalFairuz/php-build-scripts/actions/runs/3106786925 this binary fork is only supported on Linux and MacOS, I need someone to help me to build the PM php binary with snappy extension on Windows

### API changes

- Added `pocketmine\network\mcpe\compression\CompressorPool`
- Added `public function toNetworkCompressionAlgorithm() : int;` in `pocketmine\network\mcpe\compression\Compressor` interface

### DIFF
--- a/src/entity/Entity.php
+++ b/src/entity/Entity.php
@@ -832,7 +832,7 @@ abstract class Entity{
 				$pk->headYaw = $this->location->yaw;
 				$pk->flags |= MoveActorDeltaPacket::FLAG_HAS_HEAD_YAW;
 			}
-			$this->server->broadcastPackets(array_filter($this->hasSpawned, fn($player) => !isset($this->needsInitialMovement[$player->getId()])), [$pk]);
+			$this->server->broadcastPackets(array_filter($this->hasSpawned, fn($player) => !isset($this->needsInitialMovement[spl_object_id($player)])), [$pk]);
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


jasonw4331 (changes requested):
The Snappy PHP extension should be tested to ensure the required functions exist before enabling the feature

dktapps:
I think we're jumping the gun on this. We're just giving users another way to break their servers without any solid motive to do so.

- Snappy support is still experimental in the game itself.
- There's zero evidence to suggest that this will be beneficial. No performance benchmarks have been done. This is especially important since PM uses libdeflate, which is considerably faster than zlib.
- Snappy does not support compression levels or threshold-based zero compression.
- The mode of operation may also change (I personally suggested per-packet compression type selection, rather than making it global).

dktapps:
Adding to my previous comments, some specific words about php-ext-snappy:

- This extension doesn't have any support for length limiting on decompressed payloads. This means that the server could be taken down with a zip bomb attack when using Snappy.
- The extension itself isn't supported on Windows, or if it is, there's no instructions how to build it.
- The extension uses antiquated Zend APIs that are harmful to performance.
- It also performs useless copies of the output data, further wasting performance.

jasonw4331:
The commits in this PR no longer contain a snappy compression implementation

Issue #5258, opened 2022-08-27
Add support for sending encrypted resource packs


dktapps:
## Description
It's fairly trivial to add support for encrypted resource packs. All that's needed is a way for an encryption key to get from `resource_packs.yml` to `ResourcePacksDataInfo`.

**Note:** This does **not** involve actually decrypting or encrypting packs. The proposal is simply that PM should be able to load and send a resource pack that is already encrypted. Tools like https://encryptmypack.com already exist for encrypting the packs to begin with.

There is some doubt in my mind about how the keys should be provided:

- In `resource_packs.yml`, along side the pack entry.
  - While convenient, it also presents the ability to lose your encryption key by removing the pack from the config.
  - Keys can only use stuff that can be written in YAML, i.e. no raw keys, only ASCII/Unicode
  - Config schema becomes more complex and easier to mess up (we've seen countless users manage to break the config that we already have)
- In a file alongside the pack, e.g. `MyPack.zip` and `MyPack.zip.key`
  - Allows the use of raw (non-printable) encryption keys
  - Less easy to lose due to a configuration error
  - Keys are associated with a pack name, rather than an entry in a config (e.g. if you change the order of packs, you don't need to remember to also move the encryption keys)
- In `resource_packs.yml`, in a separate section
  - Has most of the same advantages as having the keys in files, with the exception of not being able to use raw keys

## Justification
Many servers are already using encrypted resource packs, and are using packet hacks to do so. It's a popular feature.

## Alternative methods
Aforementioned packet hacks, which are the reason why this needs to be implemented.

PR #5297, opened 2022-09-22
Allow specifying a key for encrypted resource packs

BrandPVP:
## Introduction
To specify a resource encryption key, put the key in the [resource].key file alongside the resource pack. Example: vanilla.zip.key

* Allows the use of raw (non-printable) encryption keys
* Less easy to lose due to a configuration error
* Keys are associated with a pack name, rather than an entry in a config (e.g. if you change the order of packs, you don't need to remember to also move the encryption keys)

### Relevant issues
* #5258

### API changes
Added `\pocketmine\resourcepacks\ResourcePack->getEncryptionKey()`
It returns empty string if the pack has no key.

### DIFF
--- a/src/network/mcpe/handler/ResourcePacksPacketHandler.php
+++ b/src/network/mcpe/handler/ResourcePacksPacketHandler.php
@@ -65,17 +65,17 @@ class ResourcePacksPacketHandler extends PacketHandler{
 	){}
 
 	public function setUp() : void{
-		$resourcePackEntries = array_map(static function(ResourcePack $pack) : ResourcePackInfoEntry{
+		$resourcePackEntries = array_map(function(ResourcePack $pack) : ResourcePackInfoEntry{
 			//TODO: more stuff
-			$encryptionKey = $pack->getEncryptionKey();
+			$encryptionKey = $this->resourcePackManager->getPackEncryptionKey($pack->getPackId());
 
 			return new ResourcePackInfoEntry(
 				$pack->getPackId(),
 				$pack->getPackVersion(),
 				$pack->getPackSize(),
-				$encryptionKey,
+				$encryptionKey ?? "",
 				"",
-				$encryptionKey === "" ? "" : $pack->getPackId(),
+				$encryptionKey === null ? "" : $pack->getPackId(), // ContentId is required for client-side validations when the pack is encrypted.
 				false
 			);
 		}, $this->resourcePackManager->getResourceStack());
--- a/src/network/mcpe/handler/ResourcePacksPacketHandler.php
+++ b/src/network/mcpe/handler/ResourcePacksPacketHandler.php
@@ -75,7 +75,7 @@ class ResourcePacksPacketHandler extends PacketHandler{
 				$pack->getPackSize(),
 				$encryptionKey ?? "",
 				"",
-				$encryptionKey === null ? "" : $pack->getPackId(), // ContentId is required for client-side validations when the pack is encrypted.
+				$pack->getPackId(),
 				false
 			);
 		}, $this->resourcePackManager->getResourceStack());
--- a/src/network/mcpe/handler/ResourcePacksPacketHandler.php
+++ b/src/network/mcpe/handler/ResourcePacksPacketHandler.php
@@ -67,7 +67,17 @@ class ResourcePacksPacketHandler extends PacketHandler{
 	public function setUp() : void{
 		$resourcePackEntries = array_map(static function(ResourcePack $pack) : ResourcePackInfoEntry{
 			//TODO: more stuff
-			return new ResourcePackInfoEntry($pack->getPackId(), $pack->getPackVersion(), $pack->getPackSize(), "", "", "", false);
+			$encryptionKey = $pack->getEncryptionKey();
+
+			return new ResourcePackInfoEntry(
+				$pack->getPackId(),
+				$pack->getPackVersion(),
+				$pack->getPackSize(),
+				$encryptionKey,
+				"",
+				$encryptionKey === "" ? "" : $pack->getPackId(),
+				false
+			);
 		}, $this->resourcePackManager->getResourceStack());
 		//TODO: support forcing server packs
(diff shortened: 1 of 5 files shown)


src/resourcepacks/ResourcePack.php
```diff
@@ -45,6 +45,12 @@ public function getPackSize() : int;
 	 */
 	public function getPackVersion() : string;
 
+	/**
+	 * Returns the key with which the pack was encrypted.
+	 * @return string empty string if the pack is not encrypted
+	 */
+	public function getEncryptionKey() : string;
```
SOF3: This is an API breaking change.

Lunarelly:
why just dont specify it in resourcepacks.yml?
like:
name: rp.zip
encrypted: true
key: yourkey

dktapps:
Read the linked issue. It's discussed there.

src/network/mcpe/handler/ResourcePacksPacketHandler.php
```diff
@@ -67,7 +67,17 @@ public function __construct(
...
-			return new ResourcePackInfoEntry($pack->getPackId(), $pack->getPackVersion(), $pack->getPackSize(), "", "", "", false);
+			$encryptionKey = $pack->getEncryptionKey();
+
+			return new ResourcePackInfoEntry(
+				$pack->getPackId(),
+				$pack->getPackVersion(),
+				$pack->getPackSize(),
+				$encryptionKey,
+				"",
+				$encryptionKey === "" ? "" : $pack->getPackId(),
```
dktapps: why is this necessary?

src/resourcepacks/ResourcePack.php
dktapps: A better solution might be to add an EncryptedResourcePack interface that extends this one, or perhaps just keep the keys in ResourcePackManager directly.

src/resourcepacks/ResourcePackManager.php
```diff
@@ -93,13 +95,19 @@ public function __construct(string $path, \Logger $logger){
 					throw new ResourcePackException("Directory resource packs are unsupported");
 				}
 
+				$encryptionKey = "";
```
dktapps: I think `null` would make more sense.

dktapps (changes requested):
This can't be targeted at a stable release since it contains BC breaks.

Branch targets:
- `stable`: Bug fixes only
- `next-minor`: New features, refactors, changes that don't break the API
- `next-major`: Disruptive changes including API breaks

src/resourcepacks/ResourcePackManager.php
```diff
@@ -49,6 +51,9 @@ class ResourcePackManager{
 	/** @var ResourcePack[] */
 	private array $uuidList = [];
 
+	/** @var array<string,string> */
```
dktapps:
```suggestion
	/** @phpstan-var array<string, string> */
```
Please also add an `@var string[]`.

They need to be maintained this way because some tools don't support this generic style of phpdoc.

src/resourcepacks/ResourcePackManager.php
```diff
@@ -106,6 +111,11 @@ public function __construct(string $path, \Logger $logger){
 				if($newPack instanceof ResourcePack){
 					$this->resourcePacks[] = $newPack;
 					$this->uuidList[strtolower($newPack->getPackId())] = $newPack;
+
+					$keyPath = Path::join($this->path, $pack . ".key");
+					if(file_exists($keyPath)){
+						$this->encryptionKeys[strtolower($newPack->getPackId())] = Utils::assumeNotFalse(file_get_contents($keyPath));
```
dktapps: This will crash if the file permissions are incorrect. `ErrorToExceptionHandler::trapAndRemoveFalse()` should be used instead. Catch the `ErrorException` it throws and rethrow it as a `ResourcePackException`.

src/resourcepacks/ResourcePackManager.php
```diff
@@ -153,4 +163,11 @@ public function getPackById(string $id) : ?ResourcePack{
 	public function getPackIdList() : array{
 		return array_keys($this->uuidList);
 	}
+
+	/**
+	 * Returns the key with which the pack was encrypted, or null if the pack has no key.
+	 */
+	public function getPackEncryptionKey(string $id) : ?string {
```
dktapps:
```suggestion
	public function getPackEncryptionKey(string $id) : ?string{
```

dktapps (commented):
Looking better

src/network/mcpe/handler/ResourcePacksPacketHandler.php
```diff
@@ -65,9 +65,19 @@ public function __construct(
...
-			return new ResourcePackInfoEntry($pack->getPackId(), $pack->getPackVersion(), $pack->getPackSize(), "", "", "", false);
+			$encryptionKey = $this->resourcePackManager->getPackEncryptionKey($pack->getPackId());
+
+			return new ResourcePackInfoEntry(
+				$pack->getPackId(),
+				$pack->getPackVersion(),
+				$pack->getPackSize(),
+				$encryptionKey ?? "",
+				"",
+				$encryptionKey === null ? "" : $pack->getPackId(), // ContentId is required for client-side validations when the pack is encrypted.
```
dktapps: Is there any reason not to just always specify this? I don't actually remember what it's for, but there's a similar Geyser PR for reference: https://github.com/GeyserMC/Geyser/pull/3263

dktapps (commented):
Looks good, just one final thing

src/network/mcpe/handler/ResourcePacksPacketHandler.php
BrandPVP: > Is there any reason not to just always specify this? I don't actually remember what it's for, but there's a similar Geyser PR for reference: [GeyserMC/Geyser#3263](https://github.com/GeyserMC/Geyser/pull/3263)

Idk, but this is completely useless if pack is not encrypted. Its set for marketplace packs

dktapps: I just don't think it makes much sense to special-case it, if it works for both encrypted and unencrypted.

PR #5303, opened 2022-09-23
Take Health boost health

UnknownOre:
Remove extra health given by health boost when removing the effect

## Introduction
When removing Health Boost it doesn't take your extra health it just sets your max health leaving you with some extra health

### Behavioural changes
Now health Boost takes the extra health from you


dktapps (approved):
Looks OK, but this does naturally beg the question of why setMaxHealth() allows the current health level to be out of bounds...

Issue #5309, opened 2022-09-27
TaskHandler::setNextRun() Blocks the priority queue


DHJ1:
### Steps to reproduce the issue
1. schedule a delayed task (Task no.1) and set delay to 0, and store the handler
2. set the next run tick to Server::getTick() + 10
3. schedule another delayed task (Task no.2) and the set delay to 5 ticks

### Issue description
- Expected result: Task no.2 should run after 5 ticks and Task no.1 should run after 10 ticks
- Actual result: Task no.1 and Task no.2 run during the same tick, after 10 ticks

### OS and versions

* PocketMine-MP version: 4.9.0 [Protocol 554]
* PHP version: 8.0.12
* Using JIT: no
* Server OS: Linux, linux
* Game version: Android

dktapps:
setNextRun is internal, it's not supposed to be used by a plugin.

DHJ1:
alright, thank you.
https://github.com/pmmp/PocketMine-MP/pull/5310

PR #5310, opened 2022-09-27
Labelled internal TaskHandler methods as @internal

DHJ1:
https://github.com/pmmp/PocketMine-MP/issues/5309#issuecomment-1259581677


dktapps (approved):
Good, though for a future version we should definitely explore redesigning this API. It stinks of Bukkit.

PR #5312, opened 2022-09-28
Implement entity damage from falling anvil

### LINKED ISSUE #4437: [API4] Anvil does not deal fall damage and does not sound when it hits the ground

IvanCraft623:
## Introduction
This PR takes up #4460 due to the author's inactivity, I have practically rewritten the whole PR.

As usual Bugrock uses different values than java to reduce damage when wearing a helmet.

Java: Removes 1/4 of the damage
Bedrock: Removes 1/3 of the damage with a maximum of 10.

### Relevant issues
* Fixes #4437

### API changes
Because anvils are not the only fallable blocks that can do damage, I have added the methods:
`Fallable::getFallDamagePerBlock()` & `Fallable::getMaxFallDamage()`
This will make it easier to implement the Stalacmites.

## Backwards compatibility

No BC breaking changes

## Follow-up
- Add the generic translation for death by falling block.

## Tests
https://youtu.be/cb2r9yOC8xs


src/block/Anvil.php
```diff
@@ -99,4 +99,12 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
 	public function tickFalling() : ?Block{
 		return null;
 	}
+
+	public function getFallDamagePerDistance() : float{
```
jasonw4331: I think this should be named `getFallDamagePerBlock` because the distance measurement is on a per-block basis. Using the word "Distance" here indicates something other than minecraft's blocks are being user for measurement.

Alternatively, this function name can be shortened to omit the "PerDistance" part with `getFallDamage()` though I personally prefer the verbosity of the former option.

src/entity/Living.php
```diff
@@ -465,7 +477,7 @@ public function damageArmor(float $damage) : void{
 		$this->armorInventory->setContents($armor);
 	}
 
-	private function damageItem(Durable $item, int $durabilityRemoved) : void{
+	public function damageItem(Durable $item, int $durabilityRemoved) : void{
```
jasonw4331: This change is not necessary and is outside the scope of this PR.

jasonw4331 (changes requested):
The `damageItem()` function should not be made public, and the function name leaves a bit open to interpretation here.

src/block/Anvil.php
dktapps: getFallDamagePerBlock() would be my preference. The latter is too ambiguous.

src/block/SnowLayer.php
```diff
@@ -105,6 +105,14 @@ public function ticksRandomly() : bool{
...
 	}
 
+	public function getFallDamagePerDistance() : float{
+		return 0.0;
+	}
+
+	public function getMaxFallDamage() : float{
+		return 0.0;
+	}
+
```
dktapps: These default implementations can be added to `FallableTrait` to avoid polluting these classes.

src/entity/object/FallingBlock.php
```diff
@@ -146,6 +150,21 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
 		return $hasUpdate;
 	}
 
+	protected function onHitGround() : ?float{
+		if($this->block instanceof Fallable){
+			$damagePerDistance = $this->block->getFallDamagePerDistance();
```
dktapps: This variable should be renamed according to the above comments

src/entity/Living.php
```diff
@@ -445,6 +449,14 @@ protected function applyPostDamageEffects(EntityDamageEvent $source) : void{
 			if($damage > 0){
 				$attacker->attack(new EntityDamageByEntityEvent($this, $attacker, EntityDamageEvent::CAUSE_MAGIC, $damage));
 			}
+
+			if($source->getModifier(EntityDamageEvent::MODIFIER_HARD_HELMET) < 0){
+				$helmet = $this->armorInventory->getHelmet();
+				if($helmet instanceof Armor){
+					$this->damageItem($helmet, (int) $source->getFinalDamage());
```
dktapps: Are you sure this is correct?

src/entity/Living.php
IvanCraft623: Sorry, copy-paste leftovers from #4460

src/entity/Living.php
IvanCraft623: No, I have not found the formula anywhere.

xRodzMC (approved):

src/entity/Living.php
```diff
@@ -416,6 +416,10 @@ public function applyDamageModifiers(EntityDamageEvent $source) : void{
 		$source->setModifier(-$source->getFinalDamage() * min(ceil(min($totalEpf, 25) * (mt_rand(50, 100) / 100)), 20) * 0.04, EntityDamageEvent::MODIFIER_ARMOR_ENCHANTMENTS);
 
 		$source->setModifier(-min($this->getAbsorption(), $source->getFinalDamage()), EntityDamageEvent::MODIFIER_ABSORPTION);
+
+		if($cause === EntityDamageEvent::CAUSE_FALLING_BLOCK && $this->armorInventory->getHelmet() instanceof Armor){
+			$source->setModifier(-min($source->getFinalDamage() / 3, 10), EntityDamageEvent::MODIFIER_HARD_HELMET);
```
dktapps: Is there a reason why we can't just throw this under MODIFIER_ARMOR?

src/block/utils/Fallable.php
```diff
@@ -33,4 +34,14 @@ interface Fallable{
...
 	 */
 	public function tickFalling() : ?Block;
+
+	/**
+	 * Returns the damage caused per fallen block.
+	 */
+	public function getFallDamagePerBlock() : float;
+
+	/**
+	 * Returns the damage caused per fallen block.
```
dktapps: This documentation looks like a copy paste.

dktapps (approved):
This looks basically OK, although I'd like to have the armour durability change amount checked before merging.

src/entity/Living.php
IvanCraft623: Mainly because it makes it easier to identify when extra damage should be applied to the helmet and `MODIFIER_ARMOR` is only used if the cause do `EntityDamageEvent::canBeReducedByArmor()` return `true` and any armor reduces damage.

dktapps: Wouldn't a combination of CAUSE_FALLING_BLOCK and MODIFIER_ARMOR achieve the same?

dktapps: Never mind, I suppose this way is less fragile.

src/entity/object/FallingBlock.php
```diff
@@ -146,6 +150,21 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
 		return $hasUpdate;
 	}
 
+	protected function onHitGround() : ?float{
+		if($this->block instanceof Fallable){
+			$damagePerBlock = $this->block->getFallDamagePerBlock();
+			if($damagePerBlock > 0 && ($fallenBlocks = round($this->fallDistance) - 1) > 0){
+				foreach($this->getWorld()->getCollidingEntities($this->getBoundingBox()) as $entity){
+					if($entity instanceof Living){
+						$ev = new EntityDamageByEntityEvent($this, $entity, EntityDamageEvent::CAUSE_FALLING_BLOCK, min($fallenBlocks * $damagePerBlock, $this->block->getMaxFallDamage()));
```
dktapps: I think the damage amount can be extracted into a variable instead of recalculating it every iteration.

src/entity/object/FallingBlock.php
```diff
@@ -154,9 +154,11 @@ protected function onHitGround() : ?float{
 		if($this->block instanceof Fallable){
 			$damagePerBlock = $this->block->getFallDamagePerBlock();
 			if($damagePerBlock > 0 && ($fallenBlocks = round($this->fallDistance) - 1) > 0){
+				$damage = min($fallenBlocks * $damagePerBlock, $this->block->getMaxFallDamage());
 				foreach($this->getWorld()->getCollidingEntities($this->getBoundingBox()) as $entity){
 					if($entity instanceof Living){
-						$ev = new EntityDamageByEntityEvent($this, $entity, EntityDamageEvent::CAUSE_FALLING_BLOCK, min($fallenBlocks * $damagePerBlock, $this->block->getMaxFallDamage()));
+						$ev = new EntityDamageByEntityEvent($this, $entity, EntityDamageEvent::CAUSE_FALLING_BLOCK
+							, $damage);
```
dktapps: unnecessary newline

dktapps:
This is just waiting on the correct armour damage calculation, if anyone fancies taking the time to investigate the formula for it.

src/block/utils/Fallable.php
xRodzMC: Lol

src/entity/Living.php
Flonja: After some digging into Java Edition 1.19, it seems to be `(int) min(damage / 4, 1.0)`

IvanCraft623: Bedrock does not match java, I was testing and in bedrock the helmet receives much more damage than in java, also seems to use random numbers

dktapps: I think using the Java solution is better than what we have right now.

dktapps: Please make this match Java edition for now, until someone can figure out what the correct formula is for Bedrock.

IvanCraft623: This is java code 1.16:
```java
this.getItemBySlot(EquipmentSlot.HEAD).<LivingEntity>hurtAndBreak((int)(damage * 4.0f + this.random.nextFloat() * damage * 2.0f), this, (java.util.function.Consumer<LivingEntity>)(entity -> entity.broadcastBreakEvent(EquipmentSlot.HEAD)));
```

IvanCraft623: > This is java code 1.16

This is the version I could find with the most similar results to bedrock (I hope they use the same formula).
Lower and higher versions do different things, for example java 1.19 applies the same damage to the helmetas it does to the player.

IvanCraft623: Changes have been implemented

dktapps: Out of curiosity, where exactly in the JE code did you find this?

IvanCraft623: > Out of curiosity, where exactly in the JE code did you find this?

JE 1.16.2:
- Class: `net/minecraft/world/entity/LivingEntity`
- Method: `hurt()`

https://github.com/Blackjack200/minecraft_client_1_16_2/blob/master/net/minecraft/world/entity/LivingEntity.java#L911

src/entity/Living.php
dktapps: My research into the current version of Bedrock suggests that this is reduced by 25%, same as Java.

src/entity/Living.php
dktapps: This appears to match Bedrock as of 1.19.51.

src/entity/Living.php
dktapps: Why is this marked as resolved?

IvanCraft623: > My research into the current version of Bedrock suggests that this is reduced by 25%, same as Java.

done.

src/event/entity/EntityDamageEvent.php
```diff
@@ -161,7 +161,6 @@ public function canBeReducedByArmor() : bool{
 			case self::CAUSE_VOID:
 			case self::CAUSE_MAGIC:
 			case self::CAUSE_SUICIDE:
-			case self::CAUSE_FALLING_BLOCK:
```
dktapps: This is incorrect. It will allow all armour pieces to reduce the damage.

IvanCraft623: > This is incorrect. It will allow all armour pieces to reduce the damage.

This is correct, This appears to match Bedrock as of 1.19.51.

dktapps: Hmm, you're right. That's pretty stupid ...

src/entity/Living.php
```diff
@@ -440,6 +441,10 @@ public function applyDamageModifiers(EntityDamageEvent $source) : void{
 		$source->setModifier(-$source->getFinalDamage() * min(ceil(min($totalEpf, 25) * (mt_rand(50, 100) / 100)), 20) * 0.04, EntityDamageEvent::MODIFIER_ARMOR_ENCHANTMENTS);
 
 		$source->setModifier(-min($this->getAbsorption(), $source->getFinalDamage()), EntityDamageEvent::MODIFIER_ABSORPTION);
+
+		if($cause === EntityDamageEvent::CAUSE_FALLING_BLOCK && $this->armorInventory->getHelmet() instanceof Armor){
+			$source->setModifier(-min($source->getFinalDamage() / 4, 10), EntityDamageEvent::MODIFIER_HARD_HELMET);
```
dktapps: My research did not show any minimum of 10.

IvanCraft623: > My research did not show any minimum of 10.

done.

IvanCraft623:
What's stopping from merging this PR? :D

src/event/player/PlayerDeathEvent.php
```diff
@@ -155,6 +156,18 @@ public static function deriveMessage(string $name, ?EntityDamageEvent $deathCaus
...
+					$e = $deathCause->getDamager();
+					if($e instanceof FallingBlock){
+						if($e->getBlock()->getTypeId() === BlockTypeIds::ANVIL){
+							return KnownTranslationFactory::death_attack_anvil($name);
+						}else{
+							return KnownTranslationFactory::death_attack_fallingBlock($name);
+						}
+					}
+				}
+
```
dktapps, line 171: break missing here

dktapps:
> What's stopping from merging this PR? :D

Sorry, I'd forgotten about it. If you can address the remaining review comments promptly I can make an exception for feature freeze (otherwise it'll have to wait until PM6).

src/event/player/PlayerDeathEvent.php
IvanCraft623: done

PR #5313, opened 2022-09-29
Spectating players no longer break blocks

### LINKED ISSUE #3636: Spectator players can destroy blocks client-side
### Issue description
In 1.16 you can break the block in gamemode 3(just in client)
- Expected result: What were you expecting to happen?
- Actual result: What actually happened?

1. ...
2. ...

### OS and versions

* PocketMine-MP:3.14.0
* PHP:7.13.4
* Server OS:Ubuntu18.06
* Game version: Android and Win10 

### Plugins

none

Nerahikada:
## Introduction
Players can no longer break blocks on the client side while spectating.

### Relevant issues
* Fixes #3636

### API changes
N/A

### Behavioural changes
The server now sends game mode `CREATIVE_VIEWER` instead of `CREATIVE` for the spectator.

## Backwards compatibility
It does not break backward compatibility.

### DIFF
--- a/src/network/mcpe/convert/TypeConverter.php
+++ b/src/network/mcpe/convert/TypeConverter.php
@@ -82,10 +82,11 @@ class TypeConverter{
 			case GameMode::SURVIVAL()->id():
 				return ProtocolGameMode::SURVIVAL;
 			case GameMode::CREATIVE()->id():
-			case GameMode::SPECTATOR()->id():
 				return ProtocolGameMode::CREATIVE;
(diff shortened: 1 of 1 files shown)


xRodzMC (approved):

dktapps:
They shouldn't be able to do this anyway.

dktapps:
This works correctly if the player is not an operator. It appears this is a client bug due to recent protocol changes, so I'm going to close this.

Issue #5318, opened 2022-10-04
Add WorldSoundEvent


aderoian:
## Description
Whenever a sound is played on a world (using World->addSound()) an event is called. In this event you can do the following:
- Get the sound, world, and target players
- Overwrite the sound and target players
- Cancel the sound

## Justification
This would be a good way to track sounds being played, implement new sounds, and change the targets. 

## Alternative methods
Currently, you need to create a listener to listen for `DataPacketSendEvent` and iterate through the packets being sent for the proper packet.

PR #5322, opened 2022-10-05
Add the event: WorldSoundEvent

aderoian:
## Introduction
Adds an API event to listen for when a Sound is played to a world using World->addSound(). This event can be cancelled, or the sound, position, and/or target players can be changed.

### Relevant issues
#5318 

### API changes
Addition of the event: ``WorldSoundEvent``

### Behavioural changes
n/a

## Backwards compatibility
n/a

## Follow-up
n/a

## Tests
Test plugin to test cancelling, changing the sound, changing the location, and changing the target players.
```php
final class Main extends PluginBase implements Listener{

	}

	/**
	 * @priority MONITOR
	 * @ignoreCancelled true
	 */
	public function onWorldSound(WorldSoundEvent $event) : void{
		// $event->cancel(); // No sounds are played (works)

		// $event->setSound(new AnvilFallSound()); // All sounds are replaced by anvil sounds (works)

		// $event->setPosition($event->getPosition()->add(0, 25, 0)); // Sound is quieter (works)

		// $event->setPlayers([]); // Sound is sent to no one (works)
	}
}
```


src/event/world/WorldSoundEvent.php
```diff
@@ -0,0 +1,81 @@
...
+
+use pocketmine\event\Cancellable;
+use pocketmine\event\CancellableTrait;
+use pocketmine\math\Vector3;
+use pocketmine\player\Player;
+use pocketmine\world\sound\Sound;
+use pocketmine\world\World;
+
+/**
+ * Called when a sound is played in a world using World->playSound()
```
dktapps:
```suggestion
 * Called when a sound is played in a world
```
This is already covered by the @see below

src/event/world/WorldSoundEvent.php
```diff
@@ -0,0 +1,81 @@
...
+	use CancellableTrait;
+
+	/**
+	 * @param Player[]|null $players
+	 */
+	public function __construct(
+		World $world,
+		private Vector3 $position,
+		private Sound $sound,
+		private ?array $players = null
```
dktapps: why does this accept null?

aderoian: https://github.com/pmmp/PocketMine-MP/blob/stable/src/world/World.php#L661-L667

If players == null the world just sends it to the viewers. Developers may only want to change the sound and not the target, so why make them have to figure out who to send that sound to.

IvanCraft623: I don't like that this can be null, it can be understood as that no player will hear the sound.

aderoian: What do we think about a field in the event such as ``toViewersOnly`` which essentially acts as if $players was null?

dktapps: The event would already provide the viewers, so your point doesn't make any sense.

aderoian: Currently, if no players are specified (``$players = null``) the server adds them to a packet buffer, and does not actually get the viewers of that chunk at that exact moment.

dktapps: Hmm, I see. That means that collecting the viewers at that point would change the internal buffering behaviour of sending sounds. Not sure what the best solution for that is.

aderoian: The solution that I used was if null was passed get the players near the sound before sending the event. The buffering behavior gets the players who can see that chunk either way.

dktapps:
I'm doubtful that allowing the position of the sound to be changed is useful or desirable.

aderoian:
Yeah, I can agree with that.

aderoian:
Any chance we can change this PR to also do the same thing for ``World::addParticle()``?

dktapps:
Better keep that for a separate PR.

aderoian:
What else is this pr waiting on?

dktapps:
> I'm doubtful that allowing the position of the sound to be changed is useful or desirable.

aderoian (commented):
Is it just me or does ``$players`` never get updated from the event, therefore the check for whether the array is changed is always `true`?

JavierLeon9966:
Yeah, you're right

src/world/World.php
```diff
@@ -677,7 +677,7 @@ public function addSound(Vector3 $pos, Sound $sound, ?array $players = null) : v
 
 		$pk = $ev->getSound()->encode($pos);
 		if(count($pk) > 0){
-			if($players === $this->getViewersForPosition($pos)){
+			if($players === $ev->getRecipients()){
```
dktapps: This is not the correct fix. The correct solution would be to overwrite $players with the result of getRecipients() before entering this code.

dktapps (approved):

PR #5326, opened 2022-10-08
SplashPotion: Use `getCollidingEntities()` for effect propagation

IvanCraft623:
## Introduction
Due to it currently uses `getNearbyEntities` erroneously, the effects is applied to entities that it should not such as spectators, also `isAlive()` is removed because `getCollidingEntities` already checks it.

### DIFF
--- a/src/entity/projectile/SplashPotion.php
+++ b/src/entity/projectile/SplashPotion.php
@@ -98,8 +98,8 @@ class SplashPotion extends Throwable{
 
 		if($hasEffects){
 			if(!$this->willLinger()){
-				foreach($this->getWorld()->getNearbyEntities($this->boundingBox->expandedCopy(4.125, 2.125, 4.125), $this) as $entity){
-					if($entity instanceof Living && $entity->isAlive()){
+				foreach($this->getWorld()->getCollidingEntities($this->boundingBox->expandedCopy(4.125, 2.125, 4.125), $this) as $entity){
+					if($entity instanceof Living){
 						$distanceSquared = $entity->getEyePos()->distanceSquared($this->location);
 						if($distanceSquared > 16){ //4 blocks
 							continue;


jasonw4331 (approved):

src/entity/projectile/SplashPotion.php
```diff
@@ -98,8 +98,8 @@ protected function onHit(ProjectileHitEvent $event) : void{
 
 		if($hasEffects){
 			if(!$this->willLinger()){
-				foreach($this->getWorld()->getNearbyEntities($this->boundingBox->expandedCopy(4.125, 2.125, 4.125), $this) as $entity){
-					if($entity instanceof Living && $entity->isAlive()){
```
dktapps, line 102: I don't see anywhere else this is checked.

dktapps:
I'm not sure this is correct. getCollidingEntities is just getNearbyEntities with extra checks for collisions: https://github.com/pmmp/PocketMine-MP/blob/stable/src/world/World.php#L1909

The effect cloud released by a splash potion does not necessarily have to collide with an entity in order to apply effects to it, as far as I understand.

src/entity/projectile/SplashPotion.php
IvanCraft623: `World::getCollidingEntities()` calls `Entity::canBeCollidedWith()` this checks that the entity is alive:
https://github.com/pmmp/PocketMine-MP/blob/stable/src/world/World.php#L1909
https://github.com/pmmp/PocketMine-MP/blob/stable/src/entity/Entity.php#L735

IvanCraft623:
> I'm not sure this is correct. getCollidingEntities is just getNearbyEntities with extra checks for collisions

I know, it is exactly those extra checks that are missing and are causing unexpected behavior.

dktapps (approved):
LGTM, I can't think of any reason to refuse this.

I think the whole "colliding" thing is way too ambiguous all over (blame @shoghicp), but that's a discussion for another time.

PR #5328, opened 2022-10-08
Correct Item IDs from Music Disc

note3crafter:
Correct Item IDs from Music Disc from here https://minecraftitemids.com/types/music-disc

the ids are reserverd by another items...

### DIFF
--- a/src/item/ItemIds.php
+++ b/src/item/ItemIds.php
@@ -718,18 +718,18 @@ final class ItemIds{
 	public const SWEET_BERRIES = 477;
 
 	public const COMPOUND = 499;
-	public const RECORD_13 = 500;
-	public const RECORD_CAT = 501;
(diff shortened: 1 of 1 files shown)


dktapps:
The IDs are not the same between Java and Bedrock. The ones you've linked are the Java IDs.

note3crafter:
ah yes i had the right ids found

    public const RECORD_13 = 534;
    public const RECORD_CAT = 535;
    public const RECORD_BLOCKS = 536;
    public const RECORD_CHIRP = 537;
    public const RECORD_FAR = 538;
    public const RECORD_MALL = 539;
    public const RECORD_MELLOHI = 540;
    public const RECORD_STAL = 541;
    public const RECORD_STRAD = 542;
    public const RECORD_WARD = 543;
    public const RECORD_11 = 544;
    public const RECORD_WAIT = 545;

PR #5333, opened 2022-10-12
Introduce Item Lock API mainly for Client-Side inventory action prevention

yuyaprgrm:
## Introduction
Currently, PocketMine-MP have no API for item lock api provided via item's NBT(minecraft:item_lock). 
The purpose of this PR is to provide an API for item locking and implement logic for server-side validation

### Relevant issues
N/A

### API changes
- add `Item::setLockMode(?ItemLockMode $lockMode) : self`
- add `Item::getLockMode() : ?ItemLockMode`
- add `ItemLockMode` enum

### Behavioural changes
N/A

## Backwards compatibility
N/A

## Follow-up
Currently, value used in NBT is hardcoded. This value may be defined in BedrockProtocol.

## Tests
Check whether it works correctly with ScriptPlugin below.
```php
<?php

namespace famima65536\ItemLockTest;

use pocketmine\item\ItemLockMode;

/**
 * @name ItemLockTest
 * @main famima65536\ItemLockTest\Main
 * @version 0.1.0
 * @api 4.9.0
 * @description A plugin makes sure that Item::setLockMode() and Item::getLockMode() works well.
 * @author famima65536
 */


		$this->getServer()->getPluginManager()->registerEvent(PlayerJoinEvent::class, function(PlayerJoinEvent $event) : void{
			$event->getPlayer()->getInventory()->setContents([
				VanillaItems::STICK()->setLockMode(ItemLockMode::SLOT()),
				VanillaItems::APPLE()->setLockMode(ItemLockMode::INVENTORY()),
			]);
	}
}
```
image
image


src/inventory/transaction/action/SlotChangeAction.php
```diff
@@ -76,6 +77,12 @@ public function validate(Player $source) : void{
 		if($this->targetItem->getCount() > $this->inventory->getMaxStackSize()){
 			throw new TransactionValidationException("Target item exceeds inventory max stack size");
 		}
+		if($this->targetItem->getLockMode()?->equals(ItemLockMode::IN_SLOT()) === true){
```
dktapps: This check doesn't make any sense.

src/inventory/transaction/action/DropItemAction.php
```diff
@@ -45,6 +45,9 @@ public function validate(Player $source) : void{
 		if($this->targetItem->getCount() > $this->targetItem->getMaxStackSize()){
 			throw new TransactionValidationException("Target item exceeds item type max stack size");
 		}
+		if($this->targetItem->getLockMode() !== null){
+			throw new TransactionValidationException("Target item is locked in inventory or slot");
+		}
```
dktapps: I don't think this check is necessary. If the origin item is actually locked in a slot, the SlotChangeAction changing it to air will be invalid anyway.

src/item/Item.php
```diff
@@ -320,6 +334,12 @@ protected function deserializeCompoundTag(CompoundTag $tag) : void{
 				$this->canDestroy[$entry->getValue()] = $entry->getValue();
 			}
 		}
+		//TODO: define hardcoded value in BedrockProtocol or BedrockData?
+		$this->lockMode = match($tag->getByte("minecraft:item_lock", 0)){
```
dktapps: This must use a constant (see the top of the file).

src/item/Item.php
```diff
@@ -320,6 +334,12 @@ protected function deserializeCompoundTag(CompoundTag $tag) : void{
...
+			1 => ItemLockMode::IN_SLOT(),
+			2 => ItemLockMode::IN_INVENTORY(),
```
dktapps: These hardcoded values should be constants.

src/item/Item.php
```diff
@@ -377,6 +397,15 @@ protected function serializeCompoundTag(CompoundTag $tag) : void{
 		}else{
 			$tag->removeTag("CanDestroy");
 		}
+		if($this->lockMode !== null){
+			$tag->setByte("minecraft:item_lock", match($this->lockMode->id()){
```
dktapps: As above.

src/item/Item.php
```diff
@@ -377,6 +397,15 @@ protected function serializeCompoundTag(CompoundTag $tag) : void{
...
+				ItemLockMode::IN_SLOT()->id() => 1,
+				ItemLockMode::IN_INVENTORY()->id() => 2,
```
dktapps: and again here

src/item/ItemLockMode.php
```diff
@@ -0,0 +1,46 @@
...
+ * @method static ItemLockMode IN_INVENTORY()
+ * @method static ItemLockMode IN_SLOT()
+ */
+final class ItemLockMode{
+	use EnumTrait;
+
+	protected static function setup() : void{
+		self::registerAll(
+			new self("in_slot"),
+			new self("in_inventory")
```
dktapps: These IN_ prefixes seem redundant.

dktapps (changes requested):
The server-side validation for this is incomplete. Since the current transaction system does not allow actions to be tracked when moving from one inventory to another, validation for this can't be easily implemented. This means that this effectively becomes a cosmetic-only change.

src/inventory/transaction/action/DropItemAction.php
famima65536: This is necessary because without this validation, target item can be dropped when target item is locked in inventory.

yuyaprgrm:
Progress: this feature looks to be implementable with new inventory transaction system in current minor-next branch.
May I work this PR with rebasing and force-pushing?

dktapps:
The core transaction system remains unchanged, and implementing core logic in the network layer is pretty dubious

dktapps:
Blocked waiting on #6505

dktapps:
~~I had a think about this, and realized it can probably be done without a revamp of the whole system.~~

~~Basically, [this code](https://github.com/pmmp/PocketMine-MP/blob/cd2a1fdc1d2c327894b3b489937e9d6abb34e472/src/inventory/transaction/InventoryTransaction.php#L150) would have to split `$needItems` and `$haveItems` into `$needItemsByInventory` and `$haveItemsByInventory` according to which inventory they came from (if they came from a `SlotChangeAction`).~~

~~Then, [this code](https://github.com/pmmp/PocketMine-MP/blob/cd2a1fdc1d2c327894b3b489937e9d6abb34e472/src/inventory/transaction/InventoryTransaction.php#L169) would have to be adapted to only look in the subset of items that are locked to the same inventory.~~

~~This way, the transaction would only balance if the same number of inventory-locked items were consumed & created by the same inventory.~~

~~Seems doable but not for the faint of heart.~~

Never mind, seems I misunderstood the function of the inventory lock. Apparently it allows moving from a chest -> player's inventory but not the other way round. It also allows moving from cursor <-> player inventory.

dktapps (dismissed):

dktapps:
PR is basically done but I think the terminology needs to be updated. `ItemLockMode` doesn't give much away, particularly since the intended function is actually to lock items specifically into a _player's inventory_. Without improvement, people might think this feature works on general inventories, like I did.

PR #5345, opened 2022-10-16
Damage anvils when they fall

IvanCraft623:
## Introduction
When hit the ground the anvil is damaged if appropriate.

### Relevant issues

* Work around #4437

### API changes
The following method has been added:
-  `Fallable::onHitGround() : ?Block`

## Tests
https://youtu.be/_nCqGEachSI

- When hitting the ground the anvils are likely to get damaged
- If is a torch underneath, the probability of damage to the item is preserved and the damaged item is dropped if necessary.


src/block/utils/FallableTrait.php
```diff
@@ -54,4 +54,11 @@ public function onNearbyBlockChange() : void{
 			$fall->spawnToAll();
 		}
 	}
+
+	public function onHitGround(FallingBlock $blockEntity) : ?Block{
+		if($this instanceof Block){
```
dktapps:
```suggestion
		if(!$this instanceof Block){
```
I really don't like this assertion though...

src/block/Anvil.php
```diff
@@ -95,4 +98,15 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
...
+
+	public function onHitGround(FallingBlock $blockEntity) : ?Block{
+		if(lcg_value() < 0.05 + (round($blockEntity->getFallDistance()) - 1) * 0.05){
+			if($this->damage !== self::VERY_DAMAGED){
+				$this->damage = $this->damage + 1;
+			}else{
+				return null;
+			}
+		}
+		return $this;
```
dktapps: Is there any particular reason to return a Block here? It seems to me that a boolean would be sufficient.

IvanCraft623: `Returns the block resulting from hitting the ground, null will cause no block to be placed`
Allow to change the resulting block, not necessary for this implementation but likely in later implementations.

dktapps: Such as?

IvanCraft623: hmm, no block in vanilla, I will do your suggestion.

dktapps (approved):

PR #5348, opened 2022-10-17
Implement FallingBlock missing sounds

IvanCraft623:
## Introduction
- Anvils hitting the ground now plays `AnvilFallSound`.
- When the block cannot be placed and is dropped as an item, `BlockBreakSound` is played.

### Relevant issues
* Work around #4437

### API changes
The following method has been added:
- `Fallable::getFallSound() : ?Sound`

## Backward Compatibility
`Fallable::getFallSound()` was added to `FallableTrait` to avoid pollution in children clesses of `Fallable`

## Tests
https://youtu.be/JVi7dG7Pc7c


src/block/utils/Fallable.php
```diff
@@ -33,4 +34,9 @@ interface Fallable{
 	 * Return null if you don't want to change the usual behaviour.
 	 */
 	public function tickFalling() : ?Block;
+
+	/**
+	 * Returns the sound that will be played when FallingBlock hits the ground.
+	 */
+	public function getFallSound() : ?Sound;
```
dktapps: This is BC-breaking and must be targeted at `next-major`.

IvanCraft623: There is a default implementation on FallableTrait: https://github.com/IvanCraft623/PocketMine-MP/blob/fallingblock-sounds/src/block/utils/FallableTrait.php#L63, so existing code should still work.

dktapps: You can't assume that implementations will use this trait.

IvanCraft623: done.

src/entity/object/FallingBlock.php
```diff
@@ -128,14 +129,19 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
...
 				if(!$block->canBeReplaced() || !$world->isInWorld($pos->getFloorX(), $pos->getFloorY(), $pos->getFloorZ()) || ($this->onGround && abs($this->location->y - $this->location->getFloorY()) > 0.001)){
 					$world->dropItem($this->location, $this->block->asItem());
+					$world->addSound($pos->add(0.5, 0.5, 0.5), new BlockBreakSound($blockResult));
 				}else{
-					$ev = new EntityBlockChangeEvent($this, $block, $blockTarget ?? $this->block);
+					$ev = new EntityBlockChangeEvent($this, $block, $blockResult);
 					$ev->call();
 					if(!$ev->isCancelled()){
 						$world->setBlock($pos, $ev->getTo());
+						if($this->onGround && $blockResult instanceof Fallable && ($sound = $blockResult->getFallSound()) !== null){
```
dktapps: Probably it would make more sense to use `$ev->getTo()` for the result, so that the correct sounds are played if the block is changed by a plugin.

src/block/utils/Fallable.php
dktapps: I think `getLandSound()` would be more clear. Coming back to this just now, I briefly thought that "fall sound" meant "sound made while falling".

src/entity/object/FallingBlock.php
IvanCraft623: done

src/block/utils/Fallable.php
IvanCraft623: > I think `getLandSound()` would be more clear. Coming back to this just now, I briefly thought that "fall sound" meant "sound made while falling".

done

dktapps (approved):

PR #5355, opened 2022-10-19
Implement new issue forms template

ShockedPlot7560:
## Introduction

Use of the forms proposed by GitHub for the issues.  
This will allow to gain in rigor by framing the issues a bit better and by removing the markdown syntax for some things.  
Documentation can be found [here](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms)

## Changes

Conversion of the current template with this new format set up by github

### DIFF
--- a/src/PocketMine.php
+++ b/src/PocketMine.php
@@ -39,11 +39,14 @@ namespace pocketmine {
 	use function extension_loaded;
 	use function function_exists;
 	use function getcwd;
+	use function is_dir;
+	use function mkdir;
 	use function phpversion;
 	use function preg_match;
 	use function preg_quote;
 	use function realpath;
 	use function version_compare;
+	use const DIRECTORY_SEPARATOR;
 
 	require_once __DIR__ . '/VersionInfo.php';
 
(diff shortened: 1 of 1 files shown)


dktapps:
My observations so far, using a fork:

- Overall these templates are very space-inefficient (compare [this](https://github.com/dktapps-pm-pl/ShockedPlot-PocketMine-MP/issues/2) and [this](https://github.com/dktapps/PocketMine-MP/issues/1))
- The bug report template needs a "Description" box. The "expected" and "actual" sections should be guidelines, rather than required sections (some bugs require detailed explanation rather than expected/actual, and expected/actual will often just waste space)
- The checkboxes cause every bug report to show with "1 of 3 tasks completed" on the list, which is kinda annoying

ShockedPlot7560:
Indeed, I see the problem of taking space and checkboxes, I'll do some tests on my side to try to fix that, otherwise this pull request will be useless as long as github doesn't propose a way to fix that.

ShockedPlot7560:
Currently, there is no other way to format data output from fields, it will necessarily be at a new line.

I have however proposed the change [here ](https://github.com/orgs/community/discussions/37651)

ShockedPlot7560:
I didn't do it on another branch, I will recreate it later

PR #5357, opened 2022-10-21
Add ServerCrashEvent

MrHoller:
## Introduction
A new event is needed so that plugins can set new actions during a server crash and receive crash data


dktapps:
This has been proposed before and I've yet to hear a good reason for it. There's nothing that can be done with this that can't be done better with a server daemon.

MrHoller:
> This has been proposed before and I've yet to hear a good reason for it. There's nothing that can be done with this that can't be done better with a server daemon.

For example, to send a notification to the creator about a crash, send the file there as well as to maintain statistics

dktapps:
> > This has been proposed before and I've yet to hear a good reason for it. There's nothing that can be done with this that can't be done better with a server daemon.
> 
> For example, to send a notification to the creator about a crash, send the file there as well as to maintain statistics

As I said, this can be better done with a server daemon, which requires no changes to PM whatsoever.

NhanAZ:
`$event->cancel()` 🤤

SOF3:
the general reason against this is that plugins are already at a degenerate state during server crash, e.g. there is a chance that the event handler lists are malformed and cannot trigger crash event correctly. crash handler should strictly be limited to code paths known to be infallible and depend on as few states as possible.

If you want to add a crash handler, simply add it to your start.sh as a step after server process completes.

PR #5359, opened 2022-10-21
Remove unnecessary Cancellable implementation

zSALLAZAR:
## Introduction
Remove unnecessary Cancellable implementation


zSALLAZAR:
Cancellable is already implemented in BaseBlockChangeEvent

SOF3:
Does it really make sense to implement Cancellable on BaseBlockChangeEvent generically?

dktapps:
I don't see why not, but it can't be changed before PM5 anyway.

zSALLAZAR:
Why did you add the label "Waiting on Author"?

PR #5360, opened 2022-10-21
'isFull' method implemented in 'PlayerInventory'

iBryanidk:
## Introduction
I think it is a useful function. I was continuing with my plugin and noticed that something like this didn't exist, and I thought it was a good idea


dktapps:
This isn't really useful. It doesn't tell you whether or not you can still add an item.

A better way is to just use addItem() and check the results.

PR #5366, opened 2022-10-26
fix if else order in number

xerenahmed:


dktapps:
Interesting, I wonder why PHPStan doesn't pick up on this.

Issue #5372, opened 2022-10-30
STUN support


SOF3:
STUN is a protocol that allows users to expose their server to the public network without configuring their NAT manually (i.e. port forwarding). This is achieved by the following steps:

1. Create a UDP socket `$sk` without binding to a specific port.
2. Send a packet from `$sk` to a STUN server that runs in the public network.
  - There are some [public STUN servers](https://gist.github.com/mondain/b0ec1cf5f60ae726202e).
  - We can also consider running an official STUN server for PocketMine users, since such a server is stateless, uses little CPU and is expected to have low bandwidth usage unless DDoS'ed. Users can be address-rate-limited if this is constantly abused by public server hosts or crash-looping servers somehow (although rate limiting will consume more CPU and memory).
4. The STUN server returns you with the IP address and port as observed from their server
5. The user sees the IP and port received from the STUN server, and they tell their "friends" to connect to this IP and port through their own methods.
  - This means this method only works if players joining the server can receive address updates promptly, i.e. this is not practical for public servers. But hosting public servers over a NAT network (such as home WiFi) has always been inappropriate anyway.

The approach above works for users under a [full cone NAT](https://en.wikipedia.org/wiki/Full_cone_NAT). However, if the players joining the server are also known (again, this is not for public servers, but for cases like connecting to your own server in a network with LAN discovery disabled), we can add the following steps to support address-restricted cone NATs as well:

6. Ask your friend to tell you their public IP from their phone, and type this into the server (e.g. through a command).
7. Send a packet from `$sk` to any port on this IP.

This is currently not feasible with the design of RakLib, which manages its own creation of UDP socket. There are two possible approaches:
- Implement STUN support inside RakLib. This is probably not desirable since RakLib is just a library for RakNet protocol implementation.
- Expose an interface `DatagramSocketFactory`, which is passed into the worker thread to construct the UDP socket. The user (i.e. PocketMine-MP, or a STUN support plugin if RakLib server is initialized after plugin initialization) can implement the STUN logic and pass the resultant UDP socket to RakLib.

PR #5373, opened 2022-10-30
Allow customizing socket instantiation

SOF3:
## Introduction

Allows plugins handling `NetworkInterfaceRegisterEvent` to change the way the RakLib `Socket` is constructed. This allows implementing features like #5372.

### API changes

- `RakLibServer` constructor signature changed to accept a `SocketFactory` instead of an `InternetAddress`.
- Expose methods to set the socket factory before the RakLib thread starts.

### Behavioural changes

No direct changes without plugins calling the new APIs.

## Backwards compatibility

- `RakLibServer` constructor parameter type changed.

## Tests

N/A

### DIFF
--- a/src/network/mcpe/raklib/RakLibInterface.php
+++ b/src/network/mcpe/raklib/RakLibInterface.php
@@ -37,6 +37,7 @@ use pocketmine\network\PacketHandlingException;
 use pocketmine\Server;
 use pocketmine\snooze\SleeperNotifier;
 use pocketmine\utils\Utils;
+use raklib\generic\Socket;
 use raklib\generic\SocketException;
 use raklib\protocol\EncapsulatedPacket;
 use raklib\protocol\PacketReliability;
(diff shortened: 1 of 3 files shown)


dktapps:
I think this could be reduced to just a simple closure.

ShockedPlot7560:
> I think this could be reduced to just a simple closure.

Can't closures be passed in a thread, no?

dktapps:
They can, subject to certain conditions:
- $this must be a ThreadSafe, or the closure must be declared `static`
- it can't use() non-thread-safe variables
- it can't use() by reference
- it can't have local static variables

dktapps:
This is stale and would need to be rewritten to work with pmmpthread v6, which has different behaviour around sending objects between threads. This version wouldn't have worked anyway due to the use of an anonymous class, which can't be passed between threads in any version of pmmpthread or pthreads.

dktapps:
If anyone wants to take another shot at this with modern-day PM, please feel free.

PR #5380, opened 2022-11-01
Fixed totem activating when having 1HP instead of 0HP

JavierLeon9966:
## Introduction

In a vanilla server/world (BDS) when you are less than or equal to half of a heart (1HP) the totem doesn't activate, only when reaching 0HP. This has been tested with full diamond armor, being half of a heart using poison and being attacked by a silverfish.

## Tests

Expected Behavior: https://3v4l.org/SKi83
Actual Behavior: https://3v4l.org/fT4v0

### DIFF
--- a/src/entity/Human.php
+++ b/src/entity/Human.php
@@ -356,7 +356,7 @@ class Human extends Living implements ProjectileSource, InventoryHolder{
 			&& ($this->inventory->getItemInHand() instanceof Totem || $this->offHandInventory->getItem(0) instanceof Totem)){
 
 			$compensation = $this->getHealth() - $source->getFinalDamage() - 1;
-			if($compensation < 0){
+			if($compensation <= -1){
 				$source->setModifier($compensation, EntityDamageEvent::MODIFIER_TOTEM);
 			}
 		}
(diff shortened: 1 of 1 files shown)


src/entity/Human.php
```diff
@@ -356,7 +356,7 @@ public function applyDamageModifiers(EntityDamageEvent $source) : void{
 			&& ($this->inventory->getItemInHand() instanceof Totem || $this->offHandInventory->getItem(0) instanceof Totem)){
 
 			$compensation = $this->getHealth() - $source->getFinalDamage() - 1;
```
dktapps, line 358: Seems to me like it should be sufficient to remove the `- 1` here. Though it looks to have been intentionally put there.

src/entity/Human.php
```diff
@@ -365,7 +365,7 @@ public function applyDamageModifiers(EntityDamageEvent $source) : void{
 	protected function applyPostDamageEffects(EntityDamageEvent $source) : void{
 		parent::applyPostDamageEffects($source);
 		$totemModifier = $source->getModifier(EntityDamageEvent::MODIFIER_TOTEM);
-		if($totemModifier < 0){ //Totem prevented death
+		if($totemModifier <= -1){ //Totem prevented death
```
dktapps: This appears to be unnecessary.

src/entity/Human.php
JavierLeon9966: the ``- 1`` is there to set the player back to 1HP when the totem activates

src/entity/Human.php
JavierLeon9966: Yeah, since it if is less than 0 it would still mean the totem activated

PR #5381, opened 2022-11-02
switch replaced to match

globalarray:
## Introduction
The match expression branches evaluation based on an identity check of a value. Similarly to a switch statement, a match expression has a subject expression that is compared against multiple alternatives. Unlike switch, it will evaluate to a value much like ternary expressions. Unlike switch, the comparison is an identity check (===) rather than a weak equality check (==). Match expressions are available as of PHP 8.0.0.

## Tests

Before: https://3v4l.org/OuGRm#v8.0.25 (switch)
Now: https://3v4l.org/PlsIs#v8.0.25 (match)
Match was faster than switch by 8.29ms as expected.

### DIFF
--- a/src/block/BlockLegacyIdHelper.php
+++ b/src/block/BlockLegacyIdHelper.php
@@ -34,203 +34,133 @@ use pocketmine\utils\AssumptionFailedError;
 final class BlockLegacyIdHelper{
 
 	public static function getWoodenFloorSignIdentifier(TreeType $treeType) : BID{
-		switch($treeType->id()){
-			case TreeType::OAK()->id():
-				return new BID(Ids::SIGN_POST, 0, ItemIds::SIGN, TileSign::class);
-			case TreeType::SPRUCE()->id():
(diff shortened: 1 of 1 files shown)


ddosnikgit:
## Introduction
The match expression branches evaluation based on an identity check of a value. Similarly to a switch statement, a match expression has a subject expression that is compared against multiple alternatives. Unlike switch, it will evaluate to a value much like ternary expressions. Unlike switch, the comparison is an identity check (===) rather than a weak equality check (==). Match expressions are available as of PHP 8.0.0.

## Tests

Before: https://3v4l.org/OuGRm#v8.0.25 (switch)
Now: https://3v4l.org/PlsIs#v8.0.25 (match)
Match was faster than switch by 8.29ms as expected.

dktapps:
Doing this for performance is premature optimisation. This code is in a cold path.
There is no practical benefit to this change, and it creates an unnecessary merge conflict with `next-major`.

In addition, this change is not suitable for `stable`. https://github.com/pmmp/PocketMine-MP/blob/stable/CONTRIBUTING.md#choosing-a-target-branch

olovink:
java the best

Issue #5358, opened 2022-10-21
Missing sounds when a projectile strikes an amethyst block


dktapps:
### Issue description

A sound should be played when an amethyst block is struck by a projectile. It looks like this was never implemented.

### Steps to reproduce the issue
1. Shoot an arrow at an amethyst block
2. Notice there is no "ting" sound, only the sound of the arrow colliding with the block

Might be easier to listen for with other projectiles.

### OS and versions

* PocketMine-MP: 5.0.0-ALPHA5+dev (git hash: d1d5020c5381f894ab132737efc9b141a484f517)
* PHP: 8.1.11
* Using JIT: no

PR #5382, opened 2022-11-02
Fix missing sound when a projectile strikes an amethyst block

zSALLAZAR:
### Relevant issues

#5358 

## Tests

https://user-images.githubusercontent.com/59490940/199513392-c85869e8-c032-4366-9432-0300b3aa2ba3.mp4


dktapps:
Are you sure this matches vanilla? The sound is barely audible in this video clip even at max volume.

zSALLAZAR:
Now it should match with vanilla

https://user-images.githubusercontent.com/59490940/199553138-ad1efe5d-4206-41b3-b98f-ce3e939c6051.mp4

dktapps:
A quick check in IDA suggests that both the punch sound _and_ the chime sound are played simultaneously, though I'm not sure why.

Issue #5168, opened 2022-07-16
Missing sounds when interacting with item frames


dktapps:
### Issue description
The sounds for inserting an item into an item frame, rotating it, and removing it are missing.

### Steps to reproduce the issue
1. Do any of the above actions.
2. Notice there's no sound.

### OS and versions

* PocketMine-MP: 4.6.0 / 5.0.0-ALPHA2

PR #5383, opened 2022-11-02
Fix missing sounds when interacting with item frames

zSALLAZAR:
### Relevant issues

#5168 

## Tests

https://user-images.githubusercontent.com/59490940/199596126-a545f8cc-c153-4112-a344-55cbea54ea20.mp4


src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -736,7 +737,9 @@ public function handleItemFrameDropItem(ItemFrameDropItemPacket $packet) : bool{
 		$blockPosition = $packet->blockPosition;
 		$block = $this->player->getWorld()->getBlockAt($blockPosition->getX(), $blockPosition->getY(), $blockPosition->getZ());
 		if($block instanceof ItemFrame && $block->getFramedItem() !== null){
-			return $this->player->attackBlock(new Vector3($blockPosition->getX(), $blockPosition->getY(), $blockPosition->getZ()), $block->getFacing());
+			$this->player->getWorld()->addSound($pos = new Vector3($blockPosition->getX(), $blockPosition->getY(), $blockPosition->getZ()), new ItemFrameRemoveItemSound());
```
JavierLeon9966: This should be in `ItemFrame->onAttack()`, you don't know if the item frame is empty

JavierLeon9966 (changes requested):

src/network/mcpe/handler/InGamePacketHandler.php
JavierLeon9966: Now that I see that code above it it does check, but you don't know if a plugin has canceled the interact action

zSALLAZAR: You're right
I'll change it tomorrow

src/block/ItemFrame.php
```diff
@@ -162,6 +171,7 @@ public function onAttack(Item $item, int $face, ?Player $player = null) : bool{
 	public function onNearbyBlockChange() : void{
 		if(!$this->getSide(Facing::opposite($this->facing))->isSolid()){
 			$this->position->getWorld()->useBreakOn($this->position);
+			$this->position->getWorld()->addSound($this->position, new ItemFrameRemoveSound());
```
dktapps: `ItemFrameDestroySound` would be more consistent, though I don't see why item frames need a dedicated sound for this. We don't play specific destroy sounds or particles for nearby updates for any other block.

dktapps:
Retargeting this to next-minor since it's adding new APIs.

src/block/ItemFrame.php
zSALLAZAR: If you remove the block behind the itemframe in vanilla a itemframe destroy sound is played, but not in Pocketmine, so I added this sound.

dktapps:
I've removed the onNearbyBlockChange changes since those don't directly pertain to interaction, and the missing sounds there are part of a larger issue which should be solved in a more generic way. The rest of the PR is good to merge though.

PR #5384, opened 2022-11-03
Implement a `Living::getDisplayName()`

IvanCraft623:
## Introduction
While working on another PR I realized that the code to get the death message was really a mess,` if-elseif` and `ternary operators`, all over the place to get only the name of the entity to be displayed. Without a method like this, when adding more specific translations it really becomes hell.

### API changes
The following api method has been added:
- `Living::getDisplayName() : string`

### DIFF
--- a/src/entity/Living.php
+++ b/src/entity/Living.php
@@ -184,6 +184,13 @@ abstract class Living extends Entity{
 		$this->attributeMap->add($this->absorptionAttr = AttributeFactory::getInstance()->mustGet(Attribute::ABSORPTION));
 	}
 
+	/**
+	 * Returns the display name of this entity to use in stuff like death messages.
+	 */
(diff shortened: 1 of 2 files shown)


src/entity/Living.php
```diff
@@ -184,6 +184,13 @@ protected function addAttributes() : void{
 		$this->attributeMap->add($this->absorptionAttr = AttributeFactory::getInstance()->mustGet(Attribute::ABSORPTION));
 	}
 
+	/**
+	 * Returns the display name of this entity to use in stuff like death messages.
```
dktapps:
```suggestion
	 * Returns the name used to describe this entity in chat and command outputs.
```
In the future we'll most likely use this for things like `/kill` and the like.

IvanCraft623: done

PR #5386, opened 2022-11-03
Show death message on death screen

IvanCraft623:
## Introduction
This PR sends the `DeathInfoPacket` packet implemented in 1.19.10 to display the death message

## Backwards compatibility
Because `NetworkSession::onServerDeath()` now requires the `$deathMessage` parameter it could be considered BC-Break but since the protocol is not part of the API, there should be no problem.

## Tests
https://cdn.discordapp.com/attachments/430364566027763744/1037818307870478346/Death_screen_test.mp4


src/network/mcpe/handler/DeathPacketHandler.php
```diff
@@ -44,6 +48,16 @@ public function setUp() : void{
 			RespawnPacket::SEARCHING_FOR_SPAWN,
 			$this->player->getId()
 		));
+
+		/** @var string[] $parameters */
+		$parameters = [];
+		if($this->deathMessage instanceof Translatable){
+			$message = $this->deathMessage->getText();
```
dktapps: This isn't reliable. The string might not be present in the vanilla files.

You should implement similar logic to what's done here: https://github.com/pmmp/PocketMine-MP/blob/fe982c697bec3265c4222b73fe5ed17c9ef3b1ce/src/player/Player.php#L1985-L1993

src/player/Player.php
```diff
@@ -2272,13 +2272,14 @@ protected function onDeath() : void{
 			$this->xpManager->setXpAndProgress(0, 0.0);
 		}
 
-		if($ev->getDeathMessage() != ""){
+		$deathMessage = $ev->getDeathMessage();
+		if($deathMessage != ""){
```
dktapps: Don't change unrelated code. You haven't changed this consistently, either.

dktapps (commented):
Looks mostly OK, I haven't tested this yet though.

dktapps (approved):
Needs some tidying around the translation handling, but that's a job for another time.

PR #5391, opened 2022-11-05
BaseInventory: Validate given array to make sure correct items are given.

alvin0319:
## Introduction
`BaseInventory::setContents()` does not validate the given array, so when the server crashes due to an invalid array given, developers may hard to find what went wrong.

### Relevant issues
N/A


dktapps:
For posterity: Validation of array types this way is only necessary if the method in question is exposed to plugins.

Internal-use stuff doesn't require validation since PHPStan will take care of it (and runtime validation hurts performance).

PR #5395, opened 2022-11-06
Added keep on death methods for items

JavierLeon9966:
## Introduction

We need something that can keep the items of a player into their inventory when they die, for example an ability plugin. This pull request will add methods to allow plugins keep items on death and makes use of the vanilla component `minecraft:keep_on_death`.
### API changes

- `Item->keepOnDeath()`
- `Item->setKeepOnDeath()`
## Tests

https://user-images.githubusercontent.com/58715544/200150544-35dd71b8-0c9d-4129-8559-96c98a6c5e28.mp4

### DIFF
--- a/src/item/Item.php
+++ b/src/item/Item.php
@@ -226,6 +226,9 @@ class Item implements \JsonSerializable{
 		}
 	}
 
+	/**
+	 * Returns whether players will retain this item on death. If a non-player dies it will be excluded from the drops.
+	 */
 	public function keepOnDeath() : bool{
 		return $this->keepOnDeath;
 	}
--- a/src/item/Item.php
+++ b/src/item/Item.php
@@ -65,6 +65,8 @@ class Item implements \JsonSerializable{
 	public const TAG_DISPLAY_NAME = "Name";
 	public const TAG_DISPLAY_LORE = "Lore";
 
+	public const TAG_KEEP_ON_DEATH = "minecraft:keep_on_death";
+
 	private ItemIdentifier $identifier;
 	private CompoundTag $nbt;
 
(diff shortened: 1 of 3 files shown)


src/player/Player.php
```diff
@@ -2257,13 +2258,13 @@ protected function onDeath() : void{
 
 			if($this->inventory !== null){
 				$this->inventory->setHeldItemIndex(0);
-				$this->inventory->clearAll();
+				$this->inventory->setContents(array_filter($this->inventory->getContents(), fn(Item $item) => $item->keepOnDeath()));
```
dktapps: You could use a closure accepting `Inventory` to avoid repeating this logic.

src/player/Player.php
```diff
@@ -2255,15 +2256,16 @@ protected function onDeath() : void{
 				$this->getWorld()->dropItem($this->location, $item);
 			}
 
+			$keepOnDeathItemsForInventory = fn(Inventory $inventory) => $inventory->setContents(array_filter($inventory->getContents(), fn(Item $item) => $item->keepOnDeath()));
```
dktapps: This variable name is a bit unpleasant. `$clearInventory` or something similar would be sufficient.

dktapps (approved):
Looks good, I just need to test it for myself.

src/item/Item.php
```diff
@@ -222,6 +226,14 @@ public function setCanDestroy(array $canDestroy) : void{
 		}
 	}
 
+	public function keepOnDeath() : bool{
```
dktapps, line 232: Please add documentation for what this signifies. I know it seems obvious in the context of the PR, but someone who randomly stumbles across this won't know that this is specifically to make players retain the item (it won't affect other entities with inventories, for example).

Clxne228:
Very good!

PR #5396, opened 2022-11-06
Implement Weeping & Twisting vines

ipad54:
## Introduction
This PR implements Weeping & Twisting vines and their functionality

## Tests
https://www.youtube.com/watch?v=vES8HIk9fHs


src/block/NetherVines.php
```diff
@@ -0,0 +1,178 @@
...
+		}
+	}
+
+	private function grow(?Player $player, int $growthAmount) : bool{
+		$top = $this->seekToTop();
+		$face = $this->getGrowthFace();
+
+		$tx = new BlockTransaction($this->position->getWorld());
+		for($i = 1; $i <= $growthAmount && $top->getSide($face, $i)->canBeReplaced(); $i++){
+			$tx->addBlock($top->getSide($face, $i)->getPosition(), (clone $top)->setAge(min($top->getAge() + $i, self::MAX_AGE)));
```
dktapps: This won't update the block itself, so the age of the block will stay the same.

src/block/NetherVines.php
```diff
@@ -0,0 +1,178 @@
...
+					$item->pop();
+				}
+			}
+			return true;
+		}
+		return false;
+	}
+
+	public function onRandomTick() : void{
+		if(mt_rand(0, 100) <= 10 && $this->age < self::MAX_AGE){
```
dktapps: This should be 0,99 if you're trying to make a 1/100 chance, and it should be <10 if you want a 10% chance.

src/block/NetherVines.php
```diff
@@ -0,0 +1,178 @@
...
+		$entity->resetFallDistance();
+		return false;
+	}
+
+	public function recalculateCollisionBoxes() : array{
+		return [];
+	}
+
+	public function getDropsForCompatibleTool(Item $item) : array{
+		if($item instanceof Shears || mt_rand(0, 100) <= 33){
```
dktapps: BlockToolType should be used here, not instanceof.

dktapps (commented):
Looks nice, but needs a bit of polish.

src/block/NetherVines.php
ipad54: What do you mean? The age of the block is updating

Video: https://www.youtube.com/watch?v=W74C13sN_Kk

dktapps: I don't think you're understanding me. The age of the bottom block (where the growth originated from) isn't being updated.

dktapps (commented):
The growth logic in this PR is very hard to understand, and doesn't make a lot of sense to me. I would expect it to work the same way as cactus and sugarcane do.

inotflying:
*****, переделывай

celishere:
> *****, переделывай

Че высрал

dktapps:
Please keep the conversation mature and in English only. if you're not here to give constructive feedback then go somewhere else.

whoqmi:
> *****, переделывай

а чё тут градиент не поменял?

celishere:
> > *****, переделывай
> 
> а чё тут градиент не поменял?
> 

рил

Salmon1279:
Дилан дай деняк блин

celishere:
> Дилан дай деняк блин
> 
> 

И мне 😃😃😃😃

src/block/NetherVines.php
```diff
@@ -0,0 +1,190 @@
...
+		return true;
+	}
+
+	public function canClimb() : bool{
+		return true;
+	}
+
+	abstract protected function getGrowthFace() : int;
+
+	protected function getSupportFace() : int{
```
dktapps:
```suggestion
	private function getSupportFace() : int{
```

src/block/NetherVines.php
```diff
@@ -0,0 +1,190 @@
...
+		$entity->resetFallDistance();
+		return false;
+	}
+
+	public function recalculateCollisionBoxes() : array{
+		return [];
+	}
+
+	public function getDropsForCompatibleTool(Item $item) : array{
+		if($item->getBlockToolType() === BlockToolType::SHEARS || mt_rand(0, 100) < 33){
```
dktapps:
```suggestion
		if($item->getBlockToolType() === BlockToolType::SHEARS || mt_rand(0, 99) < 33){
```

src/block/NetherVines.php
```diff
@@ -0,0 +1,190 @@
...
+		if(!$this->getSide($supportFace)->isSameType($this) && $this->age < self::MAX_AGE){
+			$tx->addBlock($this->position, (clone $this)->setAge($this->age + 1));
+		}
+
+		for($i = 1; $i <= $growthAmount && $top->getSide($growthFace, $i)->canBeReplaced(); $i++){
+			$pos = $top->getPosition()->getSide($growthFace, $i);
+			if(!$world->isInWorld($pos->getX(), $pos->getY(), $pos->getZ())){
+				break;
+			}
+			$top = $tx->fetchBlock($pos->getSide($supportFace));
```
dktapps: You're overwriting the $top variable here.

dktapps (changes requested):

src/block/NetherVines.php
ipad54: What about mt_rand(1, 100)? I think it's more logical

dktapps: That would make the condition incorrect.

dktapps:
Please fix the build errors.

src/block/NetherVines.php
```diff
@@ -0,0 +1,191 @@
...
+	private function grow(?Player $player, int $growthAmount) : bool{
+		$top = $this->seekToTop();
+		$world = $this->position->getWorld();
+		$growthFace = $this->getGrowthFace();
+		$supportFace = $this->getSupportFace();
+
+		$tx = new BlockTransaction($world);
+
+		if(!$this->getSide($supportFace)->isSameType($this) && $this->age < self::MAX_AGE){
+			$tx->addBlock($this->position, (clone $this)->setAge($this->age + 1));
```
dktapps: Seems to me like this won't update any of the blocks in between the base and tip, only the block that got the random update. This will lead to inconsistent growth behaviour.

ghost: What's wrong?

src/block/NetherVines.php
```diff
@@ -0,0 +1,188 @@
...
+			if($block instanceof NetherVines){
+				$ev = new BlockGrowEvent($block, (clone $block)->setAge(min($block->getAge() + 1, self::MAX_AGE)));
+				$ev->call();
+
+				if($ev->isCancelled()){
+					break;
+				}
+				$world->setBlock($pos, $ev->getNewState());
+				$grew = true;
+			}
```
dktapps, line 165: I can't make head or tail of this code. It doesn't appear to be actually growing the vines at all, just updating the age of the existing vine blocks if they exist.

ghost: In pm3 growth was normal. Now it's broken

hausemaster3734: In pm3 growth was normal. Now it's broken

dktapps: This code existed in neither PM3 nor PM4. Stop hijacking the conversation.

ipad54: This code is cloning the block, increase his age and setting it on a new position. I don't use Vanilla Blocks here because it may be weeping or twisting vine.

Should i use VanillaBlocks + match() here?

ghost: I wrote this to the fact that the mechanics of growth, which ipad54 showed normal for the first time. Honestly, I did not see any strange growth behavior.

hausemaster3734: I wrote this to the fact that the mechanics of growth, which ipad54 showed normal for the first time. Honestly, I did not see any strange growth behavior.

dktapps: > This code is cloning the block, increase his age and setting it on a new position.

BlockGrowEvent doesn't show that. It shows that the vine block is changing its own age.

This code is very confusing and needs improvement.

src/block/NetherVines.php
```diff
@@ -0,0 +1,185 @@
...
+		$entity->resetFallDistance();
+		return false;
+	}
+
+	protected function recalculateCollisionBoxes() : array{
+		return [];
+	}
+
+	public function getDropsForCompatibleTool(Item $item) : array{
+		if(($item->getBlockToolType() & BlockToolType::SHEARS) !== 0 || mt_rand(0, 99) < 33){
```
dktapps:
```suggestion
		if(($item->getBlockToolType() & BlockToolType::SHEARS) !== 0 || mt_rand(1, 3) === 1){
```

src/block/NetherVines.php
```diff
@@ -0,0 +1,185 @@
...
+					$item->pop();
+				}
+			}
+			return true;
+		}
+		return false;
+	}
+
+	public function onRandomTick() : void{
+		if(mt_rand(0, 100) < 10 && $this->age < self::MAX_AGE){
```
dktapps:
```suggestion
		if(mt_rand(1, 10) === 1 && $this->age < self::MAX_AGE){
```

src/block/NetherVines.php
```diff
@@ -0,0 +1,185 @@
...
+			if($this->getSide($this->getGrowthFace())->canBeReplaced()){
```
dktapps: Is this necessary?

src/block/NetherVines.php
```diff
@@ -0,0 +1,185 @@
...
+		if(!$this->canBeSupportedBy($blockReplace->getSide($this->getSupportFace()))){
+			return false;
+		}
+		$this->age = mt_rand(0, self::MAX_AGE - 1);
+		return parent::place($tx, $item, $blockReplace, $blockClicked, $face, $clickVector, $player);
+	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if($item instanceof Fertilizer){
+			if($this->seekToTop()->getSide($this->getGrowthFace())->canBeReplaced()){
```
dktapps: This is repeating work done by `grow()`.

src/block/NetherVines.php
ipad54: Yes, because without this check all vines can grow (not just the top one)

src/block/NetherVines.php
```diff
@@ -0,0 +1,183 @@
...
+		for($i = 1; $i <= $growthAmount; $i++){
+			$growthPos = $pos->getSide($growthDirection, $i);
+			if(!$world->isInWorld($growthPos->getFloorX(), $growthPos->getFloorY(), $growthPos->getFloorZ()) || !$world->getBlock($growthPos)->canBeReplaced()){
+				break;
+			}
+			$tx->addBlock($growthPos, (clone $top)->setAge(min(++$age, self::MAX_AGE)));
+		}
+
+		$ev = new StructureGrowEvent($top, $tx, $player);
+		$ev->call();
```
dktapps: The BlockTransaction may be empty. If no blocks are to be generated, the event should not be called.

src/block/NetherVines.php
```diff
@@ -0,0 +1,189 @@
...
+
+		for($i = 1; $i <= $growthAmount; $i++){
+			$growthPos = $pos->getSide($growthDirection, $i);
+			if(!$world->isInWorld($growthPos->getFloorX(), $growthPos->getFloorY(), $growthPos->getFloorZ()) || !$world->getBlock($growthPos)->canBeReplaced()){
+				break;
+			}
+			$tx->addBlock($growthPos, (clone $top)->setAge(min(++$age, self::MAX_AGE)));
+		}
+
+		if(count(iterator_to_array($tx->getBlocks())) > 0){
```
dktapps: You could also just have a counter variable instead of doing this unwieldy check.

dktapps (approved):
I've made a couple of adjustments to the code, but it looks ready to go.

Issue #4771, opened 2022-01-25
Allow changing the result item in PlayerBlockPickEvent


dktapps:
## Description
Since the mechanism by which inventory slots are selected (and potentially swapped) for block picking isn't exposed to the API, it's currently cumbersome to customise the outcome of picking a block.

Adding a `setResultItem()` to `PlayerBlockPickEvent` would solve this problem by removing the need for plugins to care about where the resulting item ends up, or whether the player actually has enough of said item.

## Alternative methods
You can replicate this code in a plugin and schedule a task to change the result after the fact, but that's subject to various other potential issues: https://github.com/pmmp/PocketMine-MP/blob/419bb9eba69d05c6959606915280fcb6be945828/src/player/Player.php#L1531-L1548

dktapps:
This turned out to be a big PITA, because it's possible for the picking logic to fail after the event because the result item might not be able to go into the inventory. Since we pre-cancel the event when the picked item can't go into the inventory to begin with, this would result in unpredictable behaviour for plugins - they wouldn't be able to rely on the `isCancelled` to know whether the picking would succeed.

The clearest solution seems to be to remove the pre-cancellation behaviour of `PlayerBlockPickEvent`. This means that `isCancelled` wouldn't be affected by whether the action will succeed or not. Perhaps a method `willSucceed()` could be added to the event to run the check logic. However, this kind of change would be a subtle BC break, so it would have to wait for a major version.

A decision should be made on this before merging #5397, as the behaviour of the two events should eventually become consistent.

PR #5397, opened 2022-11-07
Added entity middle-click feature

### LINKED ISSUE #4766: Add a `PlayerEntityPickEvent`
## Description
There's currently no way for a plugin to handle middle-clicking on an entity. In vanilla, this is used to acquire spawn eggs, which isn't currently implemented in PM. However, plugins may want to use this for other purposes.

## Alternative methods
Handling packets? ah, the irony...

JavierLeon9966:
## Introduction

Added entity middle-click feature, this is a vanilla feature that it hasn't been implemented for so long.

### Relevant issues
* Closes #4766 
### API changes

Added the following methods in `Entity`:
- `Entity->getCleanedNBT()`: Returns a cleaned NBT from the entity, removing tags such as position, motion, etc.
- `Entity->copyDataFromItem()`: Copies the entity NBT from the item into the entity (internal).
- `Entity->getSpawnItem()`: Returns the item that can be used to spawn the entity, by default it returns air if there isn't one.
- `Entity->getPickedItem()`: Returns an item that is able to spawn said entity, with a parameter `$addUserData` to include the entity's NBT.

Added the following methods in `Item`:
- `Item->clearCustomEntityData()`: Clears the NBT from the picked (middle-clicked) entity.
- `Item->setCustomEntityData()`: Sets the NBT of the picked (middle-clicked) entity.
- `Item->getCustomEntityData()`: Returns the NBT of the picked (middle-clicked) entity.

Added the following method in `Player`:
- `Player->pickEntity()`: Equips the middle-clicked (picked) entity in the player's hand with added NBT if requested.

Added the following event:
- `PlayerEntityPickEvent`

## Tests

https://clipchamp.com/watch/nrP09ZXooey


dktapps:
I'd prefer this wasn't dependent on NBT. Bad enough that we have to deal with that for tiles already...

src/entity/Entity.php
```diff
@@ -1535,6 +1557,19 @@ public function despawnFromAll() : void{
 		}
 	}
 
+	/**
+	 * Returns the item that players will equip when middle-clicking on this entity.
+	 * If addUserData is true, additional data may be added, such as name tag, fire ticks, etc.
+	 */
+	public function getPickedItem(bool $addUserData) : Item{
+		$item = ItemFactory::getInstance()->get(ItemIds::SPAWN_EGG, LegacyEntityIdToStringIdMap::getInstance()->stringToLegacy(static::getNetworkTypeId()));
```
dktapps: I can't accept this.

src/player/Player.php
```diff
@@ -1618,6 +1621,45 @@ public function pickBlock(Vector3 $pos, bool $addTileNBT) : bool{
...
+					$this->inventory->setItemInHand($item);
+				}elseif($firstEmpty < $this->inventory->getHotbarSize()){
+					$this->inventory->setItem($firstEmpty, $item);
+					$this->inventory->setHeldItemIndex($firstEmpty);
+				}else{
+					$this->inventory->swap($this->inventory->getHeldItemIndex(), $firstEmpty);
+					$this->inventory->setItemInHand($item);
+				}
+			}
+		}
```
dktapps, line 1739: If you want to reuse this logic, extract it from `pickBlock` into a reusable method. Don't duplicate the code.

dktapps (changes requested):
I don't like that this involves so much in the way of legacy hacks.

src/entity/Entity.php
JavierLeon9966: What could I replace this for? I didn't find anything else that could convert entity identifiers to spawn egg meta values.

dktapps: You shouldn't be using either of those things. The subclasses should declare which spawn egg type should be used.

src/entity/Entity.php
```diff
@@ -1535,6 +1555,26 @@ public function despawnFromAll() : void{
 		}
 	}
 
+	/**
+	 * Returns the item that can be used to spawn this entity, by default it returns air if there isn't one
+	 */
+	public function getSpawnItem() : Item{
```
dktapps: This could just return `null`

src/entity/Entity.php
```diff
@@ -1535,6 +1555,26 @@ public function despawnFromAll() : void{
...
+		return VanillaItems::AIR();
+	}
+
+	/**
+	 * Returns the item that players will equip when middle-clicking on this entity.
+	 * If addUserData is true, additional data may be added, such as name tag, fire ticks, etc.
+	 */
+	public function getPickedItem(bool $addUserData) : Item{
```
dktapps: This could just return `null` if the entity doesn't have an associated item

src/item/Item.php
```diff
@@ -61,6 +61,7 @@ class Item implements \JsonSerializable{
 	public const TAG_ENCH = "ench";
 	public const TAG_DISPLAY = "display";
 	public const TAG_BLOCK_ENTITY_TAG = "BlockEntityTag";
+	public const TAG_ENTITY_TAG = "EntityTag";
```
dktapps: Does this match vanilla?

dktapps (changes requested):
I'm hesitant to merge this because entities currently predominantly use the Java entity NBT format, but items use the Bedrock format. This is already a disaster for tiles.

I'm OK with having the basic entity picking, but I think the NBT stuff should be removed for now until the internals handling can be cleaned up (and preferably find a way to deal with extra data that doesn't require NBT at all).

src/item/Item.php
JavierLeon9966: This matches the Java behavior, but in Bedrock it is a mess. The entire entity NBT is merged with the item NBT

src/entity/Entity.php
```diff
@@ -1535,6 +1543,25 @@ public function despawnFromAll() : void{
...
+	/**
+	 * Returns the item that players will equip when middle-clicking on this entity.
+	 * If addUserData is true, additional data may be added, such as name tag, fire ticks, etc.
+	 */
+	public function getPickedItem() : ?Item{
+		$item = $this->getSpawnItem();
+		if($item === null){
+			return null;
+		}
+		return $item;
```
dktapps: This can be simplified to just return `getSpawnItem()` directly. Perhaps the two separate functions are redundant.

src/entity/Entity.php
```diff
@@ -1535,6 +1543,25 @@ public function despawnFromAll() : void{
...
+	/**
+	 * Returns the item that can be used to spawn this entity, by default it returns null if there isn't one
+	 */
+	public function getSpawnItem() : ?Item{
+		return VanillaItems::AIR();
+	}
+
+	/**
+	 * Returns the item that players will equip when middle-clicking on this entity.
+	 * If addUserData is true, additional data may be added, such as name tag, fire ticks, etc.
```
dktapps:
```suggestion
```

dktapps (commented):
The separation between "spawn item" and "picked item" seems redundant.

jasonw4331:
@JavierLeon9966 any updates for this PR? If not it will be closed when API 5 releases

JavierLeon9966:
Oh, whoops, I forgot about it, but I think I have the changes in my local

src/entity/Entity.php
```diff
@@ -1557,6 +1559,13 @@ public function despawnFromAll() : void{
 		$this->hasSpawned = [];
 	}
 
+	/**
+	 * Returns the item that players will equip when middle-clicking on this entity.
+	 */
+	public function getPickedItem() : ?Item{
+		return VanillaItems::AIR();
```
dktapps: Shouldn't this just return `null`?

dktapps:
This looks basically fine as it is right now, minus a few nits.

I will mention that we'll probably want to apply #4771 to `PlayerEntityPickEvent` too, but that's a problem for another time. The most important thing is for the events to be consistent.

dktapps (dismissed):

dktapps (dismissed):

src/event/player/PlayerEntityPickEvent.php
```diff
@@ -0,0 +1,53 @@
...
+namespace pocketmine\event\player;
+
+use pocketmine\entity\Entity;
+use pocketmine\event\Cancellable;
+use pocketmine\event\CancellableTrait;
+use pocketmine\item\Item;
+use pocketmine\player\Player;
+
+/**
+ * Called when a player middle-clicks on a entity to get an item in creative mode.
```
ipad54:
```suggestion
 * Called when a player middle-clicks on an entity to get an item in creative mode.
```

ipad54 (approved):

dktapps (approved):

PR #5399, opened 2022-11-07
fixed teleport phpdoc tag

Lunarelly:
## Introduction
actual doc doesnt do anything


dktapps:
The whole block is useless anyway.

Lunarelly:
then just remove it tho

PR #5401, opened 2022-11-08
Implement glow lichen

IvanCraft623:
## Introduction
Implements the glow lichen block.

## Folow-up
Implementing spreading by interacting with bone meal.

## Tests
https://cdn.discordapp.com/attachments/430364566027763744/1039563989270663168/SVID_20221108_095506_1.mp4


src/block/GlowLichen.php
```diff
@@ -0,0 +1,175 @@
...
+use pocketmine\item\ItemTypeIds;
+use pocketmine\math\AxisAlignedBB;
+use pocketmine\math\Facing;
+use pocketmine\math\Vector3;
+use pocketmine\player\Player;
+use pocketmine\world\BlockTransaction;
+use function array_key_first;
+use function count;
+
+class GlowLichen extends Transparent{
```
dktapps, line 41: This should probably be using Flowable

src/block/GlowLichen.php
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	public function getFlammability() : int{
+		return 100;
+	}
+
+	/**
+	 * @return array<int, int> $faces
+	 */
+	public function getAvailableFaces() : array{
```
dktapps: This should accept a Position rather than assuming the block's own position is set.

src/block/GlowLichen.php
```diff
@@ -0,0 +1,175 @@
...
+	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
+		$this->faces = $blockReplace instanceof GlowLichen ? $blockReplace->faces : [];
+		$availableFaces = $this->getAvailableFaces();
+
+		if(count($availableFaces) === 0){
+			return false;
+		}
+
+		$opposite = Facing::opposite($face);
+		$placedFace = isset($availableFaces[$opposite]) ? $opposite : array_key_first($availableFaces);
```
dktapps, line 114: This seems suspicious. This will produce unpredictable results for the player, and placement of blocks is rarely unpredictable. My expectation would be that only the clicked face itself would be filled, similarly to vines.

src/data/bedrock/block/convert/BlockObjectToStateSerializer.php
```diff
@@ -983,6 +984,10 @@ private function registerSerializers() : void{
 			})
 				->writeHorizontalFacing($block->getFacing());
 		});
+		$this->map(Blocks::GLOW_LICHEN(), function(GlowLichen $block) : Writer{
+			return Writer::create(Ids::GLOW_LICHEN)
+				->writeInt(StateNames::MULTI_FACE_DIRECTION_BITS, ($block->hasFace(Facing::DOWN) ? BlockLegacyMetadata::MULTI_FACE_DIRECTION_FLAG_DOWN : 0) | ($block->hasFace(Facing::UP) ? BlockLegacyMetadata::MULTI_FACE_DIRECTION_FLAG_UP : 0) | ($block->hasFace(Facing::NORTH) ? BlockLegacyMetadata::MULTI_FACE_DIRECTION_FLAG_NORTH : 0) | ($block->hasFace(Facing::SOUTH) ? BlockLegacyMetadata::MULTI_FACE_DIRECTION_FLAG_SOUTH : 0) | ($block->hasFace(Facing::WEST) ? BlockLegacyMetadata::MULTI_FACE_DIRECTION_FLAG_WEST : 0) | ($block->hasFace(Facing::EAST) ? BlockLegacyMetadata::MULTI_FACE_DIRECTION_FLAG_EAST : 0));
```
dktapps: This is horrifying to read. Please split it into multiple lines.

src/data/bedrock/block/convert/BlockStateToObjectDeserializer.php
```diff
@@ -736,6 +736,16 @@ private function registerDeserializers() : void{
 				->setFacing($in->readHorizontalFacing())
 				->setLit(false);
 		});
+		$this->map(Ids::GLOW_LICHEN, function(Reader $in) : Block{
+			$drectionFlags = $in->readBoundedInt(StateNames::MULTI_FACE_DIRECTION_BITS, 0, 63);
```
dktapps: Spelling error here

src/data/bedrock/block/convert/BlockStateToObjectDeserializer.php
```diff
@@ -736,6 +736,16 @@ private function registerDeserializers() : void{
...
+			return Blocks::GLOW_LICHEN()
+				->setFace(Facing::DOWN, ($drectionFlags & BlockLegacyMetadata::MULTI_FACE_DIRECTION_FLAG_DOWN) !== 0)
+				->setFace(Facing::UP, ($drectionFlags & BlockLegacyMetadata::MULTI_FACE_DIRECTION_FLAG_UP) !== 0)
+				->setFace(Facing::NORTH, ($drectionFlags & BlockLegacyMetadata::MULTI_FACE_DIRECTION_FLAG_NORTH) !== 0)
+				->setFace(Facing::SOUTH, ($drectionFlags & BlockLegacyMetadata::MULTI_FACE_DIRECTION_FLAG_SOUTH) !== 0)
+				->setFace(Facing::WEST, ($drectionFlags & BlockLegacyMetadata::MULTI_FACE_DIRECTION_FLAG_WEST) !== 0)
+				->setFace(Facing::EAST, ($drectionFlags & BlockLegacyMetadata::MULTI_FACE_DIRECTION_FLAG_EAST) !== 0);
+		});
```
dktapps: A helper method should be introduced for the multiface bits, since it'll be needed again.

src/data/bedrock/block/convert/BlockObjectToStateSerializer.php
dktapps: A helper method should be introduced for this, so it can be reused.

dktapps (commented):
Looks mostly OK, although the placement logic looks a bit suspect.

src/block/GlowLichen.php
```diff
@@ -0,0 +1,175 @@
...
+				$world->useBreakOn($this->position);
+			}else{
+				$world->setBlock($this->position, $this);
+			}
+		}
+	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if($item->getTypeId() === ItemTypeIds::BONE_MEAL){
+			//TODO: spreading
```
dktapps: Do you intend to implement this in the PR?

dktapps: Please implement this. I don't want to merge it while it's not complete.

dktapps:
Taking over this PR due to an overwhelming spam of blockstate errors on the PM5 test server.

The bone meal spreading doesn't look complex: it essentially can spread to any adjacent face within a taxicab distance of 1. This might be a challenge to implement though.

IvanCraft623:
Changed target branch to `minor-next`

src/block/GlowLichen.php
```diff
@@ -0,0 +1,275 @@
...
+		foreach($faces as $face){
+			Facing::validate($face);
+			$uniqueFaces[$face] = $face;
+		}
+		$this->faces = $uniqueFaces;
+		return $this;
+	}
+
+	/** @return $this */
+	public function setFace(int $face, bool $value) : self{
```
ShockedPlot7560, line 72: I find the ``$value`` a bit confusing, perhaps a name like delete or add instead?

dktapps: This is currently consistent with blocks like vines

ShockedPlot7560: When I first looked at it, I wondered what it was for, so I suggested renaming it. But maybe it's necessary to rename the vines too, or we can leave it as it is if it's clear enough.L

dktapps: It is a bit ambiguous, but I'm doubtful that it's worth the extra inconsistency. Likely no one will take the time to change vines either, and this style is used in many places.

ShockedPlot7560: Right

dktapps: For example, we might want to add a trait for these things in the future, and it would suck if we weren't able to because of one named parameter

dktapps:
This seems good to merge now.

IvanCraft623:
> This seems good to merge now.

So the double `hasFace` check is intentional?

dktapps:
No, I forgot to remove it. Thanks.

dktapps (approved):

PR #5405, opened 2022-11-10
added Internet::getIP() return type, fixed typo in doc

Lunarelly:
## Introduction
introduction

### DIFF
--- a/src/utils/Internet.php
+++ b/src/utils/Internet.php
@@ -70,13 +70,13 @@ class Internet{
 	public static $online = true;
 
 	/**
-	 * Gets the External IP using an external service, it is cached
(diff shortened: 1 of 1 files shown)


dktapps:
These changes are unrelated and should not be in the same PR. The return type change also breaks semantic BC and doesn't belong on stable.

SOF3:
What semantic BC does it change? Or just for people who somehow try to extend `Internet` and override a static method?

dktapps:
Yes. It's a stupid case for sure, but it's better to have merge rules clear. Otherwise we start down a slippery slope that leads to a big debate every time someone wants to make a probably-ok-but-not-sure change on stable.

PR #5406, opened 2022-11-11
Fixed Internet::getIP() doc typo

Lunarelly:

### DIFF
--- a/src/utils/Internet.php
+++ b/src/utils/Internet.php
@@ -70,7 +70,7 @@ class Internet{
 	public static $online = true;
 
 	/**
-	 * Gets the External IP using an external service, it is cached
+	 * Gets the External IP using an external service, if it is cached
 	 *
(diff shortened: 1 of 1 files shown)


SOF3:
I don't think the original comment implied "if it is cached". In fact, it is pretty much the opposite &mdash; if the IP is cached, it uses the cache; otherwise, it looks up the IP using an external service.

src/utils/Internet.php
```diff
@@ -70,7 +70,7 @@ class Internet{
 	public static $online = true;
 
 	/**
-	 * Gets the External IP using an external service, it is cached
+	 * Gets the External IP using an external service, if it is cached
```
SOF3:
```suggestion
	 * Lazily gets the External IP using an external service and caches the result
```

SOF3 (changes requested):

SOF3 (approved):

PR #5422, opened 2022-11-21
Candle: Fix support type

ipad54:


dktapps:
please place this next to recalculateCollisionBoxes for consistency

PR #5423, opened 2022-11-22
Implemented Structure Void

dotV3:


ghost:
Phpdoc is generating automatically.

src/block/VanillaBlocks.php
```diff
@@ -550,6 +550,7 @@
  * @method static Wood STRIPPED_OAK_WOOD()
  * @method static Log STRIPPED_SPRUCE_LOG()
  * @method static Wood STRIPPED_SPRUCE_WOOD()
+ * @method static StructureVoid STRUCTURE_VOID()
```
jasonw4331, line 553: Please do not update this doc block manually to ensure automatic method detection/generation functions properly.

src/block/VanillaBlocks.php
```diff
@@ -1114,6 +1115,7 @@ protected static function setup() : void{
 		self::register("stripped_oak_wood", $factory->get(Ids::WOOD, 8));
 		self::register("stripped_spruce_log", $factory->get(Ids::STRIPPED_SPRUCE_LOG, 0));
 		self::register("stripped_spruce_wood", $factory->get(Ids::WOOD, 9));
+		self::register("structure_void", $factory->get(Ids::STRUCUTRE_VOID, 0));
```
jasonw4331: typo'd

dotV3:
Gotta fix it sry

jasonw4331 (changes requested):
Typo + doc should not be done by hand

jasonw4331:
... Why did the PR close?

dotV3:
> Typo + doc should not be done by hand

removed it

dotV3:
> ... Why did the PR close?

I'll re-open it when I'm done

dotV3:
Completed everything

dktapps:
This can't be targeted at PM4. Structure voids weren't added until 1.13, so it would produce worlds with incompatibilities with Bedrock 1.12.

Even if this weren't the case, this change doesn't belong on stable.

dotV3:
> This can't be targeted at PM4. Structure voids weren't added until 1.13, so it would produce worlds with incompatibilities with Bedrock 1.12.
> 
> 
> 
> Even if this weren't the case, this change doesn't belong on stable.

It was just on https://github.com/pmmp/PocketMine-MP/projects/14#card-84464854 so I added it

src/block/VanillaBlocks.php
dktapps: Technically speaking, while the docblock should indeed be generated, it's not of great importance, since the CI will verify that the docblock matches what's in the registry anyway. https://github.com/pmmp/PocketMine-MP/actions/runs/3521719774/jobs/5903887073

dktapps:
As expected, this doesn't actually work (placed structure voids turn into update blocks, because PM4 can't support any blocks newer than 1.12).

With this and aforementioned comments, I'm going to close this.

PR #5424, opened 2022-11-22
Implement Cave Vines & Glow Berries

ipad54:
## Introduction
Implement Cave Vines & Glow Berries

## Tests
https://www.youtube.com/watch?v=tY_mTphlF7k


src/block/CaveVines.php
```diff
@@ -0,0 +1,200 @@
...
+	public function describeState(RuntimeDataWriter|RuntimeDataReader $w) : void{
+		$w->boundedInt(5, 0, self::MAX_AGE, $this->age);
+		$w->bool($this->hasBerries);
+		$w->bool($this->tip);
+	}
+
+	public function readStateFromWorld() : Block{
+		parent::readStateFromWorld();
+
+		$this->tip = !$this->getSide(Facing::DOWN)->isSameType($this);
```
dktapps: This would be more appropriate for onNearbyBlockChange. Usually readStateFromWorld shouldn't be used unless working around some vanilla limitation.

src/block/CaveVines.php
```diff
@@ -0,0 +1,200 @@
...
+		}
+		return false;
+	}
+
+	public function onRandomTick() : void{
+		if($this->tip && $this->age < self::MAX_AGE && mt_rand(1, 10) === 1){
+			$growthPos = $this->position->getSide(Facing::DOWN);
+			$world = $growthPos->getWorld();
+			if($world->isInWorld($growthPos->getFloorX(), $growthPos->getFloorY(), $growthPos->getFloorZ()) && $world->getBlock($growthPos)->getTypeId() === BlockTypeIds::AIR){
+				$ev = new BlockGrowEvent($this, VanillaBlocks::CAVE_VINES()
```
dktapps: This position doesn't match the position used below.

src/data/bedrock/block/convert/BlockObjectToStateSerializer.php
```diff
@@ -712,6 +713,10 @@ private function registerSerializers() : void{
 			return Writer::create(Ids::CARVED_PUMPKIN)
 				->writeLegacyHorizontalFacing($block->getFacing());
 		});
+		$this->map(Blocks::CAVE_VINES(), function(CaveVines $block) : Writer{
+			return Writer::create($block->hasBerries() ? ($block->isTip() ? Ids::CAVE_VINES_HEAD_WITH_BERRIES : Ids::CAVE_VINES_BODY_WITH_BERRIES) : Ids::CAVE_VINES)
```
dktapps: Please split the ID selection across multiple lines

src/item/StringToItemParser.php
```diff
@@ -204,6 +204,9 @@ private static function registerBlocks(self $result) : void{
 		$result->registerBlock("carrots", fn() => Blocks::CARROTS());
 		$result->registerBlock("carved_pumpkin", fn() => Blocks::CARVED_PUMPKIN());
 		$result->registerBlock("cauldron", fn() => Blocks::CAULDRON());
+		$result->registerBlock("cave_vines", fn() => Blocks::CAVE_VINES());
+		$result->registerBlock("cave_vines_body_with_berries", fn() => Blocks::CAVE_VINES()->setBerries(true));
+		$result->registerBlock("cave_vines_head_with_berries", fn() => Blocks::CAVE_VINES()->setBerries(true)->setTip(true));
```
dktapps: These should not be here if they can't be obtained via /give

dktapps (changes requested):
This implementation allows a state of cave vines which is a head but doesn't have berries. This is not representable in vanilla, so the implementation needs to be changed.

src/block/CaveVines.php
ipad54: I don't use onNearbyBlockChange() here because it's not call when the block itself is placed or changed

dktapps: That's not correct... and even if you need placement logic, you have place() for that

ipad54: place() is not called when a block has been placed using world->setBlock()

dktapps: onNearbyBlockChange is, unless you disabled updating.

dktapps: onNearbyBlockChange is called in both of the cases you mentioned.

dktapps:
Instead of two classes, you could perhaps create an enum like so:

BODY: hasBerries() => false,
BODY_WITH_BERRIES: hasBerries() => true
HEAD_WITH_BERRIES: hasBerries() => true

ipad54:
~~I thought about it, but how i should implement the $tip state in one class? It is only needed for vines with berries.~~
Never mind..

src/block/CaveVines.php
dktapps: The first argument to `BlockGrowEvent` should be the block which is being replaced. This event incorrectly suggests that $this is being replaced.

src/block/CaveVines.php
```diff
@@ -0,0 +1,203 @@
...
+		}
+		$this->age = $age;
+		return $this;
+	}
+
+	public function isAffectedBySilkTouch() : bool{
+		return true;
+	}
+
+	public function ticksRandomly() : bool{
```
dktapps: Please place this next to onRandomTick()

src/block/CaveVines.php
```diff
@@ -0,0 +1,203 @@
...
+	/** @return $this */
+	public function setAge(int $age) : self{
+		if($age < 0 || $age > self::MAX_AGE){
+			throw new \InvalidArgumentException("Age must be in range 0-" . self::MAX_AGE);
+		}
+		$this->age = $age;
+		return $this;
+	}
+
+	public function isAffectedBySilkTouch() : bool{
```
dktapps: Please place this near getSilkTouchDrops()

src/data/bedrock/block/convert/BlockStateToObjectDeserializer.php
```diff
@@ -549,6 +551,20 @@ private function registerDeserializers() : void{
 			return Blocks::CARVED_PUMPKIN()
 				->setFacing($in->readLegacyHorizontalFacing());
 		});
+		$this->map(Ids::CAVE_VINES, function(Reader $in) : CaveVines{
+			return Blocks::CAVE_VINES()
+				->setAge($in->readBoundedInt(StateNames::GROWING_PLANT_AGE, 0, 25));
```
dktapps, line 619: The type should be set here. Don't make assumptions based on the default type.

src/block/CaveVines.php
```diff
@@ -0,0 +1,203 @@
...
+	public function canClimb() : bool{
+		return true;
+	}
+
+	public function getLightLevel() : int{
+		return $this->type->hasBerries() ? 14 : 0;
+	}
+
+	private function canBeSupportedBy(Block $block) : bool{
+		return $block->getSupportType(Facing::DOWN)->hasCenterSupport() || $block->isSameType($this);
```
dktapps: I would expect that HEAD_WITH_BERRIES cannot provide support to other vine blocks. That should probably be checked here.

dktapps (commented):
I've yet to test this, but it's looking good so far.

src/block/CaveVines.php
ipad54: I re-tested this behavior in vanilla and it's more confusing than i intend. In vanilla, HEAD_WITH_BERRIES can provide support to other vines, and vanilla recalculates the type when a block has been ticked randomly, not when some block has been changed nearby

src/block/CaveVines.php
```diff
@@ -109,7 +109,7 @@ public function onNearbyBlockChange() : void{
 	}
 
 	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
-		if(!$this->canBeSupportedBy($blockReplace->getSide(Facing::UP))){
+		if(!$this->canBeSupportedBy($blockReplace->getSide(Facing::UP)) || $face !== Facing::DOWN){
```
dktapps: I don't think this should be necessary? usually blocks can figure out where they are supposed to go regardless of which face was clicked

dktapps (commented):
I'm a bit unsure what to do about the head/body with/without berries issue. There clearly exists a visual state of head without berries, and yet it's not representable as a blockstate. It's unclear to me why the Bedrock guys decided to implement it this way instead of just making 4 states like in Java. If anyone knows why this is, I'd love to hear about it.

src/block/CaveVines.php
ipad54: Usually yes, but in Bedrock, cave vines can be placed only if facing is down.

dktapps: I think that's an unnecessary restriction. There's no real reason to enforce it in PM as far as I can see.

ipad54: I mean that without this check the vines can be placed when client thinks it can't.  I send a video:

Without face check: https://www.youtube.com/watch?v=FsZUy4TqthE
With face check: https://youtu.be/7Fc_yt_Trr0

Also, this behavior is only unique for cave vines - neither weeping vines nor hanging roots have this placement behavior.

dktapps: Yes, I understand the problem. I'm just saying I don't think it's necessary to restrict it in PM.

dktapps:
After extensive thought on the subject, I haven't been able to come up with a reason for the tri-state cave vines. Sorry to mess you around on this - the original solution was probably the best. But it will require some workarounds for the unit tests.

Issue #4761, opened 2022-01-23
Add a `Furnace->getType() : FurnaceType` API method


dktapps:
## Description
Right now it's difficult to tell the difference between different types of furnaces without using the tile or legacy ID.

PR #5425, opened 2022-11-22
Add Furnace->getType() : FurnaceType method

zSALLAZAR:
### Relevant issues

#4761 

### API changes
Added `Furnace->getType() : FurnaceType`


dktapps:
Had a quick glance over it, looks OK. Will merge when I reach home as long as there are no issues.

PR #5427, opened 2022-11-22
Add `PlayerBlockPickEvent->setResultItem(Item $resultItem) : void` method

zSALLAZAR:
### Relevant issues

#4771 

### API changes
Added `PlayerBlockPickEvent->setResultItem(Item $resultItem) : void`

## Tests

```php
		$this->getServer()->getPluginManager()->registerEvent(PlayerBlockPickEvent::class, function(PlayerBlockPickEvent $event): void{
			if ($event->getBlock()->getTypeId() === VanillaBlocks::GRASS()->getTypeId()) {
				$event->setResultItem(VanillaItems::DIAMOND());
			}
	}
}
```

https://user-images.githubusercontent.com/59490940/203421216-ae4a20a5-fda2-4bef-9714-0116df5f8b96.mp4


src/player/Player.php
```diff
@@ -1585,14 +1585,13 @@ public function pickBlock(Vector3 $pos, bool $addTileNBT) : bool{
 			return true;
 		}
 
-		$item = $block->getPickedItem($addTileNBT);
-
-		$ev = new PlayerBlockPickEvent($this, $block, $item);
+		$ev = new PlayerBlockPickEvent($this, $block, $block->getPickedItem($addTileNBT));
+		$ev->call();
```
dktapps, line 1699: Why is this moved? The code below it needs to run before the event is called.

zSALLAZAR: `$existingSlot = $this->inventory->first($item);` must be executed after `$item = $ev->getResultItem();` to return the first slot number containing the setted picked item

dktapps: The code below cancels the event before calling it if the item can't be added to the inventory. Your change breaks that behaviour.

src/player/Player.php
```diff
@@ -1585,15 +1585,23 @@ public function pickBlock(Vector3 $pos, bool $addTileNBT) : bool{
...
 			$ev->cancel();
 		}
 		$ev->call();
 
+		$resultItem = $ev->getResultItem();
+		if(!$resultItem->equalsExact($pickedItem)){
+			$existingSlot = $this->inventory->first($resultItem);
+			if($existingSlot === -1 && ($this->hasFiniteResources() || $this->isSpectator())){
+				$ev->cancel();
+			}
```
dktapps: Please avoid duplicating this logic (make a private method or closure).

src/event/player/PlayerBlockPickEvent.php
```diff
@@ -50,4 +50,8 @@ public function getBlock() : Block{
 	public function getResultItem() : Item{
 		return $this->resultItem;
 	}
+
+	public function setResultItem(Item $resultItem) : void{
+		$this->resultItem = $resultItem;
```
dktapps: This should be cloned to prevent reference poisoning.

dktapps (changes requested):

src/player/Player.php
```diff
@@ -1685,15 +1685,24 @@ public function pickBlock(Vector3 $pos, bool $addTileNBT) : bool{
...
-		$item = $block->getPickedItem($addTileNBT);
+		$pickedItem = $block->getPickedItem($addTileNBT);
 
-		$ev = new PlayerBlockPickEvent($this, $block, $item);
-		$existingSlot = $this->inventory->first($item);
-		if($existingSlot === -1 && $this->hasFiniteResources()){
-			$ev->cancel();
-		}
+		$ev = new PlayerBlockPickEvent($this, $block, $pickedItem);
+		$checkExistingSlot = function(Item $item) use ($ev) : int{
```
ShockedPlot7560: I don't think that's the best name. Maybe something like ``findExistingSlot`` would be more appropriate

src/player/Player.php
```diff
@@ -1685,15 +1685,24 @@ public function pickBlock(Vector3 $pos, bool $addTileNBT) : bool{
...
+			if($existingSlot === -1 && $this->hasFiniteResources()){
+				$ev->cancel();
+			}
+			return $existingSlot;
+		};
+		$existingSlot = $checkExistingSlot($pickedItem);
 		$ev->call();
 
+		$resultItem = $ev->getResultItem();
+		if(!$resultItem->equalsExact($pickedItem)){
```
ShockedPlot7560, line 1702: Why rechecking again when you can do it all at once after the call? The item will be the same if there has been no change, or it will be modified directly.

dktapps (commented):
I find this a bit of a headache since the action may not actually take place even though `PlayerBlockPickEvent` suggests that it will.

In survival mode, the item chosen is supposed to be selected from the player's existing resources. When plugins are able to change the outcome, this becomes doubly confusing since the resulting behaviour is not clear:
- does the player receive a new item, or is an existing slot selected to move? How do plugins know what's going on here?
- If an existing slot is selected, but there are no corresponding slots of the item, the event will quietly fail despite not being cancelled. `isCancelled` is no longer trustworthy in such cases.

zSALLAZAR:
If a player is in survival mode, the player does not receive a new item.
For example, if the result item for a grass block got changed to a diamond block and a player trys to pick a grass block with no diamond blocks in the inventory nothing happens. It is the same as if the player would try to pick a diamond block with no diamond blocks in the inventory. And if the player would have diamond blocks in the inventory the held item would change to the diamond blocks.

dktapps:
Yes, I understand that. The problem is that plugin developers won't know what to expect, since the event may fail after the fact as a result of plugins changing the outcome. This makes the behaviour inconsistent, as the event may or may not be cancelled if the action can't take place.

I don't see an obvious way to solve this other than making the event always uncancelled before the event even if the initial outcome is impossible, although that would change the implications of the event.

dktapps:
Closing this as I think the concept is flawed. There needs to be a rethink and possibly some way of exposing generic picking logic on the API without binding it to a particular source. (or a rethink of PlayerBlockPickEvent's behaviour, which raises some BC issues.)

PR #5428, opened 2022-11-23
Add the event: WorldParticleEvent

aderoian:
## Introduction
Adds an API event to listen for when a Particle is broadcasted to a world using World->addParticle(). This event can be cancelled, or the particle, and/or target players can be changed.

### API changes
Addition of the event: ``WorldParticleEvent``

### Behavioural changes
n/a

## Backwards compatibility
n/a

## Follow-up
n/a

## Tests
Test plugin to test cancelling, changing the particle, and changing the target players.
```php
final class Main extends PluginBase implements Listener{

	}

	/**
	 * @priority MONITOR
	 * @ignoreCancelled true
	 */
	public function onWorldParticle(WorldParticleEvent $event) : void{
		// $event->cancel(); // No particles are shown (works)

		// $event->setParticle(new AngryVillagerParticle()); // All particles are replaced by angry villager particle (works)

		// $event->setRecipients([]); // Particle is sent to no one (works)
	}
}
```


alvin0319:
It would be good to add `getPos()` or whatever method to get the particle's position, too. (maybe add a setter too?)

dktapps:
Knowing the position is essential, yes. Though I doubt that changing the position makes sense.

dktapps (approved):

PR #5429, opened 2022-11-23
Cleans the string containing '&'

iBryanidk:


SOF3:
1. This breaks backwards compatibility directly.
2. `&` alone is not going to color the string without calling `TextFormat::colorize()`. If you want to remove all colors, call `TextFormat::colorize()` before calling `TextFormat::clean()` (or just don't colorize it).

Issue #5431, opened 2022-11-24
Add `onInteractEntity` in Item


iBryanidk:
## Description
It should be added to get player interaction with an entity (with the item in hand, like in Nukkit), the method could be called here: https://github.com/pmmp/PocketMine-MP/blob/stable/src/player/Player.php#L1807

IDK which could be better to call, `onClickEntity` or `onInteractEntity`, the goal is to add that method on `Item`

## Justification
This can be useful for NameTag where `right-click` is used to assign the name to the entity. It may have some other utility but that is separate.

I made an example to show what I mean: https://github.com/iBryanidk/PocketMine-MP/commit/de1a490fd3d3f5c7e6db635559bbee854dde66ab (I think the 'bool' return type is unnecessary, as well as in 'onDestroyBlock', since it won't be used)

## Alternative methods
n/a

dktapps:
Yes, I do think this is necessary. In fact it's one of the main reasons I wasn't happy with #5209.

iBryanidk:
> Yes, I do think this is necessary. In fact it's one of the main reasons I wasn't happy with #5209.

I can do a PR about this to see how it goes

dktapps:
This should stay open until the change has been merged

PR #5432, opened 2022-11-24
Implemented method `onInteractEntity`

iBryanidk:
## Introduction
This is related to the issue: https://github.com/pmmp/PocketMine-MP/issues/5431

It should be added as I mentioned, because currently there is no method for this, there is `onAttackEntity` but for this it is necessary to attack the entity, but `onInteractEntity` is called when the player does the interaction with `right-click`. In theory `onInteractEntity` should be called if the event is not cancelled, otherwise it would be pointless.

I think the `bool` return value is unnecessary, but I don't want to break the aesthetics of everything else.

### DIFF
--- a/src/item/Item.php
+++ b/src/item/Item.php
@@ -555,7 +555,7 @@ class Item implements \JsonSerializable{
 	}
 
 	/**
-	 * Called when a player uses the item to interact with entity, for example by placing a name tag.
+	 * Called when a player uses the item to interact with entity, for example by using a name tag.
 	 */
 	public function onInteractEntity(Player $player, Entity $entity, Vector3 $directionVector) : bool{
 		return false;
--- a/src/item/Item.php
+++ b/src/item/Item.php
@@ -555,6 +555,13 @@ class Item implements \JsonSerializable{
 	}
 
 	/**
+	 * Called when a player uses the item to interact with entity, for example by placing a name tag.
+	 */
+	public function onInteractEntity(Player $player, Entity $entity, Vector3 $directionVector) : bool{
+		return false;
+	}
+
+	/**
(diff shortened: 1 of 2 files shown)


src/item/Item.php
```diff
@@ -554,6 +554,13 @@ public function onAttackEntity(Entity $victim) : bool{
 		return false;
 	}
 
+	/**
+	 * Called when a player uses the item to interact with entity, for example by placing a name tag.
```
dktapps:
```suggestion
	 * Called when a player uses the item to interact with entity, for example by using a name tag.
```

dktapps:
> I think the bool return value is unnecessary, but I don't want to break the aesthetics of everything else.

To be honest you're probably right about this in general. In the past, these return values were used to decide whether the item should be set back into the inventory using setItem() (if modified), to avoid spamming the client with useless inventory updates. However, in PM4, the system automatically detects whether syncing the item is necessary anyway, so most of these return values are now redundant.

dktapps:
The build needs to pass.

iBryanidk:
> The build needs to pass.

Yes, for some reason the `interactEntity` method was being called instead of `onInteractEntity`, but that's it.

src/player/Player.php
```diff
@@ -1814,7 +1814,10 @@ public function interactEntity(Entity $entity, Vector3 $clickPos) : bool{
 
 		$ev->call();
 
+		$item = $this->inventory->getItemInHand();
 		if(!$ev->isCancelled()){
+			$item->onInteractEntity($this, $entity, $clickPos);
```
dktapps: This isn't enough by itself. You need to check if the item was changed, and set it back into the inventory if it was.

For example, using shears will change the durability, and using a nametag will reduce the count by 1.

Issue #5378, opened 2022-11-01
Breaking a durable item with more than one count gives it infinite durability


JavierLeon9966:
### Issue description
Getting a durable item like a diamond sword with a count more than 1 in one slot and breaking gives it infinite durability instead of resetting the durability of the item with.

### Steps to reproduce the issue
1. Give yourself a durable item with more than one count in one slot
2. Break it
3.  You will notice the item will have used all its durability but can't be broken

### OS and versions

* PocketMine-MP: 4.10.0 (git hash: 0d5287bf0bd7138113e1c154d3c3a3b83a407d82)
* PHP: 8.0.24
* Using JIT: no 
* Server OS: win
* Game version: Win10

dktapps:
I think the current implementation was supposed to mimic the vanilla behaviour at the time, but it's been done incorrectly. In any case, the vanilla implementation -- and that of PM -- are not really defined behaviour, since this should never occur without plugin intervention anyway.

PR #5436, opened 2022-11-26
Fix Durable not being broken

SlimyBoy2:
## Introduction
The `isBroken()` function of the `Durable` class is only looking at the item damage. Since `4.10.1`, `isBroken()` isn't working properly because the damage property is set to 0 in `onBroken()`. 

The function now looks at the damage but also the count of the item using `isNull()`.

### Relevant issues
This is related to [#5378](https://github.com/pmmp/PocketMine-MP/issues/5378) and [this commit](https://github.com/pmmp/PocketMine-MP/commit/fe982c697bec3265c4222b73fe5ed17c9ef3b1ce)

### API changes
- None

### Behavioural changes
- Updated the return condition of the `isBroken()` function

## Backwards compatibility
- None

## Follow-up
- None

## Tests
I used this code to test the change. Results are below
```php
$item = VanillaItems::DIAMOND_SWORD();
$item->setDamage(1561);

var_dump($item->getDamage());

$item->applyDamage(1);

var_dump($item->getDamage(), $item->isBroken());
```

- 4.10.0
```php
int(1561)
int(1562)
bool(true) <-
```

- 4.11.0
```php
int(1561)
int(0)
bool(false) <-
```

- 4.11.1 (PR)
```php
int(1561)
int(0)
bool(true) <-
```

### DIFF
--- a/src/item/Durable.php
+++ b/src/item/Durable.php
@@ -121,7 +121,7 @@ abstract class Durable extends Item{
 	 * Returns whether the item is broken.
 	 */
 	public function isBroken() : bool{
-		return $this->damage >= $this->getMaxDurability();
+		return $this->damage >= $this->getMaxDurability() || $this->isNull();
 	}
 
 	protected function deserializeCompoundTag(CompoundTag $tag) : void{


dktapps:
To be honest I think this method was misconceived. The places where this function was used should really be checking if the stack size changed as a result of damaging the item. Might be worth considering deprecating/removing it.

For now, I think this fix is acceptable.

PR #5444, opened 2022-12-05
Only reduce durability when armor reduced damage.

rasu3n:
## Introduction

Current behavior is different from vanilla.
(https://minecraft.fandom.com/wiki/Durability#Armor_durability)

### DIFF
--- a/src/entity/Living.php
+++ b/src/entity/Living.php
@@ -455,7 +455,9 @@ abstract class Living extends Entity{
 	 */
 	protected function applyPostDamageEffects(EntityDamageEvent $source) : void{
 		$this->setAbsorption(max(0, $this->getAbsorption() + $source->getModifier(EntityDamageEvent::MODIFIER_ABSORPTION)));
-		$this->damageArmor($source->getBaseDamage());
+		if($source->canBeReducedByArmor()){
+			$this->damageArmor($source->getBaseDamage());
+		}
 
 		if($source instanceof EntityDamageByEntityEvent && ($attacker = $source->getDamager()) !== null){
 			$damage = 0;


ghost:
Still, it's worth tweaking this decision a bit. Since poison, and some effects are also absorbed by armor, it is worth considering this too.

dktapps:
> Still, it's worth tweaking this decision a bit. Since poison, and some effects are also absorbed by armor, it is worth considering this too.

Pretty sure that's not the case...

dktapps (approved):
Looks fine, but I think some investigation into the original code might be warranted (I doubt I missed this when I wrote the code to begin with).

ghost:
> > Still, it's worth tweaking this decision a bit. Since poison, and some effects are also absorbed by armor, it is worth considering this too.
> 
> Pretty sure that's not the case...

Since it has not been tested, or rather I cannot test it, the author of the PR should do this. I myself am not very sure that after such changes, armor still takes damage from effects

rasu3n:
> 

As dylan said their damage is not reduced by armor.

https://github.com/pmmp/PocketMine-MP/blob/0a9b52618d4ab570be914dc8083445714ca5df2b/src/event/entity/EntityDamageEvent.php#L160-L162
https://github.com/pmmp/PocketMine-MP/blob/0a9b52618d4ab570be914dc8083445714ca5df2b/src/entity/effect/PoisonEffect.php#L49
https://github.com/pmmp/PocketMine-MP/blob/0a9b52618d4ab570be914dc8083445714ca5df2b/src/entity/effect/WitherEffect.php#L40

ghost:
Good

rasu3n:
> Looks fine, but I think some investigation into the original code might be warranted (I doubt I missed this when I wrote the code to begin with).

It looks like it's been the same behavior since the armor durability was implemented.
https://github.com/pmmp/PocketMine-MP/commit/e06b78b0ee6b23c6cb91b5f4f73e6557a086b239

PR #5448, opened 2022-12-09
Allow players to place ItemFrames up and down

supercrafter333:
## Introduction
Players can't place ItemFrames at the top or the ground because the UP and DOWN facing is blocked by pmmp. This PR allows players to place ItemFrames at the top and the ground

### API changes
None

### Behavioural changes
Players can place ItemFrames with the facing UP and DOWN.

## Backwards compatibility
If a server owner downgrades his server there will be problems with the facing of item frames with the facing UP and DOWN.


dktapps:
This is not sufficient by itself. Up/down item frames are not supported on PM4 because it still uses legacy block metadata.

PR #5452, opened 2022-12-13
Implement Ice Bomb

zSALLAZAR:
## Introduction
This is a Education Edition feature.

The ice bomb is an item that is used to freeze water into ice.

## Tests

https://user-images.githubusercontent.com/59490940/207361701-bd0be042-4580-497d-915c-c1ca03e8fecb.mp4

### DIFF
--- a/src/entity/EntityFactory.php
+++ b/src/entity/EntityFactory.php
@@ -43,6 +43,7 @@ use pocketmine\entity\projectile\Arrow;
 use pocketmine\entity\projectile\Egg;
 use pocketmine\entity\projectile\EnderPearl;
 use pocketmine\entity\projectile\ExperienceBottle;
+use pocketmine\entity\projectile\IceBomb;
 use pocketmine\entity\projectile\Snowball;
 use pocketmine\entity\projectile\SplashPotion;
 use pocketmine\item\Item;
@@ -119,6 +120,10 @@ final class EntityFactory{
 		$this->register(FallingBlock::class, function(World $world, CompoundTag $nbt) : FallingBlock{
 			return new FallingBlock(Helper::parseLocation($nbt, $world), FallingBlock::parseBlockNBT(RuntimeBlockStateRegistry::getInstance(), $nbt), $nbt);
 		}, ['FallingSand', 'minecraft:falling_block']);
+
+		$this->register(IceBomb::class, function(World $world, CompoundTag $nbt) : IceBomb{
+			return new IceBomb(Helper::parseLocation($nbt, $world), null, $nbt);
+		}, ['ThrownIceBomb', 'minecraft:ice_bomb']);
 
 		$this->register(ItemEntity::class, function(World $world, CompoundTag $nbt) : ItemEntity{
 			$itemTag = $nbt->getCompoundTag(ItemEntity::TAG_ITEM);
--- a/src/entity/EntityFactory.php
+++ b/src/entity/EntityFactory.php
@@ -123,7 +123,7 @@ final class EntityFactory{
 
 		$this->register(IceBomb::class, function(World $world, CompoundTag $nbt) : IceBomb{
 			return new IceBomb(Helper::parseLocation($nbt, $world), null, $nbt);
-		}, ['ThrownIceBomb', 'minecraft:ice_bomb']);
+		}, ['minecraft:ice_bomb']);
 
 		$this->register(ItemEntity::class, function(World $world, CompoundTag $nbt) : ItemEntity{
 			$itemTag = $nbt->getCompoundTag(ItemEntity::TAG_ITEM);
--- a/src/entity/projectile/IceBomb.php
+++ b/src/entity/projectile/IceBomb.php
@@ -37,7 +37,7 @@ use pocketmine\world\sound\BlockBreakSound;
 use const PHP_INT_MAX;
 
 class IceBomb extends Throwable{
-	public static function getNetworkTypeId() : string{ return EntityIds::ICE_BOMB; }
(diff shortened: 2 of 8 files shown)


src/entity/EntityFactory.php
```diff
@@ -112,6 +113,10 @@ public function __construct(){
 			return new FallingBlock(Helper::parseLocation($nbt, $world), FallingBlock::parseBlockNBT(BlockFactory::getInstance(), $nbt), $nbt);
 		}, ['FallingSand', 'minecraft:falling_block'], LegacyIds::FALLING_BLOCK);
 
+		$this->register(IceBomb::class, function(World $world, CompoundTag $nbt) : IceBomb{
+			return new IceBomb(Helper::parseLocation($nbt, $world), null, $nbt);
+		}, ['ThrownIceBomb', 'minecraft:ice_bomb'], LegacyIds::ICE_BOMB);
```
dktapps: Don't assume these. They need to match the vanilla IDs.

zSALLAZAR: What else are the correct vanilla ids?

dktapps: I don't know. That's for you to research.

IvanCraft623: Vanilla NBT tree:
image
`identifier`: `minecraft:ice_bomb`

dktapps: There's no evidence for any such thing as `ThrownIceBomb`. Those style of ID used to be used in Java, but Java doesn't have ice bombs.

src/entity/projectile/IceBomb.php
```diff
@@ -0,0 +1,90 @@
...
+
+class IceBomb extends Throwable{
+	public static function getNetworkTypeId() : string{ return EntityIds::ICE_BOMB; }
+
+	public function getResultDamage() : int{
+		return -1;
+	}
+
+	protected function calculateInterceptWithBlock(Block $block, Vector3 $start, Vector3 $end) : ?RayTraceResult{
+		if($block instanceof Water){
```
ipad54: I think you can use getTypeId() for that

zSALLAZAR: Why should I?

dktapps: `Water` class could be reused for other stuff which isn't supposed to behave the same way, but in general, `instanceof` should be avoided. It's not a safe way to determine a block's type.

src/entity/projectile/IceBomb.php
```diff
@@ -44,7 +43,7 @@ public function getResultDamage() : int{
 	}
 
 	protected function calculateInterceptWithBlock(Block $block, Vector3 $start, Vector3 $end) : ?RayTraceResult{
-		if($block instanceof Water){
+		if($block->getTypeId() === VanillaBlocks::WATER()->getTypeId()){
```
dktapps:
```suggestion
		if($block->getTypeId() === BlockTypeIds::WATER){
```

src/entity/projectile/IceBomb.php
```diff
@@ -80,7 +79,7 @@ protected function onHitBlock(Block $blockHit, RayTraceResult $hitResult) : void
 		for($x = $posX - 1; $x <= $posX + 1; $x++){
 			for($y = $posY - 1; $y <= $posY + 1; $y++){
 				for($z = $posZ - 1; $z <= $posZ + 1; $z++){
-					if($world->getBlockAt($x, $y, $z) instanceof Water){
+					if($world->getBlockAt($x, $y, $z)->getTypeId() === VanillaBlocks::WATER()->getTypeId()){
```
dktapps:
```suggestion
					if($world->getBlockAt($x, $y, $z)->getTypeId() === BlockTypeIds::WATER){
```

dktapps:
There are some merge errors that need fixing

src/item/StringToItemParser.php
```diff
@@ -1299,6 +1299,7 @@ private static function registerItems(self $result) : void{
 		$result->register("heart_of_the_sea", fn() => Items::HEART_OF_THE_SEA());
 		$result->register("honey_bottle", fn() => Items::HONEY_BOTTLE());
 		$result->register("honeycomb", fn() => Items::HONEYCOMB());
+		$result->register('ice_bomb', fn() => Items::ICE_BOMB());
```
dktapps: This should use double quotes

src/item/VanillaItems.php
```diff
@@ -447,6 +448,7 @@ protected static function setup() : void{
 		self::register("heart_of_the_sea", new Item(new IID(Ids::HEART_OF_THE_SEA), "Heart of the Sea"));
 		self::register("honey_bottle", new HoneyBottle(new IID(Ids::HONEY_BOTTLE), "Honey Bottle"));
 		self::register("honeycomb", new Item(new IID(Ids::HONEYCOMB), "Honeycomb"));
+		self::register('ice_bomb', new IceBomb(new IID(Ids::ICE_BOMB), 'Ice Bomb'));
```
dktapps: Same again here

jasonw4331 (approved):

ShockedPlot7560 (dismissed):

pandaaaBE:
this pr should target minor-next

a little change would be required, make getNetworkTypeId() static

zSALLAZAR:
Changing the target branch to minor-next adds commits from major-next to this PR and Idk how to remove them

dktapps:
I think this is done, I made a few minor alterations and rebased it, but code was mostly OK anyway

PR #5454, opened 2022-12-16
Implemented chain

ipad54:
## Introduction
Implemented chain

## Tests
Video: https://www.youtube.com/watch?v=PmOL3lmGSUA


src/block/Chain.php
```diff
@@ -0,0 +1,48 @@
...
+	public function getSupportType(int $facing) : SupportType{
+		return $this->axis === Axis::Y ? SupportType::CENTER() : SupportType::NONE();
+	}
+
+	public function asItem() : Item{
+		return VanillaItems::CHAIN();
+	}
+
+	protected function recalculateCollisionBoxes() : array{
+		$inset = 13 / 32;
```
dktapps: where did you get this number from?

src/block/Chain.php
```diff
@@ -0,0 +1,48 @@
...
+
+final class Chain extends Transparent{
+	use PillarRotationTrait;
+
+	public function getSupportType(int $facing) : SupportType{
+		return $this->axis === Axis::Y ? SupportType::CENTER() : SupportType::NONE();
+	}
+
+	public function asItem() : Item{
+		return VanillaItems::CHAIN();
```
dktapps: This should not be necessary

src/block/Chain.php
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	protected function recalculateCollisionBoxes() : array{
+		$inset = 13 / 32;
+		return [AxisAlignedBB::one()->contract($inset, $inset, $inset)->stretch($this->axis, $inset)];
+	}
+}
```
dktapps, line 48: Where is the logic for deciding the placement orientation?

src/item/Chain.php
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\item;
+
+use pocketmine\block\Block;
+use pocketmine\block\VanillaBlocks;
+
+class Chain extends Item{
```
dktapps: This class should not be necessary

src/item/ItemTypeIds.php
```diff
@@ -299,8 +299,9 @@ private function __construct(){
 	public const FIRE_CHARGE = 20260;
 	public const SUSPICIOUS_STEW = 20261;
 	public const TURTLE_HELMET = 20262;
+	public const CHAIN = 20263;
```
dktapps: This should not be necessary

dktapps (changes requested):
I don't know why you've added a separate item for the block. It's not necessary.

src/block/Chain.php
```diff
@@ -0,0 +1,48 @@
...
+		return $this->axis === Axis::Y ? SupportType::CENTER() : SupportType::NONE();
+	}
+
+	public function asItem() : Item{
+		return VanillaItems::CHAIN();
+	}
+
+	protected function recalculateCollisionBoxes() : array{
+		$inset = 13 / 32;
+		return [AxisAlignedBB::one()->contract($inset, $inset, $inset)->stretch($this->axis, $inset)];
```
dktapps: This logic is a bit confusing. I'd prefer if it was written something more like end rods.

ipad54:
Without adding new item, chain has a different texture and missing from creative inventory

Screenshot_20221218_220651_com android chrome

src/block/Chain.php
ipad54: I used dragonfly as source. 

https://github.com/df-mc/dragonfly/blob/master/server/block/model/chain.go

dktapps:
That doesn't mean that there needs to be a separate item in the PM core. It just means that the item has to be handled differently in ItemSerializer/Deserializer. See beds, doors, etc.

Crasher508:
register chain in ItemSerializerDeserializerRegistrar as "ItemBlock" should be enough for me

add in register1to1BlockMappings() function:
"$this->map1to1Block(Ids::CHAIN, VanillaBlocks::CHAIN());"

src/block/Chain.php
ipad54: The placement logic is included in PillarRotationTrait

src/block/Chain.php
```diff
@@ -37,12 +35,14 @@ public function getSupportType(int $facing) : SupportType{
...
-
 	protected function recalculateCollisionBoxes() : array{
 		$inset = 13 / 32;
-		return [AxisAlignedBB::one()->contract($inset, $inset, $inset)->stretch($this->axis, $inset)];
+
+		$bb = AxisAlignedBB::one();
+		foreach([Axis::Y, Axis::Z, Axis::X] as $axis){
+			$bb->squash($axis, $inset);
+		}
+		$bb->stretch($this->axis, $inset);
```
dktapps: Check for the axis in the loop and skip it, instead of doing this.

src/data/bedrock/item/ItemSerializerDeserializerRegistrar.php
```diff
@@ -497,8 +497,10 @@ function(SuspiciousStew $item, int $meta) : void{
 	 * complex to implement in a generic way.
 	 */
 	private function registerMiscBlockMappings() : void{
+		$this->deserializer?->mapBlock(Ids::CHAIN, fn() => Blocks::CHAIN());
 		$this->deserializer?->mapBlock(Ids::FRAME, fn() => Blocks::ITEM_FRAME()->setGlowing(false));
 		$this->deserializer?->mapBlock(Ids::GLOW_FRAME, fn() => Blocks::ITEM_FRAME()->setGlowing(true));
+		$this->serializer?->mapBlock(Blocks::CHAIN(), fn(Chain $chain) => new Data(Ids::CHAIN));
```
dktapps: Not like this. See `register1to1BlockMappings()`.

src/block/Chain.php
```diff
@@ -0,0 +1,49 @@
...
+use pocketmine\block\utils\PillarRotationTrait;
+use pocketmine\block\utils\SupportType;
+use pocketmine\math\Axis;
+use pocketmine\math\AxisAlignedBB;
+
+final class Chain extends Transparent{
+	use PillarRotationTrait;
+
+	public function getSupportType(int $facing) : SupportType{
+		return $this->axis === Axis::Y ? SupportType::CENTER() : SupportType::NONE();
```
dktapps: This is missing checks for the facing. It should only provide center support if the facing axis is also Y.

dktapps (approved):

PR #5455, opened 2022-12-17
Implement firework rocket & firework star

IvanCraft623:
## Introduction
Implement items: `Firework Rocket`, `Firework Star` and entity: `Firework Rocket`.

## Follow-up
- Merge https://github.com/pmmp/Color/pull/47 for a litle code reduction
- Merge https://github.com/pmmp/BedrockProtocol/pull/118 to be able to use the appropriate constant
- Merge https://github.com/pmmp/Language/pull/145 for fireworks death message

Requires translations:

| Name | Value in eng.ini |
| :--: | :---: |
| `death.attack.fireworks` | `{%0} went off with a bang` |

## Tests
https://youtu.be/kHs_nBmLRhs


src/data/bedrock/FireworkRocketTypeIdMap.php
```diff
@@ -0,0 +1,67 @@
...
+	 * @var FireworkRocketType[]
+	 * @phpstan-var array<int, FireworkRocketType>
+	 */
+	private array $idToEnum;
+
+	/**
+	 * @var int[]
+	 * @phpstan-var array<int, int>
+	 */
+	private array $enumToId;
```
dktapps: These fields are uninitialized

IvanCraft623: done

src/entity/object/FireworkRocket.php
```diff
@@ -0,0 +1,181 @@
...
+		$this->broadcastSound(new FireworkLaunchSound());
+	}
+
+	protected function entityBaseTick(int $tickDiff = 1) : bool{
+		$hasUpdate = parent::entityBaseTick($tickDiff);
+
+		if(!$this->isFlaggedForDespawn()){
+			$this->motion->x *= 1.15;
+			$this->motion->y += 0.04;
+			$this->motion->z *= 1.15;
```
dktapps: This should use addMotion instead of mutating the vector directly

src/entity/object/FireworkRocket.php
```diff
@@ -0,0 +1,181 @@
...
+
+	public function canBeCollidedWith() : bool{
+		return false;
+	}
+
+	protected function syncNetworkData(EntityMetadataCollection $properties) : void{
+		parent::syncNetworkData($properties);
+
+		//TODO: Use the appropriate constant
+		$properties->setCompoundTag(EntityMetadataProperties::MINECART_DISPLAY_BLOCK, new CacheableNbt($this->item->getNamedTag()));
```
dktapps: This has now been updated upstream

src/item/FireworkRocketExplosion.php
```diff
@@ -0,0 +1,216 @@
...
+	protected static function decodeColors(string $colorsBytes) : array{
+		$colors = [];
+
+		$dyeColorIdMap = DyeColorIdMap::getInstance();
+		for($i = 0; $i < strlen($colorsBytes); $i++){
+			$color = $dyeColorIdMap->fromInvertedId(Binary::readByte($colorsBytes[$i]));
+			if($color !== null){
+				$colors[] = $color;
+			}else{
+				//TODO: should throw an exception?
```
dktapps: `SavedDataLoadingException` should be thrown here

src/item/FireworkRocketExplosion.php
```diff
@@ -0,0 +1,216 @@
...
+			->setByteArray(self::TAG_FADE_COLORS, $this->encodeColors($this->fadeColors))
+			->setByte(self::TAG_TWINKLE, $this->twinkle ? 1 : 0)
+			->setByte(self::TAG_TRAIL, $this->trail ? 1 : 0)
+		;
+	}
+
+	/**
+	 * @param DyeColor[] $colors
+	 */
+	protected function encodeColors(array $colors) : string{
```
dktapps: This can be static, and should probably be placed near `decodeColors`

src/item/FireworkRocketExplosion.php
```diff
@@ -0,0 +1,216 @@
...
+
+	/**
+	 * Returns the main color, this defines stuff like meta.
+	 */
+	public function getMainColor() : DyeColor{
+		return $this->colors[array_key_first($this->colors)];
+	}
+
+	/**
+	 * Returns a the mix of all colors.
```
dktapps:
```suggestion
	 * Returns the mix of all colors.
```

src/item/FireworkRocketType.php
```diff
@@ -0,0 +1,71 @@
...
+		__construct as Enum___construct;
+	}
+
+	protected static function setup() : void{
+		self::registerAll(
+			new self("small_ball", fn() => new FireworkExplosionSound()),
+			new self("large_ball", fn() => new FireworkLargeExplosionSound()),
+			new self("star", fn() => new FireworkExplosionSound()),
+			new self("creeper", fn() => new FireworkExplosionSound()),
+			new self("burst", fn() => new FireworkExplosionSound()),
```
dktapps: Not sure what the point of using closures is here?

src/item/FireworkStar.php
```diff
@@ -0,0 +1,107 @@
...
+		}
+		$this->explosion = FireworkRocketExplosion::fromCompoundTag($explosionTag);
+
+		$customColor = Color::fromARGB(Binary::unsignInt($tag->getInt(self::TAG_CUSTOM_COLOR)));
+		$color = $this->explosion->getColorMix();
+		if($customColor->getA() !== $color->getA() ||
+			$customColor->getR() !== $color->getR() ||
+			$customColor->getG() !== $color->getG() ||
+			$customColor->getB() !== $color->getB()
+		){ //check that $customColor is actually custom.
```
dktapps: You can compare the toRGBA() value instead of doing this

src/item/FireworkStar.php
```diff
@@ -0,0 +1,107 @@
...
+	public function clearCustomColor() : self{
+		$this->customColor = null;
+		return $this;
+	}
+
+	protected function deserializeCompoundTag(CompoundTag $tag) : void{
+		parent::deserializeCompoundTag($tag);
+
+		$explosionTag = $tag->getTag(self::TAG_EXPLOSION);
+		if (!$explosionTag instanceof CompoundTag) {
```
dktapps:
```suggestion
		if(!$explosionTag instanceof CompoundTag){
```

src/item/FireworkStar.php
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	/** @return $this */
+	public function setCustomColor(Color $color) : self{
+		$this->customColor = $color;
+		return $this;
+	}
+
+	/** @return $this */
+	public function clearCustomColor() : self{
```
dktapps: I think we can forgo this in favour of passing `null` to `setCustomColor`. No point cluttering the API with junk methods.

src/item/FireworkStar.php
```diff
@@ -0,0 +1,107 @@
...
+	public function setExplosion(FireworkRocketExplosion $explosion) : self{
+		$this->explosion = $explosion;
+		return $this;
+	}
+
+	public function getColor() : Color{
+		return $this->customColor ?? $this->explosion->getColorMix();
+	}
+
+	public function hasCustomColor() : bool{
```
dktapps: `getCustomColor() !== null` can replace this.

src/item/FireworkStar.php
```diff
@@ -0,0 +1,107 @@
...
+		return $this->explosion;
+	}
+
+	/** @return $this */
+	public function setExplosion(FireworkRocketExplosion $explosion) : self{
+		$this->explosion = $explosion;
+		return $this;
+	}
+
+	public function getColor() : Color{
```
dktapps, line 67: what's the difference between this and the custom colour?

src/world/sound/FireworkTwinkleSound.php
```diff
@@ -0,0 +1,35 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\world\sound;
+
+use pocketmine\math\Vector3;
+use pocketmine\network\mcpe\protocol\LevelSoundEventPacket;
+use pocketmine\network\mcpe\protocol\types\LevelSoundEvent;
+
+class FireworkTwinkleSound implements Sound{
```
dktapps: A "twinkle" is a visual effect. It doesn't make me think of a sound.

src/item/FireworkRocketExplosion.php
```diff
@@ -0,0 +1,216 @@
...
+			throw new \InvalidArgumentException("Colors list cannot be empty");
+		}
+		Utils::validateArrayValueType($colors, $this->colorsValidator);
+		$this->colors = array_values($colors);
+
+		return $this;
+	}
+
+	/**
+	 * Returns the main color, this defines stuff like meta.
```
dktapps: What's the difference between "main colour" and the other colours? This documentation is really insufficient.

src/entity/animation/FireworkParticlesAnimation.php
```diff
@@ -0,0 +1,39 @@
...
+
+namespace pocketmine\entity\animation;
+
+use pocketmine\entity\object\FireworkRocket;
+use pocketmine\network\mcpe\protocol\ActorEventPacket;
+use pocketmine\network\mcpe\protocol\types\ActorEvent;
+
+final class FireworkParticlesAnimation implements Animation{
+
+	public function __construct(private FireworkRocket $entity){}
```
dktapps:
```suggestion
	public function __construct(
		private FireworkRocket $entity
	){}
```
This is the typical way we use promoted properties, as it makes it easier to see them.

src/data/bedrock/item/ItemSerializerDeserializerRegistrar.php
```diff
@@ -468,6 +470,14 @@ function(Banner $item, int $meta) : void{
 			},
 			fn(Banner $item) => DyeColorIdMap::getInstance()->toInvertedId($item->getColor())
 		);
+		$this->map1to1ItemWithMeta(
+			Ids::FIREWORK_STAR,
+			Items::FIREWORK_STAR(),
+			function(FireworkStar $item, int $meta) : void{
+				$item->getExplosion()->setColors([DyeColorIdMap::getInstance()->fromInvertedId($meta) ?? throw new ItemTypeDeserializeException("Unknown firework star meta $meta")]);
```
dktapps: what happens if the item has colours in the NBT?

dktapps (changes requested):
The implementation looks mostly OK. I haven't tested it yet.

My main sore point on this PR is the woefully inadequate documentation and ambiguous method names.

src/item/FireworkStar.php
IvanCraft623: Better use the new `Color::equals()` method :P

src/item/FireworkStar.php
IvanCraft623: By default the firework star gets the color resulting from the mixture of the colors that the explosion contains, so setting a `customColor` you can choose the color of the item regardless of the color of the explosions.

So, `getColor()` returns the color that is being displayed

This is similar behavior to` Item::getName()` and `Item::getCustomName()`

dktapps: ok, so what's the purpose of this? why can't we just use the colour mix exclusively?

IvanCraft623: Why not use this feature that is allowing mojang? What if users want to use a custom color but don't want to change the explosion?

dktapps: This isn't answering my question. What is the purpose of this? There's no documentation anywhere.

src/item/FireworkRocket.php
```diff
@@ -40,23 +40,34 @@
...
+	/**
+	 * Returns the flight duration of the firework (equals the amount of gunpowder used in crafting the rocket).
+	 */
 	public function getFlightDuration() : int{
 		return $this->flightDuration;
 	}
 
-	public function setFlightDuration(int $duration) : void{
+	/**
+	 * Sets the flight duration of the firework.
```
dktapps: This is still confusing. It's not really the duration of flight at all. Rather it's a multiplier that is multiplied by a random value to get the actual flight duration.

src/item/FireworkRocketExplosion.php
```diff
@@ -107,56 +107,39 @@ public function __construct(
...
+	 * Returns the flash color of the explosion.
 	 */
-	public function getMainColor() : DyeColor{
+	public function getFlashColor() : DyeColor{
 		return $this->colors[array_key_first($this->colors)];
 	}
 
 	/**
-	 * Returns the mix of all colors.
+	 * Returns the mixure of colors form {@link FireworkRocketExplosion::getColors()})
```
dktapps:
```suggestion
	 * Returns the mixure of colors from {@link FireworkRocketExplosion::getColors()})
```

src/item/FireworkRocket.php
IvanCraft623: I have updated it, is it better for you?

IvanCraft623:
After doing some tests with Fireworks in vanilla it seems to behave like a projectile and in fact it can also be used as one, but PM's Projectile class adds logic that breaks the firework behavior...

dktapps:
> After doing some tests with Fireworks in vanilla it seems to behave like a projectile and in fact it can also be used as one, but PM's Projectile class adds logic that breaks the firework behavior...

Can you clarify?

src/item/FireworkRocket.php
dktapps: I think the functions should still be renamed. Perhaps "flight duration multiplier" would be a better name.

IvanCraft623:
> Can you clarify?

nvm, it seems that it is not exactly a projectile, but it can be launched by crossbows, it is also able of ringing a bell.

IvanCraft623:
Missing features:
- [ ] Fireworks can ring bells
- [ ] Fireworks can be directed in any direction (can be observed when flying with elytras, or using a dispenser).

JavierLeon9966:
> Missing features:
> - [ ] Fireworks can ring bells
> - [ ] Fireworks can be directed in any direction (can be observed when flying with elytras, or using a dispenser).

Also with crossbows, but it can wait until those are implemented first.

ShockedPlot7560:
A PR doesn't have to be all in one, I think the current features are sufficient, the remaining ones are not critical.

src/data/bedrock/item/ItemSerializerDeserializerRegistrar.php
IvanCraft623: the color in the meta is actually irrelevant, it is always overwritten with the colors in the NBT. In fact we could silently ignore the meta, its existence would rather seem to be a mojang hack.

src/item/FireworkStar.php
```diff
@@ -0,0 +1,106 @@
...
+use pocketmine\block\utils\DyeColor;
+use pocketmine\color\Color;
+use pocketmine\data\SavedDataLoadingException;
+use pocketmine\nbt\tag\CompoundTag;
+use pocketmine\utils\Binary;
+
+class FireworkStar extends Item{
+
+	protected const TAG_EXPLOSION = "FireworksItem"; //TAG_Compound
+	protected const TAG_CUSTOM_COLOR = "customColor"; //TAG_Int
```
ShockedPlot7560, line 35: Is it really a `c` at the start and not a `C` ?

IvanCraft623: yep

src/item/FireworkRocketType.php
```diff
@@ -0,0 +1,46 @@
...
+	case CREEPER;
+	case BURST;
+
+	public function getExplosionSound() : Sound{
+		return match($this){
+			self::SMALL_BALL => new FireworkExplosionSound(),
+			self::LARGE_BALL => new FireworkLargeExplosionSound(),
+			self::STAR => new FireworkExplosionSound(),
+			self::CREEPER => new FireworkExplosionSound(),
+			self::BURST => new FireworkExplosionSound(),
```
ipad54:
```suggestion
		return match($this){
			self::SMALL_BALL,
			self::STAR,
			self::CREEPER,
			self::BURST => new FireworkExplosionSound(),
			self::LARGE_BALL => new FireworkLargeExplosionSound(),
```

ipad54:
For posterity: firework rocket entity also has a long metadata property `ATTACHED_ENTITY_ID` with ID 18 (Bedrock also has `MINECART_HAS_DISPLAY` boolean property with the same id, but it's inconsistency in the game itself). This property is used for elytra boosting: the firework rocket will start to follow the attached entity and give it a movement boost if it glides.

We don't have elytras implemented right now so I think we don't have to bother about this and the PR looks finished for me.

src/entity/object/FireworkRocket.php
```diff
@@ -0,0 +1,201 @@
...
+	 */
+	public function setExplosions(array $explosions) : self{
+		Utils::validateArrayValueType($explosions, function(FireworkRocketExplosion $_) : void{});
+		$this->explosions = $explosions;
+		return $this;
+	}
+
+	/**
+	 * TODO: The entity should be saved and loaded, but this is not possible.
+	 * @see https://bugs.mojang.com/browse/MCPE-165230
```
dktapps: Not liking this. We should still save & restore them in PM, regardless of vanilla bugs that don't store them.

src/entity/object/FireworkRocket.php
```diff
@@ -0,0 +1,201 @@
...
+	protected function getInitialSizeInfo() : EntitySizeInfo{ return new EntitySizeInfo(0.25, 0.25); }
+
+	protected function getInitialDragMultiplier() : float{ return 0.0; }
+
+	protected function getInitialGravity() : float{ return 0.0; }
+
+	/**
+	 * Returns maximum number of ticks this will live for.
+	 */
+	public function getLifeTicks() : int{
```
dktapps: `getMaxAgeTicks()` would be more clear IMO

src/entity/object/FireworkRocket.php
```diff
@@ -0,0 +1,201 @@
...
+		return $this->lifeTicks;
+	}
+
+	/**
+	 * Sets maximum number of ticks this will live for.
+	 *
+	 * @return $this
+	 */
+	public function setLifeTicks(int $lifeTicks) : self{
```
dktapps: ditto here

src/item/FireworkRocket.php
```diff
@@ -0,0 +1,154 @@
...
+		$position = $blockClicked->getPosition()->addVector($clickVector);
+		$position = match($face){
+			Facing::DOWN => $position->add(0, -$correction, 0),
+			Facing::UP => $position->add(0, $correction, 0),
+			Facing::NORTH => $position->add(0, 0, -$correction),
+			Facing::SOUTH => $position->add(0, 0, $correction),
+			Facing::WEST => $position->add(-$correction, 0, 0),
+			Facing::EAST => $position->add($correction, 0, 0),
+			default => throw new AssumptionFailedError("Invalid facing $face")
+		};
```
dktapps:
```suggestion
		$position = $position->add(Vector3::zero()->getSide($face)->multiply($correction));
```

If we allowed `getSide()` to accept float values, this would be even simpler.

src/item/FireworkRocket.php
```diff
@@ -0,0 +1,154 @@
...
+
+		$randomDuration = (($this->flightDurationMultiplier + 1) * 10) + mt_rand(0, 12);
```
dktapps: Prefer if these magic numbers were moved to constants.

dktapps (changes requested):

src/item/FireworkRocket.php
dktapps: Related to pmmp/Math#86

ghost:
Fireworks haven't come since 2022, someone please fix these 😥

src/item/FireworkRocketExplosion.php
```diff
@@ -0,0 +1,190 @@
...
+		$dyeColorIdMap = DyeColorIdMap::getInstance();
+		foreach($colors as $color){
+			$colorsBytes .= Binary::writeByte($dyeColorIdMap->toInvertedId($color));
+		}
+
+		return $colorsBytes;
+	}
+
+	/**
+	 * @param DyeColor[] $colors
```
kostamax27, line 102: It seems that Minecraft allow to set custom colors.

https://github.com/Mojang/bedrock-samples/blob/c32ab2ac8ebdd8b418c03990b30597d6b56bc8c9/resource_pack/texts/en_US.lang#L5121

dktapps: Not too clear how that'd work since the colors are encoded as an array of bytes

dktapps: Never mind I'm tripping - I'm pretty sure custom colours are just combinations of dye colours in different quantities

dktapps (approved):

src/entity/object/FireworkRocket.php
```diff
@@ -149,29 +149,36 @@ public function explode() : void{
...
-				foreach(VoxelRayTrace::betweenPoints($this->location, $position) as $pos){
-					if($world->getBlockAt((int) $pos->x, (int) $pos->y, (int) $pos->z)->isSolid()){
-						continue 2;
+				//cast two rays - one to the entity's feet and another to halfway up its body (according to Java, anyway)
+				//this seems like it'd miss some cases but who am I to argue with vanilla logic :>
+				$height = $entity->getBoundingBox()->getYLength();
+				for($i = 0; $i < 2; $i++){
+					$target = $position->add(0, 0.5 * $i * $height, 0);
+					foreach(VoxelRayTrace::betweenPoints($this->location, $target) as $blockPos){
+						if($world->getBlock($blockPos)->calculateIntercept($this->location, $target) !== null){
```
IvanCraft623, line 171: Dont remember exactly but pretty sure that i compared with isSolid bc it covered most cases, for example damage should be done anyways if there is any flowable block

dktapps: The old check didn't line up with Java when I compared. It didn't make much sense to me for the likes of signs to prevent firework damage.

dktapps: isSolid is also deprecated and returns false for flowables so that wouldn't have done what you wanted either

PR #5464, opened 2022-12-20
Added `Server::broadcastToastNotification()`

supercrafter333:
## Introduction
Plugin developers can use the broadcast methods for tips, titles, message but not for toast-notifications. This PR adds a method to broadcast toast-notifications.

### API changes
- Added the `broadcastToastNotification()` method to the `Server` class

## Tests
I tested this on a clear PMMP-server and it worked well.

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -1332,6 +1332,23 @@ class Server{
 
 		return count($recipients);
 	}
+	
+	/**
+     * Broadcasts a toast-notification (message in the top of the screen).
+     *
+	 * @param string $title
+	 * @param string $body
+	 * @param CommandSender[]|null $recipients
+	 */
+	public function broadcastToastNotification(string $title, string $body, ?array $recipients = null) : int{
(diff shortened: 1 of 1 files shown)


dktapps:
Actually, my intent is to simplify the broadcast system and go the other way.

You can see from the following that there are almost no real usages of the other types of broadcast:
https://poggit.pmmp.io/grepPlugins/2022-12-01-d9df5f09d7050d9d.html
https://poggit.pmmp.io/grepPlugins/2022-12-01-1891688227246b20.html
https://poggit.pmmp.io/grepPlugins/2022-12-01-57a0e3f9b7c7fca8.html

Previously written about on Discord here: https://discord.com/channels/373199722573201408/546255162222706696/1047981575230468126

In addition, this makes almost zero difference to plugin code (3 vs 1 line in a case that is almost never used in practice).

dktapps:
You can also very easily write your own version of this function in a plugin if the 3 lines bothers you this much for the rare occasions you actually need it.

Issue #4077, opened 2021-03-14
Use a LevelDB to store playerdata NBT instead of flat files


dktapps:
## Description
The server should store players' data in a LevelDB database instead of dumping them into lots of .dat files on the disk.

## Justification
The current system of playerdata storage uses flat `.dat` NBT files on the disk. This is horribly inefficient for several reasons:
- Player `.dat` files are typically very small (a couple of kb). Considering that typical filesystems use a 4KB allocation unit size, this means that a significant percentage of disk space gets wasted (e.g. a 6KB data file occupies 2 sectors, totalling 8KB)
- No readahead is possible with many flat files, so the data can't be cached in memory, leading to longer I/O wait times.
- Compression has reduced efficiency. Considering that player.dat files contain a lot of common data (mostly NBT keys), compressing them in batch would produce better compression ratios, saving more space.
- It's difficult to remove playerdata on Linux when the playerdata file count gets too big (e.g. `rm -f players/*.dat` doesn't work anymore).
- Trying to open a directory with thousands of files in it on Windows lags Explorer to death.
- On some filesystems, the playerdata file count may reach or exceed the maximum files per directory limit.
- Players' usernames are used as data keys, so player usernames are constrained by the valid characters which can appear in a filename (e.g. they can't contain / \ " . and so on).

However, it does have the following advantages which would be lost:
- It's currently easy to delete a specific player's data. A tool would need to be provided to do this if LevelDB were used for storage.

## Alternative methods
I've also considered SQLite3, but since PM4 already depends on LevelDB (and LevelDB supports binary data, unlike SQLite3), LevelDB appears to be the better option.

dhdjSYS:
After some testing it seems like leveldb won't support WSL (maybe due to filesystem?). This may be a compatibility issue.

dktapps:
WSL is Microsoft's attempt to simulate Linux within Windows. The only reason to use it is for development. I personally don't use it and we don't lose much if it's not supported anyway.

arfush:
It's a great idea, but I recommend using the UID as the key for the user data.

Player can change gametag. This is a big problem, because the player loses all his data

ColinHDev:
The XUID? What if xbox.auth is disabled?

arfush:
Oh, sorry..I hadn't thought of that.

Maybe add the ability to choose the type of key?

ColinHDev:
That would cause compatibility issues once the setting would be changed

[...]

PR #5467, opened 2022-12-23
LevelDB player data storage

dktapps:
## Introduction
This PR implements a LevelDB player data storage provider.

### Relevant issues
I previously wrote about this in #4077, including its advantages and disadvantages.

### API changes
No public APIs were changed by this PR. However, a new class `LevelDBPlayerDataProvider` has been added to the `pocketmine\player` namespace.

### Behavioural changes
- `player.default-data-format` is added to `pocketmine.yml`, with options `leveldb` and `datfile`.
- `leveldb` is now used by default for **new servers**. Servers with existing data will continue to automatically use the old `datfile` storage provider unless they delete their `pocketmine.yml` or add the new `player.default-data-format` option to their `pocketmine.yml`.

## Backwards compatibility
No backwards compatibility issues currently known.

## Follow-up
Convert data from the `datfile` storage system to `leveldb`. This would require some kind of format detection to enable this.

## Tests
Tested locally using both `datfile` and `leveldb` configuration options. Everything seems to work as expected.


IvanCraft623:
No migrator?

dktapps:
See follow-up section.

dktapps:
Main problem with this PR is that it's not possible to delete a specific player's data. The `PlayerDataProvider` interface would need to be changed to permit this. This isn't a big deal since the interface is de facto internal right now anyway (considering there's no way to provide a custom provider to the `Server` right now anyway), but this should be addressed before merging the PR.

dktapps:
Closing this as I don't think it's worth the migration effort.

Between key ordering, and storing values in the same place as keys (requiring rewriting of values on DB sort), LevelDB is really poorly suited to any form of blob data storage, as we've seen with the performance issues of large worlds.

Considering that player data total size could easily reach multiple GB on highly active servers, this change would likely cause paying a large compaction performance hit. Again, this would be for data ordering which we don't even want anyway.

It may still be worth keeping a LevelDB in the playerdata to allow mapping player identifiers to the files their data are stored in, e.g. LevelDB containing maps of:
- Player name -> .dat file name
- UUID -> .dat name
- XUID -> .dat name

This would allow solving some old issues with player data identification.

However, putting the data in LevelDB directly is likely to cause performance degradation for large servers.

An alternative like RocksDB's [BlobDB](https://rocksdb.org/blog/2021/05/26/integrated-blob-db.html) would be much better suited to a task like this.

Issue #5468, opened 2022-12-24
Add `.self` and `.other` permissions for `/gamemode`


dktapps:
## Description
Currently it's not possible to allow a player to change their own gamemode without also allowing them to change anyone's gamemode. This turned out to be inconvenient for test.pmmp.io.

## Alternative methods
You can write your own /gamemode command in a plugin, but that kinda sucks...

IvanCraft623:
Would it also make sense to implement it for `/teleport` and `/give` ?

dktapps:
For `/teleport` yes. For `/give` it's probably pointless since you can just toss the items on the ground anyway.

dktapps:
List of commands that could benefit from this:

- /gamemode
- /tp
- /give
- /enchant
- /effect
- /spawnpoint
- /title
- perhaps /kick ?

Some other commands already have permissions like this, such as /clear and /kill.

dktapps:
Implemented by #5470.

PR #5470, opened 2022-12-24
Add `.self` and `.other` permissions for gameplay-altering commands

dktapps:
## Introduction

I'm not quite sure this is the best way to enable such functionality, but it's already used for some other stuff, so I'm not too worried for now.

This PR enables more fine-tuned restriction of self- and other-player-targeting commands used to alter gameplay.

This allows the following commands to have their usage limited to self or others:
- /effect
- /enchant
- /gamemode
- /give
- /spawnpoint
- /teleport
- /title

I envision this being useful for creative mode servers, and test servers such as test.pmmp.io.

### Relevant issues
#5468 

### API changes
No public API changes. However, some permissions have been deprecated and will be removed in the future. (Perhaps automatic parent permission group declaration is called for?)

The following new core permissions have been added (these are all given to `pocketmine.group.operator` by default)
- `pocketmine.command.effect.other`
- `pocketmine.command.effect.self`
- `pocketmine.command.enchant.other`
- `pocketmine.command.enchant.self`
- `pocketmine.command.gamemode.other`
- `pocketmine.command.gamemode.self`
- `pocketmine.command.give.other`
- `pocketmine.command.give.self`
- `pocketmine.command.spawnpoint.other`
- `pocketmine.command.spawnpoint.self`
- `pocketmine.command.teleport.other`
- `pocketmine.command.teleport.self`
- `pocketmine.command.title.other`
- `pocketmine.command.title.self`

### Behavioural changes
Without plugins, the server should behave exactly as it did before this change.

If plugins add these permissions to new permission groups (for example `pocketmine.group.user`), they can allow more fine-tuned access to these features.

## Backwards compatibility

## Follow-up
- Consider having parent permission groups like `pocketmine.command.give` auto-generated by the server, so that they don't have to be manually written into the code for backwards compatibility.

## Tests
Tested that functionality works as before without any plugin interference.

I also tested this code:
```php
$perm = PermissionManager::getInstance()->getPermission(DefaultPermissionNames::GROUP_USER);
$perm->addChild(DefaultPermissionNames::COMMAND_GAMEMODE_SELF, true);
```
which grants `pocketmine.command.gamemode.self` to anyone with the `pocketmine.group.user` permission (i.e. all non-operators). This allows the player to change their own game mode, but not that of another player.


SOF3:
great, now we can properly reject all those plugins that try to register a `/c` command for quickly changing self to creative with the excuse of "refined permissions"

dktapps:
Technically speaking those plugins still have a use, because these permissions still don't limit which game modes you can change to.

Issue #5064, opened 2022-05-22
Add a configuration option to disable listing players in GS4 query responses


dktapps:
## Description
Similar to `settings.query-plugins` in `pocketmine.yml`, a `query-player-list` setting might be in order.

## Justification
Possible privacy concerns. It's currently not possible to use GS4 query without telling everyone exactly who is on your server. It's unclear whether or not this is a real risk, but it wouldn't hurt to be able to prevent this, anyway.

## Alternative methods
Disable GS4 query entirely? It's needed for server list sites to work ...

ionaqwyxa:
In response, return a list of players with replaced usernames, for example, player1, player2, and etc.

Gaprix:
> for example, player1, player2, and etc.

What is the purpose of displaying replaced names? What can they be used for?

ionaqwyxa:
> 

Some monitoring sites index players and you can find by the username on which server he was. Impossible to response number of players higher than zero (example 2) and empty player list, response from server will not be read

PR #5476, opened 2022-12-26
QueryInfo->setPlayerList() now accepts string[] instead of Player[]

BrandPVP:
## Introduction
`QueryInfo->setPlayerList()` now accepts `QueryPlayerInfo[]` instead of `Player[]`

This allows to pass fake players/proxy players to the query or protect nicknames from being exposed on online server monitorings (you cant just do `setPlayerList([])` because monitorings will assume that there are no players on the server) https://github.com/pmmp/PocketMine-MP/issues/5064

### API changes
`QueryInfo->setPlayerList()` now accepts `QueryPlayerInfo[]` instead of `Player[]`


dktapps:
Looks to me like we could just pass string[] and drop the class, no?

dktapps (approved):

PR #5481, opened 2022-12-28
style(format): set method_argument_space

xerenahmed:


SOF3:
nit: this isn't a `feat`, just a `chore`, right?

xerenahmed:
You're right, but it's likely `style`

dktapps:
Judging by the CI diff I don't think we're going to gain anything useful from this change. It's changing too many irrelevant things.

SOF3:
This indentation doesn't look reasonable.
image

PR #5482, opened 2022-12-29
Use str_starts_with and str_contains instead of strpos

Gaprix:
In PHP 8.0, the str_starts_with and str_contains functions were introduced. Their use improves the readability of the code.


SOF3 (approved):

dktapps:
There's probably a lot more places where this can be replaced by str_contains or str_ends_with, if you feel like hunting them down.

PR #5485, opened 2022-12-30
Use str_starts_with, str_ends_with and str_contains instead of strpos

Gaprix:
In PHP 8.0, the str_starts_with, str_ends_with and str_contains functions were introduced. Their use improves the readability of the code.

This Pull Request considers the missed cases in #5482

### DIFF
--- a/src/plugin/PharPluginLoader.php
+++ b/src/plugin/PharPluginLoader.php
@@ -35,8 +35,7 @@ class PharPluginLoader implements PluginLoader{
 	){}
 
 	public function canLoadPlugin(string $path) : bool{
-		$ext = ".phar";
-		return is_file($path) && str_ends_with($path, $ext);
+		return is_file($path) && str_ends_with($path, ".phar");
 	}
 
 	/**
--- a/src/plugin/PharPluginLoader.php
+++ b/src/plugin/PharPluginLoader.php
@@ -24,8 +24,7 @@ declare(strict_types=1);
 namespace pocketmine\plugin;
 
 use function is_file;
-use function strlen;
-use function substr;
+use function str_ends_with;
 
 /**
  * Handles different types of plugins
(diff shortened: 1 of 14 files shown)


src/plugin/PharPluginLoader.php
```diff
@@ -37,7 +36,7 @@ public function __construct(
 
 	public function canLoadPlugin(string $path) : bool{
 		$ext = ".phar";
```
dktapps: This variable can be inlined

src/plugin/ScriptPluginLoader.php
```diff
@@ -42,7 +41,7 @@ class ScriptPluginLoader implements PluginLoader{
 
 	public function canLoadPlugin(string $path) : bool{
 		$ext = ".php";
```
dktapps: This variable can be inlined

src/utils/Timezone.php
```diff
@@ -185,11 +184,11 @@ public static function detectSystemTimezone(){
 	 */
 	private static function parseOffset($offset){
 		//Make signed offsets unsigned for date_parse
-		if(strpos($offset, '-') !== false){
+		if(str_contains($offset, '-')){
```
dktapps: I think these can use str_starts_with()

src/utils/Timezone.php
```diff
@@ -185,11 +184,11 @@ public static function detectSystemTimezone(){
...
 			$negative_offset = true;
 			$offset = str_replace('-', '', $offset);
 		}else{
-			if(strpos($offset, '+') !== false){
+			if(str_contains($offset, '+')){
```
dktapps: and here too

SOF3 (approved):

dktapps (approved):

PR #5486, opened 2023-01-02
Fix frosted ice from melting immediately

IvanCraft623:
## Introduction
Currently, when frosted ice is placed on the block, it melts immediately if there are not at least 2 other blocks of frosted ice nearby. This is not the normal behavior in vanilla.

## Tests
Vanilla behavior: https://cdn.discordapp.com/attachments/715214348338855987/1059349977480052806/vanilla_frosted_ice_test.mp4


dktapps:
To test vanilla behaviour, you really need to observe how it behaves with frost walker boots. Frosted ice never appears as a single block like this, and there has to be a reason why that check was added.

IvanCraft623:
> To test vanilla behaviour, you really need to observe how it behaves with frost walker boots. Frosted ice never appears as a single block like this, and there has to be a reason why that check was added.

So that is a behavior of the enchantment not of the block.

dktapps:
It doesn't really matter. Frost walker is the only regular source of frosted ice. Any other source therefore has undefined behaviour anyway.

IvanCraft623:
> It doesn't really matter. Frost walker is the only regular source of frosted ice. Any other source therefore has undefined behaviour anyway.

However, I just tested the enchant and only one block is able to appear.

IvanCraft623:
https://cdn.discordapp.com/attachments/546255162222706696/1060293403503501353/frost_walker_test.mp4

IvanCraft623:
Test 2: https://cdn.discordapp.com/attachments/546255162222706696/1060295460511817849/frost_walker_test_2.mp4

dktapps:
The existing logic matches Java, but with one subtle difference: In Java, the melting only takes place if the block's age is 0. Otherwise, the ice is allowed to melt normally.

This logic is likely the same used in Bedrock, and would explain the difference in behaviour observed.

dktapps:
It seems that Bedrock indeed does not have this logic at all.

Issue #3371, opened 2020-03-21
Command alias rework


SOF3:
# Motivation
The current command system is largely derived from bukkit, suffering from the following problems:
- to customize command names, plugin has to make their own config entry and register the command dynamically
- although alias can be set in pocketmine.yml, it is often unknown to most users, and even if it is, managing a patch layer on top of names and aliases registered by plugins is a pain since it is unclear what gets used and what doesn't
- users cannot unlink an alias from a command (for whatever reason)
- plugins trying to replace a command from the core or another plugin might suffer from undefined behavior due to racing overwrites
- it is not trivial for plugins to remap command names dynamically without implementing their own CommandMap
- plugins with the same command name race to use an alias space with undefined behavior

This proposal aims to solve these problems by redefining the concepts of command name and aliases and creating an alternative alias configuration. (Sounds orthogonal, but they are closely related)

## Proposal
The command name defined by plugins is regarded as a semi-internal identifier for a command (at a slightly higher level than function names). The command name is not expected to be used at user interface, but can still be referred to with `pluginname:commandname` (this is also the full name to identify the command externally).

A command alias is a way for a user to access a command through chat window. The command name is an implicit alias for the command, along with extra aliases as specified by the command.

When a plugin registers a command, PocketMine tracks if this command (as identified by its full name) was previously registered in previous runs. If it was never registered, for each alias, PocketMine overwrites the alias' entry in a file `$DATA_PATH/commands.yml` to this command full name. PocketMine also maintains a map [full name => Command object].

The AliasMap interface is introduced, with the single method `getCommand(string $name): ?Command`.

The only builtin implementation of AliasMap is FileAliasMap, which retrieves $name from the parameterized file (specified as commands.yml in server initialization code), then resolving the full name into the Command object.

The Player class (see Unresolved section) exposes getset for an aliasMap field, which is used to resolve the command when a player executed from the chat window. It is by default initialized with FileAliasMap("$DATA_PARH/commands.tml"). This can be set by olugins to alternative implementations, such as ones that delegate to player-specific files or world-specific files. This field should be settable under all reasonable states.

# Unresolved
- How should other CommandSender implementations behave? Should they also use an AliasMap, or is it just specific to Player?
- Should PocketMine warn/error/ignore/continue if a plugin is registering an alias that overwrites an existing one?

dktapps:
tl;dr: you want a config file which allows command aliases to be edited, remapped and removed, and not just added.

[...]

PR #5487, opened 2023-01-02
Hard code vanilla command's names in their __constructor

BrandPVP:
## Introduction
So, now all the vanilla commands pass their name from their constructor like the other things (aliases, permissions, description..)

### API changes
No

### Behavioural changes
No

## Backwards compatibility
BC break, I think?


SOF3:
what is the motivation of this change?

dktapps:
Consistency, I suppose? It doesn't make any sense to declare the names outside the constructor, but the aliases inside it.

That being said, I'm also doubtful that this provides any meaningful benefit to justify the BC break.

jasonw4331:
Wouldn't it be better to supply translations for all these instead of hard coding the command names?

dktapps:
No. Command names can't be translated.

SOF3:
I wouldn't consider this to be a BC break since vanilla commands should not be part of the public API.

That said, I have been thinking about configurable command names, but in fact, command names should not be configurable, only aliases should be (see #3371 and related discussions).

SOF3 (approved):

dktapps:
> That said, I have been thinking about configurable command names, but in fact, command names should not be configurable, only aliases should be (see https://github.com/pmmp/PocketMine-MP/issues/3371 and related discussions).

I've recently considered getting rid of the idea of "command name" entirely. The only real use for `Command->getName()` is with `CommandExecutor` implementors (essentially a locally unique command ID, not unique globally). Any other usage of a "primary" command alias uses `getLabel()`.

SOF3:
Let's move that discussion to #3371 anyway. Changes to the command alias mechanism should not be a blocker for this PR.

PR #5493, opened 2023-01-04
ItemStackResponseBuilder: Remove unnecessary ternary operator

IvanCraft623:
## Introduction

This operator is unnecessary because `Item::getCustomName()` already returns an empty string if it doesn't have a custom name.


dktapps:
Hmm, you are correct. I was thinking of Item::getName().

PR #5494, opened 2023-01-05
Use existing function to notify nearby blocks of an update

IvanCraft623:
## Introduction
Just code clean up :D


ghost:
Does this breaks blocks behaviour of Fallable?

IvanCraft623:
> Does this breaks blocks behaviour of Fallable?

No, it is just a code clean up

PR #5497, opened 2023-01-05
Implement frost walker enchantment

IvanCraft623:
## Introduction
As title.

## Follow-up
See #5486 

## Tests
https://cdn.discordapp.com/attachments/1060620088920772688/1060659547682898010/Frost_walker_test.mp4


src/entity/Living.php
```diff
@@ -659,6 +663,32 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
 		return $hasUpdate;
 	}
 
+	protected function move(float $dx, float $dy, float $dz) : void{
+		parent::move($dx, $dy, $dz);
+
+		$frostWalkerLevel = $this->armorInventory->getBoots()->getEnchantmentLevel(VanillaEnchantments::FROST_WALKER());
+		if($frostWalkerLevel > 0 && (abs($dx) > 0.0001 || abs($dz) > 0.0001)){
```
dktapps: You can't use `dx` and `dz` here. These parameters are the _desired_ movement distance, not the actual distance moved.

src/entity/Living.php
```diff
@@ -659,6 +663,32 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
 		return $hasUpdate;
 	}
 
+	protected function move(float $dx, float $dy, float $dz) : void{
+		parent::move($dx, $dy, $dz);
+
+		$frostWalkerLevel = $this->armorInventory->getBoots()->getEnchantmentLevel(VanillaEnchantments::FROST_WALKER());
```
dktapps: It might be a good idea to remember this in a property somewhere and update it on armour change. move() is a hot code path, and we don't want to compromise its performance. We already have a listener on the armour inventory for broadcasting updates anyway.

src/entity/Living.php
IvanCraft623: What should I use instead?

dktapps: You would need to record the position before and after the parent::move() call.

A possible option could be to override `checkGroundState()`, but it wasn't really intended for this usage, so it might break in other ways.

src/entity/Living.php
IvanCraft623: Ughh, there is a problem with that, there is no good way to initialize the property. `Living` does not initialize the armor, e.g. if a child class (e.g. Player) spawns with boots, the only way to get the boots would be after child's `initEntity` is executed.

IvanCraft623: I mean, to initialize the property I would have to force the child classes to define it or abuse something like `onFirstUpdate`

IvanCraft623: Ignore the above I had a brain.exe not working moment, is 1am and I have slept very little in the last 24 hours.

dktapps (dismissed):
I think this is done

dktapps (dismissed):

PR #5502, opened 2023-01-08
Implement Recovery compass

IvanCraft623:
## Introduction
Implements recovery compass, this item points to the position of the player's last kill, the position is stored with the player data and send via entity properties.

### API changes

The following methods has been added to the `Player` class:
- `Player:getDeathPosition() : ?Position`
- `Player:setDeathPosition(?Position $position) : void`

## Follow up
- Update BedrockPtotocol!!!

## Tests
https://youtu.be/cdTEHqx5c7U


src/player/Player.php
```diff
@@ -2248,6 +2280,15 @@ public function getSaveData() : CompoundTag{
 			$nbt->setInt(self::TAG_SPAWN_Z, $spawn->getFloorZ());
 		}
 
+		if($this->deathPosition !== null){
+			$nbt->setString(self::TAG_DEATH_WORLD, $this->deathPosition->getWorld()->getFolderName());
+			$nbt->setInt(self::TAG_DEATH_X, $this->deathPosition->getFloorX());
+			$nbt->setInt(self::TAG_DEATH_Y, $this->deathPosition->getFloorY());
+			$nbt->setInt(self::TAG_DEATH_Z, $this->deathPosition->getFloorZ());
+		}else{
+			$nbt->removeTag(self::TAG_DEATH_WORLD, self::TAG_DEATH_X, self::TAG_DEATH_Y, self::TAG_DEATH_Z);
```
dktapps: This is redundant. The NBT is freshly created, so it won't have these tags anyway.

src/player/Player.php
```diff
@@ -2248,6 +2280,15 @@ public function getSaveData() : CompoundTag{
 			$nbt->setInt(self::TAG_SPAWN_Z, $spawn->getFloorZ());
 		}
 
+		if($this->deathPosition !== null){
+			$nbt->setString(self::TAG_DEATH_WORLD, $this->deathPosition->getWorld()->getFolderName());
```
dktapps, line 2416: This may throw an exception if the world was unloaded since the player died

src/player/Player.php
```diff
@@ -969,6 +979,27 @@ public function doChunkRequests() : void{
...
+	public function getDeathPosition() : ?Position{
+		return $this->deathPosition;
+	}
+
+	/**
+	 * @param Vector3|Position|null $pos
+	 */
+	public function setDeathPosition(?Vector3 $pos) : void{
+		if($pos !== null){
+			if($pos instanceof Position && $pos->world !== null){
```
dktapps, line 1057: I think it's fine to just assert that the position must be valid here. We shouldn't be setting death position to an unloaded world anyway.

dktapps (changes requested):

src/player/Player.php
IvanCraft623: Would you agree to store the world's name in a var on `Player::setDeathPosition()`?

dktapps: It should probably be done in a similar way to spawn position handling, where the spawn is invalidated if the world isn't loaded. I'm not sure this is the _best_ way, but it should be consistent.

src/player/Player.php
dktapps: Apparently this is the same as how the spawn setting logic works. Perhaps we need to take a hard look at why Position is allowed to have a null `World`...

dktapps (approved):

PR #5505, opened 2023-01-09
Ring bell when is hit by a projectile

IvanCraft623:
## Introduction

As title, I would like to target this to next-minor, but because `Block::onProjectileHit()` only exists in next-major it is not possible.


src/block/Bell.php
```diff
@@ -152,6 +154,22 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
 		return true;
 	}
 
+	public function onProjectileHit(Projectile $projectile, RayTraceResult $hitResult) : void{
+		$faceHit = $hitResult->getHitFace();
+		if($this->attachmentType->equals(BellAttachmentType::CEILING())){
+			$this->ring($faceHit);
+		}
+		if($this->attachmentType->equals(BellAttachmentType::FLOOR()) && Facing::axis($faceHit) === Facing::axis($this->facing)){
```
dktapps: This should be using `elseif`

src/block/Bell.php
```diff
@@ -152,6 +154,22 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+			$this->ring($faceHit);
+		}
+		if(
```
dktapps: same again here

dktapps: In fact all of these conditions can be merged, since the bodies of all of them are identical.

dktapps (changes requested):

dktapps:
The base code was changed, you'll need to merge changes.

IvanCraft623:
> The base code was changed, you'll need to merge changes.

done

src/block/Bell.php
```diff
@@ -158,4 +160,16 @@ public function ring(int $faceHit) : void{
...
+
+	private function isValidFaceToRing(int $faceHit) : bool{
+		return (
+			$this->attachmentType->equals(BellAttachmentType::CEILING()) ||
+			($this->attachmentType->equals(BellAttachmentType::FLOOR()) && Facing::axis($faceHit) === Facing::axis($this->facing)) ||
+			(
+				$this->attachmentType->equals(BellAttachmentType::ONE_WALL()) ||
+				$this->attachmentType->equals(BellAttachmentType::TWO_WALLS())) &&
+				($faceHit === Facing::rotateY($this->facing, false) || $faceHit === Facing::rotateY($this->facing, true)
+			)
```
dktapps, line 171: The braces in this block are a mess. I can't make head or tail of what's going on.

IvanCraft623: It's exactly the code you wrote

dktapps: My code:
image
Your code:
image

I don't know what happened to the formatting here, but it's a confusing mess.

IvanCraft623: Your code has been successfully copied and pasted

PR #5526, opened 2023-01-23
Add the parameter `int $step` to `Block::getAllSides()`

NhanAZ:
## Introduction
This parameter works as the `int $step` parameter of `Vector3::sides`

### DIFF
--- a/src/block/Block.php
+++ b/src/block/Block.php
@@ -640,12 +640,14 @@ class Block{
 	/**
 	 * Returns the six blocks around this block.
 	 *
+	 * @param int $step Distance in each direction to shift the vector
+	 *
 	 * @return Block[]|\Generator
 	 * @phpstan-return \Generator<int, Block, void, void>
 	 */
-	public function getAllSides() : \Generator{
+	public function getAllSides(int $step = 1) : \Generator{
 		$world = $this->position->getWorld();
-		foreach($this->position->sides() as $vector3){
+		foreach($this->position->sides($step) as $vector3){
 			yield $world->getBlock($vector3);
 		}
 	}


dktapps:
I'm not really sure what the point of this is. Beyond a step of 1, you're just getting random blocks that just happen to be on the same axis as the block. There's no use case that I can see.

SOF3:
What was the point of `$vector3->sides($steps)` to begin with?

dktapps:
Good question. I think it came about during my attempt to separate Position from Block. The steps was a random thing I was able to add when this function was first introduced to begin with.

SOF3:
@NhanAZ do you have any user stories for this? I don't see chances this is getting merged if there is no justification for such a confusing API.

NhanAZ:
A simple use of this parameter: Create a cobblestone farm using water and fence
Here is the example code:
```php
	public function onBlockUpdate(BlockUpdateEvent $event): void {
		$block = $event->getBlock();
		if ($block instanceof Fence) {
			foreach ($block->getAllSides(2) as $key => $side) {
				if ($side instanceof Water) {
					$nearBlock = $block->getSide($key);
					if ($nearBlock->isSameType(VanillaBlocks::COBBLESTONE())) {
						continue;
					}
					$nearBlockPos = $nearBlock->getPosition();
					$nearBlockPos->getWorld()->setBlock($nearBlockPos, VanillaBlocks::COBBLESTONE(), false);
				}
			}
		}
	}
```
Cobblestone is generated when near the fence there is water.

SOF3:
@NhanAZ what can't you flip the two if blocks and check the cobblestone first?

dktapps:
This doesn't seem like a realistic use case.

What your code is doing is essentially this (in 2D):

```
..W..
.....
W.F.W
.....
..W..
```
The blocks closest to the fence are ignored.

Even if this was actually desired behaviour, it doesn't seem like a good enough use case to justify this change. You can simply do getSides() and then getSide(face) inside the loop to step out one block further (which your code actually does already to check for cobblestone).

NhanAZ:
Thanks for taking the time to review this PR.

PR #5537, opened 2023-01-26
ci: add appimage build to release flow

SOF3:
Provide a simple binary download for Linux without the complexity of setting up docker or using an installer.

PR #5538, opened 2023-01-26
ci: add appimage build to release flow

SOF3:
Provide a simple binary download for Linux without the complexity of setting up docker or using an installer.

Replaces the accidentally-closed #5537

### DIFF
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -48,6 +48,10 @@ jobs:
       - name: Build PocketMine-MP.phar
         run: php -dphar.readonly=0 build/server-phar.php --git ${{ github.sha }} --build ${{ steps.build-number.outputs.BUILD_NUMBER }}
 
+      - name: Build appimage
+        run: build-appimage.sh
+        working-directory: build
+
       - name: Get PocketMine-MP release version
         id: get-pm-version
         run: |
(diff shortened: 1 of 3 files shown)


SOF3:
@dktapps https://github.com/SOF3/PocketMine-MP/releases/tag/4.993.994

dktapps:
For posterity, the related discussion from #internals-dev:

> some kind of comedy genius — 26/01/2023 16:58
yuck, it has to build the PHP binary from scratch?
i'm getting the vibe this should be a separate workflow like the docker image build

> the correctly useless dev — 26/01/2023 16:59
well, we could reuse the existing one
I just wanted it to be more reproducible
it is totally fine to reuse existing binaries, maybe I should add params for that

> some kind of comedy genius — 26/01/2023 17:00
i don't think it can reuse the binary from the action
it uses system libraries

> the correctly useless dev — 26/01/2023 17:00
what system libraries?
PHP itself is not system libs right?
I'm not doing any static linking now
it still depends on the system
so it's not as cool as docker tbh
it's just avoiding downloading a bunch of files

> some kind of comedy genius — 26/01/2023 17:01
i don't really see the point tbh
get.pmmp.io is already a piece of cake
and it will automatically build binaries for you if the existing ones aren't suitable

> the correctly useless dev — 26/01/2023 17:01
yeah but it still extracts a bunch of files

> some kind of comedy genius — 26/01/2023 17:01
yeah but id on't hear anyone complaining
it's windows we need this for, not linux

> the correctly useless dev — 26/01/2023 17:02
true, maybe there isn't really a point
yeah
was just playing around
would be cool to be able to publish it on another repository though
looks like it helps with some Linux desktop installers as well
although it's probably still pointless without an actual GUI

dktapps:
Since this has no obvious use case, I don't think it justifies adding +5 minutes to the release workflow, so I'm going to close this.

PR #5540, opened 2023-01-29
Fix packet flood attack

FazeNextuss:
## Introduction
- This fix aims to block malicious attacks (flood of packets) by adding a limit. -

## Fix
- Added a packet limit sent by a player within a given time. (More than 5 packets in 2 seconds). If the player exceeds this limit, he will be kicked for "Too fast packet". 

## Test
- This has been tested, and now blocks these attacks.


src/network/mcpe/NetworkSession.php
```diff
@@ -179,7 +178,7 @@ class NetworkSession{
 
 	/** @var Packet[] */
 	private array $sendBuffer = [];
-
+	private array $packetsReceived = [];
```
DaisukeDaisuke, line 183:
```suggestion
	/** @var array<string, array{count: int, time: int}> */
	private array $packetsReceived = [];
```

src/network/mcpe/NetworkSession.php
```diff
@@ -419,10 +418,31 @@ public function handleEncoded(string $payload) : void{
...
+		$playerName = $this->getPlayer()?->getName();
+		if (!isset($this->packetsReceived[$playerName])) {
+			$this->packetsReceived[$playerName] = [
+				'count' => 0,
+				'time' => time(),
+			];
+		}
+
+		$playerPackets = &$this->packetsReceived[$playerName];
+		$playerPackets['count']++;
```
DaisukeDaisuke, line 433: I think passing by reference causes confusion
(It can be split into two properties)

src/network/mcpe/NetworkSession.php
```diff
@@ -419,10 +418,31 @@ public function handleEncoded(string $payload) : void{
...
 		if(!($packet instanceof ServerboundPacket)){
 			throw new PacketHandlingException("Unexpected non-serverbound packet");
 		}
 
+		if (strlen($buffer) > 16500 && $packet->getName() === "InventoryTransactionPacket") {
```
DaisukeDaisuke:
```php
use pocketmine\network\mcpe\protocol\DisconnectPacket;
use pocketmine\network\mcpe\protocol\EmotePacket;
use pocketmine\network\mcpe\protocol\InventoryTransactionPacket; // new import
use pocketmine\network\mcpe\protocol\MobArmorEquipmentPacket;
```
```suggestion
		if (strlen($buffer) > 16500 && $packet instanceof InventoryTransactionPacket) {
```

FazeNextuss:
Thanks for your help.
Right now correct my mistakes.

src/network/mcpe/NetworkSession.php
```diff
@@ -419,10 +419,31 @@ public function handleEncoded(string $payload) : void{
...
+			$this->logger->debug("Huge InventoryTransactionPacket: " . base64_encode($buffer));
+			throw new PacketHandlingException("InventoryTransactionPacket too big");
+		}
+		if (strlen($buffer) > 31968) {
+			if (time() - $playerPackets['time'] >= 2 && $playerPackets['count'] > 5) {
```
NhanAZ: `16500`, `31968`
magic number

src/network/mcpe/NetworkSession.php
```diff
@@ -419,10 +419,31 @@ public function handleEncoded(string $payload) : void{
...
+				$this->getPlayer()?->kick("Too fast packet");
```
DaisukeDaisuke: Here there is no interruption
```suggestion
				throw new PacketHandlingException("Too fast packet");
```

src/network/mcpe/NetworkSession.php
```diff
@@ -419,10 +419,31 @@ public function handleEncoded(string $payload) : void{
...
 	 */
 	public function handleDataPacket(Packet $packet, string $buffer) : void{
+
+		$playerName = $this->getPlayer()?->getName();
+		if (!isset($this->packetsReceived[$playerName])) {
+			$this->packetsReceived[$playerName] = [
+				'count' => 0,
+				'time' => time(),
+			];
+		}
```
DaisukeDaisuke, line 430: After initializing the `packetsReceived` property, the `time` appears unchanged
```
array(2) {
  ["count"]=>
  int(6557)
  ["time"]=>
  int(1674995763)
}
```

dktapps:
There is already a limit.

Viska2007:
Hello, I just saw that you just closed my pull request. But your limit doesn't work. The person who has the code that allows down the servers with what I did is now impossible. You what you have done is a basic "limit" but what it does is that it sends a lot of requests in a given time. What I have done is that in 2 seconds, it cannot send 5 packets of more than approximately 30000 bits. After having tested this flaw by having put my fix, it no longer works.

## Here is the packet which is spammy and quite big which freezes the server (encrypted):
Unhandled PlayerAuthInputPacket: kAEAAMB/AADAfwAAtUP0PetCAICRQwAAAAAAAAAAAADAf4CAgIiAAQECAgAAAAAAAAAAAAAAAACg/

## By putting my fix:
devfix9999

PR #5549, opened 2023-02-02
Update Minecraft: PE header to Minecraft: BE

Gaprix:


dktapps:
I've never been much of a fan of this abbreviation. `Bedrock Edition` would be better.

PR #5559, opened 2023-02-08
Implement DataPacketPreReceiveEvent

### LINKED ISSUE #5554: DataPacketPreReceiveEvent
## Description
Currently we're forced to spend CPU time decoding junk packets from the client, because plugins might want to access their data in DataPacketPreReceiveEvent.

Cases where this is wasting resources include CraftingDataPacket, MobArmorEquipmentPacket, LevelSoundEventPacket and others. None of these packets are used by vanilla PM.

A DataPacketPreReceiveEvent would include a raw packet buffer, allowing plugins to capture and handle packets that the server would otherwise ignore. This will allow the server to drop the useless packets instead of decoding them.

## Alternative methods
Possibly, some way to detect if the current PacketHandler implements a handler for the currently received packet could be used to determine whether or not to decode the packet in advance, but this still doesn't solve the DataPacketReceiveEvent problem.

aderoian:
## Introduction
As said in #5554, some packets do not need to go through the decoding process. This implements an event that allows a set of ignored packets to be set. These packets will be intercepted and denied decoding if they are flagged as ignored.

### Relevant issues
Closes #5554 

### API changes
Creation of the event ``pocketmine\event\server\DataPdacketPreReceiveEvent``

### Behavioural changes
Some packets are not decoded, therefore not handled. Some sort of performance gain should be expected.

## Tests
```php

	public function onEnable(): void{
	}

	public function onPreDataPacketReceiveEvent(DataPacketPreReceiveEvent $event): void{
		$packets = $event->getIgnoredPackets();
		$packets[] = ProtocolInfo::INVENTORY_TRANSACTION_PACKET;
		$event->setIgnoredPackets($packets);
	}
}
```
A plugin that flags the ``Inventory Transaction Packet`` as "ignored". All inventory actions (i.e dropping items) are never handled by the server.


IvanCraft623:
This should be targeted to `next-major` since it contains BC-Break

JavierLeon9966:
I don't see any BC breaks for my part, just behavior changes. For example, plugins will no longer receive `LevelSoundEventPacket`s' for cps

src/event/server/DataPacketPreReceiveEvent.php
```diff
@@ -0,0 +1,113 @@
...
+	public const MAP_INFO_REQUEST_PACKET = ProtocolInfo::MAP_INFO_REQUEST_PACKET;
+	public const BOSS_EVENT_PACKET = ProtocolInfo::BOSS_EVENT_PACKET;
+	public const SHOW_CREDITS_PACKET = ProtocolInfo::SHOW_CREDITS_PACKET;
+	public const COMMAND_BLOCK_UPDATE_PACKET = ProtocolInfo::COMMAND_BLOCK_UPDATE_PACKET;
+	public const SUB_CLIENT_LOGIN_PACKET = ProtocolInfo::SUB_CLIENT_LOGIN_PACKET;
+	public const SERVER_SETTINGS_REQUEST_PACKET = ProtocolInfo::SERVER_SETTINGS_REQUEST_PACKET;
+	public const LAB_TABLE_PACKET = ProtocolInfo::LAB_TABLE_PACKET;
+	public const NETWORK_STACK_LATENCY_PACKET = ProtocolInfo::NETWORK_STACK_LATENCY_PACKET;
+	public const LEVEL_SOUND_EVENT_PACKET = ProtocolInfo::LEVEL_SOUND_EVENT_PACKET;
+
```
dktapps: I don't see the point of these constants.

src/event/server/DataPacketPreReceiveEvent.php
```diff
@@ -0,0 +1,113 @@
...
+		self::MAP_INFO_REQUEST_PACKET,
+		self::BOSS_EVENT_PACKET,
+		self::SHOW_CREDITS_PACKET,
+		self::COMMAND_BLOCK_UPDATE_PACKET,
+		self::SUB_CLIENT_LOGIN_PACKET,
+		self::SERVER_SETTINGS_REQUEST_PACKET,
+		self::LAB_TABLE_PACKET,
+		self::NETWORK_STACK_LATENCY_PACKET,
+		self::LEVEL_SOUND_EVENT_PACKET
+	];
```
dktapps: This shouldn't be defined here. If we decide to start using these packets in the future, this is going to be a surprising footgun for anyone who wants to use them.

dktapps (changes requested):
This is rather more complex than I'd envisioned. My thought was to pass the packet buffer string to the event directly, which would allow plugins to decide how to handle it, as well as allowing junked packets to be logged. An ignored packet would be signalled by pre-cancelling the event.

src/event/server/DataPacketPreReceiveEvent.php
aderoian: I agree, however, I feel that there should be a list of the "junk" packets somewhere. A lot of those are marked as todo and can be removed, yes.

aderoian:
I am not sure if that if statement is the best place to check for certain packets, but it's much better than the original event. Additionally, this change possibly should be in ``next-major`` as any plugin depending on one of these packets will be affected.

dktapps:
To me, it seems that deciding to blacklist some packets is not really relevant to the original intent here, which is to simply add a DataPacketPreReceiveEvent. Where and how to blacklist packets is a problem that can be discussed separately.

aderoian:
Just to be certain, we only want to be calling the event, no pre-canceling of the event?

dktapps:
for now yes

src/event/server/DataPacketPreReceiveEvent.php
```diff
@@ -0,0 +1,51 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\event\server;
+
+use pocketmine\event\Cancellable;
+use pocketmine\event\CancellableTrait;
+use pocketmine\network\mcpe\NetworkSession;
+
+class DataPacketPreReceiveEvent extends ServerEvent implements Cancellable{
```
dktapps, line 34: Please add documentation for this.

src/network/mcpe/NetworkSession.php
```diff
@@ -426,6 +427,12 @@ public function handleDataPacket(Packet $packet, string $buffer) : void{
 			throw new PacketHandlingException("Unexpected non-serverbound packet");
 		}
 
+		$ev = new DataPacketPreReceiveEvent($this, $packet->pid(), $buffer);
+		$ev->call();
+		if($ev->isCancelled()){
+			return;
+		}
+
```
dktapps: The event call should be covered by receiveDataPacket timings.

src/event/server/DataPacketPreReceiveEvent.php
```diff
@@ -0,0 +1,55 @@
...
+ */
+class DataPacketPreReceiveEvent extends ServerEvent implements Cancellable{
+	use CancellableTrait;
+
+	public function __construct(
+		private NetworkSession $origin,
+		private int $packetId,
+		private string $packetBuffer
+	){
+	}
```
dktapps: Formatting issue here (remove the extra newline)

src/network/mcpe/NetworkSession.php
```diff
@@ -428,6 +429,12 @@ public function handleDataPacket(Packet $packet, string $buffer) : void{
 
 		$timings = Timings::getReceiveDataPacketTimings($packet);
 		$timings->startTiming();
+		$ev = new DataPacketPreReceiveEvent($this, $packet->pid(), $buffer);
+		$ev->call();
+		if($ev->isCancelled()){
+			$timings->stopTiming();
```
dktapps: You can put the event call inside the `try` and avoid having this extra call (the `finally` will take care of it).

src/network/mcpe/NetworkSession.php
```diff
@@ -427,9 +428,14 @@ public function handleDataPacket(Packet $packet, string $buffer) : void{
 		}
 
 		$timings = Timings::getReceiveDataPacketTimings($packet);
-		$timings->startTiming();
-
 		try{
+			$timings->startTiming();
```
dktapps: This line can be moved outside the `try` as well

IvanCraft623:
I think event's name isn't accurate, it may not be understood the concept of "PreReceive" because to understand what the event actually does you have to read the documentation, it would make more sense "PreDecode".

dktapps:
To be honest I'd like to rename DataPacketReceiveEvent to DataPacketHandleEvent, and have this be DataPacketReceiveEvent instead. But that would cause a confusing mess for plugins.

PR #5566, opened 2023-02-11
Implement new methods in Config

victoriovm:
## Introduction
Implementation of new GET methods in config

### API changes
New methods implemented in Config
`Config->getInt()`
`Config->getString()`
`Config->getFloat()`
`Config->getArray()`
`Config->getBoolean()`

`Config->getNestedInt()`
`Config->getNestedString()`
`Config->getNestedFloat()`
`Config->getNestedArray()`
`Config->getNestedBoolean()`

## Tests

```php
<?php

declare(strict_types=1);

namespace test;

use pocketmine\utils\Config;
use function var_dump;

final class Loader extends PluginBase {
	protected function onEnable(): void {
		$test = new Config($this->getDataFolder() . 'test.yml', Config::YAML, [
			'string_value' => 'test',
			'int_value' => 1,
			'float_value' => 1.2,
			'array_value' => [1, 2, 3],
			'bool_value' => true,
			'nested_test' => [
				'string_value' => 'test',
				'int_value' => 1,
				'float_value' => 1.2,
				'array_value' => [1, 2, 3],
				'bool_value' => true
			]
		]);
		var_dump($test->getInt('int_value'));
		var_dump($test->getBoolean('bool_value'));
		var_dump($test->getArray('array_value'));
		var_dump($test->getFloat('float_value'));
		var_dump($test->getString('string_value'));

		var_dump($test->getNestedInt('nested_test.int_value'));
		var_dump($test->getNestedBoolean('nested_test.bool_value'));
		var_dump($test->getNestedArray('nested_test.array_value'));
		var_dump($test->getNestedFloat('nested_test.float_value'));
		var_dump($test->getNestedString('nested_test.string_value'));

		var_dump($test->getString('int_value', 'invalid value'));
	}
}
```
Result:
```
int(1)

bool(true)

array(3) {
  [0]=>
  int(1)
  [1]=>
  int(2)
  [2]=>
  int(3)
}

float(1.2)

string(4) "test"

int(1)

bool(true)

array(3) {
  [0]=>
  int(1)
  [1]=>
  int(2)
  [2]=>
  int(3)
}

float(1.2)

string(4) "test"

string(13) "invalid value"
```


NhanAZ (dismissed):

SOF3:
it is probably a better idea to directly validate the whole config schema in advance using libraries like Marshal

dktapps:
Due to yaml coercion, the types of things in the config might not be what you expect. For example `version: 4.0` will return a float rather than a string.

SOF3:
> Due to yaml coercion, the types of things in the config might not be what you expect. For example `version: 4.0` will return a float rather than a string.

We can't really do anything about that. The best we can do is to warn the user about it with an explicit error.

dktapps:
My point is that the behaviour of these functions silently differs from the behaviour of a `(string) getProperty("usually-a-string")`, and the code is not able to predict when the type will be wrong due to a user "error". I don't think erroring or returning default is correct here.

SOF3:
I think error makes sense on wrong type, and default should only be returned on missing.

dktapps:
I think we should allow int and float values for strings, as long as they are casted before returning. Same with int values for float properties. Otherwise we're gonna get a lot of confused people.

SOF3:
but it would cause incorrect result when people write 4.0. would it be better to just throw error with quotes?

dktapps:
I don't see how it would be incorrect. I'm talking about making YAML return values sane. Right now callers have to cast externally, and this change is going to force the stupidness of YAML onto users. I'm saying that `getString()` e.g. should cast floats and ints to string if found.

In any case, it should **definitely** not be silently returning default on wrong value types.

SOF3:
the approach I used in Capital is like this:
1. parse the whole YAML strictly 
2. if parsed type is incorrect,
  1. convert to the correct type
  2. push the error message about incorrect type to a buffer
3. after parsing everything, if the error buffer is not empty, write the converted file back into the config and display the error buffer

if the user doesn't care and just restarts, it is already fixed anyway. but users who care would have the chance to see that something got changed.

the disadvantage is that you have to parse the whole config in advance so that all corrections can be made in one restart.

src/utils/Config.php
```diff
@@ -396,6 +396,30 @@ public function getNested($key, $default = null){
 		return $this->nestedCache[$key] = $base;
 	}
 
+	public function getNestedString(string $key, string $default = null) : ?string {
+		return is_string($value = $this->getNested($key)) ? $value : $default;
```
dktapps: This needs to cast `int` and `float` to `string` (yaml may cause strings such as `4.0` to be returned as floats)

src/utils/Config.php
```diff
@@ -396,6 +396,30 @@ public function getNested($key, $default = null){
...
+	}
+
+	public function getNestedInt(string $key, int $default = null) : ?int {
+		return is_int($value = $this->getNested($key)) ? $value : $default;
+	}
+
+	public function getNestedFloat(string $key, float $default = null) : ?float {
+		return is_float($value = $this->getNested($key)) ? $value : $default;
```
dktapps: This needs cast `int` to `float` (same as above)

src/utils/Config.php
```diff
@@ -441,6 +465,30 @@ public function set($k, $v = true) : void{
 		}
 	}
 
+	public function getString(string $key, string $default = null) : ?string {
+		return is_string($value = $this->config[$key]) ? $value : $default;
```
dktapps: This needs to cast `int` and `float` to `string` (same as above).

src/utils/Config.php
```diff
@@ -441,6 +465,30 @@ public function set($k, $v = true) : void{
...
+	public function getInt(string $key, int $default = null) : ?int {
+		return is_int($value = $this->config[$key]) ? $value : $default;
+	}
+
+	public function getBoolean(string $key, bool $default = null) : ?bool {
+		return is_bool($value = $this->config[$key]) ? $value : $default;
+	}
+
+	public function getFloat(string $key, float $default = null) : ?float {
+		return is_float($value = $this->config[$key]) ? $value : $default;
```
dktapps: This needs cast `int` to `float` (same as above).

dktapps (changes requested):

jasonw4331 (dismissed):

dktapps (changes requested):
To quote previous comments:

> In any case, it should definitely not be silently returning default on wrong value types.

This has not been addressed.

dktapps (commented):
Looks OK apart from formatting. If you use PhpStorm, press ctrl+alt+L to format the file.

Otherwise, I'll deal with it when I find the time.

victoriovm:
I use vscode.

AnyBananaGAME:
> I use vscode.

SHIFT+ALT+F for vsc

dktapps:
It won't work without the proper configuration, and this repo doesn't have one for VSC. I'll deal with it later.

src/utils/Config.php
```diff
@@ -396,6 +400,46 @@ public function getNested($key, $default = null){
...
+
+	public function getNestedFloat(string $key, float $default = null) : ?float{
+		if(($value = $this->getNested($key)) !== null && !is_numeric($value)){
+			throw new \TypeError("Expected float, got " . gettype($value));
+		}
+		return $value !== null ? (float) $value : $default;
+	}
+
+	public function getNestedBoolean(string $key, bool $default = null) : ?bool{
+		if(($value = $this->getNested($key)) !== null && !is_bool($value)){
```
dktapps: I think this should also accept the integer values `1` and `0`.

src/utils/Config.php
```diff
@@ -441,6 +485,46 @@ public function set($k, $v = true) : void{
...
+
+	public function getInt(string $key, int $default = null) : ?int{
+		if(($value = $this->get($key, null)) !== null && !is_int($value)){
+			throw new \TypeError("Expected int, got " . gettype($value));
+		}
+		return $value ?? $default;
+	}
+
+	public function getBoolean(string $key, bool $default = null) : ?bool{
+		if(($value = $this->get($key, null)) !== null && !is_bool($value)){
```
dktapps: Same here.

NhanAZ (approved):

src/utils/Config.php
```diff
@@ -441,6 +486,46 @@ public function set($k, $v = true) : void{
 		}
 	}
 
+	public function getString(string $key, string $default = null) : ?string{
+		if(($value = $this->get($key, null)) !== null && !(is_string($value) || is_numeric($value))){
```
SOF3, line 490: I am still hesitant about this. `is_numeric` does not mean the string can be losslessly preserved. `1.00` is still converted incorrectly.

SOF3:
Config is a really obsolete class that we even wanted to remove (#4113). Is there any point in adding more APIs to this class? If you wanted a better config, just bundle your own library.

dktapps:
To paraphrase my previous comments on Discord about this, these functions are not solving any problem. They are simply allowing you to pretend there isn't a fundamental design problem with Config, while doing nothing to solve the actual validation concerns that lead to needing this in the first place.

This is not a fault of the PR, but simply a symptom of the fact that Config was ill-conceived and isn't suitable for any actual use csse.

dktapps:
I'm going to decline this, but don't take it the wrong way - this isn't a problem with the PR's quality, just the problem that it's trying to solve.

PR #5567, opened 2023-02-11
start.sh: Use -n instead of ! -z

zSALLAZAR:

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -51,12 +51,12 @@ if [ "$DO_LOOP" == "yes" ]; then
 		if [ ${LOOPS} -gt 0 ]; then
 			echo "Restarted $LOOPS times"
 		fi
-		"$PHP_BINARY" "$POCKETMINE_FILE" $@
+		"$PHP_BINARY" "$POCKETMINE_FILE" "$@"
 		echo "To escape the loop, press CTRL+C now. Otherwise, wait 5 seconds for the server to restart."
 		echo ""
 		sleep 5
 		((LOOPS++))
 	done
 else
-	exec "$PHP_BINARY" "$POCKETMINE_FILE" $@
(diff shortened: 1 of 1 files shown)


zSALLAZAR:
See commit message for explanation.

SOF3:
There seems to be two irrelevant changes

crischutu07:
Hey, don't forget to double quotes the "$@"
```
$ shellcheck myscript
# You're solved this one
Line 26:
        elif [[ ! -z $(type php 2> /dev/null) ]]; then
                ^-- SC2236 (style): Use -n instead of ! -z.
 
Line 54:
                "$PHP_BINARY" "$POCKETMINE_FILE" $@
                                                 ^-- SC2068 (error): Double quote array expansions to avoid re-splitting elements.
 
Line 61:
        exec "$PHP_BINARY" "$POCKETMINE_FILE" $@
                                              ^-- SC2068 (error): Double quote array expansions to avoid re-splitting elements.

$```

**Comment by dktapps** [comment]
I guess it would be a good idea to have shellcheck in the CI. I don't know why I never thought of that.

File: start.sh [hunk 1]
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 DIR="$(cd -P "$( dirname "${BASH_SOURCE[0]}" )" && pwd)"
-cd "$DIR"
+cd "$DIR" || exit 1
```
dktapps: I'm all for fail-fast, but this seems pointless. The script isn't going to fail to cd into its own directory.

dktapps (commented):
The individual changes look fine, but I'm not a fan of mixing unrelated changes together.

zSALLAZAR:
> The individual changes look fine, but I'm not a fan of mixing unrelated changes together.

Should I split this pull request into two different ones?

dktapps:
If you wouldn't mind.

PR #5574, opened 2023-02-14
Implement MoveActorDeltaPacket

### LINKED ISSUE #5568: Make use of MoveActorDeltaPacket
## Description
This has been in the protocol for ages and offers a fairly easy bandwidth win by reducing outflow of non-cacheable, poorly-compressible data (entity position updates).

This wasn't implemented initially due to technical complications with the packet design, but those have long since been resolved, and it should be pretty easy to implement this now.

AkmalFairuz:
## Introduction
Implement MoveActorDeltaPacket

### Relevant issues
Fix https://github.com/pmmp/PocketMine-MP/issues/5568


src/entity/Entity.php
```diff
@@ -785,18 +789,50 @@ protected function broadcastMovement(bool $teleport = false) : void{
...
+				if(($this->location->yaw - $this->lastBroadcastedMovement->yaw) !== 0.0){
+					$pk->yaw = $this->location->yaw;
+					$pk->flags |= MoveActorDeltaPacket::FLAG_HAS_YAW;
+					$pk->headYaw = $this->location->yaw;
+					$pk->flags |= MoveActorDeltaPacket::FLAG_HAS_HEAD_YAW;
+				}
+				$this->server->broadcastPackets($this->hasSpawned, [$pk]);
+			}
 		}
+		$this->lastBroadcastedMovement = Location::fromObject($offsetPosition, null, $this->location->yaw, $this->location->pitch);
```
dktapps: We already have `lastLocation` for this.

dktapps (changes requested):
The position used in sendSpawnPacket needs to be changed to use the last broadcast position as well, or the initial delta may accumulate errors.

src/entity/Entity.php
```diff
@@ -1477,7 +1495,7 @@ protected function sendSpawnPacket(Player $player) : void{
 			$this->getId(), //TODO: actor unique ID
 			$this->getId(),
 			static::getNetworkTypeId(),
-			$this->location->asVector3(),
+			$this->lastLocation->asVector3(),
```
dktapps: This needs to be changed in subclasses as well.

dktapps:
It occurs to me that it may not be possible to ensure that plugin devs use the correct position in custom sendSpawnPacket implementations. If an incorrect position is used, the delta position will accumulate errors and the source of the problem will be far from obvious.

Any suggestions how to combat this?

src/entity/Entity.php
```diff
@@ -785,17 +786,34 @@ protected function broadcastMovement(bool $teleport = false) : void{
...
-				$this->getOffsetPosition($this->location),
-				$this->location->pitch,
-				$this->location->yaw,
-				$this->location->yaw,
-				(
-					//TODO: if the above hack for #4394 gets removed, we should be setting FLAG_TELEPORT here
-					($this->onGround ? MoveActorAbsolutePacket::FLAG_GROUND : 0)
-				)
-			)]);
+			$deltaLoc = $offsetPosition->subtractVector($this->getOffsetPosition($this->lastLocation));
```
dktapps: This calculation is useless. With the current code, you can directly compare lastLocation->whatever with currentLocation->whatever.

src/entity/Entity.php
```diff
@@ -785,17 +786,34 @@ protected function broadcastMovement(bool $teleport = false) : void{
...
+			}
+			if($deltaLoc->y != 0.0){
+				$pk->yPos = $offsetPosition->y;
+				$pk->flags |= MoveActorDeltaPacket::FLAG_HAS_Y;
+			}
+			if($deltaLoc->z !== 0.0){
+				$pk->zPos = $offsetPosition->z;
+				$pk->flags |= MoveActorDeltaPacket::FLAG_HAS_Z;
+			}
+			if(($this->location->pitch - $this->lastLocation->pitch) !== 0.0){
```
dktapps: This subtraction is pointless. It's not telling you anything that an equality operator wouldn't.

src/entity/Entity.php
```diff
@@ -785,17 +786,34 @@ protected function broadcastMovement(bool $teleport = false) : void{
...
+				$pk->pitch = $this->location->pitch;
+				$pk->flags |= MoveActorDeltaPacket::FLAG_HAS_PITCH;
+			}
+			if(($this->location->yaw - $this->lastLocation->yaw) !== 0.0){
```
dktapps: Same here.

dktapps (changes requested):

src/entity/Entity.php
```diff
@@ -1477,7 +1495,7 @@ protected function sendSpawnPacket(Player $player) : void{
 			$this->getId(), //TODO: actor unique ID
 			$this->getId(),
 			static::getNetworkTypeId(),
-			$this->location->asVector3(),
+			$this->lastLocation->asVector3(),
 			$this->getMotion(),
 			$this->location->pitch,
 			$this->location->yaw,
```
dktapps: These also need to use `lastLocation`.

AkmalFairuz:
> It occurs to me that it may not be possible to ensure that plugin devs use the correct position in custom sendSpawnPacket implementations. If an incorrect position is used, the delta position will accumulate errors and the source of the problem will be far from obvious.
> 
> Any suggestions how to combat this?

Store players who need an absolute move packet when spawned to players, then send absolute move packet when movement packet is broadcasted. See my commit

dktapps:
You could probably reduce it to a boolean flag. Not much sense in adding the extra memory usage and complexity. It anyway might be useful to send periodic absolute updates anyway.

dktapps:
I meant a single boolean, not an array of booleans.

src/entity/Entity.php
```diff
@@ -220,6 +221,9 @@ public static function nextRuntimeId() : int{
 	/** @var int|null */
 	protected $targetId = null;
 
+	/** @var bool */
+	protected $needsAbsoluteMovement = true;
```
dktapps, line 225: Use native types for this. Also, this should be `private`.

src/entity/Entity.php
```diff
@@ -785,17 +790,48 @@ protected function broadcastMovement(bool $teleport = false) : void{
 				$this->spawnTo($player);
```
dktapps, line 790: This is no longer going to work correctly, since it's now using the lastLocation.

PR #5581, opened 2023-02-17
SurvivalBlockBreakHandler: Take swimming, flying, climbing and effects into account

zSALLAZAR:


src/player/SurvivalBlockBreakHandler.php
```diff
@@ -65,8 +67,27 @@ private function calculateBreakProgressPerTick() : float{
...
+		if(!$this->player->onGround){
+			$breakTimePerSecond *= 5;
+		}
+		if($this->player->canClimb() && $this->player->getWorld()->getBlock($this->player->getPosition())->canClimb()){
+			$breakTimePerSecond *= 5;
+		}
+
+		$effects = $this->player->getEffects();
+		if(($miningFatigue = $effects->get(VanillaEffects::MINING_FATIGUE())) !== null){
+			$breakTimePerSecond *= $breakTimePerSecond / ($breakTimePerSecond * 0.2 ** $miningFatigue->getEffectLevel());
```
dktapps: How confident are you in this? Last time I checked, the values were different for Bedrock, which is why I didn't implement this properly to begin with.

src/player/SurvivalBlockBreakHandler.php
```diff
@@ -65,8 +67,27 @@ private function calculateBreakProgressPerTick() : float{
 		if(!$this->block->getBreakInfo()->isBreakable()){
 			return 0.0;
 		}
-		//TODO: improve this to take stuff like swimming, ladders, enchanted tools into account, fix wrong tool break time calculations for bad tools (pmmp/PocketMine-MP#211)
-		$breakTimePerTick = $this->block->getBreakInfo()->getBreakTime($this->player->getInventory()->getItemInHand()) * 20;
+		$breakTimePerSecond = $this->block->getBreakInfo()->getBreakTime($this->player->getInventory()->getItemInHand());
```
dktapps: This should really be `breakTimeSeconds`, "per second" doesn't make sense. (Don't ask me why the original variable was named that way ...)

src/player/SurvivalBlockBreakHandler.php
```diff
@@ -65,8 +67,27 @@ private function calculateBreakProgressPerTick() : float{
...
+		if($this->player->canClimb() && $this->player->getWorld()->getBlock($this->player->getPosition())->canClimb()){
+			$breakTimePerSecond *= 5;
+		}
+
+		$effects = $this->player->getEffects();
+		if(($miningFatigue = $effects->get(VanillaEffects::MINING_FATIGUE())) !== null){
+			$breakTimePerSecond *= $breakTimePerSecond / ($breakTimePerSecond * 0.2 ** $miningFatigue->getEffectLevel());
+		}
+		if(($haste = $effects->get(VanillaEffects::HASTE())) !== null){
+			$breakTimePerSecond -= $breakTimePerSecond * 0.2 * min($haste->getEffectLevel(), 5);
```
dktapps: Same again here - I'm pretty sure the values differ from Java.

src/player/SurvivalBlockBreakHandler.php
SenseiTarzan:
```php
$breakTimePerTick = ceil($this->block->getBreakInfo()->getBreakTime($this->player->getInventory()->getItemInHand()) * 20);
       
        $breakTimePerTick *= 1 - (0.3 * $this->player->getEffects()->get(VanillaEffects::MINING_FATIGUE())?->getAmplifier() ?? 0);
```

src/player/SurvivalBlockBreakHandler.php
SenseiTarzan:
```php
$breakTimePerTick = ceil($this->block->getBreakInfo()->getBreakTime($this->player->getInventory()->getItemInHand()) * 20);
        $breakTimePerTick *= 1 - (0.2 * $this->player->getEffects()->get(VanillaEffects::HASTE())?->getAmplifier() ?? 0);
```

SenseiTarzan:
The code I gave is correct I used the values ​​of mcbe but also I tested and it did not create a problem and detected the correct time

src/player/SurvivalBlockBreakHandler.php
ShockedPlot7560: I don't know where is your code from but according to the wiki:
```
The mining speed is reduced to 0.30^level of the normal mining speed. For the in-game default level III, the speed is reduced to 0.33 (0.027×) of normal speed, taking 37 times longer than usual to break a block with the proper tool for that block. 
```
https://minecraft.fandom.com/wiki/Mining_Fatigue

src/player/SurvivalBlockBreakHandler.php
ShockedPlot7560: Why maximize the haste level to 5? This is normally an effect that can theoretically have a level of up to 255. The only limit is haste 3 for arm animation.

dktapps: Probably because the break time can become negative if the haste level is higher than 5

src/player/SurvivalBlockBreakHandler.php
zSALLAZAR: The values from the wiki are wrong. The mining animation was too fast and already finished although the block was not broken yet.

src/player/SurvivalBlockBreakHandler.php
zSALLAZAR: > Probably because the break time can become negative if the haste level is higher than 5

You're right. But I used the values from the wiki and it seems that they are wrong too. The animation is too fast.

zSALLAZAR:
The calculation for the mining fatigue effect is very close to vanilla, but still wrong.
And the calculation for the Haste effect was copied from the wiki, but is also wrong.
Unfortunately I don't know the correct calculations for the effects, so I would appreciate any help.

SenseiTarzan:
> The calculation for the mining fatigue effect is very close to vanilla, but still wrong.
> And the calculation for the Haste effect was copied from the wiki, but is also wrong.
> Unfortunately I don't know the correct calculations for the effects, so I would appreciate any help.

https://github.com/AID-LEARNING/SymplyPlugin/blob/master/src/SenseiTarzan/SymplyPlugin/Utils/BlockUtils.php
Here is the official MCBE code

dktapps:
Closing in favour of #6500

Issue #2540, opened 2018-11-30
Allow chest open/closed state to be changed arbitrarily


dktapps:
This could be useful for plugins in decorative environments. Currently it's necessary to add a viewer to the chest's inventory, which seems unnecessarily overcomplicated.

This proposes adding some simple `Chest->open()` and `Chest->close()` methods, or something similar.

dktapps:
Quoting from #5583:

> What about this: `?bool $lockedOpenState = null`, where `null` means "default", `false` means "stays closed", and `true` means "stays open" (or possibly an enum for the same).
>
> That being said, it's not so simple as this - you'd also need to address #2503 to have cosmetically open chests continue to display as open when new players come in range. (#2503 is currently a minor annoyance, but this feature would get the blame for it by users.)

Issue #2503, opened 2018-10-30
Teleporting to an open chest does not make the chest show as open


dktapps:
### Issue description
If a player out of your render range opens a chest and you teleport to them, they'll see the chest as open while you'll see it as closed.

### Steps to reproduce the issue

1. have player1 open a chest
2. have player2 teleport to player1
3. notice chest appears closed on player2's client

dktapps:
This requires several changes, including having chunk caches contain multiple packets.

PR #5583, opened 2023-02-19
Allow animated blocks with inventories to change their open/closed state arbitrarily

zSALLAZAR:
### Relevant issues

#2540

### API changes
Added `AnimatedBlockInventoryTrait->isClosed() : bool`
Added `AnimatedBlockInventoryTrait->open() : void`
Added `AnimatedBlockInventoryTrait->close() : void`

## Tests

https://user-images.githubusercontent.com/59490940/219906422-dab44cec-c281-4b6f-91b9-3990150798cb.mp4


dktapps:
Relates to #2503

dktapps:
Seems to me like this can be overridden quite easily by a player. A chest which is set in the "open" state with no viewers simply needs to be opened and closed by a player, and the chest's lid will close, despite having been set in the "open" state by a plugin.

zSALLAZAR:
Adding an event for opening/closing animated blocks and canceling the event might solve this?
For example, if such event is canceled and a player opens and then closes a chest that is in the open state, the chest remains in the open state.

dktapps:
What about this: `?bool $lockedOpenState = null`, where `null` means "default", `false` means "stays closed", and `true` means "stays open" (or possibly an enum for the same).

That being said, it's not so simple as this - you'd also need to address #2503 to have cosmetically open chests continue to display as open when new players come in range. (#2503 is currently a minor annoyance, but this feature would get the blame for it by users.)

I know you're probably not gonna reappear after 18 months and implement this, but I'll write it here for posterity.

dktapps:
Closing due to lack of activity

PR #5587, opened 2023-02-21
Timings tree

dktapps:
## Introduction
Currently, it's often difficult to identify problems in a timings report, since it's far from obvious which timers include which other timers.

## Changes
This PR:
- uses unique timings records for different call trees, so different parent timers will result in distinct timings records (e.g. `Command - status` will appear twice, once for the times it was called by the console, and a second time for times it was called as a result of a player using it)
- clearly identifies in timings reports which timers are counted by which other timers, allowing them to be potentially displayed in a tree breakdown

### Behavioural changes
Timings may now have a slightly larger performance impact when enabled, but I think the trade-off is more than worth it.

## Backwards compatibility
This is fully backwards compatible with timings.pmmp.io, which already aggregates timings from multiple timers with the same names.

No public non-internal APIs were changed by this PR.

## Unresolved research
It's not currently clear how "parent" timers (which are really timings groups or aggregators) will play into this, but there shouldn't be any issues.

## Follow-up
Implement a tree view for timings.pmmp.io.

## Tests
An example timings report can be seen here: https://timings.pmmp.io/?id=300261
and its raw version here: https://timings.pmmp.io/?id=300261&raw=1

You can see that `Command - status` appears multiple times in the raw view, but only once in the result.

**Note:** timings.pmmp.io does not _yet_ support displaying this information in a tree view, so it's currently expected that the timings report will look exactly the same as any other.

### DIFF
--- a/src/timings/TimingsHandler.php
+++ b/src/timings/TimingsHandler.php
@@ -56,8 +56,9 @@ class TimingsHandler{
 				"Count: $count",
 				"Avg: $avg",
 				"Violations: " . $timings->getViolations(),
-				"Id: " . $timings->getId(),
-				"ParentId: " . ($timings->getParentId() ?? "none")
+				"RecordId: " . $timings->getId(),
+				"ParentRecordId: " . ($timings->getParentId() ?? "none"),
+				"TimerId: " . $timings->getTimerId()
 			]);
 		}
 		$result = [];
--- a/src/timings/TimingsHandler.php
+++ b/src/timings/TimingsHandler.php
@@ -56,8 +56,8 @@ class TimingsHandler{
 				"Count: $count",
 				"Avg: $avg",
 				"Violations: " . $timings->getViolations(),
-				"Id: " . $timings->getTimerId(),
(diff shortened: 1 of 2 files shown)


SOF3 (approved):

src/timings/TimingsHandler.php
```diff
@@ -115,13 +131,24 @@ public function startTiming() : void{
...
-			if($this->record === null){
-				$this->record = new TimingsRecord($this);
-			}
-			$this->record->startTiming($now);
 			if($this->parent !== null){
 				$this->parent->internalStartTiming($now);
 			}
+
+			$current = TimingsRecord::getCurrentRecord();
+			if($current !== null){
```
dktapps, line 153: This extra check can be avoided with a global root `TimingsRecord`

dktapps:
An example of this new style of timings report can be seen here: https://timings.pmmp.io/?id=302629

The old list view is also retained, but I don't think there will be much need for it with the tree view.

Issue #5595, opened 2023-02-26
Add Player parameter in BlockGrowEvent


DebHugo:
## Description
Add in BlockGrowEvent an optional parameter that accept Player or null, default is null.

## Justification
If you want to prevent player to grow crops is not possible. So with this parameter, when player want to grow crops with bone meal we can autorise or not. If crops grow naturally set third parameter to null.

## Alternative methods
I know none alternative method

dktapps:
Yes, we do already have the same for StructureGrowEvent, so I suppose it makes sense.

dktapps:
Implemented by #5595

PR #5596, opened 2023-02-26
Add Player parameter in BlockGrowEvent

DebHugo:
## Introduction

#5595 

### API changes

Added an optional that accept Player or null, default is null and add a function getPlayer()


dktapps:
BlockGrowEvent is called in many more places, not just crops.

The build also needs to pass.

DebHugo:
I don't understand why there is an error in the code style check.
I'll let you fix it yourself.

dktapps:
It's because there's trailing whitespace on those lines. You can download and run [php-cs-fixer](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/releases) to fix the issues automatically.

DebHugo:
I finished. It is good for you ?

dktapps (approved):

PR #5636, opened 2023-03-14
First look at implementing ItemStackRequest

### LINKED ISSUE #3899: Implement the new transaction system
## Description
In 1.16 or so, Mojang reinvented the wheel once again and made a new transaction system for inventories.
I refused to implement this on the basis that it didn't really seem to offer any benefit over the original system.

I eventually planned to implement it since there's no doubt in my mind that the legacy system will spontaneously break at some point when nobody is expecting it.
However, it's since come to my attention that there are various reasons why it's desirable.

- Instead of sending itemstacks by-value on the wire, itemstacks are referred to by server-assigned itemstack IDs, which means that the client cannot send NBT to the server anymore. This is a big positive for bandwidth and exploitability.
- Enchanting is easily implemented with the new system, just like the good old days.
- Crafting is no longer required to be a godawful hacky mess.
- More detail is provided about how an itemstack was interacted with (e.g. move, split, drop) instead of just "old item, new item" like the old system.

While the new system has its flaws, it solves some significant issues with the current system, so it should be considered more of a need than a nice-to-have.

### LINKED ISSUE #3954: Too many packet in single Batch
Issue description

When I a craft a lot of Block at once from minerals (Lapis, diamond...).

1. Give a lot of mineral
2. Open Crfating table
3. Craft a lot of Block at once fromp the minerals (with shift-click on computer or crafting menu on mobile.

### OS and versions
PocketMine-MP: 3.16.0
PHP: 7.3
Server OS:Debian 9
Game version: Win10 1.16.101

### Plugins
- No plugins used (expect DevTools)

Capture

### More 
Use default pocketmine.yml

### LINKED ISSUE #5179: Smithing Mismatch transaction
### Issue description

- Expected result: Smithing to give netherite helmet
- Actual result: Gives "Mismatch transaction received"

1. Get Diamond Helmet and 1 netherite ingot(issue seems to only happen with 1 ingot)
2. Go into survival mode
3. place a smithing table and try upgrading it

### OS and versions

* PocketMine-MP: 5.0.0-ALPHA2
* PHP: 8.0
* Server OS: Windows
* Game version: Win10

### Plugins
0 plugins

- If you remove all plugins, does the issue still occur?: Yes issue happens without plugins, i installed 0 plugins.

Not a crash

### LINKED ISSUE #5619: Cut Copper Stairs cannot be selected in inventory (PM 5.0.0)
### Issue description

- Expected result: When selecting the stairs, you can drag them to the inventory and use them
- Actual result: Selecting them removes them from the mouse

1. Grab the stairs
2. See how it disappears the second you grab them

https://user-images.githubusercontent.com/127597296/224480140-b3e75ca7-fe55-4a44-b1f3-32df00001a7c.mp4

### OS and versions

* PocketMine-MP: 5.0.0-BETA1
* PHP: 8.0.28
* Using JIT: not supported
* Server OS: linux
* Game version: Win10

### Plugins

No plugins

dktapps:
## Introduction
This is a rough implementation of #3899.

The ItemStackRequest system replaces almost all usages of the old inventory transaction system.

Legacy inventory transactions are notably still used for dropping items when the inventory window is closed, equipping armour, and other stuff that hasn't been migrated yet. However, even in its incomplete state, it is vastly superior to the legacy system:

- Reduced bandwidth usage. The ItemStackRequest system is much more bandwidth efficient, particularly when items with large NBT are involved, such as written books or shulker boxes.
- Better performance. Since we don't have to decode serialized itemstacks from the client anymore, inventory transaction processing can be done much more efficiently.
- Reduced attack surface for hackers. The ItemStackRequest system does not require deserializing serialized items sent by the client, meaning that the deserializer pathways for items and their NBT is no longer accessible to clients (note the eliminated use of `TypeConverter::netItemStackToCore()`.
- No more transaction failures because of encoding weirdness or the client modifying its copy of the item (e.g. https://github.com/pmmp/NBT/commit/7de6bdfdfbcafe2689aafbde47f6506f265afd19)

### Relevant issues
- Closes #3899 
- Fixes #3954
- Fixes #5179 (or it did, before smithing changed again in 1.19.60 anyway)
- Fixes #5619 

### API changes
- The following API methods have been added:
  - `CraftingManager->getCraftingRecipeIndex() : array<int, CraftingRecipe>`
  - `CraftingTransaction` now accepts two additional optional parameters: `?CraftingRecipe $recipe = null, ?int $repetitions = null`

## Backwards compatibility
This PR makes significant changes to the inventory handling in the `network/mcpe` package. While these packages are considered internal, they may nonetheless break plugins.

## Follow-up
Tighten validation constraints to limit attack surface as much as possible. This system is relatively easy to constrain and validate the behaviour of.

## Tests
This has been playtested quite thoroughly and all supported PM inventory and crafting features were found working. However, more testing and review is needed.


dktapps:
There's also some lack of clarity around transaction validation - there's currently quite a lot of pre-validation in the transaction builder throwing PacketHandlingExceptions, which differs from the old behaviour of just quietly rejecting and rolling back the transaction on failure to validate.

dktapps:
The following edge cases have now been tested:
- Removing windows during InventoryTransactionEvent
- Setting contents of unrelated slots during InventoryTransactionEvent

The following problems remain:
- Simulating client mis-predictions results in broken behaviour in the client inventories. All slots affected by the request must receive a corresponding ResponseSlotInfo, but if we send one for a mispredicted slot, it can result in all kinds of peculiar behaviour. In some cases we may need to defer sending InventorySlotPackets until after ItemStackResponse has been sent.

dktapps:
In addition, we may be able to avoid sending some slot updates back to the client, since rejecting a request will cause the client to rollback all of the stuff it did in the transaction.

Sanjunaik93:
How to add recipe can any one help me?

PR #5640, opened 2023-03-17
replacing new Vector3(0, 0, 0) with Vector3::zero()

SmallkingDev:
## Introduction
Just giving method zero existing in Vector3.

## Changes
Replaced new Vector(0, 0, 0) with Vector::zero()


dktapps:
This should be targeted at `minor-next`

Issue #5683, opened 2023-04-06
Ways to get cause for EntityCombustEvent


alvin0319:
## Description
Add a new EntityIgniteEvent called when an entity is being ignited by something, e.g by the player or TNT.

## Justification
With the current API, there is no way to know who/which player/thing caused the entity to be ignited.

The following class added: `EntityIgniteEvent`

`EntityIgniteEvent->getCause() : ?Entity`: Returns an entity that caused this entity to be ignited.

#### Thing to consider

Make this event `Cancellable`:
* After this event, events such as `EntityExplodeEvent/ExplosionPrimeEvent` will be called anyway, so I don't think this event need to be cancellable, instead, we could just cancel exploding in `EntityExplodeEvent/ExplosionPrimeEvent`

## Alternative methods
N/A

dktapps:
> After this event, events such as EntityExplodeEvent/ExplosionPrimeEvent will be called anyway, so I don't think this event need to be cancellable

This depends on the behaviour of the entity when ignited.

We could also use EntityCombustEvent for the same purpose.

PR #5688, opened 2023-04-10
Implement EntityIgniteEvent

alvin0319:
## Introduction

This PR adds a new event as discussed in #5683.

### Relevant issues

* Addresses #5683 

### API changes

Added `EntityIgniteEvent`:
* `EntityIgniteEvent->getCause()`: Returns the entity that caused the entity to ignite.

### Behavioural changes

Entities such as TNT will not be ignited if the `EntityIgniteEvent` was cancelled.

## Backwards compatibility

N/A

## Follow-up

N/A

### DIFF
--- a/src/block/TNT.php
+++ b/src/block/TNT.php
@@ -80,7 +80,7 @@ class TNT extends Opaque{
 
 	public function onBreak(Item $item, ?Player $player = null) : bool{
 		if($this->unstable){
-			$this->ignite();
+			$this->ignite(cause: $player);
 			return true;
 		}
 		return parent::onBreak($item, $player);
@@ -91,7 +91,7 @@ class TNT extends Opaque{
 			if($item instanceof Durable){
 				$item->applyDamage(1);
 			}
-			$this->ignite();
+			$this->ignite(cause: $player);
 			return true;
 		}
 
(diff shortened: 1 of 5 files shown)


dktapps:
TNT might ignite because of fire, so I don't think limiting this to entities alone is the correct approach.

IvanCraft623:
I think that this event could be also used for `Entity::setOnFire()`

dktapps:
> I think that this event could be also used for `Entity::setOnFire()`

We have EntityCombustEvent for that.

dktapps:
I'm still not convinced we can't simply use EntityCombustEvent for this. Seems to me like having both EntityCombustEvent and EntityIgniteEvent will be confusing for plugin developers. Thoughts?

alvin0319:
~~In the current PM usage, `EntityCombustEvent` is called "after" the start of combustion. However, the `EntityIgniteEvent` in this PR is called before combustion starts, so I think the two are clearly different. (although the description of `EntityCombustEvent` is currently unclear, I think this should be addressed in another commit)~~ nvm, misread the code

alvin0319:
Oops, I accidentally deleted the repo while cleaning up forks. but considering the class EntityCombustEvent, I think we can make use of it instead of creating a dedicated class for these cases. I will create another PR that utilizes EntityCombustEvent.

PR #5689, opened 2023-04-10
Ticking chunks rewrite

### LINKED ISSUE #5673: Avoid recalculating ticking chunks every tick
## Description
We can reduce the overhead of chunk selection by caching the selected chunks, and only recalculating the selected area periodically, for example every 20 ticks. This would reduce the performance cost by 95%, while having almost no visible effect - chunks not in the player's current ticking radius might continue to be ticked for 1 second after the player moves away from the area, but I don't think anyone will notice this or care.

I previously attempted to implement this here: aebcfc516ff97492c58ee478a3944915a777a75e
However, I ended up reverting it soon after, because there were several edge cases I forgot to consider.

Caching the ticked chunk list adds a new state to the system, where chunks which were previously valid for ticking are no longer valid. This can happen for several reasons:

- The chunk, or one or more of its 8 surrounding chunks were unloaded
- The chunk, or one or more of its 8 surrounding chunks was locked, e.g. for world population
- The chunk, or one or more of its 8 surrounding chunks had its light cache invalidated, e.g. due to external modification
- The chunk, or one or more of its 8 surrounding chunks was replaced by setChunk() with a non-populated chunk

All of these conditions would need to be accounted for at the points when they occurred.

## Justification
`Random Chunk Ticking - Chunk Selection` timers often show up as a heavy hitter on performance profiles from factions servers and similar. This is because of sparse player distribution, resulting in many more chunks needing to be ticked.

## Alternative methods
I have not given much thought to alternative ways of achieving this goal. It's not very clear what exactly the hard-hitting logic for this is, so further investigation and profiling may be warranted.

dktapps:
## Introduction
Ticking chunks have been a pain point for a long time, due to inefficient (performance intensive) chunk selection process, and inability for plugins to define precise ticking areas, preventing the implementation of ticking areas like those seen in vanilla Bedrock.

This PR introduces new APIs for marking chunks for ticking, and also allows the system to remember whether a chunk is available for ticking, which significantly reduces the amount of work done by the server. In addition, due to the new chunk ticker implementation, the system will never check the same chunk twice, which also improves performance.

### Relevant issues
Fixes #5673

## Changes
- `ChunkSelector::selectChunks()` now yields the distance in chunks from the given center point as the key, instead of an incrementing index.
- `PlayerChunkLoader` is deprecated
- `TickingChunkLoader` is deprecated
- Added `ChunkTicker` class
- Added APIs `World::registerTickingChunk()` and `World::unregisterTickingChunk()`

### Behavioural changes
I'm expecting this to provide a noticeable performance improvement for factions servers, and other similar game modes where players are scattered over a large map.

The legacy method of ticking chunks is still supported, but is not used in the core code, and has no performance impact if not used.

## Backwards compatibility
This should be fully backwards compatible.

## Follow-up
Simplification?

## Tests
This has not yet been tested.


dktapps:
A simulation with a 24 chunk ticking radius (totalling 1884 chunks ticked) across 3 minutes with 1 player shows a 25x performance improvement in chunk selection (average 3.54 ms/tick -> 0.142 ms/tick):

Without: https://timings.pmmp.io/?id=303594
With: https://timings.pmmp.io/?id=303588

Due to some anomaly (probably shifting from P to E cores), the version with the PR shows a higher load in actual chunk ticking, despite no logic change and the same number of chunks being ticked. However, the improvement to selection performance is abundantly clear.

dktapps:
I've now fully tested this and am satisfied that it works correctly.

Situations tested:
- Performance changes
- Moving around the map
- Changing render distance (if smaller than the ticking radius, it should take precedence)
- Manually registering `ChunkTickers` to create ticking areas

PR #5694, opened 2023-04-12
CreativeInventory per Player

BrandPVP:
### API changes

Player->getCreativeInventory()
Player->setCreativeInventory(CreativeInventory)

## Tests
Just simply jump in the game and it switches your creative inventory
```php
<?php

use pocketmine\event\player\PlayerJumpEvent;
use pocketmine\inventory\CreativeInventory;
use pocketmine\item\Durable;

/**
 * @name test
 * @version 1.0.0
 * @api 4.0.0
 * @main test
 */
class test extends PluginBase implements Listener{

	private CreativeInventory $customCreative;

		// remove apples from the default creative inventory
		CreativeInventory::getInstance()->remove(VanillaItems::APPLE());

		// remove all durables from a custom creative inventory
		$this->customCreative = clone CreativeInventory::getInstance();
		foreach($this->customCreative->getAll() as $item) {
			if($item instanceof Durable) {
				$this->customCreative->remove($item);
			}
		}
	}

	public function onJump(PlayerJumpEvent $event) : void {

		$serverDefaultInventory = CreativeInventory::getInstance();
		$customInventory = $this->customCreative;

		if($player->getCreativeInventory() === $serverDefaultInventory) {
			$player->setCreativeInventory($customInventory);
		} else {
			$player->setCreativeInventory($serverDefaultInventory);
		}
	}
}
```


src/player/Player.php
```diff
@@ -2505,6 +2509,17 @@ public function getCraftingGrid() : CraftingGrid{
...
 	}
 
+	public function getCreativeInventory() : CreativeInventory{
+		return $this->creativeInventory;
+	}
+
+	public function setCreativeInventory(CreativeInventory $inventory) : void{
+		$this->creativeInventory = $inventory;
+		if($this->isConnected()){
+			$this->getNetworkSession()->getInvManager()?->syncCreative();
```
dktapps, line 2553: This might not be appropriate. If it's called during the login sequence it could lead to a client crash.

src/player/Player.php
```diff
@@ -2526,6 +2530,24 @@ public function getCraftingGrid() : CraftingGrid{
 		return $this->craftingGrid;
 	}
 
+	/**
+	 * Returns the default CreativeInventory instance if no custom one was set.
```
```suggestion
	 * Returns the creative inventory shown to the player. Unless changed by a plugin, this is usually the same for all players.
```

src/player/Player.php
```diff
@@ -2526,6 +2530,24 @@ public function getCraftingGrid() : CraftingGrid{
...
+	 */
+	public function getCreativeInventory() : CreativeInventory{
+		return $this->creativeInventory;
+	}
+
+	/**
+	 * To set a custom creative inventory, you need to make a clone of a CreativeInventory instance.
+	 * Make sure to use this method only in or after `PlayerJoinEvent` to sync the contents on the client.
```
dktapps: This doesn't seem appropriate. It will work in PlayerLoginEvent also (since that takes place before the network session sends the creative inventory initially.

dktapps (changes requested):
I'm not super happy with the documentation for these APIs.

src/player/Player.php
```diff
@@ -2526,6 +2530,25 @@ public function getCraftingGrid() : CraftingGrid{
...
+	 * Returns the creative inventory shown to the player.
+	 * Unless changed by a plugin, this is usually the same for all players.
+	 */
+	public function getCreativeInventory() : CreativeInventory{
+		return $this->creativeInventory;
+	}
+
+	/**
+	 * To set a custom creative inventory, you need to make a clone of a CreativeInventory instance.
+	 * Make sure to use this method only in or after `PlayerLoginEvent` to sync the contents on the client.
```
dktapps: This line is redundant. There's no event before PlayerLoginEvent where this could be used anyway.

dktapps (approved):
It seems like this also sets the stage for getting rid of the CreativeInventory singleton, which is a bonus.

PR #5703, opened 2023-04-16
Caching creative inventory entries

ShockedPlot7560:
## Introduction

I could notice during a debugging that the entries of the packet that is sent to the client to synchronize the creative menu, was recalculated each time.  

This is quite surprising because it is not linked to the player and therefore can only change if you add/remove items from the creative menu.

This difference is much bigger with an important number of items more than pocketmine base. 950 base items against 1700 for my case.

Implementing this caching system allows to reduce the sending of the creative inventory to just what is necessary: the encoding / network compression.

### API changes

* Added ``CreativeInventoryCache``

## Tests

* Launch server
* Active timings
* Connect player
* I did my test by alternating a change of game mode from survival to creative, creative survival, 4 times (4 commands)
* Recover your timings

You should not see a significant difference with the basic pocketmine data.


dktapps:
Seems to me like we need TypeConverter timings.

ShockedPlot7560:
It is that with or without, the systematic calculation is not necessary.

Do you mean that I have to add the timings in the pull request, or that I have to provide other things?

dktapps:
No, you don't need to worry about that. It's just an afterthought I had when looking at the PR.

I'm not a huge fan of caching the entries this way since it'll add to the server idle memory footprint. I know we already do the same thing for crafting, but I'm not a fan of it there either.

Perhaps it warrants some investigation into why this is so slow.

ShockedPlot7560:
I'm not a big fan either, but considering what the server does for this, the problem is not the conversion of items. But it comes rather from the numerous iterations to do, one for all the items to clone them and another one that comes to convert them.

dktapps:
> I'm not a big fan either, but considering what the server does for this, the problem is not the conversion of items. But it comes rather from the numerous iterations to do, one for all the items to clone them and another one that comes to convert them.

I hadn't considered that cloning items in getAll() might have performance implications for this. I don't suppose you've measured the impact of this?

From what I recall, the main issue with cloning items is that their NBT trees have to be recursively cloned, which is quite costly.

ShockedPlot7560:
I didn't measure the impact, but getting the list of creative items requires a recursive clone of all the items

ShockedPlot7560:
But I can try to measure the impact from my side and give the results here. know where the problem really comes from

dktapps:
I've just done a test run, about 40% of the overhead actually comes from Item::serializeCompoundTag(). We might be able to optimise this somewhat, but it's not likely it will improve much.

The cloning is a tiny fraction (less than 1%).

ShockedPlot7560:
It is always possible to optimize on the creative side but this masks another problem which is the serialization

src/inventory/CreativeInventoryCache.php
```diff
@@ -0,0 +1,64 @@
...
+use pocketmine\network\mcpe\convert\TypeConverter;
+use pocketmine\network\mcpe\protocol\types\inventory\CreativeContentEntry;
+use pocketmine\utils\SingletonTrait;
+
+final class CreativeInventoryCache{
+	use SingletonTrait;
+
+	/** @var CreativeContentEntry[] $entries */
+	private array $entries = [];
+	private bool $isHit = false;
```
dktapps: You can do this by making `$entries` nullable instead of having an extra field.

dktapps (changes requested):
I think this should be done in a similar manner to `CraftingDataCache`. It's not guaranteed that there will only be a singular `CreativeInventory` instance (see #5694), nor that the cache will live forever.

dktapps:
Concept ACK on this. We're already doing the same thing for crafting, so I don't think it's a problem to add a cache for this as well (the footprint should be microscopic in comparison).

ShockedPlot7560:
I just modified the system to support multiple instances of the creative inventory

src/inventory/cache/CreativeInventoryCache.php
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @author PocketMine Team
+ * @link http://www.pocketmine.net/
+ *
+ *
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\inventory\cache;
```
dktapps: This class belongs in `pocketmine\network\mcpe\cache`

src/inventory/cache/CreativeInventoryCache.php
```diff
@@ -0,0 +1,69 @@
...
+final class CreativeInventoryCache{
+	use SingletonTrait;
+
+	/**
+	 * @var CreativeContentPacket[]
+	 * @phpstan-var array<int, CreativeContentPacket>
+	 */
+	private array $caches = [];
+
+	public function getCache(CreativeInventory $inventory) : CreativeContentPacket {
```
dktapps:
```suggestion
	public function getCache(CreativeInventory $inventory) : CreativeContentPacket{
```

src/inventory/CreativeInventory.php
```diff
@@ -94,10 +104,22 @@ public function remove(Item $item) : void{
...
 	public function contains(Item $item) : bool{
 		return $this->getItemIndex($item) !== -1;
 	}
+
+	/** @phpstan-return ObjectSet<\Closure() : void> */
+	public function getContentChangedCallbacks() : ObjectSet{
+		return $this->contentChangedCallbacks;
+	}
+
+	private function onContentChange() : void {
```
dktapps:
```suggestion
	private function onContentChange() : void{
```

dktapps (changes requested):

dktapps (approved):

PR #5707, opened 2023-04-17
Avoid unnecessary event-related work in hot paths when the events have no registered handlers

dktapps:
## Introduction
this particular optimisation became possible thanks to changes in 4.19.

I observed that the allocation of `Event` objects and calling `->call()` was costing us a significant percentage of the time taken in `PlayerAuthInputPacket` handlers. This change produces a measurable 2 microsecond reduction in overhead for `PlayerAuthInputPacket` handling when players are not moving (10.7 -> 8.7 microseconds on my machine). On a server with 200 players, this translates into a 1% reduction in total CPU load. It will also benefit other packets, but not to the extent that PlayerAuthInputPacket benefits.

### API changes
Added static method `Event::hasHandlers()`

### Behavioural changes
Minor performance improvement to packet handling.

## Backwards compatibility
No backwards compatibility issues known.

## Follow-up
Analyse other events to see if they could also benefit from this kind of change too. Events such as `PlayerMoveEvent` might benefit from this as well.

This optimisation may also facilitate adding more hot events for testing purposes, without a huge performance impact.

## Tests
TBD


dktapps:
This gave rise to another optimisation: tracking the `RegisteredListenerCache` in `HandlerListManager` directly, bypassing `HandlerList` entirely.

dktapps:
It appears that DataPacketSendEvent would significantly benefit from this, particularly when there are many moving entities crowded into one area

PR #3998, opened 2021-01-08
Improving the language system

Saehing:
Introduction
This PR contributes to improving the server language system, also correcting some inconsistencies.

Relevant issues
Fixed translation of the text for the player is used. Previously, a player took as a basis the language that is selected on the server, now for each player his language pack is automatically selected, based on his language from LoginPacket.
Introduction
This PR helps improve the server's language system, as well as correcting some inconsistencies.

Related issues
Fixed text translation is used for the player. If previously the player used the language selected on the server as the basis, now each player automatically selects their language pack based on their language from LoginPacket.
API changes
Using the function "loadLanguages()" in LanguageManager you can now load your language packs, this can be used in plugins.

PR #5710, opened 2023-04-18
PluginBase Supported Translation API

jasonw4331:
## Introduction

If plugins wish to use the existing translations system, they must reflect into the `Server::getLanguage()` instance and add their translations directly, then call to `Language::translate()` before passing the output to `Player::sendMessage()` or setting the death event message (*both accept a `Translatable` already*).

### Relevant issues

Relates to #3998

### API changes

Adds `NamespacedLanguage` to keep track of Languages using namespaces
Adds `Language::merge()` to import translations from other Language objects
Adds `PluginBase::addTranslation()` to allow plugins to craft translations onEnable instead of by resource
Adds `PluginBase::getTranslations()` to return all registered `NamespacedLanguage` objects

### Behavioral changes

`Language::getLanguageList()` now supports reading phar resources
`NetworkSession::onChatMessage()` automatically translates `Translatable` if the namespace is registered to the server Language instance.
`DeathPacketHandler` automatically translates `Translatable` if the namespace is registered to the server Language instance.

## Backwards compatibility

BC 100%

## Follow-up

TODO

## Tests

Before:
```php
// onEnable
$languageAliases = ['eng', 'fre'];
foreach($languageAliases as $alias){
	$this->saveResource('/languages/' . $alias . '.ini', true);
}

$dir = scandir(Path::join($this->getDataFolder(), "languages"));
if($dir !== false){
	foreach($dir as $file){
		/** @phpstan-var array{dirname: string, basename: string, extension?: string, filename: string} $fileData */
		$fileData = pathinfo($file);
		if(!isset($fileData["extension"]) || $fileData["extension"] !== "ini"){
			continue;
		}
		self::$languages[$fileData["filename"]] = new Language($fileData["filename"], Path::join($this->getDataFolder(), "languages"));
	}
}

// add translations to existing server language instance
$serverLanguage = $this->getServer()->getLanguage();
$refClass = new \ReflectionClass($serverLanguage);
$refPropA = $refClass->getProperty('lang');
$refPropA->setAccessible(true);
/** @var string[] $serverLanguageList */
$serverLanguageList = $refPropA->getValue($serverLanguage);

$pluginLanguage = self::$languages[$serverLanguage->getLang()];
$refClass = new \ReflectionClass($pluginLanguage);
$refPropB = $refClass->getProperty('lang');
$refPropB->setAccessible(true);
/** @var string[] $pluginLanguageList */
$pluginLanguageList = $refPropB->getValue($pluginLanguage);

$refPropA->setValue($serverLanguage, array_merge($serverLanguageList, $pluginLanguageList));
```

```php
// onCommand
$language = $sender->getLanguage();
$sender->sendMessage(
	$language->translate(CustomKnownTranslationFactory::command_tpaccept_norequest())
);
```

After:
First code block unnecessary
```php
// onCommand
$sender->sendMessage(CustomKnownTranslationFactory::command_tpaccept_norequest());
```


jasonw4331:
Small test plugin
---

resources/translations/eng.ini
```ini
; English language file for PluginTranslationsTest
language.name=English
PluginTranslationsTest.onjoin.message1 = Welcome to the server, {%0}!
onjoin.message2 = You are player number {%0}!
```

src/Main.php
```php

		var_dump($this->getTranslations());
	}

	public function onJoin(PlayerJoinEvent $event) : void {
		$playerName = $player->getName();
		$player->sendMessage(CustomKnownTranslationFactory::PluginTranslationsTest_onjoin_message1($playerName));
		$player->sendMessage(CustomKnownTranslationFactory::PluginTranslationsTest_onjoin_message2(mt_rand()));
	}
}
```

src/lang/CustomKnownTranslationFactory.php
```php
final class CustomKnownTranslationFactory{
	public static function PluginTranslationsTest_onjoin_message1(Translatable|string $param0) : Translatable{
		return new Translatable(CustomKnownTranslationKeys::PLUGINTRANSLATIONSTEST_ONJOIN_MESSAGE1, [
			0 => $param0,
		]);
	}

	public static function PluginTranslationsTest_onjoin_message2(Translatable|string $param0) : Translatable{
		return new Translatable(CustomKnownTranslationKeys::PLUGINTRANSLATIONSTEST_ONJOIN_MESSAGE2, [
			0 => $param0,
		]);
	}

}
```

src/lang/CustomKnownTranslationKeys.php
```php
final class CustomKnownTranslationKeys{
	public const PLUGINTRANSLATIONSTEST_ONJOIN_MESSAGE1 = "PluginTranslationsTest.onjoin.message1";
	public const PLUGINTRANSLATIONSTEST_ONJOIN_MESSAGE2 = "PluginTranslationsTest.onjoin.message2";
}
```

onEnable var_dump output
```
array(1) {
  ["eng"]=>
  object(pocketmine\plugin\PluginTranslations)#21935 (5) {
    ["langName":protected]=>
    string(3) "eng"
    ["lang":protected]=>
    array(2) {
      ["PluginTranslationsTest.onjoin.message1"]=>
      string(28) "Welcome to the server, {%0}!"
      ["PluginTranslationsTest.onjoin.message2"]=>
      string(27) "You are player number {%0}!"
    }
    ["fallbackLang":protected]=>
    array(2) {
      ["PluginTranslationsTest.onjoin.message1"]=>
      string(28) "Welcome to the server, {%0}!"
      ["PluginTranslationsTest.onjoin.message2"]=>
      string(27) "You are player number {%0}!"
    }
    ["pluginNamespace":"pocketmine\plugin\PluginTranslations":private]=>
    string(22) "PluginTranslationsTest"
    ["fallbackName":"pocketmine\plugin\PluginTranslations":private]=>
    string(3) "eng"
  }
}
```

SOF3:
Is there a particular reason why we require plugins to use the same .lang format for translation files?

jasonw4331:
Language files are formatted as `.ini` and read using the same system as pocketmine's lang files. I don't really see a need to write an additional implementation to accept other file formats.

For plugins which have an existing implementation using a different file format, it can be easily converted because the translation system is at its base key value pairs and replaceable placeholders.

src/network/mcpe/NetworkSession.php
```diff
@@ -949,11 +950,12 @@ public function syncAvailableCommands() : void{
 
 	public function onChatMessage(Translatable|string $message) : void{
 		if($message instanceof Translatable){
-			$language = $this->player->getLanguage();
-			if(!$this->server->isLanguageForced()){
+			$language = Server::getInstance()->getLanguage();
```
dktapps: unnecessary singleton usage

src/network/mcpe/NetworkSession.php
```diff
@@ -949,11 +950,12 @@ public function syncAvailableCommands() : void{
...
+			$namespace = explode('.', $message->getText())[0] ?? '';
```
dktapps: explode() will always return a non-empty-array

src/network/mcpe/NetworkSession.php
```diff
@@ -949,11 +950,12 @@ public function syncAvailableCommands() : void{
...
+			if(!in_array($namespace, $language->getNamespaces(), true) || !$this->server->isLanguageForced()){
```
dktapps: This condition doesn't make sense. It will send untranslated pocketmine strings if the language isn't forced.

This should be &&, not ||.

src/network/mcpe/NetworkSession.php
dktapps: This is also breaking the ability to use languages per player.

src/network/mcpe/handler/DeathPacketHandler.php
```diff
@@ -52,11 +55,12 @@ public function setUp() : void{
 		/** @var string[] $parameters */
 		$parameters = [];
 		if($this->deathMessage instanceof Translatable){
-			$language = $this->player->getLanguage();
-			if(!$this->player->getServer()->isLanguageForced()){
+			$language = Server::getInstance()->getLanguage();
```
dktapps: see above

src/network/mcpe/handler/DeathPacketHandler.php
```diff
@@ -52,11 +55,12 @@ public function setUp() : void{
...
+			$namespace = explode('.', $this->deathMessage->getText())[0] ?? '';
```
dktapps: see above

src/network/mcpe/handler/DeathPacketHandler.php
```diff
@@ -52,11 +55,12 @@ public function setUp() : void{
...
+			if(!in_array($namespace, $language->getNamespaces(), true) || !$this->player->getServer()->isLanguageForced()){
```
dktapps: see above

src/network/mcpe/NetworkSession.php
```diff
@@ -949,11 +950,12 @@ public function syncAvailableCommands() : void{
...
 				//we can't send nested translations to the client, so make sure they are always pre-translated by the server
 				$parameters = array_map(fn(string|Translatable $p) => $p instanceof Translatable ? $language->translate($p) : $p, $message->getParameters());
-				$this->sendDataPacket(TextPacket::translation($language->translateString($message->getText(), $parameters, "pocketmine."), $parameters));
+				$this->sendDataPacket(TextPacket::translation($language->translateString($message->getText(), $parameters), $parameters));
```
dktapps: This breaks unforced translations.

src/network/mcpe/handler/DeathPacketHandler.php
```diff
@@ -52,11 +55,12 @@ public function setUp() : void{
...
 				//we can't send nested translations to the client, so make sure they are always pre-translated by the server
 				$parameters = array_map(fn(string|Translatable $p) => $p instanceof Translatable ? $language->translate($p) : $p, $this->deathMessage->getParameters());
-				$message = $language->translateString($this->deathMessage->getText(), $parameters, "pocketmine.");
+				$message = $language->translateString($this->deathMessage->getText(), $parameters);
```
dktapps: See above.

dktapps (changes requested):
This PR breaks several things and makes a bunch of changes that don't seem to be really necessary.

src/network/mcpe/NetworkSession.php
SOF3: also don't we have a convention of using double quotes?

src/network/mcpe/NetworkSession.php
jasonw4331: `ServerLanguage::getNamespaces()` contains the pocketmine namespace meaning this will never affect the output of translateString

dktapps: I don't think you understand what the purpose of this is. The point is to allow the client to translate non-`pocketmine` strings by itself. By removing this, you're forcing ALL the translations to be translated by the server.

src/network/mcpe/NetworkSession.php
jasonw4331: There is no validation for the Translatable text being passed in by a plugin to be in the `namespace.key.whatever` format at any point previously. It's not guaranteed except by PocketMine

src/network/mcpe/handler/DeathPacketHandler.php
jasonw4331: same answer above

src/network/mcpe/NetworkSession.php
dktapps: Whether or not it's validated is irrelevant. `explode` is guaranteed to always return an array with at least 1 element.

src/network/mcpe/handler/DeathPacketHandler.php
jasonw4331: still not guaranteed here

SOF3:
> Language files are formatted as `.ini` and read using the same system as pocketmine's lang files. I don't really see a need to write an additional implementation to accept other file formats.
> 
> For plugins which have an existing implementation using a different file format, it can be easily converted because the translation system is at its base key value pairs and replaceable placeholders.

I mean, do we even have a motivation to introduce a standard way of reading translations from plugins directly, instead of plugins themselves registering it during startup? Plugins could just bundle a translation library that knows how to convert their translations into the API understood by PocketMine; why do we have to implement this into the core? `config.yml` is already an unnecessary bloat for PluginBase.

dktapps:
> > Language files are formatted as `.ini` and read using the same system as pocketmine's lang files. I don't really see a need to write an additional implementation to accept other file formats.
> > For plugins which have an existing implementation using a different file format, it can be easily converted because the translation system is at its base key value pairs and replaceable placeholders.
> 
> I mean, do we even have a motivation to introduce a standard way of reading translations from plugins directly, instead of plugins themselves registering it during startup? Plugins could just bundle a translation library that knows how to convert their translations into the API understood by PocketMine; why do we have to implement this into the core? `config.yml` is already an unnecessary bloat for PluginBase.

+1 for this. The minimum working product here would be to just add APIs to Language to insert new key=string pairs. I do think the namespace enforcement is a good idea though.

jasonw4331:
> `config.yml` is already an unnecessary bloat for PluginBase.

Plugins can opt not to use that feature, or even not to use PluginBase at all. That's why the Plugin interface exists separately. PluginBase should have useful utilities like this to assist with ease of development though.



> The minimum working product here would be to just add APIs to Language to insert new key=string pairs.

Why should it only be the minimum to make it work?

SOF3:
@jasonwynn10 the same thing for forms API. We don't want to add a bunch of bloat into the core API, especially something core like PluginBase, when we aren't sure which one is the best approach. All we need is to provide the necessary APIs so that libraries can be built on top of it.

SOF3:
> Plugins can opt not to use that feature, or even not to use PluginBase at all. That's why the Plugin interface exists separately. PluginBase should have useful utilities like this to assist with ease of development though.

Because this means that if we want to improve the config API, we have to break backwards compatibility for all plugins. If plugins just statically linked their own libraries, multiple versions of the same library can co-exist and we have much less backwards compatibility burden.

SOF3:
Furthermore, the role of PluginBase is not just to "have useful utilities", but to integrate with the PharPluginLoader/FolderPluginLoader lifecycle. It is a hard requirement that the main class extends PluginBase so that the PluginLoader can initialize it correctly (e.g. passing the necessary arguments to the constructor, handling `setEnabled` properly, etc). You cannot replace PluginBase with a library class.

Things like `getResource` have to stick to PluginBase because only PluginBase knows how to interact with the PluginLoader and get the path to the resources (e.g. whether to add `phar://` or not). Meanwhile, things like `saveDefaultConfig` just work on top of `getResource` and did not need to be coupled with `PluginBase`; shoghi just did that because... well, composer was not a (popular) thing back then (along with some Bukkit cargo cult too), but we could do better than that now that we have dependency management techniques like composer and virions.

jasonw4331:
> the same thing for forms API

This is a bit off topic, but a Form API should also exist in the core somewhere because forms are a part of protocol.


There are approximately 100 total plugins on poggit at time of writing which implement a translation system of some sort. 41 of those plugins use the pocketmine built-in Language class with reflections to add their translations into the core. I think if ~40% of the publicly known plugins use this system, then adding an API to improve it's ease of use should be considered.

jasonw4331:
> It is a hard requirement that the main class extends PluginBase so that the PluginLoader can initialize it correctly

That's not true at all. There isn't a single reliance on PluginBase anywhere within the PharPluginLoader, FolderPluginLoader, or PluginManager classes.

> Things like `getResource` have to stick to PluginBase because only PluginBase knows how to interact with the PluginLoader and get the path to the resources (e.g. whether to add `phar://` or not).

PluginBase is completely PluginLoader agnostic interfacing with the ResourceProvider rather than the PluginLoader for its resources. `onLoad()` `onEnable()` `onDisable()` `getResource()` `saveResource()` and the config APIs are all there I'd argue as utility methods. Only the resource functions integrate with the ResourceProvider, not even the PluginLoader.

PluginBase is only a recommendation, not a requirement as is said in Plugin line 33 and therefore it should be for utility, not integration. There isn't even a single function call to a method in PluginBase that Plugin doesn't have with the exception of CrashDump via reflection.

dktapps:
> Furthermore, the role of PluginBase is not just to "have useful utilities", but to integrate with the PharPluginLoader/FolderPluginLoader lifecycle. It is a hard requirement that the main class extends PluginBase so that the PluginLoader can initialize it correctly (e.g. passing the necessary arguments to the constructor, handling `setEnabled` properly, etc). You cannot replace PluginBase with a library class.
> 
> Things like `getResource` have to stick to PluginBase because only PluginBase knows how to interact with the PluginLoader and get the path to the resources (e.g. whether to add `phar://` or not). Meanwhile, things like `saveDefaultConfig` just work on top of `getResource` and did not need to be coupled with `PluginBase`; shoghi just did that because... well, composer was not a (popular) thing back then (along with some Bukkit cargo cult too), but we could do better than that now that we have dependency management techniques like composer and virions.

PluginBase hasn't been a requirement since 4.0. It's entirely possible to make your own Plugin implementation. However, no one really wants to.

dktapps:
> Why should it only be the minimum to make it work?

Anything beyond that is core bloat for a feature that most plugins won't use anyway.

Automagically registering stuff has caused many issues for plugins in the past (see onCommand). We can choose to add more APIs later, but we can't remove them once they've been added. So minimalism is always the best approach.

matcracker:
> If plugins wish to use the existing translations system, they must reflect into the `Server::getLanguage()` instance and ...

I don't understand the cited point: why we must reflect the `Server::getLanguage()`? I was able to use the Language API without using any reflection. Here are some examples:

https://github.com/matcracker/BedcoreProtect/blob/3a23a3960af1fb41b5cdbf17432a13c9779bcb57/src/matcracker/BedcoreProtect/Main.php#L117

https://github.com/matcracker/BedcoreProtect/blob/3a23a3960af1fb41b5cdbf17432a13c9779bcb57/src/matcracker/BedcoreProtect/Main.php#L61-L64

https://github.com/matcracker/BedcoreProtect/blob/3a23a3960af1fb41b5cdbf17432a13c9779bcb57/src/matcracker/BedcoreProtect/storage/PatchManager.php#L70

For everything else I agree, if you can find a way to directly use the sendMessage passing a translatable key it would eliminate a lot of boiler code inside the plugin

dktapps:
The purpose is to be able to pass `Translatable` instead of having `translateString` boilerplate everywhere. You can't currently do that without reflecting your way into `Language`.

PR #5725, opened 2023-04-29
Add two missing parameters to doc comments

Wertzui123:


dktapps:
This is not correct. The bounds of the world are -64 to 320 in PM5.

Additionally, the Y axis bounds may be different for different dimensions, or we might add support for dynamic world height in the future, so these doc comments don't apply. In fact I'm pretty sure they were purposely removed.

Wertzui123:
Well, if it's -64 to 320 now (which I completely forgot about) and not 0 to 255 like previously, I think this comment would be at least as useful as the x and z comments. It just looked like someone forgot to add it there, but you can close this if you think it shouldn't be added.

dktapps:
Yeah, I think we should either be removing them, or noting that they are dependent on the world's Y axis bounds.

dktapps:
https://github.com/pmmp/PocketMine-MP/commit/b3e94ef1dccc7b540057a626122c87f888acda8a

PR #5726, opened 2023-04-29
PlayerDeathEvent add setScreenDeathMessage()

BrandPVP:
## Introduction

Currently its not possible to set different death messages in the chat and on the screen. Even if you try to send a new DeathInfoPacket, thats not gonna work because it can only be sent once.

Or if you want to hide the chat message and show only the screen message.
You can't just do setDeathMessage("""), because it will be sent in a DeathInfoPacket and you will not be able to send a new one.

### DIFF
--- a/src/event/player/PlayerDeathEvent.php
+++ b/src/event/player/PlayerDeathEvent.php
@@ -39,6 +39,7 @@ class PlayerDeathEvent extends EntityDeathEvent{
 	protected $player;
 
 	private Translatable|string $deathMessage;
+	private Translatable|string $screenDeathMessage;
 	private bool $keepInventory = false;
 	private bool $keepXp = false;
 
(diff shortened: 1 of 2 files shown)


ghost (approved):

src/event/player/PlayerDeathEvent.php
```diff
@@ -71,6 +73,14 @@ public function setDeathMessage(Translatable|string $deathMessage) : void{
 		$this->deathMessage = $deathMessage;
 	}
 
+	public function getScreenDeathMessage() : Translatable|string{
```
dktapps: I think `getDeathScreenMessage()` would be more idiomatic.

dktapps (changes requested):
The idea is OK, but the naming of some of this stuff is questionable

dktapps (approved):
I think this change also calls for documentation of the difference between the methods, but that can be done later.

PR #5735, opened 2023-05-04
add Player::getLowerCaseName()

Lunarelly:
Added a method to get lower case name instead of using strtolower($player->getName()) every time


dktapps:
This was actually removed during PM4 development due to premature optimisation.

Lunarelly:
ik it was removed but it does not affect performance so why dont add it back

Lunarelly:
better than doing strtolower() every time
or doing a custom player class

PR #5740, opened 2023-05-10
Fix major-next code style

IvanCraft623:
I am anxious that the checks on my commits in my PRs will fail :D


dktapps:
It's always the damn CS ...

PR #5744, opened 2023-05-11
added the possibility to overwrite BlockSerializer, BlockDeserializer, ItemSerializer

SenseiTarzan:


SenseiTarzan:
@dktapps

dktapps:
To be honest I don't see the use case for this. Overriding serializers is a sketchy hack anyway, and is a noisy-neighbour behaviour. Every time someone has asked for it, it's been for the same kind of XY problem case as we've previously seen with custom player classes and custom entity classes.

SenseiTarzan:
> To be honest I don't see the use case for this. Overriding serializers is a sketchy hack anyway, and is a noisy-neighbour behaviour. Every time someone has asked for it, it's been for the same kind of XY problem case as we've previously seen with custom player classes and custom entity classes.

I made this because it was foreseen that if I don't allow this people can't overwrite their block giving there is a problem in the item give, it gives an update_block

dktapps:
As I said, XY problem. /give has nothing to do with these serializers - that's the domain of StringToItemParser. And this code doesn't stop you from registering custom blocks.

SenseiTarzan:
> As I said, XY problem. /give has nothing to do with these serializers - that's the domain of StringToItemParser. And this code doesn't stop you from registering custom blocks.

plus the change I made was something that was supposed to be deleted as a //TODO: REMOVE ME

SenseiTarzan:
image

SenseiTarzan:
I do not make any case of Hack it is just that in the BlockFactory we can overwrite and not on the Serializer and the Deserializer whereas it is essential

SenseiTarzan:
I don't know why you're skeptical when it was planned

SenseiTarzan:
in addition register the class-string while it is possible to set a single serialize or deserialize not typeId

SenseiTarzan:
?

dktapps:
Yes, I allowed for the possibility, and since realized that it doesn't make any sense. 

> I do not make any case of Hack it is just that in the BlockFactory we can overwrite and not on the Serializer and the Deserializer whereas it is essential

This is a bad argument. Overriding blocks via BlockFactory was also sketchy and wasn't properly supported.

SenseiTarzan:
what do you want to do to make it better

SenseiTarzan:
then why did you allow in the ItemDeserializer the overwrite

dktapps:
> then why did you allow in the ItemDeserializer the overwrite

That's a bug. It's not supposed to allow that.

SenseiTarzan:
the problem now is that the block can't be overwritten and that means that the overwrite variable in blockFactory is useless if the DeSerializer doesn't have the right block

dktapps:
The mess around this has been cleaned up by https://github.com/pmmp/PocketMine-MP/commit/9621836e36b9a9bb8c3ce1d339871bd6a07d7cc1.

SenseiTarzan:
If you block the overwrite option then a good majority of plugins are broken and that means you will have to add the option to add custom blocks

SenseiTarzan:
So that's not to tell you that many are scaring away pmmp because it will no longer be possible to overwrite the blocks

ShockedPlot7560:
Block IDs have changed under pocketmine 5. These are now runtime only unique ids, so even with an override it shouldn't be possible to change their values.

PR #5747, opened 2023-05-16
Fix being able to pot flower pots

platz1de:
## Introduction
Flower pots can currently be placed into flower pots (which isn't part of the vanilla game for obvious reasons).

### Relevant issues
Caused by d9b050fb688155ec962f574388eb48342fc8f9d1

### API changes
--

### Behavioural changes
Flower pots aren't tagged as `pottable` anymore.

## Backwards compatibility
--

## Tests

1. Place a flower pot
2. Click on the flower pot using another flower pot (if it places down, repeat)
3. Break the flower pots and look if they drop two flower pots

Before:
https://github.com/pmmp/PocketMine-MP/assets/51201131/67601322-0a2f-4cd9-bdc1-5c11ed8eaa3c

After:
https://github.com/pmmp/PocketMine-MP/assets/51201131/2e4c7a67-7fdc-41e6-9fcf-d7e76be6ed2c


dktapps:
Nice find, thanks.

PR #5751, opened 2023-05-17
Fix: item frame placement with Flowable

### LINKED ISSUE #5734: Item Frame cannot be placed on Chest
### Issue description

- Expected result: 

https://user-images.githubusercontent.com/67502532/235832770-a113e6f4-3460-4439-92bb-9c84cafa93f3.mp4

- Actual result: 

https://user-images.githubusercontent.com/67502532/235832811-91fa59d0-019e-40fb-99a1-e7e9841ad58e.mp4

1. Place an ItemFrame next to a Chest block

### OS and versions

* PocketMine-MP: 5.0.0-BETA3+dev
* PHP: 8.2.3
* Using JIT: no
* Server OS: win
* Game version: Android

ShockedPlot7560:
## Introduction
After some tests in vanilla survival, I could notice that the item frames could be placed on any block except the Flowable.

As PocketMine does not yet implement the item frames at the top and bottom of the blocks, only the faces work for the moment.

### Relevant issues
* Fixes #5734

## Tests

https://github.com/pmmp/PocketMine-MP/assets/66992287/2dc0d3f8-585a-4f88-b62d-f18e547dff1b

### DIFF
--- a/src/block/ItemFrame.php
+++ b/src/block/ItemFrame.php
@@ -169,7 +169,7 @@ class ItemFrame extends Flowable{
 	}
 
 	public function onNearbyBlockChange() : void{
-		if(!$this->getSide(Facing::opposite($this->facing))->isSolid()){
+		if($this->getSide(Facing::opposite($this->facing)) instanceof Flowable){
 			$this->position->getWorld()->useBreakOn($this->position);
 		}
 	}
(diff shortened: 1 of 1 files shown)


dktapps:
This feels like a weird solution. I think there's some condition that we're missing here.

ShockedPlot7560:
This is the concern of the placement of the blocks. The chests, beds etc... don't accept any other block as support than the item frames (normally). The problem is that the item frames can't be placed anywhere: on the torches. 
Currently, torches and chests share the same reasoning for block support logic: none. 

Looking at the blocks that extend from a Flowable, all in vanilla did not support item frames.

dktapps:
isSolid is usually the correct solution in those cases.

I'll also point out that the bug report in question applies to PM5, not PM4.

ShockedPlot7560:
Indeed, I apologize, I close it because on PM4 it works properly

PR #5757, opened 2023-05-19
Implement fortune enchantment

### LINKED ISSUE #4800: Fortune Enchant
## Description

I can't add Fortune enchant to item.

## Alternative methods
You should adding the Fortune enchant to the api.

ShockedPlot7560:
### Relevant issues
* Fixes #4800

### API changes
Add the following methods to ``Block``:
- ``getFortuneDrops``: As for the silk touch or the normal drops

Add the ``FortuneTrait`` containing all fortune logics

## Tests

https://youtu.be/m-lTUq6S7fI


src/block/Carrot.php
```diff
@@ -23,19 +23,42 @@
...
+	}
+
+	public function getFortuneDrops(Item $item) : array{
+		return $this->getDropsForFortuneLevel(
+			$item->getEnchantmentLevel(VanillaEnchantments::FORTUNE())
+		);
+	}
+
+	/** @return Item[] */
+	private function getDropsForFortuneLevel(int $level = 0) : array{
```
dktapps: Avoid optional parameters. This is a recipe for bugs.

src/block/DoubleTallGrass.php
```diff
@@ -34,9 +33,20 @@ public function canBeReplaced() : bool{
...
 		return [];
 	}
+
+	public function isAffectedByFortune() : bool{
+		return true;
+	}
+
+	public function getFortuneDrops(Item $item) : array{
+		$fortune = VanillaEnchantments::FORTUNE();
+		return $fortune->grassDrops(
```
dktapps: Is this used anywhere outside of this class? I'm not super comfortable with putting specialized methods in general classes.

src/block/Leaves.php
```diff
@@ -173,4 +163,33 @@ public function getFlammability() : int{
...
+	public function getFortuneDrops(Item $item) : array{
+		if(($item->getBlockToolType() & BlockToolType::SHEARS) !== 0){
+			return parent::getDropsForCompatibleTool($item);
+		}
+
+		return $this->getDropsForFortuneLevel($item->getEnchantmentLevel(VanillaEnchantments::FORTUNE()));
+	}
+
+	/** @return Item[] */
+	private function getDropsForFortuneLevel(int $level = 0) : array{
```
dktapps: Avoid optional parameters.

src/block/NetherWartPlant.php
```diff
@@ -93,7 +96,25 @@ public function onRandomTick() : void{
...
+		$fortuneEnchantment = VanillaEnchantments::FORTUNE();
+		if ($this->age === self::MAX_AGE) {
+			return $fortuneEnchantment->discreteDrops(
+				$this->asItem(),
+				self::MINIMUM_DROPS,
+				self::MAXIMUM_DROPS,
+				$item->getEnchantmentLevel($fortuneEnchantment)
+			);
+		} else {
+			return [$this->asItem()];
```
dktapps: I feel like some logic is getting repeated here.

dktapps (changes requested):
I don't think it makes much sense to bake the helpers for this into FortuneEnchantment. It isn't offering any meaningful benefit, since the enchantment is statically referenced, and it's adding a bunch of extra boilerplate code.

src/block/Block.php
```diff
@@ -482,6 +495,13 @@ public function isAffectedBySilkTouch() : bool{
 		return false;
 	}
 
+	/**
+	 * Returns whether Fortune enchanted tools will cause this block to drop more items.
+	 */
+	public function isAffectedByFortune() : bool{
```
dktapps: I get the impression this method is probably redundant. If a block isn't affected by fortune, it won't override getFortuneDrops(), and the block's regular drops will be given anyway.

src/block/DoubleTallGrass.php
ShockedPlot7560: The drop of wheat seeds is specified in ``DoubleTallGrass`` and ``TallGrass``

dktapps (commented):
I'm also getting the impression that there's a lot of duplicated drops logic between fortune and non-fortune. My understanding (from a brief read) is that most of these blocks can reuse the fortune logic for the non-fortune cases, just using a level of 0, and avoid duplicating the regular logic.

Perhaps a trait which redirects `getDropsForCompatibleTool()` to `getFortuneDrops()` (or a helper `getFortuneDropsForLevel()` would simplify the code.

src/block/Beetroot.php
```diff
@@ -23,26 +23,45 @@
...
+			$item->getEnchantmentLevel(VanillaEnchantments::FORTUNE())
+		);
+	}
+
+	/** @return Item[] */
+	private function getDropsForFortuneLevel(int $level = 0) : array{
+		if ($this->age >= self::MAX_AGE) {
+			return array_merge(
+				[VanillaItems::BEETROOT()],
+				VanillaEnchantments::FORTUNE()->binomialDrops(
```
BrandPVP: Why isn't this a static method? It will improve performance

dktapps: That's a bad reason to make this change, it's a cold path.

ShockedPlot7560:
I'll put all the utilities in a trait dedicated to fortune

ShockedPlot7560:
I just moved all the logic of the fortune into a dedicated line. I'm still mixed on the use of constant for some blocks like ores to specify the min and max drop rate though.

src/block/Potato.php
```diff
@@ -23,23 +23,36 @@
...
+				$level,
+				1
+			);
+		} else {
+			$result = [
+				VanillaItems::POTATO()
+			];
+		}
 		if($this->age >= self::MAX_AGE && mt_rand(0, 49) === 0){
 			$result[] = VanillaItems::POISONOUS_POTATO();
```
dktapps, line 39: Does fortune have any effect on this?

src/block/Wheat.php
```diff
@@ -23,26 +23,34 @@
...
+	 * @return Item[]
+	 */
+	protected function getFortuneDropsForLevel(int $level) : array{
+		if ($this->age >= self::MAX_AGE) {
+			return array_merge(
+				[VanillaItems::WHEAT()],
+				$this->binomialDrops(
+					VanillaItems::WHEAT_SEEDS(),
+					$level
+				)
```
dktapps: Perhaps `binomialDrops` could accept `array<Item> $extraDrops` to avoid this array_merge boilerplate everywhere.

src/block/utils/FortuneTrait.php
```diff
@@ -0,0 +1,126 @@
...
+					VanillaItems::WHEAT_SEEDS()->setCount($drop)
+				];
+			}
+		}
+
+		return [];
+	}
+
+	/**
+	 * Discret drop, increases the maximum number of items that can be dropped by the fortune level.
```
dktapps:
```suggestion
	 * Discrete drop increases the maximum number of items that can be dropped by the fortune level.
```

src/block/utils/FortuneTrait.php
```diff
@@ -0,0 +1,126 @@
...
+	 *
+	 * @param int $maximumDropLimitation As minecraft doc, this is the maximum number of drops that can be dropped by this enchantment.
+	 *                                   If a drop higher than these maximums is rolled, it is rounded down to the capacity.
+	 * @return Item[]
+	 */
+	protected function discreteDrops(Item $item, int $min, int $max, int $fortuneLevel, int $maximumDropLimitation = PHP_INT_MAX) : array{
```
dktapps: These parameter names are confusing. `$max` isn't really the max drops, it's the _base_ maximum, to which the fortune level is added. `$maximumDropLimitation` is the real max value.

dktapps (changes requested):
This is looking better, but still needs work.

src/block/Potato.php
ShockedPlot7560: I don't see any information about it in the wiki. Maybe there is, but since we don't have an exact value, I think we can assume not.

ShockedPlot7560: Everywhere it is quoted that the poisoned potatoes are drop with 2% chance but without more

src/block/Block.php
```diff
@@ -566,6 +566,10 @@ public function getDrops(Item $item) : array{
 				return $this->getSilkTouchDrops($item);
 			}
 
+			if ($item->hasEnchantment(VanillaEnchantments::FORTUNE())) {
```
dktapps:
```suggestion
			if($item->hasEnchantment(VanillaEnchantments::FORTUNE())){
```

src/block/Gravel.php
```diff
@@ -25,24 +25,33 @@
...
+
+	public function tickFalling() : ?Block{
+		return null;
+	}
+
+	/**
+	 * @return Item[]
+	 */
+	protected function getFortuneDropsForLevel(int $level) : array{
+		if(mt_rand(1, 10 - 3 * $level) === 1){
```
dktapps: This will crash with fortune 4 or higher

src/block/Leaves.php
```diff
@@ -137,8 +143,42 @@ public function getDropsForCompatibleTool(Item $item) : array{
...
+		);
+	}
+
+	/**
+	 * @return Item[]
+	 */
+	protected function getFortuneDropsForLevel(int $level) : array{
 		$drops = [];
-		if(mt_rand(1, 20) === 1){ //Saplings
+		if(mt_rand(1, 20 - 4 * $level) === 1){ //Saplings
```
dktapps: This will crash with fortune 5 or higher

src/block/Leaves.php
```diff
@@ -154,29 +194,13 @@ public function getDropsForCompatibleTool(Item $item) : array{
 				$drops[] = $sapling;
 			}
 		}
-		if(($this->leavesType->equals(LeavesType::OAK()) || $this->leavesType->equals(LeavesType::DARK_OAK())) && mt_rand(1, 200) === 1){ //Apples
+		if(($this->leavesType->equals(LeavesType::OAK()) || $this->leavesType->equals(LeavesType::DARK_OAK())) && mt_rand(1, 200 - 20 * $level) === 1){ //Apples
```
dktapps: This will crash with fortune 10 or higher

src/block/Leaves.php
```diff
@@ -154,29 +194,13 @@ public function getDropsForCompatibleTool(Item $item) : array{
...
 			$drops[] = VanillaItems::APPLE();
 		}
-		if(mt_rand(1, 50) === 1){
+		if(mt_rand(1, 50 - 5 * $level) === 1){
```
dktapps: Same again here

src/block/utils/FortuneTrait.php
```diff
@@ -0,0 +1,130 @@
...
+	 * @return Item[]
+	 */
+	abstract protected function getFortuneDropsForLevel(int $level) : array;
+
+	/**
+	 * Gives a weight of 2 to a normal drop chance and adds a weight of 1 for each extra drop multiplier.
+	 *
+	 * @return Item[]
+	 */
+	protected function weightedDrops(Item $item, int $min, int $max, int $fortuneLevel) : array{
```
dktapps: A check for min < max should be added here, otherwise this will generate confusing errors on incorrect usage

src/block/utils/FortuneTrait.php
```diff
@@ -59,6 +60,8 @@ abstract protected function getFortuneDropsForLevel(int $level) : array;
 	 * @return Item[]
 	 */
 	protected function weightedDrops(Item $item, int $min, int $max, int $fortuneLevel) : array{
+		assert($min <= $max, "Minimum drop amount must be less than or equal to maximum drop amount");
```
dktapps: We generally don't use assert for argument validation. This should throw an InvalidArgumentException.

src/block/utils/FortuneTrait.php
```diff
@@ -118,6 +121,8 @@ protected function grassDrops(int $fortuneLevel) : array {
 	 * @return Item[]
 	 */
 	protected function discreteDrops(Item $item, int $minBaseAmount, int $maxBaseAmount, int $fortuneLevel, int $maximum = PHP_INT_MAX) : array{
+		assert($minBaseAmount <= $maxBaseAmount, "Minimum base drop amount must be less than or equal to maximum base drop amount");
```
dktapps: Same again here.

src/block/Gravel.php
```diff
@@ -25,24 +25,34 @@
...
 
-	public function getDropsForCompatibleTool(Item $item) : array{
-		if(mt_rand(1, 10) === 1){
+	public function isAffectedBySilkTouch() : bool{
+		return true;
+	}
+
+	public function tickFalling() : ?Block{
+		return null;
+	}
```
dktapps: Unrelated change

src/block/Leaves.php
```diff
@@ -154,29 +195,13 @@ public function getDropsForCompatibleTool(Item $item) : array{
...
-	}
-
-	public function getFlammability() : int{
-		return 60;
-	}
-
-	public function getSupportType(int $facing) : SupportType{
-		return SupportType::NONE();
-	}
 }
```
dktapps: This diff is quite hard to read due to the logic being moved. I think it would make more sense to have the fortune functions near the original drop functions rather than dumping them at the bottom of every file.

src/block/Potato.php
```diff
@@ -23,16 +23,33 @@
...
 class Potato extends Crops{
+	use FortuneTrait;
 
-	public function getDropsForCompatibleTool(Item $item) : array{
-		$result = [
-			VanillaItems::POTATO()->setCount($this->age >= self::MAX_AGE ? mt_rand(1, 5) : 1)
-		];
+	public function getPickedItem(bool $addUserData = false) : Item{
+		return VanillaItems::POTATO();
+	}
```
dktapps: This doesn't seem related to the PR. It's already handled by asItem().

Again, the diff for this file looks very chaotic due to logic moving around, making it difficult to read.

dktapps (changes requested):
I'd be happier if the fortune drop functions were in the same position in the `Block` classes as the old drops code that they are replacing (or next to them, if they aren't being replaced). Currently, this diff is quite noisy due to all the functions being dumped at the bottom of the files, which is also not great for organization.

src/block/Leaves.php
```diff
@@ -23,22 +23,29 @@
...
+use pocketmine\world\generator\object\TreeType;
 use pocketmine\world\World;
+use function max;
 use function mt_rand;
 
 class Leaves extends Transparent{
+	use FortuneTrait;
+
 	protected LeavesType $leavesType; //immutable for now
+	protected TreeType $treeType;
```
dktapps: Unrelated change

dktapps (changes requested):

ShockedPlot7560:
It seems that this PR does not implement Sweet berries. If we were to consider doing so, it might break the BC due to the need to add a parameter to the ``getBerryDropAmount`` function for fortune calculation.

src/block/Carrot.php
```diff
@@ -23,15 +23,15 @@
...
 use pocketmine\item\Item;
 use pocketmine\item\VanillaItems;
-use function mt_rand;
 
 class Carrot extends Crops{
 
 	public function getDropsForCompatibleTool(Item $item) : array{
 		return [
-			VanillaItems::CARROT()->setCount($this->age >= self::MAX_AGE ? mt_rand(1, 4) : 1)
+			VanillaItems::CARROT()->setCount($this->age >= self::MAX_AGE ? FortuneDropHelper::binomial($item, 1) : 1)
```
ShockedPlot7560, line 34: According to the wiki, it seems that 2 carrots are fixed by default

dktapps:
> It seems that this PR does not implement Sweet berries. If we were to consider doing so, it might break the BC due to the need to add a parameter to the `getBerryDropAmount` function for fortune calculation.

We can't change it due to BC breaks. Really that method shouldn't have been public ...

src/block/utils/FortuneDropHelper.php
```diff
@@ -123,10 +123,7 @@ public static function discrete(Item $usedItem, int $min, int $maxBase, int $max
 			throw new \InvalidArgumentException("Minimum base drop amount must be less than or equal to maximum base drop amount");
 		}
 
-		$max = min(
-			$maxLimit,
-			$maxBase + $usedItem->getEnchantmentLevel(VanillaEnchantments::FORTUNE())
-		);
-		return mt_rand($min, $max);
+		$drop = mt_rand($min, $maxBase + $usedItem->getEnchantmentLevel(VanillaEnchantments::FORTUNE()));
+		return min($drop, $maxLimit);
```
dktapps: I'd actually intended to remove the maxLimit from here.

src/block/utils/FortuneDropHelper.php
```diff
@@ -0,0 +1,137 @@
...
+	 *
+	 * @return int the number of items to drop
+	 */
+	public static function discrete(Item $usedItem, int $min, int $maxBase) : int{
+		if($maxBase < $min){
+			throw new \InvalidArgumentException("Minimum base drop amount must be less than or equal to maximum base drop amount");
+		}
+
+		$max = $maxBase + $usedItem->getEnchantmentLevel(VanillaEnchantments::FORTUNE());
+		return mt_rand($min, $max);
```
ShockedPlot7560, line 122: that's not right. For sealantern, for example, with fortune 3, there's a 20% chance of getting 2, 3, 4, but a 40% chance of getting 5.

ShockedPlot7560: This works well for those with a uniform distribution, but not for the rest who have a limit.

dktapps: I'm pretty sure it's supposed to work that way.

> Each level of Fortune increases the maximum number of prismarine crystals dropped. **The amount is capped at 5, so Fortune III simply increases the chance of getting 5 crystals.**

dktapps: Same story for melons:
> The actual amount is capped at 9, so Fortune III simply increases the chance of getting 9 slices, which is sufficient to craft the melon again from the slices.

For glowstone it's a bit more hazy, but the Java logic for these blocks is identical.

dktapps: A quick test case:
```php
$hits = [];

$runs = 10000;

$item = VanillaItems::DIAMOND_PICKAXE()->addEnchantment(new EnchantmentInstance(VanillaEnchantments::FORTUNE(), 3));
for($i = 0; $i < $runs; $i++){
	$amount = min(5, FortuneDropHelper::discrete($item, 2, 3));

	$hits[$amount] = ($hits[$amount] ?? 0) + 1;
}

ksort($hits, SORT_NUMERIC);
foreach($hits as $amount => $count){
	echo "$amount: " . (($count / $runs) * 100) . "%\n";
}
```
Output:
```
2: 20.03%
3: 19.28%
4: 20.68%
5: 40.01%
```

dktapps:
Live on `test.pmmp.io:19132` for testing

src/block/utils/FortuneDropHelper.php
ShockedPlot7560: > I'm pretty sure it's supposed to work that way.
> 
> > Each level of Fortune increases the maximum number of prismarine crystals dropped. **The amount is capped at 5, so Fortune III simply increases the chance of getting 5 crystals.**

that's surprising because that's not what it says on the wiki fortune's page.

dktapps: To paraphrase my Discord comments:

Fortune 3 on sea lantern would, if uncapped, have a 20% chance of dropping 6 crystals. However, it's capped by `min(5, amount)` in `SeaLantern`. This means that 6 gets rounded down to 5. The chance of 5 crystals is then 40%, since there are two possible results which would produce a drop count of 5.

Other blocks have the same behaviour.

src/block/Carrot.php
ShockedPlot7560: > The yield is calculated by a binomial distribution: 2 drops are fixed, then a drop is attempted three times with a success rate of 57.14286% to yield the extra 0–3 drops. Each level of Fortune enchantment increases the number of attempts by one.

https://minecraft.fandom.com/wiki/Carrot

dktapps: does this apply to Bedrock?

dktapps: This does not appear to apply to Bedrock from what I can see. Without Fortune I always get 1-4 carrots.

dktapps (approved):
This now seems good to go.

PR #5774, opened 2023-05-29
Addition of new minecraft colours, deletion of colours that have been removed from the game

roimee6:
## Introduction
Minecraft has just added the colours that are linked to the materials (colours slightly modified from the classic ones) and Minecraft has removed a few other colours (I don't know since when) which are the underline and the strikethrough. 

## Tests
On this photo I realised that minecraft didn't have the underline and the strikethrough (I removed them).
image

### DIFF
--- a/src/utils/Terminal.php
+++ b/src/utils/Terminal.php
@@ -37,8 +37,6 @@ abstract class Terminal{
 	public static string $FORMAT_BOLD = "";
 	public static string $FORMAT_OBFUSCATED = "";
 	public static string $FORMAT_ITALIC = "";
-	public static string $FORMAT_UNDERLINE = "";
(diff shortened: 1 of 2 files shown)


dktapps:
The entire contents of both files has been reformatted and is illegible. Don't use GitHub to edit or upload files.

roimee6:
I've just realised that github misunderstood that I had updated the file and not deleted it and then re-added it.

And also the indexing is weird

roimee6:
> The entire contents of both files has been reformatted and is illegible. Don't use GitHub to edit or upload files.

I'm redoing the pullrequest

SOF3:
Editing is fine if the change is only one file. The issue with uploading is that it preserved the CRLF encoding, so if you're doing multi-file changes, please use [a git client](https://git-scm.com).

PR #5783, opened 2023-05-31
Remove method call to ReflectionProperty::setAccessible()

Muqsit:
## Introduction

As of PHP 8.1, `ReflectionProperty::getValue()` and `ReflectionProperty::setValue()` do not require calling `ReflectionProperty::setAccessible(true)` ([rfc:make-reflection-setaccessible-no-op](https://wiki.php.net/rfc/make-reflection-setaccessible-no-op)).

### Relevant issues
No issues exist regarding PHP 8.1's reflection-setaccessible change as calling `ReflectionClass::setAccessible(true)` has [no effect](https://www.php.net/manual/en/reflectionproperty.setaccessible.php).

### API changes

This PR does not introduce API changes.

### Behavioural changes

This PR does not introduce behavioural changes.

## Backwards compatibility

This change is backward-incompatible with PHP versions < 8.1. However, this should not be an issue since 8.1 is now minimum version (https://github.com/pmmp/PocketMine-MP/commit/043350753b9c93a6a7e508fbaec0744e5d2d4beb).

## Follow-up

Similar change may be made to dependency libraries that are now using PHP 8.1 as minimum version.

## Tests

Using PHP 8.1, the following code does not result in an error due to `ReflectionProperty::setAccessible()` not being called.
https://3v4l.org/689D9#v8.1.19
```php
<?php

declare(strict_types=1);

final class IntObj{
    
    public function __construct(
        private int $value
    ){}
}

$obj = new IntObj(4);

$_value = new ReflectionProperty(IntObj::class, "value");
$value = $_value->getValue($obj);
assert($value === 4);
```

### DIFF
--- a/src/MemoryManager.php
+++ b/src/MemoryManager.php
@@ -316,9 +316,6 @@ class MemoryManager{
 					continue;
 				}
 
-				if(!$property->isPublic()){
-					$property->setAccessible(true);
-				}
 				if(!$property->isInitialized()){
 					continue;
 				}
@@ -441,9 +438,6 @@ class MemoryManager{
 								}else{
 									continue;
 								}
-							}
-							if(!$property->isPublic()){
-								$property->setAccessible(true);
 							}
 							if(!$property->isInitialized($object)){
 								continue;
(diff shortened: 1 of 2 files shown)


dktapps:
Thank you!

PR #5798, opened 2023-06-04
Implement timings for AsyncTask completion

### LINKED ISSUE #5749: Add timings for AsyncTask onCompletion handlers
## Description
Currently, timings from AsyncTask completion handlers are dumped into Server Mid-Tick Processing, with some random subtimers from paths they happen to activate (e.g. Player Send Chunks).

We need timings for individual `onCompletion` handlers to make it clear which tasks are doing heavy work on the main thread.

## Alternative methods
We can allow plugins to do their own timings, but thus far they haven't been very good at it, and this still doesn't account for core tasks.

ShockedPlot7560:
## Introduction
Allows you to identify resource-intensive tasks once back in the main thread.

### Relevant issues
* Fixes #5749

## Changes
* Add ``Timings::getAsyncTasksProgressUpdateTimings()``
* Add ``Timings::getAsyncTasksCompletionTimings()``
* Add ``Timings::getAsyncTasksErrorTimings()``

## Tests
* Launch your server 
* Send ``timings on``
* Join or do what you want
* Send ``timings paste`` and view the result: https://timings.pmmp.io/?id=306362


src/scheduler/AsyncTask.php
```diff
@@ -170,8 +171,14 @@ public function publishProgress(mixed $progress) : void{
...
-		while(($progress = $this->progressUpdates->shift()) !== null){
-			$this->onProgressUpdate(igbinary_unserialize($progress));
+		$timing = Timings::getAsyncTasksProgressUpdateTimings($this);
+		$timing->startTiming();
+		try{
+			while(($progress = $this->progressUpdates->shift()) !== null){
+				$this->onProgressUpdate(igbinary_unserialize($progress));
+			}
+		} finally{
+			$timing->stopTiming();
```
dktapps: why put the timings here? seems like they would be more at home in the AsyncPool directly

src/timings/Timings.php
```diff
@@ -299,4 +306,43 @@ public static function getEventHandlerTimings(string $event, string $handlerName
 
 		return self::$eventHandlers[$event][$handlerName];
 	}
+
+	public static function getAsyncTasksProgressUpdateTimings(AsyncTask $task) : TimingsHandler{
+		$taskClass = get_class($task);
+		if(!isset(self::$asyncTasksProgressUpdate[$taskClass])){
+			self::$asyncTasksProgressUpdate[$taskClass] = new TimingsHandler(
+				"AsyncTask Progress Update - " . $taskClass,
```
dktapps:
```suggestion
				"AsyncTask - Progress Updates - " . self::shortenCoreClassName($taskClass, "pocketmine\\"),
```

src/timings/Timings.php
```diff
@@ -299,4 +306,43 @@ public static function getEventHandlerTimings(string $event, string $handlerName
...
+				self::$schedulerAsync,
+				self::GROUP_BREAKDOWN
```
dktapps: This should probably be parameterized. We can't specify groups for this yet (we don't know which plugin it belongs to), but it'll likely be desirable in the future.

src/timings/Timings.php
```diff
@@ -299,4 +306,43 @@ public static function getEventHandlerTimings(string $event, string $handlerName
 
 		return self::$eventHandlers[$event][$handlerName];
 	}
+
+	public static function getAsyncTasksProgressUpdateTimings(AsyncTask $task) : TimingsHandler{
+		$taskClass = get_class($task);
+		if(!isset(self::$asyncTasksProgressUpdate[$taskClass])){
+			self::$asyncTasksProgressUpdate[$taskClass] = new TimingsHandler(
+				"AsyncTask Progress Update - " . $taskClass,
+				self::$schedulerAsync,
```
dktapps: This is problematic for offline/unit testing, as the timer may not be initialized

src/timings/Timings.php
```diff
@@ -299,4 +306,43 @@ public static function getEventHandlerTimings(string $event, string $handlerName
 
 		return self::$eventHandlers[$event][$handlerName];
 	}
+
+	public static function getAsyncTasksProgressUpdateTimings(AsyncTask $task) : TimingsHandler{
+		$taskClass = get_class($task);
```
dktapps: `$task::class` can be used instead of an extra variable

src/timings/Timings.php
```diff
@@ -299,4 +306,43 @@ public static function getEventHandlerTimings(string $event, string $handlerName
...
+		if(!isset(self::$asyncTasksCompletion[$taskClass])){
+			self::$asyncTasksCompletion[$taskClass] = new TimingsHandler(
+				"AsyncTask Completion - " . $taskClass,
+				self::$schedulerAsync,
+				self::GROUP_BREAKDOWN
+			);
+		}
+
+		return self::$asyncTasksCompletion[$taskClass];
+	}
```
dktapps, line 345: See comments above

src/timings/Timings.php
```diff
@@ -299,4 +306,43 @@ public static function getEventHandlerTimings(string $event, string $handlerName
...
+		if(!isset(self::$asyncTasksError[$taskClass])){
+			self::$asyncTasksError[$taskClass] = new TimingsHandler(
+				"AsyncTask Error - " . $taskClass,
+				self::$schedulerAsync,
+				self::GROUP_BREAKDOWN
+			);
+		}
+
+		return self::$asyncTasksError[$taskClass];
+	}
```
dktapps, line 359: See comments above

src/timings/Timings.php
dktapps: I wonder if we might be able to get away with removing the `AsyncTask` prefix here. It makes the timings output unnecessarily noisy in tree mode.

dktapps:
PHPUnit build stalled due to test errors

src/timings/Timings.php
ShockedPlot7560: Deleting it would certainly reduce the number of entries in Timings, but it could lead to confusion if the class name isn't so explicit and the fact that it's an AsyncTask behind it isn't self-evident.

ShockedPlot7560: Or we can have a naming scheme like this: ``SomeShortenClass::onCompletion()`` / ``SomeShortenClass::onError()`` / ``SomeShortenClass::onProgressUpdates()``

dktapps: Fair enough, probably better to let the prefix stay then.

src/scheduler/AsyncPool.php
```diff
@@ -247,7 +247,9 @@ public function collectTasksFromWorker(int $worker) : bool{
 					 * lost. Thus, it's necessary to do one last check here to make sure all progress updates have
 					 * been consumed before completing.
 					 */
-					$task->checkProgressUpdates();
+					Timings::getAsyncTasksProgressUpdateTimings($task)->time(function() use ($task) : void{
+						$task->checkProgressUpdates();
```
dktapps: This is called in multiple places. Perhaps a wrapper is needed.

src/timings/Timings.php
```diff
@@ -338,7 +338,7 @@ public static function getAsyncTasksErrorTimings(AsyncTask $task, string $group
 		if(!isset(self::$asyncTasksError[$taskClass])){
 			self::$asyncTasksError[$taskClass] = new TimingsHandler(
 				"Error - " . self::shortenCoreClassName($taskClass, "pocketmine\\"),
-				self::$schedulerAsync,
+				self::$schedulerAsync ?? null,
```
dktapps: It would be better to use `Timings::init()` to ensure the proper structure is maintained.

ShockedPlot7560: This means that in the case of unit testing, we're forming completely useless handlers. Wouldn't it just be better to have a function that retrieves the handler from the async task, if it's not defined (init not called) that creates it? Like : 
```php
public static function getSchedulerAsyncTiming() : TimingsHandler;
```

ShockedPlot7560: After all, this could also happen to other handlers in the future, so why not generalize the init of Timings even during unit tests, rather than making a specific case for AsyncTask?

dktapps: > This means that in the case of unit testing, we're forming completely useless handlers. Wouldn't it just be better to have a function that retrieves the handler from the async task, if it's not defined (init not called) that creates it? Like :
> 
> ```
> public static function getSchedulerAsyncTiming() : TimingsHandler;
> ```

This adds extra opcodes and a function call, increasing the performance influence of timings. Timings should minimally influence performance.

You can simply call Timings::init() within these functions to ensure they are available. We may want to unit test timings themselves in the future also, so hierarchy should be preserved.

src/timings/Timings.php
```diff
@@ -299,4 +306,52 @@ public static function getEventHandlerTimings(string $event, string $handlerName
 
 		return self::$eventHandlers[$event][$handlerName];
 	}
+
+	public static function getAsyncTasksProgressUpdateTimings(AsyncTask $task, string $group = self::GROUP_BREAKDOWN) : TimingsHandler{
+		if (!isset(self::$schedulerAsync)){
+			self::init();
+		}
```
dktapps: You can do this inside the isset() block below instead of doing it on every call.

src/timings/Timings.php
```diff
@@ -299,4 +306,46 @@ public static function getEventHandlerTimings(string $event, string $handlerName
 
 		return self::$eventHandlers[$event][$handlerName];
 	}
+
+	public static function getAsyncTasksProgressUpdateTimings(AsyncTask $task, string $group = self::GROUP_BREAKDOWN) : TimingsHandler{
+		$taskClass = $task::class;
+		if(!isset(self::$asyncTasksProgressUpdate[$taskClass])){
+			self::init();
+			self::$asyncTasksProgressUpdate[$taskClass] = new TimingsHandler(
+				"AsyncTask - " . self::shortenCoreClassName($taskClass, "pocketmine\\") . "::onProgressUpdate()",
```
dktapps: This isn't actually measuring `onProgressUpdate()`, as it won't be called if there are no progress updates available.

src/timings/Timings.php
```diff
@@ -310,9 +310,7 @@ public static function getEventHandlerTimings(string $event, string $handlerName
 	public static function getAsyncTasksProgressUpdateTimings(AsyncTask $task, string $group = self::GROUP_BREAKDOWN) : TimingsHandler{
 		$taskClass = $task::class;
 		if(!isset(self::$asyncTasksProgressUpdate[$taskClass])){
-			if (!isset(self::$schedulerAsync)){
-				self::init();
-			}
+			self::init();
```
ShockedPlot7560, line 322: This will reform an instance of all the other handlers when a new class is encountered, won't it?

dktapps: No. init() has its own guard variable to prevent double initialization.

ShockedPlot7560: So much for me

PR #5800, opened 2023-06-04
Implement a proper solution to stop the music when a jukebox is destroyed

IvanCraft623:
## Introduction

The current solution presents some problems as first of all Tiles should not handle the behavior of blocks, instead of using the jukebox API to eject the disc directly boradcasts `RecordStopSound`, which causes the loss of the disc by destroying the jukebox which at the same time does not match the Vanilla behavior.

This PR addresses the problems mentioned above by adding the `Block::onDestroy()` method, making the block not dependent on the Tile to manage this logic, which as a plus eliminates duplicate code in `Block::onBreak()` and `Explosion::explodeB()`.


src/block/Block.php
```diff
@@ -427,14 +427,20 @@ public function getBreakInfo() : BlockBreakInfo{
 	 * @param Item[] &$returnedItems Items to be added to the target's inventory (or dropped, if full)
 	 */
 	public function onBreak(Item $item, ?Player $player = null, array &$returnedItems = []) : bool{
-		$world = $this->position->getWorld();
-		if(($t = $world->getTile($this->position)) !== null){
-			$t->onBlockDestroyed();
-		}
-		$world->setBlock($this->position, VanillaBlocks::AIR());
+		$this->onDestroy();
```
dktapps: Overriders of onBreak might not call this.

dktapps (commented):
I agree that something like this is needed, but I'm not super happy with the "destroy" terminology. It feels too ambiguous.

src/block/Block.php
IvanCraft623: Overriders always call and should call parent, however currently if they do not call it, the block will not replaced  with air.

So if they don't call parent, it will be their fault.

dktapps: They may replace it with air themselves, assuming that that's all the parent is doing. In such case, onDestroy() won't be called.

IvanCraft623: That has been solved by moving the call of `Block::onDestroy()` to `World::destroyBlockInternal()`

src/block/Block.php
```diff
@@ -427,14 +427,19 @@ public function getBreakInfo() : BlockBreakInfo{
...
-		}
-		$world->setBlock($this->position, VanillaBlocks::AIR());
+		$this->position->getWorld()->setBlock($this->position, VanillaBlocks::AIR());
 		return true;
 	}
 
+	/**
+	 * Called when this block is destroyed either when a player breaks it or is hit by an explosion.
+	 */
+	public function onDestroy() : void{
```
ShockedPlot7560, line 493: Like @dktapps I don't think that ``destroy`` terminology is appropriate in this case. Perhaps, as with the placement logic, an `onPostBreak` would be preferable? It would be less ambiguous and not so funny in the sense that the block would already have been destroyed.

src/world/World.php
```diff
@@ -2012,11 +2012,7 @@ private function destroyBlockInternal(Block $target, Item $item, ?Player $player
 		}
 
 		$target->onBreak($item, $player, $returnedItems);
-
-		$tile = $this->getTile($target->getPosition());
-		if($tile !== null){
-			$tile->onBlockDestroyed();
-		}
+		$target->onDestroy();
```
ShockedPlot7560, line 2228: Shouldn't this function only be called if the onBreak returns true? There is a return value, but it's never used, and yet it would make sense in this case. If the broken action has been confirmed, then we can call this function, which is there to perform post-broken actions.

src/block/Block.php
dktapps: I'm not sure that's desirable. If the block was already destroyed, contents of stuff like tiles wouldn't be accessible.

Personally I would lean towards an `onExplode()` method or something similar. We already have the likes of `onIncinerate()` for fire. The problem is that this would require multiple handlers to be implemented for dealing with tile drops.

dktapps:
Has anyone considered using `Block::onBreak()` for this instead of adding an extra method? It does allow passing `null` for the player parameter.

I haven't thought much about the pros and cons of this approach, but I think it should be considered.

dktapps:
Linking #4652 as I think these are two different solutions for essentially the same problem

src/block/Block.php
```diff
@@ -483,14 +483,19 @@ public function getEnchantmentTags() : array{
 	 * @param Item[] &$returnedItems Items to be added to the target's inventory (or dropped, if full)
 	 */
 	public function onBreak(Item $item, ?Player $player = null, array &$returnedItems = []) : bool{
-		$world = $this->position->getWorld();
-		if(($t = $world->getTile($this->position)) !== null){
-			$t->onBlockDestroyed();
-		}
-		$world->setBlock($this->position, VanillaBlocks::AIR());
+		$this->position->getWorld()->setBlock($this->position, VanillaBlocks::AIR());
```
dktapps, line 486: Noting this is probably BC breaking, since the behaviour is now changed for non-core consumers of `onBreak()`

dktapps:
The disc does actually appear to drop on an explosion. It's probably because the yield is less than 100% that you didn't see that.

The main deviation from vanilla would appear to be that the disc doesn't drop when the jukebox is broken by a creative player. So something akin to #4652 would probably be better suited to dealing with that problem, or we could have it drop in `onBreak()`.

There is also the issue of containers dropping their contents in `onBlockDestroyedHook()` that nobody seems to have paid any mind to. Perhaps that should be considered for this too.

Issue #5641, opened 2023-03-17
Special-purpose workers for network packet compression


dktapps:
## Description
Async compression currently has a lot of overhead, due to allocating and deallocating a new `AsyncTask` object for every batch compressed. When there are a large number of packets to be compressed (e.g. when `async-compression: true` is used), this causes a significant amount of main thread CPU wastage.

In some cases (and especially since the introduction of [libdeflate](https://github.com/ebiggers/libdeflate), which brought significant performance improvements), the overhead of async compression is actually higher than the time saved by threading the compression, rendering it not only useless, but harmful to performance.

A naive local test suggests that dedicated threads using simple buffers would reduce overhead by as much as 85%, making async compression a lot more practical.

dktapps:
Further testing reveals that even in the common cases (compression level 6, and 300+ bytes) the overhead of async compression is _usually_ bigger than the cost of compressing on the main thread. The situation is even worse than I initially believed.

Issue #1161, opened 2017-07-01
Plugins hogging the AsyncWorkers causes significant server slowdown


dktapps:
### Issue description
A frequent problem I see these days is plugins are now often using AsyncTasks to perform lengthy blocking operations. When I say "lengthy", I mean > 1 tick.

This can become a serious problem if plugins are frequently scheduling lots of AsyncTasks, especially if they are performing blocking operations such as disk I/O or network I/O. 

Since ALPHA6, the PocketMine-MP core now always uses AsyncTasks for chunk compression and serialization to improve performance and take advantage of the whole of the CPU.

PocketMine-MP opens a thread per core of the CPU (in hyperthreaded CPUs this is one per logical core) to allow CPU-expensive operations such as compression and world generation to take advantage of the entire CPU. If a plugin schedules an AsyncTask to a worker to perform a blocking operation, that worker is subsequently unusable until the operation completes. This means that while the operation is executing, one core of your CPU is not doing anything.

This problem gets worse when plugins spam lots of blocking AsyncTasks. The tasks will be distributed evenly across the AsyncWorkers, which means you can arrive at the point where _all_ of the workers are performing blocking jobs... which means that important _core_ stuff that executes on the workers, such as chunk serialization and compression, is significantly delayed, while your CPU does nothing whatsoever and could be executing other tasks.

Increasing the worker pool size won't help the problem either, because as stated above, tasks are evenly distributed across all workers. Increasing the worker pool size will only slightly reduce the size of the problem.

The TL;DR of this is that it is a bad idea to use the AsyncWorkers heavily for blocking operations, as it will impact on the entire server performance.

### Steps to reproduce the issue

1. Make a plugin that does lots of blocking operations with lots of AsyncTasks
2. Watch server performance appear to deteriorate while actually waiting for blocking operations to finish on workers.

Misteboss:
Is it related to the infinite generation issue ?

dktapps:
That depends. Have you tested the Building Terrain problem on a server with no plugins installed?

Misteboss:
Nop I am going to try but to make it happens I need time and lots of players to join.
But it can be related cause I am using a lot of AsyncTask ( ~ 1 scheduled every 2 secs).

SOF3:
Note that the second AsyncPool must have enough workers too. Otherwise, the blocking operations will still accumulate.

SOF3:
[...]

PR #5805, opened 2023-06-05
First look at specialized network compression threads

dktapps:
## Introduction
As described in #5641, using AsyncTasks for compression is no longer cost-effective, due to the large overhead of creating an AsyncTask and scheduling it.

This PR introduces a specialized thread pool for network compression, which once again makes threaded compression practical.
It also uses Snooze to minimize latency.

The current implementation isn't optimal, but it's much faster than using AsyncTasks.
### Relevant issues
#5641 

### API changes
No public API changes - this is all network internals.

### Behavioural changes
- Several more threads are now started by the server, specifically to handle compression. These threads have a memory footprint of 1-2 MB each, which may become noticeable with large numbers of threads.
- `network.async-compression` in `pocketmine.yml` is now ignored, as is `network.async-compression-threshold`.
- A new setting `network.compression-threads` has been added, with a default of `0` (for now). This can be set to a positive integer or `auto` in a similar manner to `settings.async-workers`.

## Backwards compatibility
Fully backwards compatible.

## Follow-up
Explore whether chunk prepare and compression needs similar treatment. Chunk handling bypasses the usual systems using `ChunkRequestTask`, which is not modified here.

## Tests
This has been mostly playtested and benchmarked on Windows. Unit tests TBD.


ShockedPlot7560:
Related to: #1161

dktapps:
Yes, I'd actually forgotten about that issue. Though this change doesn't solve #1161 entirely.

dktapps:
User reports suggest that this change may offer a substantial performance improvement to certain types of servers.

dktapps:
This is potentially affected by pmmp/ext-pmmpthread#42, but profiling suggests that the issue should be minimal for the situations that the ThreadSafeArrays used here are subjected to

dktapps:
This should probably include some kind of GC mechanism to dispose of idle pools and free up memory

dktapps:
This PR may be problematic because it allows `compression-threads` to be disabled.

Currently, we force async compression during the player login sequence even if it's not enabled by config option, because a lot of large, non-latency-sensitive data is sent during login sequence. (The config option only decides whether to use async compression during main game runtime when the player is interacting with the world.)

With this PR, if `compression-threads` is set to 0, and because `CompressBatchTask` was removed, that data would all be compressed on the main thread, which would probably cause a significant performance hit if users aren't aware of this.

Options available:
- Don't allow disabling threaded compression (remove ability to disable by setting `compression-threads` to `0`)
- Preserve the legacy `CompressBatchTask`-based threaded compression

PR #5826, opened 2023-06-12
implement separated asyncPool for world

SenseiTarzan:
## Introduction
The problem he had with this system was that the generation of the worlds stopped if he had an AsyncTask that was blocked for X time, all the pieces that had to generate had to wait this time for the asyncTask to finish, but I separated the asyncPool from the Server and from each world to have a better performance in the generation of the world.

### API changes
Add the following functions to the `Server`class:

- `autoloader` is change in public readonly

Add the following functions to the `World`class:

- `workerPool` to `asyncPool`
- `getAsyncPool` can use on WorldLoadEvent

### Behavioural changes
to be tested on a larger scale
image
enables parallel generation

## Backwards compatibility

then it will have porblems in the Block or Item adds that it will also have to be synchronized in each world

## Follow-up

then I think we should make a real AsyncPool with a different name because when we turn them off and put the World logger

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -761,7 +761,7 @@ class Server{
 	}
 
 	public function __construct(
-		private ThreadSafeClassLoader $autoloader,
+		public  readonly ThreadSafeClassLoader $autoloader,
 		private AttachableThreadSafeLogger $logger,
 		string $dataPath,
 		string $pluginPath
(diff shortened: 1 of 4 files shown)


SenseiTarzan:
@dktapps

dktapps:
I don't think a pool per world is a good idea. Each new thread increases memory footprint, and when generators are involved, that's 10-20 MB per thread. It adds up really fast.

A separate pool for I/O tasks would make more sense, but typically people just use libasynql for database access, and file access won't usually take long enough to be noticeable.

src/Server.php
```diff
@@ -761,7 +761,7 @@ public static function getInstance() : Server{
 	}
 
 	public function __construct(
-		private ThreadSafeClassLoader $autoloader,
+		public  readonly ThreadSafeClassLoader $autoloader,
```
dktapps, line 764: This change is unnecessary

SenseiTarzan:
Tell me which asynctask consumes the most

SenseiTarzan:
Afterwards I had planned to do it in the worldManager but I didn't ask myself too many questions

dktapps:
> Afterwards I had planned to do it in the worldManager but I didn't ask myself too many questions

This would make more sense, yes.

dktapps:
Closing due to lack of activity. That being said, I don't disagree with the idea of having specialized worker pools for certain things to avoid stalling.

SenseiTarzan:
> Closing due to lack of activity. That being said, I don't disagree with the idea of having specialized worker pools for certain things to avoid stalling.

I would have to see if the process could be separated to avoid blockages.

SenseiTarzan:
> Closing due to lack of activity. That being said, I don't disagree with the idea of having specialized worker pools for certain things to avoid stalling.

I would have to see if the process could be separated to avoid blockages.

PR #5827, opened 2023-06-12
Implement Chiseled Bookshelf

HimmelKreis4865:
## Introduction
This PR implements functionality for the Chiseled Bookshelf, added in Minecraft 1.20

### Relevant issues
Sounds are not available as I was not able to make the sound appear.
For now it seems like there is some strange extra data.

## Tests
- Block faces are supported, break time is as in vanilla.
- The Bookshelf contains up to 6 books, book & quill, enchanted books, written books
- Drops of the block are only the books stored inside, except for breaking with a tool that is enchanted with silk touch, in this case the bookshelf is dropped too
- If the players inventory is full, the book will drop (starting from the eye position, moving towards the bookshelf)
- Books can be taken with any item, no blocks can be placed in front of the books (only possible if the player is sneaking)
- Item data did not change over several server restarts


src/block/ChiseledBookshelf.php
```diff
@@ -0,0 +1,144 @@
...
+			 /** @var WritableBookBase|Book $item */
+			 $this->setItem($index, $item->pop());
+		}
+		$this->recalculateBooksBit();
+		$this->position->getWorld()->setBlock($this->position, $this);
+		return true;
+	}
+
+	public function getDrops(Item $item) : array{
+		return ($item->hasEnchantment(VanillaEnchantments::SILK_TOUCH()) ? [$this->asItem(), ...$this->items] : $this->items);
```
ShockedPlot7560: Use ``getSilkTouchDrops`` rather than this check

HimmelKreis4865: oh yeah i actually didn't realize `BlockBreakInfo->isToolCompatible()` would return true if harvest level is 0.

src/block/ChiseledBookshelf.php
```diff
@@ -0,0 +1,147 @@
...
+		if($item === null){
+			if(isset($this->items[$index])){
+				unset($this->items[$index]);
+			}
+			return;
+		}
+		$this->items[$index] = $item;
+	}
+
+	public function recalculateBooksBit() : int{
```
dktapps: This should not be public

src/block/ChiseledBookshelf.php
```diff
@@ -0,0 +1,147 @@
...
+		$bit = 0;
+		foreach($this->items as $index => $item){
+			$bit |= 1 << $index;
+		}
+		return $this->booksBit = $bit;
+	}
+
+	public function setCachedBooksBit(int $booksBit) : self{
```
dktapps: This is premature optimisation

src/block/ChiseledBookshelf.php
```diff
@@ -0,0 +1,147 @@
...
+	use FacesOppositePlacingPlayerTrait;
+
+	private int $booksBit = 0;
+
+	/** @var (WritableBookBase|Book)[] $items */
+	private array $items = [];
+
+	protected function describeBlockOnlyState(RuntimeDataDescriber $w) : void{
+		$w->horizontalFacing($this->facing);
+		$w->int(6, $this->booksBit);
```
dktapps: This should use 6 bools, one for each occupied slot

src/block/ChiseledBookshelf.php
dktapps: In addition, this is implicitly exposing Minecraft blockstate internals to the API

src/block/ChiseledBookshelf.php
```diff
@@ -0,0 +1,147 @@
...
+		if($tile instanceof TileChiseledBookshelf){
+			$tile->setItems($this->items);
+		}
+	}
+
+	public function getItem(int $index) : WritableBookBase|Book|null{
+		return $this->items[$index] ?? null;
+	}
+
+	public function setItem(int $index, WritableBookBase|Book|null $item) : void{
```
dktapps: This should be fluent for consistency with other setters

src/block/ChiseledBookshelf.php
HimmelKreis4865: It's only usecase is BlockToObjectDeserializer actually

dktapps: Then it should be encoded in there, not here. This internal stuff shouldn't be exposed to the API, and encoding 6 bool flags doesn't require a cache.

src/block/tile/TileFactory.php
```diff
@@ -59,6 +59,7 @@ public function __construct(){
 		$this->register(BrewingStand::class, ["BrewingStand", "minecraft:brewing_stand"]);
 		$this->register(Cauldron::class, ["Cauldron", "minecraft:cauldron"]);
 		$this->register(Chest::class, ["Chest", "minecraft:chest"]);
+		$this->register(ChiseledBookshelf::class, ["Chiseled Bookshelf", "minecraft:chiseled_bookshelf"]);
```
diamond-gold:
```suggestion
		$this->register(ChiseledBookshelf::class, ["ChiseledBookshelf", "minecraft:chiseled_bookshelf"]);
```
Vanilla ID does not contain space

HimmelKreis4865:
are all problems resolved?

src/block/ChiseledBookshelf.php
```diff
@@ -0,0 +1,145 @@
...
+	 */
+	public function getItems() : array{
+		return $this->items;
+	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if($face !== $this->getFacing()){
+			return false;
+		}
+		$x = (Facing::axis($face) === Axis::X ? $clickVector->getZ() : $clickVector->getX());
```
dktapps:
```suggestion
		$x = Facing::axis($face) === Axis::X ? $clickVector->getZ() : $clickVector->getX();
```

src/block/ChiseledBookshelf.php
```diff
@@ -0,0 +1,145 @@
...
+
+	/**
+	 * @return (WritableBookBase|Book)[]
+	 */
+	public function getItems() : array{
+		return $this->items;
+	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if($face !== $this->getFacing()){
```
dktapps:
```suggestion
		if($face !== $this->facing){
```
getter is not necessary here

src/block/ChiseledBookshelf.php
```diff
@@ -0,0 +1,145 @@
...
+		}
+	}
+
+	public function getItem(int $index) : WritableBookBase|Book|null{
+		return $this->items[$index] ?? null;
+	}
+
+	public function setItem(int $index, WritableBookBase|Book|null $item) : self{
+		if($item === null){
+			if(isset($this->items[$index])){
```
dktapps: useless `isset`

src/block/ChiseledBookshelf.php
```diff
@@ -0,0 +1,145 @@
...
+		}
+		$this->items[$index] = $item;
+		return $this;
+	}
+
+	/**
+	 * @param (WritableBookBase|Book)[] $items
+	 */
+	public function setItems(array $items) : self{
+		$this->items = $items;
```
dktapps: Multiple problems here:

- The number of items must be at most 6
- The indexes in the given array must not exceed 5 (`array_values()` probably OK for this)
- The array must be ordered
- The `Item`s in the array must be `clone`d

src/block/ChiseledBookshelf.php
```diff
@@ -0,0 +1,145 @@
...
+		return $this;
+	}
+
+	/**
+	 * @return (WritableBookBase|Book)[]
+	 */
+	public function getItems() : array{
+		return $this->items;
```
dktapps: These items need to be `clone`d

src/block/ChiseledBookshelf.php
```diff
@@ -0,0 +1,145 @@
...
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if($face !== $this->getFacing()){
+			return false;
+		}
+		$x = (Facing::axis($face) === Axis::X ? $clickVector->getZ() : $clickVector->getX());
+		$x = min(match($face){
+			Facing::NORTH, Facing::EAST => 1 - $x,
+			default => $x
+		}, 0.9);
```
dktapps: I feel like this variable would be more appropriately named `horizontal`.

Also, what's the deal with this 0.9?

src/block/ChiseledBookshelf.php
```diff
@@ -0,0 +1,145 @@
...
+				$item = $this->getItem($index) ?? VanillaItems::AIR();
+
+				$leftover = $player->getInventory()->addItem($item);
+				if(count($leftover) > 0){
+					$player->getWorld()->dropItem($player->getEyePos(), $leftover[0], $player->getDirectionVector()->multiply(0.2));
+				}
+				$this->setItem($index, null);
+			}
+		}else{
+			 if($this->getItem($index) instanceof Item){
```
dktapps:
```suggestion
			 if($this->getItem($index) !== null){
```

src/block/ChiseledBookshelf.php
```diff
@@ -0,0 +1,145 @@
...
+			return false;
+		}
+		$x = (Facing::axis($face) === Axis::X ? $clickVector->getZ() : $clickVector->getX());
+		$x = min(match($face){
+			Facing::NORTH, Facing::EAST => 1 - $x,
+			default => $x
+		}, 0.9);
+		$index = ($clickVector->y < 0.5 ? 3 : 0) + (int) floor($x * 3);
+		if(!$item instanceof WritableBookBase && !$item instanceof Book){
+			if($player instanceof Player){
```
dktapps:
```suggestion
			if($player !== null){
```

src/block/tile/ChiseledBookshelf.php
```diff
@@ -0,0 +1,82 @@
...
+	protected function addAdditionalSpawnData(CompoundTag $nbt) : void{
+
+	}
+
+	public function readSaveData(CompoundTag $nbt) : void{
+		$itemsTag = $nbt->getListTag(self::TAG_ITEMS);
+		if($itemsTag !== null && $itemsTag->getTagType() === NBT::TAG_Compound){
+			/** @var CompoundTag $itemNBT */
+			foreach($itemsTag->getValue() as $itemNBT){
+				/** @var WritableBookBase|Book $item */
```
dktapps: This is a bad idea. There's no guarantee that `Item::nbtDeserialize()` will actually return a book item.

src/data/bedrock/block/convert/BlockStateDeserializerHelper.php
```diff
@@ -373,4 +378,14 @@ public static function mapWoodenSlabType(BlockStateReader $in) : Slab{
 			default => throw $in->badValueException(BlockStateNames::WOOD_TYPE, $type),
 		};
 	}
+
+	public static function mapChiseledBookshelf(BlockStateReader $in) : Block{
+		$bit = $in->readInt(StateNames::BOOKS_STORED);
+		$items = array_filter(array_fill(0, 6, VanillaItems::BOOK()),
+			fn(Book $book, int $slot) => (($bit & (1 << $slot)) === (1 << $slot)),
```
dktapps:
```suggestion
			fn(Book $book, int $slot) => ($bit >> $slot) & 1 === 1,
```

src/data/bedrock/block/convert/BlockStateDeserializerHelper.php
```diff
@@ -373,4 +378,14 @@ public static function mapWoodenSlabType(BlockStateReader $in) : Slab{
...
+			ARRAY_FILTER_USE_BOTH);
```
dktapps: This array_* magic looks really weird and it's hard to understand what's going on. A simple `foreach` or `for` would be preferable.

src/data/bedrock/block/convert/BlockStateDeserializerHelper.php
```diff
@@ -373,4 +378,14 @@ public static function mapWoodenSlabType(BlockStateReader $in) : Slab{
 			default => throw $in->badValueException(BlockStateNames::WOOD_TYPE, $type),
 		};
 	}
+
+	public static function mapChiseledBookshelf(BlockStateReader $in) : Block{
+		$bit = $in->readInt(StateNames::BOOKS_STORED);
```
dktapps: `boundedInt` should be used here.

src/data/bedrock/block/convert/BlockStateDeserializerHelper.php
```diff
@@ -373,4 +378,14 @@ public static function mapWoodenSlabType(BlockStateReader $in) : Slab{
...
+		$items = array_filter(array_fill(0, 6, VanillaItems::BOOK()),
```
dktapps: The number of books should not be hardcoded

src/data/bedrock/block/convert/BlockStateSerializerHelper.php
```diff
@@ -288,4 +290,15 @@ public static function encodeWoodenSlab(Slab $block, string $typeValue) : BlockS
 		return self::encodeSlab($block, Ids::WOODEN_SLAB, Ids::DOUBLE_WOODEN_SLAB)
 			->writeString(BlockStateNames::WOOD_TYPE, $typeValue);
 	}
+
+	public static function encodeChiseledBookshelf(ChiseledBookshelf $block) : BlockStateWriter{
+		$slots = array_keys($block->getItems());
```
dktapps: This `array_keys()` calls is redundant. A `foreach($block->getItems() as $slot => $item)` can be used instead.

src/data/bedrock/block/convert/BlockStateSerializerHelper.php
```diff
@@ -288,4 +290,15 @@ public static function encodeWoodenSlab(Slab $block, string $typeValue) : BlockS
...
+		$bit = 0;
```
dktapps: `flags` would be a more idiomatic name.

src/block/tile/ChiseledBookshelf.php
```diff
@@ -0,0 +1,82 @@
...
+use pocketmine\item\WritableBookBase;
+use pocketmine\math\Vector3;
+use pocketmine\nbt\NBT;
+use pocketmine\nbt\tag\CompoundTag;
+use pocketmine\nbt\tag\ListTag;
+use pocketmine\world\World;
+use function array_keys;
+use function array_map;
+
+class ChiseledBookshelf extends Spawnable{
```
dktapps: Is it necessary for this to be `Spawnable`? Since we're not sending any special NBT data anyway, it ought to work fine with just a regular `Tile` as far as I can tell.

src/block/ChiseledBookshelf.php
```diff
@@ -0,0 +1,145 @@
...
+				$item = $this->getItem($index) ?? VanillaItems::AIR();
+
+				$leftover = $player->getInventory()->addItem($item);
+				if(count($leftover) > 0){
```
dktapps: The common method is to use a `foreach` for this. However, this entire section could likely be avoided by using `returnedItems` (which would also not require the `player`).

dktapps (dismissed):

src/block/ChiseledBookshelf.php
HimmelKreis4865: array_values will not work as it may skip book slots, why would the array need to be ordered?

src/block/ChiseledBookshelf.php
HimmelKreis4865: 0.9 is just a number between `0.66667 < x < 1` I used it so `floor($x * 3)` is scaled to be between 0-2

src/block/ChiseledBookshelf.php
dktapps: If we foreach over it, it may be implicitly expected.

dktapps: Actually, forget about the ordering thing - looks like this is a more widespread problem anyway

src/block/ChiseledBookshelf.php
```diff
@@ -0,0 +1,145 @@
...
+					$player->getWorld()->dropItem($player->getEyePos(), $leftover[0], $player->getDirectionVector()->multiply(0.2));
+				}
+				$this->setItem($index, null);
+			}
+		}else{
+			 if($this->getItem($index) instanceof Item){
+				 return false;
+			 }
+			 /** @var WritableBookBase|Book $item */
```
dktapps: Properly written code should not be using these within internals

src/block/ChiseledBookshelf.php
dktapps: Seems to me that `(int) (fmod($x, 1) * 3)` is what's needed here.

PR #5830, opened 2023-06-13
Accept translatable permission messages in `Command`

ShockedPlot7560:
## Introduction
Currently, if a plugin wishes to modify the message directly in the event of insufficient permission, it cannot pass an instance of Translatable.

### API changes
- Accept ``Translatable`` in ``permissionMessage``

## Tests

```php
class TestCommand extends Command{
	public function __construct(string $name, Translatable|string $description = "", Translatable|string|null $usageMessage = null, array $aliases = []){
		parent::__construct($name, $description, $usageMessage, $aliases);
		$permission = new Permission("test.command", "Allows you to use the test command");
		PermissionManager::getInstance()->addPermission($permission);
		$this->setPermission("test.command");
		$this->setPermissionMessage(KnownTranslationFactory::ability_noclip());
	}

	public function execute(CommandSender $sender, string $commandLabel, array $args){
		$sender->sendMessage("You have permission to use this command!");
	}
}

// Will send No-clip because we haven't the permission
```

### DIFF
--- a/src/command/Command.php
+++ b/src/command/Command.php
@@ -112,13 +112,12 @@ abstract class Command{
 			return true;
 		}
 
+		$message = null;
 		if($this->permissionMessage instanceof Translatable || $this->permissionMessage === null){
 			$message = $this->permissionMessage ?? KnownTranslationFactory::pocketmine_command_error_permission($this->name);
 			$message = $message->prefix(TextFormat::RED);
 		}elseif($this->permissionMessage !== ""){
 			$message = str_replace("<permission>", $permission ?? implode(";", $this->permission), $this->permissionMessage);
-		}else{
-			$message = null;
 		}
 
 		if($message !== null){
--- a/src/command/Command.php
+++ b/src/command/Command.php
@@ -114,9 +114,15 @@ abstract class Command{
 
 		if($this->permissionMessage instanceof Translatable || $this->permissionMessage === null){
 			$message = $this->permissionMessage ?? KnownTranslationFactory::pocketmine_command_error_permission($this->name);
-			$target->sendMessage($message->prefix(TextFormat::RED));
+			$message = $message->prefix(TextFormat::RED);
 		}elseif($this->permissionMessage !== ""){
-			$target->sendMessage(str_replace("<permission>", $permission ?? implode(";", $this->permission), $this->permissionMessage));
+			$message = str_replace("<permission>", $permission ?? implode(";", $this->permission), $this->permissionMessage);
+		}else{
+			$message = null;
+		}
+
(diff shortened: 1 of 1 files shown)


src/command/Command.php
```diff
@@ -112,8 +112,9 @@ public function testPermission(CommandSender $target, ?string $permission = null
 			return true;
 		}
 
-		if($this->permissionMessage === null){
-			$target->sendMessage(KnownTranslationFactory::pocketmine_command_error_permission($this->name)->prefix(TextFormat::RED));
+		if($this->permissionMessage instanceof Translatable || $this->permissionMessage === null){
```
dktapps: This logic looks cursed. It would be better to move the message variable outside the condition and use it in all the branches.

src/command/Command.php
```diff
@@ -187,7 +188,7 @@ public function getAliases() : array{
 		return $this->activeAliases;
 	}
 
-	public function getPermissionMessage() : ?string{
+	public function getPermissionMessage() : null|string|Translatable{
```
dktapps: This is BC-breaking, it can't target a minor version.

src/command/Command.php
```diff
@@ -213,7 +214,7 @@ public function setDescription(Translatable|string $description) : void{
 		$this->description = $description;
 	}
 
-	public function setPermissionMessage(string $permissionMessage) : void{
+	public function setPermissionMessage(string|Translatable $permissionMessage) : void{
```
dktapps: This is BC-breaking.

src/command/Command.php
dktapps: Not really what I meant ...

ShockedPlot7560: Huh ? What do you mean by moving outside the conditions? I don't think I understand then.

dktapps: `$message = $this->permissionMessage ?? KnownTranslationFactory::whatever(...)`
then check for string or Translatable instance

src/command/Command.php
dktapps: This should be `Translatable|string|null` for consistency.

src/command/Command.php
dktapps: `Translatable|string` would be more consistent

src/command/Command.php
```diff
@@ -58,7 +58,7 @@ abstract class Command{
 
 	/** @var string[] */
 	private array $permission = [];
-	private ?string $permissionMessage = null;
+	private null|string|Translatable $permissionMessage = null;
```
dktapps: `Translatable|string|null` would be more consistent

PR #5835, opened 2023-06-14
Implemented Big & Small dripleaf

ipad54:
## Introduction
Implemented Big & Small dripleaf

## Backwards compatibility
BC Break change is in RuntimeEnumDescriber interface, but i think it's not a part of API?

## Tests
https://www.youtube.com/watch?v=n94ewbNj_xw


dktapps:
> BC Break change is in RuntimeEnumDescriber interface, but i think it's not a part of API?

Yeah, this is a design flaw I didn't think of when I initially wrote that code, but users aren't likely to implement it anyway. It should probably be explicitly marked @internal though.

src/block/BigDripleaf.php
```diff
@@ -0,0 +1,233 @@
...
+		$w->bool($this->head);
+	}
+
+	/** @return $this */
+	public function setTilt(DripleafTiltType $tilt) : self{
+		$this->tilt = $tilt;
+		return $this;
+	}
+
+	public function getTilt() : DripleafTiltType{
```
dktapps: nit: getters are usually placed before setters

src/block/BigDripleaf.php
```diff
@@ -0,0 +1,233 @@
...
+		return $this->tilt;
+	}
+
+	/** @return $this */
+	public function setHead(bool $head) : self{
+		$this->head = $head;
+		return $this;
+	}
+
+	public function isHead() : bool{
```
dktapps: same again here

src/block/BigDripleaf.php
```diff
@@ -0,0 +1,233 @@
...
+		return
+			$block->hasSameTypeId($this) ||
+			$block->getTypeId() === BlockTypeIds::CLAY ||
+			$block->hasTypeTag(BlockTypeTags::DIRT) ||
+			$block->hasTypeTag(BlockTypeTags::MUD);
+	}
+
+	public function onNearbyBlockChange() : void{
+		if(
+			!$this->head && !$this->getSide(Facing::UP)->hasSameTypeId($this) ||
```
dktapps: Potential operator precedence issue here, I assume the below is the correct solution?
```suggestion
			(!$this->head && !$this->getSide(Facing::UP)->hasSameTypeId($this)) ||
```

src/block/BigDripleaf.php
```diff
@@ -0,0 +1,233 @@
...
+		if(
+			!$world->isInWorld($up->getFloorX(), $up->getFloorY(), $up->getFloorZ()) ||
+			$world->getBlock($up)->getTypeId() !== BlockTypeIds::AIR
+		){
+			return false;
+		}
+
+		$tx = new BlockTransaction($world);
+
+		$tx->addBlock($pos, (clone $head)->setHead(false));
```
dktapps: what about the tilt?

I'm getting the impression that BIG_DRIPLEAF_STEM ought to be separated from BIG_DRIPLEAF, since the tilt property does not apply to the stem.

src/block/BigDripleaf.php
```diff
@@ -0,0 +1,233 @@
...
+		$tx->addBlock($up, VanillaBlocks::BIG_DRIPLEAF()
+			->setFacing($head->getFacing())
+			->setHead(true)
+		);
```
dktapps: Again, looks like the tilt has not been considered here

src/block/BigDripleaf.php
```diff
@@ -0,0 +1,233 @@
...
+		if(!$this->head || $this->tilt->equals(DripleafTiltType::FULL())){
+			return [];
+		}
+		return [
+			AxisAlignedBB::one()
+			->trim(Facing::DOWN, 11 / 16)
+			->trim(Facing::UP, match($this->tilt->id()) {
+				DripleafTiltType::NONE()->id(), DripleafTiltType::UNSTABLE()->id() => 1 / 16,
+				DripleafTiltType::PARTIAL()->id() => 3 / 16,
+				default => throw new AssumptionFailedError("All types should be covered")
```
dktapps: This block below `AxisAlignedBB::one()` needs to be indented by 1

src/block/BigDripleaf.php
```diff
@@ -0,0 +1,233 @@
...
+			})
+		];
+	}
+
+	public function isFlammable() : bool{
+		return true;
+	}
```
dktapps: The flame encouragement and flammability values should be set instead of overriding this - this is not supposed to be overridden

A Bedrock data dump suggests that the flammability is `100` and the flame encouragement is `15`

src/block/SmallDripleaf.php
```diff
@@ -0,0 +1,154 @@
...
+use pocketmine\math\Facing;
+use pocketmine\math\Vector3;
+use pocketmine\player\Player;
+use pocketmine\world\BlockTransaction;
+use function mt_rand;
+
+class SmallDripleaf extends Transparent{
+	use HorizontalFacingTrait;
+
+	protected bool $upperBlock = false;
```
dktapps: `top` would be the idiomatic way of naming this, for consistency with other blocks such as double plants and doors

dktapps: Perhaps this class can use `DoublePlant` as a base? Not an instruction, but a suggestion

src/block/SmallDripleaf.php
```diff
@@ -0,0 +1,154 @@
...
+			$pos = $bottomBlock->getSide(Facing::UP, $i)->getPosition();
+			if(!$world->isInWorld($pos->getFloorX(), $pos->getFloorY(), $pos->getFloorZ())){
+				break;
+			}
+			$replace = $world->getBlock($pos);
+			if(!$replace->hasSameTypeId($this) && $replace->getTypeId() !== BlockTypeIds::AIR){
+				break;
+			}
+			$tx->addBlock($pos, VanillaBlocks::BIG_DRIPLEAF()->setFacing($this->facing));
+			$growed++;
```
dktapps: `grown` would be more idiomatic

src/block/SmallDripleaf.php
```diff
@@ -0,0 +1,154 @@
...
+		$bottomBlock = $this->upperBlock ? $this->getSide(Facing::DOWN) : $this;
+		if(!$this->hasSameTypeId($bottomBlock)){
+			return false;
+		}
+		$top = $bottomBlock->getPosition();
+		$world = $top->getWorld();
+		$tx = new BlockTransaction($world);
+		$growed = 0;
+
+		for($i = 0; $i < mt_rand(2, 5); $i++){
```
dktapps: This is going to re-run `mt_rand()` for every loop iteration - probably not what you wanted

src/block/SmallDripleaf.php
```diff
@@ -0,0 +1,154 @@
...
+			}
+			$tx->addBlock($pos, VanillaBlocks::BIG_DRIPLEAF()->setFacing($this->facing));
+			$growed++;
+			$top = $pos;
+		}
+
+		if($growed > 1){
+			$block = $tx->fetchBlock($top);
+			if($block instanceof BigDripleaf && $block->getTypeId() === BlockTypeIds::BIG_DRIPLEAF){
+				$block->setHead(true);
```
dktapps: This logic feels a bit weird

src/data/bedrock/block/convert/BlockStateReader.php
```diff
@@ -349,6 +350,17 @@ public function readWallConnectionType(string $name) : ?WallConnectionType{
...
+	/** @throws BlockStateDeserializeException */
+	public function readDripleafTiltType() : DripleafTiltType{
+		return match($type = $this->readString(BlockStateNames::BIG_DRIPLEAF_TILT)){
+			StringValues::BIG_DRIPLEAF_TILT_NONE => DripleafTiltType::NONE(),
+			StringValues::BIG_DRIPLEAF_TILT_UNSTABLE => DripleafTiltType::UNSTABLE(),
+			StringValues::BIG_DRIPLEAF_TILT_PARTIAL_TILT => DripleafTiltType::PARTIAL(),
+			StringValues::BIG_DRIPLEAF_TILT_FULL_TILT => DripleafTiltType::FULL(),
+			default => throw $this->badValueException(BlockStateNames::BIG_DRIPLEAF_TILT, $type),
+		};
+	}
```
dktapps: I'm not really clear what we're gaining by putting this here, as opposed to in the `BlockStateToObjectDeserializer` directly. It's only used once.

src/data/bedrock/block/convert/BlockStateWriter.php
```diff
@@ -318,6 +319,18 @@ public function writeWallConnectionType(string $name, ?WallConnectionType $wallC
...
+	public function writeDripleafTiltType(DripleafTiltType $tiltType) : self{
+		$this->writeString(BlockStateNames::BIG_DRIPLEAF_TILT, match($tiltType->id()){
+			DripleafTiltType::NONE()->id() => StringValues::BIG_DRIPLEAF_TILT_NONE,
+			DripleafTiltType::UNSTABLE()->id() => StringValues::BIG_DRIPLEAF_TILT_UNSTABLE,
+			DripleafTiltType::PARTIAL()->id() => StringValues::BIG_DRIPLEAF_TILT_PARTIAL_TILT,
+			DripleafTiltType::FULL()->id() => StringValues::BIG_DRIPLEAF_TILT_FULL_TILT,
+			default => throw new BlockStateSerializeException("Invalid Dripleaf tilt type " . $tiltType->name())
+		});
+		return $this;
+	}
```
dktapps: Same again here.

dktapps (changes requested):
I haven't yet tested this, but I have a few thoughts about the design of this PR.

Most notably, I think that `BIG_DRIPLEAF` needs to be split into stem vs. head parts, because the head has properties (tilt) which are useless to the stems.

src/block/SmallDripleaf.php
ipad54: Double Plants are flowable blocks but small dripleaf is not

> Perhaps this class can use `DoublePlant` as a base? Not an instruction, but a suggestion

src/block/BaseBigDripleaf.php
```diff
@@ -0,0 +1,136 @@
...
+			!$world->isInWorld($up->getFloorX(), $up->getFloorY(), $up->getFloorZ()) ||
+			$world->getBlock($up)->getTypeId() !== BlockTypeIds::AIR
+		){
+			return false;
+		}
+
+		$tx = new BlockTransaction($world);
+
+		$tx->addBlock($pos, VanillaBlocks::BIG_DRIPLEAF_STEM()->setFacing($head->getFacing()));
+		$tx->addBlock($up, VanillaBlocks::BIG_DRIPLEAF()->setFacing($head->getFacing()));
```
dktapps: what about the tilt?

src/block/BaseBigDripleaf.php
```diff
@@ -0,0 +1,136 @@
...
+
+abstract class BaseBigDripleaf extends Transparent{
+	use HorizontalFacingTrait;
+
+	abstract protected function isHead() : bool;
+
+	private function canBeSupportedBy(Block $block) : bool{
+		//TODO: Moss block
+		return
+			$block instanceof BaseBigDripleaf ||
```
dktapps: I don't think this is correct. Both stems and heads should only be able to be supported by other stems. A dripleaf head should not be able to provide support to more dripleaf.

src/block/BaseBigDripleaf.php
ipad54: In vanilla tilt restores to NONE when block grows

src/data/bedrock/block/convert/BlockStateToObjectDeserializer.php
```diff
@@ -784,6 +786,26 @@ private function registerDeserializers() : void{
...
 				->setAttachmentType($in->readBellAttachmentType());
 		});
+		$this->map(Ids::BIG_DRIPLEAF, function(Reader $in) : Block{
+			$block =
+				($head = $in->readBool(StateNames::BIG_DRIPLEAF_HEAD)) ?
+					Blocks::BIG_DRIPLEAF() :
+					Blocks::BIG_DRIPLEAF_STEM();
+			$block->setFacing($in->readLegacyHorizontalFacing());
+			if($head){
+				/** @var BigDripleaf $block */
```
dktapps: This shouldn't be needed if the code is properly written.

src/block/BaseBigDripleaf.php
dktapps: honestly this sounds like a bug. have you tested it in java?

dktapps: honestly this sounds like a bug. have you tested in Java?

ipad54: Tested it on Java a minute ago, Java seems to be also work this way.

https://www.youtube.com/watch?v=3tPl22WUuXE

dktapps: Fair enough, I guess this can be chalked up to another Mojang inconsistency.

src/block/SmallDripleaf.php
dktapps: > Double Plants are flowable blocks but small dripleaf is not

Right, I didn't think of that.

src/data/bedrock/block/convert/BlockStateToObjectDeserializer.php
ipad54: PHPStan says ```Call to an undefined method pocketmine\block\BigDripleaf|pocketmine\block\BigDripleafStem::setTilt(). ```

src/block/BaseBigDripleaf.php
```diff
@@ -38,10 +38,10 @@ abstract class BaseBigDripleaf extends Transparent{
 
 	abstract protected function isHead() : bool;
 
-	private function canBeSupportedBy(Block $block) : bool{
+	private function canBeSupportedBy(Block $block, bool $head = false) : bool{
```
dktapps: Avoid optional parameters.

src/data/bedrock/block/convert/BlockStateToObjectDeserializer.php
dktapps: As I said, the code is improperly written.

src/block/BaseBigDripleaf.php
```diff
@@ -0,0 +1,136 @@
...
+
+abstract class BaseBigDripleaf extends Transparent{
+	use HorizontalFacingTrait;
+
+	abstract protected function isHead() : bool;
+
+	private function canBeSupportedBy(Block $block, bool $head) : bool{
+		//TODO: Moss block
+		return
+			$block instanceof BaseBigDripleaf && $block->isHead() === $head ||
```
dktapps: Precedence issue here
```suggestion
			($block instanceof BaseBigDripleaf && $block->isHead() === $head) ||
```

src/world/format/io/data/BedrockWorldData.php
```diff
@@ -155,18 +154,12 @@ protected function load() : CompoundTag{
...
-		}
 		if($version > self::CURRENT_STORAGE_VERSION){
 			throw new UnsupportedWorldFormatException("LevelDB world format version $version is currently unsupported");
 		}
 		//StorageVersion is rarely updated - instead, the game relies on the NetworkVersion tag, which is synced with
 		//the network protocol version for that version.
 		$protocolVersion = $worldData->getInt(self::TAG_NETWORK_VERSION, Limits::INT32_MAX);
-		if($protocolVersion === Limits::INT32_MAX){
-			throw new CorruptedWorldException(sprintf("Missing '%s' tag in level.dat", self::TAG_NETWORK_VERSION));
-		}
```
dktapps: Why is this changed?

ipad54: I didn't change it, maybe it's a github bug

dktapps: Diff suggests it came from this commit: https://github.com/pmmp/PocketMine-MP/pull/5835/commits/8d7023f9ad7ac9c72b09b55bb09cfce4f33d2ecf

ipad54: Should be fixed now

dktapps (approved):

Issue #5169, opened 2022-07-16
Find a way to properly integrate particle events which are actually particle+sound events


dktapps:
## Description
In many places, Minecraft uses so-called "level events" to trigger certain sound or visual effects. This is useful because some effects are quite complex (involving many particles, and/or arranged in particular ways), and it would be a colossal pain in the ass to have to reverse-engineer the logic for these things, as well as being wasteful for bandwidth usage.

However, the PocketMine-MP API currently doesn't have any place for these sound+particle combination events. This has forced the direct use of packets in various places, and in other cases the misclassification of certain things (e.g. BlockBreakParticle).

In addition, there are various graphical events (block crack animations, weather change) which are neither sound _or_ particle, and yet still need to be broadcasted.

We need an API for these things that properly describes what they do. However...

We already have an events system, so `WorldEvent` is too ambiguous, because it could be conflated with the actual `WorldEvent` (you know, the one from which `ChunkLoadEvent` and friends inherit from). However, if we could unchain the event registration system from plugins, we could implement these events as actual events, allowing plugins to cancel or modify them, and allowing NetworkSessions to subscribe to (until now, plugin-exclusive) events.

## Alternative methods
Add something like `pocketmine/world/soundAndParticle/` package? meh...

Govdim:
Maybe `pocketmine/world/levelevent`?
Or in this case, the "level" and the "world" are the same thing? :/

Then the event class will be called `WorldLevelEventEvent` :D
Well, or `WorldLevelEvent`

dktapps:
I think the name is too generic. What we're really dealing with is combination sfx + vfx events. We need a name that reflects that.

Govdim:
> I think the name is too generic. What we're really dealing with is combination sfx + vfx events. We need a name that reflects that.

Here are some options:

1. `SoundParticleEffect`
2. `SoundParticleFusion`
3. `SoundParticle`
Then the package name will be `pocketmine/world/soundparticle`

PR #5859, opened 2023-06-25
Implemented BoneMealParticle

Govdim:
## Introduction
Implemented BoneMealParticle

## Tests
https://youtu.be/042EgTswmVc


NhanAZ:
I think you should take a look at this plugin
https://github.com/nhanaz-pm-pl/CropGrowth
This PR needs to do more things

Govdim:
> I think you should take a look at this plugin https://github.com/nhanaz-pm-pl/CropGrowth This PR needs to do more things

This PR implements only the effect of using bone meal, and not the logic of growing plants.

src/block/Dirt.php
```diff
@@ -79,7 +80,7 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
 
 			$item->pop();
 			$world->setBlock($down->position, VanillaBlocks::HANGING_ROOTS());
-			//TODO: bonemeal particles, growth sounds
+			$world->addParticle($this->position, new BoneMealParticle());
```
dries-c, line 83: You'll need to leave the todo for the sounds, since you didn't integrate those

dries-c (changes requested):

src/block/Dirt.php
Govdim: BoneMealParticle is both a particle and a sound. I didn’t hear any other sounds in vanilla and there is no information about this on the wiki.

PrograMistV1: > I didn’t hear any other sounds in vanilla

Isn't the sound of using bonemeal being `item.bone_meal.use`?
and yes, this is the only sound that is used in growing and does not need to be integrated

dktapps: I'm not a fan of adding more dual particle/sound events that masquerade as particles. We already have problems with BlockBreakParticle because of this.

#5169 needs to be dealt with.

ShockedPlot7560:
Waiting for a solution of #5169

Issue #5767, opened 2023-05-26
Ability to get cause in EntityTeleportEvent


MINTEDCLICKS:
You should be able to get the cause of EntityTeleportEvent

It allows one to easily detect how a player was teleported to an area which is helpful in some cases.

There are alternative methods but it just seems like this should be in.

PR #5861, opened 2023-06-27
Add ability to get cause in EntityTeleportEvent (#5767)

Nyrok:
### Relevant issues

#5767

### API changes

The following method now accepts `int $cause` parameter. Defaults is 0 for EntityTeleportEvent::CAUSE_PLUGIN.
```php
Entity->teleport(Vector3 $pos, ?float $yaw, ?float $pitch, int $cause) : bool;
```

## Backwards compatibility

No BC-breaking changes, because there is a default parameter to this method signature change.

### DIFF
--- a/src/command/defaults/TeleportCommand.php
+++ b/src/command/defaults/TeleportCommand.php
@@ -27,6 +27,7 @@ use pocketmine\command\Command;
 use pocketmine\command\CommandSender;
 use pocketmine\command\utils\InvalidCommandSyntaxException;
 use pocketmine\entity\Location;
+use pocketmine\event\entity\EntityTeleportEvent;
 use pocketmine\lang\KnownTranslationFactory;
 use pocketmine\permission\DefaultPermissionNames;
 use pocketmine\player\Player;
@@ -89,7 +90,7 @@ class TeleportCommand extends VanillaCommand{
 					return true;
 				}
 
-				$subject->teleport($targetPlayer->getLocation());
+				$subject->teleport($targetPlayer->getLocation(), null, null, EntityTeleportEvent::CAUSE_COMMAND);
 				Command::broadcastCommandMessage($sender, KnownTranslationFactory::commands_tp_success($subject->getName(), $targetPlayer->getName()));
 
 				return true;
@@ -109,7 +110,7 @@ class TeleportCommand extends VanillaCommand{
 				$z = $this->getRelativeDouble($base->z, $sender, $args[2]);
 				$targetLocation = new Location($x, $y, $z, $base->getWorld(), $yaw, $pitch);
 
-				$subject->teleport($targetLocation);
+				$subject->teleport($targetLocation, null, null, EntityTeleportEvent::CAUSE_COMMAND);
 				Command::broadcastCommandMessage($sender, KnownTranslationFactory::commands_tp_success_coordinates(
 					$subject->getName(),
 					(string) round($targetLocation->x, 2),
--- a/src/command/defaults/TeleportCommand.php
+++ b/src/command/defaults/TeleportCommand.php
@@ -90,7 +90,7 @@ class TeleportCommand extends VanillaCommand{
 					return true;
 				}
 
-				$subject->teleport($targetPlayer->getLocation(), null, null, EntityTeleportEvent::CAUSE_COMMAND);
+				$subject->teleport($targetPlayer->getLocation(), cause: EntityTeleportEvent::CAUSE_COMMAND);
 				Command::broadcastCommandMessage($sender, KnownTranslationFactory::commands_tp_success($subject->getName(), $targetPlayer->getName()));
 
 				return true;
@@ -110,7 +110,7 @@ class TeleportCommand extends VanillaCommand{
 				$z = $this->getRelativeDouble($base->z, $sender, $args[2]);
 				$targetLocation = new Location($x, $y, $z, $base->getWorld(), $yaw, $pitch);
 
-				$subject->teleport($targetLocation, null, null, EntityTeleportEvent::CAUSE_COMMAND);
+				$subject->teleport($targetLocation, cause: EntityTeleportEvent::CAUSE_COMMAND);
 				Command::broadcastCommandMessage($sender, KnownTranslationFactory::commands_tp_success_coordinates(
 					$subject->getName(),
 					(string) round($targetLocation->x, 2),
--- a/src/entity/projectile/EnderPearl.php
+++ b/src/entity/projectile/EnderPearl.php
@@ -24,6 +24,7 @@ declare(strict_types=1);
 namespace pocketmine\entity\projectile;
 
 use pocketmine\event\entity\EntityDamageEvent;
+use pocketmine\event\entity\EntityTeleportEvent;
 use pocketmine\event\entity\ProjectileHitEvent;
 use pocketmine\network\mcpe\protocol\types\entity\EntityIds;
 use pocketmine\world\particle\EndermanTeleportParticle;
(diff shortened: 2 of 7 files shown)


Nyrok:
In my IDE, the [code style](https://github.com/pmmp/PocketMine-MP/actions/runs/5384277415/jobs/9771993759?pr=5861) looks corrects, if someone can improve this, it should be better.

src/event/entity/EntityTeleportEvent.php
```diff
@@ -35,10 +35,18 @@
 class EntityTeleportEvent extends EntityEvent implements Cancellable{
 	use CancellableTrait;
 
+    public const CAUSE_PLUGIN = 0;
+    public const CAUSE_ITEM = 1;
+    public const CAUSE_WORLD = 2;
+    public const CAUSE_RESPAWN = 3;
+    public const CAUSE_PATCH = 4;
```
Nyrok: Maybe we can call this CAUSE_SYNC ?

Nyrok (commented):
Suggestions

src/event/entity/EntityTeleportEvent.php
```diff
@@ -35,10 +35,18 @@
 class EntityTeleportEvent extends EntityEvent implements Cancellable{
 	use CancellableTrait;
 
+	public const CAUSE_PLUGIN = 0;
+	public const CAUSE_ITEM = 1;
+	public const CAUSE_WORLD = 2;
+	public const CAUSE_RESPAWN = 3;
+	public const CAUSE_PATCH = 4;
```
Nyrok, line 42: ddd

dktapps: This naming is very unintuitive, I have no idea what it's supposed to mean

src/entity/projectile/EnderPearl.php
```diff
@@ -40,7 +41,7 @@ protected function onHit(ProjectileHitEvent $event) : void{
 
 			$this->getWorld()->addParticle($origin = $owner->getPosition(), new EndermanTeleportParticle());
 			$this->getWorld()->addSound($origin, new EndermanTeleportSound());
-			$owner->teleport($target = $event->getRayTraceResult()->getHitVector());
+			$owner->teleport($target = $event->getRayTraceResult()->getHitVector(), null, null, EntityTeleportEvent::CAUSE_ITEM);
```
JavierLeon9966: `CAUSE_PROJECTILE` seems better

src/item/ChorusFruit.php
```diff
@@ -76,7 +77,7 @@ public function onConsume(Living $consumer) : void{
 
 			//Sounds are broadcasted at both source and destination
 			$world->addSound($origin, new EndermanTeleportSound());
-			$consumer->teleport($target = new Vector3($x + 0.5, $y + 1, $z + 0.5));
+			$consumer->teleport($target = new Vector3($x + 0.5, $y + 1, $z + 0.5), null, null, EntityTeleportEvent::CAUSE_ITEM);
```
JavierLeon9966: `CAUSE_CHORUS_FRUIT` seems better

src/event/entity/EntityTeleportEvent.php
Nyrok: - Item: this is when an item cause the teleportation, like chorus fruit or enderpearl
- World: this is when the world cause teleportation like changing world or being teleported in safe spawn due to world unload
- Respawn: this is when the player respawn cause teleportation
- Patch: I think we can merge it with "Respawn" as "Player"
So, I can suggest this changes:
```diff
	public const CAUSE_ITEM = 1;
	public const CAUSE_WORLD = 2;
-	public const CAUSE_RESPAWN = 3;
-	public const CAUSE_PATCH = 4;
+	public const CAUSE_PLAYER = 3;
```

src/player/Player.php
```diff
@@ -1339,7 +1340,7 @@ protected function processMostRecentMovements() : void{
 			}
 
 			if($to->distanceSquared($ev->getTo()) > 0.01){ //If plugins modify the destination
-				$this->teleport($ev->getTo());
+				$this->teleport($ev->getTo(), null, null, EntityTeleportEvent::CAUSE_PATCH);
```
JavierLeon9966: `CAUSE_SYNCHRONIZE` seems better

JavierLeon9966 (changes requested):

Nyrok:
This is specific, but let's see what @dktapps think about it

src/event/entity/EntityTeleportEvent.php
dktapps: This is even less clear than the initial, and additionally provides less information.

src/player/Player.php
dktapps: CAUSE_PLUGIN would be sufficient.

src/command/defaults/TeleportCommand.php
```diff
@@ -89,7 +90,7 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
 					return true;
 				}
 
-				$subject->teleport($targetPlayer->getLocation());
+				$subject->teleport($targetPlayer->getLocation(), null, null, EntityTeleportEvent::CAUSE_COMMAND);
```
ShockedPlot7560:
```suggestion
				$subject->teleport($targetPlayer->getLocation(), cause: EntityTeleportEvent::CAUSE_COMMAND);
```
Seems better

src/event/entity/EntityTeleportEvent.php
```diff
@@ -51,6 +59,14 @@ public function getTo() : Position{
 		return $this->to;
 	}
 
+	public function getCause() : int{
+		return $this->cause;
+	}
+
+	public function setCause(int $cause) : void{
```
JavierLeon9966, line 66: Why do we need this?

Nyrok: Maybe it can be used for plugin conflicts 🤷‍♂️

src/event/entity/EntityTeleportEvent.php
```diff
@@ -36,10 +36,10 @@ class EntityTeleportEvent extends EntityEvent implements Cancellable{
 	use CancellableTrait;
 
 	public const CAUSE_PLUGIN = 0;
-	public const CAUSE_ITEM = 1;
+	public const CAUSE_PROJECTILE = 1;
```
dzolnowski, line 39: CAUSE_PROJECTILE doesn't tell us what type of projectile was used. Maybe CAUSE_ENDER_PEARL would be better?

JavierLeon9966: You know, I think the cause should be a string so plugins can specify their custom causes like "morepearls:snowball". That way we can be specific while also allowing customization

dzolnowski: Yeah agree, with ints its really easy to mess up if someone was to create custom one

Nyrok:
What's now ? Should be the reason a string or a int (like seeing PlayerPreLoginEvent) ?

ShockedPlot7560:
> What's now ? Should be the reason a string or a int (like seeing PlayerPreLoginEvent) ?

I think a string would allow greater flexibility. An integer is very restrictive and doesn't give direct information for plugins that want to get the values.

dktapps:
> > What's now ? Should be the reason a string or a int (like seeing PlayerPreLoginEvent) ?
> 
> I think a string would allow greater flexibility. An integer is very restrictive and doesn't give direct information for plugins that want to get the values.

Plugins won't be able to interpret values they don't know about anyway. I think an integer is fine.

Really this should probably use an `enum`, but I don't want to introduce API inconsistencies.

src/event/entity/EntityTeleportEvent.php
```diff
@@ -35,10 +35,18 @@
 class EntityTeleportEvent extends EntityEvent implements Cancellable{
 	use CancellableTrait;
 
+	public const CAUSE_PLUGIN = 0;
+	public const CAUSE_PROJECTILE = 1;
+	public const CAUSE_WORLD = 2;
```
dktapps, line 40: I have no idea what this is supposed to mean.

PR #5863, opened 2023-06-29
Fix for #4680 - /kill not actually killing

### LINKED ISSUE #4680: /kill command doesn’t kill the player directly

jasonw4331:
## Introduction

Fixes /kill not functioning when player had more than 1000 health

### Relevant issues

- Fixes #4680
- PR #4801
### API changes

N/A

### Behavioral changes

/kill will now *always* kill the player

## Backwards compatibility

N/A

## Follow-up

N/A

## Tests

In-game testing with absorption 255 + regen 255 + health boost 255 + resistance 255 results in player death


dktapps:
My sticking point for this was that PHP_INT_MAX isn't actually the largest float value, and is also not properly representable as a float. There's also the question of what will happen when EntityDamageEvent sees such a large value, and the effects it will have on the damage mitigation calculations (armour, enchantments etc).

jasonw4331:
The absorption and armor enchantments modifiers are affected by this change. When a floating point underflow occurs in php, the resulting value is 0. This means the armor enchantments modifier will always be 0 and absorption will be restricted to max `340282346638528859811704183484516925440.00` which is always below the float64 max thereby preventing the absorption modifier from cancelling out the initial damage value.

TLDR: The player will always die using this value and there are no side effects to be seen because attribute values are recalculated on respawn.

src/command/defaults/KillCommand.php
```diff
@@ -53,7 +54,7 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
 			return true;
 		}
 
-		$player->attack(new EntityDamageEvent($player, EntityDamageEvent::CAUSE_SUICIDE, 1000));
+		$player->attack(new EntityDamageEvent($player, EntityDamageEvent::CAUSE_SUICIDE, PHP_FLOAT_MAX));
```
SOF3, line 57: why not just INF?

NhanAZ, line 57: works?
```suggestion
		$player->setHealth(0);
		$player->attack(new EntityDamageEvent($player, EntityDamageEvent::CAUSE_SUICIDE, 1000));
```

JavierLeon9966: It wouldn't make sense if the damage event gets called after the entity died. And might as well put the damage as `0`. SOFe's idea seems better. Just use `INF` damage

NhanAZ: What do you think about how PowerNukkit does it
https://github.com/PowerNukkit/PowerNukkit/blob/master/src/main/java/cn/nukkit/command/defaults/KillCommand.java#L50-L56

JavierLeon9966: Infinite damage can do just that

dktapps: I'm dubious if PHP_FLOAT_MAX will actually work and not break anything, never mind INF.

jasonw4331: This change will cause the command to be effective against creative players which is not intended.

jasonw4331: Using `INF` will cause the armor enchantments modifier to equal `-INF`. I haven't tested, but I believe this will cancel out the damage altogether when added to `INF` base damage.

dktapps: Yeah, INF is way too unpredictable. If you can verify that it works with PHP_FLOAT_MAX, I think that's tolerable, but my gut is telling me something is going to break with these insane damage amounts.

dktapps: /kill already works on Creative players in PM (which I don't intend to change, as it's useful for returning to the spawnpoint). However, setHealth(0) calls kill(), which would prevent plugins from cancelling the event.

PR #5864, opened 2023-06-29
PluginBase Supported Translation API

jasonw4331:
## Introduction

If plugins wish to use the existing translations system, they must reflect into the `Server::getLanguage()` instance and add their translations directly, then call to `Language::translate()` before passing the output to `Player::sendMessage()` or setting the death event message (*both accept a `Translatable` already*).

There are approximately 100 total plugins on poggit at time of writing which implement a translation system of some sort. 41 of those plugins use the pocketmine built-in Language class with reflections to add their translations into the core. I think if ~40% of the publicly known plugins use this system, then adding an API to improve it's ease of use should be considered.

My opinion stands that PluginBase should have useful utilities like this to assist with ease of development, but developers may still opt not to use this feature.

### Relevant issues

- Relates to #3998
- PR #5710

### API changes

Adds `NamespacedLanguage` to keep track of Languages using namespaces
Adds `Language::merge()` to import translations from other Language objects
Adds `PluginBase::addTranslation()` to allow plugins to craft translations onEnable instead of by resource
Adds `PluginBase::getTranslations()` to return all registered `NamespacedLanguage` objects

### Behavioral changes

`Language::getLanguageList()` now supports reading phar resources
`NetworkSession::onChatMessage()` automatically translates `Translatable` if the namespace is registered to the server Language instance.
`DeathPacketHandler` automatically translates `Translatable` if the namespace is registered to the server Language instance.

## Backwards compatibility

BC 100%

## Follow-up

TODO

## Tests

Before:
```php
// onEnable
$languageAliases = ['eng', 'fre'];
foreach($languageAliases as $alias){
	$this->saveResource('/languages/' . $alias . '.ini', true);
}

$dir = scandir(Path::join($this->getDataFolder(), "languages"));
if($dir !== false){
	foreach($dir as $file){
		/** @phpstan-var array{dirname: string, basename: string, extension?: string, filename: string} $fileData */
		$fileData = pathinfo($file);
		if(!isset($fileData["extension"]) || $fileData["extension"] !== "ini"){
			continue;
		}
		self::$languages[$fileData["filename"]] = new Language($fileData["filename"], Path::join($this->getDataFolder(), "languages"));
	}
}

// add translations to existing server language instance
$serverLanguage = $this->getServer()->getLanguage();
$refClass = new \ReflectionClass($serverLanguage);
$refPropA = $refClass->getProperty('lang');
$refPropA->setAccessible(true);
/** @var string[] $serverLanguageList */
$serverLanguageList = $refPropA->getValue($serverLanguage);

$pluginLanguage = self::$languages[$serverLanguage->getLang()];
$refClass = new \ReflectionClass($pluginLanguage);
$refPropB = $refClass->getProperty('lang');
$refPropB->setAccessible(true);
/** @var string[] $pluginLanguageList */
$pluginLanguageList = $refPropB->getValue($pluginLanguage);

$refPropA->setValue($serverLanguage, array_merge($serverLanguageList, $pluginLanguageList));
```

```php
// onCommand
$language = $sender->getLanguage();
$sender->sendMessage(
	$language->translate(CustomKnownTranslationFactory::command_tpaccept_norequest())
);
```

After:
First code block unnecessary
```php
// onCommand
$sender->sendMessage(CustomKnownTranslationFactory::command_tpaccept_norequest());
```


SOF3:
> Developers may opt not to use this feature still -- that's why the Plugin interface exists separately anyway. PluginBase should have useful utilities like this to assist with ease of development.

While I am not directly against this PR, this argument doesn't sound right. The ability to opt out of PluginBase doesn't imply we can add anything to PluginBase with the alternative of opting out the entire PluginBase.

I know this PR doesn't really imply so (you don't even need to use it while using PluginBase), but the "that's why the Plugin interface exists separately anyway" argument is a dangerous fallacy.

src/lang/Language.php
```diff
@@ -59,7 +62,24 @@ public static function getLanguageList(string $path = "") : array{
 			$path = \pocketmine\LOCALE_DATA_PATH;
 		}
 
-		if(is_dir($path)){
+		if(str_contains($path, ".phar/")){
```
SOF3: phar paths should be distinguished by checking `str_starts_with($path, "phar://")`, right?

src/lang/Language.php
```diff
@@ -138,6 +165,17 @@ protected static function loadLang(string $path, string $languageCode) : array{
 		throw new LanguageNotFoundException("Language \"$languageCode\" not found");
 	}
 
+	public function merge(Language $language) : void{
```
SOF3, line 143: Is "merge" really a good way to look at this? Do we really want to override/affect existing language entries? If we expect both language objects to be disjoint, "add" sounds like a more reasonable verb.

SOF3:
This PR doesn't really allow plugins to add their own translations. It forces plugins to write all their translations through the INI format. Is it really what we desire to achieve?

SOF3:
Another question I probably asked in the previous PR: Can our current translation API support something in the InfoAPI style, where translated strings are computed server-side to take in arbitrary dynamically-defined context arguments? Or do users who want to support custom message templates still need to reinvent their own translation system?

src/lang/Language.php
jasonw4331: If it's to be renamed I think I prefer `importLanguageStrings` for accuracy

dktapps: `importLanguageStrings` doesn't tell anything about overwrite vs fill-the-blanks semantics.

src/lang/Language.php
```diff
@@ -59,7 +65,24 @@ public static function getLanguageList(string $path = "") : array{
 			$path = \pocketmine\LOCALE_DATA_PATH;
 		}
 
-		if(is_dir($path)){
+		if(str_starts_with($path, "phar://")){
```
dktapps: I'm not seeing why phar needs its own logic for this.

src/network/mcpe/handler/DeathPacketHandler.php
```diff
@@ -52,7 +55,8 @@ public function setUp() : void{
 		$parameters = [];
 		if($this->deathMessage instanceof Translatable){
 			$language = $this->player->getLanguage();
-			if(!$this->player->getServer()->isLanguageForced()){
+			$namespace = explode(".", $this->deathMessage->getText())[0];
+			if(!in_array($namespace, NamespacedLanguage::getNamespaces(), true) && !$this->player->getServer()->isLanguageForced()){
```
dktapps, line 59: This feels hacky, and I'm not a big fan of the in_array either. Seems to me that this should be achievable with a regular array<string, string>

src/plugin/PluginBase.php
```diff
@@ -116,6 +132,13 @@ final public function onEnableStateChange(bool $enabled) : void{
 			$this->isEnabled = $enabled;
 			if($this->isEnabled){
 				$this->onEnable();
+				try{
+					foreach($this->getTranslations() as $translation){
+						$this->server->getLanguage()->merge($translation);
+					}
+				}catch(\InvalidArgumentException $e){
```
dktapps: Logical exceptions like this generally shouldn't be caught, as we don't know where they are coming from. If you want to catch an exception here, a specific exception should be introduced.

src/plugin/PluginBase.php
```diff
@@ -69,6 +78,13 @@ public function __construct(
 		$this->file = rtrim($file, "/" . DIRECTORY_SEPARATOR) . "/";
 		$this->configFile = Path::join($this->dataFolder, "config.yml");
 
+		$translations = Path::join($file, 'resources', 'translations');
+		if(is_dir($translations) || ($loader->getAccessProtocol() === 'phar://' && isset((new \Phar($file))['resources/translations']))){ // TODO: script plugins
```
dktapps: I'm not liking the use of `phar` specific logic here.

src/network/mcpe/handler/DeathPacketHandler.php
jasonw4331: Please clarify what you're looking for here.

dktapps:
For posterity: I think this PR is trying to do too many things at once. As mentioned above, the PluginBase stuff doesn't make a ton of sense.

I think there's a case to be made for injecting additional strings into the Language. However, I don't think this PR approaches this in a sensible fashion. This could be achieved with a small change to the Language class itself.

There's also the question of how useful fallback language would really be considering the lack of validation. (This validation is currently not required for loading PM's language files, although it probably should be present.)

Considering that the core language system is overdue a redesign itself, I'm not sure it makes sense to bolt even more stuff on it, especially since the approach of this PR isn't well-liked by various team members (including myself).

PR #5866, opened 2023-07-01
Give an extra month of life to the old auth key

IvanCraft623:
Server in some time zones are getting players unable to connect due to PM has decided to change keys before mojang starts using the new one.

This PR expands the expiration time given by PM to the old key by one month, to workaround the mentioned above.


dktapps:
For context on the July 1st expiry date: I had it on good authority from Mojang contacts that the key would be changed "a couple of weeks" after the 1.20 release. However, this has evidently not happened.

Seems like a better approach would have been to just use the old key exclusively, and then change over to the new key when everyone spams my Discord notifications. That way everyone would be forced to update to a version which only accepts the new key, and this pandemonium would have a purpose other than making me look stupid.

dktapps:
Obsoleted by 249ef9c534656e4286e5fffb6ff60705dc424414.

PR #5871, opened 2023-07-04
declare RegistryTrait's members type

Lunarelly:


SOF3:
Please fix the lint. Turns out the original doc comment doesn't look very correct either.

dktapps:
This is BC-breaking for any class that uses RegistryTrait.

dktapps:

PR #5872, opened 2023-07-04
Disallow plugins disabling other plugins

### LINKED ISSUE #2826: Disallow non-shutdown plugin disabling
## Description
Following #2823, it is now possible to further regularize the plugin loading lifecycle.

This issue proposes three alternatives:
- Plugins must not request to disable **other** plugins, but they can suicide anyway
- Plugins must not be disabled at all, except the natural disabling on shutdown or during startup crash
- Plugins must not be disabled even during startup crash

I personally tend to implement the first alternative, but the other two are also open for discussion.

## Justification
Why plugins should not disable other plugins:
- This may cause instability of other plugins. It should be more convenient to disable
- Disabling plugins should be a concept local to the plugin itself and be handled by itself. PocketMine should not be so authoritarian and force plugin disabling into their API. Plugins tend to have complex multifold mechanisms, and it is usually more desirable to only disable certain components, e.g. a player using a world edit plugin might want to turn off world edit mode for themselves only, but not to disable it for the whole server; or the owner just wants to pause a plugin from running for a while; these should be explicitly supported by plugins, not brutally deafening their event handlers.

Why plugins should not disable themselves between startup and shutdown:
- Give me a reason why they should. I have actually never seen a plugin doing this. Disabling a plugin actually just deafens their listeners, hide their commands and kill their scheduler tasks. It is not even conceptually "disabled", but just isolated from the PocketMine API.
- Because they can't crash on startup; they fetched startup data asynchronously that did not arrive immediately.
  - Let's promisify server startup, allowing plugins to hold back startup. This would be more useful even beyond the scope of this issue.

Why plugins should not disable themselves on startup:
- If they crash due to uncoverable invalid configuration etc., just throw an exception instead of explicitly calling `disablePlugin`. Let PocketMine decide whether to just disable the plugin or crash the server.
- If some config value asked the plugin to disable... Why do you even need this? If you don't want to enable a plugin, simply move it to another folder or change its file extension. You don't need to make a config value to ask to disable your own plugin!

But here comes the question: I only raised why they aren't needed (rebutting the advantages), but didn't provide a motivation to remove them (suggesting the disadvantages). Indeed, there is no motivation to remove except "make the code cleaner and easier". The first alternative has a positive impact on plugin developers, while the second and third only improve the simplicity of PocketMine itself, so I would personally support just restricting disabling other plugins.

But how to detect if it is the plugin disabling itself? Simply add an API that only expects the plugin itself to call; API design is not security design, and we don't need to defend it. But we can still make it clear that our API design does not expect killing other plugins.

JavierLeon9966:
## Introduction

Quote from @SOF3:
> Why plugins should not disable other plugins:
> 
> - This may cause instability of other plugins. It should be more convenient to disable
> - Disabling plugins should be a concept local to the plugin itself and be handled by itself. PocketMine should not be so authoritarian and force plugin disabling into their API. Plugins tend to have complex multifold mechanisms, and it is usually more desirable to only disable certain components, e.g. a player using a world edit plugin might want to turn off world edit mode for themselves only, but not to disable it for the whole server; or the owner just wants to pause a plugin from running for a while; these should be explicitly supported by plugins, not brutally deafening their event handlers.

### Relevant issues

Fixes #2826 
### API changes

None
### Behavioural changes

None
## Backwards compatibility

`PluginManager->disablePlugin()` is now `private`.
`PluginManager->disablePlugins()` is now an `@internal` method.
`PluginManager->clearPlugins()` is now removed since it isn't used by the core.
## Follow-up

None
## Tests

N/A


SOF3:
So plugins are still expected to disable themselves, but only by throwing DisablePluginException, or what?

JavierLeon9966:
Yes, they can disable themselves by throwing `DisablePluginException`

SOF3:
Does DisablePluginException just disable one plugin or stops the server? I thought we have another issue that says if one plugin can't be loaded, the whole server shouldn't start.

JavierLeon9966:
It disables the plugin and then stops the server.

alvin0319:
The server will be shutdown whenever plugin uses `disablePlugin()` to disable themselves anyway.

JavierLeon9966:
Now you can't do it *after* the server has started.

dktapps:
>  I thought we have another issue that says if one plugin can't be loaded, the whole server shouldn't start.

This was changed in 4.4.0 by https://github.com/pmmp/PocketMine-MP/commit/52e74296dec604f64bd7e8fd1011474387690a5c

In light of this change, `DisablePluginException` probably ought to be renamed to `PluginEnableException`, since it's not actually disabling anything, just aborting server startup. However, that's a debate unrelated to this PR.

dktapps (commented):
Conceptually, this seems to make sense, and I don't have any argument against it.

On a related note, since `disablePlugin()` is no longer accessible to plugins, it may be possible to simplify the code of `enablePlugin()`, since we don't have to worry about the plugin randomly disabling itself anymore.

I wonder if it even makes sense to call `onDisable()` for a plugin which failed to enable correctly... Perhaps that's a debate for another issue

ShockedPlot7560 (approved):
I don't see any counter-argument to implementing this logic. The code can be simplified at a later stage, I think, because it's indirectly linked to this PR.

> I wonder if it even makes sense to call onDisable() for a plugin which failed to enable correctly... Perhaps that's a debate for another issue

I don't think this method is appropriate in the sense that ``onDisable`` would imply that it is deactivated after activation. However, if it fails to start, another function might make more sense.

dktapps (approved):

PR #5874, opened 2023-07-04
Generic types for ``RegistryTrait``

### LINKED ISSUE #5275: Use PHPStan generics for RegistryTrait
## Description
This wasn't originally supported by PHPStan when I first wrote the code.

Adding these isn't exactly a BC break, but it will cause a disruption for downstream plugin developers using PHPStan, so it should probably target a minor release.

ShockedPlot7560:
## Introduction
Currently, static code analysis may not correctly understand RegistryTrait uses and related issues due to its generic typing.

This PR implements the specification of a generic type for these traits.

### Relevant issues
* Fixes #5275

## Backwards compatibility
It's not really, because nothing has been touched to the API. However, it may lead to disruptions in downstream plugins using PHPStan and ``RegistryTrait`` / ``CloningRegistryTrait``.


src/utils/RegistryTrait.php
```diff
@@ -90,9 +96,17 @@ private static function _registryFromString(string $name) : object{
 		if(!isset(self::$members[$upperName])){
 			throw new \InvalidArgumentException("No such registry member: " . self::class . "::" . $upperName);
 		}
-		return self::preprocessMember(self::$members[$upperName]);
+		/** @phpstan-var T $return */
+		$return = self::preprocessMember(self::$members[$upperName]);
+		// We need to force typing here, otherwise PHPStan understands that it's an object and not T
```
dktapps, line 107: This seems a bit suspicious. I really don't like using @var when the type should be obvious.

src/utils/RegistryTrait.php
```diff
@@ -117,12 +132,15 @@ public static function __callStatic($name, $arguments){
...
-	 * @phpstan-return array<string, object>
+	 * @phpstan-return array<string, T>
 	 */
 	private static function _registryGetAll() : array{
 		self::checkInit();
 		return array_map(function(object $o) : object{
-			return self::preprocessMember($o);
+			/** @phpstan-var T $return */
+			$return = self::preprocessMember($o);
+			// We need to force typing here, otherwise PHPStan understands that it's an object and not T
```
dktapps: Same again here.

dktapps:
Prefer using `phpstan-bugs.neon` instead of suppressing errors using `@var`. Use of `@var` is a recipe for hidden bugs.

ShockedPlot7560:
Waiting for https://github.com/phpstan/phpstan/issues/9630 to allow the removal of ``@var`` that are not ignorable in the baseline.

dktapps:
Closing this until the referenced PHPStan bug is fixed (which is probably gonna take a long time). This will just accumulate conflicts in the meantime.

PR #5875, opened 2023-07-04
Implemented new bamboo-wood blocks & bamboo raft

Crasher508:
## Introduction
Sorry for the wrong format of this pr introduction, i miss time. I implement and test all the bamboo blocks and items. Workaround in VanillaBlocks is not the best, but it works quit nice.

### API changes
- Enums:
  Add WoodType::BAMBOO()
  Add BoatType::BAMBOO()
- new methods:
  BoatType::BAMBOO()

### Behavioural changes
Add a check in VanillaBlocks class which lookup for WoodType::BAMBOO()

## Backwards compatibility
fully given

## Tests
- run server, place and break blocks, save and load world
- checks of github


JavierLeon9966 (changes requested):
This should target `minor-next`

Crasher508:
I hope the pr has now the correct format

src/item/BoatType.php
```diff
@@ -33,7 +33,7 @@
  * @generate-registry-docblock
  *
  * @method static BoatType ACACIA()
- * @method static BoatType BAMBOO)
+ * @method static BoatType BAMBOO()
```
dktapps, line 36: This is supposed to be auto generated. Didn't you read the warning?

src/block/VanillaBlocks.php
```diff
@@ -1249,7 +1264,13 @@ private static function registerWoodenBlocks() : void{
 			$idName = fn(string $suffix) => $woodType->name() . "_" . $suffix;
 
 			self::register($idName(mb_strtolower($woodType->getStandardLogSuffix() ?? "log", 'US-ASCII')), new Wood(WoodLikeBlockIdHelper::getLogIdentifier($woodType), $name . " " . ($woodType->getStandardLogSuffix() ?? "Log"), $logBreakInfo, $woodType));
-			self::register($idName(mb_strtolower($woodType->getAllSidedLogSuffix() ?? "wood", 'US-ASCII')), new Wood(WoodLikeBlockIdHelper::getAllSidedLogIdentifier($woodType), $name . " " . ($woodType->getAllSidedLogSuffix() ?? "Wood"), $logBreakInfo, $woodType));
+			if(WoodType::BAMBOO()->equals($woodType)) {
+				self::register("bamboo_mosaic", new Planks(new BID(Ids::BAMBOO_MOSAIC), "Bamboo Mosaic", $planksBreakInfo, $woodType));
+				self::register("bamboo_mosaic_slab", new WoodenSlab(new BID(Ids::BAMBOO_MOSAIC_SLAB), "Bamboo Slab", $planksBreakInfo, $woodType));
+				self::register("bamboo_mosaic_stairs", new WoodenStairs(new BID(Ids::BAMBOO_MOSAIC_STAIRS), "Bamboo Stairs", $planksBreakInfo, $woodType));
```
dktapps: I think it's misleading to pass the wood type here. We already have variants of these blocks which will have a BAMBOO type.

dktapps:
> Workaround in VanillaBlocks is not the best, but it works quit nice.

Yeah, I already realized a similar hack would be necessary. In the future we'll probably have to split up `LogType` and `PlanksType`.

I'm not sure about how bamboo mosaic has been implemented though.

src/item/VanillaItems.php
```diff
@@ -545,16 +549,17 @@ public function isFireProof() : bool{ return true; }
...
-			self::register($type->name() . "_boat", new Boat(new IID(match($type){
+			self::register($type->name() . "_" . (mb_strtolower($type->getSuffix() ?? "boat", 'US-ASCII')), new Boat(new IID(match($type){
 				BoatType::OAK() => Ids::OAK_BOAT,
 				BoatType::SPRUCE() => Ids::SPRUCE_BOAT,
 				BoatType::BIRCH() => Ids::BIRCH_BOAT,
 				BoatType::JUNGLE() => Ids::JUNGLE_BOAT,
 				BoatType::ACACIA() => Ids::ACACIA_BOAT,
 				BoatType::DARK_OAK() => Ids::DARK_OAK_BOAT,
 				BoatType::MANGROVE() => Ids::MANGROVE_BOAT,
+				BoatType::BAMBOO() => Ids::BAMBOO_RAFT,
```
dktapps, line 560: I'm not really sure this is appropriate. It also wasn't mentioned in the PR description.

src/item/BoatType.php
Crasher508: I read it a bit to late xD Running "composer update-registry-annotations" add this commit

src/block/VanillaBlocks.php
Crasher508: Add a new WoodType::BAMBOO_MOSAIC() doesn't make sense to me, I expand it to VanillaBlocks::registerBambooMosaicBlocks() as a possible solution

dktapps: I'm not convinced bamboo mosaic should be considered a wood type at all.

src/item/BoatType.php
```diff
@@ -54,12 +55,14 @@ protected static function setup() : void{
...
 			new self("dark_oak", WoodType::DARK_OAK()),
 			new self("mangrove", WoodType::MANGROVE()),
+			new self("bamboo", WoodType::BAMBOO(), "Raft"),
 		);
 	}
 
 	private function __construct(
 		string $enumName,
 		private WoodType $woodType,
+		private readonly ?string $suffix = null,
```
ShockedPlot7560, line 65: Rather than setting it to null here, wouldn't it be better to set ``Boat``. This would avoid all conditions later

src/item/BoatType.php
```diff
@@ -69,4 +72,8 @@ public function getWoodType() : WoodType{ return $this->woodType; }
 	public function getDisplayName() : string{
 		return $this->woodType->getDisplayName();
```
ShockedPlot7560, line 73: Wouldn't it be better to combine suffix and wood type directly in this function rather than elsewhere?

src/item/VanillaItems.php
ShockedPlot7560: The name of the PR mentions bamboo items and therefore indirectly bamboo boats, which are above all boats.

ShockedPlot7560 (changes requested):

src/item/VanillaItems.php
dktapps: I'm not convinced that it's appropriate to consider rafts as boats. They have different behaviour and appearance.

src/block/VanillaBlocks.php
```diff
@@ -1624,4 +1643,13 @@ private static function registerCauldronBlocks() : void{
...
 		self::register("potion_cauldron", new PotionCauldron(new BID(Ids::POTION_CAULDRON, TileCauldron::class), "Potion Cauldron", $cauldronBreakInfo));
 	}
+
+	private static function registerBambooMosaicBlocks() : void{
+		$woodType = WoodType::BAMBOO();
+		$planksBreakInfo = new Info(BreakInfo::axe(2.0, null, 15.0));
+
+		self::register("bamboo_mosaic", new Planks(new BID(Ids::BAMBOO_MOSAIC), "Bamboo Mosaic", $planksBreakInfo, $woodType));
+		self::register("bamboo_mosaic_slab", new WoodenSlab(new BID(Ids::BAMBOO_MOSAIC_SLAB), "Bamboo Slab", $planksBreakInfo, $woodType));
+		self::register("bamboo_mosaic_stairs", new WoodenStairs(new BID(Ids::BAMBOO_MOSAIC_STAIRS), "Bamboo Stairs", $planksBreakInfo, $woodType));
```
dktapps: The names of these blocks are incorrect. Also, as I said previously, I think considering these as WoodType::BAMBOO is misleading. From an API perspective they will appear the same as actual bamboo stairs etc. which is misleading to plugin devs.

src/item/VanillaItems.php
ShockedPlot7560: What's the difference in behavior? Because as far as I can see, only the appearance changes.

dktapps: You can suffocate on a raft when going under 1 block, but not on a boat, since you sit higher on a raft than on a boat.

src/item/BoatType.php
Crasher508: I follow the api style of WoodType enum, using a hard code string can be problematic

src/world/format/io/leveldb/LevelDB.php
```diff
@@ -199,6 +200,16 @@ protected function deserializeBlockPalette(BinaryStream $stream, \Logger $logger
...
+				$id = $blockStateNbt->getString("name");
+				if ($id != null) {
+					$item = StringToItemParser::getInstance()->parse($id);
+					if ($item != null) {
+						//$palette[] = $this->blockStateDeserializer->deserialize(BlockStateData::current($id, []));
+						$palette[] = $item->getBlock()->getStateId();
+						$logger->error("Failed to upgrade blockstate: " . $e->getMessage() . " offset $i in palette, blockstate NBT: " . $blockStateNbt->toString() . ". Fixed!");
+						continue;
+					}
+				}
```
dktapps: Unrelated change - I'm not sure what exactly is going on here ...

Crasher508: My mistake, removed by last commit

ShockedPlot7560:
In my opinion, we should wait for #5895 to have a more logical implementation of these blocks with differentiation between wood type and planks type

dktapps:
> In my opinion, we should wait for #5895 to have a more logical implementation of these blocks with differentiation between wood type and planks type

Not sure I want to wait for PM6 to have these implemented.

Regular bamboo excluding "bamboo log" and "bamboo all-sided log" seems OK to me (though obviously not ideal), but the handling of bamboo mosaic still makes me unhappy.

PR #5878, opened 2023-07-06
Use SingletonTrait in PermissionManager

Lunarelly:


IvanCraft623:
This should be targeted to `minor-next`

Lunarelly:
then merge it?

dktapps:
I'm unsure if this is desirable, since SingletonTrait also allows overwriting and resetting the singleton. Not really sure if that makes sense for PermissionManager.

Lunarelly:
maybe implement something like ReadonlySingletonTrait

src/permission/PermissionManager.php
```diff
@@ -23,19 +23,12 @@
...
-	private static ?self $instance = null;
-
-	public static function getInstance() : PermissionManager{
-		if(self::$instance === null){
-			self::$instance = new self();
-		}
-
-		return self::$instance;
-	}
+	use SingletonTrait;
```
xxAROX, line 31:
```php
use SingletonTrait{
   reset as private;
   setInstance as private;
}
```

dktapps:

PR #5894, opened 2023-07-12
Don't mention enchantments in generic StringToTParser doc-comment

Wertzui123:
The new comment may not be the best, but I'm pretty sure the thing with "enchantments" was accidentally copied over from `StringToEnchantmentParser` or something like that.

Is "an instance of T" better than "a value of T"? Or any other ideas?

### DIFF
--- a/src/utils/StringToTParser.php
+++ b/src/utils/StringToTParser.php
@@ -59,7 +59,7 @@ abstract class StringToTParser{
 	}
 
 	/**
-	 * Tries to parse the specified string into a value of T.
+	 * Tries to parse the specified string into a corresponding instance of T.
 	 * @phpstan-return T|null
 	 */
 	public function parse(string $input){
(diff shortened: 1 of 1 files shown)


src/utils/StringToTParser.php
```diff
@@ -59,7 +59,7 @@ public function override(string $alias, \Closure $callback) : void{
 	}
 
 	/**
-	 * Tries to parse the specified string into an enchantment.
+	 * Tries to parse the specified string into a value of T.
```
dktapps:
```suggestion
	 * Tries to parse the specified string into a corresponding instance of T.
```

dktapps:
> I'm pretty sure the thing with "enchantments" was accidentally copied over from StringToEnchantmentParser or something like that.

Yeah, originally StringTo*Parser was intended to replace Enchantment::fromString() and Effect::getEffectByName() and there was no generic variant. Once it started expanding to more stuff it became obvious that there was a code duplication problem.

PR #5895, opened 2023-07-13
Introduce planks type

ShockedPlot7560:
## Introduction
With Mojang's introduction of mosaic bamboo, it's necessary to have a way of distinguishing it.  
This is a somewhat direct way of dividing the blocks inheriting the planks from those inheriting the wood, but it is the most flexible way of making it easier to introduce mosaic bamboo blocks.

This PR goes hand in hand with #5875, which seems to need this API.

### API changes
- Added ``PlanksType``, it inserts ``WoodType`` values as standard, but may be extended in the future.
- Remove ``getWoodType`` for using ``getPlanksType`` in ``Planks``, ``WoodenSlab`` and ``WoodenStairs``
- Added ``PlanksLikeBlockIdHelper`` 
- Rename ``WoodType`` to ``LogType``
- Rename ``WoodTypeTrait`` to ``LogTypeTrait``
- Change ``writeLegacyWoodType`` to ``writeLegacyLogType`` in ``BlockStateWriter``
- Rename ``getWoodType`` to ``getLogType`` in ``BoatType``

## Backwards compatibility
- Api changed in `Planks`, `WoodenSlab` and `WoodenStairs`
- ``WoodType`` renamed in ``LogType``
- ``WoodTypeTrait`` renamed in ``LogTypeTrait``
- Change ``writeLegacyWoodType`` to ``writeLegacyLogType`` in ``BlockStateWriter``

## Follow-up
- Implements Bamboo Mosaic


dktapps:
WoodType should probably be scrapped and replaced with LogType

src/block/BaseSign.php
```diff
@@ -57,8 +57,8 @@ abstract class BaseSign extends Transparent{
 	/**
 	 * @param \Closure() : Item $asItemCallback
 	 */
-	public function __construct(BlockIdentifier $idInfo, string $name, BlockTypeInfo $typeInfo, WoodType $woodType, \Closure $asItemCallback){
-		$this->woodType = $woodType;
+	public function __construct(BlockIdentifier $idInfo, string $name, BlockTypeInfo $typeInfo, LogType $logType, \Closure $asItemCallback){
```
dktapps: This doesn't seem right... there are bamboo signs

src/block/FenceGate.php
```diff
@@ -36,7 +36,7 @@
 use pocketmine\world\sound\DoorSound;
 
 class FenceGate extends Transparent{
-	use WoodTypeTrait;
+	use LogTypeTrait;
```
dktapps: Same again here

src/block/PlanksLikeBlockIdHelper.php
```diff
@@ -0,0 +1,92 @@
...
+			PlanksType::ACACIA()->id() => Ids::ACACIA_STAIRS,
+			PlanksType::DARK_OAK()->id() => Ids::DARK_OAK_STAIRS,
+			PlanksType::MANGROVE()->id() => Ids::MANGROVE_STAIRS,
+			PlanksType::CRIMSON()->id() => Ids::CRIMSON_STAIRS,
+			PlanksType::WARPED()->id() => Ids::WARPED_STAIRS,
+			PlanksType::CHERRY()->id() => Ids::CHERRY_STAIRS,
+			default => throw new AssumptionFailedError("All planks types should be covered")
+		});
+	}
+}
```
dktapps: I'm not really sure what's gained by pulling this out into its own class.

src/block/WoodenDoor.php
```diff
@@ -23,8 +23,8 @@
 
 namespace pocketmine\block;
 
-use pocketmine\block\utils\WoodTypeTrait;
+use pocketmine\block\utils\LogTypeTrait;
 
 class WoodenDoor extends Door{
-	use WoodTypeTrait;
+	use LogTypeTrait;
```
dktapps: Bamboo door is a thing

src/block/WoodenFence.php
```diff
@@ -23,20 +23,20 @@
 
 namespace pocketmine\block;
 
-use pocketmine\block\utils\WoodTypeTrait;
+use pocketmine\block\utils\LogTypeTrait;
 
 class WoodenFence extends Fence{
-	use WoodTypeTrait;
+	use LogTypeTrait;
```
dktapps: and bamboo fence

src/block/WoodenPressurePlate.php
```diff
@@ -23,10 +23,10 @@
 
 namespace pocketmine\block;
 
-use pocketmine\block\utils\WoodTypeTrait;
+use pocketmine\block\utils\LogTypeTrait;
 
 class WoodenPressurePlate extends SimplePressurePlate{
-	use WoodTypeTrait;
+	use LogTypeTrait;
```
dktapps: and pressure plate

src/block/WoodenTrapdoor.php
```diff
@@ -23,10 +23,10 @@
 
 namespace pocketmine\block;
 
-use pocketmine\block\utils\WoodTypeTrait;
+use pocketmine\block\utils\LogTypeTrait;
 
 class WoodenTrapdoor extends Trapdoor{
-	use WoodTypeTrait;
+	use LogTypeTrait;
```
dktapps: and trapdoor

src/block/utils/PlanksType.php
```diff
@@ -0,0 +1,79 @@
...
+ */
+final class PlanksType{
+	use EnumTrait {
+		__construct as private Enum___construct;
+	}
+
+	protected static function setup() : void{
+		self::registerAll(
+			// TODO: bamboo mosaic
+			...self::fromLogType()
```
dktapps: This is generally not recommended

src/item/BoatType.php
```diff
@@ -47,26 +47,26 @@ final class BoatType{
...
-			new self("acacia", WoodType::ACACIA()),
-			new self("dark_oak", WoodType::DARK_OAK()),
-			new self("mangrove", WoodType::MANGROVE()),
+			new self("oak", LogType::OAK()),
+			new self("spruce", LogType::SPRUCE()),
+			new self("birch", LogType::BIRCH()),
+			new self("jungle", LogType::JUNGLE()),
+			new self("acacia", LogType::ACACIA()),
+			new self("dark_oak", LogType::DARK_OAK()),
+			new self("mangrove", LogType::MANGROVE()),
```
dktapps: boats are made from planks

dktapps (changes requested):
I think there is some confusion here over what's appropriate to use for the different types of blocks.

src/block/utils/LogTypeTrait.php
```diff
@@ -26,15 +26,15 @@
 use pocketmine\block\BlockIdentifier;
 use pocketmine\block\BlockTypeInfo;
 
-trait WoodTypeTrait{
-	private WoodType $woodType; //immutable for now
+trait LogTypeTrait{
```
dktapps, line 29: Seems like this trait is now redundant, as far as I can see

src/block/utils/PlanksType.php
```diff
@@ -0,0 +1,67 @@
...
+ * @method static PlanksType SPRUCE()
+ * @method static PlanksType WARPED()
+ */
+final class PlanksType{
+	use EnumTrait {
+		__construct as private Enum___construct;
+	}
+
+	protected static function setup() : void{
+		foreach(LogType::getAll() as $type) {
```
dktapps, line 51: This is probably not a good idea - it won't translate well to native enums

dktapps:
I feel like `PlanksType` should remain as `WoodType`. `LogType` is the outlier here, since it only applies to one block type. This would minimize unnecessary API breaks and reduce the diff size.

ShockedPlot7560:
> I feel like `PlanksType` should remain as `WoodType`. `LogType` is the outlier here, since it only applies to one block type. This would minimize unnecessary API breaks and reduce the diff size.

However, the reason for separating is to be able to distinguish between planks and logs. The problem arises with bamboo and bamboo planks.

dktapps:
Bamboo is still a wood type, it just doesn't have an associated log type.

As I said, I don't think the BC break of renaming all of this stuff to `Planks` is justified. Only logs really need to change IMO.

dktapps:
Closing this due to lack of activity and conflicts. Introducing a `LogType` would be a lot simpler and require less work.

PR #5901, opened 2023-07-14
A small optimisation on gamemode change

SmallkingDev:
## Introduction
In the method `internalSetGameMode` is performed 2 calls of the method `syncAbilities` without any need, since the method `syncGamemode` performs the method call at the end of the whole process. The validation that is in the `setGamemode` method does not consider the previous gamemode of the player. This way it validates if the player is in spectator mode and calls the `despawnFromAll` method otherwise it just calls the `spawnToAll` method even if the player's gamemode was not Spectator before.

### API changes
Added `internalSetFlying` method

### Behavioural changes
Yes! Considering that it would be 2 less packages that would be sent.

### DIFF
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -489,10 +489,14 @@ class Player extends Human implements CommandSender, ChunkListener, IPlayer{
 		return $this->blockCollision;
 	}
 
+	private function internalSetFlying(bool $value) : void{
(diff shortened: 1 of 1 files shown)


dktapps:
I don't really see what value this adds. It's a cold path (and this case is rare anyway) and the code complexity is now higher than it was before.

PR #5904, opened 2023-07-15
Remove unnecessary param in Config

Lunarelly:


skyss0fly:
930E117C-DB68-4B89-8169-0083B2F9FAC9
Have a coding meme

dktapps:
Default parameters are evil. The original code has a clearer intent for readers who may not be aware of the default.

PR #5906, opened 2023-07-16
World hopper interactions

CoderJoeW:
## Introduction
This pull request solves the following issue

https://github.com/pmmp/PocketMine-MP/projects/14#card-67670591

## Follow-up
Normally when a hopper is powered it locks the hopper. Due to redstone and powering functionality not currently being implemented I saw no need to implement this in the hopper(yet)


ShockedPlot7560:
There's already an open PR on the subject: #5869

CoderJoeW:
@ShockedPlot7560 I have looked over the original pull request and it has some issues

1. The code does not seem to handle input properly(when hopper is on top of furnance) it will throw anything into the input
2. The code has this long if/elseif/else statement

I think I could maybe build a better implementation at least I would like the chance to try to build a better implementation

ShockedPlot7560:
As mentioned in the other PR, transfer logics should be located in the blocks directly, to allow greater flexibility, or at least be externalized from the hopper class itself : https://github.com/pmmp/PocketMine-MP/pull/5869#discussion_r1261131283

This PR also seems incomplete, as it lacks the logic to retrieve items / check entities.

If it's in progress, convert it into a draft so you can tell when it's ready for review.

CoderJoeW:
@ShockedPlot7560  How do I convert it into a draft?

ColinHDev:
> @ShockedPlot7560 I have looked over the original pull request and it has some issues
> 
> 1. The code does not seem to handle input properly(when hopper is on top of furnance) it will throw anything into the input
> 2. The code has this long if/elseif/else statement
> 
> I think I could maybe build a better implementation at least I would like the chance to try to build a better implementation

It does handle the input properly, as you can see in the [wiki](https://minecraft.fandom.com/wiki/Hopper#Container_interactions): `A working hopper pointing into top of a furnace deposits only into the ingredient slot. It can push any item, including items that can't be smelted by the furnace.`

CoderJoeW:
@ColinHDev I didnt know that. Its weird thats how it is supposed to work in vanilla mc

CoderJoeW:
Perhaps this is better off being closed then as I clearly do not know as much as I need to know for this implementation

src/block/Hopper.php
```diff
@@ -92,7 +95,39 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
 	}
 
 	public function onScheduledUpdate() : void{
-		//TODO
+		$world = $this->position->getWorld();
+		$world->scheduleDelayedBlockUpdate($this->position, 8);
```
ColinHDev: If the hoppers didn't do anything, they should be scheduled for the next tick instead of the 8th.

src/block/HopperInteractable.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\block;
+
+use pocketmine\block\tile\Hopper as TileHopper;
+
+interface HopperInteractable{
+	public function pull(TileHopper $tileHopper) : void;
```
ColinHDev: This should probably use the hopper block instead of the hopper tile as the parameter to reduce the dependence on tile classes

src/block/Furnace.php
```diff
@@ -92,4 +95,50 @@ public function onScheduledUpdate() : void{
...
+
+			if($hopperFacing === Facing::DOWN && $furnanceInventory->canAddSmelting($singleItem)){
+				$this->transferItem($hopperInventory, $furnanceInventory, $singleItem, 0);
+			}elseif($hopperFacing !== Facing::DOWN && $hopperFacing !== Facing::UP && $furnanceInventory->canAddFuel($singleItem)){
+				$this->transferItem($hopperInventory, $furnanceInventory, $singleItem, 1);
+			}
+		}
+	}
+
+	private function transferItem(SimpleInventory $sourceInventory, SimpleInventory $targetInventory, Item $item, int $slot = null) : void{
```
ColinHDev: Why should the `$slot` variable be optional?

src/block/Furnace.php
```diff
@@ -92,4 +95,50 @@ public function onScheduledUpdate() : void{
...
+
+		for($i = 0; $i < 5; $i++){
+			$itemStack = $hopperInventory->getItem($i);
+
+			if($itemStack->isNull()) continue;
+
+			$singleItem = $itemStack->pop(1);
+
+			if($hopperFacing === Facing::DOWN && $furnanceInventory->canAddSmelting($singleItem)){
+				$this->transferItem($hopperInventory, $furnanceInventory, $singleItem, 0);
```
ColinHDev: You probably should use the slot constants in the `FurnaceInventory` class instead of hardcoding that value.

src/block/Furnace.php
```diff
@@ -92,4 +95,50 @@ public function onScheduledUpdate() : void{
...
+			}elseif($hopperFacing !== Facing::DOWN && $hopperFacing !== Facing::UP && $furnanceInventory->canAddFuel($singleItem)){
+				$this->transferItem($hopperInventory, $furnanceInventory, $singleItem, 1);
```
ColinHDev: same here

src/block/inventory/FurnaceInventory.php
```diff
@@ -68,4 +68,19 @@ public function setFuel(Item $item) : void{
 	public function setSmelting(Item $item) : void{
 		$this->setItem(self::SLOT_INPUT, $item);
 	}
+
+	public function canAddSmelting(Item $item) : bool{
+		if($this->getSmelting()->isNull()) return true;
+
+		return $this->getSmelting()->getTypeId() === $item->getTypeId();
```
ColinHDev: This doesn't check if the item is already at its maxed stack size (`Item::getCount() >= Item::getMaxStackSize()`) or if the items can't be stacked (e.g. due to different nbt tags).
So you should rely on the `Item::canStackWith()` method instead of comparing the type ids.

src/block/inventory/FurnaceInventory.php
```diff
@@ -68,4 +68,19 @@ public function setFuel(Item $item) : void{
...
+	}
+
+	public function canAddFuel(Item $item) : bool{
+		$typeId = $item->getTypeId();
+		if($item->getFuelTime() == 0) return false;
+
+		if($this->getFuel()->isNull()) return true;
+
+		return $this->getFuel()->getTypeId() === $item->getTypeId();
```
ColinHDev: same here

src/block/Hopper.php
```diff
@@ -92,7 +95,39 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+	}
+
+	public function pull(TileHopper $tileHopper) : void{
+		$hopperTile = $this->position->getWorld()->getTile($this->position);
+		if(!$hopperTile instanceof TileHopper) return;
+
+		$sourceInventory = $tileHopper->getInventory();
+		$targetInventory = $hopperTile->getInventory();
+
+		for($i = 0; $i < 5; $i++){
```
ColinHDev: same here regarding `Inventory::getSize()`

src/block/Hopper.php
```diff
@@ -92,7 +95,39 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+			$itemStack = $sourceInventory->getItem($i);
+
+			if($itemStack->isNull()) continue;
+
+			$singleItem = $itemStack->pop(1);
+
+			if($targetInventory->canAddItem($singleItem)){
```
ColinHDev: Your loop is stopped even if the item couldn't be added. But that would mean that it doesn't check if an item in the following slot could be added to the target inventory, just because the item in the current slot could not be added.
You should `continue` if `canAddItem()` returns false.

src/block/Furnace.php
```diff
@@ -92,4 +95,50 @@ public function onScheduledUpdate() : void{
...
+		if(!$furnanceTile instanceof TileFurnace) return;
+
+		$hopperBlock = $tileHopper->getBlock();
+		if(!$hopperBlock instanceof Hopper) return;
+
+		$hopperFacing = $hopperBlock->getFacing();
+		$hopperInventory = $tileHopper->getInventory();
+		$furnanceInventory = $furnanceTile->getInventory();
+
+		for($i = 0; $i < 5; $i++){
```
ColinHDev: You should use Inventory::getSize() instead of hardcoding it.

src/block/Furnace.php
```diff
@@ -92,4 +95,50 @@ public function onScheduledUpdate() : void{
...
+			if($itemStack->isNull()) continue;
+
+			$singleItem = $itemStack->pop(1);
+
+			if($hopperFacing === Facing::DOWN && $furnanceInventory->canAddSmelting($singleItem)){
+				$this->transferItem($hopperInventory, $furnanceInventory, $singleItem, 0);
+			}elseif($hopperFacing !== Facing::DOWN && $hopperFacing !== Facing::UP && $furnanceInventory->canAddFuel($singleItem)){
+				$this->transferItem($hopperInventory, $furnanceInventory, $singleItem, 1);
+			}
+		}
```
ColinHDev, line 117: The problem with this loop is that it doesn't stop if it added an item to the furnace's inventory.
If I would place coal in all five slots of my hopper, your implementation would always add five coal to the furnace.

ColinHDev (changes requested):
Some comments on the code, aside from the missing features (pushing logic, missing pulling logic for certain blocks like chests and item entity pick-up logic)

ColinHDev:
> @ColinHDev I didnt know that. Its weird thats how it is supposed to work in vanilla mc

It matches how the player can interact with furnaces. As a player, you also can put any item in the furnace's smelting slot, but only items like coal into the fuel one.

src/block/HopperInteractable.php
CoderJoeW: Wouldnt this(at least at the moment) just be kicking the can down the road because at some point the tile still needs to be accessed

If not in the top level scheduler then it would be in the functions

src/block/Hopper.php
CoderJoeW: I could not duplicate this I think it works correctly

src/block/Furnace.php
CoderJoeW: Maybe it is resolved by changes I made today not sure couldnt duplicate

ShockedPlot7560:
Should target ``next-minor``

ColinHDev:
I'm sorry, I didn't notice it earlier:
Hoppers _**pull**_ items from containers that are placed above them. Hoppers _**push**_ blocks into the container they are facing.
And while your method is called `pull()` it pushes items to other containers instead of pulling them into the hopper.

CoderJoeW:
@ColinHDev Your right and I was thinking about that yesterday when I was working on it

Going to make that change today

CoderJoeW:
Ready for review

src/block/Furnace.php
```diff
@@ -92,4 +95,48 @@ public function onScheduledUpdate() : void{
...
+		if(!$furnanceTile instanceof TileFurnace) return false;
+
+		$hopperBlock = $tileHopper->getBlock();
+		if(!$hopperBlock instanceof Hopper) return false;
+
+		$hopperFacing = $hopperBlock->getFacing();
+		$hopperInventory = $tileHopper->getInventory();
+		$furnanceInventory = $furnanceTile->getInventory();
+
+		for($i = 0; $i < $furnanceInventory->getSize(); $i++){
```
ColinHDev: This should be `$hopperInventory`

src/block/Hopper.php
```diff
@@ -92,7 +97,68 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+	}
+
+	public function push(TileHopper $tileHopper) : bool{
+		$hopperTile = $this->position->getWorld()->getTile($this->position);
+		if(!$hopperTile instanceof TileHopper) return false;
+
+		$sourceInventory = $tileHopper->getInventory();
+		$targetInventory = $hopperTile->getInventory();
+
+		return $this->transferItem($sourceInventory, $targetInventory);
```
ColinHDev: When pushing into another hopper, that hopper is set on an eight tick cooldown.

src/block/Hopper.php
```diff
@@ -92,7 +97,68 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
 	public function onScheduledUpdate() : void{
-		//TODO
+		$world = $this->position->getWorld();
+
+		$tile = $this->position->getWorld()->getTile($this->position);
+		if(!$tile instanceof TileHopper) { $world->scheduleDelayedBlockUpdate($this->position, 1); return; }
+
+		$facingBlock = $this->getSide($this->facing);
+		$topBlock = $this->getSide(Facing::UP);
+		if(!$facingBlock instanceof HopperInteractable && !$topBlock instanceof Chest && !$topBlock instanceof Hopper) { $world->scheduleDelayedBlockUpdate($this->position, 1); return; }
```
ColinHDev: Don't hardcode the blocks here, that's the reason the interface was implemented.

Also, this if statement is very hard to read due to the missing line breaks.

src/block/Hopper.php
```diff
@@ -92,7 +97,68 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+
+		$pullSuccess = $this->pull($tile);
```
ColinHDev: The `pull()` method should simply be implemented into the `HopperInteractable` interface. Especially since hoppers can also pull from most of the blocks they can push to.

src/block/Hopper.php
```diff
@@ -92,7 +97,68 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+
+		if($pullSuccess && !$facingBlock instanceof HopperInteractable){
+			$world->scheduleDelayedBlockUpdate($this->position, 8);
+			return;
+		}
+
+		if(!$facingBlock instanceof HopperInteractable) { $world->scheduleDelayedBlockUpdate($this->position, 1); return; }
+		$updateTime = $facingBlock->push($tile) ? 8 : 1;
```
ColinHDev: Hoppers push before they pull as stated in the wiki: `Item pushes and pulls are processed in the same game tick, but pushes are processed before pulls.`

src/block/Hopper.php
```diff
@@ -92,7 +97,68 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+
+		$world->scheduleDelayedBlockUpdate($this->position, $updateTime);
```
ColinHDev: There are a lot of `scheduleDelayedBlockUpdate()` calls in this method. This could probably be improved.

ColinHDev (changes requested):

ColinHDev:
A few other comments:
- There are some complaints by PHPStan that need to be addressed.
- What is your goal with this pr? Currently, you implemented the hopper push logic for some blocks (a lot are still missing, like brewing stands or barrels), partially the pull logic as far as I could tell and none of the item entity pick-up logic. How much do you plan to implement here?
- You still rely on the hopper tile instead of the block. I know, that the goal is / was to reduce the dependence on the tile objects in these places. Although you have a point, that those methods will have to get the tile themselves, a maintainer should probably just settle that argument.

CoderJoeW:
Okay made some big changes which I believe puts the code on a good path forward for implementing for other blocks currently the following blocks should be fully implemented

- Hopper
- Chest
- Furnace

There may still be a debate about the cooldown? I am not sure I think it works correctly now

The hopper will push before it pulls
If a push or pull is successful the next interaction happens on the 8 tick otherwise it happens on the next tick

Functionality may not be 100% correct for the following

- Blast furnace
- Smoker
- Trapped chest

src/block/HopperInteractable.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\block;
+
+use pocketmine\block\tile\Hopper as TileHopper;
+use pocketmine\inventory\BaseInventory;
+
+interface HopperInteractable{
```
ShockedPlot7560, line 31: Need docs I think

src/block/HopperInteractable.php
```diff
@@ -0,0 +1,32 @@
...
+	public function push(BaseInventory $targetInventory) : bool;
+	public function pull(TileHopper $tileHopper) : bool;
```
ShockedPlot7560: Need block instead of tile, in a logic of tile replacement, it's best to avoid adding dependencies.

src/block/Chest.php
```diff
@@ -97,4 +99,42 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
 	public function getFuelTime() : int{
 		return 300;
 	}
+
+	public function pull(TileHopper $tileHopper) : bool{
+		$currentTile = $this->position->getWorld()->getTile($this->position);
+		if(!$currentTile instanceof TileChest) return false;
```
ShockedPlot7560: CS, missing braces

src/block/Chest.php
```diff
@@ -97,4 +99,42 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+		$sourceInventory = $currentTile->getInventory();
+
+		return $this->transferItem($sourceInventory, $targetInventory);
+	}
+
+	private function transferItem(BaseInventory $sourceInventory, BaseInventory $targetInventory) : bool{
+		for($i = 0; $i < $sourceInventory->getSize(); $i++){
+			$itemStack = $sourceInventory->getItem($i);
+
+			if($itemStack->isNull()) continue;
```
ShockedPlot7560: Missing braces

src/block/Chest.php
```diff
@@ -97,4 +99,42 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+		$currentTile = $this->position->getWorld()->getTile($this->position);
+		if(!$currentTile instanceof TileChest) return false;
+
+		$sourceInventory = $currentTile->getInventory();
+
+		return $this->transferItem($sourceInventory, $targetInventory);
+	}
+
+	private function transferItem(BaseInventory $sourceInventory, BaseInventory $targetInventory) : bool{
+		for($i = 0; $i < $sourceInventory->getSize(); $i++){
```
ShockedPlot7560: Why you're not using a foreach here ?

src/block/Chest.php
```diff
@@ -97,4 +99,42 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+			$itemStack = $sourceInventory->getItem($i);
+
+			if($itemStack->isNull()) continue;
+
+			$singleItem = $itemStack->pop(1);
```
ShockedPlot7560: Unnecessary parameter

src/block/Chest.php
```diff
@@ -97,4 +99,42 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+
+			if(!$targetInventory->canAddItem($singleItem)) continue;
```
ShockedPlot7560: Can't we check the rest once the items have been added to the source inventory and then remove them?

src/block/Furnace.php
```diff
@@ -92,4 +96,68 @@ public function onScheduledUpdate() : void{
...
+		$sourceInventory = $tileHopper->getInventory();
+
+		for($i = 0; $i < $targetInventory->getSize(); $i++){
+			$itemStack = $sourceInventory->getItem($i);
+
+			if($itemStack->isNull()) continue;
+
+			$singleItem = $itemStack->pop(1);
+
+			if($hopperFacing === Facing::DOWN && $targetInventory->canAddSmelting($singleItem)){
```
ShockedPlot7560, line 110: It can pull any item, including items that can't be smelted by the furnace.

src/block/Hopper.php
```diff
@@ -92,7 +96,69 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
-		//TODO
+		$world = $this->position->getWorld();
+
+		$facingBlock = $this->getSide($this->facing);
+		$topBlock = $this->getSide(Facing::UP);
+
+		$pushSuccess = false;
+		$pullSuccess = false;
+
+		if($topBlock instanceof HopperInteractable){
```
ShockedPlot7560: The hopper starts to try to push items and next they pull.

src/block/Hopper.php
```diff
@@ -92,7 +96,69 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
 	}
 
 	public function onScheduledUpdate() : void{
-		//TODO
+		$world = $this->position->getWorld();
+
+		$facingBlock = $this->getSide($this->facing);
+		$topBlock = $this->getSide(Facing::UP);
```
ShockedPlot7560: Why set the two variables directly when they may not be needed later?

ShockedPlot7560 (changes requested):
I think to avoid confusion later. It might be necessary to rename the interface methods to : `pullFrom` and `pushTo`

By implementing the other logics, it will surely be possible to factorize the item transfer logic, which seems to be redundant.

src/block/Furnace.php
CoderJoeW: This is how it works currently the canAddSmelting only contains checks for if the item is max size or it can stack with the current item

ColinHDev:
> I think to avoid confusion later. It might be necessary to rename the interface methods to : `pullFrom` and `pushTo`
> 
> By implementing the other logics, it will surely be possible to factorize the item transfer logic, which seems to be redundant.

I already discussed it on discord: `push()` or `pushTo()` is not really self-explanatory without thinking of hoppers. My suggestion would be `onHopperPull()` / `onHopperPush()`

ColinHDev:
> Should target `next-minor`

Since the target branch got changed, might be a good time to merge `next-minor` into this one, so the "files changes" are not polluted and easier to read

ColinHDev:
Currently, you have duplicated logic in your code with the `transferItem()` method. Although this only is implemented twice, you probably have to implement it a few more times when accounting for other storage blocks like barrels or shulker boxes.
I propose a `HopperInteractableTrait` that implements basic `doHopperPull()` and `doHopperPush` methods: 
- Normal blocks like chests or barrels could just implement the `HopperInteractable` interface as well as using that trait and be done.
- Certain blocks like the hopper, which are set on an 8 tick cooldown if an item is pushed into them, could still make use of the trait and just also schedule the block update, only requiring them to contain a few lines calling the trait's method than implementing the logic themselves.
- Special blocks like the furnace blocks can still implement their logic and just don't have to use the trait.

src/block/Chest.php
CoderJoeW: I am not sure how else you would handle this

src/block/Chest.php
CoderJoeW: Im not sure what you mean here

CoderJoeW:
Alright I think that resolves the outstanding issues

src/block/Chest.php
ColinHDev: The default value of that parameter is already `1`, so `...->pop();` is sufficient: https://github.com/pmmp/PocketMine-MP/blob/stable/src/item/Item.php#L427

ColinHDev:
> Alright I think that resolves the outstanding issues

- There is still no logic for some of the blocks hoppers can interact with, e.g. brewing stands. A complete list would be [here in the wiki](https://minecraft.fandom.com/wiki/Hopper#Container_interactions)
- You switched the logic: `doHopperPush()` should handle the behaviour when a hopper pushes an item into another container. `doHopperPull()` on the other hand handles the behaviour when a hopper pulls an item from another container into its own inventory. It seems like you switched those in your implementation.
- You always used `BaseInventory` as the type for your method parameters. From what I've learned so far, is that you should generally use the base /parent type in those cases and not any of its implementations. So that would be using the Ìnventory` interface here for the type declaration. But someone might disagree with that.
- As already discussed in the #internals-chat on Discord, the `doHopperPush()` and `doHopperPull()` methods should just get the block as the parameter and not only / or in combination with the inventory. Yes, it would result in some boilerplate code where each implementation would have to get the tile instance, check if it's valid and then get the inventory. But, as discussed on Discord yesterday, the goal is to remove tiles from the core. That boilerplate code could then be simplified to just `$inventory = $block->getInventory()`.
Yes, your code would be fine for now, but in that version where tiles get removed, there would be no point in having two parameters in those methods for block and inventory. So the interface would have to change then, which would have the consequence of every implementation of `HopperInteractable` being forced to adapt, which is not only the case for classes in the core but also for every plugin that made use of that interface.

dktapps:
> So that would be using the `Inventory` interface here for the type declaration. But someone might disagree with that.

Yes, that's correct. There's no obvious reason to constrain stuff to `BaseInventory` since it's merely implementing `Inventory`. `BaseInventory` should be considered an implementation detail and shouldn't normally appear in typehints.

CoderJoeW:
@ColinHDev 

1. There is still no logic for some of the blocks hoppers can interact with, e.g. brewing stands. A complete list would be [here in the wiki](https://minecraft.fandom.com/wiki/Hopper#Container_interactions)

I am waiting on the base implementation to accepted as the standard before implementing the rest of the blocks it is painful enough testing with 3 blocks and making sure everything works correctly it would be more painful to test every block

2. You switched the logic: doHopperPush() should....

I dont think so?

3. You always used BaseInventory as the type....

This has now been updated to ```Inventory```

4. As already discussed in the #internals-chat on Discord....

This has now been corrected

ColinHDev:
> 1. There is still no logic for some of the blocks hoppers can interact with, e.g. brewing stands. A complete list would be [here in the wiki](https://minecraft.fandom.com/wiki/Hopper#Container_interactions)
> 
> I am waiting on the base implementation to accepted as the standard before implementing the rest of the blocks it is painful enough testing with 3 blocks and making sure everything works correctly it would be more painful to test every block

Fair enough.

> 2. You switched the logic: doHopperPush() should....
> 
> I dont think so?

If you look at your furnace class for example: You check in your `doHopperPull()` method how the hopper faces the furnace since different faces change the target slot in the furnace. This is done when the hopper pushes an item into the furnace.

src/block/Furnace.php
```diff
@@ -90,4 +93,77 @@ public function onScheduledUpdate() : void{
...
+			$itemStack = $sourceInventory->getItem($i);
+
+			if($itemStack->isNull()){
+				continue;
+			}
+
+			$singleItem = $itemStack->pop(1);
+
+			if($hopperFacing === Facing::DOWN && $targetInventory->canAddSmelting($singleItem)){
+				$this->transferItem($sourceInventory, $targetInventory, $singleItem, $targetInventory::SLOT_INPUT);
```
ColinHDev: It's debatable, but I believe it would be more readable to use `FurnaceInventory::SLOT_INPUT` here than `$targetInventory::SLOT_INPUT`.

src/block/Furnace.php
```diff
@@ -90,4 +93,77 @@ public function onScheduledUpdate() : void{
...
+				return true;
+			}elseif($hopperFacing !== Facing::DOWN && $hopperFacing !== Facing::UP && $targetInventory->canAddFuel($singleItem)){
+				$this->transferItem($sourceInventory, $targetInventory, $singleItem, $targetInventory::SLOT_FUEL);
```
ColinHDev: Would be the same here regarding `$targetInventory::SLOT_INPUT`

src/block/Furnace.php
```diff
@@ -90,4 +93,77 @@ public function onScheduledUpdate() : void{
...
+
+	public function doHopperPush(Inventory $targetInventory) : bool{
+		$currentTile = $this->position->getWorld()->getTile($this->position);
+		if(!$currentTile instanceof TileFurnace){
+			return false;
+		}
+
+		$sourceInventory = $currentTile->getInventory();
+
+		$itemStack = $sourceInventory->getItem($sourceInventory::SLOT_RESULT);
```
ColinHDev: This could just use `$sourceInventory->getResult();`

src/block/inventory/FurnaceInventory.php
```diff
@@ -68,4 +68,26 @@ public function setFuel(Item $item) : void{
 	public function setSmelting(Item $item) : void{
 		$this->setItem(self::SLOT_INPUT, $item);
 	}
+
+	public function canAddSmelting(Item $item) : bool{
+		if($this->getSmelting()->isNull()) return true;
+
+		$currentInput = $this->getSmelting();
```
ColinHDev: It's not that big of a deal, but you call `getSmelting()` twice here. Especially due to inventories cloning the item instance before returning it, this results in one unnecessary clone if the item is not null. So this should be changed to
```suggestion
		$currentInput = $this->getSmelting();
		if($currentInput->isNull()){
				return true;
		}
```

src/block/inventory/FurnaceInventory.php
```diff
@@ -68,4 +68,26 @@ public function setFuel(Item $item) : void{
...
+
+		if($currentInput->getCount() === $currentInput->getMaxStackSize()) return false;
```
ColinHDev: This should not only check for equality but also if the count is greater than the max stack size.
```suggestion
		if($currentInput->getCount() >= $currentInput->getMaxStackSize()){
				return false;
		}
```

src/block/inventory/FurnaceInventory.php
```diff
@@ -68,4 +68,26 @@ public function setFuel(Item $item) : void{
...
+	}
+
+	public function canAddFuel(Item $item) : bool{
+		if($item->getFuelTime() == 0) return false;
+
+		$currentFuel = $this->getFuel();
+
+		if($currentFuel->isNull()) return true;
+
+		if($currentFuel->getCount() === $currentFuel->getMaxStackSize()) return false;
```
ColinHDev: same here
```suggestion
		if($currentFuel->getCount() >= $currentFuel->getMaxStackSize()){
				return false;
		}
```

src/block/Furnace.php
```diff
@@ -90,4 +93,77 @@ public function onScheduledUpdate() : void{
...
+		$hopperFacing = $hopperBlock->getFacing();
+
+		for($i = 0; $i < $targetInventory->getSize(); $i++){
+			$itemStack = $sourceInventory->getItem($i);
+
+			if($itemStack->isNull()){
+				continue;
+			}
+
+			$singleItem = $itemStack->pop(1);
```
ColinHDev: The comment about this unnecessary value for the parameter seems to have gotten removed, so I'm just gonna mention it again.
```suggestion
			$singleItem = $itemStack->pop();
```

src/block/utils/HopperInteractableTrait.php
```diff
@@ -0,0 +1,49 @@
...
+use pocketmine\inventory\Inventory;
+
+trait HopperInteractableTrait{
+	public function transferItem(Inventory $sourceInventory, Inventory $targetInventory) : bool{
+		foreach($sourceInventory->getContents() as $itemStack){
+			if($itemStack->isNull()){
+				continue;
+			}
+
+			$singleItem = $itemStack->pop(1);
```
ColinHDev: same here
```suggestion
			$singleItem = $itemStack->pop();
```

src/block/HopperInteractable.php
ColinHDev: Yeah, at least a simple explanation of what this interface is for and when to implement it. Alongside that a small comment on the methods themselves might be useful.

src/block/HopperInteractable.php
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\block;
+
+use pocketmine\inventory\Inventory;
+
+interface HopperInteractable{
+	public function doHopperPush(Inventory $targetInventory) : bool;
```
ColinHDev: This should also use the `Hopper` block as the parameter. (Especially since you need to get the facing of the hopper when you want to implement the logic of hoppers pushing items into furnaces or brewing stands.

src/block/Furnace.php
```diff
@@ -90,4 +93,77 @@ public function onScheduledUpdate() : void{
...
+
+		$sourceInventory = $currentTile->getInventory();
+
+		$itemStack = $sourceInventory->getItem($sourceInventory::SLOT_RESULT);
+
+		if($itemStack->isNull()){
+			return false;
+		}
+
+		$singleItem = $itemStack->pop(1);
```
ColinHDev: same here
```suggestion
		$singleItem = $itemStack->pop();
```

ColinHDev (changes requested):

CoderJoeW:
@ColinHDev I made most of the changes

My only thing is im just not understanding your comments about pull vs push

ColinHDev:
Hopper pushing describes: An Hopper pushes an item from its own inventory into the inventory of the block it is facing.
Hopper pulling describes: An Hopper pulls an item from the inventory of the block above it into its own inventory.

The code in your methods `doHopperPush()` and `doHopperPull()` do the opposite: `doHopperPush()` pulls items into the hopper's inventory, `doHopperPull()` pushes items into the facing block's inventory.

CoderJoeW:
@ColinHDev That took me a minute to wrap my mind around I was not thinking about it in the sense of the hopper performing the action but the block performing the action

Anyways should be correct now

src/block/Furnace.php
```diff
@@ -90,4 +94,83 @@ public function onScheduledUpdate() : void{
...
+			if($itemStack->isNull()){
+				continue;
+			}
+
+			$singleItem = $itemStack->pop();
+
+			if($hopperFacing === Facing::DOWN && $targetInventory->canAddSmelting($singleItem)){
+				$this->transferItem($sourceInventory, $targetInventory, $singleItem, FurnaceInventory::SLOT_INPUT);
+				return true;
+			}elseif($hopperFacing !== Facing::DOWN && $hopperFacing !== Facing::UP && $targetInventory->canAddFuel($singleItem)){
```
ShockedPlot7560: Can be reduced by using ``Axis`` ?

src/block/Furnace.php
```diff
@@ -90,4 +94,83 @@ public function onScheduledUpdate() : void{
...
+				$this->transferItem($sourceInventory, $targetInventory, $singleItem, FurnaceInventory::SLOT_FUEL);
+				return true;
+			}
+		}
+
+		return false;
+	}
+
+	private function transferItem(Inventory $sourceInventory, Inventory $targetInventory, Item $item, int $slot) : void{
```
ShockedPlot7560, line 143: This isn't really the transfer logic for furnace but only for pushing system.

src/block/Furnace.php
```diff
@@ -90,4 +94,83 @@ public function onScheduledUpdate() : void{
...
+		$sourceInventory->removeItem($item);
+
+		$currentItem = $targetInventory->getItem($slot);
+
+		if($currentItem->isNull()){
+			$targetInventory->setItem($slot, $item);
+			return;
+		}
+
+		$currentItem->setCount($currentItem->getCount() + 1);
```
ShockedPlot7560, line 153: Why hardcoding ``+ 1`` here and not using the item count.

src/block/Hopper.php
ShockedPlot7560: It's not resolved. The hopper will always try to retrieve the items above before pushing the items. It should be the other way around

src/block/Hopper.php
```diff
@@ -92,7 +97,62 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+			$pushSuccess = $topBlock->doHopperPull($hopperBlock);
+		}
+
+		$facingBlock = $this->getSide($this->facing);
+		$pullSuccess = false;
+		if($facingBlock instanceof HopperInteractable){
+			$pullSuccess = $facingBlock->doHopperPush($hopperBlock);
+		}
+
+		$nextTick = ($pushSuccess || $pullSuccess ) ? 8 : 1;
```
ShockedPlot7560:
```suggestion
		$nextTick = ($pushSuccess || $pullSuccess) ? 8 : 1;
```

ShockedPlot7560 (changes requested):

ShockedPlot7560:
A lock principle must be implemented to prevent conflicts with cooldowns. It should not be implemented through updates alone.

If a hopper is placed on a tick, it will update in 8 ticks.
Suppose another hopper is placed 1 tick later under a chest and facing the previous hopper.
- The first time, it will retrieve the item from the chest above and schedule an update in 8 ticks.
- The first hopper does nothing and reschedules an update.
- The second hopper sends its item to the first hopper and schedules an update for itself and the other hopper. 
Thus, the first hopper is now subject to two updates, one tick apart.

src/block/Furnace.php
CoderJoeW: Not sure I understand this but I did make a commit to reduce the code here

src/block/Furnace.php
CoderJoeW: Maybe my lack of knowledge currently but I do not know how to accomplish the task any other way

src/block/Furnace.php
ShockedPlot7560: Instead of checking both DOWN and UP facing, it can be better to check an Axis, X ou Y

src/block/Furnace.php
ShockedPlot7560: ``$item->getCount()`` can be used.

src/block/Hopper.php
```diff
@@ -77,6 +83,9 @@ public function getSupportType(int $facing) : SupportType{
 	public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, ?Player $player = null) : bool{
 		$this->facing = $face === Facing::DOWN ? Facing::DOWN : Facing::opposite($face);
 
+		$world = $this->position->getWorld();
+		$world->scheduleDelayedBlockUpdate($blockReplace->position, 8);
```
ShockedPlot7560: Use your cooldown constant here

src/block/Hopper.php
```diff
@@ -92,7 +101,80 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+			return false;
+		}
+
+		$sourceInventory = $currentTile->getInventory();
+		$targetInventory = $tileHopper->getInventory();
+
+		return $this->transferItem($sourceInventory, $targetInventory);
+	}
+
+	private function hopperOnCooldown() : bool{
```
ShockedPlot7560: ``isOnCooldown`` may be better ?

src/block/Hopper.php
```diff
@@ -92,7 +101,80 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
-		//TODO
+		$world = $this->position->getWorld();
+
+		$hopperBlock = $this->position->getWorld()->getBlock($this->position);
+		if(!$hopperBlock instanceof Hopper){
+			return;
+		}
+
+		if($this->hopperOnCooldown()){
+			$this->currentCooldown--;
```
ShockedPlot7560: I think that rather than doing it this way, it might be better to store the time of the last interaction to allow verification.  

Thus, the hopper always updates at 8 tick when an action is performed (and updates without worrying about cooldown) and if, when updating, the difference isn't greater than the cooldown, we don't go any further.

src/block/inventory/FurnaceInventory.php
```diff
@@ -68,4 +68,30 @@ public function setFuel(Item $item) : void{
...
+
+		return $item->canStackWith($currentInput);
+	}
+
+	public function canAddFuel(Item $item) : bool{
+		if($item->getFuelTime() == 0) return false;
+
+		$currentFuel = $this->getFuel();
+
+		if($currentFuel->isNull()) return true;
```
ShockedPlot7560:
```suggestion
		if($currentFuel->isNull()){
				return true;
		}
```

src/block/inventory/FurnaceInventory.php
```diff
@@ -68,4 +68,30 @@ public function setFuel(Item $item) : void{
...
+
+		if($currentInput->getCount() >= $currentInput->getMaxStackSize()){
+			return false;
+		}
+
+		return $item->canStackWith($currentInput);
+	}
+
+	public function canAddFuel(Item $item) : bool{
+		if($item->getFuelTime() == 0) return false;
```
ShockedPlot7560:
```suggestion
		if($item->getFuelTime() === 0){
				return false;
		}
```

src/block/inventory/FurnaceInventory.php
ShockedPlot7560: Need to be changed

src/block/inventory/FurnaceInventory.php
```diff
@@ -84,13 +84,19 @@ public function canAddSmelting(Item $item) : bool{
 	}
 
 	public function canAddFuel(Item $item) : bool{
-		if($item->getFuelTime() == 0) return false;
+		if($item->getFuelTime() == 0){
```
dktapps:
```suggestion
		if($item->getFuelTime() === 0){
```
`==` should almost always be avoided.

src/block/Furnace.php
```diff
@@ -90,4 +94,83 @@ public function onScheduledUpdate() : void{
...
+
+		$singleItem = $itemStack->pop();
+
+		$sourceInventory->removeItem($singleItem);
+		$targetInventory->addItem($singleItem);
+
+		return true;
+	}
+
+	public function doHopperPush(Hopper $hopperBlock) : bool{
```
ShockedPlot7560, line 87: Maybe it's better to keep the same order for `doHopper...` as in the interface? So push and pull

src/block/Furnace.php
```diff
@@ -90,4 +94,83 @@ public function onScheduledUpdate() : void{
...
+			if($itemStack->isNull()){
+				continue;
+			}
+
+			$singleItem = $itemStack->pop();
+
+			if($hopperFacing === Facing::DOWN && $targetInventory->canAddSmelting($singleItem)){
+				$this->transferItem($sourceInventory, $targetInventory, $singleItem, FurnaceInventory::SLOT_INPUT);
+				return true;
+			}elseif($hopperFacing !== Facing::DOWN && $targetInventory->canAddFuel($singleItem)){
```
ShockedPlot7560, line 112: If its Facing down and you can't add Smelting, this condition can be true

src/block/Furnace.php
```diff
@@ -90,4 +94,83 @@ public function onScheduledUpdate() : void{
...
+		if(!$tileHopper instanceof TileHopper){
+			return false;
+		}
+
+		$sourceInventory = $tileHopper->getInventory();
+		$targetInventory = $currentTile->getInventory();
+
+		$hopperFacing = $hopperBlock->getFacing();
+
+		for($i = 0; $i < $targetInventory->getSize(); $i++){
```
ShockedPlot7560: This must use the size of the source inventory, not the target inventory.

ShockedPlot7560:
https://github.com/pmmp/PocketMine-MP/pull/5906#issuecomment-1657174520 

Still seems relevant

ShockedPlot7560:
> [#5906 (comment)](https://github.com/pmmp/PocketMine-MP/pull/5906#issuecomment-1657174520)
> 
> Still seems relevant

For your information, this PR is waiting for this comment.

CoderJoeW:
If someone wants to takeover and finish this would be great I am busy atm to finish

ShockedPlot7560:
> If someone wants to takeover and finish this would be great I am busy atm to finish

You haven't authorized the maintaner to push on your branch. I've taken over the PR locally, but I can't update your branch because of this permission problem.

[See here for the solution](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)

CoderJoeW:
@ShockedPlot7560 Couldnt figure out how to enable access from the link so I just sent you an invite to the project as a maintainer. I believe that should let you edit it

ShockedPlot7560:
> @ShockedPlot7560 Couldnt figure out how to enable access from the link so I just sent you an invite to the project as a maintainer. I believe that should let you edit it

For your knowledge, just go to your PR on the right there's a checkbox on PC, or at the very bottom for a mobile. For your futures PR, you can do this directly when you create them (it should be checked by default).

I'll look into it, thanks

ShockedPlot7560:
Only events need to be done and PR will be ready for review for me

ShockedPlot7560:
Storage in the cooldown tile and retrieval will be missing.

The only question I have is that in the wiki, a cooldown for the recovery of entities is indicated, but only the transfer of items is present in the tile. Something to think about.

PR #5912, opened 2023-07-18
Added handling for attack-air action

ipad54:
## Introduction
In vanilla when player hits air (performs a left-click-air action) this sound is played.

## API Changes
Added ```PlayerMissedSwingEvent```, called when player attacks air

## Tests
https://www.youtube.com/watch?v=iFfr5fehmjY

### DIFF
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -1899,7 +1899,7 @@ class Player extends Human implements CommandSender, ChunkListener, IPlayer{
 	 * Performs a left-click (attack) action on air and plays sound
 	 * if it's been successfully
 	 */
-	public function attackAir() : void{
+	public function missSwing() : void{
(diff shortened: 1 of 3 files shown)


dktapps:
It would be preferable to have a new `Player` method to handle this input, as well as an event for plugins to detect and cancel this sound.

src/player/Player.php
```diff
@@ -1894,6 +1895,18 @@ public function attackEntity(Entity $entity) : bool{
 		return true;
 	}
 
+	/**
+	 * Performs a left-click (attack) action on air and plays sound
+	 * if it's been successfully
+	 */
+	public function attackAir() : void{
+		$ev = new PlayerMissedSwingEvent($this);
```
dktapps, line 1903: The naming inconsistency feels weird here.

PR #4850, opened 2022-02-19
Ender eye placement in ender frame

ShockedPlot7560:
## Introduction

* Implemented Ender eye

### API changes
* Add Ender Eye in registry
* Add EndPortalFrameFillSound

## Backwards compatibility
N/A

## Tests

https://user-images.githubusercontent.com/66992287/154821093-ce338951-aab8-4d78-92e7-7d54f4051e17.mp4

PR #5913, opened 2023-07-18
Ender eye placement in portal frame

ShockedPlot7560:
## Introduction
This is an update/improvement of : #4850

### Behavioural changes
* Add an eye in a frame when a player interacts with it.

## Tests
https://github.com/pmmp/PocketMine-MP/assets/66992287/69cd6416-a0be-40c5-bacd-9ca1f68925c8


src/block/EndPortalFrame.php
```diff
@@ -58,4 +63,16 @@ public function getLightLevel() : int{
 	protected function recalculateCollisionBoxes() : array{
 		return [AxisAlignedBB::one()->trim(Facing::UP, 3 / 16)];
 	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if(!$item->equals(VanillaItems::ENDER_EYE()) || $this->eye){
```
dktapps: Less expensive check should be placed first. Also, this should use `ItemTypeIds`.

src/world/sound/EndPortalFillSound.php
```diff
@@ -0,0 +1,35 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\world\sound;
+
+use pocketmine\math\Vector3;
+use pocketmine\network\mcpe\protocol\LevelSoundEventPacket;
+use pocketmine\network\mcpe\protocol\types\LevelSoundEvent;
+
+class EndPortalFillSound implements Sound{
```
dktapps: `EndPortalFrame` would be better. `EndPortal` might refer to the portal block itself.

src/block/EndPortalFrame.php
```diff
@@ -58,4 +63,16 @@ public function getLightLevel() : int{
...
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if($this->eye || $item->getTypeId() !== ItemTypeIds::ENDER_EYE){
+			return false;
+		}
+		$world = $this->getPosition()->getWorld();
+		$world->setBlock($this->getPosition(), $this->setEye(true));
+		$world->addSound($this->getPosition(), new EndPortalFillSound());
+		$item->pop();
+
```
dktapps, line 70: A TODO should probably be added here to check if the portal is complete.

src/block/EndPortalFrame.php
```diff
@@ -58,4 +63,18 @@ public function getLightLevel() : int{
...
 		return [AxisAlignedBB::one()->trim(Facing::UP, 3 / 16)];
 	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if($this->eye || $item->getTypeId() !== ItemTypeIds::ENDER_EYE){
+			return false;
+		}
+		$world = $this->getPosition()->getWorld();
+		$world->setBlock($this->getPosition(), $this->setEye(true));
+		$world->addSound($this->getPosition(), new EndPortalFrameFillSound());
```
dktapps: We don't need (or want) to use getPosition here - a clone is neither wanted nor required

dktapps:
I stalled on this because I don't like the Eye of Ender being added as a generic item. From reading the code there's no obvious indication that there's more work to be done (e.g. adding projectile functionality).

Unsure if this might cause BC blockers for PM5 if additional functionality were to be implemented.

ShockedPlot7560:
I will not work anymore on this PR, feel free to take over

Issue #5883, opened 2023-07-07
Call an event when farmland decays to dirt due to missing hydration


Wertzui123:
## Description
There should be an event when farmland decays to dirt due to missing hydration.

## Justification
It is currently not possible to cancel this behavior (https://discord.com/channels/373199722573201408/373214753147060235/1126894765410091059), which is really inconvenient for map builders.

## Alternative methods
Allow overriding builtin blocks.

Wertzui123:
If anyone else needs this, there is a workaround for this problem:
```php
$this->getServer()->getWorldManager()->getWorldByName($world)->removeRandomTickedBlock(VanillaBlocks::FARMLAND());
```
Thanks to @JavierLeon9966 for pointing this out!

Wertzui123:
Closing this as it has been implemented in a5d8ef7a6c5ad93a385de8d3be92eef29e812494.

dktapps:
Thanks. Not sure why it didn't get automatically closed.

PR #5916, opened 2023-07-18
Add FarmlandHydrationChangeEvent

DebHugo:
## Introduction
Add suggestion of #5883 

### API changes
Add FarmlandHydrationChangeEvent class in `pocketmine\event\block` and modif `onRandomTick` function in Farmland class


dktapps:
`FarmlandHydrationChangeEvent` would be better. `wetness` sounds weird.

ShockedPlot7560 (dismissed):

src/event/block/FarmlandHydrationChangeEvent.php
```diff
@@ -0,0 +1,56 @@
...
+
+	public function __construct(
+		Block $block,
+		private int $oldWetness,
+		private int $newWetness,
+	){
+		parent::__construct($block);
+	}
+
+	public function getOldWetness() : int{
```
dktapps: This should be named `getOldHydration()` to be consistent with the event name

src/event/block/FarmlandHydrationChangeEvent.php
```diff
@@ -0,0 +1,56 @@
...
+		return $this->oldWetness;
+	}
+
+	public function getNewWetness() : int{
```
dktapps: Same again here

src/event/block/FarmlandHydrationChangeEvent.php
```diff
@@ -0,0 +1,56 @@
...
+use pocketmine\block\Farmland;
+
+/**
+ * Called when farmland wetness is updated.
+ */
+class FarmlandHydrationChangeEvent extends BlockUpdateEvent{
+
+	public function __construct(
+		Block $block,
+		private int $oldWetness,
```
ShockedPlot7560: Could be better to rename it `$oldHydration` ?

src/event/block/FarmlandHydrationChangeEvent.php
```diff
@@ -0,0 +1,56 @@
...
+		private int $newWetness,
```
ShockedPlot7560: Same here

src/event/block/FarmlandHydrationChangeEvent.php
```diff
@@ -0,0 +1,56 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\event\block;
+
+use pocketmine\block\Block;
+use pocketmine\block\Farmland;
+
+/**
+ * Called when farmland wetness is updated.
```
ShockedPlot7560:
```suggestion
 * Called when farmland hydration is updated.
```

src/event/block/FarmlandHydrationChangeEvent.php
```diff
@@ -0,0 +1,56 @@
...
+
+namespace pocketmine\event\block;
+
+use pocketmine\block\Block;
+use pocketmine\block\Farmland;
+
+/**
+ * Called when farmland hydration is updated.
+ */
+class FarmlandHydrationChangeEvent extends BlockUpdateEvent{
```
dktapps: Why is this extending BlockUpdateEvent?

DebHugo: Because I'm making a mistake. I change this now to extend `BlockEvent`

PR #5923, opened 2023-07-20
Allowed players to eat in creative mode and difficulty peaceful

NTT1906:
## Introduction
Allowed players to eat in creative mode and difficulty peaceful.

- Added **`Player::canEatWithoutHunger()`.**

## Backwards compatibility
Maybe?

## Tests

https://github.com/pmmp/PocketMine-MP/assets/54394881/961b72ef-62e2-4e46-a470-844cf21482e9


src/entity/Human.php
```diff
@@ -190,11 +190,10 @@ public function getHungerManager() : HungerManager{
 		return $this->hungerManager;
 	}
 
-	public function consumeObject(Consumable $consumable) : bool{
-		if($consumable instanceof FoodSource && $consumable->requiresHunger() && !$this->hungerManager->isHungry()){
+	public function consumeObject(Consumable $consumable, bool $forceEating = false) : bool{
+		if (($consumable instanceof FoodSource) && !$forceEating && $consumable->requiresHunger() && !$this->getHungerManager()->isHungry()){
```
ShockedPlot7560: Perhaps you should put it on several lines to make the condition more legible.

src/player/Player.php
```diff
@@ -673,6 +674,13 @@ public function setUsingItem(bool $value) : void{
 		$this->networkPropertiesDirty = true;
 	}
 
+	/**
+	 * Allow eating
+	 */
+	public function forceAllowEating() : bool{
```
ShockedPlot7560: ``allowForceEating`` seems better

src/player/Player.php
```diff
@@ -2000,6 +2008,10 @@ public function emote(string $emoteId) : void{
 		}
 	}
 
+	public function consumeObject(Consumable $consumable, bool $forceEating = false) : bool{
+		return parent::consumeObject($consumable, $this->forceAllowEating() || $forceEating);
```
ShockedPlot7560: Prefer conditions that consume less first, i.e. check the variable before the function.

src/entity/Human.php
```diff
@@ -190,11 +190,10 @@ public function getHungerManager() : HungerManager{
...
 			return false;
 		}
-
```
ShockedPlot7560, line 196: Unnecessary change

src/entity/Human.php
ShockedPlot7560: Plus prefer conditions that consume less first like the variable check before the instance.

ShockedPlot7560 (changes requested):

src/entity/Human.php
dktapps: In practice the difference between the two is likely to be negligible, but in this case the instanceof is related to the later part of the condition also.

dktapps:
Not a fan of the signature changes here. I think this should be doable without API changes.

src/entity/Human.php
NTT1906: > Plus prefer conditions that consume less first like the variable check before the instance.

Can you explain it further?

NTT1906:
By the way, how to automatically delete the use path when not using (e.g: use pocketmine\entity\Consumable;) in PhpStorm?

src/entity/Human.php
```diff
@@ -191,8 +191,13 @@ public function getHungerManager() : HungerManager{
 	}
 
 	public function consumeObject(Consumable $consumable) : bool{
-		if($consumable instanceof FoodSource && $consumable->requiresHunger() && !$this->hungerManager->isHungry()){
-			return false;
+		if ($consumable instanceof FoodSource) {
+			if ($this instanceof Player && $this->allowForceEating()) {
```
dktapps: `$this instanceof` is a sure sign that your code design is flawed.

src/entity/Human.php
```diff
@@ -170,6 +170,10 @@ public function sendSkin(?array $targets = null) : void{
 		]);
 	}
 
+	public function allowForceEating () : bool{
```
ShockedPlot7560:
```suggestion
	public function allowForceEating() : bool{
```

src/entity/Human.php
```diff
@@ -191,8 +195,13 @@ public function getHungerManager() : HungerManager{
 	}
 
 	public function consumeObject(Consumable $consumable) : bool{
-		if($consumable instanceof FoodSource && $consumable->requiresHunger() && !$this->hungerManager->isHungry()){
-			return false;
+		if ($consumable instanceof FoodSource) {
```
ShockedPlot7560: Code style, no space before and after parentheses

src/entity/Human.php
```diff
@@ -191,8 +195,13 @@ public function getHungerManager() : HungerManager{
...
+			if ($this->allowForceEating()) {
```
ShockedPlot7560: Same here

src/entity/Human.php
```diff
@@ -191,8 +195,13 @@ public function getHungerManager() : HungerManager{
...
+				return parent::consumeObject($consumable);
+			}
+			if ($consumable->requiresHunger() && !$this->getHungerManager()->isHungry()) {
```
ShockedPlot7560: Same here

src/player/Player.php
```diff
@@ -673,6 +673,10 @@ public function setUsingItem(bool $value) : void{
 		$this->networkPropertiesDirty = true;
 	}
 
+	public function allowForceEating () : bool{
```
ShockedPlot7560:
```suggestion
	public function allowForceEating() : bool{
```

src/entity/Human.php
```diff
@@ -191,8 +195,13 @@ public function getHungerManager() : HungerManager{
 	}
 
 	public function consumeObject(Consumable $consumable) : bool{
```
ShockedPlot7560, line 196: Can be simplified to have only two returns instead of duplicating the call to the parent

ShockedPlot7560 (changes requested):
I find the name of the added function rather odd, but I wouldn't know how to replace it.

src/player/Player.php
dktapps: I'm not a fan of this name. It feels way too specific.

NTT1906: > I'm not a fan of this name. It feels way too specific.

what is your idea?

NTT1906: canPlayerConsume

canPlayerSnack
canPlayerChew
canPlayerDevour
isPlayerFamished

allowForceFeast
canFreelyEating

ShockedPlot7560: Could `canEatAnytime` better ? It can't include `player` because it's linked to Human and not only Player.

dktapps: None of these address my original comment.
I'm thinking more along the lines of `canEat()`, which covers the creative case and hunger together.

NTT1906: `canEat()` seems too general, but I too go with that.

NTT1906: @ShockedPlot7560 Can I go with "canEatWithoutHunger"?

NTT1906: I think I will go with "canEatWithoutHunger"

dktapps (changes requested):
I don't like this solution. I think a better solution would be to have a switch to enable/disable HungerManager, and a way for the World to notify players when the difficulty is changed. This would allow plugins more power over it rather than hardlocking the logic into a random `Player` function.

dktapps:
Superseded by https://github.com/pmmp/PocketMine-MP/commit/12214792b32a63a0668a2ecbc175f148bd7fe222

Issue #3473, opened 2020-05-07
Remove PlayerCreationEvent


SOF3:
Applications of `PlayerCreationEvent` are basically as unreliable as using reflections. PocketMine is currently not semver-compliant because `Player` internal semantics are not polymorphism-friendly enough; most changes on `Player` methods do not consider BC with plugins using `PlayerCreationEvent`. Hence, supporting `PlayerCreeationEvent` is a maintainability burden and should be removed from the API.

Plugins currently using `PlayerCreationEvent` should seek alternative methods. In particular, for public plugins, it is always preferrable to send pull requests to extend PocketMine's API, while private developers unwilling to contribute can still patch the source directly (this is unreliable, but `PlayerCreationEvent` is unreliable as well).

dktapps:
what "alternative methods"? the whole reason I haven't put forth this proposal yet myself is because there isn't any suitable alternative methods for some scenarios.

> most changes on Player methods do not consider BC with plugins using PlayerCreationEvent

citation needed; this is actually the exact reason why #3167 has not yet been merged.

Sending PRs is fool's gold, since usually the changes needed are still specialized to one plugin or another anyway (e.g. economy plugins adding getBalance() or similar), and also because of the duration of time it takes for a PR to get from initial creation to release. That said, any application which doesn't require directly _modifying_ Player functionality doesn't belong in a custom Player class anyway.

dktapps:
Some additional things to consider, which aren't mentioned in the main issue:

- Users of `PlayerCreationEvent` are non-cooperative. Only one plugin using custom player classes can operate at a time. This goes against PM's entire plugin architecture, which is designed from ground-up to be cooperative.
  - This means that custom player classes cannot be used in public plugins, since public plugins can't know whether they'll be used in isolation or alongside other plugins.
  - Following from the above point, it's clear that only private core plugins can benefit from this feature. If the required use case is _so_ specialized that it can't be filled by a general-purpose PM API addition, private servers can fork PM and modify it to their desires anyway.
- The intended, original use case for custom player classes was to allow overriding `handleDataPacket()` in the old days when packet handling was embedded into `Player`. This is no longer the case since many years.
- The only remaining _legitimate_ use cases in PM4 for custom `Player` classes are to override `protected` or `public` methods of `Player` to alter behaviour that isn't directly exposed, or to intercept stuff that don't currently have events associated with them.
[...]

PR #5932, opened 2023-07-23
Allow to replace already registered Block and Item classes and serializers

S3v3Nice:
## Introduction
Sometimes it is very necessary to make own implementation of some block or item, which is already registered in Pocketmine. For example, to implement an enchantment table with its own logic of enchanting an item. The changes given are minimal and can't harm in any way, but can be very useful for some developers (me included).

### API changes
Added optional boolean argument $replace (false by default) for multiple methods that register block / item / serializer / deserializer.

## Backwards compatibility
Full backward compatibility

### DIFF
--- a/src/block/RuntimeBlockStateRegistry.php
+++ b/src/block/RuntimeBlockStateRegistry.php
@@ -84,12 +84,12 @@ class RuntimeBlockStateRegistry{
 	 * Maps a block type's state permutations to its corresponding state IDs. This is necessary for the block to be
 	 * recognized when fetching it by its state ID from chunks at runtime.
 	 *
-	 * @throws \InvalidArgumentException if the desired block type ID is already registered
+	 * @throws \InvalidArgumentException if the desired block type ID is already registered and $replace is set to false
 	 */
-	public function register(Block $block) : void{
+	public function register(Block $block, bool $replace = false) : void{
 		$typeId = $block->getTypeId();
 
-		if(isset($this->typeIndex[$typeId])){
+		if(!$replace && isset($this->typeIndex[$typeId])){
 			throw new \InvalidArgumentException("Block ID $typeId is already used by another block");
 		}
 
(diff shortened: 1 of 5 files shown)


JavierLeon9966:
Those items/blocks that don't have any vanilla features should be implemented in pocketmine itself by a PR, not by letting developers doing it in a plugin that will give them the feature to some people instead of here where everyone would have it.

S3v3Nice:
> Those items/blocks that don't have any vanilla features should be implemented in pocketmine itself by a PR, not by letting developers doing it in a plugin that will give them the feature to some people instead of here where everyone would have it.

I'm not talking about implementing vanilla features, I'm talking about implementing my custom functionality for vanilla blocks. In PM 4 you could do it, now you can't (maybe with events, but it's a very inconvenient and restricted solution).

S3v3Nice:
> Those items/blocks that don't have any vanilla features should be implemented in pocketmine itself by a PR, not by letting developers doing it in a plugin that will give them the feature to some people instead of here where everyone would have it.

Also, speaking specifically about vanilla features.... Some vanilla blocks and items are not yet registered in Pocketmine, and can be added by plugins. So here too you can say that people will make plugins to add these things instead of creating pull requests to implement in the core (For example, your `VanillaElytra` plugin. Why don't you add the elytra functionality to the core, but create your own plugin?). There is essentially no difference between blocks and items that are registered but not working and those that are not registered. But there is no way to implement already registered ones in your own way, except by pull request ;/

ShockedPlot7560:
> Sometimes it is very necessary to make own implementation of some block or item, which is already registered in Pocketmine. 

This logic must now be done via break / place / interaction events etc. And not via block replacement as in pm4.

dktapps:
Support for this was intentionally removed in PM5, due to the long-standing abuse of this anti-feature to implement absurdly overengineered solutions to problems in the name of "clean code" and "performance", without any actual evidence to suggest that such things are beneficial. Just the other day I dealt with a user on Discord who implemented a custom item and entity class just to override `Projectile->onProjectileHit()` instead of simply handling `ProjectileHitEvent` (the function even accepts the event directly...). There are many many more examples of this.

Most things that overrides were used for can and should be done using events, and when that's not the case, it's indicative of an API shortfall that calls for new features to be added, such as in the case of #5799. Overriding builtin stuff is invasive and non-cooperative, and has never been properly supported. It's much the same problem we have with custom player classes as discussed in #3473.

The following cases are currently unsolved problems:
- Entity intersections / collisions for portals, pressure plates etc.
- Changing the flow rate of lava in the nether

Both of these can be solved with improvements to the API, rather than allowing these plugin abuses to continue.

PR #5936, opened 2023-07-24
Implement PlayerPressurePlateTriggerEvent

AIPTU:
## Introduction
The `PlayerPressurePlateTriggerEvent` is an event that occurs when a player steps on or off a pressure plate block in the game world. It carries information about the player who triggered it and the specific pressure plate block that was activated. Other plugins or code can listen for this event and perform actions when it happens. Additionally, the event can be cancelled, preventing further processing, if needed.

### API changes
Add `PlayerPressurePlateTriggerEvent`

## Tests
```php
public function onPressurePlateTrigger(PlayerPressurePlateTriggerEvent $event) {
    $player->sendMessage("You triggered a pressure plate!");
}
```

### DIFF
--- a/src/block/PressurePlate.php
+++ b/src/block/PressurePlate.php
@@ -61,5 +61,20 @@ abstract class PressurePlate extends Transparent{
 		}
 	}
 
-	//TODO
+	public function onEntityInside(Entity $entity) : bool{
+		if($entity instanceof Player){
+			$ev = new PlayerPressurePlateTriggerEvent($entity, $this);
+			$ev->call();
+
+			if($ev->isCancelled()){
+				return false;
+			}
+		}
+
+		return parent::onEntityInside($entity);
+	}
+
+	public function hasEntityCollision() : bool{
+		return true;
+	}
 }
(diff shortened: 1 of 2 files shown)


dktapps:
I don't see the point of this change. There's nothing actually happening as a result of this event one way or another.

I also don't see why it should be constrained to just players.

dries-c:
It's nice for things like parkour plugins, where you can trigger a checkpoint or an end result for stepping onto a pressure plate. In the NG fork, I've also added this.

dktapps:
The pressure plate is not actually being triggered though. There is nothing happening as a result of this event being fired.

dktapps:
Not to mention the fact that it will spam the event every tick for as long as the player is within its full block range (also incorrect, the activation range is much less than a full block)

dries-c:
Well ideally it should only be called when walking into the block when it's not pressed down.

src/block/PressurePlate.php
```diff
@@ -61,5 +63,20 @@ public function onNearbyBlockChange() : void{
 		}
 	}
 
-	//TODO
+	public function onEntityInside(Entity $entity) : bool{
+		if($entity instanceof Player){
```
ShockedPlot7560, line 67: This should check the status of the current pressure plate, otherwise it will launch the event every tick.

dktapps:
Closing in favour of #5991

PR #5937, opened 2023-07-24
Modify and improve cake behavior to be similar to vanilla.

NTT1906:
## Introduction
This PR version makes the cake work compatible with vanilla behaviors.

## Changes
- Cake can only be able to add candle when it's not used yet, `$this-bites === 0`.
- Cake drops nothing when destroyed.
- When supporting blocks get destroyed, it will show breaking particles instead of just disappearing like before.
- CakeWithCandle can be lit up using fire charge.
- Cake does not get eaten when trying to extinguish the candle.
- Added a calculation function to check if the player trying to extinguish the candle or eat the cake.
### API changes

No
### Behavioural changes

No
## Backwards compatibility
Maybe?

## Tests
https://github.com/pmmp/PocketMine-MP/assets/54394881/f6d3f828-d4b6-4fdd-9ed1-c6d2cc92771a

### DIFF
--- a/.github/workflows/discord-release-notify.yml
+++ b/.github/workflows/discord-release-notify.yml
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Setup PHP and tools
-        uses: shivammathur/setup-php@2.25.2
+        uses: shivammathur/setup-php@2.25.4
         with:
           php-version: 8.1
 
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -18,7 +18,7 @@ jobs:
           submodules: true
 
       - name: Setup PHP
-        uses: shivammathur/setup-php@2.25.2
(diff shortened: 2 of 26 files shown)


dktapps:
This PR contains multiple unrelated changes that should be submitted separately

NTT1906:
> This PR contains multiple unrelated changes that should be submitted separately

The current code exhibits numerous small disturbances, some of which arise from inconveniences while others stem from outdated code. Personally, I find it difficult to resolve each problem individually. The presence of these inconveniences becomes disturbing, and my motivation wanes if prompt action is not taken. Consequently, I exert every effort to rectify the issues swiftly, striving to achieve my desired outcome.

Although I acknowledge that separating the problems would streamline maintenance and help future tracks, I still struggle to fully embrace this approach. I already have this kind of problems before... I kinda got frustratd about it :l

Would you have any tips for me?

dktapps:
Closing due to lack of activity on review comments and being packed with unwanted changes.

PR #5940, opened 2023-07-24
Implemented pink petals

ipad54:
## Introduction
Implemented pink petals

## Tests
https://www.youtube.com/watch?v=wdpcff7A8Yk


src/block/PinkPetals.php
```diff
@@ -0,0 +1,124 @@
...
+	/** @return $this */
+	public function setCount(int $count) : self{
+		if($count < self::MIN_COUNT || $count > self::MAX_COUNT){
+			throw new \InvalidArgumentException("Count must be in range " . self::MIN_COUNT . " ... " . self::MAX_COUNT);
+		}
+		$this->count = $count;
+		return $this;
+	}
+
+	private function canBeSupportedBy(Block $block) : bool{
```
dktapps: `canBeSupportedAt` is the preferred way for `minor-next` (see changes in 5899f2fc1d39175ce7bf0672cc6c09514779b5f2)

src/block/PinkPetals.php
```diff
@@ -0,0 +1,124 @@
...
+		//TODO: Moss block
```
dktapps, line 64: This seems to be coming up a lot. Are they difficult to implement? I haven't really looked into it.

ipad54: Moss spread logic can be a little bit difficult

NhanAZ:
Try break pink petals with count 4 in vanilla (survival mode), you will feel the difference when pink petals drop, In vanilla pink petals with count 4 when broken will drop 1 pink petals in 4 directions instead of dropping 1 stack pink petals with count 4.

dktapps:
> Try break pink petals with count 4 in vanilla (survival mode), you will feel the difference when pink petals drop, In vanilla pink petals with count 4 when broken will drop 1 pink petals in 4 directions instead of dropping 1 stack pink petals with count 4.

This is a general issue with item drops and not relevant to this PR.

src/block/PinkPetals.php
ipad54: > Moss spread logic can be a little bit difficult

To implement moss spread logic properly, firstly ```Azalea```, ```Flowering Azalea```, ```Moss Carpet``` should be implemented, because moss can generate these blocks during spreading

dktapps: Right, that makes sense. I suppose those blocks need to get bumped in priority then.

src/block/PinkPetals.php
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	private function grow(?Player $player) : bool{
+		if($this->count < self::MAX_COUNT){
+			$ev = new BlockGrowEvent($this, (clone $this)->setCount($this->count + 1), $player);
+			$ev->call();
+			if($ev->isCancelled()){
+				return false;
+			}
+			$this->position->getWorld()->setBlock($this->position, $ev->getNewState());
```
dktapps: Please update this to use the new `BlockEventHelper::grow()` API

src/block/PinkPetals.php
```diff
@@ -98,15 +99,9 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
-			$ev->call();
-			if($ev->isCancelled()){
-				return false;
-			}
-			$this->position->getWorld()->setBlock($this->position, $ev->getNewState());
-		}else{
-			$this->position->getWorld()->dropItem($this->position->add(0, 0.5, 0), $this->asItem());
+			return BlockEventHelper::grow($this, (clone $this)->setCount($this->count + 1), $player);
 		}
+		$this->position->getWorld()->dropItem($this->position->add(0, 0.5, 0), $this->asItem());
```
dktapps: I'm not sure I understand why this is in the grow() function.

ipad54: ```If there are already four pink petals in the block, it will drop a copy of itself```

I thought that this mechanic is also part of the growth process

Or is it better to separate this from the grow function?

dktapps: On closer look the grow function seems completely redundant, since there's no random-tick growth.

However, if there was random-tick growth, I wouldn't expect this to be dropping items, otherwise we'd get random dropped items scattered across the world.

dktapps (approved):

Issue #4552, opened 2021-11-05
Add PlayerChunkPreSendEvent


dktapps:
## Description
This would allow changing the chunk that gets sent, which would be useful for doing anti-xray without packet hacks. (and probably in a more performant way, too.)

## Justification
Right now, it's quite a pain in the ass to implement anti-xray, especially on PM4, because there are no events that you can intercept in order to alter the chunk that gets sent. The `ChunkCache` will always send the chunk that is in the global `World`, and doesn't offer any way to change this.

Exposing APIs to modify `ChunkCache`'s behaviour is a bad idea, since it's internal network architecture and not covered by API backwards compatibility promise.

Therefore, a logical and protocol-agnostic way of supporting something like this is an event, which allows you to replace the chunk that gets sent with a custom one.

## Caveats
Chunks are very expensive objects - they have a large memory footprint. Care would need to be taken to ensure that modifying the chunk found in the event for sending doesn't corrupt the actual terrain.

## Alternative methods
Honestly, I don't actually know if there is another way to do this on PM4 right now. You can't even use `DataPacketSendEvent` for this since you'll never receive `LevelChunkPacket` - it's prepared and sent in a different way than normal packets.

ColinHDev:
Possible in PM4, but very hacky
Involves custom player class, some ReflectionProperties, etc.

SOF3:
What about overlay worlds? Instead of exposing the entire chunk for modification, could we just let developers register a world layer that overrides specific blocks for individual players? An implementation for anti-xray would be like this:

```mermaid
graph TD
t1[nearby ore blocks\nindividually maintained for each player]
t2[all ore blocks replaced to stone\nshared by all non-admins]
t3[base world]
t1 --> t2 --> t3
```

dktapps:
I've considered something like this before, but it's not likely to be practical in a minor version

PR #5941, opened 2023-07-25
Removed PlayerCreationEvent

### LINKED ISSUE #3473: Remove PlayerCreationEvent

JavierLeon9966:
## Introduction

See https://github.com/pmmp/PocketMine-MP/issues/3473#issuecomment-1146081554
### Relevant issues

Fixes #3473 
### API changes

None
### Behavioural changes

None
## Backwards compatibility

Removed `PlayerCreationEvent`
## Follow-up

None
## Tests

N/A


dktapps:
Yeah, this is a big one. Not sure how many people we'll piss off with this change...

Lunarelly:
Don't do that. Even if pocketmine will implement some events for knockback and hits (delay specifically), it still will be slower and will affect the gameplay. Changing logic of knockback, hit delay directly into Player's class will be easier & faster.
Theres no valid point to remove it. You still can add those events, but listen to me, you can always just NOT accept plugins that using this feature (OR do a notice in description that it may break something), but keep it for private use. (...or just throw an error if multiple plugins changing player's class)

ionaqwyxa:
pmmp not ready for this

JavierLeon9966:
> Don't do that. Even if pocketmine will implement some events for knockback and hits (delay specifically), it still will be slower and will affect the gameplay. Changing logic of knockback, hit delay directly into Player's class will be easier & faster.

How much slower will it be? Is this just premature optimizing? And you don't know if a new event will make changing the KB harder than if it were to modify the player class.
> Theres no valid point to remove it. You still can add those events, but listen to me, you can always just NOT accept plugins that using this feature (OR do a notice in description that it may break something), but keep it for private use. (...or just throw an error if multiple plugins changing player's class)

Did you even read the original issue?

dktapps:
This has already been discussed at length in #internals-dev and in the linked comment on #3473. Please go to one of those places to discuss this.

JavierLeon9966:
> pmmp not ready for this

Removing the player class will allow pocketmine to find the hidden issues like missing events. In the original issue is aware of this like PlayerPostChunkEvent.

ShockedPlot7560:
Before removing this possibility, I think it would be important and correct to look at all the (useful) uses of this event. These uses would then show that the API lacks a way of doing certain things that should be implemented before even removing this. 
 
I'm thinking in particular of : 
- acceptance of Translatable in several functions such as sendPopup, sendTitle, sendToast, etc ...
- the possibility for a plugin to add its own translation lines. (#5864 probably can fix that ?)
- modification of chunk sending, for an anti-xray for example (#4552)

These are just some of the cases I've come across, but there are surely others that require a great deal of thought?

dadodasyra:
A lot of these uses cases are personnal and may be overkill/"useless", some may even be possible to implement with the API, but at least I try so that "we" don't forget anything before removing player overriding :
- The old trick used to replace " " by "_" in playername would become a pain to implement
- In my case I implement a custom translation system that adds a getLang on the player, ShockedPlot is right on it
- The implementation of some custom items needs to trick weirdly PM to works (ex. onUpdate is modified on my side to force call "breakBlock" in the playerclass when the blockBreakHandler is finished, otherwise the block is never broken as the client doesnt send necessary packets for
- Some crashs were caused by sendMessage/kick on Player offline, instead of always checking if the player is online everywhere in my code it was easier to just add a isOnline in sendMessage in player class
- Since PM4 (or PM5 ?) root operators doesnt have all permissions if these permissions weren't registered before boot, I added a check in hasPermission to change this
- The transfer function is overriden to support my proxy things only if the ip match my servers ip (guess it's possible to do it with the API)
- starting breaking block and stop breaking block (attackBlock, stopBreakBlock) aren't available in event unless by listening to every packets
- Overriding knockback is way easier (and permisive ?) by doing it from the player class
- I'm adding a cooldown on toast Notification because the client will stack them
- spawnTo can be modified to easily implement a vanish system
- removeCurrentWindow was crashing sometimes due to craftingGrid being not set ? so.. I check it myself.. 
- Implementing a scale different for the player itself and for other players is easier to implement in player class
- Implementing synchronization for proxy transfers; inventory saves and a lot of other stuffs is easier with overriden player class events

And I think the more useful thing with this overriding stuff is that you can add your custom variables to the player class that is used everywhere. Removing it mean every custom data you want to save would be only accessible by always getting a "customplayer" object from an API which returns it from a list. Not impossible but "less pretty". 

However player class overriding is infernous for IDEs like PHPStorm which doesnt understand that your CustomPlayer == Player and the developer is forced to cast it to its own class every time (or modify his PM bins with it).
I guess if it's implemented at least you should keep PlayerCreationEvent but only for information and without any way to override the Player class

ShockedPlot7560:
Most of what you say is easily done via sessions.  
It's not a question of knowing too much about what can be done more easily thanks to this functionality. Rather, it's a question of knowing what couldn't be done without using a custom class. What I mean by other would be not having to go through Reflection or other tricky manipulations.

SOF3:
@dadodasyra everything you mentioned can be done even more easily by modifying the PocketMine source code directly. And we all know this is a bad idea (I hope so).

And the core argument in #3473 and #5941 is that using PlayerCreationEvent is only barely better than modifying the source code directly.

dktapps:
If anything, modifying the code directly is probably better, as it doesn't create any type ambiguity for events or pretend to be cooperative

dktapps:
> I guess if it's implemented at least you should keep PlayerCreationEvent but only for information and without any way to override the Player class

Yeah, this might be an ok solution.

I won't address the other "use cases" listed here as they've been talked to death about in #3473. My aim is to make all of the legit cases doable using other API.

dktapps:
Closing this as it's a trivial change to do when the time comes. Please comment on #3473 if you have further inputs.

PR #5942, opened 2023-07-25
Added animation to missSwing

### LINKED ISSUE #4598: No animation when hitting the air
### Issue description

- Expected result: Animation of the hit when hitting the air
- Actual result: No animation is performed

1. Hit air

### OS and versions

* PocketMine-MP: 4.0.0-BETA13+dev
* PHP: 8.0.12
* Using JIT: no
* Server OS: Linux
* Game version: Win10

### Plugins
n/a

- If you remove all plugins, does the issue still occur? yes
  - Have you asked for help on our forums before creating an issue? yes
  - Can you provide sample, *minimal* reproducing code for the issue? no

rasu3n:
## Introduction

This adds the feature to play animation to the missSwing method added in #5912

### Relevant issues

* Fixes #4598 

## Tests

https://github.com/pmmp/PocketMine-MP/assets/76860328/05233a96-3b97-4b37-910c-1c8bf91214fb

### DIFF
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -1899,11 +1899,14 @@ class Player extends Human implements CommandSender, ChunkListener, IPlayer{
 	 * Performs actions associated with the attack action (left-click) without a target entity.
 	 * Under normal circumstances, this will play the no-damage attack sound and nothing else.
 	 */
-	public function missSwing() : void{
+	public function missSwing(bool $broadcastAnimation) : void{
 		$ev = new PlayerMissedSwingEvent($this);
 		$ev->call();
 		if(!$ev->isCancelled()){
 			$this->broadcastSound(new EntityAttackNoDamageSound());
+			if($broadcastAnimation){
+				$this->broadcastAnimation(new ArmSwingAnimation($this), $this->getViewers());
+			}
 		}
 	}
 
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -1897,7 +1897,7 @@ class Player extends Human implements CommandSender, ChunkListener, IPlayer{
 
 	/**
 	 * Performs actions associated with the attack action (left-click) without a target entity.
-	 * Under normal circumstances, this will play the no-damage attack sound and animation and nothing else.
+	 * Under normal circumstances, this will just play the no-damage attack sound and the arm-swing animation.
 	 */
 	public function missSwing() : void{
 		$ev = new PlayerMissedSwingEvent($this);
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -1897,16 +1897,17 @@ class Player extends Human implements CommandSender, ChunkListener, IPlayer{
 
 	/**
 	 * Performs actions associated with the attack action (left-click) without a target entity.
-	 * Under normal circumstances, this will play the no-damage attack sound and nothing else.
+	 * Under normal circumstances, this will play the no-damage attack sound and animation and nothing else.
 	 */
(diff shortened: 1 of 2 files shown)


src/player/Player.php
```diff
@@ -1899,11 +1899,14 @@ public function attackEntity(Entity $entity) : bool{
 	 * Performs actions associated with the attack action (left-click) without a target entity.
 	 * Under normal circumstances, this will play the no-damage attack sound and nothing else.
 	 */
-	public function missSwing() : void{
+	public function missSwing(bool $broadcastAnimation) : void{
 		$ev = new PlayerMissedSwingEvent($this);
```
rasu3n, line 1903: Should I pass $broadcastAnimation parameters to this event? (and add getter and setter to event)

dktapps:
I don't see why this needs a parameter.

rasu3n:
> I don't see why this needs a parameter.

This is because the animation is not played when using touch control in vanilla.
Is it better for the Player class to have an InputMode property?

dktapps:
> > I don't see why this needs a parameter.
> 
> This is because the animation is not played when using touch control in vanilla. Is it better for the Player class to have an InputMode property?

This sounds like a client bug to me.

dktapps:
https://bugs.mojang.com/browse/MCPE-107865

rasu3n:
> > > I don't see why this needs a parameter.
> > 
> > 
> > This is because the animation is not played when using touch control in vanilla. Is it better for the Player class to have an InputMode property?
> 
> This sounds like a client bug to me.

It sounds strange, but it has been doing this for a long time.

dktapps:
> It sounds strange, but it has been doing this for a long time.

That doesn't mean it's not a bug, and it also doesn't mean that PM should be treating this any differently.

src/player/Player.php
```diff
@@ -1899,11 +1899,14 @@ public function attackEntity(Entity $entity) : bool{
 	 * Performs actions associated with the attack action (left-click) without a target entity.
 	 * Under normal circumstances, this will play the no-damage attack sound and nothing else.
```
ipad54: Please also update this doc-block

src/player/Player.php
```diff
@@ -1897,16 +1897,17 @@ public function attackEntity(Entity $entity) : bool{
 
 	/**
 	 * Performs actions associated with the attack action (left-click) without a target entity.
-	 * Under normal circumstances, this will play the no-damage attack sound and nothing else.
+	 * Under normal circumstances, this will play the no-damage attack sound and animation and nothing else.
```
```suggestion
	 * Under normal circumstances, this will just play the no-damage attack sound and the arm-swing animation.
```

src/player/Player.php
```diff
@@ -1897,16 +1897,17 @@ public function attackEntity(Entity $entity) : bool{
...
 		$ev = new PlayerMissedSwingEvent($this);
 		$ev->call();
 		if(!$ev->isCancelled()){
 			$this->broadcastSound(new EntityAttackNoDamageSound());
-			if($broadcastAnimation){
-				$this->broadcastAnimation(new ArmSwingAnimation($this), $this->getViewers());
-			}
+			// https://bugs.mojang.com/browse/MCPE-107865
+			// In vanilla, this animation doesn't play when using touch controls, but
+			// it's marked as unresolved, so it does here
```
dktapps: I don't think we need to mention this. This is a bug in vanilla that doesn't affect PM.

ipad54 (approved):

dktapps (approved):

PR #5944, opened 2023-07-26
define RegistryTrait::__callStatic() var types

Lunarelly:


dktapps:
Again, this is BC-breaking and can't target stable.

PR #5946, opened 2023-07-26
Allow different axes to have different knockback forces

JavierLeon9966:
## Introduction

There's no way to modify the knockback of a player, or living entity, without overriding them and causing conflicts. This pull request aims to let plugins easily modify specific knockback forces so they can implement their custom knockback easily without overriding them. Another reason is that iron golems and holgins/zoglins have vertical knockback and these changes will allow them to be implemented more easily.
### Relevant issues

None
### API changes

The following methods' signatures were changed:
- Changed `EntityDamageByEntityEvent->__construct(Entity $damager, Entity $entity, int $cause, float $damage, array $modifiers = [], private float $knockBack = 0.4, private ?float $verticalKnockBackLimit = 0.4)` to `EntityDamageByEntityEvent->__construct((Entity $damager, Entity $entity, int $cause, float $damage, array $modifiers = [], private Vector3 $knockBack = new Vector3(0.4, 0.4, 0.4), private ?float $verticalKnockBackLimit = 0.4)`
- Changed `EntityDamageByEntityEvent->getKnockBack(): float` to `EntityDamageByEntityEvent->getknockBack(): Vector3`
- Changed `EntityDamageByEntityEvent->setKnockBack(float $knockBack): void` to `EntityDamageByEntityEvent->setKnockBack(Vector3 $knockBack): void`
- Changed `Living->knockBack(float $x, float $z, Vector3 $force = new Vector3(0.4, 0.4, 0.4), ?float $verticalLimit = 0.4)` to `Living->knockBack(float $x, float $z, Vector3 $force = new Vector3(0.4, 0.4, 0.4), ?float $verticalLimit = 0.4)`
### Behavioural changes

None
## Backwards compatibility

Before:
```php
$event->setKnockBack($event->getKnockBack() * 2);
//...
$player->knockBack($deltaX, $deltaZ, 10);
```
After:
```php
$event->setKnockBack($event->getKnockBack()->multiply(2));
//...
$player->knockBack($deltaX, $deltaZ, new Vector3(10, 10, 10));
```
## Follow-up

Perhaps add a new static method to `Vector3` like `Vector3::fromScalar(0.4)` which returns a `Vector3` with X, Y and Z as 0.4 to prevent duplication.
## Tests

```php
<?php

declare(strict_types=1);

namespace JavierLeon9966\KnockBackTest;

use pocketmine\event\entity\EntityDamageByEntityEvent;
use pocketmine\math\Vector3;

		$this->getServer()->getPluginManager()->registerEvent(EntityDamageByEntityEvent::class, static function(EntityDamageByEntityEvent $event): void{
			$event->setKnockBack(new Vector3(0.2, 0.7, 0.2));
		}, EventPriority::HIGHEST, $this);
	}
}
```

https://github.com/pmmp/PocketMine-MP/assets/58715544/b5be2445-5187-4034-9274-b88dbba9433c

### DIFF
--- a/src/entity/Living.php
+++ b/src/entity/Living.php
@@ -529,7 +529,8 @@ abstract class Living extends Entity{
 			//TODO: knockback should not just apply for entity damage sources
 			//this doesn't matter for TNT right now because the PrimedTNT entity is considered the source, not the block.
 			$base = $source->getKnockBack();
-			$source->setKnockBack($base - min($base, $base * $this->getHighestArmorEnchantmentLevel(VanillaEnchantments::BLAST_PROTECTION()) * 0.15));
+			$force = $base - min($base, $base * $this->getHighestArmorEnchantmentLevel(VanillaEnchantments::BLAST_PROTECTION()) * 0.15);
+			$source->setKnockBack(new Vector3($force, $force, $force));
 		}
 
 		parent::attack($source);
@@ -544,14 +545,14 @@ abstract class Living extends Entity{
 			$e = $source->getChild();
 			if($e !== null){
 				$motion = $e->getMotion();
-				$this->knockBack($motion->x, $motion->z, $source->getKnockBack());
+				$this->knockBack($motion->x, $motion->z, $source->getKnockBack(), $source->getVerticalKnockBackLimit());
 			}
 		}elseif($source instanceof EntityDamageByEntityEvent){
 			$e = $source->getDamager();
 			if($e !== null){
 				$deltaX = $this->location->x - $e->location->x;
 				$deltaZ = $this->location->z - $e->location->z;
-				$this->knockBack($deltaX, $deltaZ, $source->getKnockBack());
+				$this->knockBack($deltaX, $deltaZ, $source->getKnockBack(), $source->getVerticalKnockBackLimit());
 			}
 		}
 
@@ -565,7 +566,7 @@ abstract class Living extends Entity{
 		$this->broadcastAnimation(new HurtAnimation($this));
 	}
 
-	public function knockBack(float $x, float $z, float $force = 0.4, ?float $verticalLimit = 0.4) : void{
+	public function knockBack(float $x, float $z, Vector3 $force = new Vector3(0.4, 0.4, 0.4), ?float $verticalLimit = 0.4) : void{
 		$f = sqrt($x * $x + $z * $z);
 		if($f <= 0){
 			return;
@@ -576,11 +577,11 @@ abstract class Living extends Entity{
 			$motionX = $this->motion->x / 2;
 			$motionY = $this->motion->y / 2;
 			$motionZ = $this->motion->z / 2;
-			$motionX += $x * $f * $force;
-			$motionY += $force;
-			$motionZ += $z * $f * $force;
+			$motionX += $x * $f * $force->x;
+			$motionY += $force->y;
+			$motionZ += $z * $f * $force->z;
 
-			$verticalLimit ??= $force;
+			$verticalLimit ??= $force->y;
 			if($motionY > $verticalLimit){
 				$motionY = $verticalLimit;
 			}
--- a/src/entity/Living.php
+++ b/src/entity/Living.php
@@ -546,14 +546,14 @@ abstract class Living extends Entity{
 			$e = $source->getChild();
 			if($e !== null){
 				$motion = $e->getMotion();
-				$this->knockBack($motion->x, $motion->z, $source->getKnockBack(), $source->getVerticalKnockBackLimit());
+				$this->knockBack($motion->x, $motion->z, $source->getKnockBack());
 			}
 		}elseif($source instanceof EntityDamageByEntityEvent){
 			$e = $source->getDamager();
 			if($e !== null){
 				$deltaX = $this->location->x - $e->location->x;
 				$deltaZ = $this->location->z - $e->location->z;
-				$this->knockBack($deltaX, $deltaZ, $source->getKnockBack(), $source->getVerticalKnockBackLimit());
+				$this->knockBack($deltaX, $deltaZ, $source->getKnockBack());
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


src/entity/Living.php
```diff
@@ -531,7 +531,9 @@ public function attack(EntityDamageEvent $source) : void{
 			//TODO: knockback should not just apply for entity damage sources
 			//this doesn't matter for TNT right now because the PrimedTNT entity is considered the source, not the block.
 			$base = $source->getKnockBack();
-			$source->setKnockBack($base - min($base, $base * $this->getHighestArmorEnchantmentLevel(VanillaEnchantments::BLAST_PROTECTION()) * 0.15));
+			$knockBackResistance = $this->getHighestArmorEnchantmentLevel(VanillaEnchantments::BLAST_PROTECTION()) * 0.15;
+			$calculateNewValue = fn(float $baseValue) => $baseValue - min($baseValue, $baseValue * $knockBackResistance);
+			$source->setKnockBack(new Vector3($calculateNewValue($base->x), $calculateNewValue($base->y), $calculateNewValue($base->z)));
```
dktapps, line 536: This code seems unnecessarily confusing

src/entity/Living.php
```diff
@@ -567,7 +569,7 @@ protected function doHitAnimation() : void{
 		$this->broadcastAnimation(new HurtAnimation($this));
 	}
 
-	public function knockBack(float $x, float $z, float $force = 0.4, ?float $verticalLimit = 0.4) : void{
+	public function knockBack(float $x, float $z, Vector3 $force = new Vector3(0.4, 0.4, 0.4), ?float $verticalLimit = 0.4) : void{
```
dktapps, line 572: Is this valid code?

dktapps (commented):
I'm not sure this change makes sense. I don't really see the need for a BC break here.

I don't understand the purpose of making knockback a Vector3. The direction of knockback can already be altered by the X/Z parameters.

dktapps:
I've pushed a change to enable setting vertical knockback limit for 5.4. That doesn't require a BC break, so it shouldn't be mixed with this other stuff.

src/entity/Living.php
JavierLeon9966: Yes, since PHP 8.1

JavierLeon9966:
> I'm not sure this change makes sense. I don't really see the need for a BC break here.
> 
> I don't understand the purpose of making knockback a Vector3. The direction of knockback can already be altered by the X/Z parameters.

Is to let users make a weak horizonal force and a high vertical force like I showed in the video. Perhaps there's some use case to make the X and Z forces different, otherwise I can just make 2 forces for XZ and Y.

JavierLeon9966:
> I've pushed a change to enable setting vertical knockback limit for 5.4. That doesn't require a BC break, so it shouldn't be mixed with this other stuff.

Alright, I'll revert the vertical knock back limit change.

src/entity/Living.php
JavierLeon9966: What can make it less confusing?

dktapps:
> Perhaps there's some use case to make the X and Z forces different

No. Changing X and Z independently just changes the direction of the knockback, which can be achieved using the existing x/z parameters + the force (which is a magnitude).

Different force for Y doesn't make much sense to me either, unless there is a situation in which you can be knocked upwards and not sideways.

JavierLeon9966:
> Different force for Y doesn't make much sense to me either, unless there is a situation in which you can be knocked upwards and not sideways.

Iron golems and hoglins/zoglins have more vertical force than horizontal ones, entities that are in the same space as you just knock you vertically, and explosions such as end crystals can have more vertical force.

dktapps:
Explosions aren't affected by knockback resistance. Hoglin knockback uses specialized logic that probably can't be reproduced using this function with or without these adaptations.

JavierLeon9966:
The only reason to make the Y force different these days is for PvP servers making their custom knock back. These changes will be needed if custom players aren't a thing anymore. If this change is too specific for the core then feel free to close it.

dktapps:
> The only reason to make the Y force different these days is for PvP servers making their custom knock back.

I'm not seeing a case where simply changing the force and vertical limit wouldn't be sufficient. Do you have an example?

JavierLeon9966:
> > The only reason to make the Y force different these days is for PvP servers making their custom knock back.
> 
> I'm not seeing a case where simply changing the force and vertical limit wouldn't be sufficient. Do you have an example?

For example the video I sent you see the villager getting knocked vertically more than horizontally. I can't do that by increasing the force and vertical knock back limit because the vertical knock back force can only be less than or equal to the knock back force.

xSuperr:
> For example the video I sent you see the villager getting knocked vertically more than horizontally. I can't do that by increasing the force and vertical knock back limit because the vertical knock back force can only be less than or equal to the knock back force.

However, many pvp servers will just override the knockback function in custom player classes anyways, so I don't feel as if this change is necessary until the removal of custom player classes.

dktapps:
For fully customizing knockback, my current thought process is to introduce an API for a KnockbackHandler interface to remove the urge to override player classes. This would also be much more useful for general entities.



> > > The only reason to make the Y force different these days is for PvP servers making their custom knock back.
> > 
> > 
> > I'm not seeing a case where simply changing the force and vertical limit wouldn't be sufficient. Do you have an example?
> 
> For example the video I sent you see the villager getting knocked vertically more than horizontally. I can't do that by increasing the force and vertical knock back limit because the vertical knock back force can only be less than or equal to the knock back force.

Yes, I saw the video. I asked for an actual use case where this would be needed in a currently operating server. You can achieve weird knockback parameters by using setMotion(), which some parts of the code already do.

JavierLeon9966:
> For fully customizing knockback, my current thought process is to introduce an API for a KnockbackHandler interface to remove the urge to override player classes. This would also be much more useful for general entities.

Should this pull request be changed to add that interface or just make it change the horizontal and vertical forces separately? Now that I see the parameters can change the X and Z directions separately.

dktapps:
A feature like that should be the topic of a separate PR. I'm not yet sure if it even makes sense.

dktapps:
Closing this as I don't think a Vector3 force makes any sense. It would make more sense to add a Y direction component considering we already have X and Z components. (Perhaps the X/Z params should be replaced by a Vector3 instead of force.)

PR #5955, opened 2023-08-02
Update safe spawn API to use new world height bounds

jasonw4331:
## Introduction

`World::getSafeSpawn()` operates with the expectation the y is above 0 for some undocumented reason. For worlds which generate below Y level 0, like some flat worlds, this makes the API ignore any passed parameters. 

### API changes

`World::getSafeSpawn()` should be mentioned in the changelog with support for lower world heights

### Behavioral changes

`World::getSafeSpawn()` now accepts Vector3 objects with the Y param as low as the world minimum

## Backwards compatibility

N/A

## Follow-up

It should be considered whether to test if the max y value should also be checked by this function

## Tests

TODO


src/world/World.php
```diff
@@ -2951,7 +2951,7 @@ function() use ($resolver) : void{
 	 * @throws WorldException if the terrain is not generated
 	 */
 	public function getSafeSpawn(?Vector3 $spawn = null) : Position{
-		if(!($spawn instanceof Vector3) || $spawn->y < 1){
+		if($spawn === null || $spawn->y < $this->minY){
```
JavierLeon9966:
```suggestion
		if($spawn === null || $spawn->y < $this->minY + 1){
```
There must be at least one block where the player can stand otherwise, the player will spawn above the void.

JavierLeon9966 (changes requested):

src/world/World.php
dktapps: <= can be used instead

dktapps: This instanceof change seems unrelated

dktapps (changes requested):

dktapps (approved):

PR #5962, opened 2023-08-03
Age blocks logic moved into dedicated trait

ShockedPlot7560:
## Introduction
Currently, a lot of code is duplicated for block age logic. This PR attempts to correct this problem by centralizing functions that are mostly duplicated in a dedicated trait.

Each block using this feature must define a constant: ``MAX_AGE`` which will be used by the trait. Defining this constant externally wouldn't be that much of a problem, as phpstan will detect if this constant is missing.

### API changes
- Add trait `AgeBlockTrait`


src/block/utils/AgeBlockTrait.php
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * This trait is used for blocks that have an age property.
+ * Need to add to the block the constant MAX_AGE.
+ */
+trait AgeBlockTrait{
+	protected int $age = 0;
+
+	protected function describeBlockOnlyState(RuntimeDataDescriber $w) : void{
+		$w->boundedInt(strlen(decbin(self::MAX_AGE)), 0, self::MAX_AGE, $this->age);
```
dktapps: (int) ceil(log(2, max - min)) should be used here

src/block/utils/AgeBlockTrait.php
IvanCraft623, line 1: Maybe AgeableBlockTrait name would be more descriptive

ShockedPlot7560: This was one of the questions that lingered in the back of my mind. I was also hesitating with this name, which was the first one I used, and then I renamed it

src/block/CaveVines.php
```diff
@@ -38,9 +39,9 @@
 use function mt_rand;
 
 class CaveVines extends Flowable{
+	use AgeBlockTrait;
```
zSALLAZAR:
```suggestion
	use AgeBlockTrait;

```

src/block/ChorusFlower.php
```diff
@@ -42,28 +42,12 @@
 use function mt_rand;
 
 final class ChorusFlower extends Flowable{
+	use AgeBlockTrait;
```
zSALLAZAR:
```suggestion
	use AgeBlockTrait;

```

src/block/Crops.php
```diff
@@ -34,25 +34,9 @@
 use function mt_rand;
 
 abstract class Crops extends Flowable{
+	use AgeBlockTrait;
```
zSALLAZAR:
```suggestion
	use AgeBlockTrait;

```

src/block/Fire.php
```diff
@@ -36,25 +36,9 @@
 use function mt_rand;
 
 class Fire extends BaseFire{
+	use AgeBlockTrait;
```
zSALLAZAR:
```suggestion
	use AgeBlockTrait;

```

src/block/FrostedIce.php
```diff
@@ -23,30 +23,14 @@
 
 namespace pocketmine\block;
 
-use pocketmine\data\runtime\RuntimeDataDescriber;
+use pocketmine\block\utils\AgeBlockTrait;
 use pocketmine\event\block\BlockMeltEvent;
 use function mt_rand;
 
 class FrostedIce extends Ice{
+	use AgeBlockTrait;
```
zSALLAZAR:
```suggestion
	use AgeBlockTrait;

```

src/block/NetherVines.php
```diff
@@ -41,36 +41,17 @@
  * This class is used for Weeping & Twisting vines, because they have same behaviour
  */
 class NetherVines extends Flowable{
+	use AgeBlockTrait;
```
zSALLAZAR:
```suggestion
	use AgeBlockTrait;

```

src/block/NetherWartPlant.php
```diff
@@ -34,25 +34,9 @@
 use function mt_rand;
 
 class NetherWartPlant extends Flowable{
+	use AgeBlockTrait;
```
zSALLAZAR:
```suggestion
	use AgeBlockTrait;

```

src/block/Sugarcane.php
```diff
@@ -34,14 +34,9 @@
 use pocketmine\world\Position;
 
 class Sugarcane extends Flowable{
+	use AgeBlockTrait;
```
zSALLAZAR:
```suggestion
	use AgeBlockTrait;

```

zSALLAZAR (commented):
A blank line after the use-statement would match the code style

S3v3Nice:
What about adding an abstract function `getMaxAge()` instead of implicit constant `MAX_AGE`?

ShockedPlot7560:
> What about adding an abstract function `getMaxAge()` instead of implicit constant `MAX_AGE`?

I don't think it's appropriate to add a function returning a static result in this case.

src/block/utils/AgeBlockTrait.php
ShockedPlot7560: I don't see the point in the sense that if it's to replace the ``strlen`` it won't give a proper result.

ShockedPlot7560:
Now, I think it's necessary to have a healthier architecture than to force a constant. The only way would be to use two new functions: ``getMinAge`` which would default to 0 and ``getMaxAge`` which would need to be filled in.  
The problem is that this means we're adding yet more functions with fixed values, while we're avoiding introducing new ones as much as possible.  

Glad to hear other opinions on the subject

dktapps:
Constants are probably fine. This PR is only aimed at deduplicating code. We're not trying to introduce new functionality here.

src/block/utils/AgeableBlockTrait.php
```diff
@@ -0,0 +1,51 @@
...
+	protected int $age = 0;
+
+	protected function describeBlockOnlyState(RuntimeDataDescriber $w) : void{
+		$w->boundedInt(strlen(decbin(self::MAX_AGE)), 0, self::MAX_AGE, $this->age);
+	}
+
+	public function getAge() : int{ return $this->age; }
+
+	/** @return $this */
+	public function setAge(int $age) : self{
```
dktapps: This could probably have a `@phpstan-param int<0, self::MAX_AGE> $age`

src/block/utils/AgeableBlockTrait.php
```diff
@@ -0,0 +1,51 @@
...
+ * Need to add to the block the constant MAX_AGE.
+ */
+trait AgeableBlockTrait{
+	protected int $age = 0;
+
+	protected function describeBlockOnlyState(RuntimeDataDescriber $w) : void{
+		$w->boundedInt(strlen(decbin(self::MAX_AGE)), 0, self::MAX_AGE, $this->age);
+	}
+
+	public function getAge() : int{ return $this->age; }
```
dktapps: `@phpstan-return int<0, self::MAX_AGE>` would probably work here

src/block/utils/AgeableBlockTrait.php
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * This trait is used for blocks that have an age property.
+ * Need to add to the block the constant MAX_AGE.
+ */
+trait AgeableBlockTrait{
+	protected int $age = 0;
+
+	protected function describeBlockOnlyState(RuntimeDataDescriber $w) : void{
+		$w->boundedInt(strlen(decbin(self::MAX_AGE)), 0, self::MAX_AGE, $this->age);
```
dktapps: This is a really nasty hack. `(int) ceil(log(self::MAX_AGE, 2))` is better for this.

src/block/utils/AgeBlockTrait.php
dktapps: I had the parameters backwards. See my newer comment below.

dktapps:
I managed to fix some of the int range errors, but it looks like overall they are way too inconvenient because of https://github.com/phpstan/phpstan/issues/6810
Until this is resolved it's probably not worth bothering

I knew there had to be some reason I didn't do this in the original code...

PR #5964, opened 2023-08-04
Powdered snow

CoderJoeW:
## Introduction
The goal of the PR is to introduce powder snow

https://github.com/pmmp/PocketMine-MP/projects/14#card-84464945

Adds the following

- New block "Powder Snow"
- New bucket "Powder Snow Bucket"
- Added cauldron integration with powder snow


dktapps (commented):
You need to run `composer update-registry-annotations` to update the magic methods for VanillaBlocks and VanillaItems

src/block/PowderSnow.php
```diff
@@ -0,0 +1,52 @@
...
+
+use pocketmine\item\Item;
+use pocketmine\item\VanillaItems;
+use pocketmine\world\sound\BucketEmptyPowderSnowSound;
+use pocketmine\world\sound\BucketFillPowderSnowSound;
+use pocketmine\world\sound\Sound;
+
+class PowderSnow extends Opaque{
+
+	public function getDropsForCompatibleTool(Item $item) : array{
```
diamond-gold: Powder snow does not drop anything when mined

src/block/PowderSnow.php
```diff
@@ -0,0 +1,52 @@
...
+		return [
+			VanillaItems::SNOWBALL()->setCount(4)
+		];
+	}
+
+	public function isAffectedBySilkTouch() : bool{
+		return true;
```
diamond-gold: Powder snow does not drop anything when mined, even with Silk Touch

src/block/PowderSnowCauldron.php
```diff
@@ -0,0 +1,60 @@
...
+use pocketmine\math\Vector3;
+use pocketmine\player\Player;
+use pocketmine\world\sound\CauldronEmptyPowderSnowSound;
+use pocketmine\world\sound\CauldronFillPowderSnowSound;
+use pocketmine\world\sound\Sound;
+
+final class PowderSnowCauldron extends FillableCauldron{
+
+	public function writeStateToWorld() : void{
+		parent::writeStateToWorld();
```
diamond-gold: Unneeded override

src/block/PowderSnowCauldron.php
```diff
@@ -0,0 +1,60 @@
...
+
+	public function getEmptySound() : Sound{
+		return new CauldronEmptyPowderSnowSound();
+	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		match($item->getTypeId()){
+			ItemTypeIds::BUCKET => $this->removeFillLevels(self::MAX_FILL_LEVEL, $item, VanillaItems::POWDER_SNOW_BUCKET(), $returnedItems),
+			ItemTypeIds::POWDER_SNOW_BUCKET, ItemTypeIds::WATER_BUCKET => $this->mix($item, VanillaItems::BUCKET(), $returnedItems),
+			ItemTypeIds::LINGERING_POTION, ItemTypeIds::POTION, ItemTypeIds::SPLASH_POTION => $this->mix($item, VanillaItems::GLASS_BOTTLE(), $returnedItems),
```
diamond-gold, line 53: Unrelated change

src/world/World.php
```diff
@@ -3003,7 +3003,7 @@ function() use ($resolver) : void{
 	 * @throws WorldException if the terrain is not generated
 	 */
 	public function getSafeSpawn(?Vector3 $spawn = null) : Position{
-		if(!($spawn instanceof Vector3) || $spawn->y < 1){
+		if(!($spawn instanceof Vector3) || $spawn->y <= $this->minY){
```
diamond-gold, line 3006: Unrelated change

src/block/PowderSnow.php
```diff
@@ -0,0 +1,40 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\block;
+
+use pocketmine\world\sound\BucketEmptyPowderSnowSound;
+use pocketmine\world\sound\BucketFillPowderSnowSound;
+use pocketmine\world\sound\Sound;
+
+class PowderSnow extends Opaque{
```
ShockedPlot7560, line 30: Powder snow isn't an opaque bloc

ShockedPlot7560: The block recovery logic via bucket is missing. The fact that the player slowly falls in with an effect is also missing.

src/item/BlockBucket.php
```diff
@@ -0,0 +1,78 @@
...
+
+namespace pocketmine\item;
+
+use pocketmine\block\Block;
+use pocketmine\block\PowderSnow;
+use pocketmine\event\player\PlayerBucketEmptyEvent;
+use pocketmine\math\Vector3;
+use pocketmine\player\Player;
+
+class BlockBucket extends Item{
```
ShockedPlot7560, line 32: By the same logic, lava and water should also be BlockBuckets. I don't think trying to generalize buckets is correct in this PR.

src/item/BlockBucket.php
```diff
@@ -0,0 +1,78 @@
...
+
+	public function onInteractBlock(Player $player, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, array &$returnedItems) : ItemUseResult{
+		if(!$blockReplace->canBeReplaced()){
+			return ItemUseResult::NONE();
+		}
+
+		//TODO: move this to generic placement logic
+		$resultBlock = clone $this->block;
+
+		if(!$resultBlock instanceof PowderSnow){
```
ShockedPlot7560, line 60: Why name the BlockBucket class in this case and only accept snow, I think this class could just be renamed

ShockedPlot7560 (commented):
This PR looks OK, but a lot of logic is missing to say it's implemented.

ShockedPlot7560:
Ok I didn't see that it was an introduction, however I don't think it's the best thing to just put this half-implemented block.

resources/item_from_string_bc_map.json
```diff
@@ -1328,6 +1328,7 @@
     "potato_block": "minecraft:potatoes",
     "potatoes": "minecraft:potatoes",
     "potion": "minecraft:potion",
+    "powder_snow": "minecraft:powder_snow",
```
dktapps, line 1331: This should not be changed - this file is kept only for BC purposes to support `LegacyStringToItemParser`

src/block/PowderSnow.php
```diff
@@ -0,0 +1,40 @@
...
+
+	public function getBucketFillSound() : Sound{
```
dktapps, line 32: Are these functions used by something? I don't remember them being implemented at the base.

ShockedPlot7560: Its only used in `Blockbucket` class

dktapps: Seems a bit redundant to have them here then ...

PR #5986, opened 2023-08-10
Fixed vertical knock back not matching vanilla

JavierLeon9966:
## Introduction

I tested the knockback in a vanilla server using [gophertunnel](https://github.com/Sandertv/gophertunnel) by logging the output of `SetActorMotionPacket` and it turns out the player has a vertical knockback limit of 0.3608. 
### Relevant issues

None
### API changes

None
### Behavioural changes

None
## Backwards compatibility

None
## Follow-up

None
## Tests

```php
<?php

declare(strict_types=1);

namespace JavierLeon9966\KBTest;

use pocketmine\event\player\PlayerItemUseEvent;

		$this->getServer()->getPluginManager()->registerEvent(PlayerItemUseEvent::class, function(PlayerItemUseEvent $event): void{
			$event->getPlayer()->knockBack(lcg_value() * 2 - 1, lcg_value() * 2 - 1);
		}, EventPriority::MONITOR, $this);
	}
}
```
https://github.com/pmmp/PocketMine-MP/assets/58715544/89d9dd34-13d7-4880-bd77-27b3711acc96

### DIFF
--- a/src/entity/Living.php
+++ b/src/entity/Living.php
@@ -89,7 +89,7 @@ abstract class Living extends Entity{
 	 * Limit of an entity's vertical knockback velocity when hit by another entity. Without this limit, the entity
 	 * may be knocked far up into the air with large knockback forces.
 	 */
-	public const DEFAULT_KNOCKBACK_VERTICAL_LIMIT = 0.4;
+	public const DEFAULT_KNOCKBACK_VERTICAL_LIMIT = 0.3608;
 
 	private const TAG_LEGACY_HEALTH = "HealF"; //TAG_Float
 	private const TAG_HEALTH = "Health"; //TAG_Float


IvanCraft623:
Are you sure this isn't a randomized value?

JavierLeon9966:
> Are you sure this isn't a randomized value?

Yes. I tested multiple times and that value was always logged.

ItsZodiaX:
[dragonfly](https://github.com/df-mc/dragonfly/blob/281943e6efc446daa57c07ff34451f9942c0e034/server/player/player.go#L1483) also use this value.

dktapps:
I feel like there's some physics that we're missing here. Java uses 0.4, and Bedrock mimics Java in this regard.

Issue #3766, opened 2020-08-11
[RFC] Removing the Config class


SOF3:
Before anyone does it,
# DO NOT JUST USE THE :+1:/:-1: REACTION.
This issue contains many points, and we don't know which point you are voting on. If you want to express your opinion, no matter you agree or disagree, please post a comment below stating what you think and why you think so (it is fine even if it is duplicate). Put :+1:/:-1: on the comments below, not this issue body.

This is an RFC. See the guidelines for RFCs in [CONTRIBUTING.md](https://github.com/pmmp/PocketMine-MP/blob/stable/CONTRIBUTING.md#rfc-and-voting).
# Content

## Motivation
The Config class is a mix of unrelated methods on multiple serialization formats, having caused API unsoundness (such as `->set($key, true)` on the LIST type) and undocumented behaviour (such as `fixYAMLIndices`) and difficulty in accessing data (see lots of people on forums and discord asking how to push to mapping list property). It is proposed that the actual features of `Config` be split down into actual documented utility functions for higher reusability and maintainability.

### Deprecate the `Config` class
Deprecate and eventually remove `Config` in API 4.0.0. Most usages can be directly replaced by `yaml_parse_file`.

The `getConfig()` method in plugins is very often replaced by custom invocations on `yaml_parse`, and keeping it in the PluginBase class has little to no value. Since default config is not automatically saved by `getConfig()` anyway, `saveResource("config.yml")` followed by `$this->config = yaml_parse_file($this->getDataFolder() . "config.yml");` would be equivalent. (I would furthermore want to remove `saveDefaultConfig()` as an unnecessary specialization, but that is not relevant to this issue).

#### `getNested` and `setNested`
Provide util functions `getNested` and `setNested`, which act on an actual arrays rather than forced on a `Config` object.

Note that `getNested` and `setNested` are mostly useful for dynamic user input. From the developer perspective, it is more sound to use `$array["foo"]["bar"]` than `Utils::getNested($array, "foo.bar")`.

#### `fixYAMLIndices`
This method is very poorly documented. It is possible to keep it as an util function, but given the unreliable nature of regex replacement (doesn't work with `{x: 1, y: 2, y: 3}` inline mappings), it is probably more reliable to get a fix at the ext-yaml level or adoption of another YAML library (or use NEON?).

#### Properties format
[...]

PR #5988, opened 2023-08-11
default-unicode-for-json

nomirva:


nomirva:
I think you need to make the default unicode option for json config. So that there are no problems with non-ascii alphabets.

dktapps:
I don't see how this fixes anything. It changes the output representation, not the parsing.

nomirva:
This is exactly what I propose. I didn't quite understand what you thought. I just want unicode characters not to be escaped. In the current state, you have to override the options for saving the json config yourself. In the same yaml, a similar option is already set.

dktapps:
Considering that we're planning to get rid of Config (see #3766), I don't think it makes sense to change its behaviour this late in its life cycle. You can always add this option yourself if you want to alter the behaviour of your own JSON configs (or just stop using Config altogether).

PR #5989, opened 2023-08-12
Added ability to change max players by a command

JavierLeon9966:
## Introduction

Sometimes we want to limit the number of players that can join our server without needing to restart it and have to make them rejoin it and possibly make unwanted players join in the process. This pull request adds the /setmaxplayers command to change the max player count during runtime.
### Relevant issues

None
### API changes

None
### Behavioural changes

None
## Backwards compatibility

None
## Follow-up

Requires translations:

| Name | Value in eng.ini |
| :--: | :---: |
| `commands.setmaxplayers.success` | `Set max players to {%0}.` |
| `commands.setmaxplayers.success.lowerbound` | `(Bound to current player count)` |
| `commands.setmaxplayers.usage` | `/setmaxplayers <maxPlayers>` |
| `pocketmine.command.setmaxplayers.description` | `Sets the maximum number of players for this server.` |
| `pocketmine.permission.command.setmaxplayers` | `Allows the user to change the max player count` |

## Tests

TODO

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -262,8 +262,6 @@ class Server{
 
 	private WorldManager $worldManager;
 
-	private int $maxPlayers;
-
 	private bool $onlineMode = true;
 
 	private Network $network;
@@ -338,7 +336,7 @@ class Server{
 	}
 
 	public function getMaxPlayers() : int{
-		return $this->maxPlayers;
+		return $this->configGroup->getConfigInt("max-players", self::DEFAULT_MAX_PLAYERS);
 	}
 
 	/**
@@ -931,8 +929,6 @@ class Server{
 			@touch($bannedIpsTxt);
 			$this->banByIP = new BanList($bannedIpsTxt);
 			$this->banByIP->load();
-
-			$this->maxPlayers = $this->configGroup->getConfigInt("max-players", self::DEFAULT_MAX_PLAYERS);
 
 			$this->onlineMode = $this->configGroup->getConfigBool("xbox-auth", true);
 			if($this->onlineMode){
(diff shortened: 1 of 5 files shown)


dktapps:
I feel like this is better suited to a plugin

SOF3:
First, the changes to Server.php must be merged to make this possible through a plugin.

Second, this is a BC break. There might be plugins assuming an `new SplFixedArray($server->getMaxPlayers())` and end up overflowing.

dktapps:
> First, the changes to Server.php must be merged to make this possible through a plugin.

I hadn't actually noticed this wasn't currently possible. In that case, we should probably add some actual APIs rather than using configs for this.

I still don't think a command for this is appropriate for the core though.

JavierLeon9966:
> I hadn't actually noticed this wasn't currently possible. In that case, we should probably add some actual APIs rather than using configs for this.
> 
> I still don't think a command for this is appropriate for the core though.

At first I considered making `Server->setMaxPlayers()` with validation checks, but then I realized `Server` had no `set` methods and it used the config instead, so I decided to opt for that for consistency.


> Second, this is a BC break. There might be plugins assuming an `new SplFixedArray($server->getMaxPlayers())` and end up overflowing.

I also considered limiting it to `Limits::INT32_MAX` because Minecraft uses that for the max player count and anything above that goes into the negatives, but I chose `PHP_INT_MAX` to be consistent with the config which has no bounds checks.

dktapps:
Closing this as stale. With the proper API (as in #6261) this can be implemented by a plugin.

PR #5991, opened 2023-08-14
Implement pressure plate activation logic and events

dktapps:
## Introduction
This makes it easier for plugins to implement redstone and other custom functionality without needing to overwrite blocks.

### Relevant issues
Closes #5936

### API changes
- Added `PressurePlateUpdateEvent`
  - This event is called on pressure plate activation, signal strength change (for weighted pressure plates) and deactivation.
  - If an entity stands on the pressure plate without moving, the event will be fired once every few ticks, depending on the plate's deactivation delay (20 ticks for simple pressure plates, 10 for weighted).
  - The event extends `BaseBlockChangeEvent`, presenting `getBlock()` and `getNewState()`.
  - `getActivatingEntities()` is also provided, giving the list of entities which are causing the pressure plate to have its new state.
  - While active, the event will be fired even if the pressure plate's state has not changed.
- Deprecated `WeightedPressurePlateLight` and `WeightedPressurePlateHeavy`. These classes have no obvious purpose and will be removed in PM6.
- `WeightedPressurePlate` is no longer abstract.
- `WeightedPressurePlate->__construct()` accepts 2 new parameters: `int $deactivationDelayTicks` and `float $signalStrengthFactor`. Currently optional, these will become mandatory in PM6.
- `PressurePlate->__construct()` accepts 1 new parameter: `int $deactivationDelayTicks`. Currently optional, will become mandatory in PM6.
- Added the following methods to `PressurePlate`:
  - `protected function getActivationBox() : AxisAlignedBB`
  - `protected function hasOutputSignal() : bool` - to become abstract in PM6
  - `protected function calculatePlateState(list<Entity> $entities) : list<Entity>` - to become abstract in PM6
  - `protected function filterIrrelevantEntities(list<Entity> $entities) : list<Entity>`

### Behavioural changes
Pressure plates now behave as expected.

## Backwards compatibility
None, although some BC breaks should be made in PM6 where existing code has been changed.

## Tests
This has been playtested extensively.

Events have not yet been tested as of writing this.

### DIFF
--- a/src/block/PressurePlate.php
+++ b/src/block/PressurePlate.php
@@ -33,8 +33,8 @@ use pocketmine\math\Facing;
 use pocketmine\math\Vector3;
 use pocketmine\player\Player;
 use pocketmine\world\BlockTransaction;
-use pocketmine\world\sound\RedstonePowerOffSound;
-use pocketmine\world\sound\RedstonePowerOnSound;
+use pocketmine\world\sound\PressurePlateActivateSound;
+use pocketmine\world\sound\PressurePlateDeactivateSound;
 use function count;
 
 abstract class PressurePlate extends Transparent{
@@ -155,8 +155,8 @@ abstract class PressurePlate extends Transparent{
 				$world->setBlock($this->position, $newState);
 				if($pressedChange !== null){
 					$world->addSound($this->position, $pressedChange ?
-						new RedstonePowerOnSound() :
-						new RedstonePowerOffSound()
+						new PressurePlateActivateSound($this) :
+						new PressurePlateDeactivateSound($this)
 					);
 				}
 			}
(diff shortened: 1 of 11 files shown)


dktapps:
The event can now be tested using the following code or something like it:
```php
	public function onPressurePlateUpdate(PressurePlateUpdateEvent $event) : void{
		$newState = $event->getNewState();
		if($newState instanceof SimplePressurePlate){
			$this->getLogger()->info("Pressure plate is " . ($newState->isPressed() ? "pressed" : "not pressed"));
		}elseif($newState instanceof WeightedPressurePlate){
			$this->getLogger()->info("Pressure plate has signal strength " . $newState->getOutputSignalStrength());
		}
		$this->getLogger()->info("Activating entities: " . count($event->getActivatingEntities()));
	}
```

This has been playtested and some bugs fixed involving item entities on stone pressure plates.

dktapps:
Correct sounds still need to be added

PR #5992, opened 2023-08-15
Simplified operation

JavierLeon9966:


src/block/WeightedPressurePlate.php
```diff
@@ -55,12 +55,7 @@ protected function calculatePlateState(array $entities) : array{
 		}
 		return [
 			(clone $this)->setOutputSignalStrength($newSignalStrength),
-			//I dare anyone to simplify this without breaking it
-			match(true){
-				$this->signalStrength === 0 && $newSignalStrength !== 0 => true,
-				$this->signalStrength !== 0 && $newSignalStrength === 0 => false,
-				default => null,
-			}
+			($value = $this->signalStrength === 0) === ($newSignalStrength !== 0) ? $value : null
```
SOF3, line 58: Are you sure it is really a good idea to make this even more complex?

JavierLeon9966: No, I'm not sure. But I may come up with an alternative.

dktapps:
ChatGPT suggested something similar to me, and I chose not to use it due to taking way too much brainpower to understand.

While this takes less lines of code, it is not simpler, it is more difficult to understand.

PR #6000, opened 2023-08-17
Implement proper pre-installed sign editing

alvin0319:
## Introduction
This PR implements proper pre-installed sign editing.

Tested the following cases (which vanilla behaves):
* Interacting with dyes/(glow) ink sak
   * With same color/glowing state: Player can open sign editor.
   * Different color/glowing state: Sign color/glowing state will be changed.
* Interacting without any items on hand
   * Open sign editor.

### Relevant issues
N/A

### API changes

N/A

### Behavioural changes

* SignChangeEvent is now also fired when a player edited a pre-instlaled sign.

## Backwards compatibility

N/A

## Follow-up
N/A

### DIFF
--- a/src/block/BaseSign.php
+++ b/src/block/BaseSign.php
@@ -152,7 +152,7 @@ abstract class BaseSign extends Transparent{
 			return false;
 		}
 		$dyeColor = $this->getColorFromItem($item);
-		if($this->canOpenSign($item)){
+		if($this->canOpenSignEditor($item)){
 			$player->openSignEditor($this->position);
 			return true;
 		}
@@ -247,7 +247,7 @@ abstract class BaseSign extends Transparent{
 	/**
 	 * Returns whether player can open sign editor
 	 */
-	private function canOpenSign(Item $item) : bool{
+	private function canOpenSignEditor(Item $item) : bool{
 		$currentColor = $this->text->getBaseColor();
 		$newColor = $this->getColorFromItem($item);
 		if($newColor === null){
--- a/src/block/BaseSign.php
+++ b/src/block/BaseSign.php
@@ -151,12 +151,11 @@ abstract class BaseSign extends Transparent{
 		if($player === null){
 			return false;
 		}
-		$dyeColor = $item instanceof Dye ? $item->getColor() : match($item->getTypeId()){
-			ItemTypeIds::BONE_MEAL => DyeColor::WHITE(),
-			ItemTypeIds::LAPIS_LAZULI => DyeColor::BLUE(),
-			ItemTypeIds::COCOA_BEANS => DyeColor::BROWN(),
-			default => null
-		};
+		$dyeColor = $this->getColorFromItem($item);
+		if($this->canOpenSign($item)){
+			$player->openSignEditor($this->position);
+			return true;
+		}
 		if($dyeColor !== null){
 			$color = $dyeColor->equals(DyeColor::BLACK()) ? new Color(0, 0, 0) : $dyeColor->getRgbValue();
 			if($color->toARGB() === $this->text->getBaseColor()->toARGB()){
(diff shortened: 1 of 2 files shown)


alvin0319:
CI failure seems unrelated to this PR.

src/block/BaseSign.php
```diff
@@ -235,4 +234,35 @@ public function updateText(Player $author, SignText $text) : bool{
...
+	}
+
+	/**
+	 * Returns whether player can open sign editor
+	 */
+	private function canOpenSignEditor(Item $item) : bool{
+		$currentColor = $this->text->getBaseColor();
+		$newColor = $this->getColorFromItem($item);
+		if($newColor === null){
+			if($item->getTypeId() === ItemTypeIds::GLOW_INK_SAC || $item->getTypeId() === ItemTypeIds::INK_SAC){
```
dktapps: This condition seems redundant

src/block/BaseSign.php
```diff
@@ -235,4 +234,35 @@ public function updateText(Player $author, SignText $text) : bool{
...
+				}
+				if(!$wasGlowing && $item->getTypeId() !== ItemTypeIds::INK_SAC){
+					return false;
+				}
+			}
+			return true;
+		}
+		$color = $newColor->equals(DyeColor::BLACK()) ? new Color(0, 0, 0) : $newColor->getRgbValue();
+		return $currentColor->equals($color);
+	}
```
dktapps, line 265: This function looks like it's repeating a lot of logic. Seems to me like we should just be able to check if the previous actions did anything in `onInteract` (or simply put `openSignEditor` at the end instead of the `false` return) instead of having all this extra code.

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -756,7 +756,9 @@ public function handleBlockActorData(BlockActorDataPacket $packet) : bool{
 			}
 
 			try{
-				$text = SignText::fromBlob($textBlobTag->getValue());
+				$baseColor = $block->getText()->getBaseColor();
+				$glowing = $block->getText()->isGlowing();
+				$text = SignText::fromBlob($textBlobTag->getValue(), $baseColor, $glowing);
```
dktapps, line 761: This allows the player to change the sign properties without the requisite items.

alvin0319: I don't understand how this code allows players to change sign properties without items, the code only looks up for the sign which is actually placed on the world, not depending on what the client send.

dktapps: Then why change this at all?

alvin0319: The previous code doesn't consider the sign color and glowing state, resulting in the removal of color and glowing states when the player edited the sign.

dktapps: My bad, I misread the code.

The new variables for the formatting should be moved out of the try-catch, as they shouldn't be throwing any exceptions.

src/block/BaseSign.php
alvin0319: We can't put `openSignEditor` at the end of `onInteract()` because of the following client behavior:
* Interacting with dyes/(glow) ink sak
   * With same color/glowing state: Player can open sign editor.
   * Different color/glowing state: Sign color/glowing state will be changed.
* Interacting without any items on hand
   * Open sign editor.
  
Opening sign editor should happen before changing sign.

dktapps: I don't see how this prevents it from being placed at the end.

dktapps:
Superseded by d44e0e87d0106ff003de830a1296974cbd13e8ee

PR #6013, opened 2023-08-23
Hanging signs

dktapps:
## Introduction
Adds hanging signs

Hanging signs have 3 variants:
- 16-way rotation attached at a single point on the ceiling
- 4-way rotation attached at two points on the ceiling
- 4-way rotation attached to a solid signpost attached to a wall

This PR is not complete, but I needed to clear out my workspace for a beta release.

### API changes
TODO: fill this section once the PR is done

## Tests
Not complete yet


dktapps:
Not planning to complete this

PR #6014, opened 2023-08-23
Automatically fallback to PSR-0 when class list is unchanged

jasonw4331:
## Introduction

Pharynx is the primary supporting tool for plugins being compiled with external libraries which need shaded. The tool itself is not designed only for pm plugin use though and so outputs PSR-0 plugins without changing the plugin manifest to comply with the changes.

This PR checks whether the known class list is updated after PSR-4 tries to autoload the plugin. If the list is unchanged then PSR-0 is automatically attempted.

### Relevant issues

* https://github.com/SOF3/pharynx/issues/5

### Behavioral changes

`ThreadSafeClassLoader::addPath()` now falls back on PSR-0 when PSR-4 loading returns no results.

## Follow-up

TODO

## Tests

TODO

### DIFF
--- a/changelogs/4.23.md
+++ b/changelogs/4.23.md
@@ -60,3 +60,9 @@ Released 9th August 2023.
 - Fixed `PluginBase->saveResource()` leaking file resources when the data file already exists in the plugin's data folder. This bug existed since 2014 and was only discovered recently.
 - Fixed coral blocks becoming dead after calling `getDropsForCompatibleTool()` on them.
 - Fixed `BlockDeathEvent->getOldState()` returning a block which is already dead.
+
+# 4.23.6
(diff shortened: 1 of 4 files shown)


dktapps:
This seems like a problem for pharynx to fix.

I don't really see how your change makes any difference, anyway. This code doesn't check whether any classes exist yet. This would only change behaviour if `addPath()` is called twice for the same namespace, which we don't do.

SOF3:
I think this is neither the problem of PocketMine nor pharynx, but just of the user adding an unreasonable configuration to their plugin. This is just like cross-compiles a binary on WSL to .exe and complain that WSL isn't smart enough to automatically run .exe binaries on the native windows subsystem.

PR #6015, opened 2023-08-23
Implement waiting login sequence

ShockedPlot7560:
## Introduction
This PR introduces a new system for deferred the player creation, allowing plugins to add tasks to be done at that time. This ensures that plugins load player data correctly before the player is created and the login sequence launched.

Plugins can also cancel the player's connection by rejecting their promise.

### API changes
- Add `Promise::all()`

### Behavioural changes
- Player creation sequence is launched only when all promises added by plugins are resolved.

## Tests

```php
	}

	public function onPlayerCreated(PlayerPreCreationEvent $event) : void {
		$resolver = new PromiseResolver();
		$this->getLogger()->info("Player " . $event->getNetworkSession()->getDisplayName() . " pre creation");
		$event->addPromise($resolver->getPromise());
		$this->getScheduler()->scheduleDelayedTask(new ClosureTask(function() use ($event, $resolver) {
			$this->getLogger()->info("Player " . $event->getNetworkSession()->getDisplayName() . " pre creation (delayed)");
			$resolver->resolve(null); // Connect the player
			// OR
			$resolver->reject(); // Disconnect the player
		}), 5 * 20);
	}
}
```

With promise resolve:
```
[16:21:18.114] [Server thread/DEBUG]: [NetworkSession: ShockedPlot7560] Resource packs sequence completed
[16:21:18.609] [Server thread/INFO]: [TestPlugin] Player ShockedPlot7560 pre creation
[16:21:23.071] [Server thread/INFO]: [TestPlugin] Player ShockedPlot7560 pre creation (delayed)
```
With promise rejection: 
```
[16:46:59.301] [Server thread/INFO]: [TestPlugin] Player ShockedPlot7560 pre creation
[16:47:04.276] [Server thread/INFO]: [TestPlugin] Player ShockedPlot7560 pre creation (delayed)
[16:47:04.276] [Server thread/INFO]: [NetworkSession: ShockedPlot7560] Session fermée : Failed to create player (ID d'erreur: cfb2-7e1a-f645)
[16:47:04.276] [Server thread/ERROR]: [NetworkSession: ShockedPlot7560] Failed to create player
```


src/event/player/PlayerCreatedEvent.php
```diff
@@ -0,0 +1,66 @@
...
+ * - Loading initial player data from external sources or databases before the main
+ * gameplay loop begins.
+ * - Providing a tailored welcome experience, which might involve sending messages,
+ * granting starter items, or awarding initial rewards.
+ *
+ * Note:
+ * After each task is completed, it's essential to call the `done` method of the WaitGroup
+ * to signal its completion and allow the WaitGroup to determine when to proceed.
+ */
+class PlayerCreatedEvent extends PlayerEvent{
```
dktapps: I really don't like this event name. This is way too easy to confuse with `PlayerCreationEvent`.

dktapps:
I'd be happier if there was a way to do this without adding new events. I don't really see an obvious reason why a new event is needed here.

ShockedPlot7560:
> I'd be happier if there was a way to do this without adding new events. I don't really see an obvious reason why a new event is needed here.

This can be done via callbacks, in the same way as inventory listener or craftmanager.  
Is it better? I'm currently hesitating between the event and this solution.

dktapps:
Well, I don't see why e.g. `PlayerLoginEvent` couldn't be used for this. Perhaps even `PlayerCreationEvent` itself.

ShockedPlot7560:
The idea was that this would allow things to be executed before the event login. But the more I think about it, the more I realize that it could be done in the PlayerCreationEvent or during the login event.

ShockedPlot7560:
I moove the logic to `PlayerCreationEvent`. It's not really enviable to place it in the `PlayerLoginEvent`, in the sense that my ambition is that we can specify tasks to be executed **before** the event is launched. These tasks are executed before the login sequence is launched: spawn, log console and change handler for player spawn.  

The event's documentation needs to be reviewed, because it is no longer dedicated to changing the player's class, but also contains access to the `WaitGroup`

src/Server.php
```diff
@@ -577,25 +577,27 @@ public function createPlayer(NetworkSession $session, PlayerInfo $playerInfo, bo
...
-				}
-			);
-		}else{ //returning player with a valid position - safe spawn not required
-			$createPlayer($playerPos);
-		}
+				);
+			}else{ //returning player with a valid position - safe spawn not required
+				$createPlayer($playerPos);
+			}
+		});
```
dktapps: Side note: This code is really becoming callback hell... probably should be refactored at some point

src/utils/WaitGroup.php
```diff
@@ -0,0 +1,130 @@
...
+			$this->execute = $execute;
+		}
+	}
+
+	/**
+	 * Adds $count pending tasks to the wait group.
+	 *
+	 * New tasks cannot be added after `wait()` has been called.
+	 */
+	public function add(int $count = 1) : void{
```
dktapps: I wonder if it makes sense to allow specifying this. Feels like an opening for someone to accidentally pass a random integer and break the whole thing (add/remove underflow).

Perhaps it would make more sense to ditch the parameter. I can't think of any use-cases that would necessitate calling this with anything other than `1` anyway. Such cases can use a `for` loop if that's what they really want to do.

src/utils/WaitGroup.php
```diff
@@ -0,0 +1,130 @@
...
+namespace pocketmine\utils;
+
+use Closure;
+use InvalidArgumentException;
+use RuntimeException;
+
+/**
+ * Represents a group of tasks that can be asynchronously completed.
+ *
+ * Unlike `sync.WaitGroup` in Go, each WaitGroup can only be used once.
```
dktapps: This documentation is a bit confusing.

- Why mention golang wait group at all?
- Why can't we reuse a wait group? Seems like it shouldn't be that difficult to reset the state once the counter reaches `0`.

src/event/player/PlayerCreationEvent.php
```diff
@@ -109,4 +116,20 @@ public function setPlayerClass(string $class) : void{
...
+	 *  - Carrying out preparatory tasks such as configuring default preferences, setting
+	 * player-specific settings, or initializing in-game assets.
+	 *  - Loading initial player data from external sources or databases before the login sequence begins.
+	 *  - Providing a tailored welcome experience, which might involve sending messages,
+	 * granting starter items, or awarding initial rewards.
+	 *
+	 * Note:
+	 *  After each task is completed, it's essential to call the `done` method of the WaitGroup
+	 *  to signal its completion and allow the WaitGroup to determine when to proceed.
+	 */
```
dktapps: A brief summary line at the top would be good, something like:

`Returns a WaitGroup which allows pausing the player creation until a given number of tasks have been completed`

dktapps:
I'm getting the impression that this "wait group" is essentially just a `Promise` combined with a semaphore, except that this has no `reject()` functionality.

The documentation is perhaps too much. People are going to get scared off by seeing that much text.

ShockedPlot7560:
I've completely rewritten the system by simply adding `Promise::all`

From now on, plugins will be able to add promises that must be completed before the player is created.

src/event/player/PlayerCreationEvent.php
```diff
@@ -118,18 +117,24 @@ public function setPlayerClass(string $class) : void{
...
 	 *
-	 * Note:
-	 *  After each task is completed, it's essential to call the `done` method of the WaitGroup
-	 *  to signal its completion and allow the WaitGroup to determine when to proceed.
+	 * @return Promise[]
+	 * @phpstan-return Promise<mixed>[]
 	 */
-	public function getWaitGroup() : WaitGroup{
-		return $this->waitGroup;
+	public function getPromises() : array{
```
dktapps: Doubtful it makes sense to expose this

ShockedPlot7560: Funny you should mention that, I thought the same thing and it's fixed in 64a164d7aa300429b1d6a927626647d291bfc342

tests/phpstan/configs/actual-problems.neon
```diff
@@ -1215,3 +1215,8 @@ parameters:
 			count: 1
 			path: ../../phpunit/scheduler/AsyncPoolTest.php
 
+		-
+			message: "#^Right side of && is always true\\.$#"
+			count: 1
+			path: ../../../src/promise/Promise.php
```
dktapps, line 1211: Seems like we could avoid this error by doing `$toResolve = count($promises)`

src/promise/Promise.php
```diff
@@ -52,4 +52,54 @@ public function onCompletion(\Closure $onSuccess, \Closure $onFailure) : void{
...
+					$values[$key] = $value;
+
+					if(--$toResolve === 0 && $continue){
+						$resolver->resolve($values);
+					}
+				},
+				function() use ($resolver, &$continue) : void{
+					if($continue){
+						$continue = false;
+						$resolver->reject();
```
dktapps, line 89: ~~The other promises should be rejected here too, otherwise they'll be left dangling forever~~ nvm, I can't read

src/event/player/PlayerCreationEvent.php
```diff
@@ -109,4 +117,15 @@ public function setPlayerClass(string $class) : void{
 		Utils::testValidInstance($class, $this->baseClass);
 		$this->playerClass = $class;
 	}
+
+	/**
+	 * Adds a promise to the waiting list for the login sequence.
+	 * Once all the promises that have been added have been completed,
+	 * the player login sequence will be initiated and the player will be created.
+	 *
+	 * @phpstan-param Promise<mixed> $promise
```
dktapps: `Promise<null>` would make more sense, as we won't use any values given by plugins anyway.

(Really it ought to be `Promise<void>`, but I think that would make the signature invalid.)

src/promise/Promise.php
```diff
@@ -52,4 +52,54 @@ public function onCompletion(\Closure $onSuccess, \Closure $onFailure) : void{
...
+
+	/**
+	 * Returns a promise that will resolve only once all the Promises in
+	 * `$promises` have resolved. The resolution value of the returned promise
+	 * will be an array containing the resolution values of each Promises in
+	 * `$promises` indexed by the respective Promises' array keys.
+	 *
+	 * @phpstan-param Promise<mixed>[] $promises
+	 *
+	 * @phpstan-return Promise<array<int, mixed>>
```
dktapps: Probably makes more sense to use a `@phpstan-template TValue` for this

src/event/player/PlayerCreationEvent.php
ShockedPlot7560: I don't think `void` works properly

tests/phpstan/configs/actual-problems.neon
ShockedPlot7560: It's the continue var that poses a problem

dktapps: My bad, you are right.

src/promise/Promise.php
```diff
@@ -59,19 +59,19 @@ public function isResolved() : bool{
 	 * will be an array containing the resolution values of each Promises in
 	 * `$promises` indexed by the respective Promises' array keys.
 	 *
-	 * @phpstan-param Promise<mixed>[] $promises
+	 * @phpstan-template TTValue
```
dktapps: `TTValue`?

ShockedPlot7560: Phpstan noted that TValue conflicted with the Promise template. Could this be an isolated problem with my local configuration, idk

dktapps: No, it's correct. Seems like this could be a problem though, as there's no way for it to infer template types and it will likely complain.

src/promise/Promise.php
```diff
@@ -52,4 +53,54 @@ public function onCompletion(\Closure $onSuccess, \Closure $onFailure) : void{
...
 		return $this->shared->resolved;
 	}
+
+	/**
+	 * Returns a promise that will resolve only once all the Promises in
+	 * `$promises` have resolved. The resolution value of the returned promise
+	 * will be an array containing the resolution values of each Promises in
+	 * `$promises` indexed by the respective Promises' array keys.
+	 *
+	 * @phpstan-template TValue
```
dktapps: It's probably fine to get rid of this. As it's a static method it doesn't really matter anyway.

dktapps:
I meant to get rid of the @phpstan-template tag on Promise::all(), not the generics altogether.

ShockedPlot7560:
I don't think I've understood everything. The all type can't use the Promise template, it has no link?

BrandPVP:
If you load data in a callback and store it in memory, will the `PlayerQuitEvent` trigger to free the memory when the player leaves? Could there be a memory leak? Maybe we should have this in `PlayerLoginEvent` too?

dktapps:
> I don't think I've understood everything. The all type can't use the Promise template, it has no link?

In this case it'll be inferred from the parameters since there's no $this.

dktapps:
> If you load data in a callback and store it in memory, will the `PlayerQuitEvent` trigger to free the memory when the player leaves? Could there be a memory leak? Maybe we should have this in `PlayerLoginEvent` too?

This is outside the scope of the PR

ShockedPlot7560:
I don't know where the PHPstan error came from... would anyone know?

src/promise/Promise.php
```diff
@@ -23,10 +23,11 @@
 
 namespace pocketmine\promise;
 
+use function count;
 use function spl_object_id;
 
 /**
- * @phpstan-template TValue
+ * @phpstan-template-covariant TValue
```
dktapps: Why is this changed?

ShockedPlot7560: Without, PHPstan return: 
```
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   src/Server.php                                                                                                                                                                          
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  :595   Parameter #1 $promises of static method pocketmine\promise\Promise<mixed>::all() expects array<pocketmine\promise\Promise<mixed>>, array<int, pocketmine\promise\Promise<null>> given.  
         💡 Template type TValue on class pocketmine\promise\Promise is not covariant. Learn more: https://phpstan.org/blog/whats-up-with-template-covariant                                     
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 ```

dktapps: I see. I suppose it makes sense in this case considering that `Promise` is a container class.

ShockedPlot7560: Is this resolved so ? Or we remove template covariance from promise (strange for me)

dktapps: I personally don't love template-covariant as I find it very confusing. But in this case I think it's probably OK.

dktapps:
My main sticking point for this PR now boils down to the use of `PlayerCreationEvent`. Considering there's been discussions to scrap it, I'm not sure if repurposing it like this is a good idea.

Is there something to gain by doing this as opposed to using e.g. `PlayerLoginEvent`?

ShockedPlot7560:
We will loose monitor priority with that.

ShockedPlot7560:
Will become the same problem as https://github.com/pmmp/PocketMine-MP/pull/5056

dktapps:
Ah, because PlayerLoginEvent is mutable ... yes, you are correct, I'd forgotten how easily async events can become nightmares

ShockedPlot7560:
I notice that PlayerCreationEvent is not immutable too.
Do we need to create a new event for that? Point of discussion, I think, because pointing to next-major just to ensure that the event will be immutable doesn't seem like a good idea.

dktapps:
Yeah, the same thought crossed my mind as well. We can't guarantee that PlayerCreationEvent will be immutable in PM6 anyway, as there are several unresolved obstacles to removing custom player classes.

ShockedPlot7560:
Wall, the only way to do this, so, is to create a new `WantedPlayerCreationEvent` to define the promises and, once all the promises have been resolved, launch `PlayerCreationEvent`, etc...

ShockedPlot7560:
Well, after some testing, it's more interesting to move this functionality before launching PlayerPreLogin. I was able to carry out player disconnection tests at the time of `PlayerCreation`, `PlayerPreLogin` and `PlayerLogin`, and only the prelogin outputted the error message 100% correctly. In the other cases, a "disconnected from server" appears.  

It wouldn't be a good idea to do it at login time either, as this would mean processing the LoginPacket when the player isn't necessarily authorized to connect (in the case of a plugin checking a player's ban in the database, for example).

ShockedPlot7560:
Now requires a new translation: the message that will be displayed when the player disconnects because the promise has been rejected.

dktapps:
> Well, after some testing, it's more interesting to move this functionality before launching PlayerPreLogin. I was able to carry out player disconnection tests at the time of `PlayerCreation`, `PlayerPreLogin` and `PlayerLogin`, and only the prelogin outputted the error message 100% correctly. In the other cases, a "disconnected from server" appears.
> 
> It wouldn't be a good idea to do it at login time either, as this would mean processing the LoginPacket when the player isn't necessarily authorized to connect (in the case of a plugin checking a player's ban in the database, for example).

Whether or not the client functions correctly shouldn't be relevant. Also, doing this before PlayerPreLogin could be problematic for loading playerdata from remote databases, which is the primary use case for a feature like this.

ShockedPlot7560:
Would you suggest putting it between the prelogin and the login?

dktapps:
I'm undecided, but I don't think prelogin is the right call.

dries-c:
Ideally this should be ran after PlayerLoginEvent as to not waste resources on gathering data for players that will never connect or to not overload the database whenever people are spamming fake accounts.

ShockedPlot7560:
I would also agree

dktapps:
I'm leaning towards doing it directly after PlayerCreationEvent, to permit the player to be properly constructed with the appropriate data. It would be nice if Player::__construct() could receive all of its data and then be able to proceed directly to the spawn sequence IMO.

ShockedPlot7560:
> I'm leaning towards doing it directly after PlayerCreationEvent, to permit the player to be properly constructed with the appropriate data. It would be nice if Player::__construct() could receive all of its data and then be able to proceed directly to the spawn sequence IMO.

I don't think that passing values to the constructor is relevant for this PR, but I wouldn't agree with that too much.
We'd like to do away with custom players and allow plugins to inject values directly into the player. Perhaps this is another vision you had in mind?

dktapps:
I'm thinking of stuff like the `?CompoundTag` containing the player's data. But yes, you're probably right, since custom data would be used to configure plugin sessions, rather than the player itself.

ShockedPlot7560:
I think it would be better to rename the event into `PlayerPreCreationEvent` event because it will be called just before player creation (accorded by @dktapps' comment) and it will be called just after the `PlayerCreationEvent` (destined to disappear). 

Afterwards, I also agree with @dries-c's opinion in the sense that applying it after login would avoid doing things on a potentially invalid player, a bot. Potentially to be done in another PR to allow some flexibility to plugins and thus have the choice? There may be cases where I see the need for these two events.

dktapps:
> because it will be called just before player creation (accorded by @dktapps' comment)

This was a suggestion, not a request. Your point about removing custom player classes changed my mind.

> Afterwards, I also agree with @dries-c's opinion in the sense that applying it after login would avoid doing things on a potentially invalid player, a bot. Potentially to be done in another PR to allow some flexibility to plugins and thus have the choice? There may be cases where I see the need for these two events.

I don't see a need for two events. Invalid players won't get any further than PlayerPreLoginEvent anyway.

ShockedPlot7560:
> This was a suggestion, not a request. Your point about removing custom player classes changed my mind.  

I know, but I thought the comment was pertinent and would perhaps allow more flexibility in the future to have this event fired before the player is built and not after.

ShockedPlot7560:
We now use the `PlayerPreCreationEvent`, which is launched when the player is created. Once the promises have been completed, the `PlayerCreationEvent` is launched.

ShockedPlot7560:
What about merging this ? Is there need some modification or is it clear for the moment ?

ShockedPlot7560:
I think we need a general way of making events asynchronous before we change the logic of these events. Closing

PR #6022, opened 2023-08-25
Corrects errors in y-coordinate values caused in PlayerAuthInputPacket calculations

Nerahikada:
## Introduction
PlayerAuthInputPacket coordinate calculations were rounded and then subtracted, resulting in errors at certain coordinates.
Therefore, subtracting and then rounding the values corrects this problem.

## Tests
Move to the coordinates where the error occurs (e.g. y=7, 8, 31, 32, etc.) and output the coordinates with `var_dump()`.

### DIFF
--- a/src/network/mcpe/handler/InGamePacketHandler.php
+++ b/src/network/mcpe/handler/InGamePacketHandler.php
@@ -201,7 +201,7 @@ class InGamePacketHandler extends PacketHandler{
 		}
 
 		$hasMoved = $this->lastPlayerAuthInputPosition === null || !$this->lastPlayerAuthInputPosition->equals($rawPos);
-		$newPos = $rawPos->round(4)->subtract(0, 1.62, 0);
+		$newPos = $rawPos->subtract(0, 1.62, 0)->round(4);
 
 		if($this->forceMoveSync && $hasMoved){
 			$curPos = $this->player->getLocation();


dktapps:
I don't really understand why this fixes the problem, nor why this is a problem to begin with. There are precision errors simply because of the protocol's use of float32, while PM core uses float64.

We already compensate for floating point accuracy errors with epsilons anyway, so I don't really see what difference this makes.

Nerahikada:
For example, if the player is at y=7, the coordinates would be:
```
object(pocketmine\math\Vector3)#1495417 (3) {
  ["x"]=>
  float(253.5306)
  ["y"]=>
  float(6.999999999999999)
  ["z"]=>
  float(259.3722)
}
```
Then, when a function like `World::setBlock()` is executed, the block is placed at y=6 instead of the expected coordinates (y=7) because the coordinate values are floored.

dktapps:
One specific case where this causes desired behaviour doesn't mean it will work for all cases. I don't see a strong reason **why** this change would improve accuracy in any meaningful way. I want to be able to understand this before merging such a change.

Nerahikada:
Conversely, when is the error generated by the calculation necessary?
This is also to correct rounding errors caused by subtraction, not to improve accuracy.

Nerahikada:
This is a rounding error problem due to subtraction and can be easily reproduced: `var_dump(8.62 - 1.62);`
And the output will be: `float(6.999999999999999)`

dktapps:
I see. Then the same result could also be achieved by adding a second round() without removing the one before it. Though I see that's probably going to cost more accuracy than what this PR is doing.

Nerahikada:
> Though I see that's probably going to cost more accuracy than what this PR is doing.

What does that mean?
Google Translate and DeepL give opposite results and I'm confused 😕

dktapps:
It means you're fine, don't worry about a thing :)

PR #6027, opened 2023-08-29
Facing::UP now required to extinguish cake candle

### LINKED ISSUE #5983: Cake with candle should only be able to be extinguished by clicking the top face
### Issue description
Candles on cakes can currently be extinguished by clicking any face with an empty hand. This is problematic because it differs from the client's expectation, causing rollbacks.

This is a headache to fix due to requiring a rework of CandleTrait.

1. Click side of candle cake with empty hand
2. Notice glitchy block updates

### OS and versions

* PocketMine-MP: 5.4.2
* PHP: 8.2.8

Rorkh:
## Introduction
Fixed wrong cake with candle behavior that cake candle can be extinguished by clicking on faces other than top face.

I'm not good with codebase but I think code changes that I did isn't bad. Although issue author said that this issue is problematic to fix and I didn't found and problems fixing what can indicate that I'm very bad with codebase and dumb, I consider myself not dumb and I claim I fixed the problem in the best way possible. I did the facing check in CakeWithCandle.php because it's only cake relevant problem and this check shouldn't be done on any other candle extinguish because candles on blocks that isn't a cake can be extenguished from any face by clicking on candle.

### Relevant issues
* Fixes #5983

### Behavioural changes
Behavior of cake candle was changed

## Tests
1. Place cake
2. Place candle on cake
3. Click on face other than top face
4. Confirm cake candle can't be extinguished that way
5. Click on top face
6. Cake candle extinguished!

I swear I did this tests in-game and they are real.

### DIFF
--- a/src/block/CakeWithCandle.php
+++ b/src/block/CakeWithCandle.php
@@ -52,7 +52,7 @@ class CakeWithCandle extends BaseCake{
 	}
 
 	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
-		if($this->onInteractCandle($item, $face, $clickVector, $player, $returnedItems)){
+		if($face === Facing::UP && $this->onInteractCandle($item, $face, $clickVector, $player, $returnedItems)){
 			return true;
 		}
 


ShockedPlot7560:
I think you can light the candle by clicking anywhere on the cake, but you can only extinguish it by touching the candle (top side). Right?

Rorkh:
now I understand it's all. truly a pain.

PR #6030, opened 2023-09-03
Add generic types for TaskHandler

ShockedPlot7560:
## Introduction
Currently, if we store an handler returned by the scheduler and wish to retrieve the Task it stores, it is reduced to an Task. This PR brings a generic handler type to improve static analysis at plugin level and adds consistency to the code.

### Relevant issues
n/a

### API changes
- Implement template ``TTask`` for `TaskHandler`

## Backwards compatibility
At the static analysis level, plugins may have more errors if generic handler types are not defined.

## Follow-up
See if the problem with the `TaskScheduler`'s `isReady` function is a PHPstan problem: https://phpstan.org/r/a0269a3f-2a24-4b62-bcd3-2727e69b862b


src/scheduler/TaskScheduler.php
```diff
@@ -33,12 +33,12 @@
...
 	private bool $enabled = true;
 
-	/** @phpstan-var ReversePriorityQueue<int, TaskHandler> */
+	/** @phpstan-var ReversePriorityQueue<int, TaskHandler<Task>> */
 	protected ReversePriorityQueue $queue;
 
 	/**
-	 * @var ObjectSet|TaskHandler[]
-	 * @phpstan-var ObjectSet<TaskHandler>
+	 * @var ObjectSet|TaskHandler<Task>[]
```
SOF3: should this be added to `@var`?

ShockedPlot7560: Why shouldn't it?

SOF3: I thought generics should only go to `@phpstan-var`

dktapps: Yeah, @var is meant to be PSR-5 compliant so that tools can understand it
Generics should be confined to @phpstan-var

dktapps:
To be honest, I don't actually see what the gain of this is. There's no obvious reason to get the task instance from the handler.

src/scheduler/Task.php
```diff
@@ -26,8 +26,12 @@
 use pocketmine\utils\Utils;
 
 abstract class Task{
+	/** @phpstan-var TaskHandler<self>|null  */
 	private ?TaskHandler $taskHandler = null;
 
+	/**
+	 * @phpstan-return TaskHandler<self>|null
```
dktapps: This would have to be `static`, not `self`, and I don't think that actually works

ShockedPlot7560:
> To be honest, I don't actually see what the gain of this is. There's no obvious reason to get the task instance from the handler.

Internally, the gain are not necessarily significant. However, for plugins wishing to analyze their code with a static analysis tool or an IDE, this part of the code can be better typed.
Currently, if someone stores a TaskHandler and later calls task retrieval, the task type is lost. This means that if we want to call up a function specific to our task, the function is unknown, problematic.

dktapps:
If the type is important, you can retain the `Task` directly, and then use `Task->getHandler()` to fetch the handler, which has no type variation. So I don't really see what the point of this is, and I doubt that it's actually possible to properly typehint this, as it appears to require generics using `static`.

ShockedPlot7560:
This typehint is correct and its only purpose is to improve documentation. Even if it's possible to do otherwise, I think it might be worth doing anyway.

dktapps:
All things considered, it doesn't even really make much sense to have two different types for this. They are coupled anyway.

The only real reason to interact with the TaskHandler from the plugin side is to cancel the task. This can be better done using CancelTaskException anyway, or perhaps we could add an API method that allows a task to flag itself as cancelled, without having to expose the handler.

ShockedPlot7560:
I agree that the API would require changes, however, why should we have to wait for an API change before we can have correct typing and functional autocompletion/static analysis?

dktapps:
As I said previously I don't really see a point to this change anyway. If you want to access the Task object you can just retain it in a variable before calling scheduleTask() or whatever. I'm also not sure it's actually even possible to do this fully anyway due to the requirement for `TaskHandler<static>`.

dktapps:
> See if the problem with the TaskScheduler's isReady function is a PHPstan problem: https://phpstan.org/r/a0269a3f-2a24-4b62-bcd3-2727e69b862b

It's actually a consequence of the shitty design of SplPriorityQueue (mode of the queue affects the returned values). The error you see was actually ignored in a baseline already, but surfaced again here because of the changed types.

src/scheduler/Task.php
dktapps: OK, turns out it actually does work in this instance https://phpstan.org/r/0595cfe5-5ccd-4a13-90da-3b9db2a0e7cb

`static` in generics is just very flaky so I had my doubts

dktapps (approved):

PR #6038, opened 2023-09-06
Add a bit of customization to ConsoleSender!

VennDev:
## Introduction
- Add the ConsoleOutputEvent event to test results such as: results from commands returned through Player, Console, ...

### API changes
- Add `ConsoleOutputEvent.php` in `src\event\server`
- Update/Change code in `pocketmine\console\ConsoleCommandSender.php` at line 70


dktapps:
What's the use-case?

src/event/server/ConsoleOutputEvent.php
```diff
@@ -0,0 +1,48 @@
...
+
+namespace pocketmine\event\server;
+
+use pocketmine\event\Cancellable;
+use pocketmine\event\CancellableTrait;
+
+/**
+ * Called when the server outputs a message to the console.
+ */
+class ConsoleOutputEvent extends ServerEvent implements Cancellable{
```
dktapps: The event is marked as cancellable, yet you don't check the cancellation state, so cancelling the event will do nothing.

src/event/server/ConsoleOutputEvent.php
```diff
@@ -0,0 +1,48 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\event\server;
+
+use pocketmine\event\Cancellable;
+use pocketmine\event\CancellableTrait;
+
+/**
+ * Called when the server outputs a message to the console.
```
dktapps: This is inaccurate. The event is only fired when command output is sent to a ConsoleCommandSender.

VennDev:
For an example I need to use it to check that the command results from the Console 'dispatchCommand(new ConsoleCommandSender'

VennDev:
Example code: 
```php
        public function onCommand(CommandEvent $event): void
	{
		$command = $event->getCommand();
		$key = explode(" ", $command)[0];
		if (!$event->getSender() instanceof Player) return;
		$this->cmdHandler[$key] = $command;
	}

	public function onConsoleOutput(ConsoleOutputEvent $event): void
	{
		$messages = $event->getMessages();
		if (isset($this->cmdHandler['abc'])) {
			var_dump($messages); // Handle something here!
		}
	}
```

dktapps:
> For an example I need to use it to check that the command results from the Console 'dispatchCommand(new ConsoleCommandSender'

You should make a custom CommandSender for that.

ShockedPlot7560:
This break BC and i don't see the point of that changes.

VennDev:
> 

what BC ?

dktapps:
@VennDev I don't think you understood what I was telling you to do. I'm saying implement `CommandSender` on a custom class _in your plugin_ to capture the messages. No change to the PM core is required to do that.

This is what, for example, the RconServer plugin does: https://github.com/pmmp/RconServer/blob/4adfd71651db1e06a1914725891f2c858431f196/src/RconCommandSender.php
(though this example is a bit overcomplicated)

This PR is now full of unrelated changes, and breaks BC for any plugin that already implements `CommandSender`. Since there appears to be nothing to do here, I'm going to close this.

PR #6039, opened 2023-09-06
RuntimeDataDescriber: Support dynamically describing arbitrary enums

### LINKED ISSUE #5877: Unable to describe new enums using RuntimeDataDescriber
## Description
Please add a way to accept string/enum block states in RuntimeDataDescriber, or add all known states to RuntimeDataDescriber

## Justification
String block states could not be loaded or saved
Possibly auto generate block state enums & RuntimeDataDescriber methods from BedrockData?

Affected blocks: 
- Big dripleaf
- Pointed dripstone
- Seagrass
- Structure block
- Structure void
- Sniffer egg
- Turtle egg

Blocks states:
- BlockStateStringValues::BIG_DRIPLEAF_TILT_*
- BlockStateStringValues::CRACKED_STATE_*
- BlockStateStringValues::DRIPSTONE_THICKNESS_*
- BlockStateStringValues::SEA_GRASS_TYPE_*
- BlockStateStringValues::STRUCTURE_BLOCK_TYPE_*
- BlockStateStringValues::STRUCTURE_VOID_TYPE_*
- BlockStateStringValues::TURTLE_EGG_COUNT_*

As of 1.20.10, PM 5.3
The observer block is also affected
Block states: BlockStateStringValues::MC_FACING_DIRECTION_*

## Alternative methods
Hack!

https://github.com/diamond-gold/DummyItemsBlocks/blob/main/src/diamondgold/DummyItemsBlocks/block/hack/HackStringProperty.php

https://github.com/diamond-gold/DummyItemsBlocks/blob/main/src/diamondgold/DummyItemsBlocks/block/PointedDripstone.php

dktapps:
Previously, we were using codegen to support describing a fixed set of enums.

Instead, we implement an enum() function, allowing any native PHP enum to be described.
All enums used in runtime data have been migrated to native PHP 8.1 enums in minor-next to facilitate this.

This implementation is:
- faster
- requires way less code
- does not require a build step
- way more flexible

This fixes #5877, increasing the range of stuff that plugins are now able to do.

### Relevant issues
Fixes #5877

### API changes
- Added `RuntimeDataDescriber->enum()`. `RuntimeDataDescriber` is considered a sealed interface, so this doesn't constitute a BC break.
- Deprecated `RuntimeEnumDescriber`. This can't be directly removed due to BC breaks (plugins might use the functions it provides).

### Behavioural changes
The server behaves as before (although block data serialization will be significantly faster when enums are involved).

## Backwards compatibility
No known BC issues.

## Follow-up
Shim methods were left for the old codegen'd methods to the new `enum()` method. The trait and interface for these functions `LegacyRuntimeEnumDescriberTrait` and `RuntimeEnumDescriber` should be removed in PM6.

## Tests
This passes a full PHPUnit test run, demonstrating that the result of the changes is the same as before (as verified by `block_factory_consistency_check.json`).


dktapps:
cc @diamond-gold

dktapps:
An alternative approach that doesn't require `RuntimeDataEnum` would look like this:

```php
$w->enum($this->dripleafState, DripleafState::getAll());
```
This would be a bit slower (due to `array_map()` in `getAll()`), but more flexible (it allows using PM core enums that don't implement `RuntimeDataEnum`).

dktapps:
Better approach seems to be to just use native enums, so we can take advantage of `UnitEnum` interface.

Issue #5973, opened 2023-08-07
Make Player->getFirstPlayed() and Player->getLastPlayed() return DateTimeImmutable


dktapps:
## Justification
The current format (UNIX time in milliseconds) is opaque and undocumented. `DateTimeImmutable` is thoroughly documented and has an API that's relatively easy to understand.

PR #6042, opened 2023-09-08
Make Player->getFirstPlayed() and Player->getLastPlayed() return DateTimeImmutable

zSALLAZAR:
### Relevant issues
* #5973 

### API changes
* `public function getFirstPlayed() : ?int` now returns `?DateTimeImmutable`
* `public function getLastPlayed() : ?int` now returns `?DateTimeImmutable`

### Behavioural changes
None

## Backwards compatibility
BC break

## Follow-up
None

## Tests
```php
		$this->getServer()->getPluginManager()->registerEvent(PlayerJoinEvent::class, function(PlayerJoinEvent $event): void{

			$player->sendMessage("FirstPlayed: " . ($player->getFirstPlayed()?->format("Y-m-d H:i:s") ?? "Null"));
			$player->sendMessage("LastPlayed: " . ($player->getLastPlayed()?->format("Y-m-d H:i:s") ?? "Null"));
			$player->sendMessage("hasPlayedBefore: " . ($player->hasPlayedBefore() ? "Yes" : "No"));
	}
}
```


src/player/OfflinePlayer.php
```diff
@@ -36,12 +37,12 @@ public function getName() : string{
 		return $this->name;
 	}
 
-	public function getFirstPlayed() : ?int{
-		return ($this->namedtag !== null && ($firstPlayedTag = $this->namedtag->getTag(Player::TAG_FIRST_PLAYED)) instanceof LongTag) ? $firstPlayedTag->getValue() : null;
+	public function getFirstPlayed() : ?DateTimeImmutable{
+		return ($this->namedtag !== null && ($firstPlayedTag = $this->namedtag->getTag(Player::TAG_FIRST_PLAYED)) instanceof LongTag) ? (new DateTimeImmutable())->setTimestamp($firstPlayedTag->getValue()) : null;
```
dktapps: This is incorrect. The time is represented as UNIX time in milliseconds.

dktapps (commented):
The PR is generally incorrectly using the NBT data, which is UNIX epoch milliseconds, not seconds.

src/player/Player.php
```diff
@@ -428,19 +428,19 @@ public function getUniqueId() : UuidInterface{
...
 	 * TODO: not sure this should be nullable
 	 */
-	public function getLastPlayed() : ?int{
+	public function getLastPlayed() : ?DateTimeImmutable{
 		return $this->lastPlayed;
 	}
 
 	public function hasPlayedBefore() : bool{
-		return $this->lastPlayed - $this->firstPlayed > 1; // microtime(true) - microtime(true) may have less than one millisecond difference
+		return $this->firstPlayed->diff($this->lastPlayed)->format('Uu') > 1;
```
dktapps: Shouldn't this be `U.u`?

Also, type juggling in comparisons is discouraged

src/player/OfflinePlayer.php
zSALLAZAR: This PR should now use the NBT data correctly

src/player/Player.php
zSALLAZAR: Using `U.u` would return a float, but that was wrong anyway

src/player/Player.php
```diff
@@ -368,8 +369,13 @@ function() : void{
 			}
 		));
 
-		$this->firstPlayed = (new DateTimeImmutable())->setTimestamp($nbt->getLong(self::TAG_FIRST_PLAYED, $now = (new DateTimeImmutable())->getTimestamp()));
-		$this->lastPlayed = (new DateTimeImmutable())->setTimestamp($nbt->getLong(self::TAG_LAST_PLAYED, $now));
+		$createDateTimeImmutable = static function(string $tag) use ($nbt) : DateTimeImmutable{
+			$value = $nbt->getLong($tag, (int) (new DateTimeImmutable())->format('Uv'));
+
+			return (new DateTimeImmutable('@' . substr_replace((string) $value, '.', 10, 0)));
```
dktapps: I don't understand what this is doing. Can you explain?

zSALLAZAR: This divides the stored milliseconds by 1000. This must be done because DateTimeImmutable only accepts seconds with microseconds and not with milliseconds. And yes, there is an easier way, I already updated the code: https://github.com/pmmp/PocketMine-MP/pull/6042/commits/3819a076407aa47019d5b1256199e3c03d3e6499

src/player/Player.php
```diff
@@ -368,8 +368,11 @@ function() : void{
 			}
 		));
 
-		$this->firstPlayed = $nbt->getLong(self::TAG_FIRST_PLAYED, $now = (int) (microtime(true) * 1000));
-		$this->lastPlayed = $nbt->getLong(self::TAG_LAST_PLAYED, $now);
+		$createDateTimeImmutable = static function(string $tag) use ($nbt) : DateTimeImmutable{
+			return new DateTimeImmutable('@' . $nbt->getLong($tag, (int) (new DateTimeImmutable())->format('Uv')) / 1000);
```
dktapps: This weird default seems unnecessary. The `microtime` version was more readable.

dktapps:
Looking better now.

src/player/Player.php
```diff
@@ -2328,8 +2331,8 @@ public function getSaveData() : CompoundTag{
 		}
 
 		$nbt->setInt(self::TAG_GAME_MODE, GameModeIdMap::getInstance()->toId($this->gamemode));
-		$nbt->setLong(self::TAG_FIRST_PLAYED, $this->firstPlayed);
-		$nbt->setLong(self::TAG_LAST_PLAYED, (int) floor(microtime(true) * 1000));
+		$nbt->setLong(self::TAG_FIRST_PLAYED, (int) $this->firstPlayed->format('Uv'));
+		$nbt->setLong(self::TAG_LAST_PLAYED, (int) (new DateTimeImmutable())->format('Uv'));
```
zSALLAZAR: Should `microtime` be used here too?

src/player/Player.php
zSALLAZAR: Changed.

src/player/Player.php
dktapps: There's nothing gained by changing it to DateTime IMO.

zSALLAZAR: Changed.

ShockedPlot7560 (dismissed):

src/player/Player.php
```diff
@@ -428,19 +432,19 @@ public function getUniqueId() : UuidInterface{
...
 	 * TODO: not sure this should be nullable
 	 */
-	public function getLastPlayed() : ?int{
+	public function getLastPlayed() : ?DateTimeImmutable{
 		return $this->lastPlayed;
 	}
 
 	public function hasPlayedBefore() : bool{
-		return $this->lastPlayed - $this->firstPlayed > 1; // microtime(true) - microtime(true) may have less than one millisecond difference
+		return ((int) $this->firstPlayed->diff($this->lastPlayed)->format('%f')) > 1;
```
dktapps: This behaviour differs from the original. The previous code checked if the difference was at least 1 second.
This is now checking if the difference is at least 1 microsecond, which is much more prone to failure.

src/player/Player.php
```diff
@@ -368,8 +369,11 @@ function() : void{
 			}
 		));
 
-		$this->firstPlayed = $nbt->getLong(self::TAG_FIRST_PLAYED, $now = (int) (microtime(true) * 1000));
-		$this->lastPlayed = $nbt->getLong(self::TAG_LAST_PLAYED, $now);
+		$createDateTimeImmutable = static function(string $tag) use ($nbt) : DateTimeImmutable{
+			return new DateTimeImmutable('@' . $nbt->getLong($tag, (int) (microtime(true) * 1000)) / 1000);
```
dktapps: This is now creating two different timestamps. Previously, `$now` was reused to ensure that the two timestamps would be identical for new players.

dktapps (approved):

PR #6044, opened 2023-09-08
Static support trait

dktapps:
## Introduction
Many blocks require unconditional support from one of their neighbouring blocks.
The logic for this has been repeated and reinvented in many places, often with inconsistent results and various bugs.

This PR introduces `FixedSupportTrait`, which provides the placement precondition checks and the self-destruction behaviour when the block's support is removed.

### API changes
- Added `FixedSupportTrait`
- Deprecated protected `SweetBerryBush->canBeSupportedBy()` (this should really have been private)

### Behavioural changes
Everything should work as it did before.

## Backwards compatibility
No BC issues known.

## Follow-up
Introduce a `DynamicSupportTrait` for blocks whose support requirements depends on a state property.

## Tests
This has been briefly playtested, but further testing is needed to ensure that no regressions have been introduced.

### DIFF
--- a/src/block/Bamboo.php
+++ b/src/block/Bamboo.php
@@ -23,7 +23,7 @@ declare(strict_types=1);
 
 namespace pocketmine\block;
 
-use pocketmine\block\utils\FixedSupportTrait;
+use pocketmine\block\utils\StaticSupportTrait;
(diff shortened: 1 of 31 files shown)


dktapps:
This has now been fully playtested and bugs fixed.

PR #6045, opened 2023-09-08
Add Structure Block and Structure Void

Bqleine:
## Introduction
This pull request is looking to implement the structure void, and possibly the structure block as it is closely related.

I personally don't have the knowledge required to implement the full structure block, but I would like to get it working at least as an inanimate place-able block.

### Relevant issues
*  #5423: A previous attempt at implementing structure void
* https://github.com/pmmp/PocketMine-MP/projects/14#card-49457870
* https://github.com/pmmp/PocketMine-MP/projects/14#card-84464854

## Changes
This is a non-disruptive addition of blocks.

## Tasks
- [x] Register structure void
  - [x] Define block states (defaults as void)
  - [x] Implement serialization
- [x] Register structure block
  - [x] Implement interactions with the block
  - [x] Define block states (defaults as save mode)
  - [x] Implement serialization
  - [x] Create tile

Left for a future PR:
  - [ ] Add events related to structure block modification?
  - [ ] Create permission for using structure block?
  - [ ] Plugin API
  - [ ] saving/loading of structures to disk (probably better left for another PR for full structure API)
  - [ ] Link block and tile type
  - [ ] Unit tests

## Bugs
- [ ] Falling blocks landing inside a structure void should replace it instead of dropping
- [ ] The `air` variant of structure voids gets placed as a `void` variant in vanilla
- [ ] Sneaking and right clicking to create a structure block on another while flying will create a ghost block instead

## Notes
~I have used arbitrary values in `BlockTypeIds.php`, as there is a gap there I am not sure if it should have used a different number inside the gap.~
~I have used blast resistance values as indicated in the minecraft wiki, the block is indestructible, does it really matter ?~
~Right now, both blocks appear as update blocks to the client. I am not sure why, is the block state mandatory for them to appear correctly ?~

### Structure Void
- `/give <player> structure_void`
- place it
- save the world
### Structure Block
plugin.yml
```yaml
---
name: StructureBlockTests
version: 0.0.1
main: Baleine\StructureBlockTests\Main
api: 5.5.1
src-namespace-prefix: Baleine\StructureBlockTests
commands:
  test:
    usage: "/test"
    description: "Test"
    permission: command.test
permissions:
  command.test:
    default: true
...
```
Main.php
```php
<?php

declare(strict_types=1);

namespace Baleine\StructureBlockTests;

use pocketmine\block\StructureBlock;
use pocketmine\block\tile\StructureBlock as TileStructureBlock;
use pocketmine\block\utils\StructureAnimationMode;
use pocketmine\block\utils\StructureAxes;
use pocketmine\block\utils\StructureBlockType;
use pocketmine\block\utils\StructureRotation;
use pocketmine\math\Vector3;
use pocketmine\player\Player;

	/**
	 * @param string[] $args
	 */
	public function onCommand(CommandSender $sender, Command $command, string $label, array $args) : bool{
		if ($sender instanceof Player){
			switch ($command->getName()){
				case "test":
					var_dump($sender->getPosition());
					$world = $sender->getPosition()->getWorld();
					$position = $sender->getPosition()->add(0, -1, 0);
					$block = $world->getBlock($position);
					$tile = $world->getTile($position);
					if (!($block instanceof StructureBlock)) return false;
					if (!($tile instanceof TileStructureBlock)) return false;
					$tile->setType(StructureBlockType::LOAD);
					$tile->setShowBoundingBox(true);
					$tile->setStructureSize(new Vector3(16, 255, 16));
					$tile->setStructureOffset(new Vector3(0, -$position->getY(), 0));
					$tile->setAnimationMode(StructureAnimationMode::BY_BLOCK);
					$tile->setAnimationSeconds(5);
					$tile->setIgnoreEntities(true);
					$tile->setIntegrity(45.8);
					$tile->setIntegritySeed(76876867);
					$tile->setMirroredAxes(new StructureAxes(false, true));
					$tile->setRemoveBlocks(false);
					$tile->setRotation(new StructureRotation(180));
					$tile->setStructureName("Test123");
					$block->setType(StructureBlockType::LOAD);
					$world->setBlock($position, $block);
					break;
			}
		}
		return true;
	}
}
```

Place a structure block and stand on top of it, then execute the /test command and check that the parameters have been updated.
~Problem: if $block->setType is not called the block will not visibly update (it calls setBlock). I am not sure how to go about this.~
Plugins should call setBlock themselves after changing block properties.


dktapps:
> I have used arbitrary values in BlockTypeIds.php, as there is a gap there I am not sure if it should have used a different number inside the gap.

The values can be basically anything. The usual way is to put them at the bottom and increase the FIRST_UNUSED_TYPE_ID constant (or whatever it's called, I forget) to accommodate. However, putting them in a gap will also work.

> Right now, both blocks appear as update blocks to the client. I am not sure why, is the block state mandatory for them to appear correctly ?

You haven't registered them in `BlockStateToObjectDeserializer` and `BlockObjectToStateSerializer`.

Bqleine:
After having implemented block states and serialization/deserialization. I'm getting this error when using the block:
```
[Server thread/CRITICAL]: pocketmine\utils\AssumptionFailedError: "Unmapped blockstate returned by blockstate serializer: TAG_Compound={ "name" => TAG_String="minecraft:structure_void" "version" => TAG_Int=18090528 "states" => TAG_Compound={ "structure_void_type" => TAG_Int=0 } "PMMPDataVersion" => TAG_Long=1 }" (EXCEPTION) in "pmsrc/src/network/mcpe/convert/ItemTranslator" at line 80
--- Stack trace ---
  #0 pmsrc/src/network/mcpe/convert/ItemTranslator(57): pocketmine\network\mcpe\convert\ItemTranslator->toNetworkId(object pocketmine\item\ItemBlock#148201)
  #1 pmsrc/src/network/mcpe/convert/TypeConverter(207): pocketmine\network\mcpe\convert\ItemTranslator->toNetworkIdQuiet(object pocketmine\item\ItemBlock#148201)
  #2 pmsrc/src/network/mcpe/InventoryManager(540): pocketmine\network\mcpe\convert\TypeConverter->coreItemStackToNet(object pocketmine\item\ItemBlock#148201)
  #3 pmsrc/src/network/mcpe/InventoryManager(561): pocketmine\network\mcpe\InventoryManager->syncContents(object pocketmine\inventory\PlayerInventory#160238)
  #4 pmsrc/src/network/mcpe/handler/PreSpawnPacketHandler(141): pocketmine\network\mcpe\InventoryManager->syncAll()
  #5 pmsrc/src/network/mcpe/NetworkSession(326): pocketmine\network\mcpe\handler\PreSpawnPacketHandler->setUp()
  #6 pmsrc/src/network/mcpe/NetworkSession(820): pocketmine\network\mcpe\NetworkSession->setHandler(object pocketmine\network\mcpe\handler\PreSpawnPacketHandler#160390)
  #7 pmsrc/src/network/mcpe/NetworkSession(277): pocketmine\network\mcpe\NetworkSession->beginSpawnSequence()
  #8 pmsrc/src/promise/Promise(45): pocketmine\network\mcpe\NetworkSession->onPlayerCreated(object pocketmine\player\Player#167479)
  #9 pmsrc/src/network/mcpe/NetworkSession(234): pocketmine\promise\Promise->onCompletion(object Closure#167480, object Closure#167559)
  #10 pmsrc/src/network/mcpe/NetworkSession(815): pocketmine\network\mcpe\NetworkSession->createPlayer()
  #11 pmsrc/src/network/mcpe/handler/ResourcePacksPacketHandler(142): pocketmine\network\mcpe\NetworkSession->pocketmine\network\mcpe\{closure}()
  #12 pmsrc/vendor/pocketmine/bedrock-protocol/src/ResourcePackClientResponsePacket(61): pocketmine\network\mcpe\handler\ResourcePacksPacketHandler->handleResourcePackClientResponse(object pocketmine\network\mcpe\protocol\ResourcePackClientResponsePacket#167671)
  #13 pmsrc/src/network/mcpe/NetworkSession(445): pocketmine\network\mcpe\protocol\ResourcePackClientResponsePacket->handle(object pocketmine\network\mcpe\handler\ResourcePacksPacketHandler#167540)
  #14 pmsrc/src/network/mcpe/NetworkSession(383): pocketmine\network\mcpe\NetworkSession->handleDataPacket(object pocketmine\network\mcpe\protocol\ResourcePackClientResponsePacket#167671, string[4] ....)
  #15 pmsrc/src/network/mcpe/raklib/RakLibInterface(219): pocketmine\network\mcpe\NetworkSession->handleEncoded(string[7]
c.`a`..)
  #16 pmsrc/vendor/pocketmine/raklib-ipc/src/RakLibToUserThreadMessageReceiver(40): pocketmine\network\mcpe\raklib\RakLibInterface->onPacketReceive(int 0, string[16] ....et...\^+`.[.)
  #17 pmsrc/src/network/mcpe/raklib/RakLibInterface(111): raklib\server\ipc\RakLibToUserThreadMessageReceiver->handle(object pocketmine\network\mcpe\raklib\RakLibInterface#168038)
  #18 pmsrc/vendor/pocketmine/snooze/src/SleeperHandler(120): pocketmine\network\mcpe\raklib\RakLibInterface->pocketmine\network\mcpe\raklib\{closure}()
  #19 pmsrc/src/TimeTrackingSleeperHandler(58): pocketmine\snooze\SleeperHandler->processNotifications()
  #20 pmsrc/vendor/pocketmine/snooze/src/SleeperHandler(79): pocketmine\TimeTrackingSleeperHandler->processNotifications()
  #21 pmsrc/src/Server(1681): pocketmine\snooze\SleeperHandler->sleepUntil(float 1694260926.2298)
  #22 pmsrc/src/Server(1064): pocketmine\Server->tickProcessor()
  #23 pmsrc/src/PocketMine(334): pocketmine\Server->__construct(object pocketmine\thread\ThreadSafeClassLoader#3, object pocketmine\utils\MainLogger#6, string[30] /home/bloup/Pocketmine-Source/, string[38] /home/bloup/Pocketmine-Source/plugins/)
  #24 pmsrc/src/PocketMine(357): pocketmine\server()
--- End of exception information ---
```
It seems these mappings are taken directly from bedrock data though. Have I used an incorrect constant ?

Bqleine:
I have found the issue, I was using 0 and 1 for values instead of constant strings from `BlockStateStringValues`.

src/block/utils/StructureBlockType.php
```diff
@@ -23,37 +23,11 @@
...
-			self::EXPORT => 5,
-		};
-	}
+enum StructureBlockType : int{
+	case DATA = 0;
+	case SAVE = 1;
+	case LOAD = 2;
+	case CORNER = 3;
+	case INVALID = 4;
+	case EXPORT = 5;
```
dktapps: The Bedrock values should not be exposed to the API.

Bqleine: I've replaced it with normal enums with internal methods, what do you think ?

Bqleine:
I think this is ready for an initial review. I've added a plugin to test in the pr description :)

src/data/bedrock/block/convert/BlockStateReader.php
```diff
@@ -349,6 +351,28 @@ public function readWallConnectionType(string $name) : ?WallConnectionType{
 		};
 	}
 
+	/** @throws BlockStateDeserializeException */
+	public function readStructureVoidType() : ?StructureVoidType{
```
zSALLAZAR:
```suggestion
	public function readStructureVoidType() : StructureVoidType{
```

src/data/bedrock/block/convert/BlockStateReader.php
```diff
@@ -349,6 +351,28 @@ public function readWallConnectionType(string $name) : ?WallConnectionType{
...
+		return match($type = $this->readString(BlockStateNames::STRUCTURE_VOID_TYPE)){
+			StringValues::STRUCTURE_VOID_TYPE_VOID => StructureVoidType::VOID,
+			StringValues::STRUCTURE_VOID_TYPE_AIR => StructureVoidType::AIR,
+			default => throw $this->badValueException(BlockStateNames::STRUCTURE_VOID_TYPE, $type),
+		};
+	}
+
+	/** @throws BlockStateDeserializeException */
+	public function readStructureBlockType() : ?StructureBlockType{
```
zSALLAZAR:
```suggestion
	public function readStructureBlockType() : StructureBlockType{
```

zSALLAZAR (changes requested):
These functions should not be nullable

src/data/bedrock/block/convert/BlockStateReader.php
Bqleine: Does this mean line 342 is also incorrect ?
 ```php
	/** @throws BlockStateDeserializeException */
	public function readWallConnectionType(string $name) : ?WallConnectionType{
		return match($type = $this->readString($name)){
			//TODO: this looks a bit confusing due to use of EAST, but the values are the same for all connections
			//we need to find a better way to auto-generate the constant names when they are reused
			//for now, using these constants is better than nothing since it still gives static analysability
			StringValues::WALL_CONNECTION_TYPE_EAST_NONE => null,
			StringValues::WALL_CONNECTION_TYPE_EAST_SHORT => WallConnectionType::SHORT,
			StringValues::WALL_CONNECTION_TYPE_EAST_TALL => WallConnectionType::TALL,
			default => throw $this->badValueException($name, $type),
		};
	}
```

Comment (zSALLAZAR) [reply src/data/bedrock/block/convert/BlockStateReader.php]: No, because as you can see in line 347, this function would return null if `$type` matched `StringValues::WALL_CONNECTION_TYPE_EAST_NONE`, but nowhere in your functions does it return null, so they should not be nullable

Comment (Bqleine) [reply src/data/bedrock/block/convert/BlockStateReader.php]: Understood :+1:

File: src/data/bedrock/block/convert/BlockStateToObjectDeserializer.php [hunk 4]
```diff
@@ -1438,6 +1438,14 @@ private function registerDeserializers() : void{
 			return Blocks::STONECUTTER()
 				->setFacing($in->readHorizontalFacing());
 		});
+		$this->map(Ids::STRUCTURE_BLOCK, function(Reader $in) : Block{
+			return Blocks::STRUCTURE_BLOCK()
+				->setType($in->readStructureBlockType());
```
dktapps: These helpers seem redundant. A simple `match` would suffice.

src/data/bedrock/block/convert/BlockStateToObjectDeserializer.php
```diff
@@ -1438,6 +1438,14 @@ private function registerDeserializers() : void{
...
+		});
+		$this->map(Ids::STRUCTURE_VOID, function(Reader $in) : Block{
+			return Blocks::STRUCTURE_VOID()
+				->setType($in->readStructureVoidType());
```
dktapps: Same again here.

src/block/tile/StructureBlock.php
```diff
@@ -0,0 +1,376 @@
...
+	public const TAG_ROTATION = "rotation";
+	public const TAG_SEED = "seed";
+	public const TAG_SHOW_BOUNDING_BOX = "showBoundingBox";
+	public const TAG_STRUCTURE_NAME = "structureName";
+	public const TAG_X_STRUCTURE_OFFSET = "xStructureOffset";
+	public const TAG_Y_STRUCTURE_OFFSET = "yStructureOffset";
+	public const TAG_Z_STRUCTURE_OFFSET = "zStructureOffset";
+	public const TAG_X_STRUCTURE_SIZE = "xStructureSize";
+	public const TAG_Y_STRUCTURE_SIZE = "yStructureSize";
+	public const TAG_Z_STRUCTURE_SIZE = "zStructureSize";
```
ShockedPlot7560: To go along with the other tiles, you may need to comment the Tag type (e.g. TAG_Byte, ...)

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -1032,4 +1036,46 @@ public function handleEmote(EmotePacket $packet) : bool{
 		$this->player->emote($packet->getEmoteId());
 		return true;
 	}
+
+	public function handleRequestAbility(RequestAbilityPacket $packet) : bool{
```
ShockedPlot7560: Why is this function present?

ShockedPlot7560 (commented):
Some phpstan and style correction need to be adresses, but the logic looks good

src/network/mcpe/handler/InGamePacketHandler.php
Bqleine: It was removed after the pull request creation, it will go away when I rebase.
https://github.com/pmmp/PocketMine-MP/commit/fe3e2cc90a002b6a19e4b515d01f12dd9855cfc9#diff-f405c0ec75b0721cb66a250d6938b9267fd68285ffa443dfe8e0e156afac04e0L1035

Bqleine: Update, I have no idea what's happening and it did not go away when I rebased

Bqleine:
I've done all the changes I wanted to do for this PR, the two blocks are added and the structure block can be interacted with and changes to structure settings are saved.
The rest of the implementation requires creating a plugin api, so I am leaving this work for a future PR, as written on this PR's description.
One last thing remains to be done, and that is fixing the bug shown in unit tests, but I don't understand what the issue is. Some help would be greatly appreciated.

src/block/StructureBlock.php
```diff
@@ -0,0 +1,67 @@
...
+use pocketmine\network\mcpe\protocol\ContainerOpenPacket;
+use pocketmine\network\mcpe\protocol\types\BlockPosition;
+use pocketmine\network\mcpe\protocol\types\inventory\WindowTypes;
+use pocketmine\player\Player;
+
+class StructureBlock extends Opaque{
+	private StructureBlockType $type;
+
+	public function __construct(BlockIdentifier $idInfo, string $name, BlockTypeInfo $typeInfo){
+		$this->type = StructureBlockType::SAVE;
```
dktapps: This can be specified as property default without using constructor

src/block/StructureBlock.php
```diff
@@ -0,0 +1,67 @@
...
+		parent::__construct($idInfo, $name, $typeInfo);
+	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if ($player instanceof Player) {
+			if (!$player->isCreative(true)) return false;
```
dktapps: Single-line conditions should not be used. Also, it's generally preferred to use `getGamemode()` checks if the exact game mode is needed.

src/block/StructureBlock.php
```diff
@@ -0,0 +1,67 @@
...
+			// the 0 is dubious but any value works
+			$pk = ContainerOpenPacket::blockInv(0, WindowTypes::STRUCTURE_EDITOR, BlockPosition::fromVector3($this->getPosition()));
```
dktapps, line 49: I do not like this. Packets shouldn't be directly baked into the core code like this.

src/block/utils/StructureBlockType.php
```diff
@@ -0,0 +1,63 @@
...
+		return match($type){
+			0 => self::DATA,
+			1 => self::SAVE,
+			2 => self::LOAD,
+			3 => self::CORNER,
+			4 => self::INVALID,
+			5 => self::EXPORT,
+			default => throw new \ValueError("Unknown structure block type " . $type),
+		};
+	}
```
dktapps, line 48: These should not be visible in the API. The general way to handle these would be a map in `pocketmine\data\bedrock`.

src/block/utils/StructureBlockType.php
```diff
@@ -0,0 +1,63 @@
...
+	public function toInt() : int{
+		return match($this){
+			self::DATA => 0,
+			self::SAVE => 1,
+			self::LOAD => 2,
+			self::CORNER => 3,
+			self::INVALID => 4,
+			self::EXPORT => 5,
+		};
+	}
```
dktapps, line 62: Same again here.

src/block/utils/StructureRotation.php
```diff
@@ -0,0 +1,59 @@
...
+
+	public function getAngle() : int{
+		return $this->quarterTurns * 90;
+	}
+
+	/**
+	 * @param $angle The rotation in degrees. This must be a multiple of 90.
+	 * @throws \ValueError
+	 */
+	public function setAngle(int $angle) : void{
```
dktapps, line 45: Why is this class mutable?

src/block/utils/StructureRotation.php
```diff
@@ -0,0 +1,59 @@
...
+			throw new \ValueError("StructureRotation must be a multiple of 90, " . $angle . " given.");
+		}
+		$this->quarterTurns = ($angle / 90) % 4;
+	}
+
+	public function getQuarterTurns() : int{
+		return $this->quarterTurns;
+	}
+
+	public function setQuarterTurns(int $quarterTurns) : void{
```
dktapps, line 56: See above

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -1032,4 +1036,28 @@ public function handleEmote(EmotePacket $packet) : bool{
...
+			return false;
+		}
+
+		$tile = $world->getTileAt($pos->getX(), $pos->getY(), $pos->getZ());
+		$block = $world->getBlockAt($pos->getX(), $pos->getY(), $pos->getZ());
+		if($tile instanceof TileStructureBlock && $block instanceof StructureBlock && $this->player->canInteract($block->getPosition(), 15)){
+			try{
+				$type = StructureBlockType::fromInt($packet->structureEditorData->structureBlockType);
+				$tile->updateFromPacket($packet);
+			}catch(\ValueError $e){
```
dktapps, line 1047: Why is this being handled? This is an engine error that could be generated by any defective code.

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -1032,4 +1036,28 @@ public function handleEmote(EmotePacket $packet) : bool{
...
+				throw PacketHandlingException::wrap($e);
+			}
+			$block->setType($type);
```
dktapps, line 1050: This won't do anything unless you set the block back into the world

src/block/utils/StructureVoidType.php
```diff
@@ -0,0 +1,29 @@
...
+ *
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\block\utils;
+
+enum StructureVoidType{
+	case VOID;
+	case AIR;
```
dktapps, line 28: What's the difference between these two types?

src/block/utils/StructureAxes.php
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	public function getXMirror() : bool{
+		return $this->xMirror;
+	}
+
+	/**
+	 * @return $this
+	 */
+	public function setXMirror(bool $xMirror) : self{
```
dktapps, line 55: Again, why is this mutable?

src/block/utils/StructureAxes.php
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	public function getYMirror() : bool{
+		return $this->yMirror;
+	}
+
+	/**
+	 * @return $this
+	 */
+	public function setYMirror(bool $yMirror) : self{
```
dktapps, line 67: And again here

src/block/utils/StructureAnimationMode.php
```diff
@@ -0,0 +1,54 @@
...
+	 * @throws \ValueError
+	 */
+	public static function fromInt(int $mode) : StructureAnimationMode{
+		return match($mode){
+			0 => self::NONE,
+			1 => self::BY_LAYER,
+			2 => self::BY_BLOCK,
+			default => throw new \ValueError("Unknown structure animation mode " . $mode),
+		};
+	}
```
dktapps, line 42: These should be handled in `pocketmine\data`

src/block/utils/StructureAnimationMode.php
```diff
@@ -0,0 +1,54 @@
...
+	/**
+	 * @internal
+	 */
+	public function toInt() : int{
+		return match($this){
+			self::NONE => 0,
+			self::BY_LAYER => 1,
+			self::BY_BLOCK => 2,
+		};
+	}
```
dktapps, line 53: and again here

dktapps (changes requested):
This PR is trying to do multiple things at once. Most of the Structure Void stuff is fine, but the majority of the PR is Structure Block stuff which I find has been implemented poorly. Please split the two blocks into separate PRs.

Bqleine:
Thank you for the complete review, I will make a separate PR for the structure void.
I agree that the quality of the code is not optimal, this is my first PR and I am still learning about the inner workings of the codebase, I will take some time in the coming days to improve on the points you commented. :+1:

src/block/utils/StructureVoidType.php
Bqleine: From the [minecraft wiki](https://minecraft.wiki/w/Structure_Air#Usage):
> It exists as a data value of [structure voids](https://minecraft.wiki/w/Structure_void). It acts similarly to structure voids. When holding a structure air or structure void, it becomes visible. Structure air and structure voids can stack together, the icon of the stack is the icon of whichever block was most recently added to the stack. Transparent blocks can be placed on them just like with structure voids. In the inventory, structure air is called structure void. If you give yourself a structure air then whenever you hold a structure void it will look like a structure air in your hand in both third and first person. Structure air are copied by structure blocks. This makes structure air very useful so always use structure air for your creative builds instead of structure voids. Note: structure air becomes structure voids when placed by hand, but if placed with /[setblock](https://minecraft.wiki/w/Commands/setblock) ~ ~ ~ structure_void ["structure_void_type":"air"] it will stay as a structure air.

I was not able to completely verify this information, the difference that matters in this pr is that vanilla minecraft will forbid you from placing structure airs, and in this pr you can place them fine.

Bqleine: And you can't stack two different types in the pr

dktapps: The wiki entry is gibberish that conveys no actual information. That's why I asked.

RoyalMCPE: The difference is that structure air will replace whatever block is at that position with an air block, while structure void will just leave it be

dktapps: That's exactly what normal air does ...

Bqleine: Depends on the structure block settings iirc

dktapps: Meaning what?

Bqleine: I was thinking about the Remove Blocks setting, but it doesn't help (its just another weird setting).

After testing on 1.20.41.02, I was unable to find a useful difference between structure air and structure void. I was able to reproduce normal behavior with both:
> Structure voids can be placed as part of a structure and are ignored when the structure file is loaded. This results in the blocks at the location where the structure is placed being maintained and not being overridden.

This is exactly what happened :

> For example, if a player saves a 2 block high structure with air at the top and a structure void at the bottom, and loads it into a 2 block high area of stone, the top block becomes air but the bottom stays stone.

What did **not** happen:
 - "Structure air are copied by structure blocks."

TL;DR I was not able to find any difference in the structure block usage. I left a comment on the wiki page

src/block/StructureBlock.php
Bqleine: Should I create an Inventory extending BlockInventory and TemporaryInventory with no slots then ? Sounds odd

src/block/utils/StructureVoidType.php
dktapps: Did you try pasting structures underwater with this? I wonder if there's a difference in the water displacement behaviour.

dktapps:
Closing due to conflicts and inactivity

Issue #5960, opened 2023-08-03
Do not auto-register `PluginBase->onCommand()`


dktapps:
## Description
`onCommand()` has been a point of confusion for plugin developers for many years.

Due to its name, it implies that it should behave like an event handler, where `registerEvents()` is enough to make it work. This leads newbies to believe that it should be easy to put `onCommand()` into another class just like event handlers.

However, due to it being auto-registered, it actually behaves more like `onEnable`, and it's rather impractical to make it work in a different class.

This is a problem I personally encountered when developing my first plugin way back in 2016, and it caused no end of headaches for me as I tried to figure out how to split my `onCommand()` into different classes. In the end I ended up realizing it was more or less impossible, and writing a dispatcher that redirected commands to their appropriate places.

This issue suggests that we don't auto-register `onCommand()` at all, but instead require it to be manually registered in a similar manner to event listeners. This would make the behaviour less magical and help people to make better sense of how it works.

In addition, it would likely be better to rewrite `CommandExecutor` to accept an arbitrary `Closure`, which would make it much easier to work with.

## Problems
- People migrating to the new version in which `onCommand` doesn't magically work anymore would suddenly find their `onCommand` not working. This is not a great experience, although we could easily generate warnings that a `plugin.yml` declared command doesn't have any registered handler.
- Not auto-registering `onCommand` makes it a tiny bit more inconvenient to use `plugin.yml` declared commands, although the value of such commands might be questionable anyway.

## Upsides
- No more people trying to extend `PluginBase` in a second class because they think it will magically make `onCommand()` work
- No more people thinking that `onCommand()` is an event handler

dktapps:
Searching `onCommand author:dktapps` on Discord yields a lot of discussions over the span of many years on this topic

dktapps:
Closing this in favour of #6506, which is what we all _really_ want to do.

PR #6063, opened 2023-09-21
Added ClosureCommand

dktapps:
this is intended to replace PluginCommand and CommandExecutor, both of which are overengineered and unfit for purpose. Allowing a closure allows much greater flexibility.

We can't use this within the core yet, as plugins will expect PluginBase->getCommand() to return PluginCommand (with its associated setExecutor() and similar APIs). However, I think this is useful enough to add by itself.

### API changes
- Added a new `ClosureCommand` class. The callback accepted is similar to (and compatible with) the signatures required by `CommandExecutor` in the past, allowing easy migration.

### Behavioural changes
None. This class is not yet used in the core.

While I'd like to migrate directly to this, various plugins expect that `PluginCommand` will be returned by `PluginBase->getCommand()`.

In addition, there are questions about identifying plugin commands if this is used. This is no different a problem than we already see with regular `Command` users, but perhaps worth mentioning anyway.

## Backwards compatibility
No BC issues.

## Follow-up
- Deprecate and remove `PluginCommand` and `CommandExecutor`
- Perhaps #5960 ?

## Tests
Not directly tested, but this is trivially PHPStan-verifiable.


ShockedPlot7560 (dismissed):

src/command/ClosureCommand.php
```diff
@@ -0,0 +1,57 @@
...
+
+namespace pocketmine\command;
+
+use pocketmine\lang\Translatable;
+use pocketmine\utils\Utils;
+
+/**
+ * @phpstan-type Execute \Closure(CommandSender $sender, Command $command, string $commandLabel, list<string> $args) : mixed
+ */
+class ClosureCommand extends Command{
```
SOF3: why is this not final?

dktapps: I have no idea. My class templates are final by default ...

dktapps:
I'm not sure whether we should allow replacing the closure handler a la `CommandExecutor->setExecutor()`. `setExecutor()` is used by some plugins to provide custom handlers for commands declared via `pocketmine.yml`. The command can just be re-registered with the new handler, so I'm not sure whether this is needed or not.

SOF3:
Let's not overcomplicate the API. Mutable executor may as well be implemented within the closure as a delegation where it matters.

```php
$this->cmd = new ClosureCommand(xxx);
$this->cmd->setClosure(xxx);
register($this->cmd);
```

is barely different from

```php
$this->closure = xxx;
register(new ClosureCommand(fn(...$args) => ($this->closure)(...$args)));
```

dktapps:
Delegate won't be available if the command was registered by the server from plugin.yml. That's basically the whole use case.

dktapps:
setExecutor would become redundant if we decide to go forward with #6506

dktapps:
I think this is fine as-is. If we decide to switch `plugin.yml` commands to this class in the future (as opposed to getting rid of them entirely), we can always add `setExecutor()` then when the time comes.

PR #6064, opened 2023-09-21
Implement client-side support of downloading RP from CDN

### LINKED ISSUE #3127: Resource pack sending overloads low-bandwidth connections
(I use Google Translate.)
# Issue description

(Android(Server) => Windows(Client))
(Windows => Windows)
I'm sorry, but I couldn't reproduce it in the above environment.

(This problem is very similar to #2026.)

Expected result: resource pack is successfully sent to the client.
Actual result: The transmission of resourcepack will stop at 1MB and after a few seconds the player will be kicked out in an timeout.(A resourcepack of 1MB or less will be sent to the client normally.)

# Steps to reproduce the issue

     1. Download https://aka.ms/resourcepacktemplate and go to the resource_packs folder.
     2. Write `-Vanilla_Resource_Pack_1.12.0.zip` on resource_packs.yml to activate the resource pack.
     3. Start the server and log in to the server.
     4. When `Download World Resource Packs?` Is displayed, click `Download & Join`.
     5. The console says "Blocked 192.168.0.22 for 300 seconds", the download pack stops at 1 MB, and after a few seconds the player is kicked out with a timeout

# OS and versions
    PocketMine-MP: 3.9.4
    PHP: php 7.2.19 (Built by php-build-scripts)
    Server OS: Android
    Game version: PE (Android => Android)(loopback)
    Minecraft Version: 1.12.0
APP: https://play.google.com/store/apps/details?id=jackpal.androidterm
# Plugins
No plugins are installed.

# Crashdump, backtrace or other files
None

# php Version
```
PHP 7.2.19 (cli) (built: Jul 15 2019 19:55:21) ( ZTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
```

# Console Output
```bash
u0_a342@SOT31:/ $ cd /sdcard/www/PocketMine4/
env TMPDIR='/storage/emulated/0/PocketMine/tmp' LESMI_RESOLV_CONF_DIR=/sdcard/PocketMine/resolv.conf /data/data/jackpal.androidterm/app_HOME/php-7.2.19 -c /sdcard/PocketMine/config/php.ini /sdcard/www/PocketMine3/PocketMine-MP.phar
[06:48:55] [Server thread/WARNING]: ChunkUtils extension is missing. Anvil-format worlds will experience degraded performance.
[06:48:55] [Server thread/INFO]: Loading pocketmine.yml...
[06:48:55] [Server thread/INFO]: Loading server properties...
[06:48:55] [Server thread/ERROR]: Missing required language file phar:///storage/emulated/0/www/PocketMine3/PocketMine-MP.phar/src/pocketmine/lang/locale/us.ini
[06:48:55] [Server thread/INFO]: Selected English (us) as the base language
[06:48:55] [Server thread/INFO]: Starting Minecraft: Bedrock Edition server version v1.12.0
[06:48:55] [Server thread/NOTICE]: Online mode is enabled. The server will verify that players are authenticated to Xbox Live.
[06:48:55] [Server thread/NOTICE]: To disable authentication, set "xbox-auth" to "false" in server.properties.
[06:48:55] [Server thread/INFO]: Opening server on 0.0.0.0:19132
[06:48:55] [Server thread/DEBUG]: Server unique id: ba2aefde-c3cb-b5f3-1aa1-57341c7e8697
[06:48:55] [Server thread/DEBUG]: Machine unique id: ba1ac2d8-3b61-0aa0-f2fc-3918c85a7f47
[06:48:55] [Server thread/INFO]: This server is running PocketMine-MP version 3.9.4
[06:48:55] [Server thread/INFO]: PocketMine-MP is distributed under the LGPL License
[06:48:56] [Server thread/INFO]: Loading resource packs...
[06:48:56] [Server thread/DEBUG]: Successfully loaded 1 resource packs
[06:48:56] [Asynchronous Worker #0 thread/DEBUG]: Set memory limit to 256 MB
[06:48:56] [Server thread/INFO]: Preparing world "world"
[06:48:56] [Server thread/INFO]: Starting GS4 status listener
[06:48:56] [Server thread/INFO]: Setting query port to 19132
[06:48:56] [Server thread/INFO]: Query running on 0.0.0.0:19132
[06:48:56] [Server thread/INFO]: Default game type: Survival Mode
[06:48:56] [Server thread/INFO]: Done (0.804s)! For help, type "help" or "?"
[06:49:01] [Server thread/DEBUG]: [AutoUpdater] Async update check failed due to "Resolving timed out after 4000 milliseconds"
[06:49:12] [RakLibServer thread/DEBUG]: Created session for 192.168.0.22 41678 with MTU size 1492
[06:49:12] [Server thread/DEBUG]: Radish12345a is logged into Xbox Live
[06:49:12] [Server thread/DEBUG]: Unhandled ClientCacheStatusPacket received from Radish12345a: gQEB
[06:49:14] [RakLibServer thread/NOTICE]: Blocked 192.168.0.22 for 300 seconds
[06:49:24] [RakLibServer thread/DEBUG]: Closed session for 192.168.0.22 41678
[06:49:24] [Server thread/INFO]: Radish12345a[/192.168.0.22:41678] logged out due to timeout
```

# Other images
Screenshot_20190925-064920
### ↓
Screenshot_20190925-064929

alvin0319:
## Introduction

Implement client-side support of downloading Resource Pack from CDN.

### Relevant issues
* Alternative way to fix #3127

### API changes

* Added `ResourcePackManager::getPackURLs()`: Returns a map of uuid_version => cdn url

### Behavioural changes

The server does not send ResourcePackChunkDataPacket since the client downloads the resource pack from CDN.

## Backwards compatibility

There are no known BC breaking changes.

## Tests

Test CDN server used with this test: https://github.com/alvin-pm-pl/ResourcePackCDN

Test Pack used with this test: https://mcpedl.com/legendary-bedrock-rtx-texture-pack/

https://github.com/pmmp/PocketMine-MP/assets/32565818/34a0fe98-a6e2-47d1-ae16-5d85fa14eabe


src/resourcepacks/ResourcePackManager.php
```diff
@@ -58,6 +58,9 @@ class ResourcePackManager{
 	 */
 	private array $encryptionKeys = [];
 
+	/** @var string[] */
```
ShockedPlot7560:
```suggestion
	/** @var array<string, string> */
```

src/resourcepacks/ResourcePackManager.php
```diff
@@ -169,6 +184,14 @@ public function getResourceStack() : array{
 		return $this->resourcePacks;
 	}
 
+	/**
+	 * Returns an array of pack URLs.
+	 * @return string[]
```
ShockedPlot7560:
```suggestion
	 * @return array<string, string>
```

ShockedPlot7560 (commented):
Looks OK for me

BrandPVP:
> «However, the resource packs still need to be loaded into PocketMine.»

It should automatically download the pack from the url and cache the metadata, shouldn’t it? Like, why do users need to put it manually when PM can automatize the job

alvin0319:
> > «However, the resource packs still need to be loaded into PocketMine.»
> 
> It should automatically download the pack from the url and cache the metadata, shouldn’t it? Like, why do users need to put it manually when PM can automatize the job

I don't think it should be a PM's job. We should consider a offline environment.

ShockedPlot7560:
In fact, I'm going back on my opinion because I hadn't realized the problem that has just been raised. Although you need to be in an offline environment, it can be interesting to enable automation.

alvin0319:
Well, the problem is that:
1. We can't download the resource packs from CDN every time we turn on the server.
2. If we cache the downloaded resource pack, we have no way to detect whether resource pack content changed unless CDN provides a checksum or sha1 hash of the file.

The reason for requiring Resource Packs to be loaded into PM is that the client also expects info of remote resource packs in ResourcePacksInfoPacket. (after that, this isn't necessary since they download from external)

As an alternative way, we could let users write the UUID and version of the resource pack and optionally content key (which is the least information the client expects) in resource_packs.yml so that we don't let users load the resource pack into PM.

src/resourcepacks/ResourcePackManager.php
dktapps: This syntax should only be used in @phpstan-var.  Regular @var should be PSR-5 compliant.

src/resourcepacks/ResourcePackManager.php
dktapps: Same as above.

dktapps:
General sentiment in the dev community so far is that this feature is pretty useless for anyone who doesn't have a perfect connection (i.e. anywhere that isn't the US or Europe).

src/resourcepacks/ResourcePackManager.php
```diff
@@ -113,6 +121,13 @@ public function __construct(string $path, \Logger $logger){
 					}
 					$this->encryptionKeys[$index] = $key;
 				}
+				if(isset($remotePacks[$pack])){
+					$url = $remotePacks[$pack];
+					if(!is_string($url)){
```
jasonw4331: Please do a regex validation for URLs here too

jasonw4331 (changes requested):
I would like to see added support for URLs without the https:// header also.

dktapps:
I do think the cleanest way to do this would be to just allow putting URLs in the `resource_stack` instead of filenames. The way this is currently done seems janky to me.

> Well, the problem is that:
> 
> 1. We can't download the resource packs from CDN every time we turn on the server.
> 2. If we cache the downloaded resource pack, we have no way to detect whether resource pack content changed unless CDN provides a checksum or sha1 hash of the file.
> 
> The reason for requiring Resource Packs to be loaded into PM is that the client also expects info of remote resource packs in ResourcePacksInfoPacket. (after that, this isn't necessary since they download from external)
> 
> As an alternative way, we could let users write the UUID and version of the resource pack and optionally content key (which is the least information the client expects) in resource_packs.yml so that we don't let users load the resource pack into PM.

I'm not understanding the logic here. What's stopping PM from downloading packs from CDN whenever the server boots up? 

I don't really buy into the point about caching either. If we have users hardcode pack UUIDs into the YML, we're already caching info, just in a way that the server can't control programmatically. Vs if the pack is downloaded, at least the information is guaranteed to be correct for _some_ version of the resource pack. We can also look at `Cache-Control` headers to see when a resource pack's local cache should expire.

dktapps:
This also needs to be updated for latest protocol changes

PR #6069, opened 2023-09-29
fix(Entity): `broadcastSound()` not firing `WorldSoundEvent`

cooldogedev:
### Behavioural changes

`Entity::broadcastSound()` now fires `WorldSoundEvent`

### DIFF
--- a/src/entity/Entity.php
+++ b/src/entity/Entity.php
@@ -1686,7 +1686,7 @@ abstract class Entity{
 	 */
 	public function broadcastSound(Sound $sound, ?array $targets = null) : void{
 		if(!$this->silent){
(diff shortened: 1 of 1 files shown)


dktapps:
This should target `stable`

ShockedPlot7560:
Was there any particular reason why it was implemented this way?

dktapps:
Not that I can remember. Functionally, this changed version is the same.

That being said though, I wonder if we should have an `EntitySoundEvent` so that the source of the sound is obvious.

ShockedPlot7560:
> That being said though, I wonder if we should have an `EntitySoundEvent` so that the source of the sound is obvious.

Question for another PR, it doesnt rely on this specific case which tend to just fix a bug. But the reflection is correct and would be taken into consideration for the future yes

dktapps:
Yeah, you're right. I think this is good to merge by itself.

PR #6071, opened 2023-09-30
Fixed dirt related blocks item interaction to match up with vanilla

JavierLeon9966:
## Introduction

The grass and dirt block item interaction doesn't match up with vanilla.
### Relevant issues

None
### API changes

None
### Behavioural changes

None
## Backwards compatibility

None
## Follow-up

None
## Tests

None

### DIFF
--- a/src/block/Dirt.php
+++ b/src/block/Dirt.php
@@ -60,8 +60,8 @@ class Dirt extends Opaque{
 	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
 		$world = $this->position->getWorld();
 		if($face !== Facing::DOWN && $item instanceof Hoe){
-			$down = $this->getSide(Facing::UP);
-			if($down->getTypeId() !== BlockTypeIds::AIR){
+			$up = $this->getSide(Facing::UP);
+			if($up->getTypeId() !== BlockTypeIds::AIR){
 				return true;
 			}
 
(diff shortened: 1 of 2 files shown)


src/block/Dirt.php
```diff
@@ -59,7 +59,12 @@ public function setDirtType(DirtType $dirtType) : self{
 
 	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
 		$world = $this->position->getWorld();
-		if($face === Facing::UP && $item instanceof Hoe){
+		if($face !== Facing::DOWN && $item instanceof Hoe){
+			$down = $this->getSide(Facing::UP);
```
ShockedPlot7560: Its up, not down, no ?

JavierLeon9966: Oh, right. I didn't notice the variable name from the copy paste

ShockedPlot7560 (dismissed):

src/block/Grass.php
```diff
@@ -91,14 +91,18 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
 			TallGrassObject::growGrass($world, $this->position, new Random(mt_rand()), 8, 2);
 
 			return true;
-		}elseif($item instanceof Hoe){
+		}
+		if($face === Facing::DOWN){
```
dktapps: So it can be fertilized even from the bottom face? Seems weird

JavierLeon9966: Yes. I tested it in vanilla.

dktapps: To be honest I wouldn't be shocked if this was a bug in vanilla, but fine. I don't know if I like throwing an early return in like this though. Seems like it would be easy to get unexpected behaviour in the future if more actions need to be added.

PR #6076, opened 2023-10-07
Add flight speed

Sergittos:
## Introduction
This pull request makes it possible for developers to modify the player's speed when flying.

### Relevant issues
Closes #5155

### API changes
Added the following methods in `Player`:

* `Player->setFlightSpeedMultiplier()`: Sets the player's speed when flying.
* `Player->getFlightSpeedMultiplier()`: Returns the player's speed when flying.

Added the following constant in `Player`:

* `Player::DEFAULT_FLIGHT_SPEED_MULTIPLIER`

## Tests
https://streamable.com/vctb3y


src/player/Player.php
```diff
@@ -515,6 +519,15 @@ public function hasAutoJump() : bool{
 		return $this->autoJump;
 	}
 
+	public function setFlySpeed(float $flySpeed) : void{
```
ShockedPlot7560: Need probably a value checking, can the flySpeed negative or equal 0 ?

Sergittos: You're right. Fly speed can be 0, but not negative

dktapps:
My issue with this is the same as the past PRs on the subject. There's no explanation or documentation of what the values mean, or why it's 0.05 by default. What's the unit? Blocks per seconds? Blocks per tick?

There's also going to be questions like why doesn't `setMovementSpeed()` affect flying speed, and why the server-side flight speed doesn't change when sprint-flying (which is inconsistent with movement speed).

The name `setFlightSpeed()` would be more in line with functions like `setMovementSpeed()`.

ShockedPlot7560:
I also think the name isn't quite right. It will indeed be confusing. However, I don't think that the unity problem should hinder the merge of adding APIs. It's up to the developers to find out what they're modifying, and a warning could be added in the docs about this. That the unit is not determined and that the value should be handled with care?

src/player/Player.php
```diff
@@ -515,6 +519,17 @@ public function hasAutoJump() : bool{
 		return $this->autoJump;
 	}
 
+	public function setFlightSpeed(float $flightSpeed) : void{
+		if($this->flightSpeed !== $flightSpeed && $flightSpeed >= 0){
```
ShockedPlot7560: I would have reversed the conditions

Sergittos: Done!

ShockedPlot7560 (dismissed):
Looks ok to me

src/player/Player.php
```diff
@@ -515,6 +519,17 @@ public function hasAutoJump() : bool{
 		return $this->autoJump;
 	}
 
+	public function setFlightSpeed(float $flightSpeed) : void{
+		if($flightSpeed >= 0 && $this->flightSpeed !== $flightSpeed){
```
Gaprix: Why should we disallow setting negative flight speed values? Negative values make the player fly backwards, and this can be used in some game modes

Sergittos: Yeah you're right

Gaprix (approved):

IvanCraft623:
Apparently with tests I have done `$value * 10` are the blocks per tick that the player will fly. So the default value `0.05` sets the player's flight speed to around `0.5` blocks per tick.

I've been testing if other attributes like movementSpeed or walkSpeed can modify the flight speed, but that doesn't happen. Alosi it seems that walkSpeed only modifies the player's fov.

When a player sprints while flying his speed is duplicated, with the default value the player's flight speed will be around `1` block per tick. The curious thing is that this only happens in client-side.

TobiasGrether:
@Sergittos @ShockedPlot7560 I think this PR is fine once documentation on the value of the FlySpeed is present. As in, an explanation as to how the values are calculated.

UnknownOre:
> Apparently with tests I have done `$value * 10` are the blocks per tick that the player will fly. So the default value `0.05` sets the player's flight speed to around `0.5` blocks per tick.
> 
> I've been testing if other attributes like movementSpeed or walkSpeed can modify the flight speed, but that doesn't happen. Alosi it seems that walkSpeed only modifies the player's fov.
> 
> When a player sprints while flying his speed is duplicated, with the default value the player's flight speed will be around `1` block per tick. The curious thing is that this only happens in client-side.

if this correct can't we use blocks it can travel per tick as a parameter in the function instead of magic numbers

ShockedPlot7560:
If @IvanCraft623 researchs is correct, we can document the method with this and it will be ready to merge for me.

Sergittos:
> @Sergittos @ShockedPlot7560 I think this PR is fine once documentation on the value of the FlySpeed is present. As in, an explanation as to how the values are calculated.

Done.

ShockedPlot7560 (dismissed):
Documentation is enough for me.

src/player/Player.php
```diff
@@ -504,6 +508,39 @@ public function isFlying() : bool{
...
+	/**
+	 * Sets the player's speed when flying.
+	 *
+	 * The flight speed is calculated as `blocks per tick = flightSpeed * 10`. For example, setting the
+	 * flight speed to 0.05 means the player will fly at approximately 0.5 blocks per tick.
+	 *
+	 * If set to zero, the player will not be able to move in the xz plane when flying, and negative values
+	 * will invert the controls.
+	 *
+	 * Notes:
```
IvanCraft623: Document negative values behavour

IvanCraft623 (changes requested):

src/player/Player.php
Sergittos: The negative values behaviour is already documented on the 518 line and at the end of the 517 line, do I need to change something?

IvanCraft623: nah

IvanCraft623 (dismissed):

dktapps:
I'm wondering if "flight speed multiplier" might be more appropriate, considering that it doesn't directly map onto actual flight speed (see above comments about sprint-flying). I also don't like exposing magic values to users.

ShockedPlot7560:
Mentioning that it is a multiplier may be preferable in view of the way it works.

dktapps (approved):
@pmmp/server-developers anyone want to give this a final look over before merge?

ipad54 (approved):

PR #6078, opened 2023-10-07
Inform blocks of which adjacent blocks triggered an update

### LINKED ISSUE #5982: Inform blocks of which adjacent blocks triggered an update
## Description
Blocks like Cactus have a costly update check, checking 5 out of 6 adjacent blocks for support conditions. This could be made cheaper if the update function knew which adjacent block(s) caused the update to be fired.

## Justification
Moar performance

ShockedPlot7560:
## Introduction
Some blocks have a certain cost to check their support during NearbyBlockUpdate. This PR tends to provide a way for blocks to know which sides triggered the update.

### Relevant issues
* Fixes #5982 

### API changes
- Deprecated `Block::onNearbyBlockChange()`
- Add `Block::onNearbyBlockChange2(int $flags)` where flags is a set of flags provided by `NearbyBlockChangeFlags`

### Behavioural changes
- Improve performance for some blocks like Cactus


src/block/Block.php
```diff
@@ -454,11 +455,21 @@ public function onBreak(Item $item, ?Player $player = null, array &$returnedItem
...
 	 */
 	public function onNearbyBlockChange() : void{
 
 	}
 
+	/**
+	 * Called when this block or a block immediately adjacent to it changes state.
+	 * @phpstan-param NearbyBlockChangeFlags::FLAG_* $flags
+	 */
+	public function onNearbyBlockChange2(int $flags) : void{
```
JavierLeon9966, line 496: Instead of making a new method you should just add the parameter into the already existing one, but add the default flag `0`.

src/block/Block.php
```diff
@@ -454,11 +455,21 @@ public function onBreak(Item $item, ?Player $player = null, array &$returnedItem
...
+	 *
+	 * @deprecated {@see Block::onNearbyBlockChange2()}
 	 */
 	public function onNearbyBlockChange() : void{
 
 	}
 
+	/**
+	 * Called when this block or a block immediately adjacent to it changes state.
+	 * @phpstan-param NearbyBlockChangeFlags::FLAG_* $flags
```
JavierLeon9966:
```suggestion
	 * @phpstan-param int-mask-of<NearbyBlockChangeFlags::FLAG_*> $flags
```

JavierLeon9966 (changes requested):

src/block/Block.php
ShockedPlot7560: Its [break the BC](https://onlinephp.io?s=s7EvyCjg5eLlSs5JLC5WcMrJT85WqObl4iwoTcrJTFZIK81LLsnMz1PIz_NLTSxKqgSrcM5IzEtP1cjMK1FQSctJTC9WsFUw0LRSKMvPTAFp5gTiWl6uWoTBzonJJaXFCqkVJal5KVB7iLUGh8EqEJNtFfJSy6Hma2haAwA%2C&v=8.2.10) and the ambition is to do without. We simply need to find a better name for the function.

src/block/Cactus.php
```diff
@@ -69,9 +69,8 @@ public function onNearbyBlockChange2(int $flags) : void{
 				$this->position->getWorld()->useBreakOn($this->position);
 			}
 		}
-		foreach(NearbyBlockChangeFlags::getSides($flags) as $side){
-			if($side !== NearbyBlockChangeFlags::FLAG_UP){
-				$facing = NearbyBlockChangeFlags::toFacing($side);
+		foreach(NearbyBlockChangeFlags::getSidesAsFacing($flags) as $facing){
```
dktapps: getFaces() would be fine

src/block/Bell.php
```diff
@@ -108,16 +110,19 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
...
 			BellAttachmentType::FLOOR => [Facing::DOWN],
 			BellAttachmentType::ONE_WALL => [Facing::opposite($this->facing)],
 			BellAttachmentType::TWO_WALLS => [$this->facing, Facing::opposite($this->facing)]
-		} as $supportBlockDirection){
-			if(!$this->canBeSupportedAt($this, $supportBlockDirection)){
-				$this->position->getWorld()->useBreakOn($this->position);
-				break;
+		};
+		foreach(NearbyBlockChangeFlags::getSidesAsFacing($flags) as $supportBlockDirection){
+			if(in_array($supportBlockDirection, $sides, true)){
```
dktapps: ew no. don't ever use in_array in performance sensitive code

dktapps: This mess can be done better using `containFacing`

src/block/Block.php
```diff
@@ -454,11 +455,21 @@ public function onBreak(Item $item, ?Player $player = null, array &$returnedItem
 
 	/**
 	 * Called when this block or a block immediately adjacent to it changes state.
+	 *
+	 * @deprecated {@see Block::onNearbyBlockChange2()}
```
dktapps: I'm not sure if this actually works. Better to put the @see on its own line rather than writing it inline.

dktapps (commented):
I'm not loving this implementation. It's introducing expensive array logic into hot paths.

src/block/utils/NearbyBlockChangeFlags.php
```diff
@@ -0,0 +1,137 @@
...
+	 */
+	public static function getSidesAsFacing(int $flag) : array{
+		return array_map(fn(int $facing) => self::toFacing($facing), self::getSides($flag));
+	}
+
+	/**
+	 * @phpstan-param Flag $flag
+	 * @phpstan-param FlagValue ...$others
+	 */
+	public static function contain(int $flag, int ...$others) : bool{
```
dktapps: This function is unnecessarily costly and also boilerplaty.

e.g.
```php
NearbyBlockChangeFlags::contain($flags, NearbyBlockChangeFlags::FLAG_UP, NearbyBlockChangeFlags::FLAG_DOWN)
```
vs
```php
($flags & (NearbyBlockChangeFlags::FLAG_UP | NearbyBlockChangeFlags::FLAG_DOWN)) !== 0
```

dktapps:
I've made some adjustments and simplified the logic.

StaticSupportTrait could benefit from this, but it would require all of its users to declare which face they are supported by, which is a whole bunch of work. Fortunately, it hasn't made it into a release yet as far as I remember, and SST is already a PITA since it's easy to accidentally look at the wrong block in canBeSupportedAt(), so I was thinking about making a change which would solve this anyway.

src/block/utils/NearbyBlockChangeFlags.php
```diff
@@ -0,0 +1,125 @@
...
+			Facing::NORTH => self::FLAG_NORTH,
+			Facing::SOUTH => self::FLAG_SOUTH,
+			Facing::WEST => self::FLAG_WEST,
+			Facing::EAST => self::FLAG_EAST,
+			default => throw new \InvalidArgumentException("Unknown facing $facing"),
+		};
+	}
+
+	/**
+	 * @phpstan-param FlagValue $flag
```
ShockedPlot7560: Should be `value-of<self::ALL_FACING>`

src/world/World.php
```diff
@@ -299,8 +300,8 @@ class World implements ChunkManager{
 	/** @phpstan-var \SplQueue<int> */
 	private \SplQueue $neighbourBlockUpdateQueue;
 	/**
-	 * @var true[] blockhash => dummy
-	 * @phpstan-var array<BlockPosHash, true>
+	 * @var int[] blockhash => dummy
```
ShockedPlot7560:
```suggestion
	 * @var int[] blockhash => update reference block flags
```

dktapps: where does "reference" come into it?

src/block/utils/NearbyBlockChangeFlags.php
dktapps: To be honest I'm not sure if this function should reject FLAG_SELF. It's easy to see how this could result in crashes, and before my amendments there were a couple of cases that could have plausibly crashed.

src/world/World.php
ShockedPlot7560: Its mean the blocks referent for the update, no?

dktapps: That's a very confusing way of putting it. "cause flags" would be simpler.

src/block/BaseRail.php
```diff
@@ -220,17 +221,24 @@ private function setConnections(array $connections) : void{
...
-			$world->useBreakOn($this->position);
-		}else{
-			foreach($this->getCurrentShapeConnections() as $connection){
-				if(($connection & RailConnectionInfo::FLAG_ASCEND) !== 0 && !$this->getSide($connection & ~RailConnectionInfo::FLAG_ASCEND)->getSupportType(Facing::UP)->hasEdgeSupport()){
-					$world->useBreakOn($this->position);
+		$break = false;
+		if(($flags & NearbyBlockChangeFlags::DOWN) !== 0 && !$this->getAdjacentSupportType(Facing::DOWN)->hasEdgeSupport()){
+			$break = true;
+		}
+		foreach($this->getCurrentShapeConnections() as $connection){
```
ShockedPlot7560: We will enter the loop even if the break is true

src/block/Cactus.php
```diff
@@ -61,6 +62,21 @@ public function onEntityInside(Entity $entity) : bool{
 		return true;
 	}
 
+	public function onNearbyBlockChange2(int $flags) : void{
+		if(($flags & NearbyBlockChangeFlags::DOWN) !== 0){
+			$supportBlock = $this->getSide(Facing::DOWN);
+			if(!$supportBlock->hasSameTypeId($this) && !$supportBlock->hasTypeTag(BlockTypeTags::SAND)){
+				$this->position->getWorld()->useBreakOn($this->position);
```
ShockedPlot7560: Could be set a new `break` var to avoid entering in the loop later?

dktapps: Maybe. This whole self-destruct thing needs some cleanup work tbh, so I wouldn't think too much about it.

dktapps:
Performance testing suggests that the benefit of this is smaller than I hoped, because we're optimizing the wrong thing.

Local testing with large wheat farms showed that each block update takes about 4.2 microseconds total without the PR. However, the time spent in onNearbyBlockChange() is only about 1.4 microseconds (at least for wheat crops and farmland).

This PR reduces the total to 3.5 us with an average of 0.7 us spent in onNearbyBlockChange2() after integrating with StaticSupportTrait. It's not bad (15-20% performance improvement), but it's not the big win I was hoping for.

It seems like there needs to be more work done to:
- reduce and optimize work done in World->actuallyDoTick() for block updates
- reduce the actual number of block updates (difficult to do without breaking observers)

There are major performance gains to be had in crop farms, where the vast majority of the CPU time spent on crops is actually coming from the useless updates rather than the random growth ticks.

dktapps:
A new comparison profile needs to be done with https://github.com/pmmp/PocketMine-MP/commit/6a3ec70c72e79603b2cc618269461204fd578f2b. I suspect that this commit closes the performance gap by quite a bit.

dktapps:
> A new comparison profile needs to be done with [6a3ec70](https://github.com/pmmp/PocketMine-MP/commit/6a3ec70c72e79603b2cc618269461204fd578f2b). I suspect that this commit closes the performance gap by quite a bit.

Now shows ~7% performance improvement (after optimizations to `Block->getSide()`)
Before: https://timings.pmmp.io/?id=312959
After: https://timings.pmmp.io/?id=312960

Everything is within margin of error, except the Scheduled Block Updates timing, which shows an overall reduction from 7.44 m to 6.94 m (I ignore the average time per tick since it includes ticks where nothing happened).

dktapps:
To paraphrase #internals-dev discussions, I'm unconvinced that this much extra code is justified by a 7% performance increase, particularly since this is a minor part of the problem anyway. I'm going to leave this open for now, but it seems like other ways to improve performance need to be explored.

dktapps:
Closing this as the aforementioned minimal performance gains don't seem to justify the large amount of extra code to maintain.

PR #6080, opened 2023-10-09
Implement growth speed modifiers

### LINKED ISSUE #6070: Implement crop growth speed modifiers
## Description
Things that should influence crop growth:
- Adjacent farmland accelerates (especially if hydrated)
- Adjacent crops of the same type decelerates

## Justification
This has been a lacking feature for many years. It's probably not that difficult to implement, and is needed for vanilla farming mechanics to work as expected.

xSuperr:
## Introduction
Adds vanilla growth speed modifiers

### Relevant issues
Fixes #6070 

### Behavioural changes
Checks the surrounding blocks of the original crop, and crops will no longer grow if below an opaque block, or if that block has a light level of less than 9

## Follow-up
Cleaning up variable names, and the code

Requires translations:

## Tests
Max growth percent is 0.5 when by itself or in a row, with a minimum of 0.3, when surrounded by crops of the same type

### DIFF
--- a/src/block/Crops.php
+++ b/src/block/Crops.php
@@ -31,6 +31,9 @@ use pocketmine\item\Item;
 use pocketmine\math\Facing;
 use pocketmine\math\Vector3;
 use pocketmine\player\Player;
++use function floor;
++use function in_array;
++use function lcg_value;
 use function mt_rand;
 
 abstract class Crops extends Flowable{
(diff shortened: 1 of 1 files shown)


dktapps:
Superseded by 109673382d213222fcf18b0d08ae18b73a3e1e51.

PR #6085, opened 2023-10-10
Fix book placed in wrong slot in Chiseled Bookshelf

diamond-gold:
## Introduction
This PR corrects the slot when a book is placed in Chiseled Bookshelf

## Tests
Tested, the book is placed in the correct slot in all 4 facing directions

### DIFF
--- a/src/block/ChiseledBookshelf.php
+++ b/src/block/ChiseledBookshelf.php
@@ -97,8 +97,13 @@ class ChiseledBookshelf extends Opaque{
 			return false;
 		}
 
+		$x = Facing::axis($face) === Axis::X ? $clickVector->getZ() : $clickVector->getX();
+		$x = match ($face) {
+			Facing::NORTH, Facing::EAST => 1 - $x,
+			default => $x
+		};
 		$slot = ChiseledBookshelfSlot::fromBlockFaceCoordinates(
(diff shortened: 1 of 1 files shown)


src/block/ChiseledBookshelf.php
```diff
@@ -97,8 +97,13 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
 			return false;
 		}
 
+		$x = Facing::axis($face) === Axis::X ? $clickVector->getZ() : $clickVector->getX();
+		$x = match ($face) {
```
dktapps:
```suggestion
		$x = match($face){
```

dktapps (commented):
Good catch, thanks

src/block/ChiseledBookshelf.php
```diff
@@ -97,8 +97,13 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
 			return false;
 		}
 
+		$x = Facing::axis($face) === Axis::X ? $clickVector->getZ() : $clickVector->getX();
+		$x = match($face){
```
dktapps: The more logical way would be to check `Facing::isPositive(Facing::rotateY($face, true))`, in which case `1 - $x` is required. It took me a while to make full sense of why this applies to north and east, when the real issue concerns the axis the bookshelf's face is parallel to.

PR #6090, opened 2023-10-11
Farmland: Remember relative location of nearby water in blockstate data

dktapps:
## Introduction
Farmland can end up scanning up to 162 blocks looking for water in the worst case. This is obviously not great for huge farms where there are thousands of blocks of the stuff.

In most farms, the water won't be moved, and nor will the farmland. This means that we can avoid this costly search on random updates.

This PR implements a cache using blockstate data (only possible in PM5) which stores an index mapping to a coordinate offset where water was previously found by this farmland block. This allows the farmland to avoid water searching entirely in most cases.

This is a colossal improvement as compared to scanning the whole 9x2x9 area every time, which, on average, scans about 40 blocks to find water if the water is at the same Y coordinate. In real terms this translates into about a 8x performance improvement for farmland (see timings below).

### API changes
No public API changes.

### Behavioural changes
- Farmland is now significantly less laggy. Worlds with large farms may see significant performance gains.
- The format of block stateID has changed. The least significant 11 bits are now reserved for state data (previously 8 were reserved).

## Backwards compatibility
- Plugins storing numeric blockstate IDs from `getStateId()` in configs or databases (which they **really shouldn't be doing, as it's clearly documented to be used at RUNTIME ONLY**) will find that their stored data no longer works due to the changed size of the state data reserved size.

## Tests
this has been extensively playtested, and timings have been collected.

Without changes: https://timings.pmmp.io/?id=312635
With changes: https://timings.pmmp.io/?id=312867

Test script: https://gist.github.com/dktapps/f56f05cb03f881a5f96882a7127e05a9
Stand in the middle of the generated farm, and make sure your `chunk-ticking.tick-radius` is at least 6. If you want to accelerate the test, set `chunk-ticking.blocks-per-subchunk-per-tick` to some large number like 50.


dktapps:
Further optimization is possible by using 3x2x1 rather than 3x1x3.
This would reduce the number of blocks scanned from 9 to 6 in a cache hit, without requiring any additional bits (27 possible cache locations instead of 18, both require 5 bits).

dktapps:
Further improvements by tracking exact X/Z offsets (reducing average case to 1.5 blocks checked): https://timings.pmmp.io/?id=312679

I was able to increase state data size to 10 bits without any adverse effects. 11 bits would allow tracking the Y offset, but 11 bits caused a significant jump in the rate of hash collisions using state IDs, so that needs to be worked on further. However, adding 1 more bit only reduces the worst case from 2 blocks to 1 block, so it's diminishing returns anyway.

For now, this is an almost 7x jump in the average performance for farmland, which I'm quite happy with.

dktapps:
This is now good to merge

PR #6092, opened 2023-10-14
Generalize asynchronous events

ShockedPlot7560:
## Introduction
This PR would be more of a foretaste for #6015 in that it provides a means of generalizing asynchronous event calling.

An asynchronous event in this context is one whose action can be deferred by plugins specifying promises to be resolved.  
The integrity of event priorities is guaranteed by calling them one by one. Before moving on to the next priority, we wait until all the promises of the current one have been fulfilled.  
Only the state of the server and player context can't be guaranteed. However, the state of the event is unaffected, whether it is called asynchronous or synchronous.

## Behaviour

When a listener of an asynchronous event returns a `Promise`, the event is marked as asynchronous and placed at the end of the call stack for its priority. 
When a callAsync is called, the call flow is as follows:
```
1. for each priority:
   2. call async handlers which allows concurrent execution
   3. for each async handlers which doesn't allows concurrent execution:
       1. wait for previous promises
       2. clear all the promise
       3. call the handler and add the return promise into the waiting list
   4. wait for all promises
2. the promise returned by callAsync is resolved with their instance
```

In this way, each listener can add only one promise.
If, on the other hand, the handler is sensitive to changes in the event, it is preferable to add the @exclusiveCall tag, which specifies that it must be the only one being called.

## Event transformation

If we wanted to implement the asynchronous chat event, this would look like this: 
```php
class PlayerChatAsyncEvent extends AsyncEvent implements Cancellable{
	use CancellableTrait;
... Logic here
}
```
When calling: 
```php
$ev = new PlayerChatAsyncEvent($this, $messagePart, $this->server->getBroadcastChannelSubscribers(Server::BROADCAST_CHANNEL_USERS), new StandardChatFormatter());
$ev->callAsync()->onCompletion(function(PlayerChatAsyncEvent $ev) : void{
	if(!$ev->isCancelled()){
		$this->server->broadcastMessage($ev->getFormatter()->format($ev->getPlayer()->getDisplayName(), $ev->getMessage()), $ev->getRecipients());
	}
}, function() {
	// NOOP
});
```

### API changes
- Add `AsyncEvent` class
- Add `Promise::all()`

## Backwards compatibility
Not for the moment

## Follow-up
n/a

## Tests
```php
	}

	public function onPlayerChatAsyncConc1(PlayerChatEvent $event) : Promise {
		return $this->timeout("Player " . $event->getPlayer()->getName() . " chat async concurrent 1");
	}

	public function onPlayerChatAsyncConc2(PlayerChatEvent $event) : Promise {
		return $this->timeout("Player " . $event->getPlayer()->getName() . " chat async concurrent 2");
	}

	/**
	 * @exclusiveCall
	 */
	public function onPlayerChatAsyncNoConc(PlayerChatEvent $event) : Promise {
		return $this->timeout("Player " . $event->getPlayer()->getName() . " chat async exclusiveCall");
	}

	private function timeout(string $message) : Promise {
		$promise = new PromiseResolver();

		$this->getScheduler()->scheduleDelayedTask(new ClosureTask(function() use ($message, $promise) {
			$this->getLogger()->debug("Resolving promise for message " . $message);
			$promise->resolve(null);
		}), 5 * 20);

		return $promise->getPromise();
	}
}
```


src/promise/Promise.php
```diff
@@ -23,10 +23,11 @@
...
 namespace pocketmine\promise;
 
+use function count;
 use function spl_object_id;
 
 /**
- * @phpstan-template TValue
+ * @phpstan-template-covariant TValue
  */
 final class Promise{
```
SOF3, line 33: So are we now determined to turn `Promise` into a general-purpose async util class?

src/promise/Promise.php
```diff
@@ -23,10 +23,11 @@
 
 namespace pocketmine\promise;
 
+use function count;
 use function spl_object_id;
 
 /**
- * @phpstan-template TValue
+ * @phpstan-template-covariant TValue
```
SOF3: are you sure a covariant `TValue` really makes sense? it seems that `TValue` would be contravariant in `onCompletion`.

src/promise/Promise.php
```diff
@@ -52,4 +53,54 @@ public function onCompletion(\Closure $onSuccess, \Closure $onFailure) : void{
...
+	 */
+	public static function all(array $promises) : Promise {
+		/** @phpstan-var PromiseResolver<array<int, TPromiseValue>> $resolver */
+		$resolver = new PromiseResolver();
+		$values = [];
+		$toResolve = count($promises);
+		$continue = true;
+
+		foreach($promises as $key => $promise){
+			$values[$key] = null;
```
SOF3: Why do we need to initialize them as null?

src/promise/Promise.php
```diff
@@ -52,4 +53,54 @@ public function onCompletion(\Closure $onSuccess, \Closure $onFailure) : void{
...
 	}
+
+	/**
+	 * Returns a promise that will resolve only once all the Promises in
+	 * `$promises` have resolved. The resolution value of the returned promise
+	 * will be an array containing the resolution values of each Promises in
+	 * `$promises` indexed by the respective Promises' array keys.
+	 *
+	 * @template TPromiseValue
+	 * @phpstan-param Promise<TPromiseValue>[] $promises
```
SOF3: What about preserving the key here? `array<K, Promise<T>>` -> `Promise<array<K, T>>`

src/event/AsyncEventTrait.php
```diff
@@ -0,0 +1,84 @@
...
+
+namespace pocketmine\event;
+
+use pocketmine\promise\Promise;
+use pocketmine\promise\PromiseResolver;
+use pocketmine\utils\ObjectSet;
+use function array_shift;
+use function count;
+
+trait AsyncEventTrait {
```
SOF3: Consider extracting the contents of this trait into a class and just proxy the methods to the class. Adding too many implementation details in a trait is dangerous since traits are just copied code.

src/event/AsyncEventTrait.php
```diff
@@ -0,0 +1,84 @@
...
+
+	/**
+	 * @phpstan-param ObjectSet<Promise<null>>|null $promises
+	 */
+	private function initializePromises(?ObjectSet &$promises) : void{
+		$promises ??= new ObjectSet();
+		$this->promises = $promises;
+	}
+
+	public function addPromise(Promise $promise) : void{
```
SOF3: Is it an expected use case that a single handler adds multiple promises?

src/event/AsyncEventTrait.php
```diff
@@ -0,0 +1,84 @@
...
+		$globalResolver = new PromiseResolver();
+
+		$callable = function(int $priority) use ($event, $promiseSet) : Promise{
+			$handlers = HandlerListManager::global()->getListFor(static::class)->getListenersByPriority($priority);
+			$event->callHandlers($handlers);
+
+			$array = $promiseSet->toArray();
+			$promiseSet->clear();
+
+			return Promise::all($array);
```
SOF3: So multiple handlers of the same priority can execute concurrently? Would this easily lead to developers "forgetting" to check if another handler cancelled the event during execution, or multiple handlers concurrently overwriting each other's data? (could happen if there are suspension points between setter calls)

src/event/Event.php
```diff
@@ -47,15 +47,30 @@ final public function getEventName() : string{
...
+	 * @param RegisteredListener[]|null $handlers
+	 *
+	 * @internal used by AsyncEventTrait and Event
+	 */
+	final protected function callHandlers(?array $handlers) : void{
 		$timings = Timings::getEventTimings($this);
 		$timings->startTiming();
 
-		$handlers = HandlerListManager::global()->getHandlersFor(static::class);
+		$handlers = $handlers ?? HandlerListManager::global()->getHandlersFor(static::class);
```
SOF3: Why not just pass this directly?

src/promise/Promise.php
ShockedPlot7560: Without, all analystic tools will raised some error when we pass null on mixed attended type

src/event/AsyncEventTrait.php
ShockedPlot7560: Yes, this will be handled as if multiple handlers were adding promises. 

Added promises are only processed once all handlers of the priority have been called, so no difference will be made on this side.

SOF3: What if user calls `addPromise` on the event again after the initial dispatch? Your `Promise::all()` call won't include the new promises, and we cannot prevent them from doing so.

ShockedPlot7560: We should throw an exception if `addPromise` is called in an asynchronous context, to avoid this confusion.

src/promise/Promise.php
dktapps: I'd also prefer if it could be avoided. It's confusing and I feel like it's probably not necessary now that we can do `Promise<*>` in newer versions of PHPStan.

src/event/AsyncEventTrait.php
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	public function addPromise(Promise $promise) : void{
+		if(!isset($this->promises)){
+			throw new \RuntimeException("Cannot add promises, be sure to initialize the promises set in the constructor");
+		}
+		$this->promises->add($promise);
+	}
+
+	final public static function callAsync(AsyncEvent&Event $event, ObjectSet $promiseSet) : Promise{
```
dktapps: why is this static? we're in a trait

ShockedPlot7560:
## New behaviour

Above all, the logic of synchronous events remains the same, even for asynchronous events, as long as they are called synchronously.

When a listener of an asynchronous event returns a `Promise`, the event is marked as asynchronous and placed at the end of the call stack for its priority. 
When a callAsync is called, the call flow is as follows:
```
1. for each priority:
   1. call sync handlers (return type is different than Promise).
   2. call async handlers which allows concurrent execution
   3. for each async handlers which doesn't allows concurrent execution:
       1. wait for previous promises
       2. clear all the promise
       3. call the handler and add the return promise into the waiting list
   4. wait for all promises
2. the promise returned by callAsync is resolved
```

In this way, each listener can add only one promise.
If, on the other hand, the handler is sensitive to changes in the event, it is preferable to add the @noConcurrentCall tag, which specifies that it must be the only one being called.

src/event/AsyncEvent.php
```diff
@@ -34,20 +33,11 @@
...
 	/**
 	 * Be prudent, calling an event asynchronously can produce unexpected results.
 	 * During the execution of the event, the server, the player and the event context may have changed state.
 	 *
-	 * @phpstan-param ObjectSet<Promise<null>> $promiseSet
-	 *
 	 * @phpstan-return Promise<null>
 	 */
-	public static function callAsync(AsyncEvent&Event $event, ObjectSet $promiseSet) : Promise;
+	public function callAsync() : Promise;
```
SOF3: Does this mean the same event class can be called either synchronously or asynchronously? What happens if PlayerChatEvent is dispatched with `call()` in another plugin?

src/event/AsyncEventTrait.php
```diff
@@ -24,61 +24,17 @@
...
 use pocketmine\promise\Promise;
-use pocketmine\promise\PromiseResolver;
-use pocketmine\utils\ObjectSet;
-use function array_shift;
-use function count;
 
 trait AsyncEventTrait {
-	/** @phpstan-var ObjectSet<Promise<null>> */
-	private ObjectSet $promises;
+	private AsyncEventDelegate $delegate;
```
SOF3: Beware GC performance impact caused by circular references. Is it even necessary to store this field in the event object?

src/event/HandlerList.php
```diff
@@ -128,11 +130,26 @@ public function getListenerList() : array{
...
+				$syncListeners = array_filter($listeners, static function(RegisteredListener $listener) : bool{ return !($listener instanceof RegisteredAsyncListener); });
+				$listenersByPriority[$priority] = array_merge($listenersByPriority[$priority] ?? [], $syncListeners);
+
+				$asyncListeners = array_filter($listeners, static function(RegisteredListener $listener) : bool{ return $listener instanceof RegisteredAsyncListener; });
+				$asyncListenersByPriority[$priority] = array_merge($asyncListenersByPriority[$priority] ?? [], $asyncListeners);
 			}
 		}
+		foreach($asyncListenersByPriority as $priority => $asyncListeners){
+			usort($asyncListeners, static function(RegisteredAsyncListener $a, RegisteredAsyncListener $b) : int{
+				if($a->canBeCallConcurrently()){
```
SOF3: Grammar issue? Either `canCall` or `canBeCalled`, not `canBeCall`

src/event/HandlerList.php
```diff
@@ -128,11 +130,26 @@ public function getListenerList() : array{
...
+					return $b->canBeCallConcurrently() ? 0 : -1;
+				}else{
+					return $b->canBeCallConcurrently() ? -1 : 0;
+				}
```
SOF3: Can we simplify this function into `$a->canBeCallConcurrently() <=> $b->canBeCallConcurrently()`?

src/event/HandlerList.php
```diff
@@ -128,11 +130,26 @@ public function getListenerList() : array{
...
+				$listenersByPriority[$priority] = array_merge($listenersByPriority[$priority] ?? [], $syncListeners);
+
+				$asyncListeners = array_filter($listeners, static function(RegisteredListener $listener) : bool{ return $listener instanceof RegisteredAsyncListener; });
+				$asyncListenersByPriority[$priority] = array_merge($asyncListenersByPriority[$priority] ?? [], $asyncListeners);
 			}
 		}
+		foreach($asyncListenersByPriority as $priority => $asyncListeners){
+			usort($asyncListeners, static function(RegisteredAsyncListener $a, RegisteredAsyncListener $b) : int{
+				if($a->canBeCallConcurrently()){
+					return $b->canBeCallConcurrently() ? 0 : -1;
```
SOF3: Is it arbitrary that exclusive handlers are called first and concurrent handlers are called together last? Would like to see the justification in comments.

src/event/ListenerMethodTags.php
```diff
@@ -31,4 +31,5 @@ final class ListenerMethodTags{
 	public const HANDLE_CANCELLED = "handleCancelled";
 	public const NOT_HANDLER = "notHandler";
 	public const PRIORITY = "priority";
+	public const NO_CONCURRENT_CALL = "noConcurrentCall";
```
SOF3: Concurrent call is actually a less intuitive behavior. I would rather have concurrent handlers opt-in.

btw it is generally an antipattern to use negative names in any programming symbol. Prefer using `exclusiveCall` if we stick with opt-in excusive.

src/plugin/PluginManager.php
```diff
@@ -626,8 +629,29 @@ public function registerEvents(Listener $listener, Plugin $plugin) : void{
...
+				if(!is_a($eventClass, AsyncEvent::class, true)){
+					throw new PluginException(sprintf(
+						"Event handler %s() declares @%s for non-async event of type %s",
+						Utils::getNiceClosureName($handlerClosure),
+						ListenerMethodTags::NO_CONCURRENT_CALL,
+						$eventClass
+					));
+				}
+				switch(strtolower($tags[ListenerMethodTags::NO_CONCURRENT_CALL])){
+					case "true":
```
SOF3, line 655: since we have multiple of these, what about extracting to a function?

src/plugin/PluginManager.php
```diff
@@ -653,8 +677,25 @@ public function registerEvent(string $event, \Closure $handler, int $priority, P
 
 		$timings = Timings::getEventHandlerTimings($event, $handlerName, $plugin->getDescription()->getFullName());
 
-		$registeredListener = new RegisteredListener($handler, $priority, $plugin, $handleCancelled, $timings);
+		if(is_subclass_of($event, AsyncEvent::class) && $this->canHandleAsyncEvent($handler)){
```
SOF3: `registerEvent` is a low-level API that does not perform any reflection checks on `$handler`. It doesn't look right to start introducing magical behavior to `registerEvent`; that's the job of `registerEvents`.

src/plugin/PluginManager.php
```diff
@@ -653,8 +677,25 @@ public function registerEvent(string $event, \Closure $handler, int $priority, P
...
+	 * Check if the given handler return type is async-compatible (equal to Promise)
+	 *
+	 * @phpstan-template TEvent of Event
+	 * @phpstan-param (\Closure(TEvent) : void)|(\Closure(AsyncEvent&TEvent) : Promise<null>) $handler
+	 */
+	private function canHandleAsyncEvent(\Closure $handler) : bool {
+		$reflection = new \ReflectionFunction($handler);
+		$return = $reflection->getReturnType();
+
+		return $return instanceof \ReflectionNamedType && $return->getName() === Promise::class;
```
SOF3, line 751: Any thoughts on making async handlers opt-in in general, and just check the return type to double-confirm? i.e. either handler return type is `Promise`, it does not return a `Promise` object and it does not declare `@async`, or it declares `@async` and return type is `Promise`; if it has a return type hint of `Promise` or the function invocation returns a `Promise`, the server should crash.

src/plugin/PluginManager.php
```diff
@@ -626,8 +629,29 @@ public function registerEvents(Listener $listener, Plugin $plugin) : void{
...
+				switch(strtolower($tags[ListenerMethodTags::NO_CONCURRENT_CALL])){
+					case "true":
+					case "":
+						$noConcurrentCall = true;
+						break;
+					case "false":
+						break;
+					default:
+						throw new PluginException("Event handler " . Utils::getNiceClosureName($handlerClosure) . "() declares invalid @" . ListenerMethodTags::NO_CONCURRENT_CALL . " value \"" . $tags[ListenerMethodTags::NO_CONCURRENT_CALL] . "\"");
+				}
```
SOF3, line 663: Not related to this PR, but both the API and this handler would be much nicer if we use attributes instead:

```php
#[\pocketmine\event\AsyncHandler]
public function handler(XxxEvent $event) : Promise { ... }

#[\pocketmine\event\AsyncHandler(concurrent: true)]
public function handler(XxxEvent $event) : Promise { ... }
```

SOF3 (commented):
tldr AsyncEventDelegate, will look again when PR is ready

src/event/AsyncEventTrait.php
JavierLeon9966: That can be solved if `AsyncEventDelegate` stores a `WeakReference` of the class.

src/event/AsyncEvent.php
ShockedPlot7560: Currently, both are possible.
I was wondering whether an exception should be thrown when someone tries to call an async event in sync.  
As a result, if an async event is called in sync, all the handlers are called without taking into account the return types, i.e. without waiting for the promises of the others, etc...

src/event/HandlerList.php
ShockedPlot7560: I have no justification for the fact that these are the exclusive handlers after the concurrent handlers because I haven't been able to make up my mind? The only performance requirement is that the concurrent handlers be grouped together and that the exclusive handlers be called either before or after the concurrent ones.

src/event/HandlerList.php
ShockedPlot7560: Can be simplify into `($a->canBeCallConcurrently() <=> $b->canBeCallConcurrently()) * -1` yes

JavierLeon9966: That's just `$b->canBeCallConcurrently() <=> $a->canBeCallConcurrently()`.

src/event/AsyncEventTrait.php
SOF3: The point is that the circular reference was never necessary to begin with. Why store an object you don't use?

src/player/Player.php
```diff
@@ -1517,6 +1517,7 @@ public function chat(string $message) : bool{
 					if(!$ev->isCancelled()){
 						$this->server->broadcastMessage($ev->getFormatter()->format($ev->getPlayer()->getDisplayName(), $ev->getMessage()), $ev->getRecipients());
 					}
+
```
JavierLeon9966:
```suggestion
```

src/event/AsyncEventTrait.php
JavierLeon9966: Right. That class can just be `AsyncEventTrait`.

JavierLeon9966 (changes requested):

src/event/AsyncEventTrait.php
JavierLeon9966: Right. That class can just be `AsyncEventTrait`.

ShockedPlot7560: > The point is that the circular reference was never necessary to begin with. Why store an object you don't use?

I agree, the class property is not necessary

src/event/Event.php
```diff
@@ -47,6 +47,21 @@ final public function getEventName() : string{
...
 	 */
 	public function call() : void{
+		$this->callDepth(function(){
+			$handlers = HandlerListManager::global()->getHandlersFor(static::class);
+
+			foreach($handlers as $registration){
+				$registration->callEvent($this);
+			}
+		});
+	}
```
dktapps: This code is very performance sensitive. I don't like introducing extra function calls and closures into here.

dktapps (commented):
It seems to me like a lot of complexity went into making async events and their handlers use the same frameworks that sync events do. The net effect of this is that async events are callable either synchronously or asynchronously, which seems to me like a design flaw.

ShockedPlot7560:
> It seems to me like a lot of complexity went into making async events and their handlers use the same frameworks that sync events do. The net effect of this is that async events are callable either synchronously or asynchronously, which seems to me like a design flaw.

We can prohibit the synchronous call of a specified asynchronous event, but exiting the event will require a BC break every time we want to implement it in an event. As I said [here](https://github.com/pmmp/PocketMine-MP/pull/6092#discussion_r1367932362) synchronous and asynchronous calling is a subject for discussion, however, it's still necessary to keep synchronous handlers in the middle of synchronous ones, as not everyone needs to return a promise every time.

SOF3:
I think what we need here is an adapter at the registerEvents level to convert a void function into a promise closure. imo we are fine with breaking BC for registerEvent on PlayerChatEvent as long as we don't break registerEvents, and it is probably fine to make PlayerChatEvent no longer extend Event.

src/event/AsyncEvent.php
ShockedPlot7560: It is no longer possible to call an asynchronous event synchronously.

src/plugin/PluginManager.php
ShockedPlot7560: This is not directly related to this PR, it should be done externally.

src/event/HandlerList.php
```diff
@@ -129,24 +128,23 @@ public function getListenerList() : array{
...
 			}
 		}
-		foreach($asyncListenersByPriority as $priority => $asyncListeners){
-			usort($asyncListeners, static function(RegisteredAsyncListener $a, RegisteredAsyncListener $b) : int{
-				// concurrent listeners are sorted to the end of the list
-				return $b->canBeCalledConcurrently() <=> $a->canBeCalledConcurrently();
-			});
-			$listenersByPriority[$priority] = array_merge($listenersByPriority[$priority] ?? [], $asyncListeners);
-		}
+		$listenersByPriority = array_merge_recursive($listeners, $asyncListeners, $exclusiveAsyncListeners);
```
JavierLeon9966: Aren't exclusive async listeners supposed to be run before normal async listeners?

ShockedPlot7560: That's the debate going on in my neurons. I don't have any pros or cons, really.

src/event/Event.php
```diff
@@ -51,6 +51,9 @@ public function call() : void{
 			//this exception will be caught by the parent event call if all else fails
 			throw new \RuntimeException("Recursive event call detected (reached max depth of " . self::MAX_EVENT_CALL_DEPTH . " calls)");
 		}
+		if($this instanceof AsyncEvent){
+			throw new \InvalidArgumentException("Cannot call async event synchronously, use callAsync() instead");
+		}
```
dktapps: Again, we're still sharing architecture here. This function should simply not exist for an async event.

dktapps:
`Promise::all()` should be PR'd separately, as it's useful by itself and would reduce PR diff noise

src/plugin/PluginManager.php
ShockedPlot7560: As async events are now completely detached from synchronous events, there is no longer any need for double confirmation, as each event handler will behave in the same way.

ShockedPlot7560:
## New behavior (again)

As async event is BC breaking by default, I detached the async event from the synchronous ones. Now, an async event should be handle with an async handler : must return a Promise. It can have a tag to specify if he need to be an exclusive handler or not. 
Exclusive handlers will be called one by one, so no other handler can modify the event during the execution. Otherwise, concurrent handler will be call in parallel, priority by priority.

src/event/HandlerListManager.php
```diff
@@ -112,7 +113,7 @@ public function getListFor(string $event) : HandlerList{
 	}
 
 	/**
-	 * @phpstan-param class-string<covariant Event> $event
+	 * @phpstan-param class-string<Event|AsyncEvent> $event
```
dktapps: Shouldn't this be templated like getListFor()?

src/event/RegisteredAsyncListener.php
```diff
@@ -0,0 +1,67 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\event;
+
+use pocketmine\plugin\Plugin;
+use pocketmine\promise\Promise;
+use pocketmine\timings\TimingsHandler;
+
+class RegisteredAsyncListener extends RegisteredListener{
```
dktapps: I don't think this should inherit from RegisteredListener. There's no commonality aside from the shared properties.

dktapps (dismissed):

dktapps:
So, this is pretty much how I envisioned it should be. I think it lost me somewhere because this whole concept is just a massive headache.

To summarize what I think I'm seeing:

# Promise rejection
If an async handler at any priority rejects the promise it returned, no handlers from subsequent priorities will be called.
If an exclusive async handler rejects, no subsequent exclusive handlers will be called, and no handlers from subsequent priorities will be called either.

This is a "hard cancellation" with no ability to uncancel. Unsure if this is desirable, or how we might change this in the future if the need arose. Please discuss.

# Concurrent vs exclusive handlers
There are 2 types of async handler: concurrent (all concurrent priorities with the same priority will be run together and then their promises collectively awaited), and exclusive (each promise must resolve before the next exclusive handler is called).

**Use case explanations & documentation needed for this.**

# Separated stack for sync vs async events
Async events and sync events are fully separated. Sync events cannot have async handlers and vice versa. I personally requested this.

This avoids impact to existing events, as they otherwise would ALL need to be post-processed asynchronously (which would require massive refactors & break any plugin that calls events of its own), or the functionality would need to be enabled for select events (which causes weird problems of its own).

This also avoids breaking existing sync handler code, which weren't designed for the possibility that 2 events of the same type might be in an incomplete state at the same time. This means they aren't built to expect that, for example `LOWEST` priority handlers could be called 2 times with a certain event before `HIGHEST` priority handlers were called at all.

**However**, I think it makes sense to allow async events to have sync handlers. We just need to be aware that switching an event from sync to async or vice versa is always a BC break in one form or another.

# Caller of events
Callers of events can simply put an `onCompletion` callback pair onto the promise returned by `call()`. This is nice and simple.

# Event handlers
Async event handlers must return `Promise`. This is different from sync events, whose return types are not enforced, but which are assumed to return `void`.

Otherwise, async handlers look basically the same as sync handlers.

src/event/HandlerList.php
dktapps: The defining feature of exclusive listeners is that no other listener can be invoked before the exclusive listener's promise is resolved. The ordering isn't really significant.

src/plugin/PluginManager.php
```diff
@@ -665,4 +693,44 @@ public function registerEvent(string $event, \Closure $handler, int $priority, P
...
+	 *
+	 * @phpstan-param \Closure(AsyncEvent) : Promise<null> $handler
+	 *
+	 * @throws \ReflectionException
+	 */
+	private function canHandleAsyncEvent(\Closure $handler) : bool{
+		$reflection = new \ReflectionFunction($handler);
+		$return = $reflection->getReturnType();
+
+		return $return instanceof \ReflectionNamedType && $return->getName() === Promise::class;
```
dktapps, line 751: This probably needs to check for `allowsNull()`. Not sure what should be done about that.

dktapps: In fact I don't think we actually need to check this at all. Sync handlers on async events should be fine, it's only the inverse that's a problem.

src/event/HandlerList.php
dktapps: Looks like this function isn't actually used for async events anyway.

src/event/AsyncEvent.php
```diff
@@ -0,0 +1,126 @@
...
+/**
+ * This class is used to permit asynchronous event handling.
+ *
+ * When an event is called asynchronously, the event handlers are called by priority level.
+ * When all the promises of a priority level have been resolved, the next priority level is called.
+ */
+abstract class AsyncEvent{
+	/** @var array<class-string<AsyncEvent>, int> $delegatesCallDepth */
+	private static array $delegatesCallDepth = [];
+	private const MAX_EVENT_CALL_DEPTH = 50;
```
SOF3: What if I dispatch 50 AsyncPlayerLoginEvents within 1 second when server starts, and the handlers take more than 1 second to resolve? The server crashes immediately?

dktapps: I discussed this with Plot on Discord yesterday. I don't think even he knows how this code was supposed to work.

src/event/AsyncEvent.php
```diff
@@ -0,0 +1,126 @@
...
+			if($handler->canBeCalledConcurrently()){
+				unset($handlers[$k]);
+				$promise = $handler->callAsync($this);
+				if($promise !== null){
+					$awaitPromises[] = $promise;
+				}
+			}else{
+				if(count($awaitPromises) > 0){
+					//wait for concurrent promises to complete
+					break;
```
SOF3, line 109: Should this be `continue` instead of `break` to process other concurrent handlers at the same priority? Or do we sort handlers by `(priority, isExclusive)`?

src/event/AsyncEvent.php
```diff
@@ -0,0 +1,126 @@
...
+
+	/**
+	 * @param AsyncRegisteredListener[] $handlers
+	 * @phpstan-param \Closure() : void $resolve
+	 * @phpstan-param \Closure() : void $reject
+	 */
+	private function processRemainingHandlers(array $handlers, \Closure $resolve, \Closure $reject) : void{
+		$currentPriority = null;
+		$awaitPromises = [];
+		foreach($handlers as $k => $handler){
```
SOF3, line 77: This function is a bit verbose. Can we do this better, e.g. treating exclusive handlres as their own batches and letting handler list return a list of batches to execute?

src/event/AsyncEvent.php
SOF3: Oh I see it's sorted indeed.

src/event/AsyncHandlerListManager.php
```diff
@@ -0,0 +1,61 @@
...
+	 * @phpstan-return array<int, AsyncRegisteredListener>
+	 */
+	private static function sortSamePriorityHandlers(array $listeners) : array{
+		uasort($listeners, function(AsyncRegisteredListener $left, AsyncRegisteredListener $right) : int{
+			//While the system can handle these in any order, it's better for latency if concurrent handlers
+			//are processed together. It doesn't matter whether they are processed before or after exclusive handlers.
+			if($right->canBeCalledConcurrently()){
+				return $left->canBeCalledConcurrently() ? 0 : 1;
+			}
+			return -1;
```
SOF3: But you end up defining an order here right? Ends up `$left > $right` if `$left` is exclusive and `$right` is concurrent.

Also why do we return `-1` when both handlers are exclusive? Shouldn't it return `0` instead?

```suggestion
			//are processed together. Exclusive handlers are sorted before concurrent handlers.
			$leftConcurrent = $left->canBeCalledConcurrently();
			$rightConcurrent = $right->canBeCalledConcurrently();
			if($leftConcurrent !== $rightConcurrent){
				return $leftConcurrent <=> $rightConcurrent;
			}
			return 0;
```

src/event/AsyncEvent.php
dktapps: Linear cache bypasses the handler list entirely in most cases for performance reasons, same as for sync events. If we start doing what you suggest it will just require even more code.

src/event/AsyncHandlerListManager.php
dktapps: Yes, the order is intended to group concurrent handlers together, whether at the start or the end.

Good catch on the sameness. I knew there was something I didn't like about this code but I couldn't put my finger on it.

dktapps: I don't like specifying that exclusive handlers should come first without reasoning. There is no reason for them to come first. What we actually want is to aggregate concurrent handlers for Promise::all() to be efficiently used.

src/event/AsyncEvent.php
dktapps: Functionally it's not actually important anyway. The sorting only affects the number of times Promise::all() will be called.

src/event/AsyncEvent.php
```diff
@@ -0,0 +1,126 @@
...
+		$timings = Timings::getAsyncEventTimings($this);
+		$timings->startTiming();
+
+		++self::$delegatesCallDepth[$class];
+		try{
+			/** @phpstan-var PromiseResolver<static> $globalResolver */
+			$globalResolver = new PromiseResolver();
+
+			$handlers = AsyncHandlerListManager::global()->getHandlersFor(static::class);
+			if(count($handlers) > 0){
```
ShockedPlot7560, line 56: This check can be avoided because it will be checked in the processRemainingHandlers function and solved directly?

ShockedPlot7560:
As I understand, now, we will accept async listeners to return null or Promise ?

src/event/AsyncEvent.php
dktapps: It's a micro optimization to avoid unnecessary allocations

SenseiTarzan:
I think the event system in asynchronous will never work properly because it was not designed for it. At the time, we couldn't do asynchronous operations, and because of that, the entire pmmp codebase would need to be reworked to integrate the asynchronous part into pmmp's event loop for proper handling.
Because we can perform synchronous operations within asynchronous code, but not the reverse, especially since managing cancellations would not be possible in a very clean way.
You have to do it before PM6

dktapps:
> As I understand, now, we will accept async listeners to return null or Promise ?

Yep, they'll allow the next handler in the queue to be called immediately if returning null

dktapps:
I think the main thing still on the TODO list is to buffer incomplete events. The existing mechanism using stack depth won't cut it. We need a way to track which handlers didn't fulfil their promises so that the correct plugin can be properly blamed when an event fails to complete.

src/event/AsyncEvent.php
ShockedPlot7560: Probably TODO now, but we probably should track where all handlers stuck and throw exception if too many handlers stuck at one handler for too long? 
As said on discord, this can produce some crash if a server handler like 100 connections at a time and the login sequence is transformed into async process.

ShockedPlot7560:
To be more precise with the depth limiter. We could probably reset it after each concurrent management of priority handlers. There's less chance of a plugin making a recursion call with different priority handlers.
I'm not saying we'll cover all cases because that's impossible, but we can limit errors and that's necessary to protect the integrity of the server at least.

ShockedPlot7560:
I said nothing, mind is blowing

ShockedPlot7560:
A question remains about the global asynchronous events timing. Should we keep it as it will result in an inconsistent result as it profiles when we call the event until an asynchronous handler is encountered? The handler call is always profiled until a promise is returned, perhaps, like asyncTask, we need to profile the handler time in general? 
Open to opinion.

ShockedPlot7560:
I modified the system to detect handlers blocked at a point. 
Now, each time a handler is called, a personal counter is increased by 1 until its execution is complete. This ensures that there are not too many handlers blocked at a point, which is why the value has been doubled. 
It won't stop the server crashing when 1000 players connect at the same time when a plugin has added a long task behind it. 
Perhaps an extra function in the AsyncEvent so that you can give your personal limit? Perhaps a security issue? But for me, it's OK as it is here.

src/event/AsyncEvent.php
```diff
@@ -0,0 +1,134 @@
...
+		$awaitPromises = [];
+		foreach($handlers as $k => $handler){
+			$priority = $handler->getPriority();
+			if(count($awaitPromises) > 0 && $currentPriority !== null && $currentPriority !== $priority){
+				//wait for concurrent promises from previous priority to complete
+				break;
+			}
+
+			$currentPriority = $priority;
+			$handlerId = spl_object_id($handler) << 3 | $priority;
```
dktapps: Why is the priority included here? This seems risky and will silently break if arbitrary priority numbers are added in the future.

We shouldn't need priority to correctly blame a handler for not fulfilling event calls, should we?

ShockedPlot7560: The handler registry ensures that there are no two identical instances of the same handler in the same priority. However, between different priorities, we may encounter the same handler. Should we consider handlers to be identical even if their priority is different?

dktapps: That's impossible, their `RegisteredListener` instances would be different even if the same `Closure` was reused

PR #6106, opened 2023-10-21
Implement Elytra (old)

pandaaaBE:
## Introduction
Implement Elytra.

_There is a plugin, why should we accept your pull request?
Well, since IvanCrafter is currently working on a Fireworks PR it would be only fitting to finally add the Elytra as well - it's tedious to search out all the vanilla features on poggit._

## Todo
- [x] Add Repairing via Crafting Recipes
- [x] Disable when touching ground

## Tests
[Video](https://streamable.com/iefpxd)

## Facts
This is my first pull request so please don't laugh at me, I will of course add any features I find at minecraft.wiki.

### DIFF
--- a/src/item/Elytra.php
+++ b/src/item/Elytra.php
@@ -1,0 +1,50 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * @author PocketMine Team
+ * @link http://www.pocketmine.net/
+ *
+ *
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\item;
+
+use pocketmine\entity\Human;
+use pocketmine\entity\Living;
+
+class Elytra extends Armor{
+
+	public const MINIMUM_PITCH = -59;
+	public const MAXIMUM_PITCH = 38;
+
+	public function onTickWorn(Living $entity) : bool{
+		if($entity instanceof Human){
+			if(!$entity->isGliding()){
+				$location = $entity->getLocation();
+				if($location->pitch >= self::MINIMUM_PITCH and $location->pitch <= self::MAXIMUM_PITCH){
+					$entity->resetFallDistance();
+					return true;
+				}
+			}else{
+				$this->applyDamage(1);
+				return true;
+			}
+		}
+
+		return false;
+	}
+}
--- a/src/item/Elytra.php
+++ b/src/item/Elytra.php
@@ -38,12 +38,14 @@ class Elytra extends Armor{
 			// Check if the player's pitch is within the allowed range between -59 and 38.
 			if ($pitch >= -59 && $pitch <= 38) {
 				$entity->resetFallDistance();
+				return true;
 			}
 
 			// If the player is on the ground, reset the fall distance and disable gliding.
 			if ($entity->isOnGround()) {
 				$entity->resetFallDistance();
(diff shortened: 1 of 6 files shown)


pandaaaBE:
1. Is there something better then "onTickWorn"?
2. Is the ElytraLoopSound class relevant or is this useless?

pandaaaBE:
Should I add repairing via crafting table for elytras? I can not find a method for this

src/item/Elytra.php
```diff
@@ -0,0 +1,56 @@
...
+
+			// Check if the player's pitch is within the allowed range between -59 and 38.
+			if ($pitch >= -59 && $pitch <= 38) {
+				$entity->resetFallDistance();
+				return true;
+			}
+
+			// If the player is on the ground, reset the fall distance and disable gliding.
+			if ($entity->isOnGround()) {
+				$entity->resetFallDistance();
```
JavierLeon9966, line 46: Why is this here? Don't we already reset fall damage above?

src/item/Elytra.php
```diff
@@ -0,0 +1,56 @@
...
+
+	public function onTickWorn(Living $entity): bool {
+		if ($entity instanceof Human && $entity->isGliding()) {
+			$this->applyDamage(1);
+
+			$location = $entity->getLocation();
+			$pitch = $location->pitch;
+
+			// Check if the player's pitch is within the allowed range between -59 and 38.
+			if ($pitch >= -59 && $pitch <= 38) {
```
JavierLeon9966, line 39: I think this should be constants or use a calculation to check if it'll reset fall damage.

src/item/VanillaItems.php
```diff
@@ -626,6 +627,7 @@ private static function registerArmorItems() : void{
 		self::register("iron_chestplate", new Armor(new IID(Ids::IRON_CHESTPLATE), "Iron Chestplate", new ArmorTypeInfo(6, 241, ArmorInventory::SLOT_CHEST, material: ArmorMaterials::IRON()), [EnchantmentTags::CHESTPLATE]));
 		self::register("leather_tunic", new Armor(new IID(Ids::LEATHER_TUNIC), "Leather Tunic", new ArmorTypeInfo(3, 81, ArmorInventory::SLOT_CHEST, material: ArmorMaterials::LEATHER()), [EnchantmentTags::CHESTPLATE]));
 		self::register("netherite_chestplate", new Armor(new IID(Ids::NETHERITE_CHESTPLATE), "Netherite Chestplate", new ArmorTypeInfo(8, 593, ArmorInventory::SLOT_CHEST, 3, true, material: ArmorMaterials::NETHERITE()), [EnchantmentTags::CHESTPLATE]));
+		self::register("elytra", new Elytra(new IID(Ids::ELYTRA), "Elytra", new ArmorTypeInfo(0, 433, ArmorInventory::SLOT_CHEST, 0, false, null)));
```
JavierLeon9966, line 630: You can skip the optional parameters and the elytra class must override the `onBroken()` with an empty one to prevent it from breaking.

src/world/sound/ElytraLoopSound.php
```diff
@@ -0,0 +1,36 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\world\sound;
+
+use pocketmine\math\Vector3;
+use pocketmine\network\mcpe\protocol\LevelEventPacket;
+use pocketmine\network\mcpe\protocol\types\LevelEvent;
+
+class ElytraLoopSound implements Sound{
```
JavierLeon9966, line 30: Isn't this sound client side?

JavierLeon9966 (changes requested):
This needs to be in the minor-next branch

JavierLeon9966:
> Should I add repairing via crafting table for elytras? I can not find a method for this

That can be skipped. Dynamic crafting hasn't been added yet.

src/item/Elytra.php
```diff
@@ -0,0 +1,56 @@
...
+namespace pocketmine\item;
+
+use pocketmine\entity\Human;
+use pocketmine\entity\Living;
+
+class Elytra extends Armor{
+
+	public function onTickWorn(Living $entity): bool {
+		if ($entity instanceof Human && $entity->isGliding()) {
+			$this->applyDamage(1);
```
JavierLeon9966, line 33: Damage should only be applied every second.

src/world/sound/ElytraLoopSound.php
pandaaaBE: I think so, this is why I was asking.

Will remove the class

src/item/Elytra.php
pandaaaBE: How do you mean?

JavierLeon9966 (changes requested):

src/item/Elytra.php
JavierLeon9966: You shouldn't use hardcoded values.

PR #6109, opened 2023-10-22
Implement Elytra

pandaaaBE:
## Introduction
Implement **player-focused** Elytra.

(Waiting until someone adds this: https://github.com/pmmp/PocketMine-MP/issues/6124)

_There is a plugin, why should we accept your pull request?
Well, since IvanCrafter is currently working on a Fireworks PR it would be only fitting to finally add the Elytra as well - it's tedious to search out all the vanilla features on poggit._

## Todo
- [x] Add to damage event (kinetic energy damage)

## Tests
[Video](https://streamable.com/iefpxd)

### DIFF
--- a/src/item/Elytra.php
+++ b/src/item/Elytra.php
@@ -1,0 +1,71 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * @author PocketMine Team
+ * @link http://www.pocketmine.net/
+ *
+ *
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\item;
+
+use pocketmine\entity\Human;
+use pocketmine\entity\Living;
+
+class Elytra extends Armor{
+
+	//Check if the player's pitch is within the allowed range between -59 and 38.
+	private const MINIMUM_PITCH = -59;
+	private const MAXIMUM_PITCH = 38;
+
+	private int $applyDamageTime = 61;
+
+	public function onTickWorn(Living $entity): bool {
+		if ($entity instanceof Human && $entity->isGliding()) {
+
+			$this->applyDamageTime--;
+			if($this->applyDamageTime <= 0){
+				$this->applyDamage(1);
+				$this->applyDamageTime = 61;
+			}
+
+			$location = $entity->getLocation();
+			$pitch = $location->pitch;
+
+			// Check if the player's pitch is within the allowed range between -59 and 38.
+			if ($pitch >= self::MINIMUM_PITCH && $pitch <= self::MAXIMUM_PITCH) {
+				$entity->resetFallDistance();
+				return true;
+			}
+
+			// If the player is on the ground, reset the fall distance and disable gliding.
+			if ($entity->isOnGround()) {
+				$entity->resetFallDistance();
+				$entity->setGliding(false);
+				return true;
+			}
+
+			return true;
+		}
+
+		return false;
+	}
+
+	protected function onBroken() : void{
+		//NOOP
+	}
+}
--- a/src/item/Elytra.php
+++ b/src/item/Elytra.php
@@ -32,7 +32,8 @@ class Elytra extends Armor{
 	private const MINIMUM_PITCH = -59;
 	private const MAXIMUM_PITCH = 38;
 
-	private int $applyDamageTime = 61;
+	const APPLY_DAMAGE_INTERVAL = 61;
+	private int $applyDamageTime = self::APPLY_DAMAGE_INTERVAL;
 
 	public function onTickWorn(Living $entity) : bool {
 		if ($entity instanceof Human && $entity->isGliding()) {
@@ -40,12 +41,10 @@ class Elytra extends Armor{
 			$this->applyDamageTime--;
 			if($this->applyDamageTime <= 0){
 				$this->applyDamage(1);
-				$this->applyDamageTime = 61;
+				$this->applyDamageTime = self::APPLY_DAMAGE_INTERVAL;
 			}
 
-			$location = $entity->getLocation();
-			$pitch = $location->pitch;
-
+			$pitch = $entity->getLocation()->pitch;
 			// Check if the player's pitch is within the allowed range between -59 and 38.
 			if ($pitch >= self::MINIMUM_PITCH && $pitch <= self::MAXIMUM_PITCH) {
 				$entity->resetFallDistance();
--- a/src/item/Elytra.php
+++ b/src/item/Elytra.php
@@ -34,7 +34,7 @@ class Elytra extends Armor{
 
 	private int $applyDamageTime = 61;
 
-	public function onTickWorn(Living $entity): bool {
+	public function onTickWorn(Living $entity) : bool {
 		if ($entity instanceof Human && $entity->isGliding()) {
 
 			$this->applyDamageTime--;
--- a/src/item/Elytra.php
+++ b/src/item/Elytra.php
@@ -28,11 +28,10 @@ use pocketmine\entity\Living;
 
 class Elytra extends Armor{
 
-	//Check if the player's pitch is within the allowed range between -59 and 38.
 	private const MINIMUM_PITCH = -59;
 	private const MAXIMUM_PITCH = 38;
 
-	const APPLY_DAMAGE_INTERVAL = 61;
+	private const APPLY_DAMAGE_INTERVAL = 21;
 	private int $applyDamageTime = self::APPLY_DAMAGE_INTERVAL;
 
 	public function onTickWorn(Living $entity) : bool {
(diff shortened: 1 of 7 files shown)


src/item/Elytra.php
```diff
@@ -0,0 +1,71 @@
...
+
+	private int $applyDamageTime = 61;
+
+	public function onTickWorn(Living $entity) : bool {
+		if ($entity instanceof Human && $entity->isGliding()) {
+
+			$this->applyDamageTime--;
+			if($this->applyDamageTime <= 0){
+				$this->applyDamage(1);
+				$this->applyDamageTime = 61;
```
ShockedPlot7560: Where does this hard value of 61 come from?

src/item/Elytra.php
```diff
@@ -0,0 +1,71 @@
...
+			}
+
+			$location = $entity->getLocation();
```
ShockedPlot7560: This variable is useless

src/item/Elytra.php
pandaaaBE: A pair of elytra has 432 durability, allowing 7 minutes and 12 seconds of gliding time - in onTickWorn this is something like a "counter", if u glide 60 secounds then it will apply 1 damage

JavierLeon9966 (changes requested):
Another thing is that elytras can give you kinetic energy damage if you go too fast against a wall or roof.

pandaaaBE:
> Another thing is that elytras can give you kinetic energy damage if you go too fast against a wall or roof.

the rest is okay?

src/item/Elytra.php
```diff
@@ -0,0 +1,71 @@
...
+
+use pocketmine\entity\Human;
+use pocketmine\entity\Living;
+
+class Elytra extends Armor{
+
+	private const MINIMUM_PITCH = -59;
+	private const MAXIMUM_PITCH = 38;
+
+	private const APPLY_DAMAGE_INTERVAL = 21;
```
JavierLeon9966: Shouldn't this be `20`?

src/item/Elytra.php
```diff
@@ -0,0 +1,71 @@
...
+			if($this->applyDamageTime <= 0){
+				$this->applyDamage(1);
+				$this->applyDamageTime = self::APPLY_DAMAGE_INTERVAL;
+			}
+
+			$pitch = $entity->getLocation()->pitch;
+			// Check if the player's pitch is within the allowed range between -59 and 38.
+			if ($pitch >= self::MINIMUM_PITCH && $pitch <= self::MAXIMUM_PITCH) {
+				$entity->resetFallDistance();
+				$entity->setFallDistance(0.0);
```
JavierLeon9966: This is redundant. `resetFallDistance()` does the same thing.

src/item/Elytra.php
```diff
@@ -0,0 +1,71 @@
...
+				return true;
+			}
+
+			// If the player is on the ground, reset the fall distance and disable gliding.
+			if ($entity->isOnGround()) {
+				$entity->setGliding(false);
+				$entity->resetFallDistance();
+				$entity->setFallDistance(0.0);
```
JavierLeon9966: Same here.

src/item/Elytra.php
```diff
@@ -0,0 +1,71 @@
...
+
+			$this->applyDamageTime--;
+			if($this->applyDamageTime <= 0){
+				$this->applyDamage(1);
+				$this->applyDamageTime = self::APPLY_DAMAGE_INTERVAL;
+			}
+
+			$pitch = $entity->getLocation()->pitch;
+			// Check if the player's pitch is within the allowed range between -59 and 38.
+			if ($pitch >= self::MINIMUM_PITCH && $pitch <= self::MAXIMUM_PITCH) {
```
JavierLeon9966: This should be calculated instead of checking the pitch. You can probably find the calculation in the Java source code ot it's probably if the Y velocity is greater than the XZ velocity.

JavierLeon9966 (changes requested):

JavierLeon9966:
> > Another thing is that elytras can give you kinetic energy damage if you go too fast against a wall or roof.
> 
> the rest is okay?

A few more things.

src/item/Elytra.php
pandaaaBE: Tested it with resetFallDistance and I always get damage when gliding to the ground, in combo with setFallDistance I don‘t get fall damage..

What is the reason?

src/item/Elytra.php
pandaaaBE: > $velocity = $entity->getMotion();
> $yVelocity = $velocity->y;
> $xzVelocity = sqrt($velocity->x ** 2 + $velocity->z ** 2);
> 
> if ($yVelocity > $xzVelocity) {
> }

Like this? 🤔

JavierLeon9966: Based on tests fall damage is reset if the vertical velocity is less than 0.5 and those angles are what make it less than that. You can't rely on motion for the velocity because players always have 0 motion server-side.

src/item/Elytra.php
```diff
@@ -0,0 +1,71 @@
...
+class Elytra extends Armor{
+
+	private const MINIMUM_PITCH = -59;
+	private const MAXIMUM_PITCH = 38;
+
+	private const APPLY_DAMAGE_INTERVAL = 21;
+	private int $applyDamageTime = self::APPLY_DAMAGE_INTERVAL;
+
+	public function onTickWorn(Living $entity) : bool {
+		if ($entity instanceof Human && $entity->isGliding()) {
```
JavierLeon9966: This can extend to `Living` entities.

JavierLeon9966 (changes requested):

src/item/Elytra.php
JavierLeon9966: The reason is probably unrelated to this. I don't see how adding redundancy could fix that.

dktapps:
As with all previous PRs on the same subject (there have been many), this just relies on client-side movement, which means that elytra are useless for server-side mobs. It's not possible to, for example, make a gliding NPC.

pandaaaBE:
> > > Another thing is that elytras can give you kinetic energy damage if you go too fast against a wall or roof.
> > 
> > 
> > the rest is okay?
> 
> A few more things.

Is there something like a sample I can use to find out how I can "register" something in EntityDamageEvent?

pandaaaBE:
> Another thing is that elytras can give you kinetic energy damage if you go too fast against a wall or roof.

image

pandaaaBE:
Okay.. maybe kinetic energy is to strong currently.. 

https://streamable.com/pqx40j

JavierLeon9966:
> > > > Another thing is that elytras can give you kinetic energy damage if you go too fast against a wall or roof.
> > > 
> > > 
> > > the rest is okay?
> > 
> > 
> > A few more things.
> 
> Is there something like a sample I can use to find out how I can "register" something in EntityDamageEvent?

Just add a `CAUSE_FLY_INTO_WALL` constant in `EntityDamageEvent`.

src/item/Elytra.php
```diff
@@ -0,0 +1,85 @@
...
+	private const APPLY_DAMAGE_INTERVAL = 20;
+	private int $applyDamageTime = self::APPLY_DAMAGE_INTERVAL;
+
+	public function onTickWorn(Living $entity) : bool {
+		if ($entity->isGliding()) {
+
+			$direction = $entity->getDirectionVector();
+			$yVel = $direction->y;
+
+			if($yVel < 0.5){
```
JavierLeon9966: That isn't how you get the velocity. It's supposed to be `currentY - prevY` but there needs to be a way to get the previous position outside of the `Entity` class

src/item/Elytra.php
```diff
@@ -0,0 +1,85 @@
...
+
+			// If the player is on the ground, reset the fall distance and disable gliding.
+			if ($entity->isOnGround()) {
+				$entity->resetFallDistance();
+				$entity->setGliding(false);
+				return true;
+			}
+
+			//Calculate kinetic energy.
+			$xzVector = new Vector2($direction->x, $direction->z);
```
JavierLeon9966: Same goes here.

JavierLeon9966 (changes requested):

src/item/Elytra.php
pandaaaBE: > It's supposed to be currentY - prevY

Yes, I know but there is no method for it - at least nothing found... Is there anything outside the entity class or you didn´t know?

I could log the previous Position but idk if this is a great idea

pandaaaBE: private $previousPosition = null;

    public function onTickWorn(Living $entity): bool {
        $currentPosition = $entity->getLocation();

        if ($this->previousPosition === null) {
            // Store the current position as the previous position in the first tick
            $this->previousPosition = $currentPosition;
        }

        $deltaY = $currentPosition->getY() - $this->previousPosition->getY();
        $deltaXY = $currentPosition->distance($this->previousPosition);

        // Calculate vertical and horizontal velocities
        $yVelocity = $deltaY;
        $xyVelocity = $deltaXY;

        if ($yVelocity > $xyVelocity) {
        }

        // Update the previous position for the next tick
        $this->previousPosition = $currentPosition;

        return false;
    }

JavierLeon9966: Just make an issue about it and wait until that method gets added or make a PR that adds it.

pandaaaBE: > ue about it and wait until that method gets added or make a PR that adds it.

would it be bad if i first add myself what i sent above and later add the new method?

src/item/Elytra.php
pandaaaBE: can i add the function like in the example above my self or should i wait for a other build-in method?

pandaaaBE (commented):
Needs to be tested, would this be right with the velocity?

@JavierLeon9966

JavierLeon9966:
> Needs to be tested, would this be right with the velocity?
> 
> @JavierLeon9966

No, I don't think you're supposed to do that with items. Just make an issue about needing a way to get an entity's previous position or position Delta and Dylan will handle it.

pandaaaBE:
> > Needs to be tested, would this be right with the velocity?
> > @JavierLeon9966
> 
> No, I don't think you're supposed to do that with items. Just make an issue about needing a way to get an entity's previous position or position Delta and Dylan will handle it.

@JavierLeon9966 - now that Dylan is gone, can we introduce this by getting the previous position in the Elytra class, or do you have an idea how I could introduce this in the code

ShockedPlot7560:
> do you have an idea how I could introduce this in the code

This would be a question to be asked in the issue no ? 
Of course Dylan will be gone, but that doesn't mean we have to start issuing several features in one PR.

PR #6111, opened 2023-10-22
Create LitTrait and remove repetitive code

DebHugo:
## Introduction
Create `LitTrait` for remove repetitive code

### API changes
Create `LitTrait` trait


SOF3:
If even a getter/setter is repetitive and it is public anyway, why do we even have a getter/setter for this instead of exposing the field directly?

DebHugo:
> If even a getter/setter is repetitive and it is public anyway, why do we even have a getter/setter for this instead of exposing the field directly?

Currently, the lit value is protected. Why is it not public? I don't know.
I am making this PR to continue the logic in pmmp. Indeed, there is `AnyFacingTrait` which has the same utility as `LitTrait` which I created.

src/block/RedstoneTorch.php
```diff
@@ -23,28 +23,17 @@
 
 namespace pocketmine\block;
 
+use pocketmine\block\utils\LitTrait;
 use pocketmine\data\runtime\RuntimeDataDescriber;
 
 class RedstoneTorch extends Torch{
-	protected bool $lit = true;
+	use LitTrait;
```
dktapps: This one is problematic. Its default state should be lit=true.

dktapps:
> If even a getter/setter is repetitive and it is public anyway, why do we even have a getter/setter for this instead of exposing the field directly?

The setter is fluent, and the convention is to use functions for this for validation and such. There is an argument to be made for making the fields public directly (it would simplify implementation of the NBT blockstate serializers), but right now it's not practical and would make the API inconsistent.

ShockedPlot7560 (dismissed):

src/block/utils/LitTrait.php
```diff
@@ -0,0 +1,46 @@
...
+ *
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\block\utils;
+
+use pocketmine\data\runtime\RuntimeDataDescriber;
+
+trait LitTrait{
```
dktapps: `LightableTrait` would be more in line with other similar traits.
Welcome to suggest a better name (I don't like my suggestion much) but I don't think `LitTrait` is quite accurate.

dktapps (approved):

ShockedPlot7560 (approved):

PR #6112, opened 2023-10-23
Optimized combining of listeners

JavierLeon9966:
## Introduction

Helping the #6092 PR.

### DIFF
--- a/src/event/HandlerList.php
+++ b/src/event/HandlerList.php
@@ -130,26 +130,24 @@ class HandlerList{
 		}
 
 		$listenersByPriority = [];
+		$concurrentAsyncListenersByPriority = [];
 		$asyncListenersByPriority = [];
 		foreach($handlerLists as $currentList){
 			foreach($currentList->handlerSlots as $priority => $listeners){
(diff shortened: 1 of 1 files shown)


src/event/HandlerList.php
```diff
@@ -130,26 +129,24 @@ public function getListenerList() : array{
 		}
 
 		$listenersByPriority = [];
+		$concurrentAsyncListenersByPriority = [];
```
SOF3, line 132: Are you sure we want 34-character-long variable names?

JavierLeon9966: The other option is to make `$listenersByPriorityByType[0]`, `$listenersByPriorityByType[1]`, and `$listenersByPriorityByType[0]` but if we don't hardcode these values it would make it longer than before.

SOF3: Do we actually need to mention all four keywords ("concurrent", "async", "priority", "listener") in the same variable name?

ShockedPlot7560: I don't think so, we have to find shorter terms.

ShockedPlot7560:
Superseded by https://github.com/pmmp/PocketMine-MP/pull/6092/commits/aaa37baf2e4e6c7ae663838ac6e50a6b1c08656b

PR #6122, opened 2023-10-29
Fix loading of sign text from vanilla world

### LINKED ISSUE #6059: Sign text in vanilla worlds is no longer loaded correctly
### Issue description
This is due to the changed NBT structure in newer versions.

1. Create a world in 1.20
2. Add a sign with text
3. Load the world in PM
4. Notice the sign is blank

### OS and versions

* PocketMine-MP: 5.5.0

diamond-gold:
### Relevant issues
Fixes #6059 

## Tests
Tested, sign text is loaded correctly from vanilla world


src/block/tile/Sign.php
```diff
@@ -78,7 +78,10 @@ public function __construct(World $world, Vector3 $pos){
 	}
 
 	public function readSaveData(CompoundTag $nbt) : void{
-		if(($textBlobTag = $nbt->getTag(self::TAG_TEXT_BLOB)) instanceof StringTag){ //MCPE 1.2 save format
+		$frontTextTag = $nbt->getTag(self::TAG_FRONT_TEXT);
+		if($frontTextTag instanceof CompoundTag){
+			$this->readSaveData($frontTextTag);
```
dktapps: I don't like this recursion. A separate helper method to read the text properties would be more appropriate. We'll need such a helper to handle back text once that's implemented too.

diamond-gold:
Is there a need to maintain backwards compatibility now that its not possible for older versions to load newer worlds?

dktapps:
We need to be able to load old data from old versions, but new data saved by PM should be in the latest supported format version.

src/block/tile/Sign.php
```diff
@@ -77,22 +77,29 @@ public function __construct(World $world, Vector3 $pos){
...
 
+	private function readTextTag(CompoundTag $nbt) : void{
+		$baseColor = new Color(0, 0, 0);
+		$glowingText = false;
+		if(($baseColorTag = $nbt->getTag(self::TAG_TEXT_COLOR)) instanceof IntTag){
+			$baseColor = Color::fromARGB(Binary::unsignInt($baseColorTag->getValue()));
+		}
+		if(
+			($glowingTextTag = $nbt->getTag(self::TAG_GLOWING_TEXT)) instanceof ByteTag &&
+			($lightingBugResolvedTag = $nbt->getTag(self::TAG_LEGACY_BUG_RESOLVE)) instanceof ByteTag
```
dktapps: This tag is no longer present with the FrontText/BackText change, which means that all glowing signs stored in the new format won't be handled correctly by this change.

src/block/tile/Sign.php
```diff
@@ -107,7 +114,20 @@ public function readSaveData(CompoundTag $nbt) : void{
...
+			->setByte(self::TAG_PERSIST_FORMATTING, 1)
+		);
+		$nbt->setTag(self::TAG_BACK_TEXT, CompoundTag::create()
+			->setString(self::TAG_TEXT_BLOB, "")
+			->setInt(self::TAG_TEXT_COLOR, Binary::signInt(0xff_00_00_00))
+			->setByte(self::TAG_GLOWING_TEXT, 0)
+			->setByte(self::TAG_PERSIST_FORMATTING, 1)
+		);
+
+		$nbt->setString(self::TAG_TEXT_BLOB, implode("\n", $this->text->getLines())); //Backwards-compatibility
```
dktapps: This is unnecessary

src/block/tile/Sign.php
```diff
@@ -77,22 +77,30 @@ public function __construct(World $world, Vector3 $pos){
...
-			if(
-				($glowingTextTag = $nbt->getTag(self::TAG_GLOWING_TEXT)) instanceof ByteTag &&
-				($lightingBugResolvedTag = $nbt->getTag(self::TAG_LEGACY_BUG_RESOLVE)) instanceof ByteTag
-			){
-				//both of these must be 1 - if only one is set, it's a leftover from 1.16.210 experimental features
-				//see https://bugs.mojang.com/browse/MCPE-117835
-				$glowingText = $glowingTextTag->getValue() !== 0 && $lightingBugResolvedTag->getValue() !== 0;
-			}
-			$this->text = SignText::fromBlob(mb_scrub($textBlobTag->getValue(), 'UTF-8'), $baseColor, $glowingText);
+		$lightingBugResolved = false;
```
dktapps: This ought to be `true` by default

dktapps: Actually, you can just put this variable inside the TAG_TEXT_BLOB branch and pass `true` in the other branch

src/block/tile/Sign.php
```diff
@@ -107,14 +115,19 @@ public function readSaveData(CompoundTag $nbt) : void{
...
+			->setByte(self::TAG_PERSIST_FORMATTING, 1)
+		);
 
-		for($i = 0; $i < SignText::LINE_COUNT; ++$i){ //Backwards-compatibility
-			$textKey = sprintf(self::TAG_TEXT_LINE, $i + 1);
-			$nbt->setString($textKey, $this->text->getLine($i));
-		}
-		$nbt->setInt(self::TAG_TEXT_COLOR, Binary::signInt($this->text->getBaseColor()->toARGB()));
-		$nbt->setByte(self::TAG_GLOWING_TEXT, $this->text->isGlowing() ? 1 : 0);
 		$nbt->setByte(self::TAG_LEGACY_BUG_RESOLVE, 1);
```
dktapps: This can be removed now that we're using the new format

dktapps (approved):

PR #6134, opened 2023-11-03
Fix Iron & Gold ores not being affected by fortune

ipad54:
## Introduction
According to [Minecraft Wiki](https://minecraft.fandom.com/wiki/Fortune) and my personal test in Vanilla Bedrock, Iron & Gold ores are affected by fortune enchantement and drop 1-4 items (with fortune level 3)

## Tests
https://www.youtube.com/watch?v=EG1Q8I6FoOs

### DIFF
--- a/src/block/GoldOre.php
+++ b/src/block/GoldOre.php
@@ -23,13 +23,14 @@ declare(strict_types=1);
 
 namespace pocketmine\block;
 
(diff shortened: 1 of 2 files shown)


dktapps:
Good catch. I think the original Fortune PR targeted PM4, which would be why these were missed.

PR #6138, opened 2023-11-04
Fix ancient debris not being fireproof

ipad54:
## Introduction
Ancient debris does not burn in lava in vanilla.

## Tests
https://www.youtube.com/watch?v=cGHDNOks7sk


dktapps (approved):
I guess this probably ought to use type tags in the future rather than anonymous classes.

Having to make an anonymous class for this is a bit yucky.

PR #6146, opened 2023-11-09
Implement Border Block

pandaaaBE:
## Introduction
Implement [Border Block](https://minecraft.wiki/w/Border)

- Closed because I found no way how to add this block server-side

### API changes
`VanillaBlocks::BORDER_BLOCK()`

image

image


pandaaaBE:
Nah I think PHPUnit has some errors, not my code ;(

dktapps:
> Nah I think PHPUnit has some errors, not my code ;(

You need to run the script `tests/phpunit/block/regenerate_consistency_check.php`. It'll update a JSON file in the tests folder which should be committed.

src/block/Border.php
```diff
@@ -0,0 +1,28 @@
...
+ *
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\block;
+
+class Border extends Wall{
+
+}
```
dktapps, line 28: Why does this have its own class if you're not going to add any functionality to it?

dktapps (changes requested):
This does not implement any of the border block's useful logic (preventing entities from passing above or below) and as such is entirely dependent on the client to respect this.

pandaaaBE:
> This does not implement any of the border block's useful logic (preventing entities from passing above or below) and as such is entirely dependent on the client to respect this.

Isn't it enough if the border is already client-side or do you mean that server-side mobs can get through?

dktapps:
No, it's not enough.

alvin0319:
Relying on the client-side means the malicious client could bypass it.

pandaaaBE:
ok ok my mistake, I'll correct it

pandaaaBE:
I thinked it would be okay if it is just client-side because of this:
https://discordapp.com/channels/373199722573201408/373199722573201410/1171872447327371386

JavierLeon9966:
> Relying on the client-side means the malicious client could bypass it.

In that case `setNoClientPredictions()` should verify server side immobility but that's outside the scope of this PR.

JavierLeon9966:
> ok ok my mistake, I'll correct it

Remember some entities can cross the border block: https://minecraft.wiki/w/Border

JavierLeon9966:
> I thinked it would be okay if it is just client-side because of this: https://discordapp.com/channels/373199722573201408/373199722573201410/1171872447327371386

I thought client side was enough. Just do what Dylan says.

pandaaaBE:
> > I thinked it would be okay if it is just client-side because of this: https://discordapp.com/channels/373199722573201408/373199722573201410/1171872447327371386
> 
> I thought client side was enough. Just do what Dylan says.

But first I have to find out how I can stop the movements server-side without using barrier blocks, how this is supposed to work is currently a mystery to me

pandaaaBE:
**Ok honestly I don't understand how this is supposed to work server-side, if someone wants to continue the PR then I won't stop them.**

pandaaaBE:
Notitz: Client-side it is really useless, when rejoining it is as if the border had never existed...

JavierLeon9966:
> Notitz: Client-side it is really useless, when rejoining it is as if the border had never existed...

Yeah, you need to do something special with the LevelChunkPacket encoding.

pandaaaBE:
@dktapps how would you implement this block?

PR #6152, opened 2023-11-11
Add `Promise::all`

ShockedPlot7560:
## Introduction
This method is necessary in certain RPs, and this one tends to take out this modification, which is no longer specific to a single one.

Its aim is to make a Promise that will be resolved when all given Promises have been resolved. It is, however, rejected as soon as the first promise is rejected.

### PR needing
* #6092 
* #6015 

### API changes
* Add static method `all` to `Promise` 

## Tests
See needing PR


src/promise/Promise.php
```diff
@@ -52,4 +53,53 @@ public function onCompletion(\Closure $onSuccess, \Closure $onFailure) : void{
...
+		$values = [];
+		$toResolve = count($promises);
+		$continue = true;
+
+		foreach($promises as $key => $promise){
+			$promise->onCompletion(
+				function(mixed $value) use ($resolver, $key, &$toResolve, &$continue, &$values) : void{
+					$values[$key] = $value;
+
+					if(--$toResolve === 0 && $continue){
```
SOF3: Don't use `--$var` as expressions. It is bug prone and confusing, especially when combined with a short-circuiting `&&` operator.

src/promise/Promise.php
```diff
@@ -52,4 +53,53 @@ public function onCompletion(\Closure $onSuccess, \Closure $onFailure) : void{
...
+				},
+				function() use ($resolver, &$continue) : void{
+					if($continue){
+						$continue = false;
+						$resolver->reject();
+					}
+				}
+			);
+
+			if(!$continue){
```
SOF3, line 103: Does `onCompletion` have any side effects? Is there any observable difference if we short circuit here, or is it just a micro optimization?

ShockedPlot7560: There's no need to continue the loop if you come across a promise that's already been rejected (the only way to access this part).

SOF3: Yes, that's exactly my question. Does `onCompletion` trigger any behavior in the other promises? Would it be inconsistent if we don't call `onCompletion` there?

(This would be an inconsistency in await-generator because generators don't start executing until they are yielded, not sure if PM has any similar issues)

ShockedPlot7560: I don't think I understand what you're getting at.

dktapps: He's asking if there are any side effects from not adding a completion callback to the other promises. As far as I can see, there shouldn't be.

ShockedPlot7560: I don't think so because we reject the promise as soon as one is rejected.

src/promise/Promise.php
```diff
@@ -52,4 +53,54 @@ public function onCompletion(\Closure $onSuccess, \Closure $onFailure) : void{
...
+		$toResolve = count($promises);
+		$continue = true;
+
+		foreach($promises as $key => $promise){
+			$promise->onCompletion(
+				function(mixed $value) use ($resolver, $key, &$toResolve, &$continue, &$values) : void{
+					$values[$key] = $value;
+
+					$toResolve--;
+					if($toResolve === 0 && $continue){
```
dktapps: It should be impossible for `$toResolve` to reach `0` if any of the promises were rejected, right? So it seems to me like the right side of this condition is actually unnecessary anyway.

dktapps: I thought PHPStan was bugged in its handling of this, but I'm starting to wonder if it actually played a galaxy brain move by finding this

src/promise/Promise.php
```diff
@@ -52,4 +53,54 @@ public function onCompletion(\Closure $onSuccess, \Closure $onFailure) : void{
...
+			);
+
+			if(!$continue){
+				break;
+			}
+		}
+
+		if($toResolve === 0){
+			$continue = false;
+			$resolver->resolve($values);
```
dktapps: If all of the input promises were pre-resolved, you'll be resolving this twice here

src/promise/Promise.php
ShockedPlot7560: > It should be impossible for `$toResolve` to reach `0` if any of the promises were rejected, right? So it seems to me like the right side of this condition is actually unnecessary anyway.

If any promises were rejected, we don't need ``toResolve`` anyway ?

dktapps: No, the individual promises could still resolve correctly.
I'm saying the `$continue` check is redundant, since `$toResolve` cannot reach `0` in any condition where `$continue` becomes `false`.

ShockedPlot7560: You're right

dktapps:
After adding some unit tests I think this is OK to merge.

PR #6162, opened 2023-11-15
Fix pitcher pod wrongly registered as a block

### LINKED ISSUE #6133: Impossible to give this item pitcher_plant
Hello, in your latest update (5.8.1), you've implemented some items like pitcher plant but we can't get to them or see them in the creative menu.

diamond-gold:
## Introduction
This PR fixes the wrong registration of pitcher pod as a block
Also fixes wrong drops for pitcher crop
Fixes #6133 

## Tests
Not tested


src/block/DoublePitcherCrop.php
```diff
@@ -107,4 +108,16 @@ public function onRandomTick() : void{
...
+		if($this->age >= self::MAX_AGE){
+			return [
+				VanillaBlocks::PITCHER_PLANT()->asItem()
+			];
+		}else{
+			return [
+				VanillaItems::PITCHER_POD()
+			];
+		}
+	}
```
dktapps, line 119: This can be reduced to a ternary

src/item/StringToItemParser.php
```diff
@@ -869,7 +869,7 @@ private static function registerBlocks(self $result) : void{
 		$result->registerBlock("pink_tulip", fn() => Blocks::PINK_TULIP());
 		$result->registerBlock("piglin_head", fn() => Blocks::MOB_HEAD()->setMobHeadType(MobHeadType::PIGLIN));
 		$result->registerBlock("pitcher_plant", fn() => Blocks::PITCHER_PLANT());
-		$result->registerBlock("pitcher_pod", fn() => Blocks::PITCHER_CROP());
+		$result->registerBlock("pitcher_crop", fn() => Blocks::PITCHER_CROP());
```
dktapps: This isn't normally available via /give

diamond-gold: Blocks like wheat_block, potato_block, carrot_block are not normally available via /give too but they are registered in STIP

dktapps: That's because they were available in the legacy Item::fromString(), so they have to be kept for BC reasons. New items should not be doing this.

dktapps (changes requested):
This breaks middle-click block-picking.

src/block/PitcherCrop.php
```diff
@@ -101,4 +102,14 @@ public function onRandomTick() : void{
 			$this->grow(null);
 		}
 	}
+
+	public function getDropsForCompatibleTool(Item $item) : array{
+		return [
+			VanillaItems::PITCHER_POD()
+		];
+	}
```
dktapps: This isn't necessary with asItem() also implemented

dktapps (approved):

PR #6167, opened 2023-11-18
Position and Location function return now itself

DebHugo:
## Introduction
Allow function like `add`, `ceil`... in `Position` ans `Location` to return corresponding class instead of `Vector3`
This change allow to simplified code.

### Before:
```
$pos = new Position(1, 1, 1, $world);
$pos = Position::fromObject($pos->add(0.5, 0.5, 0.5), $pos->getWorld());
```
### After:
```
$pos = new Position(1, 1, 1, $world)->add(0.5, 0.5, 0.5);
```
$pos is now `Position` instead of `Vector3`

### API changes
Add function `add`, `multiply`, `divide`, `ceil`, `floor` in `Position` and `Location`

### DIFF
--- a/src/block/TNT.php
+++ b/src/block/TNT.php
@@ -100,7 +100,7 @@ class TNT extends Opaque{
 
 		$mot = (new Random())->nextSignedFloat() * M_PI * 2;
 
-		$tnt = new PrimedTNT(Location::fromObject($this->position->add(0.5, 0, 0.5), $world));
+		$tnt = new PrimedTNT($this->position->add(0.5, 0, 0.5));
 		$tnt->setFuse($fuse);
 		$tnt->setWorksUnderwater($this->worksUnderwater);
 		$tnt->setMotion(new Vector3(-sin($mot) * 0.02, 0.2, -cos($mot) * 0.02));
--- a/src/block/TNT.php
+++ b/src/block/TNT.php
@@ -100,7 +100,7 @@ class TNT extends Opaque{
 
 		$mot = (new Random())->nextSignedFloat() * M_PI * 2;
 
-		$tnt = new PrimedTNT($this->position->add(0.5, 0, 0.5));
+		$tnt = new PrimedTNT(Location::fromObject($this->position->add(0.5, 0, 0.5), null));
 		$tnt->setFuse($fuse);
 		$tnt->setWorksUnderwater($this->worksUnderwater);
 		$tnt->setMotion(new Vector3(-sin($mot) * 0.02, 0.2, -cos($mot) * 0.02));
--- a/src/block/utils/FallableTrait.php
+++ b/src/block/utils/FallableTrait.php
@@ -25,7 +25,6 @@ namespace pocketmine\block\utils;
 
 use pocketmine\block\Block;
 use pocketmine\block\VanillaBlocks;
-use pocketmine\entity\Location;
 use pocketmine\entity\object\FallingBlock;
 use pocketmine\math\Facing;
 use pocketmine\utils\AssumptionFailedError;
--- a/src/block/utils/FallableTrait.php
+++ b/src/block/utils/FallableTrait.php
@@ -25,6 +25,7 @@ namespace pocketmine\block\utils;
 
 use pocketmine\block\Block;
 use pocketmine\block\VanillaBlocks;
+use pocketmine\entity\Location;
 use pocketmine\entity\object\FallingBlock;
 use pocketmine\math\Facing;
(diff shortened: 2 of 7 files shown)


DebHugo:
I'm not really fan of my modification. In my opinion there is a better way that doesn't require redeclaring all functions and is faster for Location. May be a function like this:
```
#In Vector3 class
protected function redifine(float|int $x, float|int $y, float|int $z): Vector3 {
	return new Vector3($x, $y, $z);
}

#In Position class
protected function redifine(float|int $x, float|int $y, float|int $z): Position {
	return new Position($x, $y, $z, $this->world);
}

#In Location class
protected function redifine(float|int $x, float|int $y, float|int $z): Location {
	return new Location($x, $y, $z, $this->world, $this->yaw, $this->pitch);
}
```
Function `redfine` would be call in `add`, `ceil`... functions and these function would be return `static` instead of `Vector3`
I wait your opinion.

SOF3:
This is BC breaking because `$player->getLocation()->add(1,0,0)` now leaks the world.

Also it would become confusing which methods are overridden and which ones aren't. I think the root problem is that Position shouldn't have extended Vector3 to begin with.

Nerahikada:
Related: #1367, #2989

SOF3:
Also, it doesn't make sense for that `Position->add(Vector3)` returns Position but `Position->subtract(Vector3)` doesn't. Meanwhile, it makes even less sense that `Position->subtract(Position)` returns Position. So there is just no sane way to implement this in PHP.

[A white horse is not a horse](https://en.wikipedia.org/wiki/When_a_white_horse_is_not_a_horse), and a Position is not a Vector3. From what I understand, Vector3 is primarily a *relative* vector, which should not be usable as a position vector in any case.

DebHugo:
> Also, it doesn't make sense for that `Position->add(Vector3)` returns Position but `Position->subtract(Vector3)` doesn't. 

I just forgot subtract

DebHugo:
I just had an idea. Why not create a Position class that doesn't extend Vector3 and has a constructor like this
```
public function __construct(Vector3 $vector, World $world)
```

Consequently for have a vector use `Position->getVector()`
This manerie can fix issues with `Position` and `Location` class ?

SOF3:
@Dhaiven this is what we've wanted to do all the time, except it would be yet another of the biggest BC breaks of all time

dktapps:
Yeah, composing them would be the best technical solution, but would be shitty because of the amount of boilerplate code it would generate (look at the mess created from embedding Position in Block rather than extending it).

I've thought about doing this for Location specifically to split up the position and rotation parts, which would have a much smaller scope, but it would create problems for stuff like teleport().

> I just had an idea. Why not create a Position class that doesn't extend Vector3 and has a constructor like this
> 
> ```
> public function __construct(Vector3 $vector, World $world)
> ```
> 
> Consequently for have a vector use `Position->getVector()` This manerie can fix issues with `Position` and `Location` class ?

SOF3:
In either case,
- It doesn't make sense that to perform `Position->multiply()`. The quantities in a position vector don't make sense unless you compare it to another position. In my opinion we should even *disallow* `Position->multiply()` if possible since that's almost definitely a bug.
- Same for `divide`.

dktapps:
> In my opinion we should even disallow Position->multiply() if possible since that's almost definitely a bug.

I wouldn't go out of my way to do this. Without an architectural change this would violate LSP.

dktapps:
> Meanwhile, it makes even less sense that Position->subtract(Position) returns Position.

Not sure what you meant by this, but the ambiguity in deciding which Position's World gets used is problematic indeed.

SOF3:
> > Meanwhile, it makes even less sense that Position->subtract(Position) returns Position.
> 
> Not sure what you meant by this, but the ambiguity in deciding which Position's World gets used is problematic indeed.

The difference between two positions is a relative vector not a position. `(1010, 1010, 1010, world)` minus `(1001, 1001, 1001, world)` has absolutely nothing to do with the position `(9, 9, 9, world)`.

dktapps:
Yeah, that's valid. I think this should be closed.

SOF3:
If we actually have disjoint Vector3 and Position types, the following combinations involving Position would be valid:
```php
Position + Vector3 -> Position
Vector3 + Position -> Position
Position - Position -> Vector3
Position - Vector3 -> Position
```

all other operations, including all multiplication and division operations, would be unreasonable.

ceil and floor are fine, but they just don't look so nice in terms of BC when they leak the world.

PR #6176, opened 2023-11-26
AsyncTask: Fix retrieval of null data from the thread-local storage

TheNewHEROBRINEX:
## Introduction
Sometimes it can be useful to be able to store null data with `AsyncTask->storeLocal()` (e.g. if you want to make some complex parameters of the constructor optional).
Currently, `AsyncTask->storeLocal()` allows to store null data in the thread-local storage. However, an error will occur when attempting to fetch this data later on because `isset()` doesn't distinguish between non-existing data and null data. This PR uses `array_key_exists()` to fix this problem.

### API changes
None.

### Behavioural changes
Now the server doesn't crash anymore when attempting to fetch null data from the thread-local storage in an AsyncTask.

## Backwards compatibility
Yes.

## Tests
Tests have been done using a script plugin. Results can be seen here: https://gist.github.com/TheNewHEROBRINEX/516193c03a66332fede4cf126f20e8cf

### DIFF
--- a/src/scheduler/AsyncTask.php
+++ b/src/scheduler/AsyncTask.php
@@ -28,6 +28,7 @@ use pmmp\thread\Thread as NativeThread;
 use pmmp\thread\ThreadSafe;
 use pmmp\thread\ThreadSafeArray;
 use pocketmine\thread\NonThreadSafeValue;
+use function array_key_exists;
 use function assert;
 use function igbinary_serialize;
 use function igbinary_unserialize;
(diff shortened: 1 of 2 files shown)


src/scheduler/AsyncTask.php
```diff
@@ -230,7 +231,7 @@ protected function storeLocal(string $key, mixed $complexData) : void{
 	 */
 	protected function fetchLocal(string $key){
 		$id = spl_object_id($this);
-		if(!isset(self::$threadLocalStorage[$id][$key])){
+		if(!isset(self::$threadLocalStorage[$id]) || !array_key_exists($key, self::$threadLocalStorage[$id])){
```
JavierLeon9966, line 234: Couldn't this just use `array_key_exists()`? `isset()` is now redundant.

TheNewHEROBRINEX: If I remove the `isset()` check, when you try to call `fetchLocal()` while never having done `storeLocal()` it will cause an "Undefined array key" error instead of throwing the proper exception.

JavierLeon9966: Oh, I get it now

dktapps:
Please add a PHPUnit test for this

PR #6177, opened 2023-11-26
add WorldSyncTimeEvent

BrandPVP:
Allows making per-player world time

### DIFF
--- a/src/world/World.php
+++ b/src/world/World.php
@@ -907,8 +907,8 @@ class World implements ChunkManager{
 			return;
 		}
 
-		foreach($ev->getRecipients() as $player){
-			$player->getNetworkSession()->syncWorldTime($ev->getTime());
+		foreach($targets as $player){
+			$player->getNetworkSession()->syncWorldTime($ev->getTimeFor($player));
 		}
 	}
 
--- a/src/world/World.php
+++ b/src/world/World.php
@@ -901,10 +901,13 @@ class World implements ChunkManager{
 			$targets = $this->players;
 		}
 
-		foreach($targets as $player){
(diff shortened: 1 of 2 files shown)


src/world/World.php
```diff
@@ -899,8 +900,12 @@ public function sendTime(Player ...$targets) : void{
 		if(count($targets) === 0){
 			$targets = $this->players;
 		}
+
 		foreach($targets as $player){
-			$player->getNetworkSession()->syncWorldTime($this->time);
+			$ev = new WorldSyncTimeEvent($this, $player, $this->time);
```
jasonw4331: It's preferred to hold an array of all recipients instead of spamming out event calls to all handlers

jasonw4331 (changes requested):

src/world/World.php
BrandPVP: That was in the first commit. But as it turns out, it doesn't affect performance so we can call it for every player

dktapps: The original way was preferable

BrandPVP: I'm not sure if the original way was that useful.. The only thing you can do with that is cancel the event and send a packet on your own. But we can add `hasHandlers()` check I guess?

dktapps: I don't like calling an event for every player when it's so easily avoided.

BrandPVP: What about now?

dktapps:
I don't feel like this is the right way to support per-player time. Waiting for an event to do it just seems weird.

dktapps:
Closing this as I don't think it's the right way to implement the feature you want.

PR #6187, opened 2023-12-04
Implement Monster Spawner functionality

diamond-gold:
## Introduction
This PR implements the functionality of the Monster Spawner block

### API changes
Added SpawnerAttemptSpawnEvent

## Follow-up
TODO: spawn condition check (light level etc.)
TODO: spawnData, spawnPotentials

## Tests
Not Tested


src/item/SpawnEgg.php
```diff
@@ -34,7 +35,14 @@ abstract class SpawnEgg extends Item{
 
 	abstract protected function createEntity(World $world, Vector3 $pos, float $yaw, float $pitch) : Entity;
 
+	public function getEntityTypeId() : ?string{
```
ShockedPlot7560: I don't really like this function. A SpawnEgg need to be able to create an entity, however, it may haven't an EntityTypeId ?

src/block/tile/MonsterSpawner.php
```diff
@@ -130,4 +140,126 @@ protected function addAdditionalSpawnData(CompoundTag $nbt) : void{
 
 		$nbt->setFloat(self::TAG_ENTITY_SCALE, $this->displayEntityScale);
 	}
+
+	public function onUpdate() : bool{
+		if($this->closed || $this->entityTypeId === ":"){
```
ShockedPlot7560: Why `$this->entityTypeId === ":"` ?

src/block/tile/MonsterSpawner.php
```diff
@@ -130,4 +140,126 @@ protected function addAdditionalSpawnData(CompoundTag $nbt) : void{
...
+		$position = $this->getPosition();
+		$world = $position->getWorld();
+		if($world->getNearestEntity($position, $this->requiredPlayerRange, Player::class) === null){
+			return true;
+		}
+		$count = 0;
+		foreach($world->getNearbyEntities(
+			AxisAlignedBB::one()
+				->offset($position->x, $position->y, $position->z)
+				->expand($this->spawnRange * 2 + 1, 8, $this->spawnRange * 2 + 1)
```
ShockedPlot7560: This is scalable according to the entity

src/block/tile/MonsterSpawner.php
```diff
@@ -130,4 +140,126 @@ protected function addAdditionalSpawnData(CompoundTag $nbt) : void{
...
+			$ev = new SpawnerAttemptSpawnEvent($this->getBlock(), $entityTypeId);
+			$ev->call();
+			if($ev->isCancelled()){
+				return true;
+			}
+			$entityTypeId = $ev->getEntityType();
+		}
+		// TODO: spawn condition check (light level etc.)
+		for($i = 0; $i < $this->spawnPerAttempt; $i++){
+			$spawnLocation = $position->add(mt_rand(-$this->spawnRange, $this->spawnRange), 0, mt_rand(-$this->spawnRange, $this->spawnRange));
```
ShockedPlot7560: I thought height could also be varied

src/block/MonsterSpawner.php
```diff
@@ -38,10 +42,30 @@ protected function getXpDropAmount() : int{
...
 	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if($item instanceof SpawnEgg){
+			if($item->getEntityTypeId() === null){
+				return false;
+			}
+			$spawner = $this->position->getWorld()->getTile($this->position);
+			if($spawner instanceof TileSpawner){
+				$spawner->setEntityTypeId($item->getEntityTypeId());
```
ShockedPlot7560: I think its better to store the egg in the spawner instead of the entityTypeId ? So, it can be used to create an entity later ?

src/item/SpawnEgg.php
diamond-gold: Do you have a better way for the spawner to know what entity should be spawned when clicked with a spawn egg? I made it nullable for that reason

src/block/tile/MonsterSpawner.php
diamond-gold: A spawner that has no entity to spawn have no need to be ticked

src/block/tile/MonsterSpawner.php
diamond-gold: I'm not sure of the exact vanilla implementation, I got the spawnRange * 2 + 1 from wiki, how should I make it?

src/block/MonsterSpawner.php
diamond-gold: What do you mean store the egg in the spawner? Vanilla only saves the entity id

dktapps: The runtime implementation should not be dealing with entity IDs

diamond-gold: Could you perhaps suggest a way to link the spawn egg and the entity type? Should I create a registry just for that?

src/block/tile/MonsterSpawner.php
ShockedPlot7560: Zombie spawner as not the same spawn range than blaze or magma cube. If we want to implement it in a good way, we need an API to define this range

src/block/tile/MonsterSpawner.php
```diff
@@ -130,4 +146,125 @@ protected function addAdditionalSpawnData(CompoundTag $nbt) : void{
 
 		$nbt->setFloat(self::TAG_ENTITY_SCALE, $this->displayEntityScale);
 	}
+
+	public function onUpdate() : bool{
```
ShockedPlot7560: I dont think this method need to be here. We want to keep tile only for storing data. His place are in the block class.

src/item/SpawnEggEntityRegistry.php
```diff
@@ -0,0 +1,56 @@
...
+	/**
+	 * @phpstan-var array<string, SpawnEgg>
+	 * @var SpawnEgg[]
+	 */
+	private array $entityMap = [];
+
+	private function __construct(){
+		$this->register('minecraft:squid', VanillaItems::SQUID_SPAWN_EGG());
+		$this->register('minecraft:villager', VanillaItems::VILLAGER_SPAWN_EGG());
+		$this->register('minecraft:zombie', VanillaItems::ZOMBIE_SPAWN_EGG());
```
ShockedPlot7560: Don't use hardcoded ids

src/item/SpawnEggEntityRegistry.php
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	public function register(string $entitySaveId, SpawnEgg $spawnEgg) : void{
+		$this->entityMap[$entitySaveId] = $spawnEgg;
+	}
+
+	public function getEntityId(SpawnEgg $item) : ?string{
+		foreach(Utils::stringifyKeys($this->entityMap) as $entitySaveId => $spawnEgg){
```
ShockedPlot7560: No need to have the call to stringify

src/block/tile/MonsterSpawner.php
diamond-gold: That is not consistent with other tiles such as furnace and brewing stand, where all the logic is in the tile class

src/block/tile/MonsterSpawner.php
diamond-gold: Based on my observation, in vanilla it only spawns entities on the same height as the spawner

src/item/SpawnEggEntityRegistry.php
diamond-gold: It is also hardcoded in EntityFactory, should I use EntityId constants instead?

src/item/SpawnEggEntityRegistry.php
diamond-gold: phpstan will complain without it
```
------ -------------------------------------------------------------------------------------
  Line   src\item\SpawnEggEntityRegistry.php
 ------ -------------------------------------------------------------------------------------
  49     Unsafe foreach on array with key type string (they might be casted to int).
         💡 Use pocketmine\utils\Utils::stringifyKeys() for a safe Generator-based iterator.
 ------ -------------------------------------------------------------------------------------
```

src/block/tile/MonsterSpawner.php
ShockedPlot7560: Are we sure of that ?

src/block/tile/MonsterSpawner.php
ShockedPlot7560: If i remember, the ambition was to moove the furnace logics etc into the block. Maybe i'm wrong

diamond-gold: So where should I put the logic? It will still have to be called from onUpdate() if its in the block class

ShockedPlot7560: We should put the logic into the block. The tile need only to be used for storing or actions which arent possible with blocks. As we can see, in the furnace class, there is a TODO, to moove all the logic into the block.
That makes sense because this `onUpdate` function is only called in the block class, so why putting it into the tile ?

src/block/tile/SpawnerSpawnRangeRegistry.php
```diff
@@ -0,0 +1,45 @@
...
+use pocketmine\utils\SingletonTrait;
+
+final class SpawnerSpawnRangeRegistry{
+	use SingletonTrait;
+
+	/**
+	 * @phpstan-var array<string, AxisAlignedBB>
+	 * @var AxisAlignedBB[]
+	 */
+	private array $spawnRange;
```
ShockedPlot7560: You need to initialize them

src/block/tile/SpawnerSpawnRangeRegistry.php
```diff
@@ -0,0 +1,45 @@
...
+
+	public function register(string $entitySaveId, AxisAlignedBB $spawnRange) : void{
+		$this->spawnRange[$entitySaveId] = $spawnRange;
+	}
+
+	public function getSpawnRange(string $entitySaveId) : ?AxisAlignedBB{
+		return (clone $this->spawnRange[$entitySaveId]) ?? null;
```
ShockedPlot7560: Will throw an error if the entity isnt registered.

src/block/tile/SpawnerSpawnRangeRegistry.php
```diff
@@ -0,0 +1,45 @@
...
+
+	/**
+	 * @phpstan-var array<string, AxisAlignedBB>
+	 * @var AxisAlignedBB[]
+	 */
+	private array $spawnRange;
+
+	public function register(string $entitySaveId, AxisAlignedBB $spawnRange) : void{
+		$this->spawnRange[$entitySaveId] = $spawnRange;
+	}
```
ShockedPlot7560: Need to check if its already registered

src/event/block/SpawnerAttemptSpawnEvent.php
```diff
@@ -0,0 +1,44 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\event\block;
+
+use pocketmine\block\Block;
+use pocketmine\event\Cancellable;
+use pocketmine\event\CancellableTrait;
+
+class SpawnerAttemptSpawnEvent extends BlockEvent implements Cancellable{
```
ShockedPlot7560, line 30: Wouldn't it be possible to change more values? The size, the number? Is it useful ?

src/item/SpawnEggEntityRegistry.php
```diff
@@ -0,0 +1,57 @@
...
+
+	private function __construct(){
+		$this->register(EntityIds::SQUID, VanillaItems::SQUID_SPAWN_EGG());
+		$this->register(EntityIds::VILLAGER, VanillaItems::VILLAGER_SPAWN_EGG());
+		$this->register(EntityIds::ZOMBIE, VanillaItems::ZOMBIE_SPAWN_EGG());
+	}
+
+	public function register(string $entitySaveId, SpawnEgg $spawnEgg) : void{
+		$this->entityMap[$entitySaveId] = $spawnEgg;
+	}
```
ShockedPlot7560: We probably need to check if its already present

ShockedPlot7560: Would be better to store in a TypeIdMap instead of an EntityIdMap : array<typeId, entitySaveId>. This avoids having to iterate each time you search for an entity.

ShockedPlot7560 (commented):
By moving the logic to a block, I mean something like Noteblock where all the data is only read once and the tile is only used to store the data. At runtime, the block class is used, enough.

pandaaaBE:
does the author mind if i help with this pr?

diamond-gold:
> does the author mind if i help with this pr?

I appreciate any help, feel free to offer suggestions or PR

src/event/block/SpawnerAttemptSpawnEvent.php
diamond-gold: I don't think it makes sense to allow changing those in the event, those can be changed on the spawner block

src/block/MonsterSpawner.php
```diff
@@ -23,12 +23,102 @@
...
+			$this->displayEntityWidth = $tile->getDisplayEntityWidth();
+			$this->displayEntityHeight = $tile->getDisplayEntityHeight();
+			$this->displayEntityScale = $tile->getDisplayEntityScale();
+			$this->spawnDelay = $tile->getSpawnDelay();
+			$this->minSpawnDelay = $tile->getMinSpawnDelay();
+			$this->maxSpawnDelay = $tile->getMaxSpawnDelay();
+			$this->spawnPerAttempt = $tile->getSpawnPerAttempt();
+			$this->maxNearbyEntities = $tile->getMaxNearbyEntities();
+			$this->spawnRange = $tile->getSpawnRange();
+			$this->requiredPlayerRange = $tile->getRequiredPlayerRange();
```
dktapps, line 84: Not liking this. A data class to wrap all these up into one item to be passed between block & tile would be a lot cleaner.

dktapps (changes requested):
I really don't like how much this depends on Mojang internal data. Not sure if there's currently a way to avoid this, but this is why I didn't implement mob spawners myself.

PR #6194, opened 2023-12-09
feat(Player): introduce `PlayerPacketHandlerChangeEvent`

cooldogedev:
### Behavioural changes

`NetworkSession->setHandler(PacketHandler)` now fires `PlayerPacketHandlerChangeEvent`

### DIFF
--- a/src/event/player/PlayerPacketHandlerChangeEvent.php
+++ b/src/event/player/PlayerPacketHandlerChangeEvent.php
@@ -1,0 +1,44 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * @author PocketMine Team
+ * @link http://www.pocketmine.net/
+ *
+ *
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\event\player;
+
+use pocketmine\event\Event;
+use pocketmine\network\mcpe\handler\PacketHandler;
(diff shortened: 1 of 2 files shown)


dktapps:
What's the use case?

src/event/player/PlayerPacketHandlerChangeEvent.php
```diff
@@ -0,0 +1,44 @@
...
+
+	public function getNetworkSession() : NetworkSession{
+		return $this->session;
+	}
+
+	public function getHandler() : ?PacketHandler{
+		return $this->handler;
+	}
+
+	public function setHandler(?PacketHandler $handler) : void{
```
dktapps, line 41: I don't like this. This opens lots of options to break things.

src/event/player/PlayerPacketHandlerChangeEvent.php
```diff
@@ -0,0 +1,44 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\event\player;
+
+use pocketmine\event\Event;
+use pocketmine\network\mcpe\handler\PacketHandler;
+use pocketmine\network\mcpe\NetworkSession;
+
+class PlayerPacketHandlerChangeEvent extends Event{
```
dktapps, line 30: Misleading to refer to this as "player" and to put it in this namespace.

dktapps (changes requested):

cooldogedev:
> What's the use case?

[Check out this example](https://github.com/cooldogepm/WDPELoginExtras). This plugin overrides the default Login packet handler to support WaterdogPE's login extras. However, it currently listens for all incoming packets, making the override process less than optimal in terms of convenience.

src/event/player/PlayerPacketHandlerChangeEvent.php
cooldogedev: Without creating a new namespace it has to be either `event/player` or `event/server`

dktapps:
WaterdogPE's design in general is less than optimal ...

SenseiTarzan:
> WaterdogPE's design in general is less than optimal ...

I find this thing interesting because I have to do a hack to deactivate a handler because I have a synchronization system and it's me who manages the call that makes the setHandler(InGameHandler), WaterdogPE could make a new packets but they didn't and I find that a shame

dktapps:
My gut reaction is that I don't like this change. The event's naming is problematic, but also I don't like the general concept of it.

Since this is primarily intended at supporting Waterdog, I feel like a more targeted approach to dealing with the extra JWT fields should be used.

dktapps:
I think something that makes use of https://github.com/pmmp/netresearch-jsonmapper/blob/fdeac5f26eb7e61bdbab5e0948de8a71f3795102/src/JsonMapper.php#L112 would be better for WDPE than this approach. For example, if we added a `mixed[] $unknownProperties` element to the JSON models, and then added an undefined handler to collect all of them into one place, this would still allow the server to warn when there are unexpected new elements, while not preventing WDPE from working.

PR #5793, opened 2023-06-02
New colors

Brougud:
## Introduction

I added the new colors, and fixed some issues, the Terminal stuff may not be correct yet, still need help on that

### API changes

**Added Following Constants to `TextFormat`**
- `MATERIAL_QUARTZ`
- `MATERIAL_IRON`
- `MATERIAL_NETHERITE`
- `MATERIAL_REDSTONE`
- `MATERIAL_COPPER`
- `MATERIAL_GOLD`
- `MATERIAL_EMERALD`
- `MATERIAL_DIAMOND`
- `MATERIAL_LAPIS`
- `MATERIAL_AMETHYST`

PR #5775, opened 2023-05-29
Add of new minecraft colours, deletion of colours that have been removed from the game

roimee6:
## Introduction
Minecraft has just added the colours that are linked to the materials (colours slightly modified from the classic ones) and Minecraft has removed a few other colours (I don't know since when) which are the underline and the strikethrough. 

## Tests
On this photo I realised that minecraft didn't have the underline and the strikethrough (I removed them).
image

PR #6202, opened 2023-12-18
Implemented smithing table functionality

HimmelKreis4865:
## Introduction
This PR adds working smithing tables to pocketmine.
There are no known issues or problems 

### API changes
`Amor->setTrim(ArmorTrim)`
`Armor->getTrim() : ?ArmorTrim`
`Armor->clearTrim()`

`SmithingTableInventory->getInput() : Item`
`SmithingTableInventory->getAddition() : Item`
`SmithingTableInventory->getTemplate() : Item`

The new enums: `ArmorTrimPattern` & `ArmorTrimMaterial` (required to construct `ArmorTrim`)

## Tests

Tested ingame with the smithing table several times:
image

Tested the code
```php
$player->getInventory()->addItem(VanillaItems::LEATHER_CAP()->setTrim(new ArmorTrim(ArmorTrimMaterial::QUARTZ, ArmorTrimPattern::EYE)));
```


src/crafting/CraftingManagerFromDataHelper.php
```diff
@@ -292,7 +294,6 @@ public static function make(string $directoryPath) : CraftingManager{
 				$input
 			));
 		}
-
```
IvanCraft623: Unrelated change

IvanCraft623 (changes requested):

src/block/inventory/SmithingTableInventory.php
```diff
@@ -25,13 +25,39 @@
...
+
+	public function getTemplate() : Item{
+		return $this->getItem(self::SLOT_TEMPLATE);
+	}
+
+	public function getOutput() : ?Item{
+		$craftingManager = Server::getInstance()->getCraftingManager();
+		$recipe = $craftingManager->matchSmithingRecipe($this->getInput(), $this->getAddition(), $this->getTemplate());
+		return $recipe?->constructOutput($this->getInput(), $this->getAddition(), $this->getTemplate());
 	}
```
dktapps, line 40: This doesn't belong here. It's not relevant to the function of the class.

src/block/inventory/SmithingTableInventory.php
```diff
@@ -25,13 +25,39 @@
...
+	public const SLOT_INPUT = 0;
+
+	public const SLOT_ADDITION = 1;
+
+	public const SLOT_TEMPLATE = 2;
+
 	public function __construct(Position $holder){
 		$this->holder = $holder;
-		parent::__construct(3);
+		parent::__construct(4);
```
dktapps: This is not correct

src/crafting/SmithingRecipe.php
```diff
@@ -0,0 +1,37 @@
...
+ *
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\crafting;
+
+use pocketmine\item\Item;
+
+abstract class SmithingRecipe{
```
dktapps: This might as well be an interface

src/crafting/SmithingRecipe.php
```diff
@@ -0,0 +1,37 @@
...
+
+	abstract public function getInput() : RecipeIngredient;
+
+	abstract public function getAddition() : RecipeIngredient;
+
+	abstract public function getTemplate() : RecipeIngredient;
+
+	abstract public function constructOutput(Item $input, Item $addition, Item $template) : ?Item;
```
dktapps: `getResultFor()` would be more consistent with other stuff in this package

src/inventory/transaction/SmithingTransaction.php
```diff
@@ -0,0 +1,83 @@
...
+			throw new TransactionValidationException("Expected 1 output item, but received $outputCount");
+		}
+
+		$input = $addition = $template = null;
+		foreach($inputs as $item){
+			switch(true){
+				case $item instanceof Armor || $item instanceof TieredTool:
+					$input = $item;
+					break;
+				case $item->getTypeId() >= ItemTypeIds::NETHERITE_UPGRADE_SMITHING_TEMPLATE && $item->getTypeId() <= ItemTypeIds::SPIRE_ARMOR_TRIM_SMITHING_TEMPLATE:
```
dktapps: Do not compare type IDs like this

src/inventory/transaction/SmithingTransaction.php
```diff
@@ -0,0 +1,83 @@
...
+		}
+
+		if($input === null || $addition === null || $template === null){
+			throw new TransactionValidationException("The given inputs are no valid smithing ingredients");
+		}
+
+		$craftingManager = Server::getInstance()->getCraftingManager();
+		$recipe = $craftingManager->matchSmithingRecipe($input, $addition, $template);
+
+		if(($output = $recipe?->constructOutput($input, $addition, $template)) === null){
```
dktapps: an explicit null check with a separate exception would be clearer

src/inventory/transaction/SmithingTransaction.php
```diff
@@ -0,0 +1,83 @@
...
+					$template = $item;
+					break;
+				default:
+					$addition = $item;
+					break;
+			}
+		}
+
+		if($input === null || $addition === null || $template === null){
+			throw new TransactionValidationException("The given inputs are no valid smithing ingredients");
```
dktapps:
```suggestion
			throw new TransactionValidationException("The given inputs are not valid smithing ingredients");
```

src/inventory/transaction/SmithingTransaction.php
```diff
@@ -0,0 +1,83 @@
...
+		if(($inputCount = count($inputs)) !== 3){
+			throw new TransactionValidationException("Expected 3 input items, got $inputCount");
+		}
+		if(($outputCount = count($outputs)) !== 1){
+			throw new TransactionValidationException("Expected 1 output item, but received $outputCount");
+		}
+
+		$input = $addition = $template = null;
+		foreach($inputs as $item){
+			switch(true){
```
dktapps: An if/elseif would be more idiomatic and require less lines of code

src/inventory/transaction/SmithingTransaction.php
dktapps: Better yet, the exception should report which input is missing

src/item/ArmorTrim.php
```diff
@@ -0,0 +1,40 @@
...
+declare(strict_types=1);
+
+namespace pocketmine\item;
+
+class ArmorTrim{
+
+	public function __construct(
+		private readonly ArmorTrimMaterial $material,
+		private readonly ArmorTrimPattern $pattern
+	){ }
```
dktapps:
```suggestion
	){}
```

src/item/Armor.php
```diff
@@ -160,12 +184,25 @@ protected function deserializeCompoundTag(CompoundTag $tag) : void{
...
 	}
 
 	protected function serializeCompoundTag(CompoundTag $tag) : void{
 		parent::serializeCompoundTag($tag);
 		$this->customColor !== null ?
 			$tag->setInt(self::TAG_CUSTOM_COLOR, Binary::signInt($this->customColor->toARGB())) :
 			$tag->removeTag(self::TAG_CUSTOM_COLOR);
+		$this->armorTrim !== null ?
+			$tag->setTag(self::TAG_TRIM, CompoundTag::create()
+				->setString(self::TAG_TRIM_MATERIAL, $this->armorTrim->getMaterial()->value)
```
dktapps: The save data values shouldn't be directly linked to the enums. We may need to save multiple different types of save data.
A mapping in `pocketmine\data\bedrock` should be used instead.

src/item/Armor.php
```diff
@@ -160,12 +184,25 @@ protected function deserializeCompoundTag(CompoundTag $tag) : void{
...
+				->setString(self::TAG_TRIM_PATTERN, $this->armorTrim->getPattern()->value)) :
```
dktapps: Same again here

src/item/ArmorTrimPattern.php
```diff
@@ -0,0 +1,64 @@
...
+
+	private const TEMPLATE_SUFFIX = "_armor_trim_smithing_template";
+
+	public static function fromItem(Item $item) : ?ArmorTrimPattern{
+		foreach(StringToItemParser::getInstance()->lookupAliases($item) as $alias){
+			if (!str_ends_with($alias, self::TEMPLATE_SUFFIX)){
+				continue;
+			}
+			return self::tryFrom(substr($alias, 0, strlen($alias) - strlen(self::TEMPLATE_SUFFIX)));
+		}
```
dktapps: This code looks very suspicious

src/item/ArmorTrimPattern.php
```diff
@@ -0,0 +1,64 @@
...
+		return null;
+	}
+
+	public function getItemId() : string{
+		return "minecraft:" . $this->value . self::TEMPLATE_SUFFIX;
```
dktapps: Not sure what this is for, but it has no business being in the API

src/network/mcpe/handler/PreSpawnPacketHandler.php
```diff
@@ -144,6 +150,11 @@ public function setUp() : void{
 			$this->session->getLogger()->debug("Sending creative inventory data");
 			$this->inventoryManager->syncCreative();
 
+			$this->session->getLogger()->debug("Sending armor trim data");
+			$patterns = array_map(fn(ArmorTrimPattern $pattern) => new TrimPattern($pattern->getItemId(), $pattern->value), ArmorTrimPattern::cases());
+			$materials = array_map(fn(ArmorTrimMaterial $material) => new TrimMaterial($material->value, $material->getColor(), $material->getItemId()), ArmorTrimMaterial::cases());
+			$this->session->sendDataPacket(TrimDataPacket::create($patterns, $materials));
```
dktapps: Is this customizable? If so, why are we using hard enums for this?

src/network/mcpe/cache/CraftingDataCache.php
```diff
@@ -141,6 +145,33 @@ private function buildCraftingDataCache(CraftingManager $manager) : CraftingData
 			}
 		}
 
+		$index ??= 0;
```
dktapps: This looks suspicious.

src/network/mcpe/handler/ItemStackRequestExecutor.php
```diff
@@ -342,6 +344,9 @@ protected function processItemStackRequestAction(ItemStackRequestAction $action)
 					$this->specialTransaction = new EnchantingTransaction($this->player, $option, $optionId + 1);
 					$this->setNextCreatedItem($window->getOutput($optionId));
 				}
+			}elseif($window instanceof SmithingTableInventory){
+				$this->specialTransaction = new SmithingTransaction($this->player);
```
dktapps: Why isn't the recipe ID used here?

src/crafting/CraftingManager.php
```diff
@@ -304,4 +326,13 @@ public function matchBrewingRecipe(Item $input, Item $ingredient) : ?BrewingReci
 
 		return null;
 	}
+
+	public function matchSmithingRecipe(Item $input, Item $addition, Item $template) : ?SmithingRecipe{
```
dktapps: This function should not be necessary. The ItemStackRequest provides an ID which tells us the recipe selected.

dktapps (changes requested):

src/inventory/transaction/SmithingTransaction.php
HimmelKreis4865: what should I replace it with?

dktapps: Literally anything else. The numeric values can change without warning.

src/crafting/CraftingManager.php
dktapps: (unless you want to allow SmithingTransaction to auto-detect the recipe for plugins, in which case take a look at how CraftingTransaction does this)

src/block/inventory/SmithingTableInventory.php
dktapps: My comments from Discord: 
> just delete jt
if the handling was implemented correctly it would be unused anyway
look at how the CraftingTransaction works in ItemStackRequestExecutor
also, keep conversations in one place, it becomes a confusing mess of back-and-forth otherwise

src/network/mcpe/cache/CraftingDataCache.php
HimmelKreis4865: it's there to prevent errors in case no crafting recipe gets registered. the indexes are incrementing and are not reseted when switching from crafting to smithing

src/item/ArmorTrimPattern.php
HimmelKreis4865: I wouldn't say a map is a good idea either, is it?

src/item/Armor.php
HimmelKreis4865: I tried it, but I don't understand where the translation between the save-id to the real name happens...Trims are also invisible after reconnect, but I'm not able to add the same trim on one item again. I will take a few more looks into this, otherwise I will commit everything for you to fix it

HimmelKreis4865:
There's not much more I can do at this point, everthing is fixed except for save ids.

src/item/Armor.php
```diff
@@ -40,8 +42,16 @@ class Armor extends Durable{
 
 	public const TAG_CUSTOM_COLOR = "customColor"; //TAG_Int
 
+	public const TAG_TRIM = "Trim";
```
jasonw4331: please add the tag type comment for consistency

jasonw4331 (changes requested):
Looks pretty good to me, one comment.

src/crafting/CraftingManager.php
```diff
@@ -197,6 +204,23 @@ public function getPotionContainerChangeRecipes() : array{
...
+	/**
+	 * @return SmithingRecipe[]
+	 * @phpstan-return list<SmithingRecipe>
+	 */
+	public function getSmithingRecipes() : array{
+		return $this->smithingRecipes;
+	}
+
+	public function getSmithingRecipeFromIndex(int $index) : ?SmithingRecipe{
+		$craftingOffset = count($this->craftingRecipeIndex) - 1;
```
dktapps: This doesn't look reliable. If new crafting recipes are registered, the offset will change.

src/block/SmithingTable.php
```diff
@@ -31,10 +31,7 @@
 final class SmithingTable extends Opaque{
 
 	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
-		if($player !== null){
-			$player->setCurrentWindow(new SmithingTableInventory($this->position));
-		}
-
+		$player?->setCurrentWindow(new SmithingTableInventory($this->position));
```
dktapps: Unrelated change

src/item/ArmorTrimPattern.php
dktapps: It's better than this hack

src/crafting/CraftingManager.php
HimmelKreis4865: Smithing recipes are still behind crafting recipes, the relative position of the smithing recipes will not change. Also all caches are regenerated once a new recipe is registered, so I dont get your point.

src/block/SmithingTable.php
HimmelKreis4865: is this bad though?

src/item/Armor.php
HimmelKreis4865: Alright, I removed it because this is very rare in the code, but I can add it back

src/block/SmithingTable.php
dktapps: It's unrelated to the PR objective, so it shouldn't be changed.

src/crafting/CraftingManager.php
dktapps: If a recipe is registered after a player joins, the recipe IDs they send back to the server won't match up with this.

HimmelKreis4865: yeah it seems possible, what should I do though? 🤷🏼

src/network/mcpe/handler/ItemStackRequestExecutor.php
```diff
@@ -342,6 +345,13 @@ protected function processItemStackRequestAction(ItemStackRequestAction $action)
 					$this->specialTransaction = new EnchantingTransaction($this->player, $option, $optionId + 1);
 					$this->setNextCreatedItem($window->getOutput($optionId));
 				}
+			}elseif($window instanceof SmithingTableInventory){
+				$craftingManager = Server::getInstance()->getCraftingManager();
```
dktapps:
```suggestion
				$craftingManager = $this->player->getServer()->getCraftingManager();
```

src/crafting/CraftingManager.php
pandaaaBE: > If a recipe is registered after a player joins, the recipe IDs they send back to the server won't match up with this.

Which plugins add recipes after the join? but anyway, how would you solve that? @dktapps

dktapps: Not sure tbh. It seems stupid to me that these recipes share the same ID space with crafting recipes, since they can't be crafted in a crafting table. However, I expect that this is required by the client.

Perhaps a bitflag on the recipe ID would work, or a large fixed offset.

dktapps: We had a similar problem with enchanting options. An offset of 100,000 was used to prevent conflicts there. To use the same approach here you'd need to go with 200k or something like that.

We'll need a better solution for this in the longer term (particularly since this problem will likely crop up again with other recipe types), but that's a problem for another PR.

fractalllll:
how do I put this on my server?

pandaaaBE:
image
Looks finished now :)

HimmelKreis4865:
yeah, ready to be reviewed again.

src/block/SmithingTable.php
```diff
@@ -34,7 +34,6 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
 		if($player !== null){
 			$player->setCurrentWindow(new SmithingTableInventory($this->position));
 		}
-
```
ShockedPlot7560: Unrelated change

src/crafting/SmithingTransformRecipe.php
```diff
@@ -0,0 +1,73 @@
...
+
+	public function getResult() : Item{
+		return clone $this->result;
+	}
+
+	/**
+	 * @param Item[] $inputs
+	 * @phpstan-param list<Item> $inputs
+	 */
+	public function getResultFor(array $inputs) : ?Item{
```
ShockedPlot7560, line 58: I'm sceptical about the fact that it's up to the recipe to find out what the input is.

src/crafting/SmithingTrimRecipe.php
```diff
@@ -0,0 +1,77 @@
...
+
+	public function getTemplate() : RecipeIngredient{
+		return $this->template;
+	}
+
+	/**
+	 * @param Item[] $inputs
+	 * @phpstan-param list<Item> $inputs
+	 */
+	public function getResultFor(array $inputs) : ?Item{
```
ShockedPlot7560, line 55: Same here

src/crafting/SmithingTransformRecipe.php
```diff
@@ -0,0 +1,73 @@
...
+		$input = null;
+		foreach($inputs as $item){
+			if ($item instanceof Armor || $item instanceof TieredTool){
+				$input = $item;
+			}
+		}
+		if($input === null){
+			return null;
+		}
+		return $this->getResult()->setNamedTag($input->getNamedTag());
```
ShockedPlot7560: When we transform to netherite, all enchantments and lost point of durability is keeping. Here, we lost

src/item/ArmorTrim.php
```diff
@@ -0,0 +1,40 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\item;
+
+class ArmorTrim{
+
+	public function __construct(
+		private readonly ArmorTrimMaterial $material,
+		private readonly ArmorTrimPattern $pattern
```
ShockedPlot7560, line 30: Whats the purpose of setting this readonly when there isnt setter or public visibility ?

src/item/Armor.php
```diff
@@ -106,6 +114,22 @@ public function clearCustomColor() : self{
...
+	}
+
+	/** @return $this */
+	public function setTrim(ArmorTrim $trim) : self{
+		$this->armorTrim = $trim;
+		return $this;
+	}
+
+	/** @return $this */
+	public function clearTrim() : self{
```
ShockedPlot7560: Permit to set a null value instead of create a new function just for this

src/item/SmithingTemplate.php
```diff
@@ -0,0 +1,28 @@
...
+ * @link http://www.pocketmine.net/
+ *
+ *
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\item;
+
+class SmithingTemplate extends Item{
```
ShockedPlot7560: Not needed if its empty

src/utils/TextFormat.php
ShockedPlot7560, line 1: Unrelated changes

src/crafting/SmithingTransformRecipe.php
ipad54: Actually, enchantments and durability are stored in NBT. So, this is correct

ShockedPlot7560: Right

src/item/SmithingTemplate.php
ipad54: iirc it was made to determine the smithing template from other inputs in transaction

ShockedPlot7560: Use TypeId ?

ipad54: > Use TypeId ?

https://github.com/pmmp/PocketMine-MP/pull/6202#discussion_r1432552692

ShockedPlot7560: I don't think creating a new empty class just for that is viable. Another way can be done, i'm sure.

src/utils/TextFormat.php
HimmelKreis4865: what exactly do you mean?

src/item/SmithingTemplate.php
HimmelKreis4865: if you propose one, I'd be happy aswell

src/item/ArmorTrim.php
HimmelKreis4865: I agree it has no use case, iirc phpstorm proposed it, should I remove it?

src/utils/TextFormat.php
ShockedPlot7560: Its already in progess in a PR : https://github.com/pmmp/PocketMine-MP/pull/5838

src/item/SmithingTemplate.php
ipad54: I think you can use ```match()``` or ```ArmorTrimPattern::fromItem()```. What do you think about that @ShockedPlot7560?

ShockedPlot7560: fromItem could be a solution, however using gameplay API for some network validation, i dont know

src/utils/TextFormat.php
HimmelKreis4865: okay, but I do need them in `ArmorTrimMaterial` for the registration.

ShockedPlot7560: Could we assume we skip ArmorTrim colors ? Is it used anymore than just exposing API ?

HimmelKreis4865: No, I dont think so, take a look at `network/mcpe/handler/PreSpawnPacketHandler.php`, I need to send the colors at this point

ShockedPlot7560: It seems that the PR for the new color is bloqued because of the underline and other formatting format which it was removed. I guess we could keep the color here

src/network/mcpe/handler/PreSpawnPacketHandler.php
ipad54: @HimmelKreis4865 I don't know about patterns, but colours can be definitely customized. (For example, I can change the amethyst colour from purple to white). What do you think about allowing plugins to customize this?
Screenshot_393

src/crafting/SmithingTransformRecipe.php
HimmelKreis4865: tbh I don't see a problem here, it also has 2 usages so I don't know how I'd change it smoothly..

src/network/mcpe/handler/PreSpawnPacketHandler.php
HimmelKreis4865: yeah I'm aware of this, and this should be customizable but I did not yet find a solution for this problem that satisfies me

src/crafting/SmithingTrimRecipe.php
```diff
@@ -0,0 +1,77 @@
...
+			}
+		}
+
+		if($input === null || $addition === null || $template === null){
+			return null;
+		}
+		if(($material = ArmorTrimMaterial::fromItem($addition)) === null || ($pattern = ArmorTrimPattern::fromItem($template)) === null){
+			return null;
+		}
+		return $input->setTrim(new ArmorTrim($material, $pattern));
```
ipad54: Input should be cloned

HimmelKreis4865: they are already cloned in both usecases

ipad54: Right. Sorry, I forgot that ```getContents()``` also clones items.

src/network/mcpe/handler/PreSpawnPacketHandler.php
HimmelKreis4865: I pushed a commit to make this all customizable, please tell me what you think of this

src/utils/TextFormat.php
```diff
@@ -63,6 +63,16 @@ abstract class TextFormat{
...
+	public const MATERIAL_QUARTZ = TextFormat::ESCAPE . "h";
+	public const MATERIAL_IRON = TextFormat::ESCAPE . "i";
+	public const MATERIAL_NETHERITE = TextFormat::ESCAPE . "j";
+	public const MATERIAL_REDSTONE = TextFormat::ESCAPE . "m";
+	public const MATERIAL_COPPER = TextFormat::ESCAPE . "n";
+	public const MATERIAL_GOLD = TextFormat::ESCAPE . "p";
+	public const MATERIAL_EMERALD = TextFormat::ESCAPE . "q";
+	public const MATERIAL_DIAMOND = TextFormat::ESCAPE . "s";
+	public const MATERIAL_LAPIS = TextFormat::ESCAPE . "t";
+	public const MATERIAL_AMETHYST = TextFormat::ESCAPE . "u";
```
diamond-gold, line 75: Currently these formatting codes will not work in the terminal, not sure if that will be an issue for this PR
Maybe these should be added in a separate PR?
See #5775,#5793

HimmelKreis4865: yeah I think this is probably something for another pr

src/network/mcpe/handler/PreSpawnPacketHandler.php
HimmelKreis4865: > To me it looks good, but I'm not sure about passing type id and serialized item name to the constructor. Wouldn't a better to pass an Item and then serialize it when sending TrimData?
(from ipad54 on discord)

what do you guys think about this?

src/crafting/SmithingTransformRecipe.php
HimmelKreis4865: is this relevant now? If so, please give me some more information on where to move the methods

HimmelKreis4865:
what is still missing?

pandaaaBE:
> what is still missing?

nothing. just reviews 😭

src/crafting/SmithingTrimRecipe.php
```diff
@@ -0,0 +1,77 @@
...
+use pocketmine\item\ArmorTrim;
+use pocketmine\item\ArmorTrimRegistry;
+use pocketmine\item\Item;
+
+class SmithingTrimRecipe implements SmithingRecipe{
+
+	public function __construct(
+		private readonly RecipeIngredient $input,
+		private readonly RecipeIngredient $addition,
+		private readonly RecipeIngredient $template){
```
jasonw4331: Because these are `readonly` values the getter functions are redundant and these variables can be made public

src/item/ArmorTrimPattern.php
```diff
@@ -0,0 +1,62 @@
...
+	public const TIDE = "tide";
+	public const VEX = "vex";
+	public const WARD = "ward";
+	public const WAYFINDER = "wayfinder";
+	public const WILD = "wild";
+
+	public function __construct(
+		private string $identifier,
+		private string $itemName,
+		private int $typeId
```
jasonw4331: These can also be public, making the getters redundant.

src/item/ArmorTrimMaterial.php
```diff
@@ -0,0 +1,61 @@
...
+	public const LAPIS = "lapis";
+	public const NETHERITE = "netherite";
+	public const QUARTZ = "quartz";
+	public const REDSTONE = "redstone";
+
+	public function __construct(
+		private string $identifier,
+		private string $color,
+		private string $itemName,
+		private int $typeId
```
jasonw4331: These can also be public, making the getters redundant.

jasonw4331 (commented):
Here are some micro-optimizations which I believe may help the API

src/item/ArmorTrimRegistry.php
```diff
@@ -0,0 +1,142 @@
...
+	 */
+	private array $idToPatternMap = [];
+
+	public function __construct(){
+		$this->registerDefaultMaterials();
+		$this->registerDefaultPatterns();
+	}
+
+	private function registerDefaultMaterials() : void{
+		$this->registerMaterial(new ArmorTrimMaterial(ArmorTrimMaterial::AMETHYST, TextFormat::MATERIAL_AMETHYST, Names::AMETHYST_SHARD, Ids::AMETHYST_SHARD));
```
ipad54: As I said in discord, it would be better to pass an Item and then serialise it when sending TrimData instead of passing type id and serialised item name

dktapps: Agreed.

I also don't like that this class is the only way to get an `ArmorTrimMaterial` instance or `ArmorTrimPattern`. Feels very `ItemFactory`-esque in that it forces the use of ID constants and obnoxiously long call syntax that encourages plugin devs to hardcode stuff.

`ArmorTrimRegistry::getInstance()->getMaterial(ArmorTrimMaterial::COPPER)`
could be more like
`VanillaArmorTrimMaterials::COPPER()`

The job of classes like this should be to load & save data from storage, not to be used as APIs. It should be in the `pocketmine\data\item` package.

src/crafting/SmithingTransformRecipe.php
dktapps: It's fairly consistent with how recipes are generally expected to work. Wouldn't be any different for the likes of shulker box recipes.

src/crafting/SmithingTransformRecipe.php
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	/**
+	 * @param Item[] $inputs
+	 * @phpstan-param list<Item> $inputs
+	 */
+	public function getResultFor(array $inputs) : ?Item{
+		$input = null;
+		foreach($inputs as $item){
+			if ($item instanceof Armor || $item instanceof TieredTool){
```
dktapps: Probably better to use [`$this->input->accepts($item)`](https://github.com/pmmp/PocketMine-MP/blob/cd2a1fdc1d2c327894b3b489937e9d6abb34e472/src/crafting/RecipeIngredient.php#L30) rather than the instanceof checks.

src/crafting/SmithingTrimRecipe.php
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	/**
+	 * @param Item[] $inputs
+	 * @phpstan-param list<Item> $inputs
+	 */
+	public function getResultFor(array $inputs) : ?Item{
+		$input = $template = $addition = null;
+		foreach($inputs as $item){
+			if($item instanceof Armor){
```
dktapps: Again, use `RecipeIngredient->accepts()` instead of doing this

src/crafting/SmithingTrimRecipe.php
```diff
@@ -0,0 +1,77 @@
...
+				$input = $item;
+			}elseif(ArmorTrimRegistry::getInstance()->getPatternFromItem($item) !== null){
```
dktapps: Probably same again here

src/inventory/transaction/SmithingTransaction.php
```diff
@@ -0,0 +1,65 @@
...
+		$this->matchItems($outputs, $inputs);
+
+		if(($inputCount = count($inputs)) !== 3){
+			throw new TransactionValidationException("Expected 3 input items, got $inputCount");
+		}
+		if(($outputCount = count($outputs)) !== 1){
+			throw new TransactionValidationException("Expected 1 output item, but received $outputCount");
+		}
+		if(($output = $this->recipe->getResultFor($inputs)) === null){
+			throw new TransactionValidationException("Could find a matching output item for the given inputs");
```
dktapps:
```suggestion
			throw new TransactionValidationException("Couldn't find a matching output item for the given inputs");
```

src/item/ArmorTrimMaterial.php
dktapps: That would be inconsistent with PM's general code style.

src/network/mcpe/handler/PreSpawnPacketHandler.php
dktapps: See above.

src/data/bedrock/ArmorTrimPatternTypeIdMap.php
```diff
@@ -0,0 +1,107 @@
...
+	/**
+	 * @var string[]
+	 * @phpstan-var array<int, string>
+	 */
+	private array $patternToId = [];
+
+	public function __construct(){
+		foreach(Patterns::getAll() as $case){
+			$this->register(match($case){
+				Patterns::COAST() => Ids::COAST,
```
dktapps, line 58: Matching on this is unstable. The pattern is cloneable & serializable.

dktapps (changes requested):

src/data/bedrock/ArmorTrimPatternTypeIdMap.php
dktapps: There's also not much point to using this match style for registering these since there's no static analysis benefit if not using an enum.

src/data/bedrock/ArmorTrimMaterialTypeIdMap.php
```diff
@@ -53,21 +53,16 @@ final class ArmorTrimMaterialTypeIdMap{
...
+		$this->register(Ids::AMETHYST, Materials::AMETHYST());
+		$this->register(Ids::COPPER, Materials::COPPER());
+		$this->register(Ids::DIAMOND, Materials::DIAMOND());
+		$this->register(Ids::EMERALD, Materials::EMERALD());
+		$this->register(Ids::GOLD, Materials::GOLD());
+		$this->register(Ids::IRON, Materials::IRON());
+		$this->register(Ids::LAPIS, Materials::LAPIS());
+		$this->register(Ids::NETHERITE, Materials::NETHERITE());
+		$this->register(Ids::QUARTZ, Materials::QUARTZ());
+		$this->register(Ids::REDSTONE, Materials::REDSTONE());
```
dktapps: Sorry to go back and forth on this but I changed my mind about the match. I realized it'll still be useful to detect errors even if PHPStan doesn't see them.

src/crafting/CraftingManager.php
```diff
@@ -29,12 +29,15 @@
...
 use pocketmine\utils\DestructorCallbackTrait;
 use pocketmine\utils\ObjectSet;
+use function count;
 use function spl_object_id;
 use function usort;
 
 class CraftingManager{
 	use DestructorCallbackTrait;
 
+	public const SMITHING_RECIPES_OFFSET = 200000;
```
dktapps: I think I mentioned before that I don't like this being here. It should be managed in the network layer.

src/block/inventory/SmithingTableInventory.php
```diff
@@ -25,13 +25,30 @@
...
+		return $this->getItem(self::SLOT_INPUT);
+	}
+
+	public function getAddition() : Item{
+		return $this->getItem(self::SLOT_ADDITION);
+	}
+
+	public function getTemplate() : Item{
+		return $this->getItem(self::SLOT_TEMPLATE);
+	}
```
dktapps: Not much point adding wrapper functions for these. It's a stupid convention. Constants are enough.

src/data/bedrock/ArmorTrimMaterialTypeIdMap.php
```diff
@@ -0,0 +1,101 @@
...
+	public function fromId(string $id) : ?ArmorTrimMaterial{
+		return $this->idToMaterial[$id] ?? null;
+	}
+
+	public function fromItem(Item $item) : ?ArmorTrimMaterial{
+		return $this->itemToMaterial[$item->getStateId()] ?? null;
+	}
+
+	public function toId(ArmorTrimMaterial $material) : string{
+		$k = spl_object_id($material);
```
dktapps, line 89: `spl_object_id()` is unstable for this, since `ArmorTrimMaterial` is cloneable & serializable

src/data/bedrock/ArmorTrimPatternTypeIdMap.php
```diff
@@ -0,0 +1,107 @@
...
+	public function fromId(string $id) : ?ArmorTrimPattern{
+		return $this->idToPattern[$id] ?? null;
+	}
+
+	public function fromItem(Item $item) : ?ArmorTrimPattern{
+		return $this->itemToPattern[$item->getStateId()] ?? null;
+	}
+
+	public function toId(ArmorTrimPattern $pattern) : string{
+		$k = spl_object_id($pattern);
```
dktapps, line 94: `spl_object_id()` is unstable for this, since `ArmorTrimMaterial` is cloneable & serializable

src/inventory/transaction/SmithingTransaction.php
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	public function validate() : void{
+		if(count($this->actions) < 1){
+			throw new TransactionValidationException("Transaction must have at least one action to be executable");
+		}
+
+		/** @var Item[] $inputs */
+		$inputs = [];
+		/** @var Item[] $outputs */
```
dktapps: These doc comments are not necessary

dktapps (changes requested):
Looks mostly OK.

I'm a bit uneasy about allowing custom armor trim patterns because they'll generate confusing errors if the objects are duplicated. The `ArmorTrimPattern` and `ArmorTrimMaterial` classes should use the `NotCloneable` and `NotSerializable` traits.

src/network/mcpe/convert/TypeConverter.php
```diff
@@ -56,6 +56,10 @@
 class TypeConverter{
 	use SingletonTrait;
 
+	//TODO: Hack! In Bedrock smithing recipe indexes go together with regular recipes,
+	//so we make an offset to prevent conflicts.
+	public const SMITHING_RECIPES_NETWORK_OFFSET = 200000;
```
dktapps: It would be better if this was declared together with the other offset constants so that it's clear which ones have already been used. (That being said, the 100k for enchanting options is currently hardcoded in `InventoryManager`, so make of that what you will.)

kostamax27:
image

```log
[12:57:00.529] [Server thread/DEBUG]: [NetworkSession: kostamax27] Invalid inventory transaction -35: Expected an exact multiple of required Item Netherite Upgrade Smithing Template (20269:0)x2 (given: 1, needed: 2)
[12:57:00.529] [Server thread/DEBUG]: [NetworkSession: kostamax27] Detected prediction mismatch in inventory pocketmine\inventory\PlayerCursorInventory#214179 slot 0
[12:57:00.529] [Server thread/DEBUG]: [NetworkSession: kostamax27] Detected prediction mismatch in inventory pocketmine\block\inventory\CraftingTableInventory#88462 slot 0
[12:57:00.529] [Server thread/DEBUG]: [NetworkSession: kostamax27] Detected prediction mismatch in inventory pocketmine\block\inventory\CraftingTableInventory#88462 slot 1
[12:57:00.529] [Server thread/DEBUG]: [NetworkSession: kostamax27] Detected prediction mismatch in inventory pocketmine\block\inventory\CraftingTableInventory#88462 slot 2
[12:57:00.529] [Server thread/DEBUG]: [NetworkSession: kostamax27] Detected prediction mismatch in inventory pocketmine\block\inventory\CraftingTableInventory#88462 slot 3
[12:57:00.529] [Server thread/DEBUG]: [NetworkSession: kostamax27] Detected prediction mismatch in inventory pocketmine\block\inventory\CraftingTableInventory#88462 slot 4
[12:57:00.529] [Server thread/DEBUG]: [NetworkSession: kostamax27] Detected prediction mismatch in inventory pocketmine\block\inventory\CraftingTableInventory#88462 slot 5
[12:57:00.529] [Server thread/DEBUG]: [NetworkSession: kostamax27] Detected prediction mismatch in inventory pocketmine\block\inventory\CraftingTableInventory#88462 slot 6
[12:57:00.529] [Server thread/DEBUG]: [NetworkSession: kostamax27] Detected prediction mismatch in inventory pocketmine\block\inventory\CraftingTableInventory#88462 slot 7
[12:57:00.529] [Server thread/DEBUG]: [NetworkSession: kostamax27] Detected prediction mismatch in inventory pocketmine\block\inventory\CraftingTableInventory#88462 slot 8
[12:57:00.560] [Server thread/DEBUG]: [NetworkSession: kostamax27] Full inventory sync requested, sending contents of 6 inventories
```

dktapps:
For posterity: I think this got stalled because I wasn't happy about using `RegistryTrait` for armor patterns and trims.
My issue with that is mostly around wanting to use enums wherever possible, but we run into #6292. I don't feel good about introducing inconsistent APIs around this stuff, but no resolution was reached in internal discussions about this.

Will revisit this soon :tm:

src/crafting/SmithingTrimRecipe.php
```diff
@@ -0,0 +1,76 @@
...
+				$input = $item;
+			}elseif($this->addition->accepts($item)){
+				$addition = $item;
+			}elseif($this->template->accepts($item)){
+				$template = $item;
+			}
+		}
+		if($input !== null && $addition !== null && $template !== null){
+			$material = ArmorTrimMaterialTypeIdMap::getInstance()->fromItem($addition);
+			$pattern = ArmorTrimPatternTypeIdMap::getInstance()->fromItem($template);
```
dktapps, line 68: Really don't like this. `pocketmine\data` namespace is explicitly internal.
Type ID mapping is part of an internal package because plugins aren't supposed to deal with that stuff, but plugins should be able to get the trim color of an item without using internal namespaces.
Likewise, the pattern should be obtainable without using internal namespaces.

dktapps:
So, the issue with the registries makes more sense to me now.
Basically, the problem is that the patterns and materials are registered by object ID as if they were enums, but it's extremely easy for plugin devs to just do `setTrim(new ArmorTrim(new ArmorTrimMaterial(VanillaItems::SOMETHING()), ...))` and then proceed to have the server crash later when it tries to save data, because that particular trim material object's identity was never associated with a material ID.

When we use enums, it's easy to guarantee that all enum cases are saveable since there's never going to be others that weren't defined at compile time. This is not true when using registries, which makes the API here a series of footguns which will lead to confusing crashes.

This is linked to a broader issue described in #6654

PR #6204, opened 2023-12-20
Implement blocks ticking per world API

### LINKED ISSUE #6173: API for random tick speed
## Description
We already support something similar to random tick speed via pocketmine.yml (blocks-per-subchunk-per-tick) but it's not granular enough:
- it applies to the whole server
- can't be changed at runtime

An API on World to allow changing this would be beneficial for plugins.

ShockedPlot7560:
## Introduction
We already support to change the random tick speed in server, however we can't control which world or change it at runtime. This PR fix that by providing an API and adding new configuration key.

### Relevant issues
* Fixes #6173 

### API changes
Adding the following functions to `World`:
- `setTickedBlocksPerSubchunkPerTick`
- `getTickedBlocksPerSubchunkPerTick`

Adding a new event `WorldTickedBlocksChangeEvent` which is fired when the value of tickedBlocksPerSubchunkPerTick change (not at the initialisation).

## Backwards compatibility
No

## Follow-up
Change the configuration nested the worlds per-tick and the default value under a `blocks-per-subchunk-per-tick`


resources/pocketmine.yml
```diff
@@ -127,6 +127,9 @@ chunk-ticking:
   #Number of blocks inside ticking areas' subchunks that get ticked every tick. Higher values will accelerate events
   #like tree and plant growth, but at a higher performance cost.
   blocks-per-subchunk-per-tick: 3
+  blocks-per-subchunk-per-world-per-tick:
```
dktapps: I think this would make more sense in the `worlds` section

src/event/world/WorldTickedBlocksChangeEvent.php
```diff
@@ -0,0 +1,44 @@
...
+ *
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\event\world;
+
+use pocketmine\world\World;
+
+class WorldTickedBlocksChangeEvent extends WorldEvent{
```
dktapps: Not sure why we need an event for this

dktapps: The name of the event is also very ambiguous, and the API method names don't help either, nor does the lack of documentation

CicciogamerRrr:
Stop spamming

Il mer 20 dic 2023, 12:27 Dylan T. ***@***.***> ha scritto:

src/event/world/WorldTickedBlocksChangeEvent.php
ShockedPlot7560: I was also wondering, would there be any use for it?

dktapps: I can't think of any myself.

resources/pocketmine.yml
```diff
@@ -215,6 +215,7 @@ worlds:
   # seed: 404
   # generator: FLAT
   # preset: 2;bedrock,59xstone,3xdirt,grass;1
+  # blocks-per-subchunk-per-tick: xx
```
dktapps:
```suggestion
  # blocks-per-subchunk-per-tick: 3
```

dktapps (commented):
I feel like the method names are too esoteric for anyone to ever actually use them. Can we simplify these names at all without losing the meaning?

ShockedPlot7560:
> Can we simplify these names at all without losing the meaning?

The only names i see would be: 
- `initTickedBlocksPerSubchunkPerTick` → `initSubchunkTicks`
- `setTickedBlocksPerSubchunkPerTick` → `setSubchunkTicks`
- `getTickedBlocksPerSubchunkPerTick` → `getSubchunkTicks`

But we will need any docs to precise this is the number of block ticked inside each subchunk each tick. But for now, I don't see any shorter name.

dktapps:
I feel like those are too ambiguous too.

ShockedPlot7560:
I can't think of any other name without losing the original meaning of the function.

dktapps:
Perhaps the issue here is that we're exposing something that is too "internal". "Random tick speed" would be the most intuitive name, but the problem is that, because of its usage, the values don't map onto what users would expect.

We could expose a random tick speed that is just this value divided by 3, but we'd probably have to make it a decimal value to allow the level of granularity that the current setting has. This then exposes all kinds of edge cases like what happens if somebody sets a tick speed of 1.5. We can't tick 4.5 blocks per tick without a lot more complexity.

ShockedPlot7560:
Come to think of it, isn't it possible to add a warning in the console if the final value is a decimal, saying that it has been rounded up or down?

dktapps:
No one reads warnings, and even if they do, they won't know why it's being emitted if a plugin triggers it.

PR #6217, opened 2024-01-04
PocketMine-MP.phar self-extraction to decompressed cache

### LINKED ISSUE #6214: Custom `phar://` stream wrapper to prevent `/tmp` wastage
## Description
`ext-phar` decompresses compressed phar contents into temporary files in `/tmp`. This is problematic because the cached temporary file is not shared between different threads, leading to several MB of wasted `/tmp` footprint per thread started.

While this is not a unique problem to PM, it should be possible to work around this with a custom stream wrapper that either doesn't cache the files at all (the cache is only really useful when including the same file multiple times without opcache, or when loading a data file directly), or that caches them in a more thread-friendly manner.

## Justification
Recently users of Pterodactyl hosts have experienced issues with tmp space being filled due to this issue.

dktapps:
Because ext-phar sucks, tmp gets spammed by cache files for every thread when loading files from the phar on the fly.

Instead, we convert the `.phar` into a decompressed `.tar` in the tmp directory and require files from inside it. Surprisingly, this works because `ext-phar` supports `tar` and `zip` natively. No stream wrapper is required, as the `PocketMine.php` bootstrap loads files relative to its location, so the cache is automatically used for everything.

To be honest I would rather get rid of phars entirely, but they are still the easiest way to have PhpStorm load PocketMine-MP API information for now, and the alternatives are more complicated and inconvenient.

### Caveats
Everywhere that previously used `new Phar(Phar::running(false))` in the core code needs to be updated to use `PharData` for this to work correctly. Plugins don't need to do anything.

### Why not just use `Phar::decompressFiles()`?
This requires setting `phar.readonly` to `0`, which is a security issue. Technically, we could have used a subprocess to do this, but it just didn't seem right.

### WTF? `phar://` can be used on `tar` files???
Yup. I was just as surprised to find out that `require` works in such contexts.

### Relevant issues
- Closes #6214 

### API changes
None.

### Behavioural changes
Server startup will be slightly slower, as the phar has to decompress and convert itself into a `.tar`. However, testing showed that this generally takes less than 200 ms, so it should be barely noticeable.

## Backwards compatibility
No BC issues.

## Tests
Locally tested and the CI will also verify


SenseiTarzan:
I have a client who had 16 GB in tmp which came from pmmp because when you go into it he had a whole file which starts with php and with pmmp inside it decompresses into a single file

SenseiTarzan:
I have a client who had 10 GB in tmp which came from pmmp because when you go into it he had a whole file which starts with php and with pmmp inside it decompresses into a single file XD

src/PocketMine.php
```diff
@@ -175,9 +175,6 @@ function emit_performance_warnings(\Logger $logger){
 		if(((int) ini_get('zend.assertions')) !== -1){
 			$logger->warning("Debugging assertions are enabled. This may degrade performance. To disable them, set `zend.assertions = -1` in php.ini.");
 		}
-		if(\Phar::running(true) === ""){
-			$logger->warning("Non-packaged installation detected. This will degrade autoloading speed and make startup times longer.");
```
SOF3: rip phar

build/server-phar.php
```diff
@@ -159,21 +161,7 @@ function main() : void{
...
-	echo "Check that the directory exists, and that the current user has read/write permissions for it." . PHP_EOL;
-	echo "Alternatively, set 'sys_temp_dir' to a different directory in your php.ini file." . PHP_EOL;
-	exit(1);
-}
-
-require("phar://" . __FILE__ . "/src/PocketMine.php");
-__HALT_COMPILER();
-STUB
-,
+		Filesystem::fileGetContents(Path::join(__DIR__, 'server-phar-stub.php')) . "\n__HALT_COMPILER();",
```
SOF3, line 174: why not write this inside server-phar-stub.php?

build/server-phar-stub.php
```diff
@@ -0,0 +1,73 @@
...
+	if($prepare){
+		$phar = new \Phar(__FILE__);
+		echo "Preparing source cache..." . PHP_EOL;
+		$phar->extractTo($sourceDir);
+		file_put_contents($sourceDir . DIRECTORY_SEPARATOR . "phar-hash.txt", $expectedHash);
+		echo "Done!" . PHP_EOL;
+	}
+}
+
+$sourceDir = getcwd() . "/.PocketMine-MP.phar/";
```
SOF3: Why don't we use tempdir here? Classes are still loaded from the same tempdir since we won't be including the phar again in threads. You can copy with mode 0o400 if you are concerned about other users tampering the file.

build/server-phar-stub.php
```diff
@@ -0,0 +1,73 @@
...
+			echo "Removing old source cache..." . PHP_EOL;
+			recursiveUnlink($sourceDir);
+			echo "Done!" . PHP_EOL;
+		}
+	}
+
+	if($prepare){
+		$phar = new \Phar(__FILE__);
+		echo "Preparing source cache..." . PHP_EOL;
+		$phar->extractTo($sourceDir);
```
SOF3: actually we could chmod(400) all files inside the phar too so that users don't try to modify this file https://www.php.net/manual/en/pharfileinfo.chmod.php

build/server-phar.php
dktapps: No particular reason, I just felt it looked weird to have __HALT_COMPILER in a script file
I also don't know what effects trailing newlines might have

build/server-phar-stub.php
dktapps: Multiple servers might want to extract different versions of the server at once. Having the cache locally avoids the need to add discriminators to separate them

src/PocketMine.php
dktapps: To be honest I think this was always a lie. Decompressing the phar contents negates any gain had by packing everything into one file. The only way this warning would have been valid is if the phar had zero compression.

dries-c: It was probably manufactured during the time where releases only had function imports.

dktapps: I'd forgotten the phar releases used to be preprocessed. You might be right about that.

build/server-phar-stub.php
SOF3: I am mainly concerned that users are under the impression they can directly modify the source code here.

src/PocketMine.php
SOF3: I thought windows had better performance loading from a single archive.

build/server-phar-stub.php
dktapps: Yeah, the same thing occurred to me, though I'm not sure how much we need to worry about this.

dktapps: I'll also mention it's a pretty big pain to reliably differentiate these if we start going into /tmp. We'd need to derive the tmp discriminator from the PocketMine-MP.phar path to ensure reuse and proper cleanup on crash.

src/PocketMine.php
dktapps: Like I said, the I/O overhead is offset by decompression overhead. There's little observable difference in practice.

On *nix src is pretty much always going to be faster, and that's where the majority of the userbase is.

dktapps:
My main sticking point with this change is extracting source files into a directory tree. This makes for a cache that's slow to clean up and validate. It would be better if we could have a single-file cache similar to ext-phar for this purpose (but obviously one that can be shared by multiple threads).

The only ways I can see to do this are:
- Create a second temporary phar with all files decompressed - problematic because of `phar.readonly`
- Add a stream wrapper for `phar://` to access a custom cache file, overriding `ext-phar`'s shitty cache mechanism - potentially problematic because it's difficult to ensure that all threads have the stream wrapper registered

I'm wondering if we could do something with Composer bootstrap files to auto-register phar stream wrappers to do this. If we can find a way to piggyback onto the autoloader, the threading problems would go away.

dktapps:
I managed to find a simpler way to do this with a single file, allowing us to avoid potentially fragile cache reuse across multiple runs.

Essentially, we create a .tar from the .phar using `Phar::convertToData()` with `compression: Phar::NONE` (which isn't restricted by `phar.readonly`), and then `require` files from inside the .tar using the `phar://` wrapper (yes, this actually works).

This approach requires changing any places that used `new Phar(Phar::running(false))` as they'll need to use `PharData` if the server is running from a cache.

build/server-phar-stub.php
dktapps: No longer applicable with latest revision

build/server-phar-stub.php
dktapps: No longer applicable on latest revision

build/server-phar-stub.php
```diff
@@ -0,0 +1,72 @@
...
+	return [$tmp . '.tar', $tmp];
+}
+
+echo "Preparing PocketMine-MP.phar decompressed cache...\n";
+$start = hrtime(true);
+[$cacheName, $tmpReserveName] = preparePharCache();
+echo "Cache ready at $cacheName in " . number_format((hrtime(true) - $start) / 1e9, 2) . "s\n";
+register_shutdown_function(static function() use ($cacheName, $tmpReserveName) : void{
+	if(is_file($cacheName)){
+		\Phar::unlinkArchive($cacheName);
```
dktapps: Unsure what effects this might have during crashes - I'm not sure if this handler will be called before or after `Server::crashDump`

dktapps: Turns out it's called before, which is quite annoying :(

dktapps:
We also need a cache cleanup mechanism, for when servers crash and don't clean up after themselves

dktapps:
This now includes a cache cleanup mechanism, amongst other improvements (and still being littered with debug messages)

PR #6218, opened 2024-01-05
Make CocoaBlock Flowable

IvanCraft623:
## Introduction
Vanilla parity :P

### DIFF
--- a/src/block/CocoaBlock.php
+++ b/src/block/CocoaBlock.php
@@ -26,7 +26,6 @@ namespace pocketmine\block;
 use pocketmine\block\utils\AgeableTrait;
 use pocketmine\block\utils\BlockEventHelper;
 use pocketmine\block\utils\HorizontalFacingTrait;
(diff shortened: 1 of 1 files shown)


src/block/CocoaBlock.php
```diff
@@ -65,8 +64,8 @@ protected function recalculateCollisionBoxes() : array{
 		];
 	}
 
-	public function getSupportType(int $facing) : SupportType{
-		return SupportType::NONE;
+	public function isSolid() : bool{
```
dktapps: I'm not sure what you expect this to do. It's essentially the inverse of Flowable.

dktapps (changes requested):

src/block/CocoaBlock.php
IvanCraft623: Sorry, you're right, for some reason I thought solid meant it has collision.

dktapps (approved):

PR #6233, opened 2024-01-28
Remove many derprecated functions and classes

DebHugo:
## Introduction

I try to remove all deprecated function, classes... in PM6.

However, I need your help with the `getChunkPlayers` function in the `World` class.
It's written `deprecated` because the name of the function doesn't represent what the function actually does.
So I'm renaming this function to `getViewersAt` but I'm not sure if it's a good name and even, if the name change goes with this PR.

## Backwards compatibility

There are several BC-Break like in `Item`.
```
#Before Changes
final public function equals(Item $item, bool $checkDamage = true, bool $checkCompound = true) : bool{
#After changes
final public function equals(Item $item, bool $checkCompound = true) : bool{
```


DebHugo:
I don't understand where is the error. Can you help me please.

dktapps:
@deprecated doesn't mean "intended for removal". It just means "avoid using this". Most of this stuff can't be easily removed for one reason or another.

Issue #6164, opened 2023-11-17
ChunkListener::onBlockChanged() to accept 3 integers instead of Vector3


dktapps:
## Description
`Vector3` is both slower and more inconvenient for this purpose.

`setBlockAt()` currently has to allocate a `Vector3` for this purpose alone (well, there's also `createBlockUpdatePackets()`, but that could be avoided too). `Vector3` is inconvenient since it might contain float components, forcing the use of `getFloorX()` and friends.

Since we already use integers for this internally, it probably makes sense to do the same for `onBlockChanged()` too.

dktapps:
It might also be beneficial to add a way for ChunkListeners to hear about multiple block updates at once, which would permit things like [this](https://github.com/pmmp/PocketMine-MP/blob/67ad2bad175916372f9883a73188f120e66f6415/src/world/World.php#L1014-L1033) to be moved into the network system where they belong.

BajanVlogs:
foreach ($this->changedBlocks as $index => $blocks) {
    if (count($blocks) === 0) {
        continue;
    }

    World::getXZ($index, $chunkX, $chunkZ);

    // Check if the chunk is loaded before processing block updates
    if (!$this->isChunkLoaded($chunkX, $chunkZ)) {
        continue;
    }

    $chunk = $this->getChunk($chunkX, $chunkZ) ?? throw new AssumptionFailedError("We already checked that the chunk is loaded");

    // Group block updates into chunks for more efficient processing
    $chunkUpdates = [];

    foreach ($blocks as $block) {
        $blockX = $block->getX();
        $blockZ = $block->getZ();

        $chunkX = $blockX >> 4;
        $chunkZ = $blockZ >> 4;

        $chunkIndex = Level::chunkHash($chunkX, $chunkZ);
        $chunkUpdates[$chunkIndex][] = $block;
    }

    // Process chunk updates
    foreach ($chunkUpdates as $chunkIndex => $blockUpdates) {
        list($chunkX, $chunkZ) = Level::getChunkXZ($chunkIndex);

        if (count($blockUpdates) > 512) {
            foreach ($this->getChunkPlayers($chunkX, $chunkZ) as $player) {
                $player->onChunkChanged($chunkX, $chunkZ, $chunk);
            }
        } else {
            foreach ($this->createBlockUpdatePackets($blockUpdates) as $packet) {
                $this->broadcastPacketToPlayersUsingChunk($chunkX, $chunkZ, $packet);
            }
        }
[...]

PR #6237, opened 2024-01-30
Fixes Issue #6164

OfficialLilKev:
## Introduction

This modified code groups block updates based on the chunk they belong to before processing them. It aims to improve efficiency by minimizing redundant operations and sending fewer packets when updating multiple blocks within the same chunk. Keep in mind that the effectiveness of this optimization may vary based on the specific characteristics of your server and its usage patterns.

### DIFF
--- a/src/world/World.php
+++ b/src/world/World.php
@@ -1011,26 +1011,51 @@ class World implements ChunkManager{
 
 		if(count($this->changedBlocks) > 0){
 			if(count($this->players) > 0){
-				foreach($this->changedBlocks as $index => $blocks){
-					if(count($blocks) === 0){ //blocks can be set normally and then later re-set with direct send
-						continue;
-					}
-					World::getXZ($index, $chunkX, $chunkZ);
-					if(!$this->isChunkLoaded($chunkX, $chunkZ)){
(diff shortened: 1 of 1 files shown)


src/world/World.php
```diff
@@ -1011,26 +1011,51 @@ protected function actuallyDoTick(int $currentTick) : void{
...
+    $chunkUpdates = [];
+
+    foreach ($blocks as $block) {
+        $blockX = $block->getX();
+        $blockZ = $block->getZ();
+
+        $chunkX = $blockX >> 4;
+        $chunkZ = $blockZ >> 4;
+
+        // Use the instance of the Level class if available
```
dktapps, line 1038: what is this referring to?

src/world/World.php
```diff
@@ -1011,26 +1011,51 @@ protected function actuallyDoTick(int $currentTick) : void{
...
+        $chunkIndex = $chunk->chunkHash($chunkX, $chunkZ);
```
dktapps, line 1039: this function doesn't exist

dktapps:
This has nothing to do with the linked issue ...

This also changes nothing, as these network updates are already segregated by chunk.

PR #6239, opened 2024-02-03
Add lastdamage in durable item

roimee6:
## Introduction

At present, if we want to create a system for cleanly modifying items without overriding them, using events only and respecting pocketmie's "rules", we can't do it.

So I'm proposing a getLastDamage function to find out the last damage the item may have had

### DIFF
--- a/src/item/Durable.php
+++ b/src/item/Durable.php
@@ -30,6 +30,8 @@ use function min;
 
 abstract class Durable extends Item{
 	protected int $damage = 0;
+	protected int $lastDamage = 0;
+
 	private bool $unbreakable = false;
 
 	/**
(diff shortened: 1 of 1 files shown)


jasonw4331:
This is a completely unnecessary API. You can get the item's current durability at the same time as when the event fires without the item being modified.

roimee6:
> This is a completely unnecessary API. You can get the item's current durability at the same time as when the event fires without the item being modified.

Not really. And unless I didn't understand your sentence, you're out of line. But I can't know how much damage the item may have taken for the last time due to the "lgc_value" function used for unbreaking enchant.

dktapps:
This change didn't help with the unbreaking randomization anyway, since it stored the value before unbreaking was applied.

PR #6240, opened 2024-02-04
Proposed event for items when damaged

roimee6:
## Introduction

At the moment, when we want to modify the durability of an item, we have to override items

But overriding items can be inconvenient in several cases (sometimes)

The event here allows you to change the durability of items when they are damaged, but could be even more developed (for example, to modify durability via a position where the item is in use)

If you need more questions or additions to the event to be "more relevant", ask me

### DIFF
--- a/src/event/inventory/ItemDamageEvent.php
+++ b/src/event/inventory/ItemDamageEvent.php
@@ -23,62 +23,41 @@ declare(strict_types=1);
 
 namespace pocketmine\event\inventory;
 
-use pocketmine\crafting\CraftingRecipe;
 use pocketmine\event\Cancellable;
 use pocketmine\event\CancellableTrait;
 use pocketmine\event\Event;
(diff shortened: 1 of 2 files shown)


dktapps:
As I wrote on Discord, I don't think the event is useful in its current form.

It tells you that an item was damaged, and that's it. It doesn't tell you how the item was damaged, where the item is, who is using the item, what the target of the use was, etc.

I also don't think this event belongs in such a low-level function.

kaxyum:
good idea

roimee6:
I don't know how to introduce, honestly, the player in the applydamage function, because that would mean that all plugins that use the "applydamage" function would have to add the player as an argument

if anyone has an idea, please suggest it

dktapps:
Players generally damage items via a proxy method in `Living` anyway, to allow sound generation

dktapps:
Closing due to inactivity

Issue #6248, opened 2024-02-10
Allow plugins to decide resource packs per player


dktapps:
## Description
Plugins should be able to customise the resource packs offered for download by players during the login sequence.

## Justification
It's currently not possible in PM to have a resource pack devoted to accessibility (e.g. high contrast) without forcing everyone to use it.

It should be possible to offer different sets of resource packs to different players. This would allow plugins to offer optional accessibility packs to specific players that opted into it.

## Alternative methods
This is probably possible already with a bunch of cursed hacks involving custom ResourcePackManagers, but it would be much better if this (kinda obvious) feature was supported natively.

dktapps:
Implemented by #6249.

PR #6249, opened 2024-02-11
Implement player session-specific resource stacks

jasonw4331:
## Introduction
Read #6248 

### Relevant issues
Implements #6248 

### API changes
- Created `ResourcePackStackSendEvent`
  - Allows adding resource packs to the stack immediately before being sent to the client.
  - Allows assigning and reassigning encryption keys for resource packs being sent.
  - Allows plugins to declare if a pack must be accepted.
- `NetworkSession::onServerLoginSuccess()` now calls `ResourcePackStackSendEvent` before constructing `ResourcePacksPacketHandler`.
  - `ResourcePacksPacketHandler` now accepts two arrays and a bool instead of `ResourcePackManager`.
  -  `ResourcePacksPacketHandler` now holds its own known resource pack array.
  -  `ResourcePacksPacketHandler` now holds its own known pack encryption keys array.

### Behavioral changes
`ResourcePacksPacketHandler` is no longer always in sync with `ResourcePackManager`

## Backwards compatibility
non-BC changes allowed by `pocketmine\network\mcpe` internal namespace

## Tests
TODO

### DIFF
--- a/src/event/player/PlayerResourcePackOfferEvent.php
+++ b/src/event/player/PlayerResourcePackOfferEvent.php
@@ -58,15 +58,6 @@ class ResourcePackStackSendEvent extends Event{
 		return $this->resourcePackEntries;
 	}
 
-	public function setForceServerResource(bool $forceServerResources) : self{
-		$this->forceServerResources = $forceServerResources;
-		return $this;
-	}
-
-	public function forceServerResources() : bool{
-		return $this->forceServerResources;
-	}
-
 	public function setResourcePacksRequired(bool $resourcePacksRequired) : self{
 		$this->resourcePacksRequired = $resourcePacksRequired;
 		return $this;
--- a/src/event/player/PlayerResourcePackOfferEvent.php
+++ b/src/event/player/PlayerResourcePackOfferEvent.php
@@ -24,7 +24,7 @@ declare(strict_types=1);
 namespace pocketmine\event\player;
 
 use pocketmine\event\Event;
-use pocketmine\network\mcpe\NetworkSession;
+use pocketmine\player\PlayerInfo;
 use pocketmine\resourcepacks\ResourcePack;
 
 class PlayerResourcePackOfferEvent extends Event{
@@ -32,10 +32,14 @@ class PlayerResourcePackOfferEvent extends Event{
 	/**
 	 * @param ResourcePack[] $resourcePackEntries
 	 */
-	public function __construct(private readonly NetworkSession $session, private array $resourcePackEntries, private bool $resourcePacksRequired, private bool $forceServerResources){}
+	public function __construct(
+		private readonly PlayerInfo $playerInfo,
+		private array $resourcePackEntries,
+		private bool $resourcePacksRequired
+	){}
 
-	public function getSession() : NetworkSession{
-		return $this->session;
+	public function getPlayerInfo() : PlayerInfo{
+		return $this->playerInfo;
 	}
 
 	public function addResourcePack(ResourcePack $entry) : self{
--- a/src/event/player/PlayerResourcePackOfferEvent.php
+++ b/src/event/player/PlayerResourcePackOfferEvent.php
@@ -36,11 +36,11 @@ use pocketmine\resourcepacks\ResourcePack;
 class PlayerResourcePackOfferEvent extends Event{
 
 	/**
-	 * @param ResourcePack[] $resourcePackEntries
+	 * @param ResourcePack[] $resourcePacks
 	 */
 	public function __construct(
 		private readonly PlayerInfo $playerInfo,
-		private array $resourcePackEntries,
+		private array $resourcePacks,
 		private bool $resourcePacksRequired
 	){}
 
@@ -49,23 +49,23 @@ class PlayerResourcePackOfferEvent extends Event{
 	}
 
 	public function addResourcePack(ResourcePack $entry) : self{
-		$this->resourcePackEntries[] = $entry;
+		$this->resourcePacks[] = $entry;
 		return $this;
 	}
 
 	/**
-	 * @param ResourcePack[] $resourcePackEntries
+	 * @param ResourcePack[] $resourcePacks
 	 */
-	public function setResourcePackEntries(array $resourcePackEntries) : self{
-		$this->resourcePackEntries = $resourcePackEntries;
+	public function setResourcePacks(array $resourcePacks) : self{
+		$this->resourcePacks = $resourcePacks;
 		return $this;
 	}
 
 	/**
 	 * @return ResourcePack[]
 	 */
-	public function getResourcePackEntries() : array{
-		return $this->resourcePackEntries;
+	public function getResourcePacks() : array{
+		return $this->resourcePacks;
 	}
 
 	public function setResourcePacksRequired(bool $resourcePacksRequired) : self{
--- a/src/event/player/PlayerResourcePackOfferEvent.php
+++ b/src/event/player/PlayerResourcePackOfferEvent.php
@@ -38,7 +38,7 @@ class ResourcePackStackSendEvent extends Event{
 		return $this->session;
 	}
 
-	public function addResourcePackEntry(ResourcePack $entry) : self{
+	public function addResourcePack(ResourcePack $entry) : self{
 		$this->resourcePackEntries[] = $entry;
 		return $this;
 	}
--- a/src/event/player/PlayerResourcePackOfferEvent.php
+++ b/src/event/player/PlayerResourcePackOfferEvent.php
@@ -41,7 +41,7 @@ class PlayerResourcePackOfferEvent extends Event{
 	public function __construct(
 		private readonly PlayerInfo $playerInfo,
 		private array $resourcePacks,
-		private bool $resourcePacksRequired
+		private bool $mustAccept
 	){}
 
 	public function getPlayerInfo() : PlayerInfo{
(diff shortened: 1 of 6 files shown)


dktapps:
I'm not really sure why this requires modifying ResourcePackManager at all. This API seems quite clunky.

jasonw4331:
I figure for consistency it makes more sense to hold resource pack stacks in the manager

src/event/server/ResourcePackStackSendEvent.php
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @author PocketMine Team
+ * @link http://www.pocketmine.net/
+ *
+ *
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\event\server;
```
jasonw4331: Is this namespace alright to use for this event?

src/event/server/ResourcePackStackSendEvent.php
```diff
@@ -0,0 +1,78 @@
...
+use pocketmine\resourcepacks\ResourcePack;
+
+class ResourcePackStackSendEvent extends Event{
+
+	/**
+	 * @param ResourcePack[] $resourcePackEntries
+	 */
+	public function __construct(private readonly NetworkSession $session, private array $resourcePackEntries, private bool $resourcePacksRequired, private bool $forceServerResources){}
+
+	public function getSession() : NetworkSession{
```
dktapps: I don't see why we need to involve the NetworkSession here.

src/event/server/ResourcePackStackSendEvent.php
dktapps: It ought to be a Player event IMO. There's no reason for this to be network-specific.

I also think something like `PlayerResourcePackOfferEvent` would be more appropriate.

src/event/server/ResourcePackStackSendEvent.php
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	/**
+	 * @return ResourcePack[]
+	 */
+	public function getResourcePackEntries() : array{
+		return $this->resourcePackEntries;
+	}
+
+	public function setForceServerResource(bool $forceServerResources) : self{
```
dktapps: This isn't generally exposed and isn't documented. At first glance I didn't see what the difference was as compared to resourcePacksRequired.

src/event/server/ResourcePackStackSendEvent.php
```diff
@@ -0,0 +1,78 @@
...
+	/**
+	 * @param ResourcePack[] $resourcePackEntries
+	 */
+	public function __construct(private readonly NetworkSession $session, private array $resourcePackEntries, private bool $resourcePacksRequired, private bool $forceServerResources){}
+
+	public function getSession() : NetworkSession{
+		return $this->session;
+	}
+
+	public function addResourcePackEntry(ResourcePack $entry) : self{
```
dktapps: `entry` suffix feels unnecessary

src/event/server/ResourcePackStackSendEvent.php
jasonw4331: There are no player objects at this stage of the login sequence. Is PlayerInfo preferred?

dktapps: See PlayerPreLoginEvent.

src/event/player/PlayerResourcePackOfferEvent.php
```diff
@@ -27,6 +27,15 @@
...
 use pocketmine\resourcepacks\ResourcePack;
 
+/**
+ * Called after a player authenticates and is being offered resource packs to download.
+ *
+ * This event should be used to decide which resource packs to offer to the player, and whether to require them to
+ * download the packs before they can join the server.
+ *
+ * The Player object has not yet been instantiated, so PlayerInfo is used instead. The player is authenticated at this
+ * point, but the information should still be treated with caution.
```
dktapps: This last paragraph seems redundant

src/event/player/PlayerResourcePackOfferEvent.php
```diff
@@ -38,6 +47,10 @@ public function __construct(
 		private bool $resourcePacksRequired
 	){}
 
+	/**
+	 * Returns an object containing self-proclaimed information about the connecting player. The player has been
+	 * authenticated at this point, but the information should still be treated with caution.
```
dktapps: This is unnecessarily over-descriptive, and the caution is unnecessary - once they are authenticated the information can be fully trusted

src/event/player/PlayerResourcePackOfferEvent.php
```diff
@@ -0,0 +1,86 @@
...
+
+	public function addResourcePack(ResourcePack $entry) : self{
+		$this->resourcePackEntries[] = $entry;
+		return $this;
+	}
+
+	/**
+	 * @param ResourcePack[] $resourcePackEntries
+	 */
+	public function setResourcePackEntries(array $resourcePackEntries) : self{
```
dktapps: See previous comments about `entry` suffixes

src/event/player/PlayerResourcePackOfferEvent.php
```diff
@@ -0,0 +1,86 @@
...
+		$this->resourcePackEntries = $resourcePackEntries;
+		return $this;
+	}
+
+	/**
+	 * @return ResourcePack[]
+	 */
+	public function getResourcePackEntries() : array{
```
dktapps: Same again here

src/event/player/PlayerResourcePackOfferEvent.php
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	public function setResourcePacksRequired(bool $resourcePacksRequired) : self{
+		$this->resourcePacksRequired = $resourcePacksRequired;
+		return $this;
+	}
+
+	public function resourcePacksRequired() : bool{
+		return $this->resourcePacksRequired;
+	}
```
dktapps, line 105: `resourcePacksRequired` would probably be better named `mustAccept` (it's ambiguous what "required" is referring to)

src/network/mcpe/handler/ResourcePacksPacketHandler.php
```diff
@@ -65,6 +67,9 @@ public function __construct(
 	){}
 
 	public function setUp() : void{
+		assert($this->session->getPlayerInfo() !== null, "PlayerInfo should be set before ResourcePacksPacketHandler is used");
```
dktapps: Have this passed to the constructor instead

src/event/player/PlayerResourcePackOfferEvent.php
```diff
@@ -0,0 +1,85 @@
...
+
+	/**
+	 * @param ResourcePack[] $resourcePacks
+	 */
+	public function __construct(
+		?PlayerInfo $playerInfo,
+		private array $resourcePacks,
+		private bool $mustAccept
+	){
+		assert($playerInfo !== null, "PlayerInfo cannot be null");
```
dktapps: ?????

jasonw4331: It resolves with phpstan this way. Is a new rule preferred?

dktapps: No, I said pass it in the constructor of the ResourcePacksPacketHandler. Not sure what you thought I meant ...

jasonw4331: I did that with [cee89f7](https://github.com/pmmp/PocketMine-MP/pull/6249/commits/cee89f7aa464c1a2857e8f23930556e2ab360b50) but it causes phpstan analysis to fail so I moved it here

dktapps: All you did was move the assert. That's not what I said.

src/event/player/PlayerResourcePackOfferEvent.php
```diff
@@ -0,0 +1,79 @@
...
+		private array $resourcePacks,
+		private bool $mustAccept
+	){}
+
+	public function getPlayerInfo() : PlayerInfo{
+		return $this->playerInfo;
+	}
+
+	public function addResourcePack(ResourcePack $entry) : self{
+		$this->resourcePacks[] = $entry;
```
dktapps: I'm pretty sure this will put the pack at the bottom of the stack, which probably isn't desired

src/event/player/PlayerResourcePackOfferEvent.php
```diff
@@ -0,0 +1,79 @@
...
+		return $this;
```
dktapps: Fluent returns aren't standard for events

src/event/player/PlayerResourcePackOfferEvent.php
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	/**
+	 * @param ResourcePack[] $resourcePacks
+	 */
+	public function setResourcePacks(array $resourcePacks) : self{
+		$this->resourcePacks = $resourcePacks;
+		return $this;
```
dktapps: Same here

src/event/player/PlayerResourcePackOfferEvent.php
```diff
@@ -0,0 +1,79 @@
...
+	/**
+	 * @return ResourcePack[]
+	 */
+	public function getResourcePacks() : array{
+		return $this->resourcePacks;
+	}
+
+	public function setMustAccept(bool $mustAccept) : self{
+		$this->mustAccept = $mustAccept;
+		return $this;
```
dktapps: and here

dktapps:
This implementation doesn't support encrypting any dynamically inserted resource packs

src/network/mcpe/handler/ResourcePacksPacketHandler.php
```diff
@@ -72,6 +71,12 @@ public function __construct(
 	public function setUp() : void{
 		$event = new PlayerResourcePackOfferEvent($this->playerInfo, $this->resourcePackManager->getResourceStack(), $this->resourcePackManager->resourcePacksRequired());
 		$event->call();
+		foreach($event->getEncryptionKeys() as $packId => $key) {
+			$existingKey = $this->resourcePackManager->getPackEncryptionKey($packId);
+			if(is_string($existingKey) && $existingKey !== $key)
+				continue;
+			$this->resourcePackManager->setPackEncryptionKey($packId, $key);
```
dktapps: ?????

dktapps:
Please test the PR before submitting it next time

PR #6255, opened 2024-02-13
Fixed `Utils::cloneObjectArray()` template signature

JavierLeon9966:
## Introduction

The previous signature removed information about the array key type.

## Tests

Before: https://phpstan.org/r/93aaa0e4-55de-414c-9ee0-4130ee1b6666
After: https://phpstan.org/r/07d51368-d7bd-459e-a8a4-b2a7db97b4ab


src/utils/Utils.php
```diff
@@ -173,16 +173,17 @@ public static function cloneCallback() : \Closure{
 	}
 
 	/**
-	 * @phpstan-template T of object
+	 * @phpstan-template T of array-key
+	 * @phpstan-template U of object
```
dktapps: `TKey` and `TValue` would be more consistent with the rest of the project

PR #6256, opened 2024-02-14
Add new BiomeIDs

pandaaaBE:
Source: https://minecraft.wiki/w/Biome/ID

### DIFF
--- a/src/data/bedrock/BiomeIds.php
+++ b/src/data/bedrock/BiomeIds.php
@@ -111,4 +111,15 @@ final class BiomeIds{
 	public const CRIMSON_FOREST = 179;
 	public const WARPED_FOREST = 180;
 	public const BASALT_DELTAS = 181;
+	public const JAGGED_PEAKS = 182;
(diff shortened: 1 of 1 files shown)


dktapps:
As far as I remember, this file is generated from here: https://github.com/pmmp/BedrockData/blob/master/biome_id_map.json

pandaaaBE:
Okay, then this PR is irrelevant, just see it as an issue it will probably be faster if you do it when you have time or someone wants to teach me something new and tells me where to find the file to generate :)

ShockedPlot7560:
It seems that this file is generated yes. However, there are no tools or documentation to generate it. @dktapps do you have the link to this tool or does this file need to be edited manually now?

dktapps:
It's probably lurking in my local workspace somewhere, I'll dig it up later

dktapps:
Fixed by 61d0181bfd7122d76109f8ff390675eaf35f3da3 and 2616d8c5adacff29de6a8db17c8e13a075ff5ea9.

PR #6258, opened 2024-02-18
Event when a block support is broken

roimee6:
## Introduction
Currently, when a block is broken because its support is broken, we can't modify the drops as in blockbreakevent, which can create duplication bugs in our plugins

I've searched all the uses of the "useBreakOn" function without a player is linked to a broken block because its support has been broken

### DIFF
--- a/src/event/block/BlockSupportBreakEvent.php
+++ b/src/event/block/BlockSupportBreakEvent.php
@@ -1,0 +1,85 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * @author PocketMine Team
+ * @link http://www.pocketmine.net/
+ *
+ *
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\event\block;
+
+use pocketmine\block\Block;
+use pocketmine\item\Item;
+
+/**
+ * Called when a block is destroyed because its support has been destroyed.
+ */
+class BlockSupportBreakEvent extends BlockEvent{
+	/** @var Item[] */
+	protected array $blockDrops = [];
+
+	/**
+	 * @param Item[] $drops
+	 */
+	public function __construct(
+		Block $block,
+		array $drops = [],
+		protected int $xpDrops = 0
+	){
+		parent::__construct($block);
+		$this->setDrops($drops);
+	}
+
+	/**
+	 * @return Item[]
+	 */
+	public function getDrops() : array{
+		return $this->blockDrops;
+	}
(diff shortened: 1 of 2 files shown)


roimee6:
The name of the event seems a bit off, but I can't think of anything else to say in English

pandaaaBE:
What exactly do you mean by support?

pandaa-be:
Why did you open an else in World.php without a return? Then you could have just left the else, couldn't you?

By the way, I'm not a PocketMine reviewer so you don't have to listen to me.

roimee6:
> What exactly do you mean by support?

Support block, when the block holding the seeds, for example, is broken, this event is called for the seed block

kaxyum:
This could be very useful, but the name needs to be revised. Alternatively, why not modify the blockbreak event so that it can be called here?

dktapps:
Messing with existing events would break BC.

ShockedPlot7560:
I don't think this is the best name for this type of event. In any case, changing the BlockBreak event isn't the best solution either.

dktapps:
Really BlockBreakEvent itself is mis-named (it should be PlayerBlockBreakEvent).

Removal of support isn't the only case in which a block will get destroyed without a player present. For example, water flowing into a block will destroy it, triggering this event.

roimee6:
So what name could be representative?

dktapps:
I don't have a problem with it personally. My only issue is that this will currently be called in cases where there wasn't any support to begin with.

kaxyum:
> Really BlockBreakEvent itself is mis-named (it should be PlayerBlockBreakEvent).
> 
> Removal of support isn't the only case in which a block will get destroyed without a player present. For example, water flowing into a block will destroy it, triggering this event.

So why not create a PlayerBlockBreakEvent and a BlockBreakEvent ?

dktapps:
> > Really BlockBreakEvent itself is mis-named (it should be PlayerBlockBreakEvent).
> > Removal of support isn't the only case in which a block will get destroyed without a player present. For example, water flowing into a block will destroy it, triggering this event.
> 
> So why not create a PlayerBlockBreakEvent and a BlockBreakEvent ?

That's a big change that will break a lot of plugins

kaxyum:
> > > Really BlockBreakEvent itself is mis-named (it should be PlayerBlockBreakEvent).
> > > Removal of support isn't the only case in which a block will get destroyed without a player present. For example, water flowing into a block will destroy it, triggering this event.
> > 
> > 
> > So why not create a PlayerBlockBreakEvent and a BlockBreakEvent ?
> 
> That's a big change that will break a lot of plugins

Yes, so why not do it for PM6?

dktapps:
> > > > Really BlockBreakEvent itself is mis-named (it should be PlayerBlockBreakEvent).
> > > > Removal of support isn't the only case in which a block will get destroyed without a player present. For example, water flowing into a block will destroy it, triggering this event.
> > > 
> > > 
> > > So why not create a PlayerBlockBreakEvent and a BlockBreakEvent ?
> > 
> > 
> > That's a big change that will break a lot of plugins
> 
> Yes, so why not do it for PM6?

Just because we have a major version doesn't mean we should just break plugins. The benefits of making a BC break need to be weighed against the costs of changing it.

dktapps:
I think we could implement this as `BlockSupportRemoveEvent` if the PR was adapted to only fire the event when block support was actually deleted. This would (in theory) be backwards-compatible.

The best way to do this would probably be to create a wrapper for `useBreakOn()` that would fire the event, and then update all the places where lack of support causes a block to be destroyed to use the wrapper method. (or perhaps a cause parameter to `useBreakOn`.)

roimee6:
> I think we could implement this as `BlockSupportRemoveEvent` if the PR was adapted to only fire the event when block support was actually deleted. This would (in theory) be backwards-compatible.
> 
> The best way to do this would probably be to create a wrapper for `useBreakOn()` that would fire the event, and then update all the places where lack of support causes a block to be destroyed to use the wrapper method. (or perhaps a cause parameter to `useBreakOn`.)

it will be done soon, sorry for the wait I was playing more minecraft

PR #6261, opened 2024-02-19
feat(Server): add a setter for `maxPlayers`

cooldogedev:
## Introduction

This PR allows plugins to modify the server's max players count during runtime

### API changes

Added `Server::setMaxPlayers(int $maxPlayers): void`

## Tests

- [ ] Other (provide details)

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -343,6 +343,10 @@ class Server{
 		return $this->maxPlayers;
 	}
 
+	public function setMaxPlayers(int $maxPlayers) : void{
+		$this->maxPlayers = $maxPlayers;
+	}
+
 	/**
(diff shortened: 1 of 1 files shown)


SOF3:
note that this is a BC break, e.g. for plugins that do a SplFixedArray($server->getMaxPlayers())

dktapps:
does anyone do that?

BrandPVP:
there is no bc break because there is no such plugins that call setMaxPlayers()

Wraith0x10:
rip i was about to make a pr doing this too 🙏

dktapps (dismissed):
I wonder if a `ServerPlayerLimitChangeEvent` is needed. Just wondering if some plugins might need to be informed of this change to invalidate caches or recompute stuff. Just a thought, not necessary to merge the PR.

PR #6266, opened 2024-02-23
Fix: banner item has no patterns when call BaseBanner->asItem()

BrandPVP:


dktapps:
The original behaviour is expected. This change breaks the behaviour of block picking.

PR #6287, opened 2024-03-16
Added sound when picking sweet berries to match vanilla

ipad54:
## Introduction
Added sound when picking sweet berries to match vanilla

## Tests
I tested this PR by doing the following
- [X] Playtesting using a Minecraft client (https://www.youtube.com/watch?v=SF0tEbHIqNA)


dktapps:
Thanks :)

PR #6304, opened 2024-03-24
Validate transaction slots

### LINKED ISSUE #6262: A system to let transactions decide what stuff can go into a particular inventory slot
## Description
This was previously attempted in #4479, but that implementation is limited and has gone stale.

> When I originally thought about this, my idea was to have a separate InventorySlotAcceptor unit which would enforce requirements like this, avoiding circular dependencies or needing custom inventory classes, as well as being more flexible.
>
> It would have a method something like
> ```
> function accept(Inventory $inventory, Item $item) : bool
> ```
> where the return value would indicate whether the item can be put into that slot or not.
>
> Maybe it can even just be done with an `Inventory::getSlotAcceptors() : ObjectSet<Closure(Inventory, Item) : bool>` instead.

A system like this would allow any Inventory to receive new acceptors at any time, allowing plugins to introduce their own logic and potentially override builtin logic.

## Justification
Slots for armour and offhand aren't currently validated.

ShockedPlot7560:
## Introduction
At present, we can place an item in any slot in an inventory (except in exceptional cases).

It is now necessary to provide a system for verifying transactions in order to guarantee a stable internal state for each inventory.

To maintain compatibility and avoid changing the Inventory interface. A new interface has been created, introducing a new notion of : "slot safe". 
I named an inventory slot safe when we are guaranteed that the internal state is consistent with what it should do. So it's up to the core (for inventory implemented by pmmp) and plugin developers to make sure they implement the right validation logic.

### Relevant issues
* Fixes #6262

### API changes
- Add `SlotSafeInventory`
- Add `SlotValidator`

### Behavioural changes
- `ArmorInventory` checked now correctly their slots. 

## Backwards compatibility
Enforced by the `SlotSafeInventory`

## Follow-up

As proof of concept, I only implement the `ArmorInventory` logics. There is many inventory which need to check their slot.


src/inventory/ArmorInventory.php
```diff
@@ -73,4 +77,8 @@ public function setLeggings(Item $leggings) : void{
 	public function setBoots(Item $boots) : void{
 		$this->setItem(self::SLOT_FEET, $boots);
 	}
+
+	protected static function initSlotValidators(ObjectSet $validators) : void{
```
dktapps: Not sure what the point of this is. Just override the constructor.

src/inventory/SlotSafeInventory.php
```diff
@@ -0,0 +1,45 @@
...
+	 * Return a set of validators that will be used to determine whether an item can be placed in a particular slot.
+	 * All validators need to return true for the transaction to be allowed.
+	 * If one of the validators returns false, the transaction will be cancelled.
+	 *
+	 * There is no guarantee that the validators will be called in any particular order.
+	 * All validators need to be stateless and not depend on the order in which they are called.
+	 *
+	 * @phpstan-return ObjectSet<InventoryTransactionValidator>
+	 */
+	public static function getSlotValidators() : ObjectSet;
```
dktapps: why static? this seems unnecessarily limiting

src/inventory/transaction/validator/ArmorInventoryValidator.php
```diff
@@ -0,0 +1,44 @@
...
+use pocketmine\block\MobHead;
+use pocketmine\inventory\ArmorInventory;
+use pocketmine\inventory\Inventory;
+use pocketmine\item\Armor;
+use pocketmine\item\Item;
+use pocketmine\item\ItemBlock;
+
+class ArmorInventoryValidator implements InventoryTransactionValidator{
+	public function validate(Inventory $inventory, Item $item, int $slot) : bool{
+		return $inventory instanceof ArmorInventory && (
```
dktapps: This instanceof seems unnecessary

src/inventory/transaction/validator/InventoryTransactionValidator.php
```diff
@@ -0,0 +1,41 @@
...
+ * This interface is used to validate the transactions of a "safe" inventory. {@see SlotSafeInventory}
+ *
+ * In this way, each inventory can determine whether or not an item has the right to be placed in a particular slot.
+ */
+interface InventoryTransactionValidator{
+	/**
+	 * @return bool Return true, if the item CAN be placed in the given slot of the given inventory.
+	 * Otherwise, return false and the transaction will be cancelled.
+	 */
+	public function validate(Inventory $inventory, Item $item, int $slot) : bool;
```
dktapps: I don't see why we need an inventory parameter here

dktapps (commented):
This could also be done with closures

src/inventory/SlotSafeInventory.php
ShockedPlot7560: So plugins can add any validator to the set of an inventory. 
Example: you want your custom item to be added on the head slot, you call the `getSlotValidators` on the load or enable of the plugins so it will be added all the time.

src/inventory/transaction/validator/InventoryTransactionValidator.php
ShockedPlot7560: Can be useful if an item can be placed in a slot dependent of another slot.

src/inventory/SlotSafeInventory.php
dktapps: I can't see a case where we'd need to do this for a non custom inventory. For a custom one we'd just override the constructor.

This API design seems weird and it creates unnecessary limitations in other areas.

dktapps: Moreover, future scope for inventories involves deleting the classes. The type would be inferred from the context (e.g. Living->getArmorInventory()) in which case there wouldn't be a class to do static binding anyway.

I would also prefer that SlotSafeInventory were implemented by BaseInventory directly so that this functionality is automatically available for all inventories. I don't see a need to put it in a trait.

ShockedPlot7560: > I can't see a case where we'd need to do this for a non custom inventory. For a custom one we'd just override the constructor.

If any case, you want to put a red carpet on your head because you created a custom logic for this block. With the new validator you can't except if you can add validator to support then.

ShockedPlot7560: Static may not be the best solution, but we need to allow plugins adding validators.

dktapps: I agree, but I don't think static is the way to do it.

Typically I would think EntitySpawnEvent or ChunkLoadEvent would be sufficient to setup validators for vanilla things.

dktapps: Either that or just apply validators as-needed during InventoryOpenEvent... that would probably make more sense

src/inventory/ArmorInventory.php
```diff
@@ -36,6 +40,14 @@ public function __construct(
 		protected Living $holder
 	){
 		parent::__construct(4);
+
+		$this->validators->add(static function (Inventory $inventory, Item $item, int $slot) : bool{
+			return ($item instanceof Armor && $item->getArmorSlot() === $slot) ||
+				($slot === ArmorInventory::SLOT_HEAD && $item instanceof ItemBlock && (
+						$item->getBlock()->getTypeId() === BlockTypeIds::CARVED_PUMPKIN ||
+						$item->getBlock() instanceof MobHead
```
dktapps: I'd prefer checking type ID or tags here
Instanceof is rarely the right solution for this

src/inventory/BaseInventory.php
```diff
@@ -23,6 +23,7 @@
 
 namespace pocketmine\inventory;
 
+use Closure;
```
dktapps: Global namespace classes are usually referred to with a `\` instead of importing

src/inventory/BaseInventory.php
```diff
@@ -46,9 +47,12 @@ abstract class BaseInventory implements Inventory{
 	 * @phpstan-var ObjectSet<InventoryListener>
 	 */
 	protected ObjectSet $listeners;
+	/** @phpstan-var ObjectSet<Closure(Inventory $inventory, Item $item, int $slot): bool> */
```
dktapps: Might be worth declaring a `@phpstan-type` for this in `SlotSafeInventory`

src/inventory/SlotSafeInventory.php
```diff
@@ -0,0 +1,46 @@
...
+
+use Closure;
+use pocketmine\item\Item;
+use pocketmine\utils\ObjectSet;
+
+/**
+ * A "slot safe inventory" is an inventory that has a set of rules that determine whether an item can be placed in a
+ * particular slot. This ensures that the inventory's internal state is always valid and consistent.
+ */
+interface SlotSafeInventory{
```
dktapps: Probably worth merging this into `Inventory` in PM6 (perhaps we need to just get rid of the `Inventory` interface entirely...)

src/inventory/SlotSafeInventory.php
```diff
@@ -0,0 +1,46 @@
...
+
+namespace pocketmine\inventory;
+
+use Closure;
+use pocketmine\item\Item;
+use pocketmine\utils\ObjectSet;
+
+/**
+ * A "slot safe inventory" is an inventory that has a set of rules that determine whether an item can be placed in a
+ * particular slot. This ensures that the inventory's internal state is always valid and consistent.
```
dktapps: Simpler way to explain this:
```
A "slot safe inventory" has validators which may restrict items from being placed in particular slots of the inventory.
```

src/inventory/SlotSafeInventory.php
```diff
@@ -0,0 +1,46 @@
...
+ */
+interface SlotSafeInventory{
+	/**
+	 * Return a set of validators that will be used to determine whether an item can be placed in a particular slot.
+	 * One validator need to return true for the transaction to be allowed.
+	 * If all the validators returns false, the transaction will be cancelled.
```
dktapps: I believe we previously discussed this behaviour on Discord (it doesn't make sense)
Not sure if you already changed the logic and just forgot to change the documentation

src/inventory/SlotSafeInventory.php
```diff
@@ -0,0 +1,46 @@
...
+	 *
+	 * There is no guarantee that the validators will be called in any particular order.
+	 * All validators need to be stateless and not depend on the order in which they are called.
```
dktapps: This second line is redundant

src/inventory/transaction/action/SlotChangeAction.php
```diff
@@ -74,6 +76,17 @@ public function validate(Player $source) : void{
 		if($this->targetItem->getCount() > $this->inventory->getMaxStackSize()){
 			throw new TransactionValidationException("Target item exceeds inventory max stack size");
 		}
+		if($this->inventory instanceof SlotSafeInventory && !$this->targetItem->equals(VanillaItems::AIR())){
```
dktapps:
```suggestion
		if($this->inventory instanceof SlotSafeInventory && !$this->targetItem->isNull()){
```

src/inventory/transaction/action/SlotChangeAction.php
```diff
@@ -74,6 +76,17 @@ public function validate(Player $source) : void{
...
+			$result = true;
+			foreach($this->inventory->getSlotValidators() as $validator){
+				if(($result = $validator($this->inventory, $this->targetItem, $this->inventorySlot))){
```
dktapps: See above about logic inversion

dktapps (changes requested):
Concept looks fine.

However, as previously discussed on Discord, I think that all validators should accept the item in order to allow the transaction to proceed, as opposed to the current behaviour of only one. The current behaviour does not make sense to me - I think it's more likely that people will want to add extra restrictions instead of overriding existing ones

src/inventory/transaction/action/SlotChangeAction.php
ShockedPlot7560: Thats currently the behavior. all validator must accept the transaction, however it will be rejected.

src/inventory/SlotSafeInventory.php
ShockedPlot7560: Agreed, but for now, i think breaking BC isnt necessary as this functionnality is quite important to implement.

src/inventory/transaction/action/SlotChangeAction.php
ShockedPlot7560: I didn't say anything

ShockedPlot7560:
Alright, I think we are done with this.

src/inventory/SlotSafeInventory.php
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * A "slot safe inventory" has validators which may restrict items
+ * from being placed in particular slots of the inventory.
+ *
+ * @phpstan-type SlotValidators ObjectSet<\Closure(Inventory $inventory, Item $item, int $slot): bool>
+ */
+interface SlotSafeInventory{
+	/**
+	 * Return a set of validators that will be used to determine whether an item can be placed in a particular slot.
+	 * All validator need to return true for the transaction to be allowed.
```
dktapps:
```suggestion
	 * All validators need to return true for the transaction to be allowed.
```

src/inventory/SlotSafeInventory.php
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * A "slot safe inventory" has validators which may restrict items
+ * from being placed in particular slots of the inventory.
+ *
+ * @phpstan-type SlotValidators ObjectSet<\Closure(Inventory $inventory, Item $item, int $slot): bool>
+ */
+interface SlotSafeInventory{
+	/**
+	 * Return a set of validators that will be used to determine whether an item can be placed in a particular slot.
```
dktapps:
```suggestion
	 * Returns a set of validators that will be used to determine whether an item can be placed in a particular slot.
```

src/inventory/transaction/action/SlotChangeAction.php
```diff
@@ -74,6 +75,13 @@ public function validate(Player $source) : void{
 		if($this->targetItem->getCount() > $this->inventory->getMaxStackSize()){
 			throw new TransactionValidationException("Target item exceeds inventory max stack size");
 		}
+		if($this->inventory instanceof SlotSafeInventory && !$this->targetItem->isNull()){
+			foreach($this->inventory->getSlotValidators() as $validator){
+				if(!$validator($this->inventory, $this->targetItem, $this->inventorySlot)){
```
dktapps: Perhaps the validators should throw (or return) TransactionValidationException directly, so that a detailed error can be given

dktapps (commented):
Looks good overall, minus a couple of nits and suggestions

src/inventory/transaction/action/SlotChangeAction.php
ShockedPlot7560: I think throwing an exception is better? If it returns false, then it's a general error.

dktapps: We can't document a Closure to throw a specific type of exception. However, we could have it return `TransactionValidationException|null`.

ShockedPlot7560: So people is force to return an exception with an error message, if transaction denied

dktapps: Yes

src/inventory/transaction/action/SlotChangeAction.php
```diff
@@ -74,6 +75,17 @@ public function validate(Player $source) : void{
...
 			throw new TransactionValidationException("Target item exceeds inventory max stack size");
 		}
+		if($this->inventory instanceof SlotSafeInventory && !$this->targetItem->isNull()){
+			foreach($this->inventory->getSlotValidators() as $validator){
+				try{
+					if(!$validator($this->inventory, $this->targetItem, $this->inventorySlot)){
+						throw new TransactionValidationException("Target item is not accepted by the inventory");
+					}
+				}catch(TransactionValidationException $e){
+					throw new TransactionValidationException("Target item is not accepted by the inventory: " . $e->getMessage());
```
dktapps:
```suggestion
					throw new TransactionValidationException("Target item is not accepted by the inventory slot: " . $e->getMessage(), 0, $e);
```

src/inventory/ArmorInventory.php
```diff
@@ -73,4 +80,20 @@ public function setLeggings(Item $leggings) : void{
 	public function setBoots(Item $boots) : void{
 		$this->setItem(self::SLOT_FEET, $boots);
 	}
+
+	private function validate(Inventory $inventory, Item $item, int $slot) : null|TransactionValidationException{
```
ipad54:
```suggestion
	private function validate(Inventory $inventory, Item $item, int $slot) : ?TransactionValidationException{
```

src/inventory/transaction/action/SlotChangeAction.php
```diff
@@ -74,6 +75,14 @@ public function validate(Player $source) : void{
 		if($this->targetItem->getCount() > $this->inventory->getMaxStackSize()){
 			throw new TransactionValidationException("Target item exceeds inventory max stack size");
 		}
+		if($this->inventory instanceof SlotSafeInventory && !$this->targetItem->isNull()){
+			foreach($this->inventory->getSlotValidators() as $validator){
+				$ret = $validator->validate($this->inventory, $this->targetItem, $this->inventorySlot);
+				if($ret instanceof TransactionValidationException){
```
ipad54:
```suggestion
				if($ret !== null){
```

dktapps (dismissed):

src/inventory/transaction/action/validator/SlotValidator.php
```diff
@@ -0,0 +1,38 @@
...
+declare(strict_types=1);
+
+namespace pocketmine\inventory\transaction\action\validator;
+
+use pocketmine\inventory\Inventory;
+use pocketmine\inventory\transaction\TransactionValidationException;
+use pocketmine\item\Item;
+
+/**
+ * Validate a slot placement in an inventory.
```
ipad54:
```suggestion
 * Validates a slot placement in an inventory.
```

src/inventory/ArmorInventory.php
```diff
@@ -73,4 +80,20 @@ public function setLeggings(Item $leggings) : void{
...
+		if($item instanceof Armor && $item->getArmorSlot() === $slot){
+			if($item->getArmorSlot() !== $slot){
+				return new TransactionValidationException("Armor item in wrong slot");
+			}
+		}else{
+			if(!($slot === ArmorInventory::SLOT_HEAD && $item instanceof ItemBlock && (
+					$item->getBlock()->getTypeId() === BlockTypeIds::CARVED_PUMPKIN ||
+					$item->getBlock()->getTypeId() === BlockTypeIds::MOB_HEAD
+				))){
+				return new TransactionValidationException("Item not accepted in an armor slot");
```
ipad54:
```suggestion
				return new TransactionValidationException("Item is not accepted in an armor slot");
```

src/inventory/ArmorInventory.php
```diff
@@ -73,4 +80,20 @@ public function setLeggings(Item $leggings) : void{
 	public function setBoots(Item $boots) : void{
 		$this->setItem(self::SLOT_FEET, $boots);
 	}
+
+	private function validate(Inventory $inventory, Item $item, int $slot) : ?TransactionValidationException{
+		if($item instanceof Armor && $item->getArmorSlot() === $slot){
+			if($item->getArmorSlot() !== $slot){
+				return new TransactionValidationException("Armor item in wrong slot");
```
ipad54:
```suggestion
				return new TransactionValidationException("Armor item is in wrong slot");
```

dktapps (dismissed):

src/inventory/ArmorInventory.php
```diff
@@ -73,4 +80,20 @@ public function setLeggings(Item $leggings) : void{
 	public function setBoots(Item $boots) : void{
 		$this->setItem(self::SLOT_FEET, $boots);
 	}
+
+	private function validate(Inventory $inventory, Item $item, int $slot) : ?TransactionValidationException{
+		if($item instanceof Armor && $item->getArmorSlot() === $slot){
```
ipad54: Is the second part of condition (`&& $item->getArmorSlot() === $slot`) needed? If I read code correctly, the second condition (`if($item->getArmorSlot() !== $slot){`) will be always false since it’s already assumed that the item armor slot should equals the inventory slot

jasonw4331 (approved):

src/inventory/ArmorInventory.php
dktapps: Yeah, you are correct. cc @ShockedPlot7560

dries-c (approved):

ipad54 (commented):
LGTM

IvanCraft623 (approved):
LGTM, Feel free to merge it

PR #6307, opened 2024-03-24
Implement Saddle & Horse Armor

pandaaaBE:
## Introduction
Implement Saddle & Horse Armor.

(Currently this is just for cosmetic because there are no mobs in pocketmine but I need this for my mob plugin :P)

### API changes
`VanillaItems::SADDLE()`
`VanillaItems::LEATHER_HORSE_ARMOR()`
`VanillaItems::IRON_HORSE_ARMOR()`
`VanillaItems::DIAMOND_HORSE_ARMOR()`
`VanillaItems::GOLDEN_HORSE_ARMOR()`

### Tested LEATHER_HORSE_ARMOR custom colour with following code:
```
> 	public function onJoin(PlayerJoinEvent $event): void{
> 		$player = $event->getPlayer();
> 		$inventory = $player->getInventory();
> 
> 		$inventory->clearAll();
> 
> 		$inventory->addItem(VanillaItems::LEATHER_HORSE_ARMOR()->setCustomColor(new Color(500, 100, 254)));
> 		$inventory->addItem(VanillaItems::LEATHER_HORSE_ARMOR());
> 		$inventory->addItem(VanillaItems::IRON_HORSE_ARMOR());
> 		$inventory->addItem(VanillaItems::DIAMOND_HORSE_ARMOR());
> 		$inventory->addItem(VanillaItems::GOLDEN_HORSE_ARMOR());
> 		$inventory->addItem(VanillaItems::SADDLE());
> 	}
```

## Tests

- [x] Playtesting using a Minecraft client
image


pandaaaBE:
Is it enough if it stays as it is now (as I said, there are no mobs) or should I add the Horse Armor to Living::attack() anyway?

ShockedPlot7560 (commented):
You need to regenerate the static functions

pandaaaBE:
> You need to regenerate the static functions 

for whatever reason i can't run the file on my laptop.

can do that when i am back at the pc 👌

pandaaaBE:
image

Is it enough if it stays as it is now (as I said, there are no mobs) or should I do more?

ShockedPlot7560:
As the PR name and objective is to add these items and not the horse logics, its enough.

src/item/HorseArmor.php
ShockedPlot7560, line 1: Shouldn't we allow armor to be colored if it's not leather?

src/item/HorseArmor.php
```diff
@@ -0,0 +1,86 @@
...
+	public const TAG_CUSTOM_COLOR = "customColor"; //TAG_Int
+
+	protected ?Color $customColor = null;
+
+	/**
+	 * @param string[] $enchantmentTags
+	 */
+	public function __construct(ItemIdentifier $identifier, string $name, array $enchantmentTags = []){
+		parent::__construct($identifier, $name, $enchantmentTags);
+	}
```
ShockedPlot7560: No need to have constructor so

src/item/HorseArmor.php
```diff
@@ -0,0 +1,86 @@
...
+	public function setCustomColor(Color $color) : self{
+		$this->customColor = $color;
+		return $this;
+	}
+
+	/** @return $this */
+	public function clearCustomColor() : self{
+		$this->customColor = null;
+		return $this;
+	}
```
ShockedPlot7560: Permit null in the setter instead of just creating a new function for that

ipad54:
Cauldron behaviour is also missing. In vanilla, leather armour can be dyed in cauldron.

src/item/HorseArmor.php
pandaaaBE: This code is by Armor.php but I can replace it if you want

pandaaaBE:
> Cauldron behaviour is also missing. In vanilla, leather armour can be dyed in cauldron.

Jup but this is currently not implemented in pocketmine and I don´t feel like it is something for this pr..

ipad54:
> Jup but this is currently not implemented in pocketmine and I don´t feel like it is something for this pr..

What do you mean? Cauldron is implemented

src/item/HorseArmor.php
pandaaaBE: I think clearCustomColor() is needed if you want to remove the color via plugin, to me clearCustomColor() sounds more logical than setCustomColor(null), doesn't it?

pandaaaBE:
image

prove that it works

src/block/WaterCauldron.php
```diff
@@ -128,9 +129,10 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
 			}else{
 				$this->mix($item, VanillaItems::GLASS_BOTTLE(), $returnedItems);
 			}
-		}elseif($item instanceof Armor){
+		}elseif($item instanceof Armor || $item instanceof HorseArmor){
 			if($this->customWaterColor !== null){
 				if(match($item->getTypeId()){ //TODO: a DyeableArmor class would probably be a better idea, since not all types of armor are dyeable
```
pandaaaBE, line 134: Could we just use Tags instead a new class for this TODO?

but that's nothing for this pr anyway

src/item/HorseArmor.php
IvanCraft623: it is unnecessary to add a new method when you can do the same thing with a single method, besides it is inconsistent for `getCustomColor()` to return `?Color`, and `setCustomColor()` to only accept `Color`.

pandaaaBE: Oops then someone screwed up badly at Armor.php if you have this opinion.

But I'm happy to change it for you, but as I said, I think it's unnecessary

ShockedPlot7560: Its just old code, the reflexion is not new, but to remove all the clear* shit method we need to break BC

pandaaaBE: jo all good has already been adjusted

src/block/WaterCauldron.php
```diff
@@ -128,9 +129,10 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+					ItemTypeIds::LEATHER_HORSE_ARMOR,
```
ShockedPlot7560: Alphabetical order would be preferable

src/item/HorseArmor.php
```diff
@@ -0,0 +1,75 @@
...
+		}else{
+			$this->customColor = null;
+		}
+	}
+
+	protected function serializeCompoundTag(CompoundTag $tag) : void{
+		parent::serializeCompoundTag($tag);
+		$this->customColor !== null ?
+			$tag->setInt(self::TAG_CUSTOM_COLOR, Binary::signInt($this->customColor->toARGB())) :
+			$tag->removeTag(self::TAG_CUSTOM_COLOR);
```
ShockedPlot7560: Better to use real if statement here. More readable

pandaaaBE: https://github.com/search?q=repo%3Apmmp%2FPocketMine-MP%20!%3D%3D%20null%20%3F&type=code

the code exists in this form or in a similar form more often in the src...

src/block/WaterCauldron.php
pandaaaBE: Are you sure? There is nothing in the alphabetical order I would have to change the order completely and then it says irrelevant change

pandaa-be: `					ItemTypeIds::LEATHER_HORSE_ARMOR,
					ItemTypeIds::LEATHER_CAP,
					ItemTypeIds::LEATHER_TUNIC,
					ItemTypeIds::LEATHER_PANTS,
					ItemTypeIds::LEATHER_BOOTS => true,`

pandaaaBE:
I think this PR is finished

src/item/HorseArmor.php
```diff
@@ -0,0 +1,77 @@
...
+	/**
+	 * Sets the dyed colour of this armour piece. This generally only applies to leather armour.
+	 *
+	 * @return $this
+	 */
+	public function setCustomColor(?Color $color) : self{
+		if($this->getTypeId() === ItemTypeIds::LEATHER_HORSE_ARMOR){
+			$this->customColor = $color;
+		}
+		return $this;
```
ShockedPlot7560: Wouldn't be better to raise an exception instead just silently ignore ?

pandaaaBE: `throw new Expection("Custom color can only be applied on Leather armor")`

Smth like this?

ShockedPlot7560: maybe?

ipad54: Maybe create a separate class for leather horse armour?

pandaaaBE: I think it's a bad idea to have a separate class just because of the custom color, there are enough anyway.

but that's not for me to decide, I'd like to have a second opinion on it

pandaaaBE: I could rename this file to LeatherHorseArmor, the other armors doesn't need the class anyway.

Is there a way to add a item in vanillablocks with maxstacksize 1 without to create a new class?

ShockedPlot7560: After checked the actual Armor class, it seems an unnecessary headache to ban coloring for non leather

nicholass003: yup, i think it's better, throw exception if non leather

jasonw4331: Only leather horse armor needs a class. The others can use Item objects

src/item/VanillaItems.php
```diff
@@ -666,6 +673,13 @@ private static function registerArmorItems() : void{
 		self::register("netherite_leggings", new Armor(new IID(Ids::NETHERITE_LEGGINGS), "Netherite Leggings", new ArmorTypeInfo(6, 556, ArmorInventory::SLOT_LEGS, 3, true, material: ArmorMaterials::NETHERITE()), [EnchantmentTags::LEGGINGS]));
 	}
 
+	private static function registerHorseArmorItems() : void{
+		self::register("leather_horse_armor", new HorseArmor(new IID(Ids::LEATHER_HORSE_ARMOR), "Leather Horse Armor"));
+		self::register("iron_horse_armor", new HorseArmor(new IID(Ids::IRON_HORSE_ARMOR), "Iron Horse Armor"));
+		self::register("diamond_horse_armor", new HorseArmor(new IID(Ids::DIAMOND_HORSE_ARMOR), "Diamond Horse Armor"));
+		self::register("golden_horse_armor", new HorseArmor(new IID(Ids::GOLDEN_HORSE_ARMOR), "Golden Horse Armor"));
+	}
+
```
jasonw4331: Only Leather Horse armor implements special logic, making 3 of these registrations contain unused logic and this entire function redundant. The other armors should be simple Item classes.

jasonw4331 (changes requested):
Only leather armor needs its own class.

src/item/HorseArmor.php
```diff
@@ -0,0 +1,77 @@
...
+	}
```
jasonw4331: I'd like to see further implementation similar to that of the Armor class with `clearCustomColor()` for consistency.
https://github.com/pmmp/PocketMine-MP/blob/68d2e2915ecd138c1eb4c4eff41eed40584c388f/src/item/Armor.php#L86-L107

jasonw4331 (changes requested):
~~This implementation is missing interaction with the Cauldron. See The interaction between Armor and the cauldron here for an implementation guide: https://github.com/pmmp/PocketMine-MP/blob/stable/src/block/WaterCauldron.php#L132-L148~~

pandaaaBE:
> This implementation is missing interaction with the Cauldron. See The interaction between Armor and the cauldron here for an implementation guide: https://github.com/pmmp/PocketMine-MP/blob/stable/src/block/WaterCauldron.php#L132-L148

wdym? it is implemented

image

src/item/HorseArmor.php
pandaaaBE: I had it exactly like that and the others wanted me to do it differently because the code in Armor is already old and we should do the new items right away.

src/item/VanillaItems.php
pandaaaBE: The default stack size is 64 so I would need a class anyway

jasonw4331: An inline class works just as well. The precedent would be the way netherite ingots have fireproofing.
```php
new class(new IID(Ids::IRON_HORSE_ARMOR), "Iron Horse Armor") extends Item {
	public function getMaxStackSize() : int{ return 1; }
}
```

pandaaaBE: let's wait and see what nico and plot think about this

pandaaaBE:
Now I have renamed the old HorseArmor.php file to LeatherHorseArmor.php xd

ShockedPlot7560 (commented):
I'm still thinking about the fact that basic armour has no way of being easily detected. If a plugin wants to add the horse interface for example, it will have to hardcode all the ids instead of just checking a class for example? I quite agree that this makes a useless class for nothing, but is it really that useless?

pandaaaBE:
> I'm still thinking about the fact that basic armour has no way of being easily detected. If a plugin wants to add the horse interface for example, it will have to hardcode all the ids instead of just checking a class for example? I quite agree that this makes a useless class for nothing, but is it really that useless?

But without joking, is that a troll of yours?

pandaaaBE:
So, I asked again on the Discord and an extra class for the Horse Armor doesn't seem to be necessary - although the duplicated code is annoying...

If you still have any wishes, please let me know

PR #2166, opened 2018-05-04
Implementation of trade

Frago9876543210:
## Introduction
This PR adds the trade with support backward compatibility.

Sreenshots:
i1
i2

### API changes

Example:
```php
$villager = new Villager($sender->level, Villager::createBaseNBT($sender));
$villager->setCanTrade(true);
$villager->setTraderName(TextFormat::RED . "Testing");
$villager->setRecipes(
	new TradeRecipe(Item::get(Item::IRON_AXE), Item::get(Item::EMERALD)),
	new TradeRecipe(Item::get(Item::MINECART), Item::get(Item::DIAMOND), Item::get(Item::GOLD_INGOT))
);
$villager->spawnToAll();
```

## Backwards compatibility
Yes

## Tests
Works great.

PR #6310, opened 2024-03-26
Implement Villager v2 and trade functionality

alvin0319:
## Introduction

This PR implements Villager V2 and trade functionality.

### Relevant issues

* #2166 

### API changes

* Added VillagerV2:
  * Villager V2 has 15 variants: Unemployed, Farmer, Fisherman, Shepherd, Fletcher, Librarian, Cartographer, Cleric, Armorer, Weaponsmith, Toolsmith, Butcher, Leatherworker, Mason (aka Stone Mason), Nitwit
  * Unemployed and Nitwit villager can't trade, so they don't have recipes.
  * Villagers now have their recipes corresponding their professions (except unimplemented items), following [this wiki](https://minecraft.wiki/w/Trading#Bedrock_Edition_offers).
  * `VillagerV2::setProfession(VillagerProfession)`: Sets a profession of this villager, also updates recipes.
  * `VillagerV2::getProfession() : VillagerProfession`: Returns the current profession of the villager.
  * `VillagerV2::getRecipeData() : RecipeData`: Returns the recipe data of this villager.
* Added VillagerProfession and its related friends:
  * Each professions have their corresponding translation key (for their name), variant id and job block.
* Added TradeRecipeData:
  * TradeRecipeData contains necessary values for trade: recipes, current tier, trader experience, tier experience requirements
* Added TradeRecipe:
  * buyA: Item that trader wants.
  * sell: Item that trader gives to you.
  * buyB: Item that trader additionally wants, optional.
  * maxUses: Max uses of this trade. 16 is default.
  * priceMultiplier: affects price of item, not used currently. 0.05 is default.
  * rewardExp: Reward exp that player earns by completing this trade. 0 is default.
  * tier: Required tier for this trade. 0 is default.
  * traderExp: Reward exp for the villager. 0 is default.
  * uses: Amount of current uses of this recipe. 0 is default.
### Behavioural changes

Villager Spawn Egg will now spawn Villager V2 instead of legacy Villager.
## Backwards compatibility

There is no known BC breaking change.
## Follow-up

N/A

## Tests

- [ ] Other (provide details)

https://github.com/pmmp/PocketMine-MP/assets/32565818/0a892342-12f7-4f3d-aacb-2d794f8693cf

Plugin used for this test:
```php
<?php

/**
 * @name TradeTest
 * @author alvin0319
 * @main alvin0319\TradeTest\TradeTest
 * @version 1.0.0
 * @api 5.0.0
 */

declare(strict_types=1);

namespace alvin0319\TradeTest;

use pocketmine\entity\Location;
use pocketmine\entity\trade\TradeRecipe;
use pocketmine\entity\trade\TradeRecipeData;
use pocketmine\entity\Villager;
use pocketmine\inventory\TradeInventory;
use pocketmine\player\Player;

final class TradeTest extends PluginBase{

		$this->getServer()->getPluginManager()->registerEvent(PlayerJoinEvent::class, fn(PlayerJoinEvent $event) => $this->sendTrade($event->getPlayer()), EventPriority::NORMAL, $this);
	}

	/**
	 * This trade UI implements some offers of Fisherman in Bedrock Edition
	 * @see https://minecraft.wiki/w/Trading#Fisherman_2
	 */
	private function sendTrade(Player $player) : void{
		// spawn villager entity 1.5 block above the player
		$villager = new Villager(
			Location::fromObject(
				$player->getDirectionVector()->multiply(1.5)->addVector($player->getPosition()),
				$player->getWorld()
			)
		);
		$villager->spawnToAll();
		$inventory = new TradeInventory("Trade Test", $villager, new TradeRecipeData([
			new TradeRecipe(
				buyA: VanillaItems::STRING()->setCount(20),
				sell: VanillaItems::EMERALD()->setCount(1),
				maxUses: 16,
				tier: 0,
				traderExp: 2
			),
			new TradeRecipe(
				buyA: VanillaItems::EMERALD()->setCount(3),
				sell: VanillaItems::COOKED_FISH()->setCount(1),
				buyB: VanillaItems::RAW_FISH()->setCount(6),
				maxUses: 16,
				tier: 0,
				traderExp: 1
			),
			new TradeRecipe(
				buyA: VanillaItems::RAW_FISH()->setCount(15),
				sell: VanillaItems::EMERALD()->setCount(1),
				maxUses: 16,
				tier: 1,
				traderExp: 10
			),
			new TradeRecipe(
				buyA: VanillaItems::EMERALD()->setCount(1),
				sell: VanillaItems::COOKED_SALMON()->setCount(6),
				buyB: VanillaItems::RAW_SALMON()->setCount(6),
				maxUses: 16,
				tier: 1,
				traderExp: 5
			),
		], 1));
		$player->setCurrentWindow($inventory);
	}
}
```


ShockedPlot7560:
Do we really want to force an API to everyone ? I can only think about the Form API. Wouldn't it be the function of a virion to do so?

Matze997:
On the other hand people will know as long as pocketmine gets updated this feature will be supported

alvin0319:
> Do we really want to force an API to everyone ? I can only think about the Form API. Wouldn't it be the function of a virion to do so?

I don't think it is possible to implement this with virion so.

alvin0319:
Now ready to be reviewed.

pandaa-be:
I don't like that it's just an API instead of building in the vanilla system, an API is more of a plugin thing in my opinion.

src/entity/VillagerV2.php
```diff
@@ -0,0 +1,183 @@
...
+	protected function getInitialSizeInfo() : EntitySizeInfo{
+		return new EntitySizeInfo(1.9, 0.6);
+	}
+
+	protected function initEntity(CompoundTag $nbt) : void{
+		parent::initEntity($nbt);
+
+		$professionId = $nbt->getInt(self::TAG_PROFESSION, -1);
+		if($professionId === -1){
+			$professionId = random_int(VillagerProfessionTypeIds::UNEMPLOYED, VillagerProfessionTypeIds::NITWIT);
```
IvanCraft623: Do not get a random profession like this, and avoid using `random_int`, use `mt_rand` instead

src/entity/VillagerV2.php
```diff
@@ -0,0 +1,183 @@
...
+		return $this->baby;
+	}
+
+	public function setBaby(bool $baby) : void{
+		$this->baby = $baby;
+		$this->networkPropertiesDirty = true;
+	}
+
+	public function onInteract(Player $player, Vector3 $clickPos) : bool{
+		if(!$this->profession->canTrade()){
```
IvanCraft623: `!isBaby` check missing

src/entity/VillagerV2.php
```diff
@@ -0,0 +1,183 @@
...
+		$offers->setTag(self::TAG_RECIPES, new ListTag($recipesTag));
+		$tierExpRequirements = [];
+		foreach($this->recipeData->getTierExpRequirements() as $tier => $expRequirement){
+			$tierExpRequirements[] = CompoundTag::create()
+				->setInt((string) $tier, $expRequirement);
+		}
+		$offers->setTag(self::TAG_TIER_EXP_REQUIREMENTS, new ListTag($tierExpRequirements));
+		$nbt->setTag(self::TAG_OFFERS, $offers);
+		return $nbt;
+	}
```
IvanCraft623: Please use line spacing for better understanding.

src/entity/profession/ArmorerProfession.php
```diff
@@ -0,0 +1,187 @@
...
+use pocketmine\block\VanillaBlocks;
+use pocketmine\data\bedrock\VillagerProfessionTypeIds;
+use pocketmine\entity\trade\TradeRecipe;
+use pocketmine\item\VanillaItems;
+use function random_int;
+
+final class ArmorerProfession extends VillagerProfession{
+
+	public function __construct(){
+		parent::__construct(VillagerProfessionTypeIds::ARMORER, "entity.villager.armor", VanillaBlocks::BLAST_FURNACE());
```
IvanCraft623, line 33: This is a translation key, make a PR in https://github.com/pmmp/Language to later use a constant

src/entity/profession/ArmorerProfession.php
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	/** @phpstan-return list<TradeRecipe> */
+	public function getRecipes(int $biomeId) : array{
```
IvanCraft623, line 37: In vanilla only a few recipes are selected per group, see https://github.com/Mojang/bedrock-samples/blob/main/behavior_pack/trading/armorer_trades.json, for more details. Probably with a function that generates recipes is not enough and will have to be extended to a dedicated class.

src/entity/trade/TradeRecipe.php
```diff
@@ -0,0 +1,156 @@
...
+	public const TAG_SELL = "sell";
+	public const TAG_REWARD_EXP = "rewardExp";
+	public const TAG_TIER = "tier";
+	public const TAG_TRADER_EXP = "traderExp";
+	public const TAG_MAX_USES = "maxUses";
+	public const TAG_PRICE_MULTIPLIER_A = "priceMultiplierA";
+	public const TAG_PRICE_MULTIPLIER_B = "priceMultiplierB";
+	public const TAG_USES = "uses";
+
+	public const TAG_NET_ID = "netId";
```
IvanCraft623: Tag constans must be private if they are not needed outside the class, plugins should not mess with NBT anyway.

src/entity/trade/TradeRecipe.php
```diff
@@ -0,0 +1,156 @@
...
+		return new self(
+			$buyA,
+			$sell,
+			$buyB,
+			$nbt->getInt(self::TAG_MAX_USES, 16),
+			$nbt->getFloat(self::TAG_PRICE_MULTIPLIER_A, 0.05),
+			$nbt->getInt(self::TAG_REWARD_EXP, 0),
+			$nbt->getInt(self::TAG_TIER, 0),
+			$nbt->getInt(self::TAG_TRADER_EXP, 0),
+			$nbt->getInt(self::TAG_USES, 0)
```
IvanCraft623: These default values are used several times, putting them in a constant would be fine

src/entity/trade/TradeRecipeData.php
```diff
@@ -0,0 +1,89 @@
...
+ * @link http://www.pocketmine.net/
+ *
+ *
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\entity\trade;
+
+final class TradeRecipeData{
```
IvanCraft623, line 26: I don't like the name of the class, it gives me to understand that it contains the data of a single recipe, or that it's for data processing, maybe something like TradeRecipeManager, TradeRecipeGroup,  or TradeRecipesHolder would be better.

src/event/player/PlayerItemTradeEvent.php
```diff
@@ -0,0 +1,54 @@
...
+	public function __construct(
+		Player $player,
+		private readonly Item $buyA,
+		private readonly Item $sell,
+		private readonly ?Item $buyB = null
+	){
+		$this->player = $player;
+	}
+
+	public function getBuyA() : Item{
```
IvanCraft623, line 46: doc what's a buyA, for plugin developers

src/event/player/PlayerItemTradeEvent.php
```diff
@@ -0,0 +1,54 @@
...
+		return clone $this->buyA;
+	}
+
+	public function getBuyB() : ?Item{
```
IvanCraft623, line 53: Same here

src/inventory/EntityInventory.php
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\inventory;
+
+use pocketmine\entity\Living;
+use pocketmine\network\mcpe\protocol\ClientboundPacket;
+
+abstract class EntityInventory extends SimpleInventory{
```
IvanCraft623, line 29: What other uses would this class have?

IvanCraft623 (commented):
A quick review, I still have to validate with BDS behavior

src/entity/profession/ArmorerProfession.php
alvin0319: This is a client-side translation key, so I don't think PM-side translations are needed?

src/inventory/EntityInventory.php
alvin0319: I'm not sure if there are any other use cases (including future), as far as I know trade is the only one which doesn't use ContainerOpen for opening inventory.

src/entity/profession/ArmorerProfession.php
alvin0319: So as far as I understand, (also according to [bedrock.dev wiki](https://wiki.bedrock.dev/loot/trade-tables.html#trade-groups)), recipes are separated something called "groups" and they are randomly selected whenever trade recipes are generated, right? Perhaps we could make a script that generates recipe class from behaviour pack.

src/entity/profession/ArmorerProfession.php
IvanCraft623: Avoid hardcode translations, this is not the only place where translation is only used client-side, e.g. names of enchantments.

src/entity/profession/ArmorerProfession.php
IvanCraft623: Yes, it would be best if this is self-generated. I think that for now we only need to make an api to implement trade tables with their groups and leave the generation of recipes for another PR.

IvanCraft623:
The pull request is very bloated by the recipes generation, so perhaps it would be nice to leave them for a separate PR

src/entity/trade/TradeRecipeData.php
alvin0319: I'm still not sure about this class name. TradeRecipeManager feels like it's too generic (which people might understand it manages all trade recipes), TradeRecipeGroup doesn't seem valid since it doesn't only contain trade recipes. TradeRecipesHolder might be good but I think we still need better name than these. (or separate other values like tier, exp to other class?)

ShockedPlot7560:
At the moment I'm still dubious about the fact that the API has been implemented right down to the NPC itself. 

Is this a mission of pocketmine itself, or in the same way as the forms, we should let the community create libraries instead of imposing one.

alvin0319:
It isn't same with forms. Trade needs a lot of reflection hacks and closure hacks to work on plugins.

pandaaaBE:
> At the moment I'm still dubious about the fact that the API has been implemented right down to the NPC itself. 
> 
> 
> 
> Is this a mission of pocketmine itself, or in the same way as the forms, we should let the community create libraries instead of imposing one.

by the way, villager trading is not only an api but also implements the normal trades now, unlike before.

i find it nice that the pr also offer an api to be able to make trades, if there were several different plugins the support on the discord server would be annoying just like with scoreboard, bossbar and forms apis - it's annoying if someone uses jojo's api but you don't know the api yourself it would be better if there was one in the core that everyone uses

alvin0319:
Closing due to lack of activity and time - Will be bringing another PR if i get enough time and proper implementation for recipe generation.

Issue #6311, opened 2024-03-29
Implementing new blocks is a pain in the ass


dktapps:
## Description
The current design of the blocks system is currently disastrously complex and painful to maintain, and needs to be redesigned.

Many people have criticized the system (for the wrong reasons), but those people weren't entirely wrong. The current system is an awful mess, and it's not one of my finer works.

## Justification
The current system is awful to work with.

- [x] Information of how to translate NBT into PM internal structures is repeated twice per block
- [ ] We have to edit minimum 4 enormous files to add a new block (and that's before thinking about StringToItemParser et al)
- [ ] We end up with giant repeated blocks of crap like [this](https://github.com/pmmp/PocketMine-MP/blob/11f119551dec2771f0af8b9ae25e88a67d854b16/src/data/bedrock/block/convert/BlockStateToObjectDeserializer.php#L462) because there's no way to dynamically (e.g. fetch a certain type of wood block)
- [x] We end up with giant repeated blocks of crap like [this](https://github.com/pmmp/PocketMine-MP/blob/11f119551dec2771f0af8b9ae25e88a67d854b16/src/data/bedrock/block/convert/BlockStateToObjectDeserializer.php#L140) because we have no way to extract information from the ID, even when the PM internals support dynamically setting said information (e.g. color)
- [ ] Implementing a new block also requires manually declaring a new type ID in BlockTypeIds, adding it to VanillaBlocks, and adding it to StringToItemParser

It ended up this way because it was being designed all the way back in 2019, and took 3 years to come to fruition. I attempted to solve several other problems at the same time (such as the inability to provide dynamic APIs for stuff like glazed terracotta due to the 1:1 linking of block type <-> minecraft block id, the heavy dependence on minecraft internal meta). While this redesign was partially successful (e.g. runtime state data is now far easier to work with than in prior versions), it's undeniable that the current system is not nice for maintenance or extension.

For what it's worth, I knew that the system I started to build out in 2019 was going to be a hunk of shit, but I pushed through with it anyway once LB got behind me, because it needed to be done, and I knew I wouldn't be able to see how better to do it until making the first version.

What we gained from the new system:
- Internal system no longer directly depends on Mojang's awful mess of a block system, allowing us to implement nicer APIs (e.g. GlazedTerracotta->setColor())
- We don't have to break BC when Mojang keep changing the block IDs one by one each month (for the love of god, please just get it over and done with)
[...]

PR #6313, opened 2024-03-29
Auto generate BlockTypeIds & ItemTypeIds

dktapps:
## Introduction
Currently, implementing new items requires manual changes to both `VanillaItems` and `ItemTypeIds`.
The same is true for `VanillaBlocks` and `BlockTypeIds`.

This PR removes `VanillaItems` dependency on the `ItemTypeIds::*` constants and `VanillaBlocks` on `BlockTypeIds::*` constants.
This permits the type IDs to be defined by `VanillaItems` and `VanillaBlocks` dynamically, and then regenerating `BlockTypeIds` and `ItemTypeIds` is just an afterthought.

This change also makes the code of registering new blocks & items far less obnoxious to read and write, since the information is repeated less times, and the code isn't as wide.

### Problems
Adding a new block or item anywhere except right at the end of the `setup()` function will cause type IDs of existing blocks & items to change. This causes a few issues:
  - ~~Messy diffs (resolved by making const values relative to previous)~~
  - ~~`block_factory_consistency_check` can't properly detect changes anymore (needs to be rethought or possibly deleted)~~
  - Forgetting to regenerate the type IDs could cause wacky behaviour of vanilla features during local testing (automated tests will ensure this doesn't happen on production)

I haven't figured out how to deal with these issues yet. Perhaps the type IDs could be derived from the block type name or something of that nature (but this would be a bit more complicated to implement).

Despite these issues, I think this change is probably still worth making, for the reduction in tedious bullshit when implementing new blocks & items.

### Relevant issues
Partially addresses #6311

### API changes
None

### Behavioural changes
None

## Backwards compatibility
Values of type ID constants have changed. Plugins should not be storing these (as per the giant warning at the top of the files), but this will no doubt catch someone off guard anyway.

## Follow-up
Consider if we can get rid of the type ID constants entirely. Perhaps redesigning `VanillaBlocks` might provide a way to do this.

I've considered the possibility of using enum objects instead of type IDs, which would have various benefits.

## Tests
CI tests are sufficient


build/generate-type-id-consts.php
```diff
@@ -0,0 +1,119 @@
...
+require dirname(__DIR__) . '/vendor/autoload.php';
+
+/**
+ * @param int[]    $stringToTypeIdMap
+ * @param string[] $backwardsCompatibilityConstants
+ *
+ * @phpstan-param array<string, int> $stringToTypeIdMap
+ * @phpstan-param array<string>      $backwardsCompatibilityConstants
+ */
+function patchTypeIds(string $file, array $stringToTypeIdMap, array $backwardsCompatibilityConstants) : void{
```
ShockedPlot7560, line 59: What's the point of getting a TypeIdMap here, the name isn't enough?

dktapps: The values are generated by VanillaBlocks, so we need to receive them here to generate the constant values

ShockedPlot7560: Oh yes, to order them, I've got it

dktapps:
Only remaining issue is how to avoid wacky runtime bugs when type IDs get changed due to adding new things

PHPUnit tests will ensure that we don't push broken type IDs to a release, but it could still cause some WTFs in local testing and I'm not sure what should be done to avoid that

dktapps:
It seems that longer term, it would make more sense to use strings for type IDs instead of integers. The names used for `VanillaBlocks` and `VanillaItems` could be used directly as type IDs.

We'd more or less end up with the same ID system that Minecraft has natively (however, we wouldn't be able to use `minecraft:` IDs directly since the vanilla IDs don't directly map onto PM blocks due to differences in implementation, such as dynamic colour and other nice APIs that avoid Mojang inconsistencies.)

This would have a few benefits
- No need for generating anything
- Easier to transmit between threads (no need to manually claim IDs on every thread)
- Far easier for custom blocks to avoid collisions
- Less variables needed for registering new blocks

Downsides:
- Plugins would be able to hardcode IDs again
- Plugins would be able to make assumptions about the structure of ID values

We would have to dynamically compress these string IDs to integers at runtime for storing on chunks, but that would be far cleaner than the clusterfuck we currently have.

However, with all that said, we won't be able to do a change like this until PM6 due to the BC breaks required.

dktapps:
Probably makes sense to separate the reflection type ID linking from this PR, since that's useful by itself and doesn't change anyone's workflow. Autogenerating type IDs doesn't have such an obvious benefit.

dktapps:
Closing in favour of #6498

PR #6333, opened 2024-04-19
Async task timings

### LINKED ISSUE #6166: AsyncTask timings
## Description
It should be possible to add separate tables on timings to insert timings from async workers.
These would be separate from Full Server Tick, so they wouldn't be counted towards main server load.

This would give a more complete picture of server performance.

dktapps:
## Introduction
This PR adds support for asynchronously collecting timings, and for collecting timings from AsyncWorkers for AsyncTasks.

The implemented API allows custom collect callbacks to be registered, which allows plugins to include custom thread timings in timings reports. This is necessary because other threads' timings records can't be directly accessed from the main thread.

### Relevant issues
Closes #6166 

### API changes
- `TimingsHandler::printTimings()` is now deprecated
- Added `getToggleCallbacks()`, `getReloadCallbacks()` and `getCollectCallbacks()` to `TimingsHandler`

### Behavioural changes
- Timings reports now include AsyncTask timings.
- Timings may now take slightly longer to generate as the report generator has to wait for other threads to respond.
- Timings report version 3 is now used. This tells the timings viewer to use the `ThreadId` part of timer group names to namespace timer IDs (necessary since timer IDs could otherwise collide between different threads).

## Backwards compatibility
Should be fully backwards compatible

## Tests

- [x] Testing via command and pasting reports

## Samples
This requires a version of the timings viewer including the following commit: https://github.com/pmmp/timings/commit/13cefa6279ee8866dc584b9a7977a9c09cecd732

An example report can be viewed here: https://timings.pmmp.io/?id=338211


dktapps:
There are some unrelated changes in the PR which I made to facilitate this feature, such as `Promise::all()` supporting 0 promises. These should be separated before merging.

dktapps:
This is finished for now.

However, it would be worth making some changes to RakLib to enable some basic timings to be collected from it too. This would probably involve making [`Server->tick()`](https://github.com/pmmp/RakLib/blob/abbb8ed48f709adf51d6998f913915a3aa787277/src/server/Server.php#L167) and [`Server->receivePacket()`](https://github.com/pmmp/RakLib/blob/abbb8ed48f709adf51d6998f913915a3aa787277/src/server/Server.php#L202) protected to allow a PM class to extend and override them for timings encapsulation. A custom `ServerEventSource` to do timings would probably be a good idea too.

Currently RakLib's architecture (and the IPC system) are not flexible enough for timings integration, so they'll be skipped for this PR. This should be done as a follow-up.

PR #6345, opened 2024-05-12
WoodenStairs can be a fuel 

DebHugo:
## Introduction
Wooden Stairs can be used as a fuel in furnaces
source: https://minecraft.wiki/w/Wooden_Stairs#Fuel

### API changes
Modify `getFuelTime` function in `WoodenStairs`


dktapps:
Weird that WoodTypeTrait doesn't cover this. Maybe someone should look into that.

PR #6347, opened 2024-05-14
Add onEntityCollide function when entity are above a block

### LINKED ISSUE #2041: Magma blocks don't work any more

DebHugo:
## Introduction
PocketMine don't have function to detect if an entity are above a block.
So i add a function and fix the bug with magma block.
But there are other bug of this type (with top of cactus...), so i must to be fix in this pr or open an other pr to fix them ?

### Relevant issues

Fix #2041 

### API changes

Add `Block::onEntityCollide()`

## Tests

https://github.com/pmmp/PocketMine-MP/assets/55756021/b196bbad-4dce-4118-9a2e-d1546f597df0


IvanCraft623:
This does not really check if there is a collision so if there is a slab between the magma block and the player, the player will be burned.

DebHugo:
> This does not really check if there is a collision so if there is a slab between the magma block and the player, the player will be burned.

Thanks for the feedback, it's fixed

src/block/Block.php
```diff
@@ -868,6 +868,12 @@ public function onEntityInside(Entity $entity) : bool{
 		return true;
 	}
 
+	/**
+	 * Called when an entity's is above this block
+	 */
+	public function onEntityCollide(Entity $entity) : void{
```
IvanCraft623: It would be useful to pass the colliding face as an argument as well.

IvanCraft623 (dismissed):
The current implementation only works for collsions on down side. An appropriate solution to the problem would be to detect when colliding with any face of the block, not just one.

DebHugo:
> The current implementation only works for collsions on down side. An appropriate solution to the problem would be to detect when colliding with any face of the block, not just one.

Fixed

src/block/Magma.php
```diff
@@ -27,23 +27,19 @@
...
 	}
 
-	public function hasEntityCollision() : bool{
-		return true;
-	}
-
-	public function onEntityInside(Entity $entity) : bool{
-		if($entity instanceof Living && !$entity->isSneaking()){
+	public function onEntityCollide(Entity $entity, int $face) : void{
+		if($face === Facing::DOWN && $entity instanceof Living && !$entity->isSneaking()){
```
ShockedPlot7560: Isn't that trivial? If the entity is above the block, we get damage and the face is below. Have I misunderstood something?

DebHugo: No, that's good.
I add the condition ``$face === Facing::DOIWN`` after suggesting @IvanCraft623 to allow detection for each face and not just downwards.
Without this verification, magma blocks cause damage on all faces and not only on the down face

ShockedPlot7560: Perhaps we've misunderstood each other, but what I mean is that the $face parameter located in the block class subject to be the face of the collided block and not the "face" of the collided entity (which is what I'm supposed to be now).

DebHugo: Indead, i fix this

src/entity/Entity.php
```diff
@@ -658,6 +658,18 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
...
 			$this->attack($ev);
 			$hasUpdate = true;
+		}else{
+			$entityBlock = $this->getWorld()->getBlock($this->getLocation());
+			$entityBox = $this->getBoundingBox()->expandedCopy(0.001, 0.1, 0.001);
+			foreach (Facing::ALL as $face) {
+				$block = $entityBlock->getSide($face);
+				foreach ($block->getCollisionBoxes() as $blockBox) {
+					if ($entityBox->intersectsWith($blockBox)) {
+						$block->onEntityCollide($this, Facing::opposite($face));
```
nicholass003: should this case add this?
```php 
$hasUpdate = true;
```

DebHugo: May be but if i realy good understand, `hasUpdate` variable is set to true only if entity is update. However, in function `onCollide`, we are not sure that entity is update.
So, i don't realy sure

ShockedPlot7560 (dismissed):
Look OK to me

src/block/Block.php
```diff
@@ -868,6 +868,12 @@ public function onEntityInside(Entity $entity) : bool{
 		return true;
 	}
 
+	/**
+	 * Called when an entity collide on this block
+	 */
+	public function onEntityCollide(Entity $entity, int $face) : void{
+	}
```
dries-c, line 880: you might want to add //NOOP since it's convention

dries-c (dismissed):

src/block/Block.php
DebHugo: which convention ?

nicholass003: > which convention ?

```php
	public function onEntityCollide(Entity $entity, int $face) : void{
	           //NOOP
	}
```

src/entity/Entity.php
dktapps: Normally we add `bool` returns to these kinds of hooks so that the caller knows whether or not something happened.

src/entity/Entity.php
```diff
@@ -658,6 +658,18 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
...
+						break;
```
dktapps: I'm not convinced we should break here. Entities can collide with multiple blocks at once.

dktapps (changes requested):
This might raise issues considering the name `onEntityCollide()` was used in previous PM versions for a different thing. We also still have `hasEntityCollision()` from those days, and people will almost definitely get confused about the usage of that with this new function in the mix.

My gut instinct is that a better name is probably called for.

src/entity/Entity.php
```diff
@@ -658,6 +658,18 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
 			$ev = new EntityDamageEvent($this, EntityDamageEvent::CAUSE_VOID, 10);
 			$this->attack($ev);
 			$hasUpdate = true;
+		}else{
+			$entityBlock = $this->getWorld()->getBlock($this->getLocation());
+			$entityBox = $this->getBoundingBox()->expandedCopy(0.001, 0.1, 0.001);
```
dktapps: Why is this different for the Y axis than X/Z?

ipad54:
> My gut instinct is that a better name is probably called for.

How about `onEntityStep()`?

dktapps:
Doesn't really fit since it'll be called by a collision from any side.

src/entity/Entity.php
DebHugo: If player collide with an AABB of block, we don't need to check all AABB of this block so we break. 
All faces are checked with the foreach above

dktapps: That may be the case right now as only magma uses this. But in the future more blocks might use it with different behaviours.

src/entity/Entity.php
DebHugo: Initially, I started testing with 0.001 for each axis but I found a bug on the Y axis with a slab. Indeed, if you put a block of magma under a slab and a player steps on this slab, with 0.001 on the Y axis, the collision is not detected so I put 0.1 to detect it

DebHugo:
> We also still have `hasEntityCollision()` from those days

Why not keep this name and call this function just if `hasEntityCollision` return `true` ?
Moreover, this can be increase performance

src/entity/Entity.php
DebHugo: For exemple, if we take cauldron that return multiple AABB in `getCollisionBoxes()`. If player collide with two AABB for the same face, we have two `onEntityCollide` call for the same face. It's illogic

src/entity/Entity.php
dktapps: That doesn't make any sense. Is magma supposed to hurt you from underneath a slab? And if it is, why does 0.1 make a difference since the BB of a slab is 0.5 thick?

dktapps:
> > We also still have `hasEntityCollision()` from those days
> 
> Why not keep this name and call this function just if `hasEntityCollision` return `true` ? Moreover, this can be increase performance

Because this will cause unexpected behaviour. No one expected these two functions to be bound together in the first place.

src/entity/Entity.php
IvanCraft623: @dktapps He's breaking `foreach ($block->getCollisionBoxes() as $blockBox) {`, in case a collision with that block has already been found, `foreach (Facing::ALL as $face) {` will continue iterating.

dktapps: Right. In that case it should be using this: https://github.com/pmmp/PocketMine-MP/blob/9cab72ed5907b3d26040e88f89b7353b0de03ec3/src/block/Block.php#L841

src/entity/Entity.php
```diff
@@ -666,6 +666,7 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
 				foreach ($block->getCollisionBoxes() as $blockBox) {
 					if ($entityBox->intersectsWith($blockBox)) {
 						$block->onEntityCollide($this, Facing::opposite($face));
+						$hasUpdate = true;
```
dktapps, line 670: This shouldn't be unconditionally set. You should add a return value to `onEntityCollide` so it knows whether to set this or not.

dktapps: I also just realized this is in entityBaseTick...

src/entity/Entity.php
```diff
@@ -665,8 +665,7 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
 				$block = $entityBlock->getSide($face);
 				foreach ($block->getCollisionBoxes() as $blockBox) {
 					if ($entityBox->intersectsWith($blockBox)) {
-						$block->onEntityCollide($this, Facing::opposite($face));
-						$hasUpdate = true;
+						$hasUpdate = $block->onEntityCollide($this, Facing::opposite($face));
```
dktapps: Don't do this. If the block isn't updated, it'll trash any previous value for $hasUpdate. (This is also why I questioned doing this in entityBaseTick...)

src/entity/Entity.php
DebHugo: After some reseach, player take damage on magma block if block between player and magma block has AABB below or equal to 13/16 like trapdoor (https://minecraft.wiki/w/Magma_Block#Damage). So the y value is 1-13/16 = 3/16.

dktapps: Then it seems like a collide function isn't appropriate for this. We need something more like onEntityNearby() or onEntityAdjacent().

dktapps (changes requested):
It looks like what we actually need for this is a function that gets called when an entity is inside an adjacent block. Then, it should be up to the block to decide how much distance is allowed between the entity and the block to apply the effects.

dktapps:
A little update on this: I attempted to implement this with an "adjacent" check, but it turns out that the extended AABB collision scanning has a significant performance impact. I probably should've realized this sooner considering how much effort I previously invested into optimising AABB collision checking. Not sure what other options there are except specializing for this particular case.

DebHugo:
I tried to create an `onEntityInNearbyBlock` that is called on every tick and implements the collision logic in that function for each block. I stopped because I'm not convinced by this solution and I can't find a better solution without performance issues

PR #6350, opened 2024-05-23
fix : syntax error in switch case statement

nicholass003:
## Introduction

The pull request improves error handling by throwing exceptions directly.

## Tests

- [ ] Other (provide details)


src/data/bedrock/block/convert/BlockStateToObjectDeserializer.php
```diff
@@ -1437,7 +1437,7 @@ private function registerDeserializers() : void{
 					$in->ignored(StateNames::PILLAR_AXIS);
 					return Blocks::SMOOTH_QUARTZ();
 				default:
-					return throw $in->badValueException(StateNames::CHISEL_TYPE, $type);
```
SOF3, line 1440: `throw $value` is an expression since recent PHP versions, so this is technically not a syntax error. But the `return` part would be dead code indeed.

PR #6351, opened 2024-05-25
Adjust pretty name of closures on PHP 8.4

staabm:
## Introduction

related to https://github.com/php/php-src/pull/13550

see analog symfony change: https://github.com/symfony/symfony/pull/54614

## Tests

- [ ] Other (provide details)

### DIFF
--- a/src/utils/Utils.php
+++ b/src/utils/Utils.php
@@ -80,7 +80,7 @@ use function preg_match_all;
 use function preg_replace;
 use function shell_exec;
 use function spl_object_id;
-use function str_ends_with;
+use function str_contains;
 use function str_pad;
 use function str_split;
 use function str_starts_with;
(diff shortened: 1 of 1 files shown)


ShockedPlot7560:
PHP 8.4 is not officially supported

staabm:
I was just crawling thru github for this particular php compat problem. 

Its not important for me to make this repo PHP 8.4 compatible

dktapps:
I think this would be better filled by using [`ReflectionFunction::isAnonymous`](https://www.php.net/manual/en/reflectionfunction.isanonymous.php) in 8.2 and up

staabm:
> I think this would be better filled by using [`ReflectionFunction::isAnonymous`](https://www.php.net/manual/en/reflectionfunction.isanonymous.php) in 8.2 and up

thats right, but it would require raising the [min-php version requirement](https://github.com/pmmp/PocketMine-MP/blob/08c6e63aac45a450fe3b437616ee8aa2500e04e2/composer.json#L8)

dktapps:
Not if we gate it with a `PHP_VERSION_ID` check. This change makes me uneasy. This check was sketchy enough to begin with.

dktapps:
I don't like this, but it's less work than trying to get PHPStan to not complain about any version-conditional logic I tried to implement...

Issue #6324, opened 2024-04-05
Automatically declare parent permission groups


dktapps:
## Description
Currently, the permission system doesn't make any assumptions about the strings used to identify permissions. However, in practice, permissions typically look like this: `pocketmine.command.gamemode.self`

However, if trying to set the permission `pocketmine.command.gamemode` to override all of the sub permissions, nothing will happen.

This issue proposes that `.` should be considered as a hierarchy separator by the internal system, and parent permission nodes automatically declared to allow users to get the expected behaviour.

This means that registering `pocketmine.command.gamemode.self` would generate the following hierarchy:

```
`pocketmine`
-> `pocketmine.command`
  -> `pocketmine.command.gamemode`
    -> `pocketmine.command.gamemode.self`
```

dktapps:
This would also be useful for backwards compatibility purposes. Currently, if we want to divide an existing permission into sub permissions, the original permission declaration has to be explicitly retained, as seen here: https://github.com/pmmp/PocketMine-MP/commit/567bd8abb5dfbd4a035a081663cba820f45887e7#diff-28f7c5edac5d9f0ad9a1eabb7aada678eb60c4d2f8ca5da16a68acad45fb724c

This shouldn't be necessary, as the "deprecated" permission should be auto declared.

PR #6357, opened 2024-06-08
fix: version command issues

### LINKED ISSUE #6322: "/ver", "/version" and "/about" commands issue.
When a player runs "/pl" or "/plugins", they can't see the plugin list because they don't have perms for that. But what's the issue is that when a player runs one of those 3 commands ("/ver", "/version" or "/about") and they type a random letter behind that, they can see all plugins and info about the plugins that contain the letter they write behind that.

nicholass003:
## Introduction

This pull request is to handle users using the version command for detailed things, such as checking plugin information.

### Relevant issues

Fixes https://github.com/pmmp/PocketMine-MP/issues/6322

### API changes

Added `DefaultPermissionNames::COMMAND_VERSION_DETAIL` -> Allows the user to view the detailed version of the server.

## Tests

- [ ] Other (provide details)


src/command/defaults/VersionCommand.php
```diff
@@ -82,6 +82,10 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
 			$pluginName = implode(" ", $args);
 			$exactPlugin = $sender->getServer()->getPluginManager()->getPlugin($pluginName);
 
+			if(!$this->testPermission($sender, DefaultPermissionNames::COMMAND_VERSION_DETAIL)){
+				return true;
+			}
```
ShockedPlot7560: Put this near the else no ?

nicholass003: > Put this near the else no ?

yes your'e right, now it's correct

ShockedPlot7560 (dismissed):
Need translation follow up

nicholass003:
> Need translation follow up

just added english for now, https://github.com/pmmp/Language/pull/182

dktapps (changes requested):
I think "detailed version" and the general "detail" wording is too vague. Be more specific - this is for restricting access to plugin versions.

nicholass003:
> I think "detailed version" and the general "detail" wording is too vague. Be more specific - this is for restricting access to plugin versions.

Is there some suitable name for this?
i have several names :
- `pocketmine.permission.command.version.plugins` - Allows the user to view information about all server plugins.
- `pocketmine.permission.command.version.info` - Allows the user to view detailed information about all plugins used by the server.
- `pocketmine.permission.command.version.pluginInfo` - Allows the user to access information for all plugins on the server.

dktapps:
First one seems the most obvious to me. Don't overthink it.

dktapps:
that being said, those names look AI generated ...

nicholass003:
> that being said, those names look AI generated ...

for the desc yes 😁

src/permission/DefaultPermissions.php
```diff
@@ -106,6 +106,7 @@ public static function registerCorePermissions() : void{
 		self::registerPermission(new Permission(Names::COMMAND_UNBAN_IP, l10n::pocketmine_permission_command_unban_ip()), [$operatorRoot]);
 		self::registerPermission(new Permission(Names::COMMAND_UNBAN_PLAYER, l10n::pocketmine_permission_command_unban_player()), [$operatorRoot]);
 		self::registerPermission(new Permission(Names::COMMAND_VERSION, l10n::pocketmine_permission_command_version()), [$everyoneRoot]);
+		self::registerPermission(new Permission(Names::COMMAND_VERSION_PLUGINS, l10n::pocketmine_permission_command_version_plugins()), [$operatorRoot]);
```
dktapps, line 109: This will cause problems for people who set `pocketmine.command.version` expecting it to set this permission as well.

It's probably better to have:
- `pocketmine.command.version.server`
- `pocketmine.command.version.plugins`

and then have `pocketmine.command.version` as a deprecated backwards-compatible group permission for the remainder of PM5. See how this was done for `/effect` in PM4: https://github.com/pmmp/PocketMine-MP/blob/ba89ae5bf20dcc4aa4d40698429535aad26cc79d/src/permission/DefaultPermissions.php#L71-L73

This will permit people to still explicitly set `pocketmine.command.version` to get the same functionality as it would previously, but since it won't be given to any groups, default functionality will be controlled by these new permissions.

(Side note for maintainers: If #6324 were implemented, explicitly registering deprecated group roots wouldn't be necessary anymore.)

dktapps (changes requested):

dktapps:
Closing this as I don't think it's the right solution for issues like #6322.

The core problem in that issue is that server owners don't expect this to be possible. If they were aware it was possible, they'd likely just ban `/version` entirely. It doesn't provide enough utility to warrant adding subpermissions.

Better approaches would be:
- Remove this feature altogether
- Move the functionality under the `/plugins` command instead, so that plugin info is only accessible by a single command
- Have `VersionCommand` check if the user has `pocketmine.command.plugins` permission so that the level of information access is consistent

PR #6366, opened 2024-06-23
Implement new 1.21 copper blocks

IvanCraft623:
## Introduction

Implement the new 1.21 copper blocks types!!!

#### Blocks
- Added the following new blocks:
  - All types of Copper Bulb
  - All types of Copper Door
  - All types of Copper Trapdoor
  - All types of Chiseled Copper
  - All types of Copper Grate

## Tests
   Copper blocks
- [x] Writing a test plugin (provide the code)
  - To be able to test Copper Bulb light activation I wrote this hack code:
```php
public function onWorldSoundEvent(WorldSoundEvent $event) : void {
	$sound = $event->getSound();
	if (!($powerOn = $sound instanceof RedstonePowerOnSound) && !$sound instanceof RedstonePowerOffSound) {
		return;
	}

	$world = $event->getWorld();
	$block = $world->getBlock($event->getPosition());
	if (!$block instanceof Button) {
		return;
	}

	$support = $block->getSide(Facing::opposite($block->getFacing()));
	if (!$support instanceof CopperBulb) {
		return;
	}

	$world->setBlock($support->getPosition(), $support->togglePowered($powerOn));
}
```


cjmustard (approved):

src/data/bedrock/block/convert/BlockStateDeserializerHelper.php
```diff
@@ -102,24 +106,24 @@ public static function decodeComparator(RedstoneComparator $block, BlockStateRea
 	}
 
 	/**
-	 * @phpstan-template TBlock of Copper|CopperSlab|CopperStairs
+	 * @phpstan-template TBlock of Copper|CopperSlab|CopperStairs|CopperBulb|CopperDoor|CopperTrapdoor|CopperGrate
```
Gaprix: Maybe it's worth creating an abstract class or something like that to combine all kinds of copper blocks?

src/block/CopperDoor.php
```diff
@@ -0,0 +1,63 @@
...
+			$world = $this->position->getWorld();
+			if ($other instanceof CopperDoor) {
+				$other->setOxidation($this->oxidation);
+				$other->setWaxed($this->waxed);
+				$world->setBlock($other->position, $other);
+			}
+			return true;
+		}
+
+		return false;
```
Gaprix: The player will not be able to open the door if he presses the shift and there is no item in his hand. In vanilla it's possible. This code fixes this issue
```
		if($player->isSneaking() && $this->onInteractCopper($item, $face, $clickVector, $player, $returnedItems)){
			//copy copper properties to other half
			$other = $this->getSide($this->top ? Facing::DOWN : Facing::UP);
			if($other instanceof CopperDoor){
				$other->setOxidation($this->oxidation);
				$other->setWaxed($this->waxed);
				$world->setBlock($other->position, $other);
			}
			return true;
		}elseif($player !== null){
			return parent::onInteract($item, $face, $clickVector, $player, $returnedItems);
		}

		return false;
```

Note: The onInteract function in this class is not called if the player holds an item in his hand with the shift key pressed. This means that he will not be able to make the door waxed.

src/block/CopperTrapdoor.php
```diff
@@ -0,0 +1,43 @@
...
+	use CopperTrait{
+		onInteract as onInteractCopper;
+	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if ($player !== null && !$player->isSneaking()) {
+			return parent::onInteract($item, $face, $clickVector, $player, $returnedItems);
+		}
+
+		return $this->onInteractCopper($item, $face, $clickVector, $player, $returnedItems);
```
Gaprix: As with the door, the player will not be able to open the trapdoor if he has the shift pressed and there is no item in his hand. Although it is possible in vanilla. Here is the fix
```
		if($player->isSneaking() && $this->onInteractCopper($item, $face, $clickVector, $player, $returnedItems)){
			return true;
		}elseif($player !== null){
			return parent::onInteract($item, $face, $clickVector, $player, $returnedItems);
		}

		return false;
```

Note: The onInteract function in this class is not called if the player holds an item in his hand with the shift key pressed. This means that he will not be able to make the trapdoor waxed.

Gaprix (changes requested):

src/data/bedrock/block/convert/BlockStateDeserializerHelper.php
IvanCraft623: Yeah, we could create an interface for that, I'm just not sure what name it would have

IvanCraft623: Although this would really be work for another PR, because we need to edit other's blocks code.

src/block/CopperDoor.php
IvanCraft623: > The player will not be able to open the door if he presses the shift and there is no item in his hand

You are right

Gaprix (approved):

src/block/CopperBulb.php
```diff
@@ -0,0 +1,68 @@
...
+	/** @return $this */
+	public function togglePowered(bool $powered) : self{
+		if($powered === $this->powered){
+			return $this;
+		}
+		if ($powered) {
+			$this->setLit(!$this->lit);
+		}
+		$this->setPowered($powered);
+		return $this;
```
ShockedPlot7560, line 54: The name "toggle" suggest we don't have parameters and it just toggle the actual state ?

src/data/bedrock/block/convert/BlockObjectToStateSerializer.php
```diff
@@ -1221,6 +1225,40 @@ private function registerSerializers() : void{
...
+					Ids::WAXED_EXPOSED_CHISELED_COPPER,
+					Ids::WAXED_WEATHERED_CHISELED_COPPER,
+					Ids::WAXED_OXIDIZED_CHISELED_COPPER
+				) :
+				Helper::selectCopperId($oxidation,
+					Ids::CHISELED_COPPER,
+					Ids::EXPOSED_CHISELED_COPPER,
+					Ids::WEATHERED_CHISELED_COPPER,
+					Ids::OXIDIZED_CHISELED_COPPER
+				)
```
ShockedPlot7560, line 1276: To be consistent, maybe inline this lines ?

src/data/bedrock/block/convert/BlockObjectToStateSerializer.php
```diff
@@ -1221,6 +1225,40 @@ private function registerSerializers() : void{
...
+					Ids::WAXED_WEATHERED_COPPER_GRATE,
+					Ids::WAXED_OXIDIZED_COPPER_GRATE
+				) :
+				Helper::selectCopperId($oxidation,
+					Ids::COPPER_GRATE,
+					Ids::EXPOSED_COPPER_GRATE,
+					Ids::WEATHERED_COPPER_GRATE,
+					Ids::OXIDIZED_COPPER_GRATE
+				)
+			);
```
ShockedPlot7560, line 1294: Same

src/data/bedrock/block/convert/BlockObjectToStateSerializer.php
```diff
@@ -1288,6 +1326,67 @@ private function registerSerializers() : void{
...
+					Ids::WAXED_COPPER_BULB,
+					Ids::WAXED_EXPOSED_COPPER_BULB,
+					Ids::WAXED_WEATHERED_COPPER_BULB,
+					Ids::WAXED_OXIDIZED_COPPER_BULB) :
+				Helper::selectCopperId($oxidation,
+					Ids::COPPER_BULB,
+					Ids::EXPOSED_COPPER_BULB,
+					Ids::WEATHERED_COPPER_BULB,
+					Ids::OXIDIZED_COPPER_BULB
+				))
```
ShockedPlot7560, line 1376: same

src/data/bedrock/block/convert/BlockObjectToStateSerializer.php
```diff
@@ -1288,6 +1326,67 @@ private function registerSerializers() : void{
...
+						Ids::WAXED_WEATHERED_COPPER_DOOR,
+						Ids::WAXED_OXIDIZED_COPPER_DOOR
+					) :
+					Helper::selectCopperId(
+						$oxidation,
+						Ids::COPPER_DOOR,
+						Ids::EXPOSED_COPPER_DOOR,
+						Ids::WEATHERED_COPPER_DOOR,
+						Ids::OXIDIZED_COPPER_DOOR
+					)
```
ShockedPlot7560, line 1398: same

src/data/bedrock/block/convert/BlockObjectToStateSerializer.php
```diff
@@ -1288,6 +1326,67 @@ private function registerSerializers() : void{
...
+						Ids::WAXED_WEATHERED_COPPER_TRAPDOOR,
+						Ids::WAXED_OXIDIZED_COPPER_TRAPDOOR
+					) :
+					Helper::selectCopperId(
+						$oxidation,
+						Ids::COPPER_TRAPDOOR,
+						Ids::EXPOSED_COPPER_TRAPDOOR,
+						Ids::WEATHERED_COPPER_TRAPDOOR,
+						Ids::OXIDIZED_COPPER_TRAPDOOR
+					)
```
ShockedPlot7560, line 1420: same

src/data/bedrock/block/convert/BlockObjectToStateSerializer.php
dktapps: More important that the code isn't inconveniently wide so it doesn't require horizontal scrolling.

IvanCraft623: I prefer better visualization than “consistence”

src/block/CopperBulb.php
IvanCraft623: It is important to have the parameter, in case of implementing redstone it needs to indicate if it is being powered and work as a T flip-flop. Besides it is normal that toggle* functions require a parameter in PM.

ShockedPlot7560 (approved):

PR #6377, opened 2024-06-26
1.21.0 Wind Charge Implementation

cjmustard:
## Introduction
Implementation for the 1.21.0 `Wind Charge`

### Relevant issues
N/A

## Details
- Non-iron doors and trapdoors are flipped.
- Fence gates are flipped.
- Buttons are pressed.
- Levers are flipped.
- Bells are rung and swung.
- Chorus flowers and pointed dripstone blocks break, dropping in item form.
- Lit candles (both standalone and on cake) are extinguished.

Wind charges can be deflected if hit with another projectile or a melee attack, similar to ghast fireballs.
Wind charges are distance attenuated by liquids such as water and lava. They also appear to fly in a straight line, similar to fireballs
Wind charges are set to despawn after five minuets due to the gravity less nature of the entity.

https://minecraft.wiki/w/Wind_Charge

## Changes
One method has been added to pocketmine\block\Block:
```php
	/**
	 * Called when this block is updated by a state altering projectile in the world.
	 *
	 * @param pocketmine\entity\projectile\Projectile $projectile The projectile affecting the block.
	 */
	public function onProjectileInteraction(pocketmine\entity\projectile\Projectile $projectile) : void {

	}
```

Both class' for the Wind Charge item and Projectile have been created:
- pocketmine\entity\projectile\WindCharge.php
- pocketmine\item\WindCharge.php

One class has been created to handle the wind charge explode particle:
- pocketmine\world\particle\WindExplosionParticle.php

One class has been added to handle wind charge burst sound:
- pocketmine/world/sound/WindChargeBurstSound.php

The Wind Charge has been implemented as a vanilla item.
The Wind Charge Entity has been implemented as an Entity.

### API changes
N/A

### Behavioural changes
N/A

## Backwards compatibility
N/A

## Follow-up
N/A

## Tests
- [ ] Other (provide details)

# Behavior With Other Blocks
https://github.com/pmmp/PocketMine-MP/assets/80650734/e27ca3d6-f086-432b-bff1-72be8c252557

# Behavior With Players (shows less effective under water.)
https://github.com/pmmp/PocketMine-MP/assets/80650734/2576fc31-f9cb-46de-a2c6-28e39964849a

# Behavior Multi Hit
https://github.com/pmmp/PocketMine-MP/assets/80650734/0250cdab-350e-4586-bf7a-2340195e0429


pandaaaBE:
image

When I am standing still and aim to the ground.. In Vanilla i get boosted maybe 2.5 blocks and with this implementation nothing happens

(already fixed)

pandaaaBE:
image

I can spam the wind charges to the sky and they will not despawn, this should be fixed somehow

(already fixed)

pandaaaBE:
The sound "wind_charge.burst" is missing

(already added lol)

cjmustard:
all issues mentioned previously should be fixed now!

src/entity/projectile/WindCharge.php
```diff
@@ -0,0 +1,242 @@
...
+			$world->addSound(
+				$affectedBlock->getPosition()->add(0.5, 0.5, 0.5),
+				$affectedBlock->isActivated() ? new RedstonePowerOnSound() : new RedstonePowerOffSound()
+			);
+
+			return;
+		}
+
+		if($affectedBlock instanceof Bell) {
+			/** @var Bell $affectedBlock */
```
SOF3: Is this redundant? Most static analysis tools can infer this from the `instanceof` line.

cjmustard: yes very much so

src/entity/projectile/WindCharge.php
```diff
@@ -0,0 +1,227 @@
...
+		$maxX = (int) ceil($source->x + $radius + 1);
+		$minY = (int) floor($source->y - $radius - 1);
+		$maxY = (int) ceil($source->y + $radius + 1);
+		$minZ = (int) floor($source->z - $radius - 1);
+		$maxZ = (int) ceil($source->z + $radius + 1);
+
+		return new AxisAlignedBB($minX, $minY, $minZ, $maxX, $maxY, $maxZ);
+	}
+
+	private function checkAffect(Block $affectedBlock) : void {
```
Matze997: More customizations should be possible here
Maybe event(s)?

Also method name could be missleading

src/entity/projectile/WindCharge.php
```diff
@@ -0,0 +1,227 @@
...
+use pocketmine\world\sound\WindChargeBurstSound;
+
+use function ceil;
+use function floor;
+use function round;
+use function time;
+
+class WindCharge extends Throwable{
+	protected float $radius = 2.5;
+	protected float $damage = 0.5;
```
nicholass003: damage value is 1 (half heart)

IvanCraft623: These should be constants instead of variables

src/entity/projectile/WindCharge.php
```diff
@@ -0,0 +1,227 @@
...
+
+	public static function getNetworkTypeId() : string{ return EntityIds::WIND_CHARGE_PROJECTILE; }
+
+	protected function getInitialDragMultiplier() : float{ return 0; }
+	protected function getInitialGravity() : float{ return 0; }
+
+	protected function initEntity(CompoundTag $nbt) : void{
+		parent::initEntity($nbt);
+
+		$this->createdAt = time();
```
IvanCraft623: All minecraft related timing uses ticks, you should be doing like this

IvanCraft623: Also there's already a variable that tracks entity's lived ticks: https://github.com/pmmp/PocketMine-MP/blob/585aee9386a787c95e73dd0a05ffca8329606b68/src/entity/Entity.php#L137

src/entity/projectile/WindCharge.php
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	public function attack(EntityDamageEvent $source) : void{
+		if(!$source instanceof EntityDamageByEntityEvent) return;
```
IvanCraft623:
```suggestion
		if(!$source instanceof EntityDamageByEntityEvent || ($entity = $source->getDamager()) === null){
				return;
		}
```

src/entity/projectile/WindCharge.php
```diff
@@ -0,0 +1,227 @@
...
+
+		if(($entity = $source->getDamager()) == null) return;
```
IvanCraft623: If you do the suggestion above you can remove this line

src/entity/projectile/WindCharge.php
```diff
@@ -0,0 +1,227 @@
...
+		$this->setOwningEntity($entity);
+
+		$this->setMotion($entity->getDirectionVector()->multiply(1.5));
+	}
+
+	protected function onHitEntity(Entity $entityHit, RayTraceResult $hitResult) : void{
+		if($this->getOwningEntity() === null)$ev = new EntityDamageByEntityEvent($this, $entityHit, EntityDamageEvent::CAUSE_PROJECTILE, $this->damage);
+		else $ev = new EntityDamageByChildEntityEvent($this->getOwningEntity(), $this, $entityHit, EntityDamageEvent::CAUSE_PROJECTILE, $this->damage);
```
IvanCraft623: Always use `if-else` whith brackets `{}`

src/entity/projectile/WindCharge.php
```diff
@@ -0,0 +1,227 @@
...
+		return true;
+	}
+
+	protected function onHit(ProjectileHitEvent $event) : void{
+		$source = $this->getLocation();
+
+		$this->getWorld()->addSound($event->getRayTraceResult()->getHitVector(), new WindChargeBurstSound());
+		$this->getWorld()->addParticle($source, new WindExplosionParticle());
+
+		$list = $source->getWorld()->getNearbyEntities($this->getBound($source, $this->radius));
```
IvanCraft623: Don't create a variable if you are only going to use a value once

IvanCraft623: Use `World::getCollidingEntities()` instead of `getNearbyEntities` since this behavior should check if the entity is colliderable. Also pass `$this` as the `entity` algument, so this way you can be sure that this entity is not on the list.

src/entity/projectile/WindCharge.php
```diff
@@ -0,0 +1,227 @@
...
+
+		foreach($list as $entity){
+			if($entity == $this) continue;
```
IvanCraft623: If you do mentioned above you can remove this

src/entity/projectile/WindCharge.php
```diff
@@ -0,0 +1,227 @@
...
+
+			$entityPos = $entity->getPosition();
+			$distance = $entityPos->distance($source) / $this->radius;
+			$motion = $entityPos->subtractVector($source)->normalize();
+
+			($entity->isUnderwater()) ? $exposure = 0.5 : $exposure = 1.5;
+			$impact = (1 - $distance) * $exposure;
+
+			if ($impact <= 0) continue;
```
IvanCraft623: Avoid using on single line if, on pocketmine we use brackets code-style

src/entity/projectile/WindCharge.php
```diff
@@ -0,0 +1,227 @@
...
+		}
+	}
+
+	private function getBound(Vector3 $source, float $radius) : AxisAlignedBB {
+		$minX = (int) floor($source->x - $radius - 1);
+		$maxX = (int) ceil($source->x + $radius + 1);
+		$minY = (int) floor($source->y - $radius - 1);
+		$maxY = (int) ceil($source->y + $radius + 1);
+		$minZ = (int) floor($source->z - $radius - 1);
+		$maxZ = (int) ceil($source->z + $radius + 1);
```
IvanCraft623: Again, don't create variables if you are gonna use it once

src/entity/projectile/WindCharge.php
IvanCraft623: Agree about method name

IvanCraft623: Would be more consistent create a new method on the Block class and each block will decide it's own behavour

IvanCraft623 (changes requested):
Quick review about code quality, Haven't reviewd a vanilla parity behavour

src/entity/projectile/WindCharge.php
cjmustard: ProjectileLaunchEvent & ProjectileHitEvent already exist

cjmustard:
this is a clean version.

src/entity/projectile/WindCharge.php
IvanCraft623: Don't close a conversation if you haven't resolved it.

src/entity/projectile/WindCharge.php
IvanCraft623: Same here

cjmustard: done !

src/entity/projectile/WindCharge.php
cjmustard: sorry !

src/entity/projectile/WindCharge.php
```diff
@@ -41,8 +41,6 @@
 use function round;
 
 class WindCharge extends Throwable{
-	protected float $radius = 2.5;
```
IvanCraft623: I told you about making them constants, no remvonving them

This is a constant btw
https://github.com/pmmp/PocketMine-MP/blob/585aee9386a787c95e73dd0a05ffca8329606b68/src/entity/projectile/Arrow.php#L52

src/block/Button.php
```diff
@@ -91,4 +92,16 @@ public function onNearbyBlockChange() : void{
 	private function canBeSupportedAt(Block $block, int $face) : bool{
 		return $block->getAdjacentSupportType(Facing::opposite($face))->hasCenterSupport();
 	}
+
+	public function onWindChargeInteraction(WindCharge $windCharge) : void{
+		if($this->getTypeId() == BlockTypeIds::IRON_DOOR) {
```
Matze997: Why do you check for an iron door in the button class?

src/block/Door.php
```diff
@@ -176,4 +177,15 @@ public function getAffectedBlocks() : array{
 	private function canBeSupportedAt(Block $block) : bool{
 		return $block->getAdjacentSupportType(Facing::DOWN)->hasEdgeSupport();
 	}
+
+	public function onWindChargeInteraction(WindCharge $windCharge) : void{
+		if($this->getTypeId() == BlockTypeIds::IRON_DOOR) {
```
Matze997: if(!self instanceof WoodenDoor){
    return;
}

Would make a bit more sense here

cjmustard: careless porting from a separate file, thank you for this catch!

cjmustard: but that wouldnt work beacuse of copper doors!

src/block/Button.php
AIPTU: Shouldn't it using '===' instead?

IvanCraft623:
@ShockedPlot7560 What do you think about using `Block::onProjectileHit()` instead of creating a new function only for specific projectile logic. Would have to change it to accept null in $hitResult parameter, so it would be BC Break

pandaaaBE:
> @ShockedPlot7560 What do you think about using `Block::onProjectileHit()` instead of creating a new function only for specific projectile logic. Would have to change it to accept null in $hitResult parameter, so it would be BC Break

I think a BC Break for such an item would be unnecessary.. Maybe it's something for another PR. 
Someone could make a PR for it after this PR has been merged, otherwise you would have to wait until PM6 for the item because of the BC break

src/block/Block.php
```diff
@@ -526,7 +536,7 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
 	/**
 	 * Called when this block is attacked (left-clicked) by a player attempting to start breaking it in survival.
 	 *
-	 * @return bool if an action took place, prevents starting to break the block if true.
+	 * @return bool if an action took place, prs starting to break the block if true.
```
ShockedPlot7560: unrelated change

src/block/Door.php
ShockedPlot7560: use ===

src/block/Trapdoor.php
```diff
@@ -91,4 +92,15 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+
+	public function onWindChargeInteraction(WindCharge $windCharge) : void{
+		if($this->getTypeId() == BlockTypeIds::IRON_TRAPDOOR) {
+			return;
+		}
+
+		$this->open = !$this->open;
+		$world = $this->position->getWorld();
+		$world->setBlock($this->position, $this);
+		$world->addSound($this->position, new DoorSound());
```
ShockedPlot7560: Need to be moved into a function to be used here and in onInteract

src/block/Lever.php
```diff
@@ -101,6 +102,16 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
 
+	public function onWindChargeInteraction(WindCharge $windCharge) : void{
+		$this->activated = !$this->activated;
+		$world = $this->position->getWorld();
+		$world->setBlock($this->position, $this);
+		$world->addSound(
+			$this->position->add(0.5, 0.5, 0.5),
+			$this->isActivated() ? new RedstonePowerOnSound() : new RedstonePowerOffSound()
+		);
+	}
```
ShockedPlot7560: same here, moove the logic into one function

src/block/Door.php
```diff
@@ -176,4 +177,15 @@ public function getAffectedBlocks() : array{
...
+			return;
+		}
+
+		$this->open = !$this->open;
+		$world = $this->position->getWorld();
+		$world->setBlock($this->position, $this);
+		$world->addSound($this->position, new DoorSound());
```
ShockedPlot7560: need to used the same logic as interact into one single function to be sure to update the top door block

src/block/Button.php
```diff
@@ -91,4 +92,16 @@ public function onNearbyBlockChange() : void{
...
+			return;
+		}
+
+		$this->pressed = true;
+		$world = $this->position->getWorld();
+		$world->setBlock($this->position, $this);
+		$world->scheduleDelayedBlockUpdate($this->position, $this->getActivationTime());
+		$world->addSound($this->position->add(0.5, 0.5, 0.5), new RedstonePowerOnSound());
```
ShockedPlot7560: move it in a single function to use it in interact with the same logic

ShockedPlot7560 (changes requested):

ShockedPlot7560:
> @ShockedPlot7560 What do you think about using `Block::onProjectileHit()` instead of creating a new function only for specific projectile logic. Would have to change it to accept null in $hitResult parameter, so it would be BC Break

Thats not really a projectileHit, I don't think we need to change this function signature. It's more a projectile effect or smth like this.

composer.lock
ShockedPlot7560, line 1: no need to be updated

src/block/Button.php
```diff
@@ -91,4 +89,19 @@ public function onNearbyBlockChange() : void{
...
 		return $block->getAdjacentSupportType(Facing::opposite($face))->hasCenterSupport();
 	}
+
+	public function onProjectileInteraction(Projectile $projectile) : void{
+		if($projectile instanceof WindCharge) {
+			$this->toggle();
+		}
+	}
+
+	public function toggle() : void {
```
ShockedPlot7560: Thats not really a toggle, more a press or smth like this on the button.

src/block/Door.php
```diff
@@ -176,4 +168,34 @@ public function getAffectedBlocks() : array{
...
+		if($projectile instanceof WindCharge) {
+			if($this->getTypeId() === BlockTypeIds::IRON_DOOR) {
+				return;
+			}
+
+			$this->open = !$this->open;
+
+			$world = $this->position->getWorld();
+			$world->setBlock($this->position, $this);
+			$world->addSound($this->position, new DoorSound());
```
ShockedPlot7560: use toggle ?

src/block/Lever.php
```diff
@@ -91,19 +93,28 @@ public function onNearbyBlockChange() : void{
...
+		if($projectile instanceof WindCharge) {
+			$this->toggle();
+		}
+	}
+
 	private function canBeSupportedAt(Block $block, int $face) : bool{
 		return $block->getAdjacentSupportType($face)->hasCenterSupport();
 	}
 
-	//TODO
```
ShockedPlot7560: redstone isnt implemented, so it still to do

cjmustard:
@ShockedPlot7560 the issues you have mentioned have been fixed, thanks for pointing those out!

src/entity/projectile/WindCharge.php
```diff
@@ -0,0 +1,150 @@
...
+use pocketmine\event\entity\EntityDamageByEntityEvent;
+use pocketmine\event\entity\EntityDamageEvent;
+use pocketmine\event\entity\ProjectileHitEvent;
+use pocketmine\math\AxisAlignedBB;
+use pocketmine\math\RayTraceResult;
+use pocketmine\math\Vector3;
+use pocketmine\network\mcpe\protocol\types\entity\EntityIds;
+use pocketmine\world\particle\WindExplosionParticle;
+use pocketmine\world\sound\WindChargeBurstSound;
+
```
nicholass003: fix code style here

cjmustard: what exactly am I changing here?

cjmustard: the line thank you @pandaa-be

ShockedPlot7560:
Why have you merged the stable into it ? Its now not up to date with minor-next and the diff reading is difficult with a lot of unrelated changes.

cjmustard:
@ShockedPlot7560 sorry, I am new to the whole git ecosystem, have I fixed this issue or is there something else I would need to do?

cjmustard:
bump

composer.lock
ShockedPlot7560, line 1: Unrelated changes

src/block/Door.php
```diff
@@ -176,4 +168,29 @@ public function getAffectedBlocks() : array{
...
 		return $block->getAdjacentSupportType(Facing::DOWN)->hasEdgeSupport();
 	}
+
+	public function onProjectileInteraction(Projectile $projectile) : void{
+		if($projectile instanceof WindCharge) {
+			if($this->getTypeId() === BlockTypeIds::IRON_DOOR) {
+				return;
+			}
+
+			$this->toggle();
```
ShockedPlot7560, line 178: Is the door open when the wind charge interacts with it?

src/block/Lever.php
```diff
@@ -91,18 +93,29 @@ public function onNearbyBlockChange() : void{
...
+
+	public function onProjectileInteraction(Projectile $projectile) : void{
+		if($projectile instanceof WindCharge) {
+			$this->toggle();
+		}
+	}
+
+	private function canBeSupportedAt(Block $block, int $face) : bool{
+		return $block->getAdjacentSupportType($face)->hasCenterSupport();
+	}
```
ShockedPlot7560: unrelated changes

src/block/Lever.php
```diff
@@ -91,18 +93,29 @@ public function onNearbyBlockChange() : void{
...
+
+	public function toggle() : void {
 		$this->activated = !$this->activated;
 		$world = $this->position->getWorld();
 		$world->setBlock($this->position, $this);
 		$world->addSound(
 			$this->position->add(0.5, 0.5, 0.5),
-			$this->activated ? new RedstonePowerOnSound() : new RedstonePowerOffSound()
+			$this->isActivated() ? new RedstonePowerOnSound() : new RedstonePowerOffSound()
```
ShockedPlot7560: Direct access to the property is enough

composer.lock
cjmustard: I am not sure how this became out of sync, and can not seem to make it match. could you please help me out

src/block/Door.php
cjmustard: it should toggle the door either way it is facing

cjmustard:
Closing to rebase at composer.lock level

Issue #6349, opened 2024-05-22
Container windows aren't closed by the server when the player is too far away


dktapps:
## Description
If the player opens an inventory window and moves away from the container (e.g. by water currents), they will still be able to interact with the container inventory from any distance as long as they don't do certain actions (e.g. breaking blocks).

We should probably have a distance check to close the window when the player goes too far away. However I don't know what the appropriate distance would be.

CoderJoeW:
Vanilla seems to be 12 blocks away is when I got kicked out of the container window

PR #6382, opened 2024-07-03
Close block windows when player is too far away

CoderJoeW:
## Introduction
If a player is inside of a container window and is moved over 6 blocks from the container the window will automatically be closed

In vanilla minecraft when I was testing this it closed the window after be moving 12 blocks(I could have miscounted) so that is what I implemented here

By running the following this change can be disabled

`$inventory->setMaxDistanceFromContainer(0)`

### Relevant issues
This is a resolution for the issue

#6349 

### Test

https://file.io/mphsWsJrFfcy


src/player/Player.php
```diff
@@ -1374,6 +1378,14 @@ protected function processMostRecentMovements() : void{
 			$this->logger->debug("Exceeded movement rate limit, forcing to last accepted position");
 			$this->sendPosition($this->location, $this->location->getYaw(), $this->location->getPitch(), MovePlayerPacket::MODE_RESET);
 		}
+
+		if ($this->openContainerPosition !== null) {
```
jasonw4331: There should be an API for plugins to be able to disable this behavior on their custom inventories

jasonw4331 (changes requested):
The composer lock file needs removed from the PR and some other API might be nice, but it looks clean.

src/player/Player.php
```diff
@@ -1374,6 +1378,14 @@ protected function processMostRecentMovements() : void{
...
+			$distance = $this->location->distance($this->openContainerPosition);
+
+			if ($distance > self::MAX_DISTANCE_FROM_CONTAINER) {
```
ShockedPlot7560: Should this distance be configurable by each inventory in the case of plugin customization?

src/player/Player.php
```diff
@@ -2646,6 +2658,12 @@ public function setCurrentWindow(Inventory $inventory) : bool{
 		$inventoryManager->onCurrentWindowChange($inventory);
 		$inventory->onOpen($this);
 		$this->currentWindow = $inventory;
+
+		if ($inventory instanceof BlockInventory) {
+			$block = $inventory->getHolder();
```
nicholass003: perhaps this should be named $holder for consistent code

CoderJoeW:
I am not super familiar with the API implementation so that might take me a min to figure out

Do you think we should have some method like setWindowCloseDistance and setting it to 0 would disable the functionality?

CoderJoeW:
@ShockedPlot7560 @nicholass003 @jasonw4331 

I believe I addressed all concerns and included a video of testing the server against a vanilla world

I found that in further testing vanilla seems to kick you out of an inventory window > 6 blocks distance so thats what I also updated this PR to do

src/player/Player.php
```diff
@@ -296,6 +298,9 @@ public static function isValidUserName(?string $name) : bool{
 
 	protected ?SurvivalBlockBreakHandler $blockBreakHandler = null;
 
+	protected ?SimpleInventory $openContainerInventory = null;
```
ShockedPlot7560: We already have this information

src/player/Player.php
```diff
@@ -2646,6 +2659,17 @@ public function setCurrentWindow(Inventory $inventory) : bool{
 		$inventoryManager->onCurrentWindowChange($inventory);
 		$inventory->onOpen($this);
 		$this->currentWindow = $inventory;
+
+		if ($inventory instanceof SimpleInventory) {
+			if ($inventory->getMaxDistanceFromContainer() > 0) {
```
ShockedPlot7560: I think we can introduce a new interface to give the `getMaxDistanceFromContainer` to avoid checking SimpleInventory which is not guaranteed

src/block/ProximityRestricted.php
```diff
@@ -0,0 +1,39 @@
...
+
+	/**
+	 * Returns the max distance the player can be away
+	 */
+	public function getMaxDistance() : int;
+
+	/**
+	 * Returns the max distance the player can be away
+	 */
+	public function setMaxDistance(int $maxDistance) : void;
```
ShockedPlot7560: I don't think we need this in the API, its more a readonly value

src/block/ProximityRestricted.php
```diff
@@ -0,0 +1,39 @@
...
+ * @link http://www.pocketmine.net/
+ *
+ *
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\block;
+
+interface ProximityRestricted {
```
ShockedPlot7560: This need to be placed in the inventory namespace

src/block/ProximityRestricted.php
```diff
@@ -0,0 +1,39 @@
...
+
+	public const MAX_DISTANCE = 6;
+
+	/**
+	 * Returns the max distance the player can be away
+	 */
+	public function getMaxDistance() : int;
```
ShockedPlot7560: Max distance against ? need some documentation or name change

src/block/ProximityRestricted.php
ShockedPlot7560: And i'm not a big fan of this interfaces name

src/block/ProximityRestricted.php
ShockedPlot7560: Could be typed with non-zero-int as if we want to disable them, just don't put the interface ?

src/block/ProximityRestricted.php
CoderJoeW: @ShockedPlot7560 Any suggestions on the name?

src/block/ProximityRestricted.php
CoderJoeW: @ShockedPlot7560 Maybe so but it never hurts to give people the ability to change it

ShockedPlot7560: Any other point of view on that ?

dktapps: I'd prefer to have it read-only, and then explore adding modification support later if people ask for it. Mutability always causes headaches in ways you don't expect.

ShockedPlot7560:
I've taking back this PR to try this to be merged.

I've made some changes to fix stale reviews but still has question about we have to include this utility.
For now, we introduce a new `ProximityRestricted` interface which need to be combined with the `BlockInventory` to be functionnal. We can't include this interface into the `BaseInventory` bc its linked to a block inventory and we can't include it into the `BlockInventory` without any BC breaks. Do we keep a new interface which need to be included everywhere we have a BlockInventory or anything else come in your minds ?

dktapps:
> Do we keep a new interface which need to be included everywhere we have a BlockInventory or anything else come in your minds ?

I think the whole thing is overengineered. Most likely the only thing needed is to check if the block is still within the player's reach distance. I don't think this should be the Block's responsibility at all.

dktapps:
Also worth asking if we really want to make an invasive change like this considering the upcoming #6533.

dktapps:
Closing this as I think the solution is way too complicated for the requirements.

PR #6383, opened 2024-07-03
6161 implement interface common block properties

### LINKED ISSUE #6161: Add interfaces for blocks with common properties
## Description
Many blocks have properties such as facing, on/off etc.

However, it's not currently possible to tell which blocks support e.g. horizontal facing using `instanceof`, despite the actual API implementations being the same for every block, because the property implementations have been done using traits only.

This leads to developers of e.g. world-edit plugins having a really bad time trying to use the API for stuff like rotating blocks.

I've seen various kludgy workarounds for this including `method_exists()` and checking for specific instances, but really we just need to have interfaces to deal with this.

I'd planned to do this even as far back as 4.0.0, but somehow never got around to it.

CoderJoeW:
## Introduction
The goal of this is to give plugins an easier way of accessing common properties of blocks such as facing and on/off

### Relevant issues
Fixes #6161

### DIFF
--- a/src/block/FacingInterface.php
+++ b/src/block/FacingInterface.php
@@ -27,14 +27,11 @@ interface FacingInterface {
 
 	/**
 	 * Get the current facing direction of the block
-     * @return int
 	 */
 	public function getFacing() : int;
 
     /**
      * Set the current facing direction of the block
-     * @param int $facing
-     * @return object
      */
     public function setFacing(int $facing) : object;
 }
--- a/src/block/FacingInterface.php
+++ b/src/block/FacingInterface.php
@@ -34,7 +34,7 @@ interface FacingInterface {
     /**
      * Set the current facing direction of the block
      * @param int $facing
-     * @return void
+     * @return object
      */
     public function setFacing(int $facing) : object;
 }
--- a/src/block/FacingInterface.php
+++ b/src/block/FacingInterface.php
@@ -30,8 +30,8 @@ interface FacingInterface {
 	 */
 	public function getFacing() : int;
 
-    /**
-     * Set the current facing direction of the block
-     */
-    public function setFacing(int $facing) : object;
+	/**
+	 * Set the current facing direction of the block
+	 */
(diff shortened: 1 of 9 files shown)


CoderJoeW:
@dktapps  Just want to get a quick review of this to make sure its headed in the intended direction

cjmustard:
Would this be better utilized trait, I think the is the direction most items like this follow.

[pocketmine\block\Bell.php#42](https://github.com/pmmp/PocketMine-MP/blob/77db7a8941d3d91c7e452cb3ca469f59fbd8f4df/src/block/Bell.php#L42)

CoderJoeW:
> Would this be better utilized trait, I think the is the direction most items like this follow.
> 
> [pocketmine\block\Bell.php#42](https://github.com/pmmp/PocketMine-MP/blob/77db7a8941d3d91c7e452cb3ca469f59fbd8f4df/src/block/Bell.php#L42)

Probably should use the trait  as there is Horizontal and AnyFacing traits that could be used

Possibly out of scope? Would likely fit more into another PR for cleaning up the code

src/block/FacingInterface.php
```diff
@@ -0,0 +1,37 @@
...
+
+	/**
+	 * Get the current facing direction of the block
+	 */
+	public function getFacing() : int;
+
+	/**
+	 * Set the current facing direction of the block
+	 */
+	public function setFacing(int $facing) : object;
```
ShockedPlot7560, line 36: Its block not object

src/entity/object/Painting.php
```diff
@@ -40,7 +41,7 @@
 use pocketmine\world\World;
 use function ceil;
 
-class Painting extends Entity{
+class Painting extends Entity implements FacingInterface{
```
ShockedPlot7560, line 44: Design error, you're using an interface made for blocks in an entity, I don't think you should do that

dktapps:
Closing due to lack of response to review comments

PR #6390, opened 2024-07-08
Added ICopper interface to reduce duplications

ipad54:
## Introduction
This PR implements ```ICopper``` interface to reduce duplications and make further additions of copper blocks (such as #6366) easier.

### Relevant issues
#6366 

### API changes
Added ```ICopper``` interface


src/block/utils/Copper.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\block\utils;
+
+/**
+ * Represents copper blocks that have oxidized and waxed variations.
+ */
+interface Copper{
```
jasonw4331: This should follow the existing interface naming scheme seen by other common trait/interface pairs in pocketmine.

jasonw4331 (changes requested):
A rename of the interface should be done to remove the aliases everywhere.

src/block/utils/Copper.php
ipad54: The problem is that it’s difficult to find a good naming. I can suggest the following names:
- CopperInterface
- ICopper
- CopperBase
- CopperBlock
- Cupreous
- OxidizableBlock
- OxidizableCopper

IvanCraft623: I would prefer `ICopper` or `CopperBase`

nicholass003: ```ICopper``` is great

ipad54: > I would prefer `ICopper` or `CopperBase`

I think `ICopper` would be a good name because `CopperBase` sounds more like an abstract class rather than interface

nicholass003 (approved):

ShockedPlot7560 (approved):
Looks OK to me

NhanAZ (approved):

IvanCraft623 (approved):
Thanks for your contribution!

dktapps:
I don't much like the name of this interface. Are there no better options? It's not yet released (still stuck on `minor-next`) so there's time to improve it.

ipad54:
> Are there no better options?

There are some other naming options I came up with, but I doubt they are better than `ICopper`:
- CopperInterface
- CopperBase
- CopperBlock
- Cupreous

dktapps:
Maybe `CopperMaterial` or `CopperLike`, I'm not sure.

ipad54:
`CopperMaterial` sounds good to me. I've created a branch in my fork repo with rename (https://github.com/ipad54/PocketMine-MP/commit/87a533710b6455ac105bc3262f9c07feb87d8a74). If it looks good to you, I can open a PR.

dktapps:
Just make the PR so it can be reviewed here

PR #6392, opened 2024-07-09
VanillaBlocks: Properly name Mud Brick Slab

IvanCraft623:
Seems to be a dylan's typo :P


ShockedPlot7560 (approved):

NhanAZ (approved):

Gaprix (approved):

dktapps:
It's not a typo: https://github.com/pmmp/PocketMine-MP/blob/616f96a7039a71559e977cd2a18348b7d90444b2/src/block/Slab.php#L40

Doubtful that logic should be hidden away like that, but it is what it is.

PR #6396, opened 2024-07-10
Split glow lichen multi face logic into traits

ipad54:
## Introduction
This PR splits glow lichen multi face logiс into 2 traits like it's done with other blocks (such as `HorizontalFacingTrait`, `StaticSupportTrait`, etc). 

These changes are necessary to remove code duplication and make easier implementing other blocks that have multi face logic (such as sculk vein, but Mojang can add more blocks in the future updates).

### API changes
Added `MultiFacingTrait`, `MultiSupportTrait`

## Tests
PHPStan tests are OK.
In-game multi face logic also works fine.


src/block/utils/MultiSupportTrait.php
```diff
@@ -0,0 +1,95 @@
...
+use pocketmine\player\Player;
+use pocketmine\world\BlockTransaction;
+use function array_key_first;
+use function count;
+
+/**
+ * Used by blocks that have multiple support requirements in the area of one solid block, such as covering three sides of a corner.
+ * Prevents placement if support isn't available and automatically destroys a block side if it's support is removed.
+ */
+trait MultiSupportTrait{
```
dktapps: I feel like this trait should `use MultiFacingTrait`. It's evidently dependent.

dktapps (dismissed):

dktapps:
Can we change these to `MultiAnyFacingTrait` and similar? I can bet there'll be a day when we need to have `MultiHorizontalFacingTrait` with the functionality from classic vines.

dktapps (approved):

PR #6398, opened 2024-07-11
Workflow for automatic merging stable into child branches

IvanCraft623:
## Introduction

Recently the merging of stable into minor-next and major-next has given problems, because you have to create a PR manually and wait for revisions.

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -29,4 +29,5 @@ jobs:
           base: minor-next
           title: Merge branch 'stable' into minor-next
           committer: github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com>
+          author: ${{ github.actor }} <${{ github.actor_id }}+${{ github.actor }}@users.noreply.github.com>
           body:
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -25,6 +25,8 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # Fetch all branches and their history
 
     - name: Check if head and base are the same
       id: check-branches
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -25,8 +25,8 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
-        with:
-          fetch-depth: 0 # Fetch all branches and their history
+      with:
+        fetch-depth: 0 # Fetch all branches and their history
 
     - name: Check if head and base are the same
       id: check-branches
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -25,5 +25,7 @@ jobs:
         uses: peter-evans/create-pull-request@v6
         with:
           token: ${{ secrets.AUTO_MERGE_TOKEN }}
+          branch: merge/stable-into-minor-next
           base: minor-next
           title: Merge branch 'stable' into minor-next
+          author: ${{ github.actor }} <${{ github.actor_id }}+${{ github.actor }}@users.noreply.github.com>
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -43,7 +43,7 @@ jobs:
     - name: Attempt to merge head into base
       id: merge
       run: |
-        git merge --no-ff --no-commit origin/${{ github.event.inputs.head }} || echo "Merge conflict detected" > conflict.txt
+        git merge --no-ff --no-commit ${{ github.event.inputs.head }} || echo "Merge conflict detected" > conflict.txt
 
     - name: Check for merge conflicts
       id: check-conflict
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -61,6 +61,6 @@ jobs:
         fi
 
     - name: Push changes
-      if: steps.check-conflict.outcome == 'success'
+      if: steps.check-errors.outcome == 'success'
       run: |
         git push origin ${{ github.event.inputs.base }}
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -28,8 +28,8 @@ jobs:
       with:
         fetch-depth: 0 # Fetch all branches and their history
 
-    - name: Check if head and base are the same
-      id: check-branches
+    - name: Validate input branches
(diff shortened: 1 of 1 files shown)


SOF3:
I assume AUTO_MERGE_TOKEN should be a bot account that has and only has exclusive access to push to the merge branch? 

For extra safety, is it possible to push the merge branch to the fork repo of the bot account?

SOF3:
Tentatively approved, but I'm not pressing the approve button yet as we need to create the corresponding bot account setup

Expecting further comment from others before I proceed with the administrative stuff.

IvanCraft623:
> I assume AUTO_MERGE_TOKEN should be a bot account that has and only has exclusive access to push to the merge branch?

The perms I granted were: `workflow`, this automatically grants `repo` perms too

> For extra safety, is it possible to push the merge branch to the fork repo of the bot account?

I can try, but I really don't think it's possible.

alvin0319:
I would recommend separating major-next from the job and give different steps to them. Since major-next *could* have BC breaking changes that has conflict with stable.

Not sure if entire jobs will be cancelled if any of them fails to merge, if so I don't think putting major-next and minor-next into same step

IvanCraft623:
> I would recommend separating major-next from the job and give different steps to them. Since major-next _could_ have BC breaking changes that has conflict with stable.

In case of conflict it would be resolved manually


> Not sure if entire jobs will be cancelled if any of them fails to merge

It should continue running

.github/workflows/auto-merge.yml
```diff
@@ -0,0 +1,36 @@
...
+        with:
+          fetch-depth: 0 # Fetch all branches and their history
+
+      - name: Fetch all branches
+        run: git fetch --all
+
+      - name: Merge changes
+        run: |
+          git checkout ${{ matrix.branch }}
+          git merge stable
```
dktapps: The usual merge strategy is

`stable` -> `minor-next`
`minor-next` -> `major-next`

It doesn't make sense to use a matrix for this, since the branches should be merged sequentially.

This also doesn't make any considerations for if a conflict arises.

SOF3: If a conflict arises, this will just fail the CI. Do we need to do anything about it?

dktapps: It still shouldn't be using a matrix as the branches need to be sequentially merged.

This workflow will also generate 2 PRs for every commit pushed, which seems to me like it would be a major annoyance.

dktapps (changes requested):
I don't like the PR stuff. It'll create a whole lot of notification noise for stuff no one cares about. I'd favour having the action push changes directly if the merges succeeded. `${{ github.token }}` should be sufficient for that.

There's also no git config username or email for the merge commit, so it will likely just fail immediately.

As I said above, this should also not be using a matrix, since `stable` should be merged to `minor-next` and then `minor-next` to `major-next`. The workflow should be aborted if any step of this fails due to a conflict.

dktapps:
I'd also favour a `workflow_dispatch` manual option (and perhaps a cron job) instead of doing it automatically on every push, otherwise the `minor-next` will get a merge commit for every commit pushed to stable, and `major-next` might end up with 2 merges for every actual commit.

IvanCraft623:
I rewrote the entire workflow. So I'll list the changes because I messed up the commit history.
- A PR is no longer created for branch merging, so `create-pull-request@v6` action is no longer used.
- Workflow can only be executed with `workflow_dispatch`
- Matrix strategy changed to inputs:
  - `head`: `stable` and `minor-next` (branch to merge from).
  - `base`: `minor-next` and `major-next` (branch to merge into).
  - This change allow the merge strategy: `stable` -> `minor-next` -> `major-next`.
  - Validate that `head` and `base` are not the same branch.
- Removed dependency of a Personal Access Token
- Now a git config is set up (previously this was set by `create-pull-request` action automatically).
- Changes are pushed only if there was no conflict/error.

dktapps:
I don't think a matrix is necessary at all.

.github/workflows/auto-merge.yml
```diff
@@ -43,7 +43,7 @@ jobs:
     - name: Attempt to merge head into base
       id: merge
       run: |
-        git merge --no-ff --no-commit ${{ github.event.inputs.head }} || echo "Merge conflict detected" > conflict.txt
+        git merge --no-ff --no-commit ${{ github.event.inputs.head }} --allow-unrelated-histories || echo "Merge conflict detected" > conflict.txt
```
dktapps: why?

.github/workflows/auto-merge.yml
```diff
@@ -0,0 +1,66 @@
...
+        options:
+          - stable
+          - minor-next
+      base:
+        description: 'Branch to merge into'
+        required: true
+        type: choice
+        options:
+          - minor-next
+          - major-next
```
dktapps, line 19: These inputs are unnecessary. The workflow needs only to do this:

```
git checkout minor-next
git merge stable || exit 1
git checkout major-next
git merge minor-next || exit 1
git push origin minor-next major-next
```

The rest is just unnecessary inconvenience.

dktapps (changes requested):

.github/workflows/auto-merge.yml
IvanCraft623: I personally consider that the current form is more beneficial, if a specific merger causes conflicts, it can be handled and resolved in isolation, without interrupting or complicating other mergers.

In the proposed implementation an error in one merger will block the whole process.

dktapps: If a conflict arises, you'll have to re-run a workflow anyway. The only thing this achieves is making it more inconvenient to do so, while creating extra merges between outdated branches (e.g. outdated `minor-next` to `major-next` and then a second merge later on once you resolve the conflict between `stable` and `minor-next`).

In any case, a matrix for this is an unnecessary amount of clicks. At most it needs 2 options:
- stable -> minor-next
- minor-next -> major-next

dktapps:
To summarize internal discussions: There are security concerns with this. It would require the action to be able to push to protected branches. Since the action can be run from any branch, this means that a rogue actor can edit the action on an unprotected branch to push malicious changes to the protected branch. There doesn't seem to be a way to generally restrict modifications of the workflow to prevent this, nor a way to restrict which branch the workflow can be run from.

It seems we'll have to do one of the following:
- Set up the workflow in a separate repo where only admins can modify it, and have a workflow in this repo that can trigger it via `repository_dispatch`
- Set up a private script on a VPS somewhere and have a web API that a workflow in this repo could call.

dktapps:
This has been superseded by https://github.com/pmmp/RestrictedActions/actions/workflows/pocketmine-mp-branch-sync.yml

PR #6399, opened 2024-07-11
Update world protocol version

Gaprix:


dries-c:
I'm not sure if this makes sense given the bedrock data hasn't changed at all

src/data/bedrock/block/BlockStateData.php
```diff
@@ -42,8 +42,8 @@ final class BlockStateData{
 	public const CURRENT_VERSION =
 		(1 << 24) | //major
 		(21 << 16) | //minor
-		(0 << 8) | //patch
-		(3); //revision
+		(2 << 8) | //patch
+		(2); //revision
```
dktapps: This is only supposed to change if the block palette was updated (and a new blockstate upgrade schema added)

dktapps (changes requested):

src/data/bedrock/block/BlockStateData.php
Gaprix: But at the same time, PocketMine will not be able to open the vanilla world 1.21.2 without this change

Gaprix: Or do you mean a change specifically in this class?

dktapps: This specific change is invalid

dktapps (approved):

IvanCraft623 (approved):

PR #6403, opened 2024-07-14
fix(Explosion): don't apply motion if the event's cancelled

cooldogedev:


dries-c:
I remember suggesting this ages ago. The reason this never got through because technically motion isn't related to damage, so it can't fall under this event.

Matze997:
New event: EntityExplosionMotionEvent   haha

IvanCraft623:
vanilla also applies motion even if the damage has been cancelled

Matze997:
But isn't the goal of pocketmine to customise the game by changing certain behaviours?

IvanCraft623:
> But isn't the goal of pocketmine to customise the game by changing certain behaviours?

Yes, but specifically this customization should not be done through EntityDamageEvent.

ShockedPlot7560:
The aim is not to introduce behaviors that don't fit in with the api and the gameplay either

IvanCraft623:
`EntityDamageByEntityEvent` has a `knockback` property that is not being used for explosions, we could standardize the use of a knockback property in `EntityDamageEvent` that defaults to zero and take use of the property for explosions. So in this way we could cancel motion of explosions.

I'm pretty sure there was an issue proposing to move knockback property to EntityDamageEvent, I just couldn't find it.

dktapps:
Not really possible to knockback on base EntityDamageEvent since it doesn't have a source to knockback away from. It's possible it could be supported from EntityDamageByBlockEvent. However using existing knockback event APIs in general is a bit problematic because explosion knockback works differently than combat knockback.

PR #6405, opened 2024-07-17
Implemented a proper way to handle items cooldown

ipad54:
## Introduction
This PR implements a proper way to handle item coodowns. These changes are necessary to make goat horns PR (and later shields) work correctly.

### API changes
Added `Item::getCooldownTag() : ?string` to identify item cooldown group
Added `ItemCooldownTags` that stores a list of vanilla cooldown groups.
### Behaviour changes
Item cooldowns are now synchronized with client via `PlayerStartItemCooldownPacket`.
Item cooldown tags are now used to store item cooldowns. If item doesn't have such, stateId will be used.

### Relevant issues
#5232 

## Tests
Tested with goat horns PR.
Screenshot_20240717_142451_com mojang minecraftpe


IvanCraft623:
pocketmine in server-side is applying the cooldown to the items with the same `stateId`, while through `PlayerStartItemCooldownPacket` the cooldown is applied to all items with the same name.

This means for example that for example when using a goat horn, the server is applying cooldown to an specific variant of the item, while the client is applying cooldown for all goat horns.

So we have a mismatch here.

IvanCraft623:
Also i'm not sure if vanilla sends `PlayerStartItemCooldownPacket` to tell clients about cooldown.

ipad54:
> Also i'm not sure if vanilla sends `PlayerStartItemCooldownPacket` to tell clients about cooldown.

From my tests in BDS vanilla sends `PlayerStartItemCooldownPacket` for all new items that have cooldown (shields and goat hoarns). However, it doesn't send this packet for old items that were added back in MCPE days (chorus fruits and ender pearls). They have their cooldown implemented client-side. I think we can skip this and send the packet for all items?

ipad54:
> pocketmine in server-side is applying the cooldown to the items with the same `stateId`, while through `PlayerStartItemCooldownPacket` the cooldown is applied to all items with the same name.
> 
> This means for example that for example when using a goat horn, the server is applying cooldown to an specific variant of the item, while the client is applying cooldown for all goat horns.
> 
> So we have a mismatch here.

It looks like vanilla bothers only about item type. Would it be correct to change the logic in pocketmine to use `typeId` to store cooldown?

dktapps:
There's no direct link between PM item types and Minecraft item IDs. In some cases multiple IDs are flattened under the same item in PM, and in other cases a single ID is split into several PM-side items. Basically you can't rely on either typeID or stateID for this.

Possibly a better approach might be to introduce a "cooldown tag". PM would start a cooldown using the item's cooldown tag if it has one. Then any other items with the same cooldown tag can't be used until the cooldown expires.

ShockedPlot7560 (dismissed):
Looks ok to me

src/item/BaseCooldownTags.php
```diff
@@ -0,0 +1,40 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\item;
+
+/**
+ * Tags used by items to determine their cooldown group.
+ * @see Item::getCooldownTag()
+ */
+final class BaseCooldownTags{
```
dktapps: Why this naming?

ipad54: This class contains a set of cooldown tags for base vanilla items with cooldown

dktapps: It's not consistent with other similar classes

ShockedPlot7560 (dismissed):

src/item/VanillaCooldownTags.php
```diff
@@ -0,0 +1,40 @@
...
+
+	private function __construct(){
+		//NOOP
+	}
+
+	public const CHORUS_FRUIT = "chorus_fruit";
+	public const ENDER_PEARL = "ender_pearl";
+	public const SHIELD = "shield";
+	public const GOAT_HORN = "goat_horn";
+}
```
dktapps: It should be documented that:
> These tag values are not related to Minecraft internal IDs.
They only share a visual similarity because these are the most obvious values to use.
Any arbitrary string can be used.

Otherwise people might get confused and want to link them with the network IDs, and we really don't want that (no protocol stuff in the API).

(It's also possible we could make an interface and require people to use unbacked enums, but strings are easier).

src/item/VanillaCooldownTags.php
```diff
@@ -0,0 +1,40 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\item;
+
+/**
+ * Tags used by items to determine their cooldown group.
+ * @see Item::getCooldownTag()
+ */
+final class VanillaCooldownTags{
```
dktapps: `ItemCooldownTags` would be more consistent with other similar classes like `BlockTypeTags`

dktapps (changes requested):
See review comments.

src/item/ItemCooldownTags.php
```diff
@@ -27,12 +27,17 @@
...
 
 	private function __construct(){
 		//NOOP
 	}
 
+	/**
+	 * These tag values are not related to Minecraft internal IDs.
+	 * They only share a visual similarity because these are the most obvious values to use.
+	 * Any arbitrary string can be used.
+	 */
```
dktapps: This comment would be better placed in the class docblock so it shows up in the correct place on doc.pmmp.io

dktapps (approved):

IvanCraft623 (approved):
LGTM, thanks for your constribution!

ShockedPlot7560 (approved):

PR #6407, opened 2024-07-24
Add ShellCheck

zSALLAZAR:
## Introduction
This PR adds [ShellCheck](https://github.com/koalaman/shellcheck) to the CI

## Follow-up

Fix the checks that are probably going to fail

## Tests

- [ ] Other (provide details)


SOF3:
Is travis.sh still used?

dktapps:
Yep, it's used for integration tests: https://github.com/pmmp/PocketMine-MP/blob/df4ada81e5d74a14046f27cf44a37dcee69d657e/.github/workflows/main-php-matrix.yml#L119

I just never bothered to rename it when we shifted to GitHub Actions

ShockedPlot7560:
Your CI is failing

zSALLAZAR:
> Your CI is failing

Fixed

IvanCraft623:
Looks OK! Thanks for your contribution!

IvanCraft623 (approved):

tests/travis.sh
```diff
@@ -45,7 +48,7 @@ if [ "$result" != "" ]; then
 	echo "$result"
 	echo Some tests did not complete successfully, changing build status to failed
 	exit 1
-elif [ $(grep -c "ERROR\|CRITICAL\|EMERGENCY" "$DATA_DIR/server.log") -ne 0 ]; then
+elif [ "$(grep -c "ERROR\|CRITICAL\|EMERGENCY" "$DATA_DIR/server.log")" -ne 0 ]; then
```
dktapps, line 51: Do we not need to escape the inner quotes here?

ShockedPlot7560: Maybe the CI failling if not ? I wouldn't know why else

dktapps: Yeah you're right. The PR is its own proof of validity I guess.

dktapps: Looks like there's no escaping needed according to this: https://unix.stackexchange.com/questions/118433/quoting-within-command-substitution-in-bash

dktapps (approved):

PR #6414, opened 2024-08-07
Fix effect bubbles not showing

IvanCraft623:
As of 1.21 bubbles for each effect are displayed separately (it is no longer a mix of all visible effect colors), so it requires that we send all visible effect ids.

## Tests
- [x] Ensured that the values generated were identical to those generated by BDS.

### DIFF
--- a/src/entity/Living.php
+++ b/src/entity/Living.php
@@ -63,20 +63,19 @@ use pocketmine\world\sound\EntityLandSound;
 use pocketmine\world\sound\EntityLongFallSound;
 use pocketmine\world\sound\EntityShortFallSound;
 use pocketmine\world\sound\ItemBreakSound;
-use function array_filter;
-use function array_map;
 use function array_shift;
+use function array_slice;
 use function atan2;
 use function ceil;
 use function count;
 use function floor;
+use function ksort;
 use function lcg_value;
 use function max;
 use function min;
 use function mt_getrandmax;
 use function mt_rand;
 use function round;
-use function sort;
 use function sqrt;
 use const M_PI;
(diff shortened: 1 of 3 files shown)


src/entity/Living.php
```diff
@@ -887,22 +886,21 @@ protected function sendSpawnPacket(Player $player) : void{
...
+		}
 
-		//TODO: HACK! the client may not be able to identify effects if they are not sorted.
-		sort($visibleEffectIds, SORT_NUMERIC);
+		ksort($visibleEffects, SORT_NUMERIC);
 
 		$effectsData = 0;
-		foreach ($visibleEffectIds as $effectId) {
-			$effectsData = ($effectsData << 7) | ($effectId << 1);
+		foreach (array_slice($visibleEffects, 0, 8, true) as $effectId => $isAmbient) {
```
dktapps: This allocates an extra array, a counter to break would be better

dktapps (commented):
Not sure why my review was requested for this. I don't know anything about this new stuff. I only stopped by to leave a code style comment.

dries-c:
Seems like this also resolves the issue where the second time the same effect is issued it won't show up anymore.

dries-c (dismissed):

ShockedPlot7560 (dismissed):

IvanCraft623:
@dries-c @ShockedPlot7560 Re-review :D

ShockedPlot7560:
:D

ShockedPlot7560 (approved):

dries-c (approved):

PR #6418, opened 2024-08-10
Implement anvil

ShockedPlot7560:
## Introduction
As the title suggests, this PR implements the anvils

### Relevant issues
* #6305 PR abandoned and reworked

### API changes
* Introduce a new `Durable::isValidRepairMaterial` enabling everyone to implement their own verification logic 
* Add `repairMaterials` to `ArmorMaterial` and `ToolTier`
* Add repairCost functions in `Item`

### Behavioural changes
* Anvil works

## Tests
- [X] Playtesting using a Minecraft client (provide screenshots or a video)
- [ ] Other (provide details)

https://youtu.be/d3vcwi_45DY


src/block/utils/AnvilHelper.php
```diff
@@ -0,0 +1,189 @@
...
+use pocketmine\item\enchantment\Rarity;
+use pocketmine\item\Item;
+use pocketmine\player\Player;
+use function ceil;
+use function floor;
+use function max;
+use function min;
+use function strlen;
+
+class AnvilHelper{
```
IvanCraft623: This should be final

src/item/Item.php
```diff
@@ -282,6 +284,23 @@ public function clearNamedTag() : Item{
...
+	/**
+	 * Sets the repair cost of the item.
+	 *
+	 * @return $this
+	 */
+	public function setRepairCost(int $cost) : self{
+		$this->repairCost = $cost;
+		return $this;
+	}
+
```
IvanCraft623, line 311: This should be placed on Durable class

src/item/TieredTool.php
```diff
@@ -61,4 +63,8 @@ public function getFuelTime() : int{
 	public function isFireProof() : bool{
 		return $this->tier === ToolTier::NETHERITE;
 	}
+
+	public function isValidRepairMaterial(Item $material) : bool{
```
IvanCraft623: i don't like this name, feel that the material terminology doesn't fit, instead i would name it something like canBeRepairedBy or isRepairedBy

src/block/utils/AnvilHelper.php
IvanCraft623: I would like this class to be divided into actions, for example an AnvilRapairAction, in the actions would define properties such as the cost and process the item

IvanCraft623 (changes requested):
Just a quick overview

src/item/Item.php
ShockedPlot7560: Any item can have repairCost, e.g EnchantedBook

src/block/utils/AnvilHelper.php
ShockedPlot7560: Like the idea

src/event/player/PlayerUseAnvilEvent.php
Dumont-Julien, line 1: implement possibility to set item and not juste get ?

src/item/Item.php
dadodasyra: How do you repair an enchanted book ?

src/item/TieredTool.php
dadodasyra: Looks more like french tbh 😂

src/event/player/PlayerUseAnvilEvent.php
dadodasyra: Maybe there's some client side prediction about recipe (I honestly dont know) which may prevent it to work

src/item/Item.php
ipad54: > How do you repair an enchanted book ?

it's more like “anvil cost” rather than actual repair cost since it also increases when combining enchantments / renaming.

src/item/TieredTool.php
ShockedPlot7560: This is the name used in the wiki to refer to materials.
But just specifying, `canBeRepairedBy` can lead to confusion in thinking that it can be both the materials and the tool/armour itself. Whereas what we're expecting here is just the materials

src/item/Item.php
ShockedPlot7560: So should we move away from vanilla and go in the direction of AnvilCost and not RepairCost in the item tags? I personally don't think so.

ipad54: Consistency within PM is more important, right? I think we can use anvil cost terminology and just document that in vanilla it is called repair cost.

ShockedPlot7560: Yeah but the tag need to stay at RepairCost

src/block/anvil/AnvilAction.php
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @author PocketMine Team
+ * @link http://www.pocketmine.net/
+ *
+ *
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\block\anvil;
```
ipad54: Does it need to be placed in block namespace? Wouldn’t be better to move anvil inventory actions logic into the `inventory\transaction` namespace since it’s more associated with anvil transactions rather than block itself?

src/item/Item.php
IvanCraft623: Okay, so document and/or name this function properly, when I saw it I thought it was the experience cost of the item to be repaired.

src/event/player/PlayerUseAnvilEvent.php
IvanCraft623: yep, no posibility for custom recipes :(

src/block/anvil/AnvilAction.php
IvanCraft623: `inventory\transaction` namespace is intended for handling player's requested ones, but in this use case  what we are actually doing is creating fictitious actions for the use of the anvil.

src/block/anvil/AnvilActionsFactory.php
```diff
@@ -0,0 +1,71 @@
...
+
+	/**
+	 * Return all available actions for the given items.
+	 *
+	 * @return AnvilAction[]
+	 */
+	public function getActions(Item $base, Item $material, ?string $customName) : array{
+		$actions = [];
+		foreach($this->actions as $class => $_){
+			$action = new $class($base, $material, $customName);
```
dktapps: Oh, I really don't like this.

src/inventory/transaction/AnvilTransaction.php
```diff
@@ -0,0 +1,157 @@
...
+			$world->addSound($inventory->getHolder(), new AnvilUseSound());
+		}
+	}
+
+	protected function callExecuteEvent() : bool{
+		if($this->baseItem === null){
+			throw new AssumptionFailedError("Expected that baseItem are not null before executing the event");
+		}
+
+		$ev = new PlayerUseAnvilEvent($this->source, $this->baseItem, $this->materialItem, $this->expectedResult->getResult() ?? throw new \AssertionError(
```
dktapps: This should be `PlayerAnvilUseEvent` to match other event naming conventions

src/item/Armor.php
```diff
@@ -172,4 +173,8 @@ protected function serializeCompoundTag(CompoundTag $tag) : void{
 			$tag->setInt(self::TAG_CUSTOM_COLOR, Binary::signInt($this->customColor->toARGB())) :
 			$tag->removeTag(self::TAG_CUSTOM_COLOR);
 	}
+
+	public function isValidRepairMaterial(Item $material) : bool{
+		return in_array($material->getTypeId(), $this->armorInfo->getMaterial()->getRepairMaterials(), true);
```
dktapps: don't use in_array :(

src/item/ToolTier.php
```diff
@@ -50,23 +51,38 @@ enum ToolTier{
...
+				ItemTypeIds::fromBlockTypeId(BlockTypeIds::SPRUCE_PLANKS),
+				ItemTypeIds::fromBlockTypeId(BlockTypeIds::BIRCH_PLANKS),
+				ItemTypeIds::fromBlockTypeId(BlockTypeIds::JUNGLE_PLANKS),
+				ItemTypeIds::fromBlockTypeId(BlockTypeIds::ACACIA_PLANKS),
+				ItemTypeIds::fromBlockTypeId(BlockTypeIds::DARK_OAK_PLANKS),
+				ItemTypeIds::fromBlockTypeId(BlockTypeIds::CRIMSON_PLANKS),
+				ItemTypeIds::fromBlockTypeId(BlockTypeIds::WARPED_PLANKS),
+				ItemTypeIds::fromBlockTypeId(BlockTypeIds::CHERRY_PLANKS),
+				ItemTypeIds::fromBlockTypeId(BlockTypeIds::MANGROVE_PLANKS)
+			]),
```
dktapps: Not liking this either.

dktapps (commented):
My gut reaction to this is: I don't like the code.

I don't think we should be using custom actions for this since the actions are only usable with specific types of transactions. By the way other stuff has been done, different AnvilTransaction classes should be implemented.

I'm also not liking the logic of some of the item comparisons being limited to checking type IDs. Plugins will likely want to implement more complex logic, like using tags or NBT checking.

dktapps (commented):
I think it would also probably be preferable to have more events, e.g. `PlayerItemRenameEvent`, `PlayerItemRepairEvent`, `PlayerItemCombineEvent` (not sure about this last one).  (That being said, if there were transaction types for this instead of actions, it would be a good amount nicer).

Doesn't make much sense to make plugins do complex logic to figure out the action type when we can just tell them directly.

src/inventory/transaction/AnvilTransaction.php
```diff
@@ -0,0 +1,155 @@
...
+				$anvilBlock = $world->getBlock($inventory->getHolder());
+				if($anvilBlock instanceof Anvil){
+					$newDamage = $anvilBlock->getDamage() + 1;
+					if($newDamage > Anvil::VERY_DAMAGED){
+						$newBlock = VanillaBlocks::AIR();
+						$world->addSound($inventory->getHolder(), new AnvilBreakSound());
+					}else{
+						$newBlock = $anvilBlock->setDamage($newDamage);
+					}
+					$world->setBlock($inventory->getHolder(), $newBlock);
```
kostamax27: It would be nice to be able to cancel this

reacherurmom:
I did a little test:
I enchanted two swords through the /enchant command and when I went to combine them in the anvil the cost was 1 xp

ShockedPlot7560:
Here is a TODO list to complete this PR for me: 
- [x] Add an event to be able to cancel the anvil damage
- [ ] Add an event or adjust events to be able to change the custom name of an item
- [ ] Clear and precise unit tests for anvil crafting recipes
- [ ] Validate in game the gameplay
- [ ] Validate Anvil cost with unit tests & in game
- [x] Include book combinaison

kostamax27:
> Here is a TODO list to complete this PR for me:

How about an event for canceling or changing an item’s name?

Possible uses: the server doesn’t allow colored names, censorship, etc.

ShockedPlot7560:
You already have `PlayerUseAnvilEvent` to cancel this but you can't change anything yep

JkqzDev:
> Here is a TODO list to complete this PR for me:
> 
> * [ ]  Add an event to be able to cancel the anvil damage
> * [ ]  Add an event or adjust events to be able to change the custom name of an item
> * [ ]  Clear and precise unit tests for anvil crafting recipes
> * [ ]  Validate in game the gameplay
> * [ ]  Validate Anvil cost with unit tests & in game

I have tested this fork and with the changes you have made I think that the combinations of items with books still do not work. 
I have added it in my fork but I don't know if it is to your taste: https://github.com/RequiemPvP/PocketMine-MP/blob/dbbb4f8fb1af2ef45a77fcc495715bae9bd94aef/src/crafting/AnvilCraftingManagerDataFiller.php#L76

ShockedPlot7560:
Thank you @JkqzDev. 
From what I can see, your implementation looks good, but it raises me one question.
Can any items really be combined with an enchanted book for Minecraft Bedrock?

JkqzDev:
> Thank you @JkqzDev. From what I can see, your implementation looks good, but it raises me one question. Can any items really be combined with an enchanted book for Minecraft Bedrock?

Good question @ShockedPlot7560 
When I implemented it, I also had the same question, but I think this answers the question: https://github.com/RequiemPvP/PocketMine-MP/blob/dbbb4f8fb1af2ef45a77fcc495715bae9bd94aef/src/crafting/ItemCombineRecipe.php#L53
I'm not sure if it actually works the way it's supposed to.

dktapps:
Makes sense to me, though it might be worth checking if there are any primary vs secondary restrictions

JkqzDev:
@ShockedPlot7560 I remember that I sent you a message about a problem that happened with Anvil: the xp cost that appeared in the UI was not always the real cost.

ShockedPlot7560:
It was part of the unit tests which currently failed a lot

JkqzDev:
https://github.com/JkqzDev/PocketMine-MP/blob/3f7d270c41aeb6cd0d4c6f5738b6ea6db776f1a7/src/crafting/ItemCombineRecipe.php#L93

I have made changes to this PR. I'm not sure it has been solved because unit tests continue to fail but within the game (from what I have tested) it works fine.
@ShockedPlot7560

PR #6421, opened 2024-08-13
Fixed incorrect function visibility modifier in ChiseledBookshelf tile class

ipad54:
## Introduction
`writeSaveData()` is protected in the base class.

## Backwards compatibility
This is not a BC break change since no plugin should use this internal method.

## Tests
PHPStan tests are OK.


dktapps (approved):

IvanCraft623 (approved):
Looks OK! Thanks for your contribution :D

PR #6422, opened 2024-08-13
Release patch update 5.17.1

IvanCraft623:


changelogs/5.17.md
```diff
@@ -23,3 +23,15 @@ Consider using the `mcpe-protocol` directive in `plugin.yml` as a constraint if
...
 - Bells always drops themselves even when using an incompatible tool.
+
+# 5.17.1
+Released 13th August 2024.
+
+## Documentation
+- Added a note about `BlockStateData::CURRENT_VERSION`.
+
+## Fixes
+- Fixed outdated `BedrockWorldData` version, this was preventing use newer worlds.
```
dktapps, line 35: Anvil rotation too

dktapps: never mind, looks like this branch doesn't have that change

IvanCraft623: Oops, seems to it was merged just as I was writing the changelogs

dktapps: Yeah I figured. The release process really needs to be rethought.

IvanCraft623: Solved

PR #6427, opened 2024-08-16
Wind Charge Implementation

cjmustard:
## Introduction
Implementation for the 1.21.0 `Wind Charge`

### Relevant issues
N/A

## Details
- Non-iron doors and trapdoors are flipped.
- Fence gates are flipped.
- Buttons are pressed.
- Levers are flipped.
- Bells are rung and swung.
- Chorus flowers and pointed dripstone blocks break, dropping in item form.
- Lit candles (both standalone and on cake) are extinguished.

Wind charges can be deflected if hit with another projectile or a melee attack, similar to ghast fireballs.
Wind charges are distance attenuated by liquids such as water and lava. They also appear to fly in a straight line, similar to fireballs
Wind charges are set to despawn after five minuets due to the gravity less nature of the entity.

https://minecraft.wiki/w/Wind_Charge

## Changes
One method has been added to pocketmine\block\Block:
```php
	/**
	 * Called when this block is updated by a state altering projectile in the world.
	 *
	 * @param pocketmine\entity\projectile\Projectile $projectile The projectile affecting the block.
	 */
	public function onProjectileInteraction(pocketmine\entity\projectile\Projectile $projectile) : void {

	}
```

Both class' for the Wind Charge item and Projectile have been created:
- pocketmine\entity\projectile\WindCharge.php
- pocketmine\item\WindCharge.php

One class has been created to handle the wind charge explode particle:
- pocketmine\world\particle\WindExplosionParticle.php

One class has been added to handle wind charge burst sound:
- pocketmine/world/sound/WindChargeBurstSound.php

The Wind Charge has been implemented as a vanilla item.
The Wind Charge Entity has been implemented as an Entity.

### API changes
N/A

### Behavioural changes
N/A

## Backwards compatibility
N/A

## Follow-up
N/A

## Tests
- [ ] Other (provide details)

# Behavior With Other Blocks
https://github.com/pmmp/PocketMine-MP/assets/80650734/e27ca3d6-f086-432b-bff1-72be8c252557

# Behavior With Players (shows less effective under water.)
https://github.com/pmmp/PocketMine-MP/assets/80650734/2576fc31-f9cb-46de-a2c6-28e39964849a

# Behavior Multi Hit
https://github.com/pmmp/PocketMine-MP/assets/80650734/0250cdab-350e-4586-bf7a-2340195e0429


cjmustard (commented):
Newly re-based PR...

src/world/sound/WindChargeBurstSound.php
```diff
@@ -0,0 +1,34 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\world\sound;
+
+use pocketmine\math\Vector3;
+use pocketmine\network\mcpe\protocol\LevelSoundEventPacket;
+use pocketmine\network\mcpe\protocol\types\LevelSoundEvent;
+
+final class WindChargeBurstSound implements Sound{
```
dadodasyra, line 30: Missing leading space after class definition

dadodasyra (commented):
Except this it looks good to me, I don't understand why does Github web preview looks like indentation is off but it seems not to.

src/world/sound/WindChargeBurstSound.php
cjmustard: this is how php-cs-fixer formatted it 🤣

src/block/ChorusFlower.php
```diff
@@ -118,6 +119,12 @@ private function allHorizontalBlocksEmpty(World $world, Vector3 $position, ?int
 		return true;
 	}
 
+	public function onProjectileInteraction(Projectile $projectile) : void{
+		if($projectile instanceof WindCharge) {
+			$this->position->getWorld()->useBreakOn($this->position);
```
ipad54: From my tests Chorus Flowers aren't affected by this. They only break on direct projectile hit which is already covered. So, this is unnecessary.

src/entity/projectile/WindCharge.php
```diff
@@ -0,0 +1,144 @@
...
+
+		$bound = $this->getBound($this->getPosition(), self::RADIUS - 1);
+		for($x = $bound->minX; $x <= $bound->maxX; $x++) {
+			for($y = $bound->minY; $y <= $bound->maxY; $y++) {
+				for($z = $bound->minZ; $z <= $bound->maxZ; $z++) {
+					$block = $this->getWorld()->getBlockAt((int) floor($x), (int) floor($y), (int) floor($z));
+
+					// This is to avoid calling Door::onProjectileInteraction() twice due to two tiles.
+					if($block instanceof Door) {
+						if($block->isTop()) {
```
ipad54: I think it would be better to check this in `onProjectileInteraction()` in Door class.

src/item/WindCharge.php
```diff
@@ -0,0 +1,48 @@
...
+
+use pocketmine\entity\Location;
+use pocketmine\entity\projectile\Throwable;
+use pocketmine\entity\projectile\WindCharge as WindChargeEntity;
+use pocketmine\player\Player;
+
+class WindCharge extends ProjectileItem{
+
+	public function getMaxStackSize() : int{
+		return 64;
```
ipad54: This is unnecessary. Items already have their max stack size set to 64 by default.

src/block/Candle.php
```diff
@@ -98,6 +100,17 @@ protected function getCandleIfCompatibleType(Block $block) : ?Candle{
 		return $block instanceof Candle && $block->hasSameTypeId($this) ? $block : null;
 	}
 
+	public function onProjectileInteraction(Projectile $projectile) : void{
+		if(!$this->lit) {
+			return;
+		}
+
+		$newCandle = $this->setLit(false);
```
ipad54: I think you should check here too that projectile is wind charge.

src/block/ChorusFlower.php
CJMustard1452: @ipad54 it seems to work on my machine (joke intended)

https://streamable.com/mtr10y

cjmustard: nevermind you meant in vanilla 😆

ipad54:
It's seems like fall distance compensation is missing.

In vanilla, a player launched by wind charges shot by a player accumulates fall damage only when falling below the Y level where the wind burst hit.

src/block/Candle.php
```diff
@@ -98,6 +101,19 @@ protected function getCandleIfCompatibleType(Block $block) : ?Candle{
 		return $block instanceof Candle && $block->hasSameTypeId($this) ? $block : null;
 	}
 
+	public function onProjectileInteraction(Projectile $projectile) : void{
+		if($projectile instanceof WindCharge) {
+			if(!$this->lit) {
+				return;
+			}
```
ShockedPlot7560: It would be preferable to move on to the condition above.

src/block/CakeWithCandle.php
```diff
@@ -59,6 +62,18 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
 		return parent::onInteract($item, $face, $clickVector, $player, $returnedItems);
 	}
 
+	public function onProjectileInteraction(Projectile $projectile) : void{
+		if($projectile instanceof WindCharge) {
+			if(!$this->lit) {
+				return;
+			}
```
ShockedPlot7560: same

src/block/Button.php
```diff
@@ -91,4 +89,19 @@ public function onNearbyBlockChange() : void{
 	private function canBeSupportedAt(Block $block, int $face) : bool{
 		return $block->getAdjacentSupportType(Facing::opposite($face))->hasCenterSupport();
 	}
+
+	public function onProjectileInteraction(Projectile $projectile) : void{
+		if($projectile instanceof WindCharge) {
```
ShockedPlot7560, line 94: need to check if its not pressed

src/block/Door.php
```diff
@@ -176,4 +168,34 @@ public function getAffectedBlocks() : array{
...
 	}
+
+	public function onProjectileInteraction(Projectile $projectile) : void{
+		if($projectile instanceof WindCharge) {
+			if($this->getTypeId() === BlockTypeIds::IRON_DOOR) {
+				return;
+			}
+
+			// This is to avoid calling Door::onProjectileInteraction() twice due to two tiles.
+			if($this->top) {
```
ShockedPlot7560, line 179: So, if the projectile hit the upper one, no toggle will be happened ?

src/block/Trapdoor.php
```diff
@@ -85,10 +87,26 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
...
+		$this->toggle();
+
+		return true;
+	}
+
+	public function onProjectileInteraction(Projectile $projectile) : void{
+		if($projectile instanceof WindCharge){
+			if($this->getTypeId() === BlockTypeIds::IRON_TRAPDOOR) {
+				return;
+			}
```
ShockedPlot7560: Would be better to moove it into the condition above?

ShockedPlot7560:
An additional comment: isn't it better to check whether it's a windcharge by the typeId rather than the instance?

src/block/Door.php
cjmustard: Yeah, this is my best attempt at handling this issue, if it hits both blocks the door hits, (open, close)

ShockedPlot7560: Can be useful before toggling to check the other state. If its not the same as ours when projectile hits, we can ignore ? Just a theory

src/block/Button.php
cjmustard: $this->pressed(); in this context, thereby extending the pushed duration if already down, I do not see a need to change this, thoughts?

src/block/Door.php
cjmustard: feels a little hacky to be honest

ShockedPlot7560: Maybe hacky but can avoid to have unexpected side effect if only the top block is hitten by the effect area.

src/block/Button.php
ShockedPlot7560: by checking that it is not activated, you will avoid extra sound.

src/item/WindCharge.php
```diff
@@ -0,0 +1,44 @@
+<?php
```
KnosTx, line 1: Wow this is a really great idea 👍👍👍👍

src/block/Trapdoor.php
```diff
@@ -85,10 +87,22 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
 	}
 
 	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		$this->toggle();
+
+		return true;
+	}
+
+	public function onProjectileInteraction(Projectile $projectile) : void{
+		if($projectile instanceof WindCharge && $this->getTypeId() !== BlockTypeIds::IRON_TRAPDOOR){
```
dktapps, line 96: Not liking the iron trapdoor check hardcoded in here.

src/block/Trapdoor.php
```diff
@@ -85,10 +87,22 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
...
+			$this->toggle();
+		}
+	}
+
+	public function toggle() : void {
```
dktapps, line 101: This should be `private` or `protected`, making it public has implications beyond this PR's scope

src/block/Lever.php
```diff
@@ -91,14 +93,25 @@ public function onNearbyBlockChange() : void{
...
+		return true;
+	}
+
+	public function onProjectileInteraction(Projectile $projectile) : void{
+		if($projectile instanceof WindCharge) {
+			$this->toggle();
+		}
+	}
+
+	public function toggle() : void {
```
dktapps, line 107: This should be `private` or `protected`, making it public has implications beyond this PR's scope

src/block/Door.php
```diff
@@ -176,4 +168,34 @@ public function getAffectedBlocks() : array{
...
+				return;
+			}
+
+			$this->toggle();
+		}
+	}
+
+	public function toggle() : void {
```
dktapps, line 187: This should be `private` or `protected`, making it public has implications beyond this PR's scope

src/block/Door.php
```diff
@@ -176,4 +168,34 @@ public function getAffectedBlocks() : array{
 	private function canBeSupportedAt(Block $block) : bool{
 		return $block->getAdjacentSupportType(Facing::DOWN)->hasEdgeSupport();
 	}
+
+	public function onProjectileInteraction(Projectile $projectile) : void{
+		if($projectile instanceof WindCharge) {
+			if($this->getTypeId() === BlockTypeIds::IRON_DOOR) {
```
dktapps, line 174: Not liking this check, same as before with trapdoors

src/block/Button.php
```diff
@@ -91,4 +89,19 @@ public function onNearbyBlockChange() : void{
...
+			$this->press();
+		}
+	}
+
+	public function press() : void {
```
dktapps, line 99: This should be `private` or `protected`, making it public has implications beyond this PR's scope

src/block/Block.php
```diff
@@ -513,6 +513,15 @@ public function onScheduledUpdate() : void{
 
 	}
 
+	/**
+	 * Called when this block is updated by a state altering projectile in the world.
```
dktapps, line 517: This sentence doesn't mean anything to me. It doesn't tell me why an update was triggered, nor what kind of state alterations should take place.

The function is trying to look and sound generic, but it's not.
"interaction" is also too vague.

dktapps (changes requested):

src/block/Door.php
dktapps: Yep, only checking the one half will cause bugs. It's the same as if a player were clicking the top or bottom half, so the logic should be basically the same.

cjmustard:
open to anyone willing to work on this pr, no further work will come from me as of this point.

PR #6429, opened 2024-08-17
Implemented vanilla /xp command.

ipad54:
## Introduction
This PR implements /xp command from vanilla. Also, I think it would be convenient to have a built-in command that can give you experience levels to test enchanting tables / anvils instead of farming experience from bottles.

### API changes
Added `XpCommand`
Added 2 new permissions:
- pocketmine.command.xp.other
- pocketmine.command.xp.self

## Follow-up
Before accepting this PR you need to add new translations into `pmmp/Language` repo

See https://github.com/pmmp/Language/pull/185

### Requires translations:

| `pocketmine.command.xp.description` | `Adds or removes player experience` |

| `pocketmine.permission.command.xp.other` | `Allows the user to add or remove experience of other players` |
| `pocketmine.permission.command.xp.self` | `Allows the user to add or remove their experience` |
| `pocketmine.command.xp.usage` | `/xp <experience[L]> [player]` |

| `commands.xp.failure.widthdrawXp` | `Cannot give player negative experience points` |
| `commands.xp.success` | `Gave {%0} experience to {%1}` |
| `commands.xp.success.levels` | `Gave {%0} experience levels to {%1}` |
| `commands.xp.success.negative.levels` | `Taken {%0} levels from {%1}` |

## Tests
I tested this PR by doing the following:
- [X] Playtesting using a Minecraft client 
Screenshot_20240817_143222


src/command/defaults/ExperienceCommand.php
```diff
@@ -0,0 +1,86 @@
...
+			"/xp <experience[L]> [player]"
+		);
+		$this->setPermissions([
+			DefaultPermissionNames::COMMAND_XP_SELF,
+			DefaultPermissionNames::COMMAND_XP_OTHER
+		]);
+	}
+
+	public function execute(CommandSender $sender, string $commandLabel, array $args){
+		if(count($args) < 1){
```
ShockedPlot7560:
```suggestion
		if(count($args) !== 2){
```

src/command/defaults/ExperienceCommand.php
```diff
@@ -0,0 +1,86 @@
...
+			if($xp >= 0){
+				$xpManager->setXpLevel(min($xpManager->getXpLevel() + $xp, $maxXp));
+				$sender->sendMessage("Gave $xp experience levels to " . $player->getName());
+			}else{
+				$xp = abs($xp);
+				$xpManager->setXpLevel(max($xpManager->getXpLevel() - $xp, 0));
+				$sender->sendMessage("Taken $xp levels from " . $sender->getName());
+			}
+			return true;
+		}
```
ShockedPlot7560: Delete this return and place the following code in an else. More conventional

src/command/defaults/ExperienceCommand.php
```diff
@@ -0,0 +1,86 @@
...
+			throw new InvalidCommandSyntaxException();
+		}
+		$player = $this->fetchPermittedPlayerTarget($sender, $args[1] ?? null, DefaultPermissionNames::COMMAND_XP_SELF, DefaultPermissionNames::COMMAND_XP_OTHER);
+		if($player === null){
+			return true;
+		}
+		$xpManager = $player->getXpManager();
+		if(str_ends_with($args[0], "L")){
+			$xpAttr = $player->getAttributeMap()->get(Attribute::EXPERIENCE_LEVEL) ?? throw new AssumptionFailedError();
+			$maxXp = (int) $xpAttr->getMaxValue();
```
ShockedPlot7560: this is maxXp level

src/command/defaults/ExperienceCommand.php
```diff
@@ -0,0 +1,86 @@
...
+			$xp = $this->getInteger($sender, substr($args[0], 0, -1), -$maxXp, $maxXp);
```
ShockedPlot7560: same here, this is the xpLevel, not just the xp

ShockedPlot7560: Btw, the min is the invert of the current level. We can't remove less than level we have. The max is the max minus the current level.

src/command/defaults/ExperienceCommand.php
```diff
@@ -0,0 +1,86 @@
...
+			if($xp >= 0){
+				$xpManager->setXpLevel(min($xpManager->getXpLevel() + $xp, $maxXp));
```
ShockedPlot7560: If we apply the min/max logics into the getInteger, we can remove this one

src/command/defaults/ExperienceCommand.php
```diff
@@ -0,0 +1,86 @@
...
+				$sender->sendMessage("Gave $xp experience levels to " . $player->getName());
+			}else{
+				$xp = abs($xp);
+				$xpManager->setXpLevel(max($xpManager->getXpLevel() - $xp, 0));
```
ShockedPlot7560: same here

src/command/defaults/ExperienceCommand.php
ipad54: I did this because in vanilla it prints the number of entered xp in command output, no matter how much xp you were actually given, but i think we can step back off these rules since we don't have vanilla command parser.

src/command/defaults/ExperienceCommand.php
ipad54: This is incorrect. “Player” is optional argument and equals to sender by default

ShockedPlot7560: you're right

src/command/defaults/ExperienceCommand.php
```diff
@@ -0,0 +1,85 @@
...
+		if(count($args) < 1){
+			throw new InvalidCommandSyntaxException();
+		}
+		$player = $this->fetchPermittedPlayerTarget($sender, $args[1] ?? null, DefaultPermissionNames::COMMAND_XP_SELF, DefaultPermissionNames::COMMAND_XP_OTHER);
+		if($player === null){
+			return true;
+		}
+		$xpManager = $player->getXpManager();
+		if(str_ends_with($args[0], "L")){
+			$xpLevelAttr = $player->getAttributeMap()->get(Attribute::EXPERIENCE_LEVEL) ?? throw new AssumptionFailedError();
```
IvanCraft623:
```suggestion
			/** @var Attribute $xpLevelAttr */
			$xpLevelAttr = $player->getAttributeMap()->get(Attribute::EXPERIENCE_LEVEL);
```

ipad54: IIRC, `@var` in PM code shouldn’t be used for such assertions. Also, I’ve found similar assertion via `AssumptionFailedError` in the code for another attribute that should always exist by default too (https://github.com/pmmp/PocketMine-MP/blob/stable/src/network/mcpe/handler/InGamePacketHandler.php#L510)

dktapps: Yep, `@var` should only be used if the IDE can't figure out what's going on. AssumptionFailedError is correct here.

src/command/defaults/ExperienceCommand.php
```diff
@@ -0,0 +1,85 @@
...
+use pocketmine\permission\DefaultPermissionNames;
+use pocketmine\utils\AssumptionFailedError;
+use pocketmine\utils\Limits;
+use pocketmine\utils\TextFormat;
+use function abs;
+use function count;
+use function str_ends_with;
+use function substr;
+
+class ExperienceCommand extends VanillaCommand{
```
dktapps: `XpCommand` would probably be more consistent with other XP-related things

dktapps:
Translations are available as of 2d9cee3d6232b12991a1ec1f9bf1cf8e8f7e72b3

dktapps (approved):

PR #6433, opened 2024-08-22
Fix shift crafting

dries-c:
## Introduction
In versions before 1.21.20, the client will send multiple CraftRecipeStackRequestActions to craft all the items in your inventory. However, since the 1.21.20 shift crafting would throw an error since it now uses repetitions instead of multiple actions and you'd be placing more items in your inventory than you actually crafted from that one action.

## Tests
https://github.com/user-attachments/assets/70b2dcfe-8fdd-4452-be3e-54881cfcc4d5

- [ ] Other (provide details)

### DIFF
--- a/src/network/mcpe/handler/ItemStackRequestExecutor.php
+++ b/src/network/mcpe/handler/ItemStackRequestExecutor.php
@@ -343,7 +343,7 @@ class ItemStackRequestExecutor{
 					$this->setNextCreatedItem($window->getOutput($optionId));
 				}
 			}else{
-				$this->beginCrafting($action->getRecipeId(), 1);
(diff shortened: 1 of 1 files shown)


ShockedPlot7560 (commented):
Simple, but could this have any side effects ?

Superomarking:
> Simple, but could this have any side effects ?

From what I checked, not really

ShockedPlot7560 (approved):

dktapps:
What about the creative item creation, or are you saving that for a separate PR?

dries-c:
I still have to do that, but I'm quite busy now. Maybe someone else can take that.

PR #6440, opened 2024-09-01
Implement missing last interacted slot property in chiseled bookshelf

ipad54:
## Introduction
This PR implements missing `last interacted slot` property that is used in vanilla worlds for redstone power. Without this PR chiseled bookshelves loaded from vanilla worlds will discard their redstone power level.

Source: https://minecraft.wiki/w/Bedrock_Edition_level_format/Block_entity_format#Chiseled_Bookshelf

### API changes
Added the following methods:
- `ChiseledBookshelf::getLastInteractedSlot() : ?ChiseledBookshelfSlot`
- `ChiseledBookshelf::setLastInteractedSlot(?ChiseledBookshelfSlot $lastInteractedSlot) : self`

## Tests
Since PM doesn't have redstone implemented, I added some debug messages to the chat locally to test this.

Screenshot_20240901_120441
Screenshot_20240901_120339


IvanCraft623 (dismissed):
LGTM, thanks for your constribution!

src/block/ChiseledBookshelf.php
```diff
@@ -48,11 +48,30 @@ class ChiseledBookshelf extends Opaque{
...
 	protected function describeBlockOnlyState(RuntimeDataDescriber $w) : void{
 		$w->horizontalFacing($this->facing);
 		$w->enumSet($this->slots, ChiseledBookshelfSlot::cases());
 	}
 
+	public function readStateFromWorld() : Block{
+		$tile = $this->position->getWorld()->getTile($this->position);
+		if($tile instanceof TileChiseledBookshelf){
+			$this->lastInteractedSlot = $tile->getLastInteractedSlot();
+		}
```
dktapps, line 64: This should have an `else` to set the property to `null` if the tile is missing

ipad54: This property is initialized to null by default. It isn't enough?

dktapps: On closer examination it appears some other blocks do the same thing you've done. Probably ought to be fixed, but that's for another time.

In core code this is only called in `getBlockAt()`, but there's no telling what plugins might do. There's no guarantee that the property is `null`.

dktapps (approved):

IvanCraft623 (approved):

PR #6442, opened 2024-09-02
Release 5.18.1 (patch)

IvanCraft623:


SOF3:
@IvanCraft623 are you sure it is a good idea to name a branch `5.18.1` when we are going to release a tag with the same name?

dktapps:
it shouldn't make any difference

dktapps (approved):

SOF3:
There are some places that mix the branch and tag namespaces, e.g. GitHub's `/blob/<tree>`, `git checkout <tree>`, etc. However it should probably be fine after deleting the branch indeed.

PR #6445, opened 2024-09-04
Regeneration 5 => 2 matching Java

WavyCraftNetwork:
## Introduction

In recent updates the Enchanted Golden Apple has been nerfed to match Java edition 

### API changes

Nerfed Enchanted Golden Apple to match Java edition

### Behavioural changes

No since it only affects 1 item

## Tests

- [ ] Other (provide details)


src/item/GoldenAppleEnchanted.php
```diff
@@ -30,7 +30,7 @@ class GoldenAppleEnchanted extends GoldenApple{
 
 	public function getAdditionalEffects() : array{
 		return [
-			new EffectInstance(VanillaEffects::REGENERATION(), 600, 4),
+			new EffectInstance(VanillaEffects::REGENERATION(), 600, 2),
```
nicholass003: the amplifier value should be 1, according to this
https://github.com/pmmp/PocketMine-MP/blob/stable/src/entity/effect/EffectInstance.php#L100

changelogs: https://feedback.minecraft.net/hc/en-us/articles/29937458432397-Minecraft-Beta-Preview-1-21-40-20

WavyCraftNetwork: I see it already applied 1 level so when a user inputs 0 it’ll automatically be 1.

ShockedPlot7560 (approved):
Looks ok

dktapps (approved):

PR #6450, opened 2024-09-10
EnderPearls now teleports to the correct world

dadodasyra:
## Introduction
EnderPearls currently teleports the player where it landed in the player world when the projectile hit something.
Normal behaviour should be that player is teleported to the coordinates AND the world **of the ender pearl**.

### API changes
None

### Behavioural changes
None except what's written in introduction.

## Backwards compatibility
Should not be a thing.

## Follow-up
This may be completed by a system to avoid teleporting player after it's death.

- [x] Other (This was written in PhpStorm, no syntax issues, CIs runs well and there's no reason that it broke something.)

### DIFF
--- a/src/entity/projectile/EnderPearl.php
+++ b/src/entity/projectile/EnderPearl.php
@@ -27,8 +27,8 @@ use pocketmine\event\entity\EntityDamageEvent;
 use pocketmine\event\entity\ProjectileHitEvent;
 use pocketmine\network\mcpe\protocol\types\entity\EntityIds;
 use pocketmine\world\particle\EndermanTeleportParticle;
-use pocketmine\world\sound\EndermanTeleportSound;
 use pocketmine\world\Position;
+use pocketmine\world\sound\EndermanTeleportSound;
 
 class EnderPearl extends Throwable{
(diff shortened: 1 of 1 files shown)


dktapps:
I think it would make more sense to not teleport the player if they are in a different world. Can't really think of a reason why anyone would want to be teleported back to the world they just left.

dktapps:
Really this ought to be locked behind a teleport counter and is generally a bigger problem than multiworld. See https://github.com/pmmp/PocketMine-MP/issues/3397#issuecomment-624227904

dktapps (changes requested):
New behaviour doesn't make much sense. See comments on https://github.com/pmmp/PocketMine-MP/issues/3397#issuecomment-624227904 and related issues.

dadodasyra:
After some singleplayer vanilla testing it seems that : 
- The ender pearl teleport is cancelled between dimensions
- Death doesn't change anything to the current thrown ender pearls. If the player dies and respawn fast enough he's still teleported.

Will take a look at it when I have the time. For the time I move to a draft

dktapps:
What vanilla does is irrelevant. The expected behaviour should be the behaviour that makes the most sense.

dktapps:
Closing due to inactivity

PR #6456, opened 2024-09-21
Explosion: implemented underwater explosion

nicholass003:
## Introduction

This PR implements underwater explosion for Underwater TNT

https://github.com/user-attachments/assets/c02f2734-b310-4278-9c2e-5a3883bf67ba

## Tests

- [ ] Other (provide details)


dktapps:
A better approach might be to allow the constructor to accept a list of BlockTypeTags (or type IDs) that it skips over, and then check those tags (or IDs) when looking at blocks to destroy. That would make it more flexible and enable custom use cases.

nicholass003:
Is this the correct approach? I have several options for the variable that stores the list of `BlockTypeIds` namely :
- `$excludedBlockTypeIds`
- `$ignoredBlockTypeIds`

However, I think the name `BlockTypeIds` is too long, would you prefer `$excludedBlockIds`?

src/world/Explosion.php
```diff
@@ -125,20 +120,17 @@ public function explodeA() : bool{
 							}
 
 							$state = $subChunk->getBlockStateId($vBlockX & SubChunk::COORD_MASK, $vBlockY & SubChunk::COORD_MASK, $vBlockZ & SubChunk::COORD_MASK);
+							if(in_array($blockFactory->fromStateId($state)->getTypeId(), $this->excludedBlockTypeIds, true)){
```
dktapps: Don't use in_array, it's slow.

nicholass003: Now I use `isset`, from some tests `isset` is the fastest.

nicholass003:
Should `$excludedBlockTypeIds` be public?

dries-c:
Well if people wanted to change the blocks affected by the explosion they can always cancel the EntityPreExplodeEvent and make their own explosion

src/world/Explosion.php
```diff
@@ -118,6 +129,9 @@ public function explodeA() : bool{
 							}
 
 							$state = $subChunk->getBlockStateId($vBlockX & SubChunk::COORD_MASK, $vBlockY & SubChunk::COORD_MASK, $vBlockZ & SubChunk::COORD_MASK);
+							if(isset($this->excludedBlockTypeIds[$blockFactory->fromStateId($state)->getTypeId()])){
```
dktapps: This check can be avoided for the common case by checking if the count of excluded IDs is > 0. Reading stuff from BlockFactory is costly.

src/world/Explosion.php
```diff
@@ -118,6 +130,9 @@ public function explodeA() : bool{
 							}
 
 							$state = $subChunk->getBlockStateId($vBlockX & SubChunk::COORD_MASK, $vBlockY & SubChunk::COORD_MASK, $vBlockZ & SubChunk::COORD_MASK);
+							if(count($this->excludedBlockTypeIds) > 0 && isset($this->excludedBlockTypeIds[$blockFactory->fromStateId($state)->getTypeId()])){
+								continue;
+							}
```
dktapps, line 155: I don't think this is right. It might end up making explosions more powerful in water since the blast force isn't being decreased. I think the previous check by overriding the blast resistance was more correct.

dktapps (changes requested):

src/world/Explosion.php
```diff
@@ -71,6 +81,8 @@ public function __construct(
 			throw new \InvalidArgumentException("Explosion radius must be greater than 0, got $radius");
 		}
 		$this->subChunkExplorer = new SubChunkExplorer($this->world);
+
+		$this->excludedBlockTypeIds = array_flip($excludedBlockTypeIds);
```
dktapps: Prefer `array_fill_keys($keys, true)` instead of doing this, it has less confusing behaviour

src/world/Explosion.php
dktapps: On balance it looks like this was actually the correct solution since it makes water behave the same as air (which has a blast resistance of -1). Sorry for the confusion.

src/world/Explosion.php
```diff
@@ -86,6 +101,14 @@ public function __construct(
 			throw new \InvalidArgumentException("Explosion radius must be greater than 0, got $radius");
 		}
 		$this->subChunkExplorer = new SubChunkExplorer($this->world);
+
+		$this->excludedBlockTypeIds = array_fill_keys($excludedBlockTypeIds, true);
+
+		if($this->what instanceof PrimedTNT || $this->what instanceof TNT){
+			if($this->what->worksUnderwater()){
+				$this->source->y += self::UNDERWATER_TNT_Y_OFFSET;
```
dktapps: Why this change?

nicholass003: I don't really remember why I did that, my bad.

dktapps:
I stalled on merging this because I don't like the solution, but I don't currently have a better one. It's not been at the forefront of my mind lately. Will revisit it soon.

PR #6461, opened 2024-09-26
fix: canPlaceOn and canDestroy will be passed on conversion

### LINKED ISSUE #6460: canPlaceOn, canDestroy have no effect on clients
### Issue description

- Expected result: Using `Item::setCanPlaceOn()`, `Item::setCanDestroy()` should be able to change the item's behavior
- Actual result: These changes are not reflected on the client and nothing happens

This issue is caused by this code:
https://github.com/pmmp/PocketMine-MP/blob/0a7cbdd56df893c1778b72ca0bd452555ccf7c4d/src/network/mcpe/convert/TypeConverter.php#L226

1. Create an item and give it to the player
```php
$item = VanillaItems::IRON_HOE();
$item->setCanDestroy(["minecraft:grass_block"]);
$player->getInventory()->addItem($item);
```

### OS and versions
* PocketMine-MP: 5.19.0

Nerahikada:
## Introduction
`TypeConverter` now also uses `canPlaceOn` and `canDestroy` parameters for item conversions.

### Relevant issues
* Fixes #6460

## Follow-up
This is enough to fix #6460, but still has the following issues:

- Crashes when using new IDs such as `minecraft:grass_block`
- `ItemStack` extra data contains unnecessary NBT data:
  - Minecraft vanilla: `//8BCgAAAwYARGFtYWdlAAAAAAAAAAAAAQAAABUAbWluZWNyYWZ0OmdyYXNzX2Jsb2Nr`
  - pmmp: `//8BCgAACQoAQ2FuRGVzdHJveQgBAAAAFQBtaW5lY3JhZnQ6Z3Jhc3NfYmxvY2sAAAAAAAEAAAAVAG1pbmVjcmFmdDpncmFzc19ibG9jaw==`

## Tests
- [ ] Other (provide details)

```php
$item = VanillaItems::IRON_HOE();
$item->setCanDestroy(["grass"]);  // minecraft:grass_block or grass_block will crash
$player->getInventory()->addItem($item);
```

### DIFF
--- a/src/network/mcpe/convert/TypeConverter.php
+++ b/src/network/mcpe/convert/TypeConverter.php
@@ -222,8 +222,8 @@ class TypeConverter{
 		}
 
 		$extraData = $id === $this->shieldRuntimeId ?
-			new ItemStackExtraDataShield($nbt, canPlaceOn: [], canDestroy: [], blockingTick: 0) :
-			new ItemStackExtraData($nbt, canPlaceOn: [], canDestroy: []);
+			new ItemStackExtraDataShield($nbt, canPlaceOn: $itemStack->getCanPlaceOn(), canDestroy: $itemStack->getCanDestroy(), blockingTick: 0) :
+			new ItemStackExtraData($nbt, canPlaceOn: $itemStack->getCanPlaceOn(), canDestroy: $itemStack->getCanDestroy());
 		$extraDataSerializer = PacketSerializer::encoder();
 		$extraData->write($extraDataSerializer);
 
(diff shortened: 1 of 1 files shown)


dries-c:
Would be nice if we could validate whatever is set in setCanDestroy.

dktapps:
In PM these arrays can accept arbitrary values (by design). Same can't be said for the Minecraft ones, those have to contain actual block IDs. Not sure what the proper solution would be. Perhaps some clunky hack using StringToItemParser and the blockstate serializer to retrieve the ID, but the behaviour still might differ between client & server because Minecraft block IDs don't map 1:1 onto PM block types.

Personally I think this feature is better not bothered with.

dktapps:
Also worth mentioning that PM does enforce these, but only in adventure mode.

dktapps:
Closing this as WF since there doesn't seem to be an obvious way to control this by the item's data beside its ID, making it difficult to use with PM-side items without directly exposing Mojang internals to the API. Considering the limited benefit I don't think this is worth bothering with.

PR #6464, opened 2024-10-03
Add observers

Joshy3282:
## Introduction

Introduces observer block

## Tests

- [ ] Other (provide details)
image


RezzaAbdillah:
Why don't you add a header license?

IvanCraft623:
the whole redstone logic is missing, which is in essence the main function of the Observer, so without a redstone implementation, there is no point in implementing it. In addition, if there is a plugin that intends to implement it properly, it will encounter the difficulty that it will not be able to overwrite the PM implementation.

Thanks for your contrubution, however redstone components cannot be implemented if there is no redstone functionality.

RezzaAbdillah:
Yeah

Pada Rab, 9 Okt 2024 22.49, IvanCraft623 ***@***.***> menulis:

PR #6467, opened 2024-10-07
Fix floating point issue

GameParrot:
## Introduction
Fixes blocks not placing right next to the player at certain positions

### Relevant issues
Currently you cannot place blocks right next to yourself at some positions due to floating point issues

### API changes
No significant API changes

### Behavioural changes
Might affect some extremely precise checks

## Backwards compatibility
No Backwards compatibility issues

## Tests

- [ ] Other (provide details)

### DIFF
--- a/src/world/World.php
+++ b/src/world/World.php
@@ -2328,7 +2328,7 @@ class World implements ChunkManager{
 		for($x = $minX; $x <= $maxX; ++$x){
 			for($z = $minZ; $z <= $maxZ; ++$z){
 				foreach($this->getChunkEntities($x, $z) as $ent){
-					if($ent !== $entity && $ent->boundingBox->intersectsWith($bb)){
+					if($ent !== $entity && $ent->boundingBox->intersectsWith($bb, 0.01)){
 						$nearby[] = $ent;
 					}
 				}


IvanCraft623:
I'm not seeing the issue, please mention reproduction steps

Superomarking:
I believe this is what he's referring to:
https://cdn.discordapp.com/attachments/480650036972224513/1287745560690032761/Minecraft_2024.09.23_-_12.57.42.12.mov?ex=6707c202&is=67067082&hm=1942120fba163c42cab91f092d7cd6a77a3916a25047dcbee388d961017ff0ff&

GameParrot:
> I believe this is what he's referring to: https://cdn.discordapp.com/attachments/480650036972224513/1287745560690032761/Minecraft_2024.09.23_-_12.57.42.12.mov?ex=6707c202&is=67067082&hm=1942120fba163c42cab91f092d7cd6a77a3916a25047dcbee388d961017ff0ff&

Yep, that's the issue


https://github.com/user-attachments/assets/1f8ef0aa-6d9b-4548-9bf1-abab6a1a145a

dktapps:
This function already has an epsilon built in, it shouldn't be needed. Also not clear if this might cause unintended side effects.

dktapps:
> Fixes blocks not placing right next to the player at certain positions

Would be nice to have some examples so this can be tested

joinasau:
I have exactly the same problem

dktapps:
I don't like this change. It'll affect all callers of getNearbyEntities(), which could have all kinds of unintended side effects.

Glitchiness at large coordinates is to be expected, since there's a falloff in the ability of floats to represent high precision fractions when the number is very large. There's only so much we can compensate for that anyway.

It would be nice to know what specific coordinates this occurs at so it can be investigated properly. As I said, an epsilon is already used for intersectsWith() so this issue should've been fixed long ago.

PR #6446, opened 2024-09-08
Mace Implementation

AzliRynz:
## Introduction

Implementation for the 1.21.0 `Mace`
### Relevant issues

N/A

### Details
https://minecraft.wiki/w/Mace

## Changes
Both class' for the Mace item have been created:

* pocketmine\item\Mace.php

One class has been added to handle mace smash ground sound:

* pocketmine/world/sound/MaceSmashGroundSound.php

And one class has been added to handle mace smash air sound:

* pocketmine/world/sound/MaceSmashAirSound.php

The Mace has been implemented as a vanilla item.

### API changes

N/A
### Behavioural changes

N/A
## Backwards compatibility

N/A
## Follow-up
N/A

## Mojang Screenshot

Noor holding a Mace

Noor using the mace against a zombie.

## Video

https://github.com/user-attachments/assets/d8f636a1-8b6e-4ff1-bc0f-b7c7b26a8585

## Tests

I tested this PR by doing the following (tick all that apply):
- [ ] Writing PHPUnit tests (commit these in the `tests/phpunit` folder)
- [x] Playtesting using a Minecraft client (provide screenshots or a video)
- [ ] Writing a test plugin (provide the code and sample output)
- [ ] Other (provide details)

PR #6468, opened 2024-10-20
feat: implements mace item

AzaleeX:
## Introduction

Added a new weapon class, the **Mace**, which functions both as a tool and a weapon with specific properties. This implementation includes changes to the item serialization system and new constants to support the Mace.

https://minecraft.wiki/w/Mace

### Relevant issues

* No specific issues mentioned.

### API changes
Added API for the Mace weapon in several files, including `ItemSerializerDeserializerRegistrar.php` and `ItemTypeIds.php`.

### Behavioural changes
Introduced the behaviour of the **Mace**, which has a 1.5x faster mining rate when used on blocks, and takes damage when attacking entities or breaking blocks.

## Backwards compatibility
* No noted backwards-incompatible changes.

## Screenshot
image
> [!NOTE]
> Don't worry i also implement the compass recovery
image

## Follow-up
Prepare unit tests or in-game playtesting sessions to confirm that the Mace functions as intended.

> [!CAUTION]
What's missing at the moment is item crafting and the main function for calculating damage based on the height of an entity. 

- [x] Writing PHPUnit tests (commit these in the `tests/phpunit` folder)


AzliRynz:
1. Smash particles that increase per distance fallen
2. Damage to the entity increased per distance fallen
3. Knock back to nearby entities to the smashed entity
4. The player falling motion is reset when smashing something
5. And there is a sound from the mace item

src/item/ItemTypeIds.php
```diff
@@ -324,8 +324,9 @@ private function __construct(){
 	public const SPIRE_ARMOR_TRIM_SMITHING_TEMPLATE = 20285;
 	public const PITCHER_POD = 20286;
 	public const NAME_TAG = 20287;
+	public const MACE = 20289;
```
AzliRynz, line 328: Why from 20297 straight to 20289?

src/block/BlockToolType.php
```diff
@@ -40,5 +40,6 @@ private function __construct(){
 	public const AXE = 1 << 3;
 	public const SHEARS = 1 << 4;
 	public const HOE = 1 << 5;
+	public const MACE = 1 << 6;
```
jasonw4331, line 43: The tool type only must be added here for variants like diamond and iron versions of the item.

src/item/ItemTypeIds.php
```diff
@@ -324,8 +324,9 @@ private function __construct(){
...
 
-	public const FIRST_UNUSED_ITEM_ID = 20288;
+	public const FIRST_UNUSED_ITEM_ID = 20290;
```
jasonw4331, line 329: This file should be autogenerated using the composer script included with pocketmine's build tools.

src/item/Mace.php
```diff
@@ -0,0 +1,61 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\item;
+
+use pocketmine\block\Block;
+use pocketmine\block\BlockToolType;
+use pocketmine\entity\Entity;
+
+class Mace extends TieredTool {
```
jasonw4331, line 30: Extending TieredTool implies variants of the tool such as an Iron Mace or Diamond Mace item, which does not exist.

src/item/ToolTier.php
```diff
@@ -47,6 +48,7 @@ enum ToolTier{
 	case IRON;
 	case DIAMOND;
 	case NETHERITE;
+	case MACE;
```
jasonw4331, line 51: It should be obvious MACE is not a tier which can be applied to existing tools in the game.

jasonw4331 (changes requested):
Mace should be implemented as a single tool, not as a tiered tool. The item should be autogenerated where applicible. Please make the requested changes and request a re-review.

src/item/Mace.php
AzliRynz: Yeah, I have ever. In pull request #6446

src/item/ItemTypeIds.php
ClousCloud: Isn't it still in pull request, not merged yet?
#6313 

Relax

dktapps:
Closing due to lack of activity

PR #6471, opened 2024-10-24
Assemble 1.21.40

dries-c:
## Introduction
This PR updates PM5 to 1.21.40

## Backwards compatibility
See https://github.com/pmmp/BedrockProtocol/pull/277


dktapps (changes requested):
Looks like the mushroom_stem changes aren't accounted for

src/network/mcpe/handler/PreSpawnPacketHandler.php
```diff
@@ -98,7 +98,7 @@ public function setUp() : void{
 				$this->server->getMotd(),
 				"",
 				false,
-				new PlayerMovementSettings(PlayerMovementType::SERVER_AUTHORITATIVE_V1, 0, false),
+				new PlayerMovementSettings(ServerAuthMovementMode::SERVER_AUTHORITATIVE_V2, 0, false),
```
dktapps, line 101: Why is this changed to v2?

cooldogedev: It is correct, the old [one](https://github.com/pmmp/BedrockProtocol/blob/9825e3abdd43f13f54aa359cce5020c5e303541c/src/types/PlayerMovementType.php#L20)'s value was `1` and so is the [new](https://github.com/pmmp/BedrockProtocol/blob/6aa7cbeb4a7ec6fa58f9024aeaddad7c5c65a459/src/types/ServerAuthMovementMode.php#L21)

dktapps: Thanks

DevPapo (commented):
Ok

joinasau:
...
merge?

ShockedPlot7560 (dismissed):

AzaleeX:
> ... merge?

We need a second approval before we can merge, this is a security feature.

smell-of-curry:
looks good?

AzaleeX:
> looks good?

Yes, I think it is, but there is no one to do the review, so it takes time, I think.

dktapps:
There are still unaddressed review comments

AzaleeX:
> There are still unaddressed review comments

What else does he have to change?

dktapps (approved):

PR #6472, opened 2024-10-25
Add caching mechanism to avoid unnecessary repeated reviews in GitHub

AzaleeX:


ShockedPlot7560:
This won't work, the problem is that every time there is a change to the PR, the approval is dismissed, requiring re-approval for merging and avoid 2 team members to approve when it's a team PR.

Your caching system will approve once, but even if there are changes, it will not be re-approved, making a merge impossible because it will require 2 approvals.

PR #6473, opened 2024-10-25
Add dynamic build date using time() function in VersionInfo

AzaleeX:
This update introduces a dynamic build date in the `VersionInfo` class using the `time()` function, allowing the build date to be set at runtime. This change ensures that the build date is always current without needing manual updates.

- remove from Todos list


jasonw4331:
The server isn't built during runtime. It's built during build time. Please ensure the build time is accurate to a time generated during the build step rather than when the server is ran.

src/VersionInfo.php
```diff
@@ -103,4 +104,8 @@ public static function VERSION() : VersionString{
 		}
 		return self::$fullVersion;
 	}
+
+	public static function BUILD_DATE() : int {
+		return time();
```
dktapps: It's supposed to be the time the phar was built, not the time the function was called.
This isn't as simple as just adding an extra function.

dktapps (changes requested):

src/VersionInfo.php
```diff
@@ -103,4 +104,21 @@ public static function VERSION() : VersionString{
...
+	//I don't know if it would really work
+	public static function BUILD_DATE() : int {
+		static $buildDate = null;
+		if ($buildDate === null) {
+			$buildDate = time();
+			if (\Phar::running(true) !== "") {
+				$pharPath = \Phar::running(false);
+				$phar = \Phar::isValidPharFilename($pharPath) ? new \Phar($pharPath) : new \PharData($pharPath);
+				$meta = $phar->getMetadata();
+				if (isset($meta["date"])) {
```
dktapps, line 117: This isn't going to magically appear in the phar metadata

dktapps:
Closing due to lack of activity on review comments and little need for this change anyway

PR #6391, opened 2024-07-09
Implement all new 1.21 tuff blocks

IvanCraft623:
## Introduction

Implement the new 1.21 tuff blocks types!!!

#### Blocks
- Added the following new blocks:
  - Tuff Slab
  - Tuff Stairs
  - Tuff Wall
  - Chiseled Tuff
  - Tuff Bricks
  - Tuff Brick Slab
  - Tuff Brick Stairs
  - Tuff Brick Wall
  - Chiseled Tuff Bricks
  - Polished Tuff
  - Polished Tuff Slab
  - Polished Tuff Stairs
  - Polished Tuff Wall

## Tests
I tested this PR by doing the following (tick all that apply):
- [x] Playtesting using a Minecraft client (provide screenshots or a video)
   Tuff blocks

PR #6474, opened 2024-10-25
Add Bamboo Block Mappings

AzaleeX:
## Introduction
Added mappings for Bamboo blocks, including Bamboo Door and Bamboo Sign, to the ItemSerializerDeserializerRegistrar.

> [!WARNING]
> I'm not sure it's legal to put IDSs directly in functions such as getLogIdentifier() or getAllSidedLogIdentifier() because banboo has nothing similar and has no log.

## Tests

- [ ] Other (provide details)


src/block/WoodLikeBlockIdHelper.php
```diff
@@ -88,9 +90,13 @@ public static function getSlabIdentifier(WoodType $type) : BID{
 			WoodType::CRIMSON => Ids::CRIMSON_SLAB,
 			WoodType::WARPED => Ids::WARPED_SLAB,
 			WoodType::CHERRY => Ids::CHERRY_SLAB,
+			WoodType::BAMBOO => Ids::BAMBOO_SLAB,
 		});
 	}
 
+	/**
+	 * @throws \Exception
+	 */
```
dktapps: Unrelated change

src/block/WoodLikeBlockIdHelper.php
```diff
@@ -103,6 +109,7 @@ public static function getLogIdentifier(WoodType $treeType) : BID{
 			WoodType::CRIMSON => Ids::CRIMSON_STEM,
 			WoodType::WARPED => Ids::WARPED_STEM,
 			WoodType::CHERRY => Ids::CHERRY_LOG,
+			WoodType::BAMBOO => 59000,
```
dktapps: No

src/block/WoodLikeBlockIdHelper.php
```diff
@@ -118,6 +125,7 @@ public static function getAllSidedLogIdentifier(WoodType $treeType) : BID{
 			WoodType::CRIMSON => Ids::CRIMSON_HYPHAE,
 			WoodType::WARPED => Ids::WARPED_HYPHAE,
 			WoodType::CHERRY => Ids::CHERRY_WOOD,
+			WoodType::BAMBOO => 59001,
```
dktapps: Also no

src/data/bedrock/item/ItemSerializerDeserializerRegistrar.php
```diff
@@ -136,6 +136,7 @@ private function register1to1BlockMappings() : void{
 		$this->map1to1Block(Ids::CAULDRON, Blocks::CAULDRON());
 		$this->map1to1Block(Ids::CHAIN, Blocks::CHAIN());
 		$this->map1to1Block(Ids::CHERRY_DOOR, Blocks::CHERRY_DOOR());
+		$this->map1to1Block(Ids::BAMBOO_DOOR, Blocks::BAMBOO_DOOR());
```
dktapps: This list is supposed to be in alphabetical order

src/data/bedrock/item/ItemSerializerDeserializerRegistrar.php
```diff
@@ -187,6 +188,7 @@ private function register1to1ItemMappings() : void{
 		$this->map1to1Item(Ids::CHAINMAIL_LEGGINGS, Items::CHAINMAIL_LEGGINGS());
 		$this->map1to1Item(Ids::CHARCOAL, Items::CHARCOAL());
 		$this->map1to1Item(Ids::CHERRY_SIGN, Items::CHERRY_SIGN());
+		$this->map1to1Item(Ids::BAMBOO_SIGN, Items::BAMBOO_SIGN());
```
dktapps: Same again here

src/item/StringToItemParser.php
```diff
@@ -147,6 +147,16 @@ private static function registerBlocks(self $result) : void{
 		$result->registerBlock("azure_bluet", fn() => Blocks::AZURE_BLUET());
 		$result->registerBlock("bamboo", fn() => Blocks::BAMBOO());
 		$result->registerBlock("bamboo_sapling", fn() => Blocks::BAMBOO_SAPLING());
```
dktapps: Alphabetical order

src/item/VanillaItems.php
```diff
@@ -387,6 +388,7 @@ protected static function setup() : void{
 		self::register("carrot", new Carrot(new IID(Ids::CARROT), "Carrot"));
 		self::register("charcoal", new Coal(new IID(Ids::CHARCOAL), "Charcoal"));
 		self::register("cherry_sign", new ItemBlockWallOrFloor(new IID(Ids::CHERRY_SIGN), Blocks::CHERRY_SIGN(), Blocks::CHERRY_WALL_SIGN()));
+		self::register("bamboo_sign", new ItemBlockWallOrFloor(new IID(Ids::BAMBOO_SIGN), Blocks::BAMBOO_SIGN(), Blocks::BAMBOO_WALL_SIGN()));
```
dktapps: Alphabetical order

src/item/VanillaItems.php
```diff
@@ -331,6 +331,7 @@
  * @method static WritableBook WRITABLE_BOOK()
  * @method static WrittenBook WRITTEN_BOOK()
  * @method static SpawnEgg ZOMBIE_SPAWN_EGG()
+ * @method static ItemBlockWallOrFloor BAMBOO_SIGN()
```
dktapps: This docblock is supposed to be generated by a script, not hand-edited

dktapps (changes requested):
I think the reason we didn't already add these was because of the issue around logs, but this is not the right way to fix it.

src/block/WoodLikeBlockIdHelper.php
AzaleeX: What should I do?

src/block/WoodLikeBlockIdHelper.php
AzaleeX: What should I do?

src/block/WoodLikeBlockIdHelper.php
nicholass003: You have to update or create new ids here
https://github.com/pmmp/PocketMine-MP/blob/stable/src/block/BlockTypeIds.php

nicholass003: Example :
https://github.com/pmmp/PocketMine-MP/pull/6391/files#diff-b7d3ed653a6892f8d17adae002eb6840e0d6ab5fba40e9a7afce040bd3ea5094

src/block/WoodLikeBlockIdHelper.php
ClousCloud: You have to add the id to the src/block/BlockTypeIds.php file

src/item/VanillaItems.php
AzliRynz: Yeah, he should run the `composer update-codegen` command

src/block/WoodLikeBlockIdHelper.php
dktapps: There is no bamboo log or "wood", but this isn't the way to fix it

AzaleeX: So we're just going to create an id with an existing block?

ShockedPlot7560:
Same problem as #5875 we can't introduce bamboo as a wood type

PR #6483, opened 2024-11-05
TextFormat: Add colors and harmonize class

dadodasyra:
## Introduction
The new bedrock colors in chat are not usable on server side (console and command/chat parsing) due to TextFormat not being up to date.
The class was mixing self:: and TextFormat:: use reference the class itself. I opted to harmonize it to self::

### Relevant issues
None

### API changes
None

### Behavioural changes
The colors will now be parsed and be shown correctly server side.

## Backwards compatibility
None

The names and specific colors where extract from the "semi-official" wiki https://minecraft.wiki/w/Formatting_codes

### DIFF
--- a/src/utils/TextFormat.php
+++ b/src/utils/TextFormat.php
@@ -46,23 +46,34 @@ abstract class TextFormat{
 	public const ESCAPE = "\xc2\xa7"; //§
 	public const EOL = "\n";
 
-	public const BLACK = TextFormat::ESCAPE . "0";
-	public const DARK_BLUE = TextFormat::ESCAPE . "1";
-	public const DARK_GREEN = TextFormat::ESCAPE . "2";
-	public const DARK_AQUA = TextFormat::ESCAPE . "3";
-	public const DARK_RED = TextFormat::ESCAPE . "4";
-	public const DARK_PURPLE = TextFormat::ESCAPE . "5";
-	public const GOLD = TextFormat::ESCAPE . "6";
-	public const GRAY = TextFormat::ESCAPE . "7";
-	public const DARK_GRAY = TextFormat::ESCAPE . "8";
-	public const BLUE = TextFormat::ESCAPE . "9";
-	public const GREEN = TextFormat::ESCAPE . "a";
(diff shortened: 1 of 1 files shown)


jasonw4331:
Sadly I believe this has the same blocker as #5838

ShockedPlot7560:
The unique blocked is the BC on removing underline and strikethrough

dadodasyra:
> Sadly I believe this has the same blocker as #5838

Yup my bad I tought this was a more recent thing (the color update) so I didn't checked PR that's old.



> The unique blocked is the BC on removing underline and strikethrough

So what's next ? Basically his PR is more detailed than mine (and does not forget the whole terminal stuff and regexes)..

ShockedPlot7560:
First, when you need to add an and word in your PR it mean that you need to create multiple PRs.

But for now, if we want to include the new colour, feel free to add only the colours and not to remove the underline and strikethrough, which are BC (not until PM6). But each time the PRs do everything at the same time, which blocks the merge.

dktapps:
> First, when you need to add an add word in your PR it mean that you need to create multiple PRs.

You mean an "and" word?

dktapps:
> But for now, if we want to include the new colour, feel free to add only the colours and not to remove the underline and strikethrough, which are BC (not until PM6).

This still creates problems for terminal usage, since people using the new colour constants will get unexpected text effects on the terminal. There's a reason no one added these already. It's BC-breaking no matter what change is made.

dadodasyra:
I mean, is it really an issue that the old underline and strikethrough change to a color ? No one is using it anymore since it's not usable in game.. The BC break seems "acceptable" to me

dktapps:
No BC breaks on minor or patch versions. Those are the rules.

dadodasyra:
Can we even qualify a color change on a console (so an administrator tool) a BC break ?

AzaleeX:
Alternatively, why not block colors directly in the terminal by resetting the color with `§r`

dktapps:
Closing this as #5838 is more complete and doesn't contain unrelated changes.

PR #6485, opened 2024-11-06
Update WallSign.php

TrinitroToluen0:
## Introduction
I, as a plugin developer needed to use this function to get the block from where the WallSign was standing, to make a subclaim system, for what i've seen i don't think this method has any reason to be a protected method.

### API changes
- `WallSign::getSupportingFace()` is now a public method

### Behavioural changes
None

## Backwards compatibility
None

## Follow-up
- Review the other blocks that have a protected `getSupportingFace()` method.

### DIFF
--- a/src/block/WallSign.php
+++ b/src/block/WallSign.php
@@ -34,7 +34,7 @@ use pocketmine\world\BlockTransaction;
 final class WallSign extends BaseSign{
 	use HorizontalFacingTrait;
 
-	protected function getSupportingFace() : int{
+	public function getSupportingFace() : int{
 		return Facing::opposite($this->facing);
 	}
 


dktapps:
I don't see a need for this change. You can just do `Facing::opposite($block->getFacing())`.

TrinitroToluen0:
Yes that is what i did, but if we can improve the API by abstracting it more, why not do it? Anyway this PR is incomplete since we would need to do this with every protected `getSupportingFace()` method to preserve consistency.

dktapps:
This function is intended for internal usage by BaseSign, not by plugins.

dktapps:
I'd intended to introduce something like a `DynamicSupportTrait` for blocks that behave like this, but never got around to it. It's outside the scope of this PR though.

Making this function public increases the API surface, and we wouldn't be able to get rid of it if someone wanted to make such functionality consistent across the code. I'll leave it to other reviewers to decide, but I'm personally against it.

ipad54:
I personally don’t like this change. There are way more places than just Sign (for example `growthFace` in `NetherVines` doesn’t have getter at all). This function exists only because it is needed for internal usage. As Dylan said above, plugin developers can use `Facing::opposite($block->getFacing())` to get the supporting face.

By the way, the problem with block faces and properties itself is much wider. Right now plugin developers still need to do `instanceof` check of exact block type, so implementing #6161 instead will be more efficient.

PR #6486, opened 2024-11-06
Adjust Sugar Cane to break when there is no water

ghost:
## Introduction
This PR allows Sugar Cane to break when it cannot find water.

https://github.com/user-attachments/assets/f1dec797-1eaf-4179-a057-0be68490c837
## Tests

### DIFF
--- a/src/block/Sugarcane.php
+++ b/src/block/Sugarcane.php
@@ -99,7 +99,7 @@ class Sugarcane extends Flowable{
 	public function onRandomTick() : void{
 		$down = $this->getSide(Facing::DOWN);
 		if(!$down->hasSameTypeId($this)){
-			if(!$this->hasNearbyWater($down)) {
+			if(!$this->hasNearbyWater($down)){
 				$this->position->getWorld()->useBreakOn($this->position, createParticles: true);
 				return;
 			}
@@ -133,7 +133,7 @@ class Sugarcane extends Flowable{
 	private function hasNearbyWater(Block $down) : bool{
 		foreach($down->getHorizontalSides() as $sideBlock){
 			$blockId = $sideBlock->getTypeId();
-			if($blockId === BlockTypeIds::WATER || $blockId === BlockTypeIds::FROSTED_ICE) {
+			if($blockId === BlockTypeIds::WATER || $blockId === BlockTypeIds::FROSTED_ICE){
 				return true;
 			}
 		}
--- a/src/block/Sugarcane.php
+++ b/src/block/Sugarcane.php
@@ -33,6 +33,8 @@ use pocketmine\math\Vector3;
 use pocketmine\player\Player;
 use pocketmine\world\BlockTransaction;
 use pocketmine\world\Position;
+use pocketmine\world\particle\BlockBreakParticle;
+use pocketmine\world\sound\BlockBreakSound;
 
 class Sugarcane extends Flowable{
 	use AgeableTrait;
@@ -98,6 +100,15 @@ class Sugarcane extends Flowable{
 
 	public function onRandomTick() : void{
 		if(!$this->getSide(Facing::DOWN)->hasSameTypeId($this)){
+			if (!$this->isSupportedByWater()) {
+				$world = $this->position->getWorld();
+			
+				$world->addParticle($this->position, new BlockBreakParticle($this));
+				$world->addSound($this->position, new BlockBreakSound($this));
+			
+				$world->useBreakOn($this->position);
+				return;
+			}
 			if($this->age === self::MAX_AGE){
 				$this->grow($this->position);
 			}else{
@@ -123,4 +134,15 @@ class Sugarcane extends Flowable{
 
 		return false;
 	}
+
+	private function isSupportedByWater() : bool{
+		 $down = $this->getSide(Facing::DOWN);
+		 foreach(Facing::HORIZONTAL as $side){
+			 $sideBlock = $down->getSide($side);
+			 if($sideBlock instanceof Water || $sideBlock instanceof FrostedIce){
+				 return true;
+			 }
+		 }
+		 return false;
+	 }
 }
--- a/src/block/Sugarcane.php
+++ b/src/block/Sugarcane.php
@@ -144,7 +144,7 @@ class Sugarcane extends Flowable{
 		$down = $this->getSide(Facing::DOWN);
 		if(!$this->canBeSupportedAt($this)){
 			$this->position->getWorld()->useBreakOn($this->position, createParticles: true);
-		}elseif (!$down->hasSameTypeId($this) && !$this->hasNearbyWater($down)) {
+		}elseif(!$down->hasSameTypeId($this) && !$this->hasNearbyWater($down)){
 			$this->position->getWorld()->useBreakOn($this->position, createParticles: true);
 		}
 	}
--- a/src/block/Sugarcane.php
+++ b/src/block/Sugarcane.php
@@ -36,9 +36,9 @@ use pocketmine\world\Position;
 
 class Sugarcane extends Flowable{
 	use AgeableTrait;
-	use StaticSupportTrait {
-        onNearbyBlockChange as onNearbyBlockChangeTrait;
-    }
+	use StaticSupportTrait{
+		onNearbyBlockChange as onNearbyBlockChangeTrait;
+	}
 
 	public const MAX_AGE = 15;
 
@@ -147,6 +147,6 @@ class Sugarcane extends Flowable{
 		if(!$down->hasSameTypeId($this) && !$this->hasNearbyWater($down)){
 			$this->position->getWorld()->useBreakOn($this->position, createParticles: true);
 		}
-        $this->onNearbyBlockChangeTrait();
+		$this->onNearbyBlockChangeTrait();
 	}
 }
--- a/src/block/Sugarcane.php
+++ b/src/block/Sugarcane.php
@@ -142,4 +142,11 @@ class Sugarcane extends Flowable{
 		 }
 		 return false;
 	 }
+
+    public function onNearbyBlockChange() : void {
+        if(!$this->isSupportedByWater()) {
+            $world = $this->position->getWorld();
+			$world->useBreakOn($this->position, createParticles: true);
+        }
+    }
 }
--- a/src/block/Sugarcane.php
+++ b/src/block/Sugarcane.php
@@ -142,11 +142,11 @@ class Sugarcane extends Flowable{
 		}
 		return false;
 	}
-	
-	public function onNearbyBlockChange() : void {
-        if(!$this->isSupportedByWater()) {
-            $world = $this->position->getWorld();
-			$world->useBreakOn($this->position, createParticles: true);
+
+	public function onNearbyBlockChange() : void{
+		if (!$this->isSupportedByWater()) {
+        	$world = $this->position->getWorld();
+            $world->useBreakOn($this->position, createParticles: true);
         }
-    }
+	}
 }
--- a/src/block/Sugarcane.php
+++ b/src/block/Sugarcane.php
@@ -142,7 +142,7 @@ class Sugarcane extends Flowable{
 
 	public function onNearbyBlockChange() : void {
 		$down = $this->getSide(Facing::DOWN);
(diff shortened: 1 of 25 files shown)


src/block/Sugarcane.php
```diff
@@ -123,4 +133,15 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
 
 		return false;
 	}
+
+	private function isSupportedByWater() : bool{
+		 $down = $this->getSide(Facing::DOWN);
+		 foreach(Facing::HORIZONTAL as $side){
+			 $sideBlock = $down->getSide($side);
+			 if($sideBlock instanceof Water || $sideBlock instanceof FrostedIce){
```
dktapps: Really not supposed to use instanceof for this

dktapps (commented):
Is this vanilla behaviour?

ghost:
> vanilla behaviour?

yes it is true vanilla behavior :D

src/block/Sugarcane.php
ghost: then how can i do it .-.

jasonw4331: check the id value

ghost: Okay.

src/block/Sugarcane.php
```diff
@@ -98,6 +100,14 @@ public function ticksRandomly() : bool{
 
 	public function onRandomTick() : void{
 		if(!$this->getSide(Facing::DOWN)->hasSameTypeId($this)){
+			if(!$this->isSupportedByWater()) {
+				$world = $this->position->getWorld();
+				$world->addParticle($this->position, new BlockBreakParticle($this));
+				$world->addSound($this->position, new BlockBreakSound($this));
+				$world->useBreakOn($this->position);
```
kostamax27: maybe it will be better

```php
$world->useBreakOn($this->position, createParticles: true);
```

ghost: yes this is better :D

src/block/Sugarcane.php
```diff
@@ -98,6 +98,12 @@ public function ticksRandomly() : bool{
 
 	public function onRandomTick() : void{
 		if(!$this->getSide(Facing::DOWN)->hasSameTypeId($this)){
+			if(!$this->isSupportedByWater()) {
```
kostamax27: This should happen not only during onRandomTick but also if something nearby is updated, for example, when a block is placed nearby.

https://github.com/user-attachments/assets/d6becf94-1c18-46d3-aae2-ecfc2c13fd4a

ghost: I added onNearbyBlockChange for this thanks 👍

src/block/Sugarcane.php
```diff
@@ -123,4 +129,24 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
 
 		return false;
 	}
+
+	private function isSupportedByWater() : bool{
+		$down = $this->getSide(Facing::DOWN);
+		foreach(Facing::HORIZONTAL as $side){
```
kostamax27: https://github.com/pmmp/PocketMine-MP/blob/b3b8aaddffe3b3a1eee8dcc62dd08fcd10581916/src/block/Block.php#L788

ghost: okay let's try this

ghost (commented):
._.

src/block/Sugarcane.php
```diff
@@ -98,6 +98,12 @@ public function ticksRandomly() : bool{
 
 	public function onRandomTick() : void{
 		if(!$this->getSide(Facing::DOWN)->hasSameTypeId($this)){
+			if(!$this->isSupportedByWater()) {
+				$world = $this->position->getWorld();
+				$world->useBreakOn($this->position, createParticles: true);
```
ipad54: There is no need to create an additional `$world` variable. You can inline this.

src/block/Sugarcane.php
```diff
@@ -123,4 +129,28 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
 
 		return false;
 	}
+
+	private function isSupportedByWater() : bool{
+		$down = $this->getSide(Facing::DOWN);
+		foreach($down->getHorizontalSides() as $sideBlock){
+			$blockId = $sideBlock->getTypeId();
+
```
ipad54:
```suggestion
```

src/block/Sugarcane.php
```diff
@@ -123,4 +129,28 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
...
+			if ($blockId === BlockTypeIds::WATER || $blockId === BlockTypeIds::FROSTED_ICE) {
+				return true;
+			}
+		}
+		return false;
+	}
+
+	public function onNearbyBlockChange() : void {
+		$bottom = $this->seekToBottom();
+		if (!$this->position->equals($bottom)) {
```
ipad54: It would be better if you combine these 2 conditions into one (e.g `if($this->position->equals($bottom) && !$this->isSupportedByWater()){}`)

src/block/Sugarcane.php
```diff
@@ -123,4 +129,28 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
...
+			return;
+		}
+
+		if (!$this->isSupportedByWater()) {
+			$world = $this->position->getWorld();
```
ipad54: Same here. There is no need to create additional `$world` variable.

src/block/Sugarcane.php
ipad54: By the way, you can just check that the down side of the block doesn't have same type id instead of seeking to the bottom (like it is done in `onRandomTick()`). It would be a little bit faster in large crop farms.

ipad54 (changes requested):

src/block/Sugarcane.php
```diff
@@ -123,4 +128,22 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
 
 		return false;
 	}
+
+	private function isSupportedByWater() : bool{
```
dktapps: `hasNearbyWater` would be a better name for this

src/block/Sugarcane.php
```diff
@@ -123,4 +128,22 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
...
+		$down = $this->getSide(Facing::DOWN);
```
dktapps: `getSide(down)` is getting called outside of where this is used in both usages. Perhaps it might be worth passing it as a parameter instead of calling it a second time.

src/block/Sugarcane.php
```diff
@@ -123,4 +128,22 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
...
+		foreach($down->getHorizontalSides() as $sideBlock){
+			$blockId = $sideBlock->getTypeId();
+			if ($blockId === BlockTypeIds::WATER || $blockId === BlockTypeIds::FROSTED_ICE) {
```
dktapps:
```suggestion
			if($blockId === BlockTypeIds::WATER || $blockId === BlockTypeIds::FROSTED_ICE){
```

src/block/Sugarcane.php
```diff
@@ -123,4 +128,22 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
...
+				return true;
+			}
+		}
+		return false;
+	}
+
+	public function onNearbyBlockChange() : void {
+		$downBlock = $this->getSide(Facing::DOWN);
+		if (!$downBlock->hasSameTypeId($this) && !$this->isSupportedByWater()) {
```
dktapps:
```suggestion
		if(!$downBlock->hasSameTypeId($this) && !$this->isSupportedByWater()){
```

src/block/Sugarcane.php
```diff
@@ -98,6 +98,11 @@ public function ticksRandomly() : bool{
 
 	public function onRandomTick() : void{
 		if(!$this->getSide(Facing::DOWN)->hasSameTypeId($this)){
+			if(!$this->isSupportedByWater()) {
```
dktapps:
```suggestion
			if(!$this->isSupportedByWater()){
```

dktapps (changes requested):

dktapps (dismissed):

kostamax27 (commented):
looks good

src/block/Sugarcane.php
```diff
@@ -123,4 +129,21 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
...
+		foreach($down->getHorizontalSides() as $sideBlock){
+			$blockId = $sideBlock->getTypeId();
+			if($blockId === BlockTypeIds::WATER || $blockId === BlockTypeIds::FROSTED_ICE){
+				return true;
+			}
+		}
+		return false;
+	}
+
+	public function onNearbyBlockChange() : void {
```
ipad54: Sugarcane uses `StaticSupportTrait`, so overriding `onNearbyBlockChange()` without calling trait`s method will make sugarcane to stay after supporting block removed.

Screenshot_20241113_111810

ipad54 (changes requested):

src/block/Sugarcane.php
ipad54:
```suggestion
	public function onNearbyBlockChange() : void{
```

Also formatting issue.

ghost: I think I found it

ipad54: You can import trait's method as alias and call it after your code. Something like [here](https://github.com/pmmp/PocketMine-MP/blob/09bf203267aca9b83a09640e17bfac0fc7b58ff8/src/block/PinkPetals.php#L40) but with `onNearbyBlockChange()`

dktapps: Good catch @ipad54, I didn't spot that

src/block/Sugarcane.php
```diff
@@ -36,7 +36,9 @@
 
 class Sugarcane extends Flowable{
 	use AgeableTrait;
-	use StaticSupportTrait;
+	use StaticSupportTrait{
+		onNearbyBlockChange as onNearbyBlockChangeTrait;
```
ipad54: Better to name it `onSupportBlockChange`

src/block/Sugarcane.php
```diff
@@ -36,7 +36,9 @@
 
 class Sugarcane extends Flowable{
 	use AgeableTrait;
-	use StaticSupportTrait;
+	use StaticSupportTrait{
```
ipad54:
```suggestion
	use StaticSupportTrait {
```

src/block/Sugarcane.php
```diff
@@ -123,4 +131,22 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
...
+			}
+		}
+		return false;
+	}
+
+	public function onNearbyBlockChange() : void{
+		$down = $this->getSide(Facing::DOWN);
+		if(!$down->hasSameTypeId($this) && !$this->hasNearbyWater($down)){
+			$this->position->getWorld()->useBreakOn($this->position, createParticles: true);
+		}
```
ipad54, line 151: Please add `else` here - we need to make sure the block won't be destroyed twice.

ipad54 (changes requested):

src/block/Sugarcane.php
ghost:
```
if(!$down->hasSameTypeId($this) && !$this->hasNearbyWater($down)) {
            $this->position->getWorld()->useBreakOn($this->position, createParticles: true);
        } else {
			//Won't be destroyed twice.
	}
}
```
Is this exactly how it should be? 🤕

ipad54:
```php
 if(!$down->hasSameTypeId($this) && !$this->hasNearbyWater($down)){
            $this->position->getWorld()->useBreakOn($this->position, createParticles: true);
        }else{
			$this->onSupportBlockChange();
	}
}```

ipad54 (approved):

ShockedPlot7560 (approved):

PR #6497, opened 2024-11-12
start.sh: print warnings on unusual exit codes from the server process

dktapps:
## Introduction
This has been sitting idle on a branch for a couple of years. Not sure why it hasn't been merged as it looks basically fine as far as I can tell.

### Behavioural changes
`start.sh` will now print warnings if the server experiences any of the following errors: SIGSEGV, SIGABRT, SIGKILL, any other anomalous exit code

## Tests
Not tested, just found this while cleaning out old branches and thought it might be worth formally proposing

- [ ] Other (provide details)


start.sh
```diff
@@ -44,6 +44,27 @@ fi
...
 LOOPS=0
 
+handle_exit_code() {
+	local exitcode=$1
+	if [ "$exitcode" -eq 134 ] || [ "$exitcode" -eq 139 ]; then #SIGABRT/SIGSEGV
+		echo ""
+		echo "ERROR: The server process was killed due to a critical error (code $exitcode) which could indicate a problem with PHP."
+		echo "Updating your PHP binary is recommended."
+		echo "If this keeps happening, please open a bug report."
+		echo ""
```
ShockedPlot7560, line 54: Segmentation fault which is due to plugin error could lead to a lot of bug report, no ?

dktapps: PHP code shouldn't be causing segfaults

ipad54: Segfault can be caused in PHP code only via third-party PHP extensions. There is an open issue with `ext-pmmpthread`, but plugin devs are not supposed to use it that way.

ipad54 (approved):

start.sh
dktapps: It's still an issue with the PHP side no matter which way you slice it.

ShockedPlot7560: > PHP code shouldn't be causing segfaults

If you make some tricky code in a thread causing segfault

dktapps: PHP code should not be able to cause a segfault under any conditions. If it can, that's a fault of PHP or an extension.

dktapps:
Sorry for the spam, just testing github actions

Issue #211, opened 2016-12-29
Block->getBreakTime() returns incorrect values for wrong tool tiers, Block->canBeBrokenWith() not implemented correctly


dktapps:
Try to break stone with your bare hand: because Block->canBeBrokenWith() isn't correctly implemented, it calculates time to break the block as if it could be broken by hand. Same thing applies when using wrong tools on blocks.

cancrackmurder:
I would create a pull request, with the correct hardness and resistance values for currently existing blocks. (so that you can focus on other stuff instead of checking the wiki for every block)

And thx for banning my second account btw, i simply were trying to help, no idea why you don't want help on a community project.
Ban this one and i come back with a new one. Have fun.

SOF3:
Looking forward to actual help as opposed to claims to create pull requests. Thank you.

dktapps:
This is important for block-breaking animations to work correctly in 1.1.

zKoz210:
@dktapps You can add it to master?

PR #6500, opened 2024-11-13
Block break progress improvement

GameParrot:
Fixes block break progress being inaccurate when not on ground or swimming. Also fixes syncing block break speed with client.

### Related issues & PRs
* Related to #211 

### API changes
Added Player::isActuallyOnGround(), more accurate than Player::isOnGround() but slower.

### Behavioural changes
Block breaking progress is now much closer to vanilla.

## In-Game Testing
Before:  
https://github.com/user-attachments/assets/22f50b6e-f994-4a17-8dfc-53df70a8d6eb  
After:  
https://github.com/user-attachments/assets/3341eaaf-2bbc-4ac1-ad35-39947e907080

### DIFF
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -865,6 +865,29 @@ class Player extends Human implements CommandSender, ChunkListener, IPlayer{
 		Timings::$playerChunkSend->stopTiming();
 	}
 
+	/**
+	 * Checks if the player is currently on the ground. This is more accurate than {@link Player::isOnGround()} but slower.
+	 */
+	public function isActuallyOnGround() : bool{
+		$bb = $this->boundingBox->expandedCopy(0, 0.001, 0);
+		$maxY = (int) floor($this->location->y);
+		$minY = $maxY - 1;
+		$floorMinX = (int) floor($bb->minX);
+		$floorMinZ = (int) floor($bb->minZ);
+		$floorMaxX = (int) floor($bb->maxX);
+		$floorMaxZ = (int) floor($bb->maxZ);
+		for ($x = $floorMinX ; $x <= $floorMaxX ; $x++) {
+			for ($y = $minY ; $y <= $maxY ; $y++) {
+				for ($z = $floorMinZ ; $z <= $floorMaxZ ; $z++) {
+					if ($this->getWorld()->getBlockAt($x, $y, $z)->collidesWithBB($bb)) {
+						return true;
+					}
+				}
+			}
+		}
+		return false;
+	}
+
 	private function recheckBroadcastPermissions() : void{
 		foreach([
 			DefaultPermissionNames::BROADCAST_ADMIN => Server::BROADCAST_CHANNEL_ADMINISTRATIVE,
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -869,7 +869,7 @@ class Player extends Human implements CommandSender, ChunkListener, IPlayer{
 	 * Checks if the player is currently on the ground. This is more accurate than {@link Player::isOnGround()} but slower.
 	 */
 	public function isActuallyOnGround() : bool{
-		$bb = $this->boundingBox->expandedCopy(0, 0.001, 0);
+		$bb = $this->boundingBox->expandedCopy(-0.001, 0.001, -0.001);
 		$maxY = (int) floor($this->location->y);
 		$minY = $maxY - 1;
 		$floorMinX = (int) floor($bb->minX);
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -876,10 +876,10 @@ class Player extends Human implements CommandSender, ChunkListener, IPlayer{
 		$floorMinZ = (int) floor($bb->minZ);
 		$floorMaxX = (int) floor($bb->maxX);
 		$floorMaxZ = (int) floor($bb->maxZ);
-		for ($x = $floorMinX ; $x <= $floorMaxX ; $x++) {
-			for ($y = $minY ; $y <= $maxY ; $y++) {
-				for ($z = $floorMinZ ; $z <= $floorMaxZ ; $z++) {
-					if ($this->getWorld()->getBlockAt($x, $y, $z)->collidesWithBB($bb)) {
+		for ($x = $floorMinX ; $x <= $floorMaxX ; $x++){
+			for ($y = $minY ; $y <= $maxY ; $y++){
+				for ($z = $floorMinZ ; $z <= $floorMaxZ ; $z++){
+					if ($this->getWorld()->getBlockAt($x, $y, $z)->collidesWithBB($bb)){
 						return true;
 					}
 				}
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -1787,7 +1787,7 @@ class Player extends Human implements CommandSender, ChunkListener, IPlayer{
 			return true;
 		}
 
-		if(!$this->isCreative() && !$block->getBreakInfo()->breaksInstantly()){
+		if(!$this->isCreative() && !$target->getBreakInfo()->breaksInstantly()){
 			$this->blockBreakHandler = new SurvivalBlockBreakHandler($this, $pos, $target, $face, 16);
 		}
 
--- a/src/player/SurvivalBlockBreakHandler.php
+++ b/src/player/SurvivalBlockBreakHandler.php
@@ -76,14 +76,19 @@ final class SurvivalBlockBreakHandler{
 		}
 		if($breakTimePerTick > 0){
 			$progressPerTick = 1 / $breakTimePerTick;
-			if ($this->player->getEffects()->has(VanillaEffects::HASTE())){
-				$amplifier = $this->player->getEffects()->get(VanillaEffects::HASTE())->getAmplifier() + 1;
+
+			$haste = $this->player->getEffects()->get(VanillaEffects::HASTE());
+			if ($haste !== null){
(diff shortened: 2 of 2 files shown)


dktapps:
Duplicates #5581

dktapps:
@GameParrot what's your source of information for the data in this PR?

GameParrot:
> @GameParrot what's your source of information for the data in this PR?

Haste speed modifier was from the wiki, mining fatigue speed modifier was found by testing (wiki numbers inaccurate)

dktapps:
I'd test those thoroughly if I were you. The last guy that tried to do this in #5581 couldn't get them to fully match up, not with wiki sources and not by guesswork either.

Even better if you can pull up an old BDS in IDA and find the code that handles this in vanilla.

GameParrot:
> I'd test those thoroughly if I were you. The last guy that tried to do this in #5581 couldn't get them to fully match up, not with wiki sources and not by guesswork either.
> 
> Even better if you can pull up an old BDS in IDA and find the code that handles this in vanilla.

Haste and mining fatigue seem pretty much perfect from my testing (tested with both low and high levels)

src/player/SurvivalBlockBreakHandler.php
```diff
@@ -65,11 +67,29 @@ private function calculateBreakProgressPerTick() : float{
...
+		if ($this->player->isUnderwater() && !$this->player->getArmorInventory()->getHelmet()->hasEnchantment(VanillaEnchantments::AQUA_AFFINITY())){
+			$breakTimePerTick *= 5;
+		}
 		if($breakTimePerTick > 0){
-			return 1 / $breakTimePerTick;
+			$progressPerTick = 1 / $breakTimePerTick;
+
+			$haste = $this->player->getEffects()->get(VanillaEffects::HASTE());
+			if ($haste !== null){
+				$amplifier = $haste->getAmplifier() + 1;
```
dktapps: Prefer using getEffectLevel() for this (and rename the variables).

src/player/SurvivalBlockBreakHandler.php
```diff
@@ -65,11 +67,29 @@ private function calculateBreakProgressPerTick() : float{
...
+				$progressPerTick *= (1 + 0.2 * $amplifier) * (1.2 ** $amplifier);
+			}
+
+			$miningFatigue = $this->player->getEffects()->get(VanillaEffects::MINING_FATIGUE());
+			if ($miningFatigue !== null){
+				$amplifier = $miningFatigue->getAmplifier() + 1;
```
dktapps: Same again here. Same goes for the variable names.

dktapps (changes requested):

src/player/Player.php
```diff
@@ -865,6 +865,29 @@ static function() : void{
 		Timings::$playerChunkSend->stopTiming();
 	}
 
+	/**
+	 * Checks if the player is currently on the ground. This is more accurate than {@link Player::isOnGround()} but slower.
+	 */
+	public function isActuallyOnGround() : bool{
```
ShockedPlot7560, line 871: Something like calculate or smth in the name can be better than this function combined with docs ?

src/player/Player.php
```diff
@@ -1764,7 +1787,7 @@ public function attackBlock(Vector3 $pos, int $face) : bool{
 			return true;
 		}
 
-		if(!$this->isCreative() && !$block->getBreakInfo()->breaksInstantly()){
+		if(!$this->isCreative() && !$target->getBreakInfo()->breaksInstantly()){
```
ShockedPlot7560: Why this change ?

src/player/SurvivalBlockBreakHandler.php
```diff
@@ -65,11 +67,29 @@ private function calculateBreakProgressPerTick() : float{
 		if(!$this->block->getBreakInfo()->isBreakable()){
 			return 0.0;
 		}
-		//TODO: improve this to take stuff like swimming, ladders, enchanted tools into account, fix wrong tool break time calculations for bad tools (pmmp/PocketMine-MP#211)
 		$breakTimePerTick = $this->block->getBreakInfo()->getBreakTime($this->player->getInventory()->getItemInHand()) * 20;
-
+		if (!$this->player->isActuallyOnGround() && !$this->player->isFlying()){
```
ShockedPlot7560, line 71: If you're flying i remember you break slower, why this is checked ?

src/player/SurvivalBlockBreakHandler.php
```diff
@@ -65,11 +67,29 @@ private function calculateBreakProgressPerTick() : float{
...
+			$breakTimePerTick *= 5;
+		}
 		if($breakTimePerTick > 0){
-			return 1 / $breakTimePerTick;
+			$progressPerTick = 1 / $breakTimePerTick;
+
+			$haste = $this->player->getEffects()->get(VanillaEffects::HASTE());
+			if ($haste !== null){
+				$hasteLevel = $haste->getEffectLevel();
+				$progressPerTick *= (1 + 0.2 * $hasteLevel) * (1.2 ** $hasteLevel);
```
ShockedPlot7560, line 83: where are these value from ?

src/player/SurvivalBlockBreakHandler.php
```diff
@@ -65,11 +67,29 @@ private function calculateBreakProgressPerTick() : float{
...
+			}
+
+			$miningFatigue = $this->player->getEffects()->get(VanillaEffects::MINING_FATIGUE());
+			if ($miningFatigue !== null){
+				$miningFatigueLevel = $miningFatigue->getEffectLevel();
+				$progressPerTick *= 0.21 ** $miningFatigueLevel;
```
ShockedPlot7560, line 89: same here, where are these from?

ShockedPlot7560 (changes requested):

src/player/Player.php
GameParrot: The check should be done on the block being broken (fixes break animation not showing when breaking a block behind an insta break block such as a torch)

src/player/SurvivalBlockBreakHandler.php
GameParrot: > If you're flying i remember you break slower, why this is checked ?

From my testing break speed while flying in survival is the same as when on the ground

src/player/SurvivalBlockBreakHandler.php
GameParrot: > where are these value from ?

Wiki

src/player/SurvivalBlockBreakHandler.php
GameParrot: > same here, where are these from?

My testing (see https://github.com/pmmp/PocketMine-MP/pull/6500#issuecomment-2479218330)

src/player/Player.php
dktapps: This looks like a bug, please submit a separate PR to fix it

GameParrot: > This looks like a bug, please submit a separate PR to fix it

https://github.com/pmmp/PocketMine-MP/pull/6528

src/player/Player.php
dktapps: Yeah, I'm not a fan of this naming either. Also, why do we actually need this?

GameParrot: > Yeah, I'm not a fan of this naming either. Also, why do we actually need this?

We need it because when using the normal is on ground method the break progress indicator is still too fast when jumping. What should it be renamed to?

src/player/Player.php
```diff
@@ -869,7 +869,7 @@ static function() : void{
 	 * Checks if the player is currently on the ground. This is more accurate than {@link Player::isOnGround()} but slower.
 	 */
 	public function isActuallyOnGround() : bool{
-		$bb = $this->boundingBox->expandedCopy(0, 0.001, 0);
+		$bb = $this->boundingBox->expandedCopy(-0.001, 0.001, -0.001);
```
dktapps, line 872: This change doesn't look related to the PR. Why would this be needed for ground calculation?

GameParrot: Fixes inaccurate progress at high x/z coordinates

dktapps: Many things break down at large coordinates. I doubt this is the proper fix for the problem.

src/player/Player.php
dktapps: Sounds like that warrants investigation into regular onGround behaviour.

dktapps: Legend has it that `PlayerAuthInputPacket` should allow making onGround checks more accurate, so this hack wouldn't be needed anymore.

dktapps: I think the PR should use the built-in onGround for now. It doesn't make sense to add an extra function specifically for this.

Once the proper onGround is fixed per #6547 we won't need this workaround. I don't like adding stuff to the API that we know we're gonna delete anyway.

dktapps (changes requested):
Please change this to use the normal `onGround()` and remove `isActuallyOnGround()`.

Fixing `onGround()` needs to be done separately from this change as it affects a variety of other functions and the way this PR is doing it isn't the right way.

Including this change in here is blocking the PR.

dktapps:
Adjusted and manually merged via 0a9a45a126ff30bf385a53d5b170b80071c6198a. Your fork doesn't seem to allow maintainer updates so I couldn't merge it the normal way (probably because it's a fork of a fork).

PR #6501, opened 2024-11-13
Added default permission "console"

### LINKED ISSUE #6499: Ability to give `plugin.yml` permissions to console only
## Description
Now if you specify the permission, marked as "default: false", the console won't be able to run the command
```yml
permissions:
  myplugin.command.test:
    default: false
```
I suggest allowing the ConsoleCommandSender to execute such commands

## Justification
It is often necessary to restrict a dangerous (or unwanted) command even to server operators. In this case, access to this command can be given manually to players through the permission manager. But in that case, the ConsoleCommandSender will lose the access to it, which is confusing

## Alternative methods
Code like this may be used to force-grant the permission
```php
$server = Server::getInstance();
$refObj = new \ReflectionObject($server);
$field = $refObj->getProperty("consoleSender");
$field->setAccessible(true);

$permission = PermissionManager::getInstance()->getPermission("myplugin.command.test");

$consoleSender = new ConsoleCommandSender($server, $server->getLanguage());
$consoleSender->addAttachment($this)->setPermission($permission, true);

$field->setValue($server, $consoleSender);
```

fsetinh:
Added a new default permissions value `console` to make sensitive commands restricted to console only.

Related to this #6499

### Related issues & PRs
Fixes #6499 

### Behavioural changes
Any command with `default: console` permission added in its `plugin.yml` will only work and show in console.

### DIFF
--- a/src/permission/PermissionParser.php
+++ b/src/permission/PermissionParser.php
@@ -31,6 +31,10 @@ class PermissionParser{
 
 	public const DEFAULT_OP = "op";
 	public const DEFAULT_NOT_OP = "notop";
+
+	public const DEFAULT_CONSOLE = "console";
+
+	public const DEFAULT_NOT_CONSOLE = "notconsole";
 	public const DEFAULT_TRUE = "true";
 	public const DEFAULT_FALSE = "false";
 
@@ -41,6 +45,8 @@ class PermissionParser{
 		"isoperator" => self::DEFAULT_OP,
 		"admin" => self::DEFAULT_OP,
 		"isadmin" => self::DEFAULT_OP,
+		"console" => self::DEFAULT_CONSOLE,
+		"isconsole" => self::DEFAULT_CONSOLE,
 
 		"!op" => self::DEFAULT_NOT_OP,
 		"notop" => self::DEFAULT_NOT_OP,
@@ -48,6 +54,8 @@ class PermissionParser{
 		"notoperator" => self::DEFAULT_NOT_OP,
 		"!admin" => self::DEFAULT_NOT_OP,
 		"notadmin" => self::DEFAULT_NOT_OP,
+		"!console" => self::DEFAULT_NOT_CONSOLE,
+		"notconsole" => self::DEFAULT_NOT_CONSOLE,
 
 		"true" => self::DEFAULT_TRUE,
 		"false" => self::DEFAULT_FALSE,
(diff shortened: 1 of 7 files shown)


src/permission/PermissionParser.php
```diff
@@ -31,6 +31,10 @@ class PermissionParser{
 
 	public const DEFAULT_OP = "op";
 	public const DEFAULT_NOT_OP = "notop";
+
+	public const DEFAULT_CONSOLE = "console";
+
+	public const DEFAULT_NOT_CONSOLE = "notconsole";
```
dktapps, line 37: Don't bother with this. You haven't implemented support for it and no one will use it. (Trust me, no one understands what `notop` is supposed to be used for either.)

src/permission/PermissionParser.php
```diff
@@ -41,13 +45,17 @@ class PermissionParser{
 		"isoperator" => self::DEFAULT_OP,
 		"admin" => self::DEFAULT_OP,
 		"isadmin" => self::DEFAULT_OP,
+		"console" => self::DEFAULT_CONSOLE,
+		"isconsole" => self::DEFAULT_CONSOLE,
```
dktapps, line 49: Don't add stuff that doesn't exist.

dktapps (changes requested):

src/permission/PermissionParser.php
Gaprix: The "notconsole" may be useful for commands that are not supported to run in cli mode. For example, a command that opens a form for the sender

dktapps: Inverted permissions aren't well-defined.

I also don't think denying permission is the right way to address that, since it'll give confusing messages.

dktapps: You'd also still end up needing to check instanceof of command sender within the execute() anyway to ensure that you're not letting in other types of command senders, or console senders which have had this permission overridden to true.

ShockedPlot7560:
I don't see the real purpose of the PR apart from adding an extra permission that may never be used and that won't solve the problem of having to check the CommandSender instance.

dktapps:
> I don't see the real purpose of the PR apart from adding an extra permission that may never be used and that won't solve the problem of having to check the CommandSender instance.

It's actually used internally already for `/dumpmemory`. There's definitely a use case for restricting access to terminal-only.

PR #6522, opened 2024-11-17
Fix: implicitly nullable parameter declarations deprecated in PHP 8.4

AkmalFairuz:
https://wiki.php.net/rfc/deprecate-implicitly-nullable-types


dktapps:
Is this the only occurrence?

AkmalFairuz:
I think yes. I used this regex in PhpStorm to find the occurrence
```
(?<!\?)\b(\w+)\b\s*(\&?\$\w+)\s*=\s*null(?=\s*\))
```

dktapps (approved):

PR #6525, opened 2024-11-19
Added the new colors

WavyCraft:
Added the ”new” colors within the TextFormat and also removed useless chat formats “STRIKETHROUGH” and “UNDERLINE”

Source: https://minecraft.fandom.com/wiki/Formatting_codes

Added to Bedrock edition a while back and it would be cool to add them to Pocketmine-MP.


ShockedPlot7560:
Duplication of #6483 #5838

PR #6529, opened 2024-11-23
Implemented the correct walking speed for the slime block

WavyCraft:
This PR implements the default walking speed of (0.4) as the “TODO” suggest and added new methods to modify the walking speed or get the walking speed of the slime block.

I sorta guessed what that “TODO” meant.

### DIFF
--- a/src/block/Slime.php
+++ b/src/block/Slime.php
@@ -28,6 +28,8 @@ use pocketmine\entity\Living;
 
 final class Slime extends Transparent{
 
+	private float $speedMultiplier = 0.4;
+
 	public function getFrictionFactor() : float{
 		return 0.8; //???
 	}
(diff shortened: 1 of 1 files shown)


src/block/Slime.php
```diff
@@ -40,5 +42,19 @@ public function onEntityLand(Entity $entity) : ?float{
 		return -$entity->getMotion()->y;
 	}
 
-	//TODO: slime blocks should slow entities walking on them to about 0.4x original speed
+	public function onEntityWalking(Entity $entity) : void{
+        $motion = $entity->getMotion();
+        $entity->setMotion($motion->multiply($this->speedMultiplier));
+    }
+
+    public function setSpeedMultiplier(float $multiplier) : void{
```
dktapps, line 50: You can't set this. Block objects aren't persistent.

src/block/Slime.php
```diff
@@ -40,5 +42,19 @@ public function onEntityLand(Entity $entity) : ?float{
 		return -$entity->getMotion()->y;
 	}
 
-	//TODO: slime blocks should slow entities walking on them to about 0.4x original speed
+	public function onEntityWalking(Entity $entity) : void{
```
dktapps, line 45: This isn't going to magically work. Nothing is going to call it.

dktapps (changes requested):

src/block/Slime.php
ShockedPlot7560: If this is really needed, you should do it across events

dktapps: It's not reliable under any conditions as the cache for blocks may be destroyed any time. We don't have setters for any other type-level properties like this. Unless this was stored in a blockstate property or tile property, it will just vanish and you'll be none the wiser why it doesn't work.

dktapps:
Closing due to lack of response

PR #6532, opened 2024-11-24
Introduce Utils::getRandomFloat()

### LINKED ISSUE #6521: lcg_value() is deprecated in PHP 8.4
## Description
lcg_value() is deprecated in PHP 8.4. I wanted to make a PR about this, but I see that the replacement is (new Randomizer())->getFloat(...), so maybe we need another replacement (maybe create a `Utils::getRandomFloat()`?).

`Deprecated: Function lcg_value() is deprecated since 8.4, use \Random\Randomizer::getFloat() instead in phar:///tmp/PocketMine-MP-phar-cache.0/PMMPiq4ekripkj4b26HJHGV.tar/src/world/World.php on line 2004`

AkmalFairuz:
### Related issues & PRs
* Fix https://github.com/pmmp/PocketMine-MP/issues/6521

### API changes
- Added `pocketmine\utils\Utils::getRandomFloat(float $min = 0.0, float $max = 1.0): float`


src/utils/Utils.php
```diff
@@ -675,4 +677,11 @@ function_exists('opcache_get_status') &&
 		//jit not available
 		return null;
 	}
+
+	public static function getRandomFloat(float $min = 0.0, float $max = 1.0) : float{
+		if($min > $max){
+			throw new \InvalidArgumentException("Minimum value cannot be greater than maximum value");
+		}
+		return $min + (mt_rand() / mt_getrandmax()) * ($max - $min);
```
dktapps: This is overengineered. These extra parameters are not needed.

src/entity/Living.php
```diff
@@ -490,7 +490,7 @@ protected function applyPostDamageEffects(EntityDamageEvent $source) : void{
 				$helmet = $this->armorInventory->getHelmet();
 				if($helmet instanceof Armor){
 					$finalDamage = $source->getFinalDamage();
-					$this->damageItem($helmet, (int) round($finalDamage * 4 + lcg_value() * $finalDamage * 2));
+					$this->damageItem($helmet, (int) round($finalDamage * (4 + Utils::getRandomFloat(0.0, 2.0))));
```
dktapps: This doesn't look right

dktapps (changes requested):
This would've been a lot more straightforward to review if you hadn't introduced unnecessary extra changes

src/utils/Utils.php
AkmalFairuz: Removed. Just `Utils::getRandomFloat(): float`

AkmalFairuz:
Would it be better if we implement `Math::random(): float` in [pmmp/Math](https://github.com/pmmp/Math) like Java and Javascript?

dktapps:
Well, really the randoms should be derived from a World-local Random object instead of on a process-global Random state, since the current way makes it basically impossible to reproduce random-driven events with a seed. However that's outside the scope of this PR.

Basically I don't think there's a need for modifying the Math library, although others may disagree.

dktapps (dismissed):

ShockedPlot7560 (dismissed):

ipad54 (dismissed):
This needs to target `minor-next`

SOF3:
> Basically I don't think there's a need for modifying the Math library, although others may disagree.

The Math library right now is more like a linalg/minecraft-coordinates library. I don't think it is a good idea to add random math-related functions into that library.

dktapps (approved):

Issue #6147, opened 2023-11-09
Positionless tiles


dktapps:
## Description
Positionless blocks has already been planned since PM3, as they have various advantages over the current design.

Positionless tiles would be able to be created and used in AsyncTasks, as well as being able to be freely copied. However, there are some obstacles:

- Block inventories require positions. So far as I can tell, there's no obvious reason for this other than giving plugins information in events such as `InventoryOpenEvent`, which can be achieved in other ways. #5033 
- Some blocks, such as brewing stands, need to register InventoryListeners that know their current position in order to notify the World to schedule an update on the block.
  - This could be solved by requiring any updater of the inventory (e.g. an InventoryWrapper in #5033, or a hopper setting items into another inventory) to schedule an update on the block in question instead of doing it automatically, but this feels clunky, is less nice than the way this currently works, and prone to bugs if someone forgets to schedule the update.
  - However, it would also allow a plugin to, for example, put items into a brewing stand without activating it, e.g. if someone wanted to make brewing stands redstone-activated or something similar.

## Justification
Described above. Less dependencies = more usability + easier copying. Tile data should be copyable just as easily as block data.

Issue #2684, opened 2019-01-21
OfflinePlayer inventory access


SOF3:
## Description
Ability to access and modify the OfflinePlayer inventory

## Justification
Scheduled awards. Asynchronous item manipulation. Late transactions. Dozens of reasons; XenForo trophy system provides some nice example usages.

## Alternative methods
Edit NBT directly? It is not even documented...

dktapps:
Amusingly enough `OfflinePlayer` currently doesn't even give you access to the NBT at all, so you couldn't do that anyway. I'm gunning to get rid of all the "API"ish NBT in the core for future releases, so we don't want users doing that.

The internal architecture of inventories is too complicated to permit this currently (due to cyclic dependencies between inventory and its holder).

dktapps:
Current intent is to address this in PM5 by removing holder information from the inventory itself (this will fix a bunch of other issues too).

The only time when holder information is actually useful is in:
- InventoryOpenEvent
- InventoryCloseEvent
- InventoryTransactionEvent

In all of these cases, an ephemeral wrapper around the inventory which tracks the holder specifically for the event would be more practical and less of a pain in the ass.

dktapps:
Dependent on #5033. Big headache for anyone willing to tackle that...

PR #6533, opened 2024-11-24
Separate inventory holder info from container & player inventories

### LINKED ISSUE #5033: Breaking cyclic reference between Inventory and its holder
## Description
Many places in the PM core (such as [here](https://github.com/pmmp/PocketMine-MP/blob/7c3b78b0a0ae848a8fb034251ef53edb7b0c08e4/src/player/Player.php#L2143-L2144)) feature hacks to destroy the inventory to destroy cyclic references and prevent memory leaks.

This is problematic for static analysis, makes post-dispose use more problematic (technically not supported use case, but it should still work), and is caused by an entirely unnecessary requirement.

The only practical purpose for inventories to contain a reference to its `holder` is to allow users of events such as `InventoryTransactionEvent` to identify the owner of an inventory. Outside of events, it has no practical purpose, since you typically have to go through the holder of an inventory to fetch the inventory in the first place - e.g. getting a `Player`'s inventory requires use of `Player->getInventory()`, which requires a `Player` reference to begin with, so why would you use `getHolder()`?

This issue proposes that:
- Inventories will no longer contain a reference to their holder. Instead, actors with inventories will hold `SimpleInventory` directly.
- When a transaction occurs, these inventories will be wrapped in a `new InventoryWrapper(inventory, holder)` to allow plugins using events to identify them.

Going further, we could also move window type information to `InventoryWrapper`, enabling completely dynamic window types to be used for inventories from any source.

dktapps:
This PR breaks the cyclic dependency between `Inventory` and its holder, which unblocks a lot of new developments.

### Related issues & PRs
- Fixes #5033
- Removes a blocker for #6147 (which in turn means that async tasks will eventually be able to work with tiles)
- Removes a blocker for #2684

### API changes
- `Player->getCurrentWindow()` now returns `?InventoryWindow` instead of `?Inventory`
- `Player->setCurrentWindow()` now accepts `?InventoryWindow` instead of `?Inventory`
- `InventoryWindow` introduced, which is created for each player viewing the inventory, provides decorative information like holder info for `InventoryTransactionEvent`, and is destroyed when the window is closed, eliminating cyclic references
- Added:
  - `player\InventoryWindow`
  - `player\PlayerInventoryWindow` - wraps all permanent inventories of Player with type info for transactions
  - `inventory\Hotbar` - replaces all hotbar usages in `PlayerInventory`
  - `Human->getHotbar()`
  - `Human->getMainHandItem()`, `Human->setMainHandItem()`, `Human->getOffHandItem()`, `Human->setOffHandItem()`
  - `block\utils\AnimatedContainerLike` & `block\utils\AnimatedContainerLikeTrait` (for chests, shulkerboxes, etc)
  - `block\utils\Container` & `block\utils\ContainerTrait` for blocks containing items (chests, etc)
  - `block\utils\MenuAccessor` implemented by all blocks that can open inventory menus
  - `block\utils\MenuAccessorTrait` used by blocks with menus but without inventories (anvils, crafting tables etc)
- Removed:
  - `inventory\DelegateInventory` (only used for ender chests)
  - `inventory\PlayerInventory`,
  - `inventory\PlayerOffHandInventory`,
  - `inventory\PlayerCraftingInventory`,
  - `inventory\PlayerCursorInventory` - these have all been internally replaced by `SimpleInventory` & they will appear as `PlayerInventoryWindow` in transactions (check `getType()` against the `PlayerInventoryWindow::TYPE_*` constants to identify them)
  - `block\inventory\AnimatedBlockInventoryTrait`, (blocks now handle this logic directly using `AnimatedContainer` and `AnimatedContainerTrait`)
  - `block\inventory\BlockInventoryTrait`,
  - `block\inventory\BlockInventory`
- Most `BlockInventory` classes have been transitioned to `InventoryWindow` wrappers
- Tiles now all use `SimpleInventory` internally (no cyclic references) except for `Chest` (which uses `CombinedInventory`, without holder info)
- `InventoryOpenEvent` and `InventoryCloseEvent` now provide `InventoryWindow` instead of `Inventory` (to provide type information)
- `InventoryTransaction` and `SlotChangeAction` now provide `InventoryWindow` instead of `Inventory`
- Renamed `TransactionBuilderInventory` to `SlotChangeActionBuilder`
- `TransactionBuilderInventory->getBuilder()` now accepts `InventoryWindow` instead of `Inventory`
- `DoubleChestInventory` superseded by `CombinedInventory` - this new class allows combining any number of inventories behind a single object; mainly used for double chests but plugins could use it to do lots of fun things

### Impacts to plugins
Plugins can now do the following:
```php
$block = $world->getBlockAt($x, $y, $z);
if($block instanceof MenuAccessor){
    $block->openToUnchecked($player);
}
```
As compared to the old way:
```php
$tile = $world->getTileAt($x, $y, $z);
if($tile instanceof Container){
    $player->setCurrentWindow($tile->getInventory());
}
```

#### Advantages
- No tile access needed
- Works for menu blocks without inventories as well as container blocks
- Less code

### Behavioural changes
Inventories no longer keep permanent cyclic references to their holders.

## Backwards compatibility
This makes significant BC breaks. However, all changes are able to be adapted to and the same amount of information is present on all APIs and events.

## Follow-up
- Implement #6147 
- Support inventory inheritance when copying blocks from one position to another

## Tests
Briefly tested in-game.


src/block/inventory/window/EnchantingTableInventoryWindow.php
```diff
@@ -0,0 +1,110 @@
...
+			},
+			onContentChange: static function() use ($weakThis) : void{
+				if(($strongThis = $weakThis->get()) !== null){
+					$strongThis->regenerateOptions();
+				}
+			}
+		);
+		$this->inventory->getListeners()->add($listener);
+
+		$this->listener = \WeakReference::create($listener);
```
dktapps: Probably need to clean this up, think I forgot about it

ShockedPlot7560:
I have no real problem with the current implementation. It seems fine to me and is a first step towards a better inventory system.

Another follow-up I would suggest concerns the viewers of the inventory. I can see the point of including viewers in this system but I still have a sticking point because we're including a reference to the player in the inventory interface. Wouldn't it be good after this to think of a system where we have classes that only contain storage logic / listeners and that's it ?   
This may also be a bad idea, potentially making the code into spaghetti for inventories.

dktapps:
> I have no real problem with the current implementation. It seems fine to me and is a first step towards a better inventory system.
> 
> Another follow-up I would suggest concerns the viewers of the inventory. I can see the point of including viewers in this system but I still have a sticking point because we're including a reference to the player in the inventory interface. Wouldn't it be good after this to think of a system where we have classes that only contain storage logic / listeners and that's it ? This may also be a bad idea, potentially making the code into spaghetti for inventories.

Yeah, that's the ideal world. Unfortunately we need the Inventory to still track the viewers so that plugins can see all the viewers of an inventory, and also so that the inventory menus get closed when a block is deleted. I don't see another obvious solution to this problem. We could pull the viewer list out of `Inventory` and have the container itself manage it, but I'm not sure if that would improve anything.

dktapps:
Few issues remain with the PR:

- [x] `BlockInventoryWindow` subclasses have minimum 3 constructor parameters - annoying for plugins, lots of room to break things - I'm thinking something like `$chest->openTo(Player $player)` would be a good idea to simplify the usage
- [x] I haven't figured out a solution for stuff like `Furnace` which uses an `InventoryListener` to send itself a scheduled block update on a position. That's a remaining blocker for #6147.
~~Possibly could use `Block` in `BlockInventoryWindow` - this is fine now that the windows aren't referenced by holders anymore, we don't need to worry about cyclic references~~
- [x] No inventory type info in windows for single chest, shulker, barrel & ender chest - this isn't needed for internal use but would probably be desired by plugins; could just let them check the block type but I don't know if this should be bound to blocks

~~Possibly could use `Block` in `BlockInventoryWindow` - this is fine now that the windows aren't referenced by holders anymore, we don't need to worry about cyclic references~~

dktapps:
Double chests have also proven to be a pain in the ass and will likely be an obstacle for #6147 as well

dktapps:
Only remaining major issue is around the lengthened usage for opening inventory windows to players.

I didn't quite decide how to deal with it. There are a few things to consider:
- Not all blocks which open windows have inventories of their own (e.g. anvils, ender chests, crafting tables, etc.)
- Blocks which _do_ have inventories of their own may not have those inventories present (e.g. due to missing tiles) and so a mode of allowed failure would need to be introduced (e.g. `Chest->openTo()` would need to be allowed to fail). This sucks, since it will _not_ fail in 99.9% of cases (only those special cases where something hasn't been done correct by a plugin, or old worlds).
- Blocks which have inventories also usually have a feature to test for container lock. The player must use an item with a matching custom name to open such containers. Should `openTo()` check this kind of logic? If not, how do we avoid duplicated logic fetching tile in `onInteract()` vs `openTo()`? (Side note: I suppose plugins already need this boilerplate duplicated logic on stable branches, since we don't offer a simple way to get the inventory anyway.)

dktapps:
I think finishing this with nice APIs may require differentiating between "container" and "menu" blocks.
Container blocks have inventories (chest, shulkerbox, hopper, furnace, etc) while menu blocks don't (anvil, crafting table, loom, etc).

Opening a menu block should never fail since these blocks don't have inventories and cannot have locks.
Opening container blocks may fail due to incorrect item key or the tile being missing.

I also think the changes wrt. pushing inventory updates through `World` should probably be reverted also, since these aren't related to the window refactor and I'm also doubtful about the way it's been done. We'll need something like it in the future for tile de-positioning (if that's even actually possible), but for this specific PR it's not required.

dktapps:
I'm also wondering whether the player's main inventory window handling should be redone as part of this PR. We'd probably need to introduce a separate "hotbar" PlayerInventoryWindow type to facilitate this.

src/player/Player.php
```diff
@@ -2785,10 +2803,10 @@ public function setCurrentWindow(Inventory $inventory) : bool{
 		if(($inventoryManager = $this->getNetworkSession()->getInvManager()) === null){
 			throw new \InvalidArgumentException("Player cannot open inventories in this state");
 		}
-		$this->logger->debug("Opening inventory " . get_class($inventory) . "#" . spl_object_id($inventory));
-		$inventoryManager->onCurrentWindowChange($inventory);
-		$inventory->onOpen($this);
-		$this->currentWindow = $inventory;
+		$this->logger->debug("Opening inventory " . get_class($window) . "#" . spl_object_id($window));
```
ShockedPlot7560: This can probably be changed in `Opening window` to keep the right vocabulary

src/player/Player.php
```diff
@@ -2797,7 +2815,7 @@ public function removeCurrentWindow() : void{
 		if($this->currentWindow !== null){
 			$currentWindow = $this->currentWindow;
 			$this->logger->debug("Closing inventory " . get_class($this->currentWindow) . "#" . spl_object_id($this->currentWindow));
```
ShockedPlot7560: Same here

src/player/Player.php
```diff
@@ -2806,20 +2824,31 @@ public function removeCurrentWindow() : void{
...
-		foreach($inventories as $inventory){
-			$inventory->onOpen($this);
-			$this->permanentWindows[spl_object_id($inventory)] = $inventory;
+	/**
+	 * @param PlayerInventoryWindow[] $windows
+	 */
+	protected function addPermanentWindows(array $windows) : void{
+		foreach($windows as $window){
+			$window->onOpen();
+			$this->permanentWindows[spl_object_id($window)] = $window;
```
ShockedPlot7560, line 2862: This can probably now create confusion ? Before, the inventory is normally the same instance while now the window could be different with the same inventory instance.

ShockedPlot7560 (changes requested):

src/player/Player.php
dktapps: No, these windows are always open, they can't be changed

dktapps:
Alright. I think I'm finally done with this

dktapps:
In the interest of avoiding branch rot I'm going to get this merged, the final result won't look too different from this anyway. Pls let me know if there are issues found after the fact.

Issue #5628, opened 2023-03-12
Disable server thread automatic cyclic garbage collection, and let PM MemoryManager exclusively decide when to run GC


dktapps:
## Description
As of PHP 7.3, GC roots are always recorded. Before 7.3, only the first 10k roots were recorded while automatic GC was turned off, and the remainder would be ignored, leading to potential memory leaks.

Since the GC can be triggered automatically seemingly at random when the GC root buffer reaches a certain size, seemingly random lag spikes can occur in code that wouldn't otherwise be slow, causing weird outputs on timings reports.

Instead, I think it makes sense to disable automatic GC and have the server decide when to do it. This would have the following benefits:

- No random lag spikes due to GC kicking in when an object happens to get destructed. This will improve performance in hot code paths.
- Time spent by the garbage collector can be properly accounted for in timings, since `MemoryManager` periodic GC is covered by timings already. Currrently, timings can only account for manual GC runs, and automatic ones triggered by Zend are hidden in other timings.

Downsides:
- It's possible that hot loops in plugin code may generate enough cyclic references to cause an OOM before MemoryManager can clean up. However, if this does happen, it should be obvious to plugin devs that their code has a problem.

dktapps:
My initial attempts at this involved using `gc_disable()`, and then temporarily enabling it during `MemoryManager::tick()`, and then attempting to trigger GC by destroying an object to allow PHP's regular GC heuristics to decide whether to try running GC or not.
However, this was unsuccessful, since I wasn't able to find a way to trigger a GC run without creating a cycle, which adds objects to the GC root buffer and ends up triggering more frequent GC runs.

dktapps:
This may become viable with PHP 8.3 thanks to new information added to `gc_status()`

dktapps:
> My initial attempts at this involved using `gc_disable()`, and then temporarily enabling it during `MemoryManager::tick()`, and then attempting to trigger GC by destroying an object to allow PHP's regular GC heuristics to decide whether to try running GC or not. However, this was unsuccessful, since I wasn't able to find a way to trigger a GC run without creating a cycle, which adds objects to the GC root buffer and ends up triggering more frequent GC runs.

Some more info on this, for posterity: The objective here was to keep the PHP default collection behaviour, but to just control _when_ it gets run. However, this turned out to be a real pain in the ass.

[...]

PR #6554, opened 2024-12-02
PM managed GC

dktapps:
This is a first look at #5628.

In a nutshell: Server `MemoryManager` now controls when GC is called. The mechanism by which it's activated remains the same as PHP's own, but instead of being called in random code paths & causing huge lag spikes, it gets called at a predictable location & can therefore be tracked in timings reports.

This required re-implementing PHP's garbage collection threshold algorithm, as it's not possible to reliably make PHP update its own internal threshold when GC isn't always enabled.

A couple of considerations I came across while working on this. It seems like PHP's GC algorithm is rather suboptimal for PM's use case.
- The average GC run in PM takes about 10-15ms. To my surprise, the biggest factor influencing performance is actually the depth of referenced objects. This means that anything referencing `Server` getting unref'd causes a hit to GC performance because `Server` will land in the GC root buffer, and `Server` is a gateway to everything.
- Long-lived complex objects like `Server` will frequently end up taking up space in the GC root buffer (whenever their refcounts go down), and then the GC wastes a ton of CPU time trying to scan them for dependencies. There's no real solution for this beyond requesting PHP to implement generations in the GC (php/php-src#17131).
- Acyclic objects flood the root buffer and cause the threshold to go sky-high. In itself, this doesn't affect the GC's performance that much (e.g. a Vector3 takes only tens of nanoseconds per object to dispose), but the filling up of the GC root buffer causes more frequent runs, causing complex objects to get repeatedly scanned as in the above point.

## Follow-up
- Explore how much performance benefit we really get from the World's block cache. These cached objects tend to flood the GC root buffer, causing collection sweeps that collect barely any objects.
- After re-implementing the PHP native threshold algorithm, I realized it probably wasn't optimal for PM's use case. For example, the default root buffer size should probably be significantly larger (around 100k or so) to allow for the World block cache to not constantly trigger GC.


dktapps:
Related issues:

https://github.com/php/php-src/issues/17127 (Acyclic objects landing in GC root buffer)
https://github.com/php/php-src/issues/17131 (No generational GC to avoid wasting CPU time on long-lived objects)

Issue #4640, opened 2021-12-11
No drops from lantern.


hausemaster3734:
### Issue description

- Expected result: When you break the block under (or) above the lantern, he had to fall as an item
- Actual result: When under (or) above the  lantern block breaks, it breaks, but does not fall out as a item

### Steps to reproduce the issue
1. Put the lantern block.
2. Break block under lantern.
3. No drops.

### OS and versions

* PocketMine-MP: 4.0.0
* PHP: 8.0
* Using JIT: No
* Server OS: Linux Mint

### Plugins
* None

### Crashdump, backtrace or other files

* None

dktapps:
Same underlying issue as #4271

PR #6378, opened 2024-06-26
Massive block's hardness & blast resistance update

IvanCraft623:
## Introduction

It is important to keep these values up to date, not only for vanilla-parity, anti-cheats may also depend on them.

## Follow-up
Add a file to bedrock-data that contains block's hardness & blast resistance values, to be able to make a PHPUnit test.

## Tests
I tested this PR by doing the following (tick all that apply):
- [x] Writing a test plugin (provide the code and sample output)

Vanilla hardness & blast resistance values(need to be droped in plugin_data) file: [blocks_hardness.json](https://github.com/user-attachments/files/15994442/blocks_hardness.json)
(File was provided by @TwistedAsylumMC)

Code:
```php
public function onEnable() : void {
	$ignoredStates = [
		"minecraft:air" => true,
		"minecraft:barrier" => true,
		"minecraft:flowing_lava" => true, //destructible in vanilla???
		"minecraft:flowing_water" => true, //destructible in vanilla???
		"minecraft:light_block" => true,
	];

	$blocksData = json_decode(file_get_contents($this->getDataFolder() . "blocks_hardness.json"), true);
	$serializer = GlobalBlockStateHandlers::getSerializer();
	foreach (VanillaBlocks::getAll() as $block) {
		$stateName = $serializer->serialize($block->getStateId())->getName();
		if (isset($ignoredStates[$stateName])) {
			continue;
		}
		foreach ($blocksData as $blockData) {
			if ($blockData["id"] !== $stateName) {
				continue;
			}
			$breakInfo = $block->getBreakInfo();
			$hardness = round($blockData["base_hardness"], 5);
			$blastResistance = round($blockData["blast_resistance"], 5) * 5;
			if ($hardness != $breakInfo->getHardness() ||
				$blastResistance != $breakInfo->getBlastResistance()
			) {
				$this->getLogger()->warning(
					"\n\nBlock(" . $block->getName() . ", class: " . basename(str_replace('\\', '/', $block::class)) . ", stateName: $stateName) has a mismatch hardness:\n" .
					"PM hardness: " . $breakInfo->getHardness() . "\n" .
					"PM blast_resistance: " . $breakInfo->getBlastResistance() . "\n" .
					"Vanilla hardness: " . $hardness . "\n" .
					"Vanilla blast_resistance: " . $blastResistance
				);
			}
			continue 2;
		}
		$this->getLogger()->error("Block with state id $stateName not found");
	}
	$this->getLogger()->info("Finished hardness test!");
}
```
[...]

PR #6573, opened 2024-12-13
Remove some blocks tool tier to match the recent vanilla changes

### LINKED ISSUE #4271: Iron doors don't drop when the supporting block is broken
### Issue description
In general, doors are broken and will drop themselves when the block they are placed on top of is destroyed. However, due to the tool checking logic within PM, iron doors don't drop, since they require a pickaxe to normally be broken, and the internal system considers this kind of breaking as the same as if the door was broken with an empty hand.

1. Place an iron door
2. Break the block beneath it
3. Observe the lack of any drops

### OS and versions

* PocketMine-MP: 3.21.0, 4.0.0
* PHP: irrelevant
* Server OS: irrelevant
* Game version: irrelevant

ipad54:
In the 1.21.50 release Mojang changed drop behaviour of some blocks and now they don't require tool tier and can be broken by hand.
This affects the following blocks:
 - Copper Doors
 - Iron Door
 - Heavy Weighted Pressure Plate
 - Light Weighted Pressure Plate
 - Polished Blackstone Pressure Plate
 - Stone Pressure Plate
 - Bell
 - Brewing Stand
 - Ender Chest
 - Lanterns

### Related issues
It seems like this Mojang change fixes #4271 & #4640

## Tests
https://www.youtube.com/watch?v=h53angduaxE

### DIFF
--- a/src/block/VanillaBlocks.php
+++ b/src/block/VanillaBlocks.php
@@ -879,12 +879,12 @@ final class VanillaBlocks{
 		self::register("bedrock", fn(BID $id) => new Bedrock($id, "Bedrock", new Info(BreakInfo::indestructible())));
 
 		self::register("beetroots", fn(BID $id) => new Beetroot($id, "Beetroot Block", new Info(BreakInfo::instant())));
-		self::register("bell", fn(BID $id) => new Bell($id, "Bell", new Info(BreakInfo::pickaxe(5.0, ToolTier::WOOD))), TileBell::class);
+		self::register("bell", fn(BID $id) => new Bell($id, "Bell", new Info(BreakInfo::pickaxe(5.0))), TileBell::class);
 		self::register("blue_ice", fn(BID $id) => new BlueIce($id, "Blue Ice", new Info(BreakInfo::pickaxe(2.8))));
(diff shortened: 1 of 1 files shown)


dktapps:
The question is whether AIR should be considered a compatible tool for these blocks because it affects the break time. There are other blocks that do require a tool, and yet still drop themselves when broken by hand anyway, but these blocks follow the break time of an incompatible tool.

Basically you need to test whether the break time animation still follows what the client expects when using wrong or no tools.

ipad54:
Tested with AntiInstaBreak plugin. Everything matches the client predicted result except the lanterns. After further research, it seems that lanterns have incorrect hardness according to BedrockData block properties dump. Changing the hardness from 5 to 3.5 fixed this issue

Yeah, we really need some kind of block properties test...

dktapps:
Many of the hardness values are outdated. Ivan was working on it in #6378 but he abandoned it and the PR accumulated a ton of conflicts.

I wonder if it might be worth adding a `BlockHardnessValues` class to the autogen in `pocketmine\data\bedrock`. That way the registrations could just reference the constants and we'd not have to manually change them all the time.

If anyone wants to tackle the hardness updates, make sure to coordinate it with the team so we can get it merged quickly, otherwise it'll just end up the same again.

dktapps (approved):

Issue #6580, opened 2024-12-20
LevelDB compaction performance issue with large worlds (especially converted worlds)


dktapps:
## Problem description
LevelDB automatic level compaction tends to continuously hammer the disk with larger worlds (several GB).

Compacting 10 GB of data (at level 4) can take around 30 minutes of continuous I/O usage. LevelDB attempts to distribute this work in order to not max out I/O, but this appears to manifest as continuous background load on disks.

## Wtf is compaction? Why is this happening?
Compaction orders all the data keys in a DB level according to the DB's comparator (by default a simple bytewise comparator that sorts like alphabetically but for byte values).
When a level in the DB fills up, the data from the lower level is merged with the higher level, and the higher level's data may be rebuilt to order it correctly.

During compaction, any files whose key ranges do not overlap with the higher level are directly moved to the higher level without rebuilding. This is cheap, so we want this to be the norm.
Files whose key ranges _do_ overlap are costly to merge into the higher level, because the new data must be sorted with the set of existing data in the higher level.

**This means that key structure is critical to getting best compaction performance.** Keys that are accessed at similar times (e.g. chunks with similar coordinates) should have leading key bits that are as similar as possible to minimize range overlap.

However, thanks to Mojang's choice of key format, the data from lower levels is practically guaranteed to overlap because the key ordering is so disorganized relative to actual data usage patterns. This means that every compaction is I/O intensive. This is a big problem with large worlds, where the I/O cost becomes very obvious and sometimes continuous.

There are two main problems:
- Keys are formatted XXXXZZZZ, so chunks with the same X coordinate will appear next to each other regardless of how far apart they are in the actual world. e.g. if you have chunks `0,0`,`0,1`,`1,0`,`1,1`, chunk `0,2` will land in the middle of this order, requiring a lot of data to be moved. Scale this up 10 million times and you can see why large DBs are a problem.
- Keys use little-endian packed int32s, meaning that chunks with similar coordinates will appear about as far apart in the DB as they possibly could (e.g. X coordinates would be ordered like 0, 256, 65536, 1, 257, 65537, etc) since the keys are compared byte-by-byte. This is like having an index at a library having the names of books sorted according to `strrev(name)`.

## Proposed solution

There's a couple of ways to solve this problem:
1) Use a custom comparator to sort the keys optimally - Not super portable because the whole DB must be rebuilt if comparator was changed
2) Change the key structure -also not ideal, would _also_ require a full DB rebuild
[...]

PR #6582, opened 2024-12-21
EXPERIMENTAL! Introduced a custom LevelDB impl with regions and better key format

dktapps:
This new impl (which is not loadable by vanilla) is targeted at very large worlds, which experience significant I/O performance issues due to a variety of issues described in https://github.com/pmmp/PocketMine-MP/issues/6580.

Two main changes are made in RegionizedLevelDB:
- First, multiple LevelDBs are used, which cover a fixed NxN chunk segment of terrain, similar to Anvil in Java. However, there's no technical constraint on these region sizes, unlike Anvil (which is limited to 32x32 chunks). Several experimental sizes are supported by default in WorldProviderManager.
- Second, bigEndianLong(morton2d(chunkX, chunkZ)) is used for chunk keys instead of littleEndianInt(chunkX).littleEndianInt(chunkZ). This new scheme has much better cache locality than Mojang's version, which reduces overlap and costly DB compactions.

The following new provider options are available as a result of this change:
- custom-leveldb-regions-128
- custom-leveldb-regions-256

128 will probably be the sweet spot. 256 will generate 4x fewer regions (might be better for space & file handle usage), but will probably experience more costly compactions (though still far less expensive than multi-GB worlds in regular leveldb format).

Note that the different variations of custom-leveldb-regions-* are not cross-compatible.
Conversion between the different formats is necessary if you want to change formats.

### Related issues & PRs
Related to #6580 

## Backwards compatibility
Should be fully backwards compatible. While significant changes were made to `LevelDB`, this was mainly in the interest of extracting a base class for `RegionizedLevelDB` to inherit from.

## Follow-up
- Add warnings for users with large worlds that their world performance will probably suffer if using `leveldb`
- Make regionized leveldb the default
- Add an easier way to export a world in a specific format from PM (instead of using `convert-world.php` which requires a src install)

## Tests
TBD

### DIFF
--- a/src/world/format/io/WorldProviderManager.php
+++ b/src/world/format/io/WorldProviderManager.php
@@ -53,7 +53,7 @@ final class WorldProviderManager{
 
 		//any arbitrary size is supported, but powers of 2 are best
 		//these are the most likely to be useful
-		foreach([32, 64, 128, 256] as $regionLength){
+		foreach([128, 256] as $regionLength){
 			$this->addProvider(new WritableWorldProviderManagerEntry(
 				fn(string $path) => RegionizedLevelDB::isValid($path, $regionLength),
 				fn(string $path, \Logger $logger) => new RegionizedLevelDB($path, $logger, $regionLength),
--- a/src/world/format/io/WorldProviderManager.php
+++ b/src/world/format/io/WorldProviderManager.php
@@ -25,9 +25,11 @@ namespace pocketmine\world\format\io;
 
 use pocketmine\utils\Utils;
 use pocketmine\world\format\io\leveldb\LevelDB;
+use pocketmine\world\format\io\leveldb\RegionizedLevelDB;
 use pocketmine\world\format\io\region\Anvil;
 use pocketmine\world\format\io\region\McRegion;
 use pocketmine\world\format\io\region\PMAnvil;
+use pocketmine\world\WorldCreationOptions;
 use function strtolower;
 use function trim;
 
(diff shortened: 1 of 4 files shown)


dktapps:
Turns out 128x128 has a significantly larger performance benefit than I expected when converting worlds.

On account of the average region size at 128x128 chunks being about 50 MB, this means that level 3 compactions are avoided in the vast majority of cases. (256x256 hovers at about 200 MB, which exceeds the threshold for level 3 compactions).

In particular, converting to `custom-leveldb-regions-128` from any format is several times faster  than `custom-leveldb-regions-256` for a 10 GB world (4 hours vs 90-100 mins).

dktapps:
For posterity: I hesitate to keep going with this change, because I think LevelDB is just fundamentally unsuited to the task of storing worlds, regardless of whether a single DB is used or multiple.

While Mojang's usage of it also couldn't be any more suboptimal (about the worst possible cache locality on account of the poor key structure), LevelDB is also just not suited to this job.

Compactions are always going to be costly for storing blob data (usually a few KB per key in Bedrock), because keys and values are stored together. Values have to be copied during sorting, which greatly increases the I/O cost of compactions. [RocksDB BlobDB](https://rocksdb.org/blog/2021/05/26/integrated-blob-db.html) would be a much better solution for this.

Splitting up into many DBs reduces the amount of live data at any given time, meaning that only live DBs will get compaction work done. However, since compaction is all about sorting data (which doesn't really benefit Bedrock anyway since iteration & range scans are not really required beyond the scope of a single chunk), I'm questioning whether it even makes sense to use a KV-store at all.

I'm also hesitant to once again introduce a custom world format which is only supported by PocketMine-MP. The onus would be on PMMP to maintain support for it forever just like PMAnvil, except this would be rather more complicated.

If we do decide to introduce a custom format into mainline, we really need to get this right.

yoraze:
Is there any implementations of RocksDB on PHP tho?
I found one that is 8-9 years old (https://github.com/phpize7/php-rocksdb)
This could take time to update to PHP8 for implement & use but is this gonna worth it?

dktapps:
> Is there any implementations of RocksDB on PHP tho? I found one that is 8-9 years old (https://github.com/phpize7/php-rocksdb) This could take time to update to PHP8 for implement & use but is this gonna worth it?

Not clear. RocksDB with universal compaction would probably be better (and probably sharded into multiple DBs as shown here). But I don't have any plan to continue working on this anyway, since it's a ton of unpaid effort.

AzaleeX:
And why not do both? Let me explain: a detection of the structuring of the world file will activate a conversion.

This would allow the 2 types of world, LevelDB and RocksDB BlobDB, to be always in Pocketmine-MP. The problem is that PocketMine will grow in size.

dktapps:
Closing this as I don't plan to continue working on it

Issue #4981, opened 2022-04-21
Pooling skin data


dktapps:
## Description
As skins get more and more complex, their size grows considerably, to the point where it's normal for marketplace skins to exceed 2MB. In many cases these skins share the same geometry and imagery.

On large servers with many players, this stacks up, consuming large amounts of memory.

It's possible to reduce the memory consumption of skins by pooling the data of skins and refcounting them, and reusing the same byte arrays instead of keeping separate copies for every player. Similarly, the geometry JSON can be deduplicated this way, potentially saving a considerable amount of memory.

ghost:
So skins will consume less memory by using shared pixels?

dktapps:
Yes, that's the idea.

StefanFox-Dev:
norm

PR #6589, opened 2024-12-28
Process SkinData asynchronously

Blackjack200:
### Summary

https://github.com/pmmp/PocketMine-MP/blob/fbaa125d0ce21ffef98fc1630881a92bedfbaa73/src/entity/Skin.php#L69-L74

This pull request optimizes the processing of player geometry data by offloading skin decoding to a `ProcessSkinTask`. This change addresses potential server freezes caused by the decoding of overly complex JSON structures.

### Existing Problems
Decoding geometry data directly in the main thread was a time-intensive operation, especially when the JSON data was highly complex, leading to server performance issues or even temporary freezes.

- **Before patch**: [Timings report](https://timings.pmmp.io/?id=339888)  
- **After patch**: [Timings report](https://timings.pmmp.io/?id=339887)  

### Changes
- Offloaded the skin geometry decoding to an asynchronous task (`ProcessSkinTask`) to prevent the main thread from being froze.

### Behavioural Changes
- Players' skin geometry data will now be processed asynchronously.
- Reduced risk of server performance degradation during skin decoding.

### Backwards Compatibility
This change is fully backward compatible, as it does not modify the API or the format of geometry data. Existing player skins and data handling will continue to work without modification.

### Follow-up
- Monitor server performance post-deployment to ensure stability.

### Tests
- Verified that skin geometry data is correctly decoded when processed via `ProcessSkinTask`.
- Tested gameplay to confirm no adverse effects on player skins or entity rendering.
- Compared server performance before and after the patch, as reflected in the timings reports.

### DIFF
--- a/src/network/mcpe/ProcessSkinTask.php
+++ b/src/network/mcpe/ProcessSkinTask.php
@@ -31,7 +31,7 @@ use pocketmine\scheduler\AsyncTask;
 use pocketmine\thread\NonThreadSafeValue;
 
 class ProcessSkinTask extends AsyncTask{
-	private const KEY_ON_COMPLETION = "onCompletion";
+	private const TLS_KEY_ON_COMPLETION = "onCompletion";
 
 	private ?string $error = "Unknown";
 	/** @var NonThreadSafeValue<SkinData> */
@@ -45,7 +45,7 @@ class ProcessSkinTask extends AsyncTask{
 		\Closure $callback,
 	){
 		$this->skinData = new NonThreadSafeValue($skinData);
-		$this->storeLocal(self::KEY_ON_COMPLETION, $callback);
+		$this->storeLocal(self::TLS_KEY_ON_COMPLETION, $callback);
 	}
 
 	public function onRun() : void{
@@ -62,7 +62,7 @@ class ProcessSkinTask extends AsyncTask{
 		/** @var Skin|null $result */
 		$result = $this->getResult();
 		/** @var \Closure(?Skin $skin,?string $error) : void $callback */
-		$callback = $this->fetchLocal(self::KEY_ON_COMPLETION);
+		$callback = $this->fetchLocal(self::TLS_KEY_ON_COMPLETION);
 		($callback)($result, $this->error);
 	}
 }
--- a/src/network/mcpe/ProcessSkinTask.php
+++ b/src/network/mcpe/ProcessSkinTask.php
@@ -59,10 +59,8 @@ class ProcessSkinTask extends AsyncTask{
 	}
 
 	public function onCompletion() : void{
+		/** @var Skin|null $result */
 		$result = $this->getResult();
-		if(!($result instanceof Skin)){
-			$result = null;
-		}
 		/** @var \Closure(?Skin $skin,?string $error) : void $callback */
 		$callback = $this->fetchLocal(self::KEY_ON_COMPLETION);
 		($callback)($result, $this->error);
--- a/src/network/mcpe/ProcessSkinTask.php
+++ b/src/network/mcpe/ProcessSkinTask.php
@@ -63,6 +63,6 @@ class ProcessSkinTask extends AsyncTask{
 		$result = $this->getResult();
 		/** @var \Closure(?Skin $skin,?string $error) : void $callback */
 		$callback = $this->fetchLocal(self::TLS_KEY_ON_COMPLETION);
-		($callback)($result, $this->error);
+		$callback($result, $this->error);
 	}
 }
--- a/src/network/mcpe/ProcessSkinTask.php
+++ b/src/network/mcpe/ProcessSkinTask.php
@@ -1,0 +1,70 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * @author PocketMine Team
+ * @link http://www.pocketmine.net/
+ *
+ *
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\network\mcpe;
+
+use pocketmine\entity\InvalidSkinException;
+use pocketmine\entity\Skin;
+use pocketmine\network\mcpe\convert\TypeConverter;
+use pocketmine\network\mcpe\protocol\types\skin\SkinData;
+use pocketmine\scheduler\AsyncTask;
+use pocketmine\thread\NonThreadSafeValue;
+
+class ProcessSkinTask extends AsyncTask{
+	private const KEY_ON_COMPLETION = "onCompletion";
+
+	private ?string $error = "Unknown";
+	/** @var NonThreadSafeValue<SkinData> */
+	private NonThreadSafeValue $skinData;
+
+	/**
+	 * @param \Closure(?Skin $skin,?string $error) : void $callback
+	 */
+	public function __construct(
+		SkinData $skinData,
+		\Closure $callback,
+	){
+		$this->skinData = new NonThreadSafeValue($skinData);
+		$this->storeLocal(self::KEY_ON_COMPLETION, $callback);
+	}
+
+	public function onRun() : void{
+		try{
+			$skin = TypeConverter::getInstance()->getSkinAdapter()->fromSkinData($this->skinData->deserialize());
+			$this->setResult($skin);
+			$this->error = null;
+		}catch(\InvalidArgumentException|InvalidSkinException $e){
+			$this->error = $e->getMessage();
+		}
+	}
+
+	public function onCompletion() : void{
+		$result = $this->getResult();
+		if(!($result instanceof Skin)){
+			$result = null;
+		}
+		/** @var \Closure(?Skin $skin,?string $error) : void $callback */
+		$callback = $this->fetchLocal(self::KEY_ON_COMPLETION);
+		($callback)($result, $this->error);
+	}
+}
--- a/src/network/mcpe/handler/InGamePacketHandler.php
+++ b/src/network/mcpe/handler/InGamePacketHandler.php
@@ -848,6 +848,9 @@ class InGamePacketHandler extends PacketHandler{
 		$this->player->getServer()->getAsyncPool()->submitTask(new ProcessSkinTask(
 			$packet->skin,
 			function(?Skin $skin, ?string $error) use ($packet) : void{
+				if(!$this->session->isConnected()){
+					return;
+				}
 				if($error !== null){
 					$this->session->disconnectWithError(
 						reason: "Invalid skin: " . $error,
--- a/src/network/mcpe/handler/InGamePacketHandler.php
+++ b/src/network/mcpe/handler/InGamePacketHandler.php
@@ -29,7 +29,7 @@ use pocketmine\block\tile\Sign;
 use pocketmine\block\utils\SignText;
 use pocketmine\entity\animation\ConsumingItemAnimation;
 use pocketmine\entity\Attribute;
-use pocketmine\entity\InvalidSkinException;
+use pocketmine\entity\Skin;
 use pocketmine\event\player\PlayerEditBookEvent;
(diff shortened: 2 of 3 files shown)


src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -843,12 +845,23 @@ public function handlePlayerSkin(PlayerSkinPacket $packet) : bool{
...
 		$this->session->getLogger()->debug("Processing skin change request");
-		try{
-			$skin = $this->session->getTypeConverter()->getSkinAdapter()->fromSkinData($packet->skin);
-		}catch(InvalidSkinException $e){
-			throw PacketHandlingException::wrap($e, "Invalid skin in PlayerSkinPacket");
-		}
-		return $this->player->changeSkin($skin, $packet->newSkinName, $packet->oldSkinName);
+		$this->player->getServer()->getAsyncPool()->submitTask(new ProcessSkinTask(
+			$packet->skin,
+			function(?Skin $skin, ?string $error) use ($packet) : void{
```
dktapps: The player may not be connected anymore, or they may have changed to another different skin before the task finished

src/network/mcpe/handler/LoginPacketHandler.php
```diff
@@ -70,19 +72,33 @@ public function handleLogin(LoginPacket $packet) : bool{
...
-		try{
-			$skin = $this->session->getTypeConverter()->getSkinAdapter()->fromSkinData(ClientDataToSkinDataHelper::fromClientData($clientData));
-		}catch(\InvalidArgumentException | InvalidSkinException $e){
-			$this->session->disconnectWithError(
-				reason: "Invalid skin: " . $e->getMessage(),
-				disconnectScreenMessage: KnownTranslationFactory::disconnectionScreen_invalidSkin()
-			);
+		$this->server->getAsyncPool()->submitTask(new ProcessSkinTask(
+			ClientDataToSkinDataHelper::fromClientData($clientData),
+			function(?Skin $skin, ?string $error) use ($packet, $clientData, $extraData){
```
dktapps, line 85: Session might not be connected anymore here

dktapps (changes requested):
This is probably a good idea. Only other way we could bring down this cost is if there was some faster way to strip and validate the JSON, which as far as I'm aware there isn't.

However, there's some potential issues that need to be addressed.

src/network/mcpe/handler/LoginPacketHandler.php
Blackjack200:
```php
if(!$this->session->isConnected()){
	return;
}
```

is added for connection state check now

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -843,12 +845,26 @@ public function handlePlayerSkin(PlayerSkinPacket $packet) : bool{
...
+				if(!$this->session->isConnected()){
```
dktapps, line 851: This should also check if the old skin is the same as before the task, so it doesn't overwrite if multiple tasks run at the same time

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -859,6 +859,10 @@ function(?Skin $skin, ?string $error) use ($packet) : void{
 					return;
 				}
 				if($skin !== null){
+					if ($currentFullSkinId !== $this->lastRequestedFullSkinId) {
```
dktapps:
```suggestion
					if($currentFullSkinId !== $this->lastRequestedFullSkinId){
```

dktapps (dismissed):
@pmmp/server-developers anyone else want to cast an eye over this?

src/network/mcpe/ProcessSkinTask.php
```diff
@@ -0,0 +1,70 @@
...
+			$this->error = null;
+		}catch(\InvalidArgumentException|InvalidSkinException $e){
+			$this->error = $e->getMessage();
+		}
+	}
+
+	public function onCompletion() : void{
+		$result = $this->getResult();
+		if(!($result instanceof Skin)){
+			$result = null;
```
dktapps: Not sure I understand the purpose of this check. It's never going to be anything but `Skin|null`. A `@var` would be fine.

(Yes, I know I usually say don't use `@var`, but in this specific case it's the norm because `AsyncTask` is poorly designed.)

src/network/mcpe/handler/LoginPacketHandler.php
```diff
@@ -70,19 +72,36 @@ public function handleLogin(LoginPacket $packet) : bool{
...
 
-		try{
-			$skin = $this->session->getTypeConverter()->getSkinAdapter()->fromSkinData(ClientDataToSkinDataHelper::fromClientData($clientData));
-		}catch(\InvalidArgumentException | InvalidSkinException $e){
-			$this->session->disconnectWithError(
-				reason: "Invalid skin: " . $e->getMessage(),
-				disconnectScreenMessage: KnownTranslationFactory::disconnectionScreen_invalidSkin()
-			);
+		$this->server->getAsyncPool()->submitTask(new ProcessSkinTask(
+			ClientDataToSkinDataHelper::fromClientData($clientData),
```
ipad54: This doesn’t cover the case when base64 data is invalid. See https://github.com/pmmp/BedrockProtocol/blob/f5daac3c11b7a1f5ba2437ecc09f9ad5ebf72aa8/src/types/login/ClientDataToSkinDataHelper.php#L33

src/network/mcpe/ProcessSkinTask.php
```diff
@@ -0,0 +1,68 @@
...
+
+use pocketmine\entity\InvalidSkinException;
+use pocketmine\entity\Skin;
+use pocketmine\network\mcpe\convert\TypeConverter;
+use pocketmine\network\mcpe\protocol\types\skin\SkinData;
+use pocketmine\scheduler\AsyncTask;
+use pocketmine\thread\NonThreadSafeValue;
+
+class ProcessSkinTask extends AsyncTask{
+	private const KEY_ON_COMPLETION = "onCompletion";
```
ipad54: @dktapps in other async tasks all keys have `TLS` prefix. Is this a convention or something?

ipad54 (changes requested):

src/network/mcpe/ProcessSkinTask.php
dktapps: Yeah, it's not required, just a style thing

src/network/mcpe/ProcessSkinTask.php
```diff
@@ -0,0 +1,68 @@
...
+			$this->error = $e->getMessage();
+		}
+	}
+
+	public function onCompletion() : void{
+		/** @var Skin|null $result */
+		$result = $this->getResult();
+		/** @var \Closure(?Skin $skin,?string $error) : void $callback */
+		$callback = $this->fetchLocal(self::TLS_KEY_ON_COMPLETION);
+		($callback)($result, $this->error);
```
dktapps:
```suggestion
		$callback($result, $this->error);
```

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -840,15 +842,33 @@ public function handlePlayerSkin(PlayerSkinPacket $packet) : bool{
...
+		$currentFullSkinId = $this->lastRequestedFullSkinId = $packet->skin->getFullSkinId();
 
 		$this->session->getLogger()->debug("Processing skin change request");
-		try{
-			$skin = $this->session->getTypeConverter()->getSkinAdapter()->fromSkinData($packet->skin);
-		}catch(InvalidSkinException $e){
-			throw PacketHandlingException::wrap($e, "Invalid skin in PlayerSkinPacket");
-		}
-		return $this->player->changeSkin($skin, $packet->newSkinName, $packet->oldSkinName);
+		$this->player->getServer()->getAsyncPool()->submitTask(new ProcessSkinTask(
```
ipad54, line 848: This would probably make the server to freeze / crash if a malicious client floods tons of skin change requests per tick.

Blackjack200: Would the existing `$this->gamePacketLimiter` be sufficient for skin change packets, or would it be better to implement a separate rate limiter specifically for `PlayerSkinPacket` to prevent potential flooding?

Blackjack200: If a separate rate limiter is needed, how should we set the values for `INCOMING_SKIN_PACKET_PER_TICK = 1` and `INCOMING_SKIN_PACKET_BUFFER_TICKS = 100`?
Would these settings be appropriate, or should we adjust them further?

dktapps: > Would the existing `$this->gamePacketLimiter` be sufficient for skin change packets, or would it be better to implement a separate rate limiter specifically for `PlayerSkinPacket` to prevent potential flooding?

Probably not. It's supposed to limit all packets, so it has quite generous limits.

More so the issue here is memory overload if the client spams a lot of skin change packets (it could cause a large backlog in the worker pool & hog a lot of memory).

My thought was to limit the number of concurrent `ProcessSkinTask`s per session, but I'm not sure if this is the best way.

Blackjack200: Limiting the rate of `PlayerSkinPacket` packets seems easier to implement and would have a similar effect to limiting the number of concurrent `ProcessSkinTask`.

In my preliminary testing, setting `INCOMING_SKIN_PACKET_PER_TICK = 0.1` and `INCOMING_SKIN_PACKET_BUFFER_TICKS = 20` seems promising (though it still needs sufficient testing). However, this would require changing the `$averagePerTick` in `PacketLimiter` from an `int` to a `float` or both. Would that be acceptable?

dktapps: No, I don't think so. The number of packets isn't the only factor we need to consider. We also have to be aware of how long each packet takes to process.

Under normal conditions the packets might take 10 ms each to process, but an attacker could craft packets that take 10-100x longer to process and still be subject to the same packets-per-tick limits (getting more lag per packet). That's why I said we probably need to limit the number of tasks, rather than the number of packets.

CicciogamerRrr:
Bro stop spamming idk wtff you talking abttt

Il dom 29 dic 2024, 16:01 Dylan T. ***@***.***> ha scritto:

dktapps:
#4981 would also enable skipping some validation, e.g. if we had a map<rawData, validatedOptimizedData>
However this would only be helpful for dealing with common geometries

dktapps:
This PR would be better if we could capture `PlayerSkinPacket` immediately after `DataPacketDecodeEvent` and have it handed off to an async task in its plain buffer form.

This would:
- eliminate main thread skin decoding performance cost entirely
- avoid unnecessary serialize cost of `SkinData` (since it would never exist on the main thread)

However, it would break plugins that use `DataPacketReceiveEvent` to capture the packet.

On an unrelated note, worth mentioning (this only just occurred to me this second) is that this change would be disruptive for any plugins which set custom `SkinAdapter`s in `TypeConverter`, as these won't be configured on the other threads.

src/network/mcpe/handler/InGamePacketHandler.php
Blackjack200: > No, I don't think so. The number of packets isn't the only factor we need to consider. We also have to be aware of how long each packet takes to process.
> 
> 
> 
> Under normal conditions the packets might take 10 ms each to process, but an attacker could craft packets that take 10-100x longer to process and still be subject to the same packets-per-tick limits (getting more lag per packet). That's why I said we probably need to limit the number of tasks, rather than the number of packets.

Aside from the better methods for now, we could consider handling these potential time-bomb data by spawning new processes, similar to ConsoleReaderChildProcess. This would allow the main thread to control their runtime, using OS-level process killing (with proper abstraction for generalization—I’ve already implemented this in my private project, leveraging opis/closure). However, this approach does come with challenging compatibility issues.

dktapps: That sounds overengineered to me.

Blackjack200:
> This PR would be better if we could capture `PlayerSkinPacket` immediately after `DataPacketDecodeEvent` and have it handed off to an async task in its plain buffer form.
> 
> This would:
> 
> * eliminate main thread skin decoding performance cost entirely
> * avoid unnecessary serialize cost of `SkinData` (since it would never exist on the main thread)
> 
> However, it would break plugins that use `DataPacketReceiveEvent` to capture the packet.
> 
> On an unrelated note, worth mentioning (this only just occurred to me this second) is that this change would be disruptive for any plugins which set custom `SkinAdapter`s in `TypeConverter`, as these won't be configured on the other threads.

Therefore, we could preprocess the data packets in other threads first and then pass them to the main thread for processing. This seems to be a more common approach, rather than focusing solely on skin processing, we could also implement asynchronous encryption and asynchronous packets decoding/encoding(assuming the cost of serialize/unserialize is lower than encode/decking them) on the basis.

Also packets’ life cycle is clear so will there a possible way to transfer the ownership of objects between threads? This could improve this process significantly.

dktapps:
Don't overengineer it. That's well outside the scope of the PR and not practical anyway.

Blackjack200:
I have no idea on a proper implementation approach. I think this should be implemented by a plugin.

PR #6596, opened 2025-01-17
Noteblock instrument changes from 1.21.50

dries-c:
In 1.21.50 Mojang seems to have swapped some instruments for the sake of it.


dktapps (approved):
Good thing we're not saving these on disk. Begs the question what it's doing in this repo though.

PR #5953, opened 2023-07-31
Implement enchanting system

S3v3Nice:
## Introduction

This adds the long-awaited functionality of enchanting items via an enchanting table.

### API changes
- Add to Human: `getEnchantmentSeed()`, `setEnchantmentSeed(int)`, `generateEnchantmentSeed()`
- ItemFlags is now deprecated
- `getPrimaryItemFlags`, `getSecondaryItemFlags`, `hasPrimaryItemType`, `hasSecondaryItemType` in Enchantment are now deprecated
- Add ItemEnchantmentTags
- Add ItemEnchantmentTagRegistry
- Add to Item: `getEnchantmentTags()`, `getEnchantability()`
- Add to BlockTypeInfo: `getEnchantmentTags()` (some block items, such as carved pumpkin and mob heads, can be enchanted (with enchanted books in the anvil when the appropriate functionality is added))
- Add to Block: `getEnchantmentTags()`
- Add to EnchantInventory: `getInput()`, `getLapis()`, `getOutput(int)`, `getOption(int)`
- Add EnchantmentHelper
- Add to Enchantment: `isCompatibleWith(Enchantment)`, `getMinEnchantingPower(int)`, `getMaxEnchantingPower(int)`
- Add IncompatibleEnchantmentGroups
- Add IncompatibleEnchantmentRegistry
- Add AvailableEnchantmentRegistry
- Add EnchantTransaction
- Add EnchantedBook item
- Add ArmorMaterial (to hold the value of `enchantability` for each armor material)
- Add VanillaArmorMaterials registry
- Add to ArmorTypeInfo: `getMaterial()`
- Add to Armor: `getMaterial()`
- Add to ToolTier: `getEnchantability()`
- Add EnchantOption
- Add to InventoryManager: `syncEnchantingTableOptions`, `getEnchantingTableOptionIndex`
- Add PlayerEnchantOptionsRequestEvent
- Add PlayerItemEnchantEvent

### Behavioural changes
Add enchanting functionality for an enchanting table.

## Follow-up
- Maybe do something with CraftRecipeAutoStackRequestAction and CraftingConsumeInputStackRequestAction (these actions are now used for both crafting and enchanting, and in the future for anvil repair by idea).
[...]

PR #6600, opened 2025-01-24
Adding item flags in VanillaEnchantments

XGDavid:
why this isnt already used?


dktapps:
It was deprecated by #5953 and will be removed in 6.0

Issue #6604, opened 2025-01-27
A faster way to check if a block's AABBs can exceed its 1x1x1 cell


dktapps:
### Problem description

https://github.com/pmmp/PocketMine-MP/blob/f60120a75ea8e80fadbedfad0148da10ea29c91f/src/world/World.php#L1570-L1578

This code multiplies the performance impact of block AABB collision checking (used for entity movement).
It's only needed to ensure blocks with oversized AABBs like fences are correctly accounted for during collision checks.
However, these blocks are the exception and not the norm, so paying a performance penalty on _all_ blocks is pretty unpleasant.

### Proposed solution

If we had a way to tell from the block's stateID whether its AABB may exceed the cell boundaries, we could avoid a lot of unnecessary work in this function. Unfortunately I don't see a nice way to implement this without BC breaks.

Blocks currently aren't required to declare this information, so we'd have to infer it by looking at the results of `getCollisionBoxes()` for each state. This is OK since `recalculateCollisionBoxes()` is supposed to only be influenced by state anyway (not coordinates).

However, AABB overflow can also be influenced by stuff like `getModelPositionOffset()`, which is currently not restricted in any way.

Options are:
- `Block::mayExceedCellBounds()` which returns `true` by default, overridden to `false` for most PM blocks (boilerplaty but preserves BC)
- `RuntimeBlockStateRegistry->oversizedAABB` (`array<stateID, bool>`), which would need reflection to check whether `getModelPositionOffset()` is overridden (shitty hack but little code needed)

Issue #6551, opened 2024-12-01
`Position` should not be allowed to have a null World (remove `Position` from `Block`)


dktapps:
## Problem description
This is bizarre and unexpected.

## Proposed solution
It's only needed because `Block`s contain `Position`s that need to initialize with _something_ when they aren't in a world. Really, this is an issue about removing `Position` from `Block` and having it injected via function parameters where it's needed, because the current design has resulted in a multitude of issues.

However, removing `Position` from `Block` is a huge change that I've unsuccessfully tried to do several times before. It would involve:
- Everything that uses `Block` as `Position` to be updated
- All plugin APIs that expose a position via `Block` would have to be changed
- New APIs on events to include positions separated from blocks
- Many BC breaks within the `Block` package to pass `World` and `Vector3` params

dktapps:
Turns out `Position` in `Block` is also a significant factor in the poor performance of GC. Since Position ref World ref Server, and Server is an expensive object to do GC on. These objects cause Server to get frequently scanned for GC.

Issue #6605, opened 2025-01-27
Require blocks to declare that they need dynamic state recalculation on read


dktapps:
### Problem description

https://github.com/pmmp/PocketMine-MP/blob/f60120a75ea8e80fadbedfad0148da10ea29c91f/src/world/World.php#L1934-L1940

This could be avoided if we knew in advance whether the block has any dynamic state info (typically calculated in `readStateFromWorld()`).
It would also allow fixing #6115 since we could just not cache AABBs for blocks known to have dynamic state info.

### Proposed solution

Store a cached `array<stateID, true>` in `RuntimeBlockStateRegistry` that signifies whether dynamic state recalculation is needed for that state ID.

Currently this would have to be achieved using reflection, e.g. `(new ReflectionFunction($block->readStateFromWorld(...))->getClosureScopeClass()->getName() !== Block::class`. Shitty hack but I don't see a better way currently. The performance gains should be worth it in any case.

PR #6606, opened 2025-01-27
Optimise collision box checks by caching some basic info

dktapps:
This PR significantly improves performance of entity movement calculation.

Previous attempts to optimise this were ineffective, as they used a cache to mitigate the cost of recomputing AABBs. Entities tend to move around randomly, so the non-cached pathway really needed to be optimized.

This change improves performance on multiple fronts:
1) avoiding Block allocations for blocks with 1x1x1 AABBs and with no AABBs (the most common)
2) avoiding Block allocations and overlapping intersection checks unless a stateID is specifically known to potentially exceed its cell boundaries (like fences)
3) avoiding overlapping AABB checks when overlaps can't make any difference anyway (cubes)

Together, these changes improve the performance of item entity movement (non-cached) by about 200%. We may be able to get further performance gains, but this is a good start.

### Related issues & PRs
- Fixes (hackily) #6604

### API changes
Added public `RuntimeBlockStateRegistry->collisionInfo` and related constants

### Behavioural changes
This change is built on the assumption that `Block->recalculateCollisionBoxes()` and its overrides don't interact with any world. This is technically possible due to the crappy design of the `Block` architecture, but should be avoided. As a world is not available during `RuntimeBlockStateRegistry` initialization, attempting to interact with a world during `recalculateCollisionBoxes()` will now cause a crash.

This turned out to be a problem for `ChorusPlant`, which was fixed by 70fb9bbdfd06c7eda00b4cd2c2c3840755e6b8f6. The correct solution in this case was to use dynamic states similar to how we currently deal with fence connections.

As a result of this change, it's possible plugins implementing custom blocks may break if their blocks' AABB calculations depend on `$this->position`. However, I don't expect this to be an issue, and the problem will become quickly apparent during startup anyway.

Given the massive performance advantage on offer from this change, I think this is an OK risk to take.

## Backwards compatibility
See above section on behavioural changes.

## Follow-up
We may be able to get further improvements by tracking more info about oversized blocks, but this is a good enough start.

In the future, it might be worth requiring blocks to explicitly declare something like `Block->mayExceedCellBounds() : bool` when their AABBs could go outside the cell bounds, and have this enforced, throwing errors if `recalculateCollisionBoxes()` ever generates out-of-bounds AABBs. However, this would be tricky to implement due to the headaches presented by `getModelPositionOffset()`, which can currently offset an AABB outside of the block's cell.

We also really, really need to implement #6551. It should never have been possible to implement position-dependent logic in `recalculateCollisionBoxes`. This problem has appeared in a variety of issues by now.

## Tests
Tested in-game with item entities. My local server started to lag without the change at around 400 dropped items; with the change, around 1200.

### DIFF
--- a/src/block/RuntimeBlockStateRegistry.php
+++ b/src/block/RuntimeBlockStateRegistry.php
@@ -113,6 +113,70 @@ class RuntimeBlockStateRegistry{
 		}
 	}
 
+	/**
+	 * Checks if the given class method overrides a method in Block.
+	 * Used to determine if a block might need to disable fast path optimizations.
+	 *
+	 * @phpstan-param anyClosure $closure
+	 */
+	private static function overridesBlockMethod(\Closure $closure) : bool{
+		$declarer = (new \ReflectionFunction($closure))->getClosureScopeClass();
+		return $declarer !== null && $declarer->getName() !== Block::class;
+	}
+
+	/**
+	 * A big ugly hack to set up fast paths for handling collisions on blocks with common shapes.
+	 * The information returned here is stored in RuntimeBlockStateRegistry->collisionInfo, and is used during entity
+	 * collision box calculations to avoid complex logic and unnecessary block object allocations.
+	 * This hack allows significant performance improvements.
+	 *
+	 * TODO: We'll want to redesign block collision box handling and block shapes in the future, but that's a job for a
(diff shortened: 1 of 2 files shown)


dktapps:
Linking #6605 since dynamic states are also a confounding factor - anything which lazy-updates on read may also have unusual AABBs

PR #6619, opened 2025-02-04
Added ArrayAccess implementation

HardlinePeak:
I think that many, like me, will find it more convenient to work with array elements than with object properties, so I propose adding the ArrayAccess implementation (of course, leaving the magic methods __get, __set, __isset and __unset).
```php
<?php
$config = new Config("example.json");
$config->set("abc", 123);
$config->abc = 321;
$config["abc"] = "What do you say?";
```

### DIFF
--- a/src/utils/Config.php
+++ b/src/utils/Config.php
@@ -23,6 +23,7 @@ declare(strict_types=1);
 
 namespace pocketmine\utils;
 
+use ArrayAccess;
 use pocketmine\errorhandler\ErrorToExceptionHandler;
 use Symfony\Component\Filesystem\Path;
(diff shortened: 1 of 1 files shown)


dktapps:
Why does everyone want to mess with Config all of a sudden...?

We're planning to get rid of it in PM6 (see #3766). If you want to work on arrays directly, just use Config->getAll() (or don't use Config to begin with).

PS: Don't put unrelated changes in a PR if you want to get it merged. That means don't change unrelated typehints or functions.

HardlinePeak:
I admit, I should have found out about future plans. Thanks for the answer)

PR #6627, opened 2025-02-15
Bedrock 1.21.60

dries-c:
This PR updates PM5 to 1.21.60

### API changes

Adds a group argument to adding items to the creative inventory

## Backwards compatibility

See https://github.com/pmmp/BedrockProtocol/pull/294


src/inventory/CreativeInventory.php
```diff
@@ -75,15 +101,27 @@ public function clear() : void{
...
 	public function getAll() : array{
-		return Utils::cloneObjectArray($this->creative);
+		return Utils::cloneObjectArray($this->items);
+	}
+
+	/**
+	 * @return CreativeGroup[]
+	 * @phpstan-return array<int, CreativeGroup>
+	 */
+	public function getItemGroup() : array{
```
dktapps:
```suggestion
	public function getItemGroups() : array{
```

src/inventory/CreativeInventory.php
```diff
@@ -75,15 +101,27 @@ public function clear() : void{
 	 * @phpstan-return array<int, Item>
 	 */
 	public function getAll() : array{
-		return Utils::cloneObjectArray($this->creative);
+		return Utils::cloneObjectArray($this->items);
+	}
+
+	/**
+	 * @return CreativeGroup[]
+	 * @phpstan-return array<int, CreativeGroup>
```
dktapps: This return value should have a description, it's not obvious how this works from the type alone

src/inventory/CreativeInventory.php
```diff
@@ -24,40 +24,65 @@
...
+		] as $category => $categoryId){
+			$groups = CraftingManagerFromDataHelper::loadJsonArrayOfObjectsFile(
+				Path::join(\pocketmine\BEDROCK_DATA_PATH, "creative", $category . ".json"),
+				CreativeGroupData::class
+			);
+
+			foreach($groups as $groupData){
+				$group = new CreativeGroup(
+					$categoryId,
+					$groupData->group_name,
```
dktapps: This should use `Translatable`, we don't work on translation keys as plain strings anymore

src/inventory/CreativeInventory.php
```diff
@@ -75,15 +101,27 @@ public function clear() : void{
...
+	public function getItemGroup() : array{
+		return $this->groups;
 	}
 
 	public function getItem(int $index) : ?Item{
-		return isset($this->creative[$index]) ? clone $this->creative[$index] : null;
+		return isset($this->items[$index]) ? clone $this->items[$index] : null;
+	}
+
+	public function getGroup(int $index) : ?CreativeGroup{
```
dktapps: `getItemGroupByIndex` would be a better name. A doc comment with a `@see CreativeInventory::getItemIndex()` would probably be a good idea too.

src/inventory/CreativeInventory.php
```diff
@@ -96,9 +134,14 @@ public function getItemIndex(Item $item) : int{
...
 	 */
-	public function add(Item $item) : void{
-		$this->creative[] = clone $item;
+	public function add(Item $item, ?CreativeGroup $group = null) : void{
+		$this->items[] = $item;
+		$this->groups[] = $group ?? $this->defaultGroup;
 		$this->onContentChange();
+
+		if($group !== null){ // We need to create a new default group if another group is used.
+			$this->defaultGroup = CreativeGroup::anonymous(CreativeCategory::ITEMS);
```
dktapps: If several items are added with a defined group, this will create a useless group object for every call.
Better if `defaultGroup` is nullable, and then we only initialize it when it's needed.

src/inventory/data/CreativeGroup.php
```diff
@@ -0,0 +1,67 @@
...
+
+use pocketmine\inventory\CreativeCategory;
+use pocketmine\item\Item;
+
+final class CreativeGroup{
+	private CreativeCategory $categoryId;
+	private string $name;
+	private ?Item $icon;
+
+	public function __construct(CreativeCategory $categoryId, string $name, ?Item $icon = null){
```
dktapps: `Translatable|string` should probably be used (although that also depends whether plain non-translated strings are allowed by the client too)

Also not clear why $name is required but $icon is optional

src/inventory/data/CreativeGroup.php
```diff
@@ -0,0 +1,67 @@
...
+
+	public static function anonymous(CreativeCategory $categoryId) : self{
+		return new self($categoryId, "");
+	}
+
+	public function getCategoryId() : CreativeCategory{
+		return $this->categoryId;
+	}
+
+	public function setCategoryId(CreativeCategory $categoryId) : void{
```
dktapps: Don't like this

src/inventory/data/CreativeGroup.php
```diff
@@ -0,0 +1,67 @@
...
+		$this->categoryId = $categoryId;
+	}
+
+	public function getName() : string{
+		return $this->name;
+	}
+
+	public function setName(string $name) : void{
```
dktapps: Don't like this

src/inventory/data/CreativeGroup.php
```diff
@@ -0,0 +1,67 @@
...
+		$this->name = $name;
+	}
+
+	public function getIcon() : ?Item{
+		return $this->icon;
+	}
+
+	public function setIcon(?Item $icon) : void{
```
dktapps: Don't like this

src/inventory/data/CreativeGroup.php
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\inventory\data;
+
+use pocketmine\inventory\CreativeCategory;
+use pocketmine\item\Item;
+
+final class CreativeGroup{
```
dktapps: This whole class could be simplified down to a few lines using `readonly` properties and promoted constructor parameters

tools/generate-bedrock-data-from-packets.php
```diff
@@ -234,31 +253,75 @@ private static function sort(mixed $object) : mixed{
...
 	}
 
 	public function handleCreativeContent(CreativeContentPacket $packet) : bool{
 		echo "updating creative inventory data\n";
-		$items = array_map(function(CreativeContentEntry $entry) : array{
-			return self::objectToOrderedArray($this->itemStackToJson($entry->getItem()));
-		}, $packet->getEntries());
-		file_put_contents($this->bedrockDataPath . '/creativeitems.json', json_encode($items, JSON_PRETTY_PRINT) . "\n");
+
+		/** @var array<int, ItemStackData[]> $groupItems */
```
dktapps:
```suggestion
		/**
		 * @var ItemStackData[][] $groupItems
		 * @phpstan-var array<int, list<ItemStackData>> $groupItems
		 */
```

We don't use generics in general @var, they aren't part of the PSR standard

tools/generate-bedrock-data-from-packets.php
```diff
@@ -234,31 +253,75 @@ private static function sort(mixed $object) : mixed{
...
+		$groups = $packet->getGroups();
+
+		static $typeMap = [
+			CreativeContentPacket::CATEGORY_CONSTRUCTION => "construction",
+			CreativeContentPacket::CATEGORY_NATURE => "nature",
+			CreativeContentPacket::CATEGORY_EQUIPMENT => "equipment",
+			CreativeContentPacket::CATEGORY_ITEMS => "items",
+		];
+
+		/** @var array<int, CreativeGroupData[]> $groupCategories */
```
dktapps:
```suggestion
		/**
		 * @var CreativeGroupData[][] $groupCategories
		 * @phpstan-var array<int, list<CreativeGroupData>>
		 */
```
Same as above

Though the doc comment may not be needed at all, try deleting it and see what happens if you do `\PHPStan\dumpType()`

composer.json
```diff
@@ -33,10 +33,10 @@
       "composer-runtime-api": "^2.0",
       "adhocore/json-comment": "~1.2.0",
       "netresearch/jsonmapper": "~v5.0.0",
-      "pocketmine/bedrock-block-upgrade-schema": "~5.0.0+bedrock-1.21.40",
-      "pocketmine/bedrock-data": "~2.15.0+bedrock-1.21.50",
+      "pocketmine/bedrock-block-upgrade-schema": "~5.1.0+bedrock-1.21.60",
+      "pocketmine/bedrock-data": "dev-bedrock-1.21.60",
       "pocketmine/bedrock-item-upgrade-schema": "~1.14.0+bedrock-1.21.50",
-      "pocketmine/bedrock-protocol": "~35.0.0+bedrock-1.21.50",
+      "pocketmine/bedrock-protocol": "dev-bedrock-1.21.60",
```
dktapps: These should be normalized to a release version before merge

dktapps (changes requested):

src/inventory/data/CreativeGroup.php
dries-c: readonly classes are PHP 8.2 sadly, so i'll just do the properties for convenience ig

dries-c:
Screenshot 2025-02-15 at 14 54 48

We can't do normal translations here since we don't have the player's language. I personally don't think it's a good idea to scrap the entire caching, so how do we best deal with this.

src/inventory/data/CreativeContentCache.php
```diff
@@ -0,0 +1,80 @@
...
+use pocketmine\inventory\CreativeCategory;
+use pocketmine\lang\Translatable;
+use pocketmine\network\mcpe\NetworkSession;
+use pocketmine\network\mcpe\protocol\CreativeContentPacket;
+use pocketmine\network\mcpe\protocol\types\inventory\CreativeGroupEntry;
+use pocketmine\network\mcpe\protocol\types\inventory\CreativeItemEntry;
+use pocketmine\network\mcpe\protocol\types\inventory\ItemStack;
+use function array_reduce;
+
+final class CreativeContentCache{
```
dktapps: Not clear why we need a separate class for this.

dktapps (changes requested):

src/inventory/data/CreativeContentCache.php
dries-c: Returning and storing arrays in arrays is less clear

dktapps: I don't see why we need two arrays in the first place.

src/inventory/data/CreativeContentCache.php
```diff
@@ -0,0 +1,85 @@
...
+
+				if($group->name instanceof Translatable){
+					$player = $session->getPlayer();
+
+					if($player === null){
+						throw new \LogicException("Cannot send creative content for a player that is not yet created");
+					}
+
+					if(!$player->getServer()->isLanguageForced()){
+						[$message, ] = $player->getNetworkSession()->prepareClientTranslatableMessage($group->name);
```
iAmFrogger: Use `$session` instead of `$player->getNetworkSession()` lol

TwistedAsylumMC (dismissed):

PR #6628, opened 2025-02-15
Utils(docs): fix incorrect case referencing an internal constant value of Linux OS

Yexeed:
The "linux" constant reference has wrong case ("linux" !== "Linux"). This got me in a micro stutter when looking at the doc.

## Changes

This small change to a Utils::getOS() phpdoc doesn't seem to break anything, including BC or FC.

### DIFF
--- a/src/utils/Utils.php
+++ b/src/utils/Utils.php
@@ -269,7 +269,7 @@ final class Utils{
 	 * MacOS => mac
 	 * iOS => ios
 	 * Android => android
-	 * Linux => Linux
+	 * Linux => linux
 	 * BSD => bsd
 	 * Other => other
 	 */


dktapps:
The doc is correct: https://github.com/pmmp/PocketMine-MP/blob/5ef89200c683f3b06d54b2d1509400a0f95152f4/src/utils/Utils.php#L110

Yexeed:
> The doc is correct:
> 
> https://github.com/pmmp/PocketMine-MP/blob/5ef89200c683f3b06d54b2d1509400a0f95152f4/src/utils/Utils.php#L110

@dktapps, I don't think you catch the point.

The constant OS_WINDOWS = "win", which is ALSO (e.g. doesn't has any impact to the constant value) told by the Utils::getOS doc:

https://github.com/pmmp/PocketMine-MP/blob/5ef89200c683f3b06d54b2d1509400a0f95152f4/src/utils/Utils.php#L268

So when I've read the documentation of Utils::getOS, I get the knowledge that, if I have a runtime running on a Windows machine, I would get a "win" result from Utils:getOS. Which is correctly does what the doc says:
https://github.com/pmmp/PocketMine-MP/blob/5ef89200c683f3b06d54b2d1509400a0f95152f4/src/utils/Utils.php#L106

The documentation implies that 
`Windows machine => Utils::getOs() => Utils::OS_WINDOWS => "win"`
That is correct!

Now, documentation of Utils::getOs() have another "result mapping":
https://github.com/pmmp/PocketMine-MP/blob/5ef89200c683f3b06d54b2d1509400a0f95152f4/src/utils/Utils.php#L272

Thus, if I have a Linux machine, I would get a "Linux" (L uppercase) as a return value? No! :
https://github.com/pmmp/PocketMine-MP/blob/5ef89200c683f3b06d54b2d1509400a0f95152f4/src/utils/Utils.php#L110

The actual result chain would be:
`Linux machine => Utils::getOS() => Utils::OS_LINUX => "linux"` (L is lowercase here!)

But the doc misleads me to that Linux machine returns a "Linux", but the actual is "linux", and this is the problem which this PR was going to resolve.

dktapps:
Ah, seems like I misread the diff. You are correct.

Though I think it really ought to be referring to the OS_* constants instead of hardcoded strings.

dktapps:
I've addressed this separately. Thanks for bringing it to our attention :)

PR #6629, opened 2025-02-15
Unit test block hardness & blast resistance values

ipad54:
This PR adds a block hardness & blast resistance test by using [BedrockData block properties table](https://github.com/pmmp/BedrockData/blob/master/block_properties_table.json) to make sure the values are always up to date because of Mojang frequent changes.

## Tests
This has been briefly tested in the game. I didn't notice any behaviour changes with block breaking and explosions.


dktapps:
Good work on this so far.

I didn't realize how much this was going to suck for shared properties though. Since we have constants for every block, we're not able to share break infos between different blocks. That's made me hesitate with this.

ipad54:
Yes, but we don't actually know about Mojang plans, maybe they will add its own hardness to every block in the same group. Moreover, [I've found some TODOs about this](https://github.com/pmmp/PocketMine-MP/blob/9402a20ee3de0232761433565fc984953939c754/src/block/VanillaBlocks.php#L1275). The only thing I didn't solve yet is wood type hardness registration. It seems like we will have to return something like `WoodLikeBlockIdHelper` but with hardness & blast resistance, but for now all wood blocks use oak values until we discuss what to do about this.

dktapps:
> It seems like we will have to return something like WoodLikeBlockIdHelper but with hardness & blast resistance, but for now all wood blocks use oak values until we discuss what to do about this.

I really don't want to go down that road. I'm starting to see why you wanted to unit-test this instead of doing it this way.

ShockedPlot7560:
Good work
I hope for your mind that you've managed to find an easy way to replace the VanillaBlock :/
It's sad for wood, but adding another class with mapping doesn't seem like a very good solution, especially if we want to withdraw from this type of class.
Having one constant per block doesn't shock me that much. It allows more flexibility if mojang decides to change the hardness separately for groups of blocks, as ipad says.

ShockedPlot7560 (dismissed):
The ambition behind one constant per block seems OK to me. The question of wood seems to me to be one to be asked when Mojang has decided to separate the game, for later.

dktapps:
I thought this was a good idea originally, but after seeing the changes it requires I think it's a bad idea to go forward. I think we'd be better with a unit test to not have to add a bunch of noise to the code or use the constants in weird ways.

src/block/BlockBreakInfo.php
```diff
@@ -73,7 +73,7 @@ public static function instant(int $toolType = BlockToolType::NONE, int $toolHar
 		return new self(0.0, $toolType, $toolHarvestLevel, 0.0);
 	}
 
-	public static function indestructible(float $blastResistance = 18000000.0) : self{
+	public static function indestructible(float $blastResistance = 18000003.75) : self{
```
dktapps, line 76: ???

src/block/VanillaBlocks.php
```diff
@@ -1157,27 +1155,30 @@ public function isAffectedBySilkTouch() : bool{
...
+		self::register("cut_sandstone_slab", fn(BID $id) => new Slab($id, "Cut Sandstone", $stoneSlabBreakInfo));
+		self::register("mossy_cobblestone_slab", fn(BID $id) => new Slab($id, "Mossy Cobblestone", $stoneSlabBreakInfo));
+		self::register("purpur_slab", fn(BID $id) => new Slab($id, "Purpur", $stoneSlabBreakInfo));
+		self::register("smooth_red_sandstone_slab", fn(BID $id) => new Slab($id, "Smooth Red Sandstone", $stoneSlabBreakInfo));
+		self::register("smooth_quartz_slab", fn(BID $id) => new Slab($id, "Smooth Quartz", $stoneSlabBreakInfo));
+		self::register("stone_slab", fn(BID $id) => new Slab($id, "Stone", $stoneSlabBreakInfo));
+
+		self::register("end_stone_brick_slab", fn(BID $id) => new Slab($id, "End Stone Brick", new Info(BreakInfo::pickaxe(3.0, ToolTier::WOOD, 30.0))));
+
+		$stoneSlabBreakInfo = new Info(BreakInfo::pickaxe(1.5, ToolTier::WOOD, 30.0));
```
dktapps: This should use a different variable name, otherwise it's going to get confusing if we ever need to debug this

src/block/VanillaBlocks.php
```diff
@@ -1272,22 +1274,26 @@ public function getBreakTime(Item $item) : float{
...
-		self::register("diorite_wall", fn(BID $id) => new Wall($id, "Diorite Wall", $wallBreakInfo));
-		self::register("end_stone_brick_wall", fn(BID $id) => new Wall($id, "End Stone Brick Wall", $wallBreakInfo));
-		self::register("granite_wall", fn(BID $id) => new Wall($id, "Granite Wall", $wallBreakInfo));
-		self::register("mossy_stone_brick_wall", fn(BID $id) => new Wall($id, "Mossy Stone Brick Wall", $wallBreakInfo));
 		self::register("mossy_cobblestone_wall", fn(BID $id) => new Wall($id, "Mossy Cobblestone Wall", $wallBreakInfo));
 		self::register("nether_brick_wall", fn(BID $id) => new Wall($id, "Nether Brick Wall", $wallBreakInfo));
-		self::register("prismarine_wall", fn(BID $id) => new Wall($id, "Prismarine Wall", $wallBreakInfo));
 		self::register("red_nether_brick_wall", fn(BID $id) => new Wall($id, "Red Nether Brick Wall", $wallBreakInfo));
+
+		$wallBreakInfo = new Info(BreakInfo::pickaxe(1.5, ToolTier::WOOD, 30.0));
```
dktapps: same here

src/block/VanillaBlocks.php
```diff
@@ -1272,22 +1274,26 @@ public function getBreakTime(Item $item) : float{
...
+		self::register("stone_brick_wall", fn(BID $id) => new Wall($id, "Stone Brick Wall", $wallBreakInfo));
+		self::register("mossy_stone_brick_wall", fn(BID $id) => new Wall($id, "Mossy Stone Brick Wall", $wallBreakInfo));
+		self::register("granite_wall", fn(BID $id) => new Wall($id, "Granite Wall", $wallBreakInfo));
+		self::register("diorite_wall", fn(BID $id) => new Wall($id, "Diorite Wall", $wallBreakInfo));
+		self::register("andesite_wall", fn(BID $id) => new Wall($id, "Andesite Wall", $wallBreakInfo));
+		self::register("prismarine_wall", fn(BID $id) => new Wall($id, "Prismarine Wall", $wallBreakInfo));
+
+		$wallBreakInfo = new Info(BreakInfo::pickaxe(0.8, ToolTier::WOOD, 4.0));
```
dktapps: again

tests/phpunit/block/BlockTest.php
```diff
@@ -95,6 +101,55 @@ public function testLightFiltersValid() : void{
 		}
 	}
 
+	public function testBlockBreakInfo() : void{
+		$propertiesTable = json_decode(Filesystem::fileGetContents(BedrockDataFiles::BLOCK_PROPERTIES_TABLE_JSON), true, 3, JSON_THROW_ON_ERROR);
+		if(!is_array($propertiesTable)){
+			throw new AssumptionFailedError("Block properties table must be an array");
+		}
+		$exceptions = array_flip([
```
dktapps: Better use `array_fill_keys($ids, true)` for this

dktapps (changes requested):

src/block/BlockBreakInfo.php
ipad54: For some reason there is an inconsistency in blast resistance between indestructible blocks. All new indestructible blocks added since 1.2+ use 18000003.75, but older ones (bedrock, portal frames) use 1800000.0

src/block/VanillaBlocks.php
```diff
@@ -1157,28 +1155,31 @@ public function isAffectedBySilkTouch() : bool{
...
-		self::register("mossy_stone_brick_slab", fn(BID $id) => new Slab($id, "Mossy Stone Brick", $stoneSlabBreakInfo));
+		self::register("mossy_cobblestone_slab", fn(BID $id) => new Slab($id, "Mossy Cobblestone", $stoneSlabBreakInfo));
+		self::register("purpur_slab", fn(BID $id) => new Slab($id, "Purpur", $stoneSlabBreakInfo));
+		self::register("smooth_red_sandstone_slab", fn(BID $id) => new Slab($id, "Smooth Red Sandstone", $stoneSlabBreakInfo));
 		self::register("smooth_quartz_slab", fn(BID $id) => new Slab($id, "Smooth Quartz", $stoneSlabBreakInfo));
 		self::register("stone_slab", fn(BID $id) => new Slab($id, "Stone", $stoneSlabBreakInfo));
 
+		self::register("end_stone_brick_slab", fn(BID $id) => new Slab($id, "End Stone Brick", new Info(BreakInfo::pickaxe(3.0, ToolTier::WOOD, 30.0))));
+
+		$lightStoneSlabBreakInfo = new Info(BreakInfo::pickaxe(1.5, ToolTier::WOOD, 30.0));
```
ipad54, line 1171: This name might sound a bit odd, but I couldn't find any shared categories or signs with this group (like I found with walls). "light" means here that their hardness is less than that on the slabs group above.

src/block/VanillaBlocks.php
```diff
@@ -1006,16 +1004,16 @@ public function getBreakTime(Item $item) : float{
 		self::register("item_frame", fn(BID $id) => new ItemFrame($id, "Item Frame", $itemFrameInfo), TileItemFrame::class);
 		self::register("glowing_item_frame", fn(BID $id) => new ItemFrame($id, "Glow Item Frame", $itemFrameInfo), TileGlowingItemFrame::class);
 
-		self::register("jukebox", fn(BID $id) => new Jukebox($id, "Jukebox", new Info(BreakInfo::axe(0.8))), TileJukebox::class); //TODO: in PC the hardness is 2.0, not 0.8, unsure if this is a MCPE bug or not
+		self::register("jukebox", fn(BID $id) => new Jukebox($id, "Jukebox", new Info(BreakInfo::axe(2.0, blastResistance: 30.0))), TileJukebox::class); //TODO: in PC the hardness is 2.0, not 0.8, unsure if this is a MCPE bug or not
```
dktapps: This TODO comment can be removed

src/block/VanillaBlocks.php
```diff
@@ -1065,14 +1063,15 @@ public function getBreakTime(Item $item) : float{
...
 
 		$quartzBreakInfo = new Info(BreakInfo::pickaxe(0.8, ToolTier::WOOD));
+		$smoothQuartzBreakInfo = new Info(BreakInfo::pickaxe(2.0, ToolTier::WOOD, 30.0));
 		self::register("quartz", fn(BID $id) => new Opaque($id, "Quartz Block", $quartzBreakInfo));
 		self::register("chiseled_quartz", fn(BID $id) => new SimplePillar($id, "Chiseled Quartz Block", $quartzBreakInfo));
 		self::register("quartz_pillar", fn(BID $id) => new SimplePillar($id, "Quartz Pillar", $quartzBreakInfo));
-		self::register("smooth_quartz", fn(BID $id) => new Opaque($id, "Smooth Quartz Block", $quartzBreakInfo));
-		self::register("quartz_bricks", fn(BID $id) => new Opaque($id, "Quartz Bricks", $quartzBreakInfo));
+		self::register("smooth_quartz", fn(BID $id) => new Opaque($id, "Smooth Quartz Block", $smoothQuartzBreakInfo));
+		self::register("quartz_bricks", fn(BID $id) => new Opaque($id, "Quartz Bricks", new Info(BreakInfo::pickaxe(0.8, ToolTier::WOOD))));
```
dktapps: This can use $quartzBreakInfo

src/block/VanillaBlocks.php
```diff
@@ -1237,25 +1238,26 @@ public function getBreakTime(Item $item) : float{
...
 		self::register("cut_sandstone", fn(BID $id) => new Opaque($id, "Cut Sandstone", $sandstoneBreakInfo));
-		self::register("smooth_sandstone", fn(BID $id) => new Opaque($id, "Smooth Sandstone", $sandstoneBreakInfo));
+		self::register("smooth_sandstone", fn(BID $id) => new Opaque($id, "Smooth Sandstone", $smoothSandstoneBreakInfo));
 
 		self::register("glazed_terracotta", fn(BID $id) => new GlazedTerracotta($id, "Glazed Terracotta", new Info(BreakInfo::pickaxe(1.4, ToolTier::WOOD))));
 		self::register("dyed_shulker_box", fn(BID $id) => new DyedShulkerBox($id, "Dyed Shulker Box", $shulkerBoxBreakInfo), TileShulkerBox::class);
 		self::register("stained_glass", fn(BID $id) => new StainedGlass($id, "Stained Glass", $glassBreakInfo));
 		self::register("stained_glass_pane", fn(BID $id) => new StainedGlassPane($id, "Stained Glass Pane", $glassBreakInfo));
-		self::register("stained_clay", fn(BID $id) => new StainedHardenedClay($id, "Stained Clay", $hardenedClayBreakInfo));
+		self::register("stained_clay", fn(BID $id) => new StainedHardenedClay($id, "Stained Clay", new Info(BreakInfo::pickaxe(1.25, ToolTier::WOOD, 6.25))));
```
dktapps, line 1260: Are we sure this is correct? Seems weird for there to be a difference here

src/block/VanillaBlocks.php
```diff
@@ -1623,7 +1629,7 @@ public function isFireProofAsItem() : bool{ return true; }
 		self::register("soul_fire", fn(BID $id) => new SoulFire($id, "Soul Fire", new Info(BreakInfo::instant(), [Tags::FIRE])));
 
 		//TODO: soul soul ought to have 0.5 hardness (as per java) but it's 1.0 in Bedrock (probably parity bug)
```
dktapps: This can be removed

src/block/VanillaBlocks.php
ipad54: Yes. They have different blast resistance 
IMG_1442
IMG_1441

dktapps: Oh, I forgot they consider these terracotta now. So maybe this is more consistent with glazed terracotta, I guess it doesn't make sense for these to be more blast-resistant than that... though it still leaves the question why stained clay is weaker than non-stained clay??? Is this a documented difference?

ipad54: Wiki only says that terracotta has the same blast resistance as most other stone blocks. According to wiki, on Java they all have the same blast resistance though. Also, on Java non-stained clay is called terracotta too, while on bedrock it is still called hardened clay

dktapps: Confusing

ipad54: Yes, maybe it’s another bug or parity issue like with mangrove log…

dktapps (approved):

PR #6636, opened 2025-02-18
Repair the generate-bedrockdata-path-consts.php and server-phar.php files

TheColorDroid:


ShockedPlot7560:
There is no fix here

PR #6638, opened 2025-02-20
Always use `Permission` objects instead of `string`

### LINKED ISSUE #6536: Always use `Permission` objects instead of `string`
## Problem description
There are various places where strings have to be looked up as permissions before using them:

https://github.com/pmmp/PocketMine-MP/blob/8cdc7d7ee16482d630e49af2765dc958062b8693/src/permission/PermissibleInternal.php#L142
https://github.com/pmmp/PocketMine-MP/blob/8cdc7d7ee16482d630e49af2765dc958062b8693/src/permission/PermissibleInternal.php#L191
https://github.com/pmmp/PocketMine-MP/blob/8cdc7d7ee16482d630e49af2765dc958062b8693/src/plugin/PluginManager.php#L191
https://github.com/pmmp/PocketMine-MP/blob/8cdc7d7ee16482d630e49af2765dc958062b8693/src/command/Command.php#L106

This is problematic because it requires redundant checks for null when working with always-defined core permissions, and also allows passing in strings that can't be checked statically, allowing typos that only crash when the code is run.

## Proposed solution
In all of these places, we should just use `Permission` objects directly. Using strings presents the possibility that the permission may not exist or may not have been registered. (To be fair, using `Permission` objects also doesn't guarantee registration. We should probably fix that too.)

Referencing permissions by their string IDs should only be done when loading stuff from configs, like in a permission management plugin. The IDs are intended for user use, not code use. Ergo, `PermissionManager` should only be interacted with to look up a permission by its ID.

`DefaultPermissions` should turn into a `RegistryTrait` user (or perhaps an enum with metadata) to remove the need for `DefaultPermissionNames` (which should never have been needed to begin with).

DebHugo:
### Related issues & PRs
Fix #6536 

### API changes

- All functions thath use Permission now just accept `Permission` instead of `Permission|string`
- Remove `DefaultPermissionNames` (use `DefaultPermissions`)
- Now `DefaultPermissions` use `RegisteryTrait` -> contains all permissions added by pocketmine
## Backwards compatibility

BC Break (see API changes)

## Tests

Test Ingame with vanilla commands but without commands added by plugins


src/command/Command.php
```diff
@@ -93,30 +94,30 @@ public function getName() : string{
...
+		Utils::validateArrayValueType($permissions, function (Permission $permission) : void{});
 		$this->permission = $permissions;
 	}
 
-	public function setPermission(?string $permission) : void{
-		$this->setPermissions($permission === null ? [] : explode(";", $permission));
+	public function setPermission(?Permission $permission) : void{
+		$permissions = [];
+		if ($permission !== null){
+			$permissions = $this->permission;
```
dktapps: This is not correct, it should replace the existing permissions, not append

src/command/defaults/VanillaCommand.php
```diff
@@ -50,8 +51,8 @@ protected function fetchPermittedPlayerTarget(CommandSender $sender, ?string $ta
 			return null;
 		}
 		if(
-			($player === $sender && $this->testPermission($sender, $selfPermission)) ||
-			($player !== $sender && $this->testPermission($sender, $otherPermission))
+			($player === $sender && ($selfPermission === null || $this->testPermission($sender, $selfPermission))) ||
+			($player !== $sender && ($otherPermission === null || $this->testPermission($sender, $otherPermission)))
```
dktapps: Not clear why this is changed when testPermission() already handles it

src/permission/DefaultPermissions.php
```diff
@@ -24,95 +24,170 @@
...
 		foreach($grantedBy as $permission){
-			$permission->addChild($candidate->getName(), true);
+			$permission->addChild($candidate, true);
 		}
 		foreach($deniedBy as $permission){
-			$permission->addChild($candidate->getName(), false);
+			$permission->addChild($candidate, false);
 		}
 		PermissionManager::getInstance()->addPermission($candidate);
+		self::_registryRegister(str_replace(".", "_", $candidate->getName()), $candidate);
```
dktapps: I think we can do away with the `POCKETMINE` prefix here

src/permission/Permission.php
```diff
@@ -36,15 +36,17 @@ class Permission{
...
 	 */
+	private array $children = [];
+
+	/**
+	 * Creates a new Permission object to be attached to Permissible objects
+	 */
 	public function __construct(
 		private string $name,
 		Translatable|string|null $description = null,
-		private array $children = []
```
dktapps, line 47: Not clear why this is removed

src/permission/PermissionAttachment.php
```diff
@@ -81,29 +80,8 @@ public function clearPermissions() : void{
...
-
-	/**
-	 * @param string[] $permissions
-	 */
-	public function unsetPermissions(array $permissions) : void{
-		foreach($permissions as $node){
-			unset($this->permissions[$node]);
-		}
-		$this->recalculatePermissibles();
-	}
```
dktapps, line 103: Not sure why this is removed

src/plugin/PluginBase.php
```diff
@@ -174,7 +175,7 @@ private function registerYamlCommands() : void{
 
 			$newCmd->setAliases($aliasList);
 
-			$newCmd->setPermission($data->getPermission());
+			$newCmd->setPermission(PermissionManager::getInstance()->getPermission($data->getPermission()));
```
dktapps: This introduces a security hole, `PermissionManager::getPermission()` may return `null`

dktapps (changes requested):
After seeing the changes I'm less convinced of the value of doing this. The issue is that using `Permission` objects does not guarantee that the permission in question is actually registered.

src/permission/Permission.php
DebHugo: It's remove because currently array signature is `array<string, bool>` but now with Permission we can't put `array<Permission, bool>` so i remove this array and use `$this->addChild()` at replacement

src/plugin/PluginBase.php
DebHugo: No because this permission is already registered at https://github.com/pmmp/PocketMine-MP/blob/stable/src/plugin/PluginManager.php#L195

src/permission/PermissionAttachment.php
DebHugo: These are the same reasons as above. We remove string usage for Permission so we can't set/remove multiple Permissions just we they name, we need the Permission object. It's can be replace by for-each loop by call `setPermission`/`unsetPermission`

dktapps: `unsetPermissions()` isn't affected by this.

src/plugin/PluginBase.php
dktapps: The permission may not have been declared.

src/permission/PermissionAttachment.php
DebHugo: why ? I think it's illogic to remove usage of string but not here

src/plugin/PluginBase.php
DebHugo: so i need to create new Permission here ?

src/permission/PermissionAttachment.php
dktapps: There's no need to remove it since there's no technical reason to.

src/plugin/PluginBase.php
dktapps: No, it needs to check whether the permission exists before proceeding. Previously `setPermission()` would have checked it.

src/permission/DefaultPermissions.php
```diff
@@ -24,95 +24,172 @@
...
-abstract class DefaultPermissions{
-	public const ROOT_CONSOLE = Names::GROUP_CONSOLE;
-	public const ROOT_OPERATOR = Names::GROUP_OPERATOR;
-	public const ROOT_USER = Names::GROUP_USER;
+/**
+ * @method static Permission BROADCAST_ADMIN()
+ * @method static Permission BROADCAST_USER()
+ * @method static Permission COMMAND_BAN_IP()
+ * @method static Permission COMMAND_BAN_LIST()
+ * @method static Permission COMMAND_BAN_PLAYER()
```
ShockedPlot7560: This does not appear to be self-generating

src/permission/DefaultPermissions.php
```diff
@@ -24,95 +24,172 @@
...
+		self::registerCorePermissions();
+	}
+
 	public static function registerCorePermissions() : void{
-		$consoleRoot = self::registerPermission(new Permission(self::ROOT_CONSOLE, l10n::pocketmine_permission_group_console()));
-		$operatorRoot = self::registerPermission(new Permission(self::ROOT_OPERATOR, l10n::pocketmine_permission_group_operator()), [$consoleRoot]);
-		$everyoneRoot = self::registerPermission(new Permission(self::ROOT_USER, l10n::pocketmine_permission_group_user()), [$operatorRoot]);
+		$consoleRoot = self::registerPermission(new Permission("pocketmine.group.console", l10n::pocketmine_permission_group_console()));
+		$operatorRoot = self::registerPermission(new Permission("pocketmine.group.operator", l10n::pocketmine_permission_group_operator()), [$consoleRoot]);
+		$everyoneRoot = self::registerPermission(new Permission("pocketmine.group.user", l10n::pocketmine_permission_group_user()), [$operatorRoot]);
```
ShockedPlot7560: why removing the use of the const ?

ShockedPlot7560 (changes requested):
I'm on the way of @dktapps for the real value of this. As Permission doesn't guarantee that the object itself is registered, I don't necessarily see the need to manipulate objects. After that, if you manage to find a system that permit it, why not, but I don't see any solution for the moment.

DebHugo:
> After that, if you manage to find a system that permit it, why not, but I don't see any solution for the moment.

I thought at 2 solutions:
- create `Permission::create` function with same parameters thath the constructor and put the constructor `private`. It's like the constructor but register automaticly the `Permission` to `PermissionManager` (Big BC-Break)
- In `Permission` constructor, registery automaticly the permission to `PermissionManager` (No BC-Brreak for this but it's less logic)

ShockedPlot7560:
The best solution would be the PermissionManager returning the Permission object and you need to create it via PermissionManager to use them. But idk how this could be applied in a clean way

src/permission/DefaultPermissions.php
dktapps: Constants aren't needed because the permissions will be referenced by registry. It's no different to the likes of VanillaItems and VanillaBlocks.

dktapps:
> The best solution would be the PermissionManager returning the Permission object and you need to create it via PermissionManager to use them. But idk how this could be applied in a clean way

This wouldn't prevent people from doing `new Permission` and not registering it

src/permission/DefaultPermissions.php
ShockedPlot7560: The ROOT_* constants at the top of this class can then be deleted?

dktapps: Probably yes, although I didn't look closely if they were still used anywhere

src/Server.php
```diff
@@ -1007,8 +1007,6 @@ public function __construct(
 			)));
 			$this->logger->info($this->language->translate(KnownTranslationFactory::pocketmine_server_license($this->getName())));
 
-			DefaultPermissions::registerCorePermissions();
-
```
dktapps: Only just caught this, but this could be problematic if plugins register permissions that conflict with the built-in ones. `RegistryTrait` will only register the permissions when the first permission is accessed. Sure, technically `SimpleCommandMap` should trigger it but I don't feel happy about relying on that.

src/permission/DefaultPermissions.php
ShockedPlot7560: In any case, if I've understood the PR correctly, if you have to use const, you're doing it wrong because the aim of this PR is to remove the use of string in favour of the Permission object.
As far as I'm concerned, const should be removed, to be confirmed.

dktapps: Yep

src/Server.php
ShockedPlot7560: Protecting the namespace `pocketmine.` for registering permission could be a solution. I don't really like it but I don't see another way to do it atm

dktapps: Not sure how that would work without pre-registering the permissions somehow. Maybe `PermissionManager::make()` should do it similar to how we setup states for `RuntimeBlockStateRegistry`.

src/permission/PermissibleBase.php
```diff
@@ -28,12 +28,8 @@ final class PermissibleBase implements Permissible{
...
 	private PermissibleInternal $permissibleBase;
 
-	/**
-	 * @param bool[] $basePermissions
-	 * @phpstan-param array<string, bool> $basePermissions
-	 */
-	public function __construct(array $basePermissions){
-		$this->permissibleBase = new PermissibleInternal($basePermissions);
+	public function __construct(){
+		$this->permissibleBase = new PermissibleInternal();
```
DebHugo, line 32: I'm not sure of this change. What's the utility of `permissibleBase` ?

dktapps: It's for getting rid of cyclic references. Not related to the PR topic.

src/Server.php
DebHugo: I think plugins don't must remove/override a default permission

dktapps: It's not good enough to just say "pls don't do this". It shouldn't be possible to begin with.

DebHugo: Yes, it's must be add in other PR. Maybe, i can do this if it's necessary for submit this pr

dktapps: It's a problem that this PR is introducing. Because of the use of `RegistryTrait`, the permissions will be lazily loaded, which could easily end up being after plugins load if someone makes a wrong refactor. Relying on lazy loading for something like this is risky.

The old code's explicit initialization made sure this wouldn't be a problem, but now it's gone.

DebHugo: Ok so i need to re register `DefaultPermissions` after loadings of plugin ? it's that ?

dktapps: No, it needs to be initialized explicitly here, as it was originally.

DebHugo:
How do you want me to solve this problem ? I see solutions but it's horible solutions

dktapps:
Without breaking BC, the only way I see is to have `Permission::__construct()` call `PermissionManager::addPermission()`. I can't think of a use-case for constructing an unregistered permission anyway.

src/permission/PermissionAttachment.php
```diff
@@ -123,8 +111,8 @@ public function setPermission(Permission|string $name, bool $value) : void{
 		$this->recalculatePermissibles();
 	}
 
-	public function unsetPermission(Permission|string $name) : void{
-		$name = $name instanceof Permission ? $name->getName() : $name;
+	public function unsetPermission(Permission $permission) : void{
+		$name = $permission->getName();
 		if(isset($this->permissions[$name])){
```
ShockedPlot7560, line 116: Would be great to keep the ability to unset by the name.

src/plugin/PluginBase.php
```diff
@@ -174,7 +175,10 @@ private function registerYamlCommands() : void{
 
 			$newCmd->setAliases($aliasList);
 
-			$newCmd->setPermission($data->getPermission());
+			$permission = PermissionManager::getInstance()->getPermission($data->getPermission());
+			if ($permission !== null) {
+				$newCmd->setPermission($permission);
+			}
```
ShockedPlot7560, line 181: An exception should be raised in this case, no ?

src/permission/Permission.php
ShockedPlot7560: Removing the ability to define children in the constructor may prove more complex. Could we imagine keeping this and calling addChild? Or make addChild / removeChild chainable by returning them themselves.

ShockedPlot7560 (changes requested):

PR #6639, opened 2025-02-21
Add blocks interfaces for commons properties

DebHugo:
### Related issues & PRs

Add #6161 

### API changes

Add:
- `AgeableInterface`
- `AnalogRedstoneSignalEmitterInterface`
- `ColoredInterface`
- `CoralTypeInterface`
- `LightableInterface`
- `MultiFacingInterface`
- `PillarRotationInterface`
- `PoweredByRedstoneInterface`
- `SignLikeRotationInterface`
- `SingleFacingInterface`
- `WoodTypeInterface`

## Backwards compatibility

No BC Break


src/block/utils/SingleFacingInterface.php
```diff
@@ -0,0 +1,41 @@
...
+ *
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\block\utils;
+
+use pocketmine\math\Facing;
+
+interface SingleFacingInterface{
```
dktapps: This should be split into two variants `AnyFacing` and `HorizontalFacing`
The value sets for get/set are different so we shouldn't use the same interface

dktapps (changes requested):
I think the names of the interfaces are too verbose. I understand why (e.g. `WoodType` would conflict with the `WoodType` enum), but I don't like it much.

DebHugo:
> I think the names of the interfaces are too verbose. I understand why (e.g. `WoodType` would conflict with the `WoodType` enum), but I don't like it much.

May be replace `Interface` by `Property` (e.g. `WoodTypeProperty`, `HorizontalFacingProperty`...)

ShockedPlot7560:
> I think the names of the interfaces are too verbose. I understand why (e.g. `WoodType` would conflict with the `WoodType` enum), but I don't like it much.

The aim is also to follow the name of the trait with the name of the corresponding interface, I think. If we were to rename these interfaces, it would probably make more sense to rename the trait accordingly.
Then, it might be possible to name IXxxx instead of XxxxInterface? Idk

dktapps:
I'd personally favour e.g. `HasWoodType` but I'm not sure about it.

DebHugo:
> I'd personally favour e.g. `HasWoodType` but I'm not sure about it.

In my opinion, syntax `$block instanceof HasWoodType` is not the best

dktapps:
> > I'd personally favour e.g. `HasWoodType` but I'm not sure about it.
> 
> In my opinion, syntax `$block instanceof HasWoodType` is not the best

yeah I consider that but I don't really like any of the alternatives

dktapps:
Perhaps for wood specifically we can follow copper's lead with `WoodMaterial`. I don't recall if there are any other potential conflicts.

DebHugo:
> Perhaps for wood specifically we can follow copper's lead with `WoodMaterial`. I don't recall if there are any other potential conflicts.

So you want i remove `Interface` for names and for wood i put `WoodMaterial` ?

dktapps:
Yeah, that's where I'm leaning right now. Welcome to suggest better alternatives.

dktapps (changes requested):
The `self` returns need to be documented as `@return $this`

ipad54:
Just my opinion, but what do you think about making facing interfaces to end on `able` word? Like `MultiFaceable`, `HorizontalFaceable`, `SignLikeRotatable`?

dktapps:
I don't have much opinion on that, aside from preferring consistency with the traits (since the traits and interfaces will typically be used together).

src/block/Wood.php
```diff
@@ -32,7 +33,7 @@
 use pocketmine\player\Player;
 use pocketmine\world\sound\ItemUseOnBlockSound;
 
-class Wood extends Opaque{
+class Wood extends Opaque implements PillarRotation{
 	use PillarRotationTrait;
 	use WoodTypeTrait;
```
ShockedPlot7560: WoodMaterial is missing

src/block/utils/AnalogRedstoneSignalEmitter.php
```diff
@@ -0,0 +1,35 @@
...
+declare(strict_types=1);
+
+namespace pocketmine\block\utils;
+
+interface AnalogRedstoneSignalEmitter{
+
+	public function getOutputSignalStrength() : int;
+
+	/**
+	 * @throws \InvalidArgumentException if `$signalStrength` is out of bounds
```
ShockedPlot7560: Should we really notice `InvalidArgumentException` ?

src/block/utils/SignLikeRotation.php
```diff
@@ -0,0 +1,40 @@
...
+	/**
+	 * @return int between 0 and 15
+	 */
+	public function getRotation() : int;
+
+	/**
+	 * @param int $rotation between 0 and 15
+	 *
+	 * @throws \InvalidArgumentException if `$rotation` is out of hounds
+	 * @return $this
```
ShockedPlot7560: If we give bounds here, we need to give it for the others

ShockedPlot7560 (changes requested):

src/block/utils/AnalogRedstoneSignalEmitter.php
dktapps: No, we typically don't document this since it's not supposed to be caught

DebHugo: I nned t o remove InvalidArgumentException for all interfaces ?

src/block/utils/SignLikeRotation.php
DebHugo: there is just this interface that have bounds

src/block/utils/AnalogRedstoneSignalEmitter.php
dktapps: Yes

dktapps (approved):
LGTM

ShockedPlot7560 (approved):
OK for me

PR #6643, opened 2025-02-28
Fixed entity spawn position

leolee3914:
https://github.com/pmmp/PocketMine-MP/blob/3a2d0d77d184760dc5dd2b57c874cd3757de0d5e/src/entity/Entity.php#L791

### DIFF
--- a/src/entity/Entity.php
+++ b/src/entity/Entity.php
@@ -1495,7 +1495,7 @@ abstract class Entity{
 			$this->getId(), //TODO: actor unique ID
 			$this->getId(),
 			static::getNetworkTypeId(),
-			$this->location->asVector3(),
+			$this->getOffsetPosition($this->location->asVector3()),
 			$this->getMotion(),
 			$this->location->pitch,
 			$this->location->yaw,


leolee3914:
PrimedTNT spawns without offset position

https://github.com/user-attachments/assets/c397aa9d-9eda-46da-9fd6-c675de79cd17

PrimedTNT spawns with offset position

https://github.com/user-attachments/assets/4c6bf373-4d7f-47cb-8cf1-7b253c276be9

seems the position should only be fixed on AddActorPacket
Human->sendSpawnPacket() for players using AddPlayerPacket has no offset issue
i am not sure for ItemEntity because the offset is small

dktapps:
This should be tested with entities such as zombies and villagers

leolee3914:
zombie and villager have zero offset position.  nothing can be tested

dktapps:
Strange this hasn't been picked up on before. Do you know if it changed in a recent version?

leolee3914:
i think this issue has existed for many years
nobody noticed it because Entity->broadcastMovement() will correct the position shortly

dktapps:
I doubt that. The TNT position error is pretty obvious, someone would've noticed.

dktapps (approved):

PR #6646, opened 2025-03-06
Implemented Respawn Anchor

b1zeyofficial:
## Introduction
This PR implements Respawn Anchor
this is just a contribution to the community

### API changes

Added class `BlockExplodeEvent` called when a block explodes, after the explosion impact has been calculated
Added class `BlockPreExplodeEvent` called when a block wants to explode, before the explosion impact is calculated
Added class `PlayerRespawnAnchorUseEvent` called when a player interacts with a respawn anchor

### Behavioural changes
When RespawnAnchor explodes, Average TPS may drop by 19.24+-.

## Tests
https://www.youtube.com/watch?v=aI3ZItR3FY8


b1zeyofficial:
I think someone will find a use for this, but for now it's closed.

b1zeyofficial:
It seems to have been fixed

b1zeyofficial:
I checked php cs fixer locally but found no errors..

src/world/Explosion.php
```diff
@@ -221,4 +299,121 @@ public function explodeB() : bool{
...
+
+	private function boundary(float $start, float $distance) : float{
+		if($distance === 0.0){
+			return INF;
+		}
+
+		return $distance < 0 ?
+			($start - floor($start)) / -$distance :
+			(1 - ($start - floor($start))) / $distance;
+	}
```
dktapps: This code is basically identical to `VoxelRayTrace` in `pocketmine\math`.

src/world/Explosion.php
```diff
@@ -327,7 +327,8 @@ private function getSeenPercent(Vector3 $source, Entity $entity) : float{
 						$bb->minZ + $k * ($bb->maxZ - $bb->minZ) + $zOffset
 					);
 
-					if(!$this->raycastHit($source, $target)){
+					$generator = VoxelRayTrace::betweenPoints($source, $target);
+					if (!$generator->valid()) {
```
dktapps: That is not how generators work

b1zeyofficial: Can I use next() or current() ? Will it fix the error? In my IDE phpstan complained about the old code calling VoxelRayTrace but valid() fixed it.

b1zeyofficial: > Can I use next() or current() ? Will it fix the error? In my IDE phpstan complained about the old code calling VoxelRayTrace but valid() fixed it.

or just use foreach, i will try that when i get home

b1zeyofficial:
It seems like the generator starts and works.

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,99 @@
...
+
+use pocketmine\data\runtime\RuntimeDataDescriber;
+use pocketmine\event\block\BlockPreExplodeEvent;
+use pocketmine\item\Item;
+use pocketmine\math\Vector3;
+use pocketmine\player\Player;
+use pocketmine\world\Explosion;
+use pocketmine\world\sound\AnchorChargeSound;
+
+class RespawnAnchor extends Opaque{
```
ipad54: The class must be final

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,99 @@
...
+	protected int $charges = 0;
+
+	protected function describeBlockOnlyState(RuntimeDataDescriber $w) : void{
+		$w->boundedIntAuto(0, 4, $this->charges);
```
ipad54: It would be better to constify min and max charge values

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,99 @@
...
+	public function setCharges(int $charges) : self{
+		$this->charges = $charges;
+		return $this;
+	}
+
+	public function getLightLevel() : int{
+		return match ($this->charges){
+			1 => 3,
+			2 => 7,
+			default => ($this->charges > 2 ? 15 : 0),
```
ipad54: This is incorrect. Respawn anchor has light level 11 with 3 charges. The light calculation formula here is `min(0, 3 + 4 * ($this->charges - 1))`

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,99 @@
...
+		};
+	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if($item->getTypeId() === VanillaBlocks::GLOWSTONE()->asItem()->getTypeId() && $this->charges < 4){
```
ipad54: You can use here `ItemTypeIds::fromBlockTypeId()`

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,99 @@
...
+			if($player !== null && !$player->isCreative()){
```
ipad54: This check isn't necessary

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,99 @@
...
+	protected int $charges = 0;
+
+	protected function describeBlockOnlyState(RuntimeDataDescriber $w) : void{
+		$w->boundedIntAuto(0, 4, $this->charges);
+	}
+
+	public function getCharges() : int{ return $this->charges; }
+
+	public function setCharges(int $charges) : self{
+		$this->charges = $charges;
```
ipad54: You need to check that the charges amount is in range 0-4, otherwise throw an exception

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,99 @@
...
+			$this->position->getWorld()->addSound($this->position, new AnchorChargeSound());
+
+			return true;
+		}
+
+		if($player === null){
+			return false;
+		}
+
+		if($this->charges >= 1){
```
ipad54:
```suggestion
		if($this->charges > 0){

```

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,99 @@
...
+			$this->explode($player);
+			$this->position->getWorld()->setBlock($this->position, VanillaBlocks::AIR(), false);
+			return true;
+		}
+
+		return false;
+	}
+
+	public function explode(Player $player) : void{
```
ipad54: This function should be private

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,99 @@
...
+			return true;
+		}
+
+		if($player === null){
+			return false;
+		}
+
+		if($this->charges >= 1){
+			$this->explode($player);
+			$this->position->getWorld()->setBlock($this->position, VanillaBlocks::AIR(), false);
```
ipad54: This is not needed. You already set the air block in `explode()`

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,99 @@
...
+				$item->pop();
+			}
+			$this->charges++;
+			$this->position->getWorld()->setBlock($this->position, $this);
+			$this->position->getWorld()->addSound($this->position, new AnchorChargeSound());
+
+			return true;
+		}
+
+		if($player === null){
```
ipad54: Why do you limit this? Player can be optional in BlockPreExlodeEvent

src/world/Explosion.php
dktapps: Yes, use foreach

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,99 @@
...
+		}
+
+		return false;
+	}
+
+	public function explode(Player $player) : void{
+		$ev = new BlockPreExplodeEvent($this, 5, $player);
+		$ev->setIncendiary(true);
+	
+		if($ev->isCancelled()){
```
ipad54: The event will never be fired because `$ev->call()` is missing

src/block/VanillaBlocks.php
```diff
@@ -1642,6 +1643,9 @@ public function getLightLevel() : int{ return 10;}
 		self::register("warped_roots", fn(BID $id) => new NetherRoots($id, "Warped Roots", $netherRootsInfo));
 
 		self::register("chain", fn(BID $id) => new Chain($id, "Chain", new Info(BreakInfo::pickaxe(5.0, ToolTier::WOOD))));
+
+		$respawnAnchorBreakInfo = new Info(BreakInfo::pickaxe(50.0, ToolTier::DIAMOND, 1200.0));
```
ipad54: There is no point to make a break info variable just for one block

src/entity/object/EndCrystal.php
```diff
@@ -130,6 +130,11 @@ public function explode() : void{
 		$ev->call();
 		if(!$ev->isCancelled()){
 			$explosion = new Explosion($this->getPosition(), $ev->getRadius(), $this);
+			$explosion->setFireChance(0);
+
+			if(!isset($explosion->fireIgnitions)){
+				$explosion->fireIgnitions = [];
+			}
```
ipad54: I don't understand why do you need this. The properties are initialized by default

src/entity/object/PrimedTNT.php
```diff
@@ -122,6 +122,11 @@ public function explode() : void{
 		if(!$ev->isCancelled()){
 			//TODO: deal with underwater TNT (underwater TNT treats water as if it has a blast resistance of 0)
 			$explosion = new Explosion(Position::fromObject($this->location->add(0, $this->size->getHeight() / 2, 0), $this->getWorld()), $ev->getRadius(), $this);
+			$explosion->setFireChance(0);
+
+			if(!isset($explosion->fireIgnitions)){
+				$explosion->fireIgnitions = [];
+			}
```
ipad54: Same here

src/world/Explosion.php
```diff
@@ -78,13 +92,23 @@ public function __construct(
 	 * will be destroyed.
 	 */
 	public function explodeA() : bool{
+		if($this->what instanceof Explosive){
+			if($this->what instanceof Entity && $this->what->isUnderwater()){
+				$this->doesDamage = false;
+				return true;
+			}
+		}
```
ipad54: I don't think this has anything to do with the respawn anchor explosion.

src/world/Explosion.php
```diff
@@ -78,13 +92,23 @@ public function __construct(
...
 			return false;
 		}
 
 		$blockFactory = RuntimeBlockStateRegistry::getInstance();
 
 		$mRays = $this->rays - 1;
+		$incendiary = $this->fireChance > 0;
+		if($incendiary && !isset($this->fireIgnitions)){
+			$this->fireIgnitions = [];
+		}
```
ipad54: The property is always initialized

src/world/Explosion.php
```diff
@@ -119,19 +143,23 @@ public function explodeA() : bool{
...
 										$_block = $this->world->getBlockAt($vBlockX, $vBlockY, $vBlockZ, true, false);
 										foreach($_block->getAffectedBlocks() as $_affectedBlock){
 											$_affectedBlockPos = $_affectedBlock->getPosition();
-											$this->affectedBlocks[World::blockHash($_affectedBlockPos->x, $_affectedBlockPos->y, $_affectedBlockPos->z)] = $_affectedBlock;
+											$this->affectedBlocks[World::blockHash((int) $_affectedBlockPos->x, (int) $_affectedBlockPos->y, (int) $_affectedBlockPos->z)] = $_affectedBlock;
 										}
 									}
 								}
 							}
+							if($incendiary && mt_rand() / mt_getrandmax() <= $this->fireChance){
```
ipad54: `mt_rand() / mt_getrandmax()` can be replaced with `Utils::getRandomFloat()`

src/world/Explosion.php
```diff
@@ -191,6 +240,19 @@ public function explodeB() : bool{
...
+						if($item instanceof TieredTool && $item->getTier() === ToolTier::NETHERITE){
+							$netheritePieces++;
+						}
+					}
+					$netheriteReduction = 1 - (0.125 * $netheritePieces);
+					$netheriteReduction = max(0.5, $netheriteReduction);
+
+					$impact *= $netheriteReduction;
+				}
+
```
ipad54: This is not related to the respawn anchor

src/world/Explosion.php
```diff
@@ -170,18 +220,17 @@ public function explodeB() : bool{
...
 
 			if($distance <= 1){
-				$motion = $entityPos->subtractVector($this->source)->normalize();
+				$impact = max(0, (1 - $distance) * $this->getSeenPercent($this->source, $entity));
 
-				$impact = (1 - $distance) * ($exposure = 1);
+				$motion = $entityPos->subtractVector($this->source)->normalize();
 
-				$damage = (int) ((($impact * $impact + $impact) / 2) * 8 * $explosionSize + 1);
+				$damage = $this->doesDamage ? max((int) (((($impact * $impact + $impact) / 2) * 8 * $explosionSize) + 1), 0) : 0;
```
ipad54: This is not related to the respawn anchor

src/world/Explosion.php
```diff
@@ -209,10 +271,26 @@ public function explodeB() : bool{
...
 				}
-				if(($t = $this->world->getTileAt($pos->x, $pos->y, $pos->z)) !== null){
+
+				if(($t = $this->world->getTileAt((int) $pos->x, (int) $pos->y, (int) $pos->z)) !== null){
 					$t->onBlockDestroyed(); //needed to create drops for inventories
 				}
-				$this->world->setBlockAt($pos->x, $pos->y, $pos->z, $airBlock);
+
+				$this->world->setBlockAt((int) $pos->x, (int) $pos->y, (int) $pos->z, $airBlock);
+			}
```
ipad54: Unrelated changes

src/world/Explosion.php
```diff
@@ -119,19 +143,23 @@ public function explodeA() : bool{
...
-							if($blastResistance >= 0){
+							if($blastResistance >= 0 && $block->getTypeId() !== VanillaBlocks::AIR()->getTypeId()){
 								$blastForce -= ($blastResistance / 5 + 0.3) * $this->stepLen;
 								if($blastForce > 0){
 									if(!isset($this->affectedBlocks[World::blockHash($vBlockX, $vBlockY, $vBlockZ)])){
 										$_block = $this->world->getBlockAt($vBlockX, $vBlockY, $vBlockZ, true, false);
 										foreach($_block->getAffectedBlocks() as $_affectedBlock){
 											$_affectedBlockPos = $_affectedBlock->getPosition();
-											$this->affectedBlocks[World::blockHash($_affectedBlockPos->x, $_affectedBlockPos->y, $_affectedBlockPos->z)] = $_affectedBlock;
+											$this->affectedBlocks[World::blockHash((int) $_affectedBlockPos->x, (int) $_affectedBlockPos->y, (int) $_affectedBlockPos->z)] = $_affectedBlock;
```
ipad54: Unrelated change

src/world/Explosion.php
```diff
@@ -119,19 +143,23 @@ public function explodeA() : bool{
 
 							$state = $subChunk->getBlockStateId($vBlockX & SubChunk::COORD_MASK, $vBlockY & SubChunk::COORD_MASK, $vBlockZ & SubChunk::COORD_MASK);
 
+							$block = $this->world->getBlockAt($vBlockX, $vBlockY, $vBlockZ, true, false);
 							$blastResistance = $blockFactory->blastResistance[$state] ?? 0;
-							if($blastResistance >= 0){
+							if($blastResistance >= 0 && $block->getTypeId() !== VanillaBlocks::AIR()->getTypeId()){
```
ipad54: You don't need to check for air here. Air block always has negative blast resistance value

src/world/Explosion.php
```diff
@@ -54,6 +64,10 @@ class Explosion{
 	/** @var Block[] */
 	public array $affectedBlocks = [];
 	public float $stepLen = 0.3;
+	private bool $doesDamage = true;
+	public float $fireChance = 0.0;
+	/** @var Block[] */
+	public array $fireIgnitions = [];
```
ipad54: These properties probably should be private

src/world/sound/AnchorChargeSound.php
```diff
@@ -0,0 +1,35 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\world\sound;
+
+use pocketmine\math\Vector3;
+use pocketmine\network\mcpe\protocol\LevelSoundEventPacket;
+use pocketmine\network\mcpe\protocol\types\LevelSoundEvent;
+
+class AnchorChargeSound implements Sound{
```
ipad54: I think `RespawnAnchorChargeSound` would be more clear name

src/event/block/BlockPreExplodeEvent.php
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	/**
+	 * Establish the probability of fire
+	 */
+	public function setIncendiary(bool $incendiary) : void{
+		if(!$incendiary){
+			$this->fireChance = 0;
+		}else{
+			if($this->fireChance <= 0){
```
ipad54: You can combine this into simple `elseif` block

src/event/block/BlockPreExplodeEvent.php
```diff
@@ -0,0 +1,119 @@
...
+	 */
+	public function getFireChance() : float{
+		return $this->fireChance;
+	}
+
+	/**
+	 * Establish the probability of fire
+	 */
+	public function setFireChance(float $fireChance) : void{
+		$this->fireChance = $fireChance;
```
ipad54, line 119: You need to do a check here that the fire chance value is between 0-1

src/event/entity/EntityExplodeEvent.php
```diff
@@ -42,14 +42,17 @@ class EntityExplodeEvent extends EntityEvent implements Cancellable{
...
+	 * @param float   $yield     0-100 The explosion yield
+	 * @param Block[] $ignitions The list of blocks affected by fire ignitions
 	 */
 	public function __construct(
 		Entity $entity,
 		protected Position $position,
 		protected array $blocks,
-		protected float $yield
+		protected float $yield,
+		/** @var Block[] $ignitions */
```
ipad54: This var isn't necessary

src/event/entity/EntityExplodeEvent.php
```diff
@@ -98,4 +101,22 @@ public function setYield(float $yield) : void{
...
 		$this->yield = $yield;
 	}
+
+	/**
+	 * Sets the blocks that will be ignited by the explosion.
+	 *
+	 * @param Block[] $ignitions
+	 */
+	public function setIgnitions(array $ignitions) : void{
+		$this->ignitions = $ignitions;
```
ipad54, line 111: You need to validate array contents first

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,99 @@
...
+use pocketmine\data\runtime\RuntimeDataDescriber;
+use pocketmine\event\block\BlockPreExplodeEvent;
+use pocketmine\item\Item;
+use pocketmine\math\Vector3;
+use pocketmine\player\Player;
+use pocketmine\world\Explosion;
+use pocketmine\world\sound\AnchorChargeSound;
+
+class RespawnAnchor extends Opaque{
+	protected int $charges = 0;
```
ipad54: This property must be private

src/event/block/BlockExplodeEvent.php
```diff
@@ -0,0 +1,137 @@
...
+		return $this->blocks;
+	}
+
+	/**
+	 * Set the set of blocks affected by fire ignitions
+	 *
+	 * @param Block[] $blocks The set of blocks to be affected by fire ignitions
+	 */
+	public function setAffectedBlocks(array $blocks) : void{
+		$this->blocks = $blocks;
```
ipad54, line 101: Array contents must be validated

src/event/block/BlockExplodeEvent.php
```diff
@@ -0,0 +1,137 @@
...
+		return $this->ignitions;
+	}
+
+	/**
+	 * Set the set of blocks that can be ignited by the explosion
+	 *
+	 * @param Block[] $ignitions The set of blocks to set as ignited
+	 */
+	public function setIgnitions(array $ignitions) : void{
+		$this->ignitions = $ignitions;
```
ipad54, line 120: Same here

src/event/block/BlockExplodeEvent.php
```diff
@@ -0,0 +1,137 @@
...
+		return $this->fireChance;
+	}
+
+	/**
+	 * Set the fire chance of the explosion
+	 *
+	 * @param float $fireChance The fire chance to set
+	 */
+	public function setFireChance(float $fireChance) : void{
+		$this->fireChance = $fireChance;
```
ipad54: You need to do a check here that the fire chance value is between 0-1

src/world/Explosion.php
```diff
@@ -209,10 +271,26 @@ public function explodeB() : bool{
...
+
+			foreach($this->fireIgnitions as $fireBlock){
+				$firePos = $fireBlock->getPosition();
+				$x = (int) $firePos->x;
+				$y = (int) $firePos->y;
+				$z = (int) $firePos->z;
+
+				$toIgnite = $this->world->getBlockAt($x, $y, $z);
+
+				if($toIgnite->getTypeId() === VanillaBlocks::AIR()->getTypeId() &&
```
ipad54:
```suggestion
				if($toIgnite->getTypeId() === BlockTypeIds::AIR &&
```

src/world/Explosion.php
```diff
@@ -209,10 +271,26 @@ public function explodeB() : bool{
...
+					$toIgnite->getSide(Facing::UP)->isSolid()){
+					$this->world->setBlockAt($x, $y, $z, VanillaBlocks::FIRE());
```
ipad54: You can make a variable with fire block before the loop to save performance

ipad54 (changes requested):

ipad54:
I've mentioned the code paths in the review that aren't related to the respawn anchor. I think you should probably make another PR with your vanilla explosion improvements, because we don't accept multiple changes on different topics in one PR, it's too difficult to review multiple things at once.

src/block/RespawnAnchor.php
b1zeyofficial: should const be private too?

b1zeyofficial:
> I've mentioned the code paths in the review that aren't related to the respawn anchor. I think you should probably make another PR with your vanilla explosion improvements, because we don't accept multiple changes on different topics in one PR, it's too difficult to review multiple things at once.

Ok, I'll create two separate PR for this, but for now check for updates regarding the respawn anchor to make things easier for me... Ummm, can I close this?

src/block/RespawnAnchor.php
b1zeyofficial: PHPStan complains about this, should i remove getPlayer from BlockPreExplodeEvent class?

src/world/sound/AnchorChargeSound.php
b1zeyofficial: This name looks long but oh well

ipad54:
>  Ummm, can I close this?

Keep this open, just remove your explosion changes that not related to the respawn anchor

src/event/block/BlockExplodeEvent.php
```diff
@@ -96,6 +96,11 @@ public function getAffectedBlocks() : array{
 	 * @param Block[] $blocks The set of blocks to be affected by fire ignitions
 	 */
 	public function setAffectedBlocks(array $blocks) : void{
+		foreach($blocks as $block){
```
ipad54: You can just use `Utils::validateArrayValueType()`

src/block/RespawnAnchor.php
```diff
@@ -26,52 +26,50 @@
...
 			return true;
 		}
 
-		if($player === null){
-			return false;
-		}
-
-		if($this->charges >= 1){
+		if($this->charges > self::MIN_CHARGES){
 			$this->explode($player);
```
ipad54: You need to make `explode()` to accept a nullable parameter.

src/world/Explosion.php
```diff
@@ -106,8 +108,8 @@ public function explodeA() : bool{
 
 		$mRays = $this->rays - 1;
 		$incendiary = $this->fireChance > 0;
-		if($incendiary && !isset($this->fireIgnitions)){
-			$this->fireIgnitions = [];
+		if($incendiary){
+			$this->fireIgnitions = $this->fireIgnitions ?? [];
```
ipad54: I don’t understand what are you trying to do here, this is not necessary

b1zeyofficial:
the only thing i don't like is the explosion resistance of the stone, it shouldn't behave like that, vanilla has more exploding stones

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,92 @@
...
+use pocketmine\item\Item;
+use pocketmine\item\ItemTypeIds;
+use pocketmine\math\Vector3;
+use pocketmine\player\Player;
+use pocketmine\world\Explosion;
+use pocketmine\world\sound\RespawnAnchorChargeSound;
+
+final class RespawnAnchor extends Opaque{
+	protected const MIN_CHARGES = 0;
+	protected const MAX_CHARGES = 4;
```
ipad54: The constants must be private too

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	public function getLightLevel() : int{
+		return $this->charges > 0 ? 3 + 4 * ($this->charges - 1) : 0;
+	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if($item->getTypeId() === ItemTypeIds::fromBlockTypeId(BlockTypeIds::GLOWSTONE) && $this->charges < self::MAX_CHARGES){
+			$this->charges++;
+			$this->position->getWorld()->setBlock($this->position, $this);
```
ipad54: You can write this in single line by doing `$this->position->getWorld()->setBlock($this->position, $this->setCharges($this->charges + 1));`

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,92 @@
...
+
+		if($this->charges > self::MIN_CHARGES){
+			$this->explode($player);
+			return true;
+		}
+
+		return false;
+	}
+
+	public function explode(?Player $player) : void{
```
ipad54: This function needs to be private

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,92 @@
...
+		$ev = new BlockPreExplodeEvent($this, 5, $player);
+		$ev->setIncendiary(true);
+
```
ipad54: I don't understand why did you delete the `isCancelled()` check. Bring it back please and just add `$ev->call();` before it

src/data/bedrock/block/convert/BlockObjectToStateSerializer.php
```diff
@@ -1746,6 +1747,9 @@ private function registerSerializers() : void{
 			return Writer::create(Ids::RESIN_CLUMP)
 				->writeFacingFlags($block->getFaces());
 		});
+		$this->map(Blocks::RESPAWN_ANCHOR(), fn(RespawnAnchor $block) => Writer::create(Ids::RESPAWN_ANCHOR)
+		->writeInt(StateNames::RESPAWN_ANCHOR_CHARGE, $block->getCharges())
+		);
```
ipad54: You should probably make this in a single line or use a full anonymous function

src/data/bedrock/block/convert/BlockStateToObjectDeserializer.php
```diff
@@ -1597,6 +1597,9 @@ private function registerDeserializers() : void{
 		$this->mapStairs(Ids::RESIN_BRICK_STAIRS, fn() => Blocks::RESIN_BRICK_STAIRS());
 		$this->map(Ids::RESIN_BRICK_WALL, fn(Reader $in) => Helper::decodeWall(Blocks::RESIN_BRICK_WALL(), $in));
 		$this->map(Ids::RESIN_CLUMP, fn(Reader $in) => Blocks::RESIN_CLUMP()->setFaces($in->readFacingFlags()));
+		$this->map(Ids::RESPAWN_ANCHOR, fn(Reader $in) => Blocks::RESPAWN_ANCHOR()
+		->setCharges($in->readBoundedInt(StateNames::RESPAWN_ANCHOR_CHARGE, 0, 4))
+		);
```
ipad54: You should probably make this in a single line or use a full anonymous function

src/entity/object/EndCrystal.php
```diff
@@ -130,6 +130,8 @@ public function explode() : void{
 		$ev->call();
 		if(!$ev->isCancelled()){
 			$explosion = new Explosion($this->getPosition(), $ev->getRadius(), $this);
+			$explosion->setFireChance(0);
```
ipad54: This isn't necessary. The fire chance is 0 by default

src/entity/object/PrimedTNT.php
```diff
@@ -122,6 +122,8 @@ public function explode() : void{
 		if(!$ev->isCancelled()){
 			//TODO: deal with underwater TNT (underwater TNT treats water as if it has a blast resistance of 0)
 			$explosion = new Explosion(Position::fromObject($this->location->add(0, $this->size->getHeight() / 2, 0), $this->getWorld()), $ev->getRadius(), $this);
+			$explosion->setFireChance(0);
```
ipad54: Same here

src/event/block/BlockPreExplodeEvent.php
```diff
@@ -0,0 +1,120 @@
...
+class BlockPreExplodeEvent extends BlockEvent implements Cancellable{
+	use CancellableTrait;
+
+	private bool $blockBreaking = true;
+
+	public function __construct(
+		Block $block,
+		protected float $radius,
+		private readonly ?Player $player = null,
+		protected float $fireChance = 0.0
```
ipad54: All these properties must be private

ipad54: Also add a fire chance validation here too

src/event/block/BlockPreExplodeEvent.php
```diff
@@ -0,0 +1,120 @@
...
+	 */
+	public function getFireChance() : float{
+		return $this->fireChance;
+	}
+
+	/**
+	 * Establish the probability of fire
+	 */
+	public function setFireChance(float $fireChance) : void{
+		if($fireChance < 0 || $fireChance > 1){
```
ipad54: You also need to check that the fire chance isn't NaN or INF

src/event/entity/EntityExplodeEvent.php
```diff
@@ -98,4 +100,27 @@ public function setYield(float $yield) : void{
...
 		$this->yield = $yield;
 	}
+
+	/**
+	 * Sets the blocks that will be ignited by the explosion.
+	 *
+	 * @param Block[] $ignitions
+	 */
+	public function setIgnitions(array $ignitions) : void{
+		foreach($ignitions as $ignition){
```
ipad54: You also need to use `Utils::validateArrayValueType()` here

src/event/block/BlockExplodeEvent.php
```diff
@@ -0,0 +1,147 @@
...
+	 * @param Block[] $blocks
+	 * @param Block[] $ignitions
+	 */
+	public function __construct(
+		Block $block,
+		protected Position $position,
+		protected array $blocks,
+		protected float $yield,
+		private array $ignitions,
+		protected float $fireChance
```
ipad54: These properties also need to be private

ipad54: Also add a fire chance validation here too

src/event/block/BlockExplodeEvent.php
```diff
@@ -0,0 +1,147 @@
...
+		return $this->ignitions;
+	}
+
+	/**
+	 * Set the set of blocks that can be ignited by the explosion
+	 *
+	 * @param Block[] $ignitions The set of blocks to set as ignited
+	 */
+	public function setIgnitions(array $ignitions) : void{
+		foreach($ignitions as $ignition){
```
ipad54: You also need to use here `Utils::validateArrayValueType()`

src/event/block/BlockExplodeEvent.php
```diff
@@ -0,0 +1,147 @@
...
+		return $this->fireChance;
+	}
+
+	/**
+	 * Set the fire chance of the explosion
+	 *
+	 * @param float $fireChance The fire chance to set
+	 */
+	public function setFireChance(float $fireChance) : void{
+		if($fireChance < 0.0 || $fireChance > 1.0){
```
ipad54: Same here about NaN/INF

src/world/Explosion.php
```diff
@@ -150,13 +163,35 @@ public function explodeB() : bool{
 		$yield = min(100, (1 / $this->radius) * 100);
 
 		if($this->what instanceof Entity){
-			$ev = new EntityExplodeEvent($this->what, $this->source, $this->affectedBlocks, $yield);
+			$ev = new EntityExplodeEvent($this->what, $this->source, $this->affectedBlocks, $yield, $this->fireIgnitions);
+
+			$ev->setIgnitions($this->fireIgnitions);
```
ipad54: There is no point to call `$ev->setIgnitions()` because you already passed them in the event constructor

src/world/Explosion.php
```diff
@@ -150,13 +163,35 @@ public function explodeB() : bool{
...
+
+			$yield = $ev->getYield();
+			$this->affectedBlocks = $ev->getBlockList();
+		}elseif($this->what instanceof Block){
+			$ev = new BlockExplodeEvent(
+				$this->what,
+				$this->source,
+				$this->affectedBlocks,
+				$yield,
+				$this->affectedBlocks,
```
ipad54: Why do you use `$this->affectedBlocks` again instead of your fire ignitions array?

src/world/Explosion.php
```diff
@@ -214,11 +249,31 @@ public function explodeB() : bool{
...
 		}
 
 		$this->world->addParticle($source, new HugeExplodeSeedParticle());
 		$this->world->addSound($source, new ExplodeSound());
 
 		return true;
 	}
+
+	public function setFireChance(float $fireChance) : void{
+		$this->fireChance = $fireChance;
```
ipad54, line 283: Validation is missing here

ipad54 (changes requested):

src/world/Explosion.php
b1zeyofficial: After my research, adding `$this->fireIgnitions` gives a load on explosion respawnanchor. Maybe I should remove it and leave only one use of `$this->affectedBlocks`

src/event/block/BlockExplodeEvent.php
b1zeyofficial: I made private only `private array $ignitions,
protected float $fireChance` Does it make sense to do this for other properties?

src/world/Explosion.php
ipad54: > After my research, adding `$this->fireIgnitions` gives a load on explosion respawnanchor.

This is because you include air blocks that not affected by the explosion to the potential ignition list. To save performance, you need to add blocks to fire ignition in the place where they are added to the `affectedBlocks`

src/event/block/BlockExplodeEvent.php
ipad54: > Does it make sense to do this for other properties?

I don’t see any point why it doesn’t. Private properties can be always turned into protected or public if there is a reason to do that, but not vice versa

src/world/Explosion.php
b1zeyofficial: I don't quite understand where exactly to change the code
```
if($incendiary && isset($this->affectedBlocks[World::blockHash($vBlockX, $vBlockY, $vBlockZ)]) && Utils::getRandomFloat() <= $this->fireChance){
    $this->fireIgnitions[spl_object_id($block)] = $block;
}
```

but I don't think that's true

b1zeyofficial: Check out the latest updates, but the fire has become smaller

src/entity/object/EndCrystal.php
```diff
@@ -130,6 +130,7 @@ public function explode() : void{
 		$ev->call();
 		if(!$ev->isCancelled()){
 			$explosion = new Explosion($this->getPosition(), $ev->getRadius(), $this);
+
```
ipad54: Please remove this line addition

src/entity/object/PrimedTNT.php
```diff
@@ -122,6 +122,7 @@ public function explode() : void{
 		if(!$ev->isCancelled()){
 			//TODO: deal with underwater TNT (underwater TNT treats water as if it has a blast resistance of 0)
 			$explosion = new Explosion(Position::fromObject($this->location->add(0, $this->size->getHeight() / 2, 0), $this->getWorld()), $ev->getRadius(), $this);
+
```
ipad54: Please remove this line addition

src/world/Explosion.php
```diff
@@ -214,11 +246,34 @@ public function explodeB() : bool{
...
+			foreach($this->fireIgnitions as $ignition){
+				$firePos = $ignition->getPosition();
+				$x = (int) $firePos->x;
+				$y = (int) $firePos->y;
+				$z = (int) $firePos->z;
+
+				$toIgnite = $this->world->getBlockAt($x, $y, $z);
+
+				if($toIgnite->getTypeId() === BlockTypeIds::AIR &&
+					$toIgnite->getSide(Facing::UP)->isSolid()){
```
ipad54: Why do you use `UP` here instead of `DOWN`? Also, `isSolid()` isn't fire block support requirement. It requires the supporting block to provide `SupportType::FULL`

b1zeyofficial:
I dont understand, is this the expected behavior of fire or not?
image

src/world/Explosion.php
```diff
@@ -150,13 +164,33 @@ public function explodeB() : bool{
...
-			$ev = new EntityExplodeEvent($this->what, $this->source, $this->affectedBlocks, $yield);
+			$ev = new EntityExplodeEvent($this->what, $this->source, $this->affectedBlocks, $yield, $this->fireIgnitions);
+
+			$ev->call();
+			if($ev->isCancelled()){
+				return false;
+			}
+
+			$yield = $ev->getYield();
+			$this->affectedBlocks = $ev->getBlockList();
```
ipad54, line 182: fire ignitions update from EntityExplodeEvent is missing here

src/world/Explosion.php
```diff
@@ -214,11 +248,34 @@ public function explodeB() : bool{
 				}
 				$this->world->setBlockAt($pos->x, $pos->y, $pos->z, $airBlock);
 			}
+
+			$fireBlock = VanillaBlocks::FIRE();
+
+			foreach($this->fireIgnitions as $ignition){
```
ipad54: I don’t like that you have to iterate over `fireIgnitions` on each `affectedBlocks` iteration. I think you could have `World::blockHash()` as a `fireIgnitions` element key instead of spl_object_id() (similar to how `affectedBlocks` are stored) and then check if `fireIgnitions` has that key on `affectedBlocks` iteration. If so and block below can support fire block, set fire instead of air

src/world/Explosion.php
```diff
@@ -125,6 +136,9 @@ public function explodeA() : bool{
 								if($blastForce > 0){
 									if(!isset($this->affectedBlocks[World::blockHash($vBlockX, $vBlockY, $vBlockZ)])){
 										$_block = $this->world->getBlockAt($vBlockX, $vBlockY, $vBlockZ, true, false);
+										if($incendiary && Utils::getRandomFloat() <= $this->fireChance){
```
ipad54: This should be done inside the `getAffectedBlocks()` loop to include all block parts if it’s a double block

src/world/Explosion.php
```diff
@@ -54,6 +61,9 @@ class Explosion{
 	/** @var Block[] */
 	public array $affectedBlocks = [];
 	public float $stepLen = 0.3;
+	private float $fireChance = 0.0;
```
ipad54: I would also like to have an ability to pass fire chance in the explosion constructor

b1zeyofficial:
> I dont understand, is this the expected behavior of fire or not?

Vanilla 1.21.62
image

ipad54 (changes requested):

src/world/Explosion.php
b1zeyofficial: I'm not sure we need to use World::blockHash twice, instead I would prefer using $affectedHash variable like

ipad54: Yes, you can create `$posHash` variable

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,95 @@
...
+		$ev->call();
+		if($ev->isCancelled()){
+			return;
+		}
+
+		$this->position->getWorld()->setBlock($this->position, VanillaBlocks::AIR());
+
+		$explosion = new Explosion($this->position, $ev->getRadius(), $this);
```
ipad54: According to Minecraft Java code, the explosion happens at the centered position, so block X, Y, Z need to be adjusted by 0.5

src/world/Explosion.php
```diff
@@ -214,11 +250,35 @@ public function explodeB() : bool{
 				}
 				$this->world->setBlockAt($pos->x, $pos->y, $pos->z, $airBlock);
```
ipad54: You can just do `$this->world->setBlockAt($pos->x, $pos->y, $pos->z,
				 	isset($this->fireIgnitions[$hash]) &&
					$block->getSide(Facing::DOWN)->getSupportType(Facing::UP) === SupportType::FULL ? $fireBlock : $airBlock
				);` here instead of adding another foreach

ipad54:
The block also needs to be registered in `StringToItemParser`

src/block/RespawnAnchor.php
b1zeyofficial: I don't quite understand this, because we need to create an object `$position = new Position`.. Does it make sense to do this if everything is fine

ipad54: It's necessary because without adjustment the explosion won't be able to destroy much blocks with good blast resistance (such as stone).

Explosion without adjustment:
Screenshot_20250316_163905

Explosion with adjustment:
Screenshot_20250316_164045

src/event/block/BlockExplodeEvent.php
```diff
@@ -0,0 +1,147 @@
...
+		return $this->fireChance;
+	}
+
+	/**
+	 * Set the fire chance of the explosion
+	 *
+	 * @param float $fireChance The fire chance to set
+	 */
+	public function setFireChance(float $fireChance) : void{
+		if(!is_finite($fireChance) || $fireChance < 0.0 || $fireChance > 1.0){
```
dktapps: Use `Utils::checkFloatNotInfOrNaN()`

src/event/block/BlockExplodeEvent.php
```diff
@@ -0,0 +1,147 @@
...
+	public function getYield() : float{
+		return $this->yield;
+	}
+
+	/**
+	 * Set the explosion yield (intensity).
+	 *
+	 * @param float $yield The intensity of the explosion (0-100)
+	 */
+	public function setYield(float $yield) : void{
```
dktapps, line 77: This also needs to use `Utils::checkFloatNotInfOrNaN()`

src/event/block/BlockExplodeEvent.php
```diff
@@ -0,0 +1,147 @@
...
+		private float $fireChance
+	){
+		parent::__construct($block);
+
+		if($yield < 0.0 || $yield > 100.0){
+			throw new \InvalidArgumentException("Yield must be in range 0.0 - 100.0");
+		}
+		if($fireChance < 0.0 || $fireChance > 1.0){
+			throw new \InvalidArgumentException("Fire chance must be between 0 and 1.");
+		}
```
dktapps: and again here for yield and fireChance

src/event/block/BlockExplodeEvent.php
```diff
@@ -53,11 +52,13 @@ public function __construct(
...
 		parent::__construct($block);
 
+		Utils::checkFloatNotInfOrNaN("fireChance", $fireChance);
+		Utils::checkFloatNotInfOrNaN("yield", $yield);
 		if($yield < 0.0 || $yield > 100.0){
 			throw new \InvalidArgumentException("Yield must be in range 0.0 - 100.0");
 		}
 		if($fireChance < 0.0 || $fireChance > 1.0){
-			throw new \InvalidArgumentException("Fire chance must be between 0 and 1.");
+			throw new \InvalidArgumentException("Fire chance must be a infinite or nan number between 0 and 1.");
```
dktapps: You don't ned to mention infinite or nan here, it's already reported by Utils if needed

src/event/block/BlockExplodeEvent.php
```diff
@@ -139,8 +141,9 @@ public function getFireChance() : float{
 	 * @param float $fireChance The fire chance to set
 	 */
 	public function setFireChance(float $fireChance) : void{
-		if(!is_finite($fireChance) || $fireChance < 0.0 || $fireChance > 1.0){
-			throw new \InvalidArgumentException("Fire chance must be a finite number between 0 and 1.");
+		Utils::checkFloatNotInfOrNaN("fireChance", $fireChance);
+		if($fireChance < 0.0 || $fireChance > 1.0){
+			throw new \InvalidArgumentException("Fire chance must be a inf or nan number between 0 and 1.");
```
dktapps: Same again here

src/event/block/BlockPreExplodeEvent.php
```diff
@@ -49,8 +49,9 @@ public function __construct(
 		if($radius <= 0){
 			throw new \InvalidArgumentException("Explosion radius must be positive");
 		}
+		Utils::checkFloatNotInfOrNaN("fireChance", $fireChance);
 		if($fireChance < 0.0 || $fireChance > 1.0){
-			throw new \InvalidArgumentException("Fire chance must be between 0 and 1.");
+			throw new \InvalidArgumentException("Fire chance must be a inf or nan number between 0 and 1.");
```
dktapps: and here

src/event/block/BlockPreExplodeEvent.php
```diff
@@ -109,8 +110,9 @@ public function getFireChance() : float{
 	 * Establish the probability of fire
 	 */
 	public function setFireChance(float $fireChance) : void{
-		if(!is_finite($fireChance) || $fireChance < 0.0 || $fireChance > 1.0){
-			throw new \InvalidArgumentException("Fire chance must be a finite number between 0 and 1.");
+		Utils::checkFloatNotInfOrNaN("fireChance", $fireChance);
+		if($fireChance < 0.0 || $fireChance > 1.0){
+			throw new \InvalidArgumentException("Fire chance must be a inf or nan number between 0 and 1.");
```
dktapps: and here

src/world/Explosion.php
```diff
@@ -264,8 +264,9 @@ public function explodeB() : bool{
 	}
 
 	public function setFireChance(float $fireChance) : void{
-		if(!is_finite($fireChance) || $fireChance < 0.0 || $fireChance > 1.0){
-			throw new \InvalidArgumentException("Fire chance must be a finite number between 0 and 1.");
+		Utils::checkFloatNotInfOrNaN("fireChance", $fireChance);
+		if($fireChance < 0.0 || $fireChance > 1.0){
+			throw new \InvalidArgumentException("Fire chance must be a inf or nan number between 0 and 1.");
```
dktapps: and here

src/world/Explosion.php
```diff
@@ -199,7 +232,8 @@ public function explodeB() : bool{
 		$air = VanillaItems::AIR();
 		$airBlock = VanillaBlocks::AIR();
 
-		foreach($this->affectedBlocks as $block){
+		$this->affectedBlocks = Utils::promoteKeys($this->affectedBlocks);
```
ipad54: You can use promoteKeys() in the foreach instead of this

src/event/block/BlockPreExplodeEvent.php
```diff
@@ -0,0 +1,126 @@
...
+ * @see BlockExplodeEvent
+ */
+class BlockPreExplodeEvent extends BlockEvent implements Cancellable{
+	use CancellableTrait;
+
+	private bool $blockBreaking = true;
+
+	public function __construct(
+		Block $block,
+		private float $radius,
```
ipad54, line 46: Also check the radius not to be NaN/INF

src/event/block/BlockPreExplodeEvent.php
```diff
@@ -0,0 +1,126 @@
...
+			throw new \InvalidArgumentException("Fire chance must be a number between 0 and 1.");
+		}
+		parent::__construct($block);
+	}
+
+	public function getRadius() : float{
+		return $this->radius;
+	}
+
+	public function setRadius(float $radius) : void{
```
ipad54, line 65: And here

src/event/block/BlockPreExplodeEvent.php
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	/**
+	 * Establish the probability of fire
+	 */
+	public function setIncendiary(bool $incendiary) : void{
+		if(!$incendiary){
+			$this->fireChance = 0;
+		}elseif($this->fireChance <= 0){
+			$this->fireChance = 1.0 / 3.0;
```
ipad54: It would be better to create a private `DEFAULT_FIRE_CHANCE` constant with this value

src/world/Explosion.php
```diff
@@ -54,13 +58,16 @@ class Explosion{
...
+	private array $fireIgnitions = [];
 
 	private SubChunkExplorer $subChunkExplorer;
 
 	public function __construct(
 		public Position $source,
 		public float $radius,
-		private Entity|Block|null $what = null
+		private Entity|Block|null $what = null,
+		private float $fireChance = 0.0
```
ipad54, line 75: Also check fire chance here for INF/NaN

src/world/Explosion.php
```diff
@@ -212,13 +245,28 @@ public function explodeB() : bool{
 				if(($t = $this->world->getTileAt($pos->x, $pos->y, $pos->z)) !== null){
 					$t->onBlockDestroyed(); //needed to create drops for inventories
 				}
-				$this->world->setBlockAt($pos->x, $pos->y, $pos->z, $airBlock);
```
ipad54, line 215: You should still set the block inside this condition. Fire can't be placed on the TNT block pos in vanilla

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,96 @@
...
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if($item->getTypeId() === ItemTypeIds::fromBlockTypeId(BlockTypeIds::GLOWSTONE) && $this->charges < self::MAX_CHARGES){
+			$this->position->getWorld()->setBlock($this->position, $this->setCharges($this->charges + 1));
+			$this->position->getWorld()->addSound($this->position, new RespawnAnchorChargeSound());
+			return true;
+		}
+
+		if($this->charges > self::MIN_CHARGES){
+			$this->explode($player);
```
ipad54: Please add `//TODO: Implement the ability to set a respawn point in Nether` before calling explode()

ipad54 (changes requested):

ipad54:
Please remove your changes in composer.lock & fix the tests by running php-cs-fixer

b1zeyofficial:
sad

b1zeyofficial:
I don't understand how to remove the file change, maybe I should just recreate the PR?

b1zeyofficial:
mm, that's it

b1zeyofficial:
Is there any point in fixing this PHPStan?

src/world/Explosion.php
```diff
@@ -214,11 +250,27 @@ public function explodeB() : bool{
...
 				$this->world->setBlockAt($pos->x, $pos->y, $pos->z, $airBlock);
 			}
+
+			$this->world->setBlockAt(
+				$pos->x, $pos->y, $pos->z,
+				isset($this->fireIgnitions[$hash]) &&
+				$block->getSide(Facing::DOWN)->getSupportType(Facing::UP) === SupportType::FULL
+					? VanillaBlocks::FIRE()
+					: $airBlock
+			);
```
dktapps: Why are you setting the block twice?

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,97 @@
...
+	public function setCharges(int $charges) : self{
+		if($charges < self::MIN_CHARGES || $charges > self::MAX_CHARGES){
+			throw new \InvalidArgumentException("Charges must be between " . self::MIN_CHARGES . " and " . self::MAX_CHARGES . ", given: $charges");
+		}
+		$this->charges = $charges;
+		return $this;
+	}
+
+	public function getLightLevel() : int{
+		return $this->charges > 0 ? 3 + 4 * ($this->charges - 1) : 0;
```
dktapps: Looks like this could just be
```suggestion
		return ($this->charges * 4) - 1;
```

Am I wrong?

ipad54: You should still check that the charges > 0, otherwise the light level will be -1 for 0 charges

b1zeyofficial: you can take into account the check > 0

b1zeyofficial: i think you can take into account the check > 0

src/event/block/BlockPreExplodeEvent.php
```diff
@@ -0,0 +1,129 @@
...
+ * Event triggered before a block explosion, allowing modifications to the explosion radius, block destruction, and fire chances.
+ * This event is used to customize the behavior of explosions before they happen.
+ *
+ * @see BlockExplodeEvent
+ */
+class BlockPreExplodeEvent extends BlockEvent implements Cancellable{
+	use CancellableTrait;
+
+	private bool $blockBreaking = true;
+	private const DEFAULT_FIRE_CHANCE = 1.0 / 3.0;
```
ipad54: By convention constants are usually placed above properties

src/world/Explosion.php
```diff
@@ -214,11 +251,25 @@ public function explodeB() : bool{
 				}
 				$this->world->setBlockAt($pos->x, $pos->y, $pos->z, $airBlock);
 			}
+
+			if(isset($this->fireIgnitions[$hash]) &&
+				$block->getSide(Facing::DOWN)->getSupportType(Facing::UP) === SupportType::FULL
+			){
+				$this->world->setBlockAt($pos->x, $pos->y, $pos->z, $fireBlock);
```
ipad54: You are still calling setBlockAt() twice. Do this check in the place where air block is being set (on line 252) and set fire instead if the check is true.

src/block/RespawnAnchor.php
ipad54: So basically it’ll be `return $this->charges > 0 ? ($this->charges * 4) - 1 : 0;`

dktapps: > You should still check that the charges > 0, otherwise the light level will be -1 for 0 charges

True. Good catch.

ipad54 (dismissed):

src/event/block/BlockExplodeEvent.php
```diff
@@ -0,0 +1,122 @@
...
+		return $this->blocks;
+	}
+
+	/**
+	 * Sets the blocks destroyed by the explosion.
+	 *
+	 * @param Block[] $blocks
+	 */
+	public function setAffectedBlocks(array $blocks) : void{
+		Utils::validateArrayValueType($blocks, fn(Block $block) => null);
```
ShockedPlot7560, line 100: static fn can be maded probably here ?

src/event/block/BlockExplodeEvent.php
```diff
@@ -0,0 +1,122 @@
...
+		return $this->ignitions;
+	}
+
+	/**
+	 * Set the list of blocks that will be replaced by fire.
+	 *
+	 * @param Block[] $ignitions
+	 */
+	public function setIgnitions(array $ignitions) : void{
+		Utils::validateArrayValueType($ignitions, fn(Block $block) => null);
```
ShockedPlot7560, line 119: same here ?

src/event/block/BlockExplodeEvent.php
dktapps: No need, it's not persisted

src/event/block/BlockExplodeEvent.php
dktapps: As above

ShockedPlot7560 (dismissed):
Looks OK to me

src/world/Explosion.php
```diff
@@ -198,8 +235,9 @@ public function explodeB() : bool{
 
 		$air = VanillaItems::AIR();
 		$airBlock = VanillaBlocks::AIR();
+		$fireBlock = VanillaBlocks::FIRE();
 
-		foreach($this->affectedBlocks as $block){
+		foreach(Utils::promoteKeys($this->affectedBlocks) as $hash => $block){
```
dktapps: This shouldn't be needed, `$hash` is an integer

You probably ought to add `@phpstan-var array<int, Block>` to the property declaration

src/world/Explosion.php
```diff
@@ -57,7 +57,7 @@ class Explosion{
 	private int $rays = 16;
 	public World $world;
 
-	/** @var Block[] */
+	/** @phpstan-var array<int, Block> */
```
dktapps: The original should be kept also

src/world/Explosion.php
```diff
@@ -57,6 +57,7 @@ class Explosion{
 	private int $rays = 16;
 	public World $world;
 
+	/** @var Block[] */
```
dktapps: Not like this.
It should be
```
/**
 * @var Block[]
 * @phpstan-var array<int, Block>
 */
```

b1zeyofficial: oops, I updated the file from the browser editor, my bad

dktapps:
Per internal discussion we think it would be a good idea to have a PlayerRespawnAnchorUseEvent with actions EXPLODE, SET_SPAWN and NONE. The default behaviour in the absence of dimensions would be EXPLODE, but this would allow plugins to alter the behaviour based on custom conditions (and perhaps custom dimensions).

b1zeyofficial:
> Per internal discussion we think it would be a good idea to have a PlayerRespawnAnchorUseEvent with actions EXPLODE, SET_SPAWN and NONE. The default behaviour in the absence of dimensions would be EXPLODE, but this would allow plugins to alter the behaviour based on custom conditions (and perhaps custom dimensions).

Maybe with this I can implement respawn set point for respawn achor? In nether
But nether is not implemented yet, I will use DimensionsIds. How do you look at this?

dktapps:
No, I'm saying that the logic for setting the spawn point should be implemented, but that the behaviour should be to EXPLODE by default. This will let plugins change it without having to implement extra logic themselves.

b1zeyofficial:
I found a problem that I can't solve. When exploding, using glowstone, when we charge the anchor for 5 charges, it explodes but the GLOWSTONE block itself is placed, this is not Vanilla behavior. I dont understand what the problem is.. 

P.S: This works on all blocks that can be placed

src/block/RespawnAnchor.php
```diff
@@ -68,9 +73,45 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
-			//TODO: Implement the ability to set a respawn point in Nether
-			$this->explode($player);
-			return true;
+			if($player === null){
+				return false;
+			}
+
+			$ev = new PlayerRespawnAnchorUseEvent($player, $this, PlayerRespawnAnchorUseEvent::ACTION_EXPLODE);
+
+			if($this->position->getWorld()->getId() === DimensionIds::NETHER){
```
dktapps: I said _don't_ do this. This is never going to be the case.

The aim is to make plugins able to change the outcome.

src/block/RespawnAnchor.php
```diff
@@ -68,9 +73,45 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+
+				case PlayerRespawnAnchorUseEvent::ACTION_SET_SPAWN:
+					if($this->charges >= 4){
+						if($player->getSpawn() !== null && $player->getSpawn()->equals($this->position)){
+							return true;
+						}
+
+						$player->setSpawn($this->position);
+						$this->position->getWorld()->addSound($this->position, new RespawnAnchorSetSpawnSound());
+						$player->sendMessage(TextFormat::GRAY . "Respawn point set");
```
dktapps: This needs translations

src/block/RespawnAnchor.php
```diff
@@ -68,9 +73,45 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
...
+				return false;
+			}
+
+			switch($ev->getAction()){
+				case PlayerRespawnAnchorUseEvent::ACTION_EXPLODE:
+					$this->explode($player);
+					return false;
+
+				case PlayerRespawnAnchorUseEvent::ACTION_SET_SPAWN:
+					if($this->charges >= 4){
```
dktapps: This should use a constant

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,139 @@
...
```
nicholass003: To set the player spawn point, it doesn't have to be full, use this
 ```php
if($this->charges > self::MIN_CHARGES)
```

File: src/block/RespawnAnchor.php (hunk 90 above)
Comment (nicholass003) [reply src/block/RespawnAnchor.php]: Also ```World->getId()``` doesn't return ```DimensionIds```, it returns a unique world identifier.
https://github.com/pmmp/PocketMine-MP/blob/e88b81a4cbd9274aa3b6faa0b3f847c13ca0ecd5/src/world/World.php#L616

Comment (dktapps) [reply src/block/RespawnAnchor.php]: Honestly we probably ought to deprecate this function. It serves no obvious purpose. But that's a topic for another issue.

File: src/block/RespawnAnchor.php (hunk 91 above)
Comment (b1zeyofficial) [reply src/block/RespawnAnchor.php]: can i create PR Language for this now?

**Comment by b1zeyofficial** [comment]
im thinking of making it so that before each player dies, the anchor charge is taken away and the sound RespawnAnchorDepleteSound is played

File: src/block/RespawnAnchor.php (hunk 93 above)
Comment (b1zeyofficial) [reply src/block/RespawnAnchor.php]: We still need to use >= because the logic requires the number of charges to be equal to or greater than the max value to set the respawn point. Otherwise the respawn point may not be set

Comment (nicholass003) [reply src/block/RespawnAnchor.php]: > We still need to use >= because the logic requires the number of charges to be equal to or greater than the max value to set the respawn point. Otherwise the respawn point may not be set

Where did you get this logic from?
This is the basis that I took as a reference.
https://minecraft.wiki/w/Respawn_Anchor#Respawning

Comment (b1zeyofficial) [reply src/block/RespawnAnchor.php]: My bad, i thought that the anchor should be fully charged before setting the respawn point. I checked the videos on YouTube how it works in gameplay

File: src/block/RespawnAnchor.php [hunk 94]
```diff
@@ -0,0 +1,126 @@
...
+		}
+
+		if($this->charges > self::MIN_CHARGES){
+			if($player === null){
+				return false;
+			}
+
+			$ev = new PlayerRespawnAnchorUseEvent($player, $this, PlayerRespawnAnchorUseEvent::ACTION_EXPLODE);
+			$ev->call();
+			if($ev->isCancelled()){
```
dktapps, line 81: Having this as well as NONE doesn't make any sense.

src/block/RespawnAnchor.php
```diff
@@ -0,0 +1,126 @@
...
+					return false;
+
+				case PlayerRespawnAnchorUseEvent::ACTION_SET_SPAWN:
+					if($player->getSpawn() !== null && $player->getSpawn()->equals($this->position)){
+						return true;
+					}
+
+					$player->setSpawn($this->position);
+					$this->position->getWorld()->addSound($this->position, new RespawnAnchorSetSpawnSound());
+					$player->sendMessage(TextFormat::GRAY . "Respawn point set");
```
dktapps: Still need translations for this

dktapps (changes requested):
It doesn't make any sense for the event to be both cancellable and have a NONE action

dktapps (dismissed):
Looks good, tested and seems working well

ShockedPlot7560 (dismissed):

ipad54 (approved):

dktapps (approved):

Yexeed:
This pr actually breaks compatibility, which was never specified, neither here or changelog: EntityExplodeEvent got new (required) field in it's constructor, breaking plugins that was using it earlier.

Yexeed:
Also, Explosion.php got new field, which is private. How did it even get through? It literally says good luck with expanding and making custom logic for explosions (e.g. different calculation, overriding functions etc.)
And another: BlockExplodeEvent (which is new, that's ok), seems to have similar fields to EntityExplodeEvent: `$blocks` and `$yield`. However, those were `protected` in EntityExplode, but in BlockExplode, they are `private`. Is this also ok, @dktapps ?

dktapps:
Private is favoured unless there's a specific reason to increase their visibility. We can increase visibility from `private` without breaking compatibility, but the reverse breaks BC. In the case of events there's not really a good reason to make them more than `private` anyway.

In the case of `Explosion`, it's a mess anyway. It already has private fields so this isn't an isolated case. I wouldn't be against increasing the visibility though.

Good catch on the `EntityExplodeEvent` BC break though. I guess no one spotted that.

Issue #6556, opened 2024-12-03
A way for world generators to not be forced to run in an async task


dktapps:
## Problem description
Some generators, like `Flat`, do not require an async task, and using an async task to just copy a chunk is a waste of resources.

There may be other cases where it might be desirable to run parts of, or all, of generation on the main thread, such as when creating entities or tiles during a generation step (e.g. paintings or chests).

## Proposed solution
Make `Generator->generateChunk()` and `Generator->populateChunk()` return `Promise`, so that individual generators can decide how they want to do async generation. This could be a promise resolved immediately on the main thread with a chunk copy, a `PopulationTask`, or perhaps some other mechanism entirely.

It may be more complex than this (e.g. a chunk source may be needed for adjacent chunks), but that's the basic premise.

dktapps:
Implemented in #6682

PR #6668, opened 2025-04-05
Extract GeneratorExecutor system from World

dktapps:
This is aimed at enabling chunk generators to exist which don't use async tasks, as described in #6556.

### Related issues & PRs
#6556

### API changes
- `AsyncGeneratorExecutor` class added that encapsulates the logic of generating chunks using async tasks as previously
- `GeneratorExecutor` interface added that can be implemented to provide chunks in other ways
- `SyncGeneratorExecutor` which invokes the generator directly on the main thread, useful for simple generators like `Flat` where async tasks are not needed
- Some redundant APIs were removed from `World` (these will probably come back as deprecated stubs for the remainder of 5.x, but I was having too much fun deleting code)
- Removed internal `World->registerGeneratorToWorker()` (no longer useful)

### Behavioural changes
- `Flat` world generator now runs on the main thread, which is faster, has less overhead, and uses less memory

## Backwards compatibility
Backwards compatible, minus some easily stubbable API method removals.

## Follow-up
Allow providing a `World` with a custom `ChunkGenerator` impl? May be useful for chunk server implementations (e.g. using BDS as a generator source)

## Tests
Manually tested with `normal` and `flat` generators

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   build:
     name: Update Docker Hub images
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     steps:
       - name: Set up Docker Buildx
--- a/.github/workflows/draft-release-pr-check.yml
+++ b/.github/workflows/draft-release-pr-check.yml
@@ -24,7 +24,7 @@ permissions:
 jobs:
   check-intent:
     name: Check release trigger
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     outputs:
       valid: ${{ steps.validate.outputs.DEV_BUILD == 'false' }}
@@ -43,7 +43,7 @@ jobs:
     #don't do these checks if this isn't a release - we don't want to generate unnecessary failed statuses
     if: needs.check-intent.outputs.valid == 'true'
 
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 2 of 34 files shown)


ShockedPlot7560 (dismissed):
As it look to a lot of cut code from World to AsyncGenerator, it look OK to me. I didn't analyse the logic in its entirety (I compared each function, seeing that the vast majority of them were copied and pasted)

I like the fact that it's fully BC compatible with current registered generators

You probably had a lot of fun deleting so much code from World 😂

dktapps:
Couple of issues remain:

- I don't like the way I avoided the cycle between World and GeneratorExecutor
- PopulationTask's invocation, chunk unlocking etc. is duplicated between the two executor types. However the code is dissimilar enough that I'm not sure how to generalize it.
- `World->getId()` becomes conditionally useless with this change. We ought to use some ID specifically designated by the `AsyncGeneratorExecutor`, since that's the only place that actually uses it. We'd be able to deprecate numeric world IDs once this is done.

dktapps:
This might also be useful for generating chunks when plugins need to place objects that aren't thread safe, such as tiles. In such cases it might also be desirable to decide whether the individual stages of generation should be async or not. For example it might be useful to have base terrain generation run asynchronously and have (some) populators run on the main thread. The current architecture isn't flexible enough for this, but it would be nice if an easy way to do this were to pop out of the woodwork.

dktapps:
Also need to take a closer look at the sync generator to make sure it sets the dirty flags properly on new chunks, otherwise new chunks won't be saved correctly

dktapps:
Closing in favour of the (greatly simplified) #6682

Issue #6637, opened 2025-02-19
Add EntityExtinguishEvent


portfolec:
### Problem description

Its not possible to detect right now when entity was extinguished

### Proposed solution

Add EntityExtinguishEvent to Entity::extinguish()

dktapps:
Good idea

AzaleeX:
@dktapps issues finish

dktapps:
Thanks

PR #6671, opened 2025-04-12
Add EntityExtinguishEvent

zSALLAZAR:
Add EntityExtinguishEvent

### Related issues & PRs
#6637

### API changes
Added class `pocketmine\event\entity\EntityExtinguishEvent` -  called when an entity on fire gets extinguished

### DIFF
--- a/src/event/entity/EntityExtinguishEvent.php
+++ b/src/event/entity/EntityExtinguishEvent.php
@@ -37,6 +37,8 @@ class EntityExtinguishEvent extends EntityEvent{
 	public const CAUSE_RESPAWN = 3;
 	public const CAUSE_FIRE_PROOF = 4;
 	public const CAUSE_TICKING = 5;
+	public const CAUSE_RAIN = 6;
+	public const CAUSE_POWDER_SNOW = 7;
 
 	public function __construct(
 		Entity $entity,
--- a/src/event/entity/EntityExtinguishEvent.php
+++ b/src/event/entity/EntityExtinguishEvent.php
@@ -31,8 +31,21 @@ use pocketmine\entity\Entity;
  * @phpstan-extends EntityEvent<Entity>
  */
 class EntityExtinguishEvent extends EntityEvent{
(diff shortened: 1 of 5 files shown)


dktapps (dismissed):

dries-c:
Shouldn't this event be cancellable?

dktapps:
Not sure how that would work. Most often it'll be called when the fire ticks reaches 0, and it's not obvious what should be done if the event were cancelled.

A non-cancellable event doesn't require much effort to reason about.

ipad54 (dismissed):
I think this PR is fine in its current form because there are a lot of complicated things to discuss if we were to make this event cancellable, such as what do to when fire ticks reach zero or a player switches to creative. If plugin developers want to keep player burning, they can set him on fire again after the event.

dktapps:
How much of a pain would it be to add support for cause info in the event? I feel like it'd be useful for plugins to be able to tell the difference between timeout extinguish vs extinguish by water/cauldron/whatever.

zSALLAZAR:
Like this?

src/event/entity/EntityExtinguishEvent.php
```diff
@@ -31,8 +31,21 @@
  * @phpstan-extends EntityEvent<Entity>
  */
 class EntityExtinguishEvent extends EntityEvent{
+	public const CAUSE_CUSTOM = 0;
+	public const CAUSE_WATER = 1;
+	public const CAUSE_WATER_CAULDRON = 2;
+	public const CAUSE_SPAWN = 3;
```
dktapps: `RESPAWN` would probably make more sense

IvanCraft623: Although not implemented, powder snow and rain are missing.

dktapps (approved):
LGTM

ShockedPlot7560 (approved):
I would have also made it cancellable, but I see it has been a topic of debate recently. I agree, so it seems fine to me.

Issue #6614, opened 2025-02-01
No Way to Disable Frost Walker Water Freezing


kostamax27:
### Problem description

Currently, the **Frost Walker** enchantment in PocketMine-MP always freezes water when a player with the enchantment walks over it. However, there is no built-in way to prevent this behavior programmatically. This makes it impossible for plugins to disable or modify the effect dynamically.  

### Proposed solution

Introduce an API method or event to allow plugin developers to control water freezing by **Frost Walker**:  

A cancellable event, e.g., `EntityFreezeWaterEvent`, so plugins can prevent freezing.

dktapps:
Implemented by #6673

PR #6673, opened 2025-04-14
Add EntityFrostWalkerEvent

zSALLAZAR:
Add EntityWaterFreezeEvent

### Related issues & PRs
#6614 

### API changes
Added class `pocketmine\event\entity\EntityWaterFreezeEvent` -  called when an entity with boots enchanted with frost walker freezes nearby water.


dktapps:
`EntityWaterFreezeEvent` would fit better with the most commonly used naming conventions (subject object verb event)

src/entity/Living.php
```diff
@@ -733,7 +734,11 @@ protected function applyFrostWalker(int $level) : void{
 				){
 					continue;
 				}
-				$world->setBlockAt($x, $y, $z, $frostedIce);
+				$ev = new EntityWaterFreezeEvent($this, $block);
```
ipad54: It’s probably worth to add `hasHandlers()` check here since it is a hot path.

dktapps:
I'm wondering whether we want to fire an event for every block, vs firing an event once.

e.g. if we had something like EntityFrostWalkerEvent, we could have stuff like getRadius(), getBlocks() etc.

Currently this event doesn't give any context, so the only real thing a plugin can do is just cancel it (which means cancelling somewhere in the ballpark of 30+ events per tick).

zSALLAZAR:
> I'm wondering whether we want to fire an event for every block, vs firing an event once.
> 
> e.g. if we had something like EntityFrostWalkerEvent, we could have stuff like getRadius(), getBlocks() etc.
> 
> Currently this event doesn't give any context, so the only real thing a plugin can do is just cancel it (which means cancelling somewhere in the ballpark of 30+ events per tick).

Good idea but im unsure how this would look like

dktapps:
Right, I didn't realize it would be a pain to include the block list.

I suppose there are 2 choices:
1) make the blocks list mutable and radius immutable
2) get rid of blocks list and allow changing radius only

I'm slightly leaning towards number 2. I think being able to set the radius would be more ergonomic than having to manually recalculate the blocks list. Possibly a `setTargetBlock()` could be useful too, in case plugins want to change the result block from frosted ice to something else (or a different state of frosted ice).

A filter for blocks to change would be nice too (e.g. to allow freezing lava) but I'm not sure how that would work on the API side.

src/entity/Living.php
```diff
@@ -721,19 +722,32 @@ protected function applyFrostWalker(int $level) : void{
...
+		}
+		if($cancelled){
+			return;
+		}
 		for($x = $baseX - $radius; $x <= $baseX + $radius; $x++){
 			for($z = $baseZ - $radius; $z <= $baseZ + $radius; $z++){
 				$block = $world->getBlockAt($x, $y, $z);
 				if(
-					!$block instanceof Water ||
+					!$block instanceof $liquid ||
```
dktapps: This shouldn't be using instanceof. Prefer isSameState() instead and that way you also don't need to do an isSource check.

src/event/entity/EntityFrostWalkerEvent.php
```diff
@@ -0,0 +1,84 @@
...
+namespace pocketmine\event\entity;
+
+use pocketmine\block\Block;
+use pocketmine\block\Liquid;
+use pocketmine\entity\Living;
+use pocketmine\event\Cancellable;
+use pocketmine\event\CancellableTrait;
+
+/**
+ * Called when an entity with boots enchanted with frost walker freezes nearby water.
```
dktapps:
```suggestion
 * Called when an entity moves horizontally while wearing boots enchanted with Frost Walker.
```

dktapps (changes requested):

src/entity/Living.php
```diff
@@ -721,19 +722,31 @@ protected function applyFrostWalker(int $level) : void{
...
+			$ev = new EntityFrostWalkerEvent($this, $radius, $liquid, $targetBlock);
+			$ev->call();
+			$cancelled = $ev->isCancelled();
+			$radius = $ev->getRadius();
+			$liquid = $ev->getLiquid();
+			$targetBlock = $ev->getTargetBlock();
+		}
+		if($cancelled){
+			return;
+		}
```
ShockedPlot7560: this can be moved in the `hasHandlers` if block

ShockedPlot7560 (changes requested):

dktapps (approved):
Looks good to me, seems to work as expected

ShockedPlot7560 (approved):

PR #6674, opened 2025-04-14
Rename `PlayerExhaustEvent` to `EntityExhaustEvent`

zSALLAZAR:
Renamed `PlayerExhaustEvent` to `EntityExhaustEvent`

### Related issues & PRs
#5086 
#5089 

### API changes
Renamed `pocketmine\event\player\PlayerExhaustEvent` to `pocketmine\event\entity\EntityExhaustEvent`
Removed `EntityExhaustEvent::getPlayer()`


dktapps (dismissed):

ShockedPlot7560 (commented):
Is it not better to call it as HumanExhaustEvent ? As it is not really all entities which is linked to this event.

dktapps:
Not sure. It could probably be applied to other entities without much changes.

src/event/entity/EntityExhaustEvent.php
```diff
@@ -47,20 +46,13 @@ class PlayerExhaustEvent extends EntityEvent implements Cancellable{
 	public const CAUSE_CUSTOM = 11;
 
 	public function __construct(
-		protected Human $human,
+		Human $human,
```
IvanCraft623: Since the event name implies this is no longer limited to Humans, it would be nice to remove this limitation from the __construct so that other entities can use the event.

dktapps (approved):

ShockedPlot7560 (approved):

PR #6676, opened 2025-04-20
Item nbtDeserialize change

### LINKED ISSUE #5868: Server crashed while loging in. "No deserializer found for ID minecraft:crimson_button"
Link to crashdump:  https://crash.pmmp.io/view/8811670

I have 2 players logging in simultaneously for a long time.
After updating the server one of the two cant log in anymore or the server will crash.

I removed all plugins just to be sure.
But it still crashes.

ghost:
This change will destroy an unknown item in the player's inventory.

### Related issues & PRs
Fix: #5868 

## Tests
I tried it on my local server and it works!

### DIFF
--- a/src/item/Item.php
+++ b/src/item/Item.php
@@ -753,6 +753,10 @@ class Item implements \JsonSerializable{
 	 */
 	public static function nbtDeserialize(CompoundTag $tag) : Item{
 		$itemData = GlobalItemDataHandlers::getUpgrader()->upgradeItemStackNbt($tag);
+		if($itemData === null){
+			return VanillaItems::AIR();
+		}
+
 		try{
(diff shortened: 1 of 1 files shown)


dries-c:
Let's say a ItemTypeDeserializeException would happen in the future because of invalid implementations no error would be thrown, the item would be replaced into air and we would never know or be able to reverse it. I don't think this is a good solution

dktapps:
Yep, I also don't like rethrowing generic Exception as SavedDataLoadingException

ghost:
So what can we do in that case? Do we really have to delete the player's data? That's really a bad thing 😞

bonbionseker:
> So what can we do in that case? Do we really have to delete the player's data? That's really a bad thing 😞

dktapps:
No, but there are better ways to fix it than this

PR #6678, opened 2025-04-26
Item: make setter methods fluent

SmallkingDev:
Make `Item` setter methods fluent

### API Changes
- Updated the return types of the following `Item` setter methods to `$this`:
  - `setCanPlaceOn`
  - `setCanDestroy`
  - `setKeepOnDeath`

#### Before:
```php
$item = VanillaItems::DIAMOND_PICKAXE();
$item->setCanPlaceOn([]);
$item->setCanDestroy([]);
$item->setKeepOnDeath(true);
```

#### After:
```php
$item = VanillaItems::DIAMOND_PICKAXE()
    ->setCanPlaceOn([])
    ->setCanDestroy([])
    ->setKeepOnDeath(true);
```


dktapps:
These changes are BC breaking for inheritors if they override any of these functions

SmallkingDev:
> These changes are BC breaking for inheritors if they override any of these functions

The chance of someone doing this must be slim, but it's a valid concern.

dktapps:
Semver doesn't care about chances, objectively it is BC breaking, so it can't target a minor version.

dktapps (approved):

ShockedPlot7560 (approved):

PR #6682, opened 2025-04-30
Extract GeneratorExecutor system from World, v2

dktapps:
Net result is basically the same as #6668, but with way less complexity for the GeneratorExecutor side.


ShockedPlot7560 (approved):
Looks good to me, nothing more to say

Issue #6508, opened 2024-11-16
Break cycle between `Command` and `CommandMap` (aliases, label)


dktapps:
## Description
Have `CommandMap` track commands' active aliases and "label" (primary alias) instead of `Command`

## Justification
Since a command's active aliases are tracked by `Command` internally, a `Command` needs to ensure it is only registered to a single `CommandMap` at a time.

This is a dumb limitation and requires extra complexity. We could just have the active aliases tracked by `CommandMap` itself, since they're only actually important to the `CommandMap` anyway.

Going further, aliases could be entirely removed from `Command`. They could just be passed as a `string[]` to `CommandMap->register()`.

In addition, if #6506 were implemented, `Command->getName()` (primary alias) could be deleted too, as its only actual use is for `onCommand()`.

https://github.com/pmmp/PocketMine-MP/blob/48a908ee8caf3d41df694def05912dc5867504d9/src/command/Command.php#L139-L181
https://github.com/pmmp/PocketMine-MP/blob/48a908ee8caf3d41df694def05912dc5867504d9/src/command/Command.php#L205-L210

PR #6685, opened 2025-05-04
Rework command alias handling, step 1

dktapps:
This PR started out as an effort to decouple Command and CommandMap, but it's turned into a bit more than that.

A summary of changes:

## UX
- Added `cmdalias create`, `cmdalias delete` and `cmdalias list` commands
- `/help` now shows prefixed names such as `pocketmine:help`
- Prefixed command name (e.g. `pocketmine:help`) are now visible to Minecraft clients
- Permission denied messages are now able to show more useful context when e.g. checking subcommand permissions
- Multiple commands claiming an alias make the alias unusable (an error will be shown when used telling the user to pick from the namespaced names), instead of whichever plugin loaded last getting lucky

## API
- Added `CommandAliasMap`, which handles mapping of aliases to namespaced command IDs
- Added `CommandSender->getCommandAliasMap()` for user-specific aliases
- Added `CommandMap->getAliasMap()` for global fallback aliases
- `Command` no longer tracks its own registered aliases (now the job of `CommandMap`), breaking circular dependency
- Aliases must now be provided to `CommandMap->register()`
- Aliases can now be individually registered and unregistered without re-registering/unregistering the whole command using `CommandAliasMap` APIs
- Aliases are no longer namespaced, only the main command name (e.g. `pocketmine:?` is now gone while `pocketmine:help` still exists)
- `Command` now requires a `$namespace` parameter, which replaces the old `$fallbackPrefix` parameter of `register()`. It should be set to the name of the plugin.

Relevant issues
- #6508 
- #3371 - this PR doesn't implement storage, but allows configuration of per-user aliases during server runtime


src/command/CommandMapEntry.php
```diff
@@ -0,0 +1,48 @@
...
+		public readonly array $aliases
+	){}
+
+	public function getPreferredAlias() : string{
+		return $this->aliases[0];
+	}
+
+	public function getUsage() : Translatable|string{
+		//TODO: usage messages ought to use user-specified alias, not command preferred
+		//command-preferred is confusing if the user used a different alias
```
NTT1906: maybe adding an alias input parameter to the getUsage() method.
also update the default command usage in the language file from:
> commands.tp.usage=/tp [target player] <destination player> OR /tp [target player] <x> <y> <z> [<y-rot> <x-rot>]

to:
> commands.tp.usage=/{%0} [target player] <destination player> OR /{%0} [target player] <x> <y> <z> [<y-rot> <x-rot>]

then, translate `{%0}` to the used alias?

dktapps: Yes, it's just a lot of work

NTT1906:
image
> `Prefixed aliases are now visible to Minecraft clients`

I don't like the look of it. Sending fully prefixed/namespaced commands to clients feels spammy. It should be cleaner by default.

1. The last registered command should take the name (without the namespace), except for help.
2. Only show namespace versions if more than 1 command with the same name.
3. Maybe add an option for the command to choose whether they wanted to show or not the namespace.
4. While the `pocketmine` namespace is consistent, using `pmmp` is a shorter, cleaner alternative.
5. Maybe a bad idea, as it might cause confusion, but by allowing namespace commands to still work silently—no hints, they can still be executed.

6. Yet another bad idea, and idk how it could work with the current command system in PMMP. Maybe introduce some kind of bundled command? Create a super command class where commands with the same name get bundled into a single command data. This allows for registering multiple overloads at the cost of losing command descriptions and needs a way to separate the command args when passing down the command args.

dktapps:
Yeah, I'm not sure about the fully-qualified names either. I think the way that those are handled probably needs to be rethought.
For example I don't like prefixing the aliases. Probably we should only prefix the primary label.

Longer term I think I want to have those move towards being internalised command IDs which are used by aliases to locate a command (see #3371), rather than being able to be directly invoked. But that's too big of a change for one PR.

dktapps:
> I don't like the look of it. Sending fully prefixed/namespaced commands to clients feels spammy.

To be honest this was more of a side effect of changing the tracking of the registered aliases. Previously the command registration wasn't aware of its own prefixed aliases, which lead to all kinds of interesting surprises for people trying to block commands by name rather than using permissions, since the aliases weren't visible in `/help` nor in client-side hints. (Not that they should have been doing that, but it's still surprising to have these hidden things.)

Filtering out the prefixed aliases is a bit problematic since we currently don't restrict aliases or fallback prefixes from containing `:`, so there's no foolproof way to tell them apart.

src/command/CommandMapEntry.php
dktapps: Clarification on this for posterity: It's in the plans to have usage messages generated from command signatures (once we have a proper arg parsing system), so it doesn't seem to make much sense to make a lot of mess with this crappy legacy system

PR #6687, opened 2025-05-07
Protocol changes for 1.21.80

dries-c:
This PR updates PM5 to 1.21.80

## Backwards compatibility

See https://github.com/pmmp/BedrockProtocol/pull/312


src/world/biome/model/BiomeDefinitionEntryData.php
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\world\biome\model;
+
+/**
+ * Model for loading biome definition entries data from JSON.
+ */
+final class BiomeDefinitionEntryData implements \JsonSerializable{
```
dktapps: Don't think this is needed since all the fields are public anyway

src/world/biome/model/BiomeDefinitionEntryData.php
```diff
@@ -0,0 +1,54 @@
...
+	/**
+	 * @var string[]
+	 * @phpstan-var list<string>
+	 */
+	public array $tags;
+
+	/**
+	 * @return mixed[]
+	 */
+	public function jsonSerialize() : array{
```
dktapps: Same again here

src/world/biome/model/ColorData.php
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\world\biome\model;
+
+/**
+ * Model for loading color data from JSON.
+ */
+final class ColorData implements \JsonSerializable{
```
dktapps: Same again here

src/world/biome/model/ColorData.php
```diff
@@ -0,0 +1,41 @@
...
+	public int $r;
+	public int $g;
+	public int $b;
+	public int $a;
+
+	/**
+	 * @return mixed[]
+	 */
+	public function jsonSerialize() : array{
```
dktapps: and here

src/world/biome/model/BiomeDefinitionEntryData.php
```diff
@@ -0,0 +1,54 @@
...
+	public float $whiteAshDensity;
+	public float $depth;
+	public float $scale;
+	public ColorData $mapWaterColour;
+	public bool $rain;
+	/**
+	 * @var string[]
+	 * @phpstan-var list<string>
+	 */
+	public array $tags;
```
dktapps, line 68: These fields should all be marked as @required

src/world/biome/model/ColorData.php
```diff
@@ -0,0 +1,41 @@
...
+namespace pocketmine\world\biome\model;
+
+/**
+ * Model for loading color data from JSON.
+ */
+final class ColorData implements \JsonSerializable{
+	public int $r;
+	public int $g;
+	public int $b;
+	public int $a;
```
dktapps: These need to be marked as @required

dktapps (changes requested):

IcyEndymion004:
Just tried running this, and got the following error

[Crash.txt](https://github.com/user-attachments/files/20093791/New.Text.Document.txt)

dktapps:
Yeah, that'll be because of this commit: https://github.com/pmmp/BedrockData/commit/a8932e40388c7280783ab78a8853d9491cb5dd40

dktapps (approved):

PR #6691, opened 2025-05-10
BlockStateData: fix version

Gaprix:


dktapps:
This is wrong, you can verify it by loading a world from vanilla. The version doesn't change unless there were changes to blockstates.

PR #6702, opened 2025-05-23
Entity: make stepHeight accessable

### LINKED ISSUE #6672: Make Entity::$stepHeight flexible
### Problem description

Right now, it is always `0.0` for all entities, only Player overrides this variable to `0.6` - no other entities modify this behavior inside pmmp. This renders the whole stepping code in the method `Entity::move()` useless for most entities, except maybe custom ones.
I personally want to be able to modify the step height of some entities dynamically for the following reasons:
- I apply the scale attribute to the entity, for example for baby Players. I do not want these to be able to step up the same height as full grown mobs
- Some of my vehicle entities can walk up a full block all the time, for example horses, whilst others should depend on their current behavior components

Since you can't easily override the check in the `Entity::move()` without having to copy the whole method (because parent call still checks the original value) and your only chance to really override the values is reflections i propose the following:

### Proposed solution

```
	public function getStepHeight() : float{
		return $this->stepHeight;
	}

	public function setStepHeight(float $stepHeight) : void{
		$this->stepHeight = $stepHeight;
	}
```

`public float $stepHeight = 0.0;`

ItzxDwi:
### Related issues & PRs
* Fixes #6672 

### API changes
Added:
```Entity->setStepHeight(float $stepHeight)```
```Entity->getStepHeight()```

### Behavioural changes
None

## Backwards compatibility
None

### DIFF
--- a/src/entity/Entity.php
+++ b/src/entity/Entity.php
@@ -1187,12 +1187,12 @@ abstract class Entity{
 
 			$moveBB->offset(0, 0, $dz);
 
-			if($this->stepHeight > 0 && $fallingFlag && ($wantedX !== $dx || $wantedZ !== $dz)){
+			if($this->getStepHeight() > 0 && $fallingFlag && ($wantedX !== $dx || $wantedZ !== $dz)){
 				$cx = $dx;
 				$cy = $dy;
 				$cz = $dz;
 				$dx = $wantedX;
-				$dy = $this->stepHeight;
+				$dy = $this->getStepHeight();
 				$dz = $wantedZ;
 
 				$stepBB = clone $this->boundingBox;
@@ -1260,6 +1260,14 @@ abstract class Entity{
 		//TODO: vehicle collision events (first we need to spawn them!)
 
 		Timings::$entityMove->stopTiming();
+	}
+
+	public function setStepHeight(float $stepHeight) : void {
+		$this->stepHeight = $stepHeight;
+	}
+
+	public function getStepHeight() : float {
+		return $this->stepHeight;
 	}
 
 	protected function checkGroundState(float $wantedX, float $wantedY, float $wantedZ, float $dx, float $dy, float $dz) : void{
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/entity/Entity.php [hunk 1]
```diff
@@ -1187,12 +1187,12 @@ protected function move(float $dx, float $dy, float $dz) : void{
...
 			$moveBB->offset(0, 0, $dz);
 
-			if($this->stepHeight > 0 && $fallingFlag && ($wantedX !== $dx || $wantedZ !== $dz)){
+			if($this->getStepHeight() > 0 && $fallingFlag && ($wantedX !== $dx || $wantedZ !== $dz)){
 				$cx = $dx;
 				$cy = $dy;
 				$cz = $dz;
 				$dx = $wantedX;
-				$dy = $this->stepHeight;
+				$dy = $this->getStepHeight();
```
dktapps: Could move this into a variable so we don't call the function twice

src/entity/Entity.php
```diff
@@ -1262,6 +1262,14 @@ protected function move(float $dx, float $dy, float $dz) : void{
 		Timings::$entityMove->stopTiming();
 	}
 
+	public function setStepHeight(float $stepHeight) : void {
```
dktapps:
```suggestion
	public function setStepHeight(float $stepHeight) : void{
```

src/entity/Entity.php
```diff
@@ -1262,6 +1262,14 @@ protected function move(float $dx, float $dy, float $dz) : void{
...
+		$this->stepHeight = $stepHeight;
+	}
+
+	public function getStepHeight() : float {
```
dktapps:
```suggestion
	public function getStepHeight() : float{
```

dktapps (approved):

dktapps (approved):

ShockedPlot7560 (approved):

dktapps:
This shouldn't have been merged into stable

ItzxDwi:
> This shouldn't have been merged into stable

Since you revert the changes, should i create a new pull request targeting ```minor-next``` with the same changes?

dktapps:
No, I only reverted it for stable

PR #6706, opened 2025-05-25
Implement underwater explosions

ItzxDwi:
### Related issues & PRs
* Related to #6456 PR abandoned 

### API changes
Explosion now accepts an optional new parameter:
```$excludedBlockTypeIds``` list of array containing excluded block type ids

### Behavioural changes
Underwater TNT now works in underwater

## Backwards compatibility
None

## Tests
https://youtu.be/NfTTrxWKHas

### CODE REVIEW

**Comment by ItzxDwi** [comment]
I don't know why the texture of the underwater tnt the same as the regular one.

**Comment by nicholass003** [comment]
> I don't know why the texture of the underwater tnt the same as the regular one.

Iirc this is a client issue, and I've also reported it to bugs.mojang.com but when I tried to search for it, it's now inaccessible.
https://github.com/pmmp/PocketMine-MP/issues/6494

**Review by ShockedPlot7560 — DISMISSED** [review]
Looks OK to me

**Comment by ItzxDwi** [comment]
> Looks OK to me

Hey, could you also review pull #6702

**Comment by dktapps** [comment]
If I remember, this doesn't happen in vanilla anymore, but I wasn't able to find the cause

**Comment by dktapps** [comment]
The underwater TNT is not supposed to destroy water blocks. It's not enough to override the blast resistance.

The previous PR attempted to fix this, but created additional issues (since the blast force wouldn't decrease). This needs to be addressed.

File: src/world/Explosion.php [hunk 1]
```diff
@@ -71,6 +81,8 @@ public function __construct(
 			throw new \InvalidArgumentException("Explosion radius must be greater than 0, got $radius");
 		}
 		$this->subChunkExplorer = new SubChunkExplorer($this->world);
+
+		$this->excludedBlockTypeIds = array_flip($excludedBlockTypeIds);
```
dktapps: Prefer `array_fill_keys($keys, true)` instead of doing this, it has less confusing behaviour

ItzxDwi:
> The underwater TNT is not supposed to destroy water blocks. It's not enough to override the blast resistance.
> 
> The previous PR attempted to fix this, but created additional issues (since the blast force wouldn't decrease). This needs to be addressed.

How would i implement that?

ItzxDwi:
New test:
https://youtu.be/kZ4_wHBXtuU

ItzxDwi:
@dktapps, sorry for the ping, would you mind reviewing this PR when you have a moment?

src/world/Explosion.php
```diff
@@ -83,6 +99,7 @@ public function explodeA() : bool{
 		}
 
 		$blockFactory = RuntimeBlockStateRegistry::getInstance();
+		$waterTypeIds = BlockTypeIds::WATER;
```
dktapps: Why is this here?

dktapps:
I think the way this is implemented needs to be reconsidered, since the current way will have a negative effect on performance for underwater TNT. Copying a new block instance from `RuntimeBlockStateRegistry` every time we want to check if a block has an overridden blast resistance is suboptimal. In addition it'll do the override check potentially multiple times if different rays touch the same block.

Also, the overridden blast resistance should be -1 to make it treat water like air. (Technically this is the same as a `continue` which your original version did do... sorry for the confusion about that)

src/world/Explosion.php
ItzxDwi: > Why is this here?

Placed outside the loop for avoiding redundant call

dktapps: The place where it's used seems redundant too.

ItzxDwi:
> I think the way this is implemented needs to be reconsidered, since the current way will have a negative effect on performance for underwater TNT. Copying a new block instance from `RuntimeBlockStateRegistry` every time we want to check if a block has an overridden blast resistance is suboptimal. In addition it'll do the override check potentially multiple times if different rays touch the same block.
> 
> Also, the overridden blast resistance should be -1 to make it treat water like air. (Technically this is the same as a `continue` which your original version did do... sorry for the confusion about that)

No rush, but could tou take a look at the PR when you have a moment? I made the suggested changes

ItzxDwi:
@dktapps @ShockedPlot7560  Sorry to bother you both and for the ping, but would you mind giving this PR another look when you have a chance? It's been a couple of weeks, and I've addressed the feedback from Dylan, Thanks!

src/world/Explosion.php
```diff
@@ -68,11 +70,22 @@ class Explosion{
...
+	/**
+	 * @param int[] $excludedBlockTypeIds
+	 * @phpstan-param array<int> $excludedBlockTypeIds
+	 */
 	public function __construct(
 		public Position $source,
 		public float $radius,
 		private Entity|Block|null $what = null,
-		private float $fireChance = 0.0
+		private float $fireChance = 0,
```
ShockedPlot7560, line 87: Why `0.0` was changed ?

src/world/Explosion.php
```diff
@@ -68,11 +70,22 @@ class Explosion{
...
 
+	/**
+	 * @var true[] excludedBlockTypeIds => dummy
+	 * @phpstan-var array<int, true>
+	 */
+	private array $excludedBlockTypeIds = [];
+
+	/**
+	 * @param int[] $excludedBlockTypeIds
+	 * @phpstan-param array<int> $excludedBlockTypeIds
```
ShockedPlot7560, line 81: The phpstan-params is redundant

src/world/Explosion.php
```diff
@@ -135,7 +150,15 @@ public function explodeA() : bool{
...
 
+							if(count($this->excludedBlockTypeIds) > 0){
+								$typeId = $blockFactory->fromStateId($state)->getTypeId();
+								if(isset($this->excludedBlockTypeIds[$typeId])){
+									continue;
+								}
+							}
+
 							$blastResistance = $blockFactory->blastResistance[$state] ?? 0;
+
```
ShockedPlot7560, line 161: unrelated change

src/world/Explosion.php
```diff
@@ -86,6 +99,8 @@ public function __construct(
 			throw new \InvalidArgumentException("Explosion radius must be greater than 0, got $radius");
 		}
 		$this->subChunkExplorer = new SubChunkExplorer($this->world);
+
+		$this->excludedBlockTypeIds = array_fill_keys($excludedBlockTypeIds, true);
```
ShockedPlot7560, line 103:
```suggestion
		$this->excludedBlockTypeIds = array_fill_keys($excludedBlockTypeIds, true);
```

ShockedPlot7560 (changes requested):
I don't know more the goal of the underwater TNT. As I understand, it's a normal TNT exploding where if it's in water, blocks wouldn't be ignored to exploded ?

If it is what it is, in vanilla, if underater TNT exploded in a non still water, does it have the same effect ? Or not ? If this happen only in a still water, this current implementation will be a problem.

dktapps:
> If it is what it is, in vanilla, if underater TNT exploded in a non still water, does it have the same effect ? Or not ? If this happen only in a still water, this current implementation will be a problem.

Still and flowing come under the same type ID in PM, so it would work in both.

> As I understand, it's a normal TNT exploding where if it's in water, blocks wouldn't be ignored to exploded ?

Water has a blast resistance high enough that the origin block will normally suppress an explosion underwater. Underwater TNT treats water like it's air and skips over it when checking for blocks to destroy.

nicholass003:
Should my PR be closed, because there are 2 similar PRs?
https://github.com/pmmp/PocketMine-MP/pull/6456

Issue #6704, opened 2025-05-24
Allowing custom command to handle/parse the command line by itself.


NTT1906:
### Problem description

Currently, SimpleCommandMap hardcoded the handling of command line when parsing them to args and there is not a non-hacky way to handle the raw inputs data without messing too much with PMMP internal.

### Proposed solution

Allow command itself to dispatch/parse command line to args by itself via an internal method that can be overridden. The SimpleCommandMap only take the first token from the command line to find the command entry.
```php
<?php
class Command{
    ...
    public function parseCommandLine(string $commandLine) : array{
        return CommandHelper::parseQuoteAware($commandLine);
    }
    ...
}

class CustomCommand extends Command{
    public function parseCommandLine(string $commandLine) : array{
        // Custom parse
    }
}

class SimpleCommandMap{
  ...
  public function dispatch(CommandSender $sender, string $commandLine) : bool{
    $sentCommandLabel = "";
    $parts = preg_split('/\s+/u', $commandLine, 2);
    if($parts !== false) {
      $sentCommandLabel = $parts[0];
      if ($target = $this->getEntry($sentCommandLabel)) !== null){
         $args = $target->command->handleCommandLine($parts[1]);
         // Command execute and stuff
      }
    }
    ...
}
```

### Alternative solutions or workarounds

A plugin can create an extend of SimpleComamndMap,  move over the data and then use reflection API/Closure bind to change the commandMap in Server instance.

dktapps:
This seems like an OK idea, although I wonder why have it pass the args back out to the `CommandMap` at that point. A wrapper `executeRaw()` would also be fine. This would also permit alternative command signatures that don't have to use `string[]`.

dktapps:
Superseded by #6837, which will allow `Parameter`s to decide how to do arg parsing with significantly more upside.

Issue #6518, opened 2024-11-17
Command overloading, & automatic command argument parsing & validation system


dktapps:
## Description
This has been discussed many times over the years, but it's never been done in a satisfying way.

Currently, all commands receive their arguments as `string[]` and have to do boilerplate validation themselves, including argument count checking, conversion to appropriate types (e.g. targets), checking if the targets exist/are valid/etc, checking if the parameters are part of an accepted set of values, etc.

You need only look at [this](https://github.com/pmmp/PocketMine-MP/blob/b2aa6396c3cc2cafdd815eacc360e1ad89599899/src/command/defaults/TeleportCommand.php#L65-L78) nasty code to see why we need this. I remember spending hours and hours to get to those few lines of code.

Before I get into this, I'll mention that it's possible there are general-purpose libraries for supporting this kind of feature. Processing commands is not unique to Minecraft, as it's also a common need of applications to parse CLI arguments, which would have pretty much the same requirements.

It should be possible to implement this system without breaking BC, even just within a custom `Command::execute()`, as this is essentially a subcommands system combined with automatic argument validation & parsing. However, it may be desirable to introduce an `OverloadedCommand` to allow overload information to be exported for use in features like #6517.

This issue proposes the following:

- Commands should be allowed to have multiple executors (or overloads), which each declare the types and constraints of arguments that they want in their function signature (and perhaps attributes if needed), like so:
  - tp to player: `function(Player $sender, Player $target)`
  - tp to xyz: `function(Player $sender, Vector3 $target)`
  - tp to xyz+rotation: `function(Player $sender, Vector3 $target, float $yaw, float $pitch)`
  - tp player1 to player2: `function(CommandSender $sender, Player $subject, Player $target)`
  - tp player1 to xyz: `function(CommandSender $sender, Player $subject, Vector3 $target)`
  - tp player1 to xyz+rotation: `function(CommandSender $sender, Player $subject, Vector3 $target, float $yaw, float $pitch)`
- Requirements for each overload are extracted from the function signature using reflection. It may be necessary to support attributes for certain things, like dynamic enums.
- Overload is selected based on the number of args given, and whether the given args can be parsed according to the requirements of the overload. e.g. in `tp player1 player2`, `player2` cannot be parsed as a `Vector3`, so overload 2 would be skipped. Overloads should be checked from most arguments to least arguments.
[...]

Issue #4193, opened 2021-04-27
CommandEvent should provide resolved command and args


SOF3:
## Motivatoin
CommandEvent is not very useful because it only provides the whole command string and allows plugins to overwrite the whole command string, after which the command map parses the command string again. This does not offer plugins a reliable way to know which command it is resolved into or what arguments are passed.

## Proposed change
It is proposed that we fire `CommandEvent` in `CommandMap::dispatch` instead of `Server::dispatchCommand`, *after* resolving the command and parsing arguments. Then we remove the `setCommand` function from `CommandEvent`, and instead add `getArgs` and `setArgs` to the event so that plugins can reliably change the final arguments passed into the command.

dktapps:
I'm not sure this is a good idea. Right now, CommandEvent is fully independent of any command parser, which allows it to be re-routed to a different command system where it might be differently parsed.

SOF3:
The point of this proposal is to allow the different parser to be used to parse the command. The `getCommand()` method is still kept, with which the different parser can generate its string array to pass into the `setArgs` function. Meanwhile, the `getArgs` method can be used at a higher priority to use `getArgs` directly.

dktapps:
What exactly is the use case?

SOF3:
Example 1: a selector plugin would want to replace `@p` to sender name like this:

```php
public function _onCommand(CommandEvent $event) {
    $args = $event->getArgs();
    foreach($args as &$arg) {
        if($arg === "@p") $arg = $event->getSender()->getName();
    }
    unset($arg);
    $event->setArgs($args);
}
```

Example 2: an alternative command parser (e.g. one that uses `\` + space instead of `""`) would want to implement their functionality like this:

```php
public function _onCommand(CommandEvent $event) {
    $command = $event->getCommand();
    $args = explode(" ", $command);
    for($i = 0; $i < count($args); $i++) {
        while(substr($args[$i], -1) === "\\")) {
            $next = array_splice($args, $i + 1, 1, []);
            $args[$i] = substr($args, 0, -1) . " " . $next;
        }
    }
    $event->setArgs($args);
}
```

In either case, it is more preferable to use `setArgs` instead of `setCommand`.

dktapps:
Reopening this as migrating back from Discussions (which sucks, by the way).

dktapps:
Not too sure how this will interact with #6837 currently. Since the arguments won't be `string[]` anymore.

SOF3:
[...]

PR #6708, opened 2025-05-26
Allowing custom command to handle raw command line inputs

NTT1906:
Current PMMP handles the parsing of command lines in a hardcoded way without any non-hacky method for commands to handle the raw inputs. This PR tweaks the API a little so it will allow custom commands to do such things without harming the performance.

### Related issues & PRs
* Related to #6704

### API changes
- Added a new method named `Command->executeRaw()` to Command class. This allowing custom command to handle the raw command inputs by override the parent method.

## Backwards compatibility
Yes

### DIFF
--- a/src/command/Command.php
+++ b/src/command/Command.php
@@ -82,8 +82,8 @@ abstract class Command{
 	}
 
 	public function executeRaw(CommandSender $sender, string $commandLabel, string $argLine) : void{
-		$args = CommandStringHelper::parseQuoteAware($argLine);
 		if($this->testPermission($sender)){
+			$args = CommandStringHelper::parseQuoteAware($argLine);
 			$this->execute($sender, $commandLabel, $args);
 		}
 	}
--- a/src/command/Command.php
+++ b/src/command/Command.php
@@ -27,6 +27,7 @@ declare(strict_types=1);
 namespace pocketmine\command;
 
 use pocketmine\command\utils\CommandException;
+use pocketmine\command\utils\CommandStringHelper;
 use pocketmine\lang\KnownTranslationFactory;
 use pocketmine\lang\Translatable;
 use pocketmine\permission\PermissionManager;
@@ -78,6 +79,13 @@ abstract class Command{
 		$this->setDescription($description);
 		$this->usageMessage = $usageMessage ?? ("/" . $name);
 		$this->setAliases($aliases);
+	}
+
+	public function executeRaw(CommandSender $sender, string $commandLabel, string $argLine) : void{
+		$args = CommandStringHelper::parseQuoteAware($argLine);
+		if($this->testPermission($sender)){
+			$this->execute($sender, $commandLabel, $args);
+		}
 	}
 
 	/**
--- a/src/command/Command.php
+++ b/src/command/Command.php
@@ -82,10 +82,8 @@ abstract class Command{
 	}
 
 	public function executeRaw(CommandSender $sender, string $commandLabel, string $argLine) : void{
-		if($this->testPermission($sender)){
-			$args = CommandStringHelper::parseQuoteAware($argLine);
-			$this->execute($sender, $commandLabel, $args);
-		}
+		$args = CommandStringHelper::parseQuoteAware($argLine);
+		$this->execute($sender, $commandLabel, $args);
 	}
 
 	/**
--- a/src/command/Command.php
+++ b/src/command/Command.php
@@ -81,8 +81,8 @@ abstract class Command{
 		$this->setAliases($aliases);
 	}
 
-	public function executeRaw(CommandSender $sender, string $commandLabel, string $argLine) : void{
-		$args = CommandStringHelper::parseQuoteAware($argLine);
+	public function executeRaw(CommandSender $sender, string $commandLabel, string $rawArgs) : void{
+		$args = CommandStringHelper::parseQuoteAware($rawArgs);
 		$this->execute($sender, $commandLabel, $args);
 	}
 
--- a/src/command/SimpleCommandMap.php
+++ b/src/command/SimpleCommandMap.php
@@ -207,26 +207,25 @@ class SimpleCommandMap implements CommandMap{
 	}
 
 	public function dispatch(CommandSender $sender, string $commandLine) : bool{
-		$args = CommandStringHelper::parseQuoteAware($commandLine);
+		$sentCommandLabel = "";
+		if(($parts = preg_split('/\s+/u', $commandLine, 2)) !== false){
+			$sentCommandLabel = $parts[0];
+			if(($target = $this->getCommand($sentCommandLabel)) !== null){
+				$timings = Timings::getCommandDispatchTimings($target->getLabel());
+				$timings->startTiming();
 
-		$sentCommandLabel = array_shift($args);
-		if($sentCommandLabel !== null && ($target = $this->getCommand($sentCommandLabel)) !== null){
-			$timings = Timings::getCommandDispatchTimings($target->getLabel());
-			$timings->startTiming();
-
-			try{
-				if($target->testPermission($sender)){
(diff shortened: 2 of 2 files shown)


src/command/Command.php
```diff
@@ -80,6 +81,13 @@ public function __construct(string $name, Translatable|string $description = "",
 		$this->setAliases($aliases);
 	}
 
+	public function executeRaw(CommandSender $sender, string $commandLabel, string $argLine) : void{
+		$args = CommandStringHelper::parseQuoteAware($argLine);
```
alvin0319: I think we can move `$args` below `if` statement so that we don't waste CPU time doing `preg_*` methods when the sender doesn't have permission to execute a command.

NTT1906: It wouldn't be much but yeah, it should. Thank you.

src/command/Command.php
```diff
@@ -80,6 +81,13 @@ public function __construct(string $name, Translatable|string $description = "",
 		$this->setAliases($aliases);
 	}
 
+	public function executeRaw(CommandSender $sender, string $commandLabel, string $argLine) : void{
+		if($this->testPermission($sender)){
```
dktapps: Permission checking should not be moved here

dktapps (dismissed):

dktapps (commented):
I feel like the SimpleCommandMap changes are more than what's needed. The diff shouldn't need to be so many lines.

NTT1906:
> I feel like the SimpleCommandMap changes are more than what's needed. The diff shouldn't need to be so many lines.

- Eh... Isn't the result check of `preg_split` return shifted a big part of the code? and I don't know if there exists any cases of it return false.
Edit: there is :l
```php
$cmd = "\xC3"; // invalid UTF-8 byte
var_dump(preg_split('/\s+/u', $cmd, 2)); //  false
```

- I did set `$sentCommandLabel` to `""`, as they got null coalesced as `""` in the later. If you don't want that then I will revert it back to `null`.

Also, wala - PHP magic
```php
$parts = preg_split('/\s+/u', $commandLine, 2) ?: [""]; // ensure the $parts will always be an array of string
$sentCommandLabel = $parts[0];
if(!empty($sentCommandLabel) && ($target = $this->getCommand($sentCommandLabel)) !== null){
	$timings = Timings::getCommandDispatchTimings($target->getLabel());
	$timings->startTiming();

	try{
		if($target->testPermission($sender)){
			$target->executeRaw($sender, $sentCommandLabel, $parts[1] ?? "");
		}
	}catch(InvalidCommandSyntaxException $e){
		$sender->sendMessage($sender->getLanguage()->translate(KnownTranslationFactory::commands_generic_usage($target->getUsage())));
	}finally{
		$timings->stopTiming();
	}
	return true;
}

$sender->sendMessage(KnownTranslationFactory::pocketmine_command_notFound($sentCommandLabel, "/help")->prefix(TextFormat::RED));
return false;
```

dktapps:
You're overthinking it. A simple `explode()` call would suffice.

NTT1906:
> You're overthinking it. A simple `explode()` call would suffice.

The orginal code `CommandStringHelper::parseQuoteAware()` used regex to split what ever regex treated as space (`\s`) with flag unicode turned on. Using explode would be simple but that will only do that for (` `, `\t`, manual listing...).

`preg_splits` was there to keep up the consistency with `CommandStringHelper::parseQuoteAware()`

Kinda unrelated to this PR, but if only ascii whitespace ` ` is acceptable -  I would like for `CommandStringHelper::parseQuoteAware()` to only accept ascii whitespace. Current regex accept all ascii spaces and all unicode spaces...

dktapps:
This is really overcomplicating it. All that's needed is to explode by spaces into 2 parts. Trim the parts and be done with it. No one is putting a tab after the command name.

src/command/Command.php
```diff
@@ -80,6 +81,11 @@ public function __construct(string $name, Translatable|string $description = "",
 		$this->setAliases($aliases);
 	}
 
+	public function executeRaw(CommandSender $sender, string $commandLabel, string $argLine) : void{
```
dktapps: `$rawArgs` would be a more obvious name for this parameter

dktapps (commented):
Looks mostly OK now

ShockedPlot7560:
To be noted, if we changed to that, we probably need to switch execute to protected. If its public, This suggests that one or other could be called whereas the plugin can implement either.
BC is keeped with the execute (while plugins don't override executeRaw) but execute can be called alone without executeRaw.

NTT1906:
> To be noted, if we changed to that, we probably need to switch execute to protected. If its public, This suggests that one or other could be called whereas the plugin can implement either. 

Well, eh... Ain't those devs decided to code it that way? Maybe adding some docs?
> BC is keeped with the execute (while plugins don't override executeRaw) but execute can be called alone without executeRaw.

I don't think `Command::execute()` being called alone is a problem as they still provide args as array.
The whole point of this `Command::executeRaw()` to me is not an alternative version of execute but making the handling of raw args possible.

ShockedPlot7560:
I share your point of you in the handling of raw args. But, we can't rely on only docs where a creepy dev can create a plugin calling execute by its way skipping all rawArgs calls
Anyway, with your implementation, for me, breaking BC is needed by switching visibility
Maybe we can find a better way because I'm not a big find breaking a such large codebase.

NTT1906:
> I share your point of you in the handling of raw args. But, we can't rely on only docs where a creepy dev can create a plugin calling execute by its way skipping all rawArgs calls 

As I mentioned before, that creepy dev already coded it like that. If you want to stop such practices, you can only stop them from being released into productions or advise them using `SimpleCommandMap::dispatch()` or `Command::executeRaw()`.
> Anyway, with your implementation, for me, breaking BC is needed by switching visibility. Maybe we can find a better way because I'm not a big find breaking a such large codebase.

Previously in issue #6704, `executeRaw()` my suggestion was just a simple parse function like `Command::parseQuoteAware()`. It returns args from rawArgs, then passes those args to execute and still allows for custom handling of rawArgs.

And I still don't think it's a major problem, and at its worst only causing some stupid behaviors for those practicing it.
The syntax checking should be coded to execute in most cases; you cannot stop those creepy devs anyway.

dktapps:
Well, the other option that was originally proposed is having the command expose a function that parses the args to an array. But I don't really like that approach since it still falls down to `string[]`.

Doing it with `executeRaw()` would allow implementing, for example, #6518 inside a custom command type. So it definitely has its upsides.

dktapps (commented):
Looks OK in principle, how I envisioned it. However I don't have a good answer to the concern raised by @ShockedPlot7560 around how `execute()` should be treated following this change. It probably should become protected but that will break BC, so I'm not sure how it should be handled.

ShockedPlot7560:
With our internal discussions against this PR, we don't really have better design than this for the moment. As it's planned to rework the command design, maybe, it should be OK to merge as it is. But I'm not a really big fan of this solution by my concern raised in the past and not a big fan to merge as it is too.
Should this PR be merged, closed or staled awaiting a rework ?

dktapps:
Perhaps we should save it for PM6?

dktapps:
Worth noting this is a bit of a pain for stuff like FormattedCommandAlias, since we may want to be able to pass pre-parsed arguments to commands as well. Not clear how all these pieces should fit together. See also #4193 and https://github.com/pmmp/PocketMine-MP/blob/335fcc8ed1ba8ea24c49f85d43d0dc203de6cafd/src/command/FormattedCommandAlias.php#L88-L92

dktapps:
Superseded by #6837.

Since this would require BC break anyway (otherwise the whole `execute` vs `executeRaw` issue mentioned earlier would be a headache), it makes more sense to just go all-in on reworking the command system fully to maximise the gain for the BC breaks. There's a bunch of other issues that that solves too.

Thanks for making an effort to contribute, though!

PR #6718, opened 2025-05-29
Move to newer systems for movement and block break handling

dktapps:
MS have informed us that the legacy movement and block break handling options will be removed in upcoming releases.

This PR:
- provisionally uses v3 server auth movement
- supports server auth block breaking

it does not:
- support rewind
- set the ticks in various packets needed to fully support rewind (this is a hassle to support because of packet broadcasting)

No behavioural changes are expected as a result of these changes.

### DIFF
--- a/src/network/mcpe/handler/InGamePacketHandler.php
+++ b/src/network/mcpe/handler/InGamePacketHandler.php
@@ -136,6 +136,8 @@ class InGamePacketHandler extends PacketHandler{
 	protected ?float $lastPlayerAuthInputPitch = null;
 	protected ?BitSet $lastPlayerAuthInputFlags = null;
 
+	protected ?BlockPosition $lastBlockAttacked = null;
+
 	public bool $forceMoveSync = false;
 
 	protected ?string $lastRequestedFullSkinId = null;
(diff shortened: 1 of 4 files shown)


src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -681,6 +686,7 @@ private function handlePlayerActionFromData(int $action, BlockPosition $blockPos
 
 		switch($action){
 			case PlayerAction::START_BREAK:
+			case PlayerAction::CONTINUE_DESTROY_BLOCK: //destroy the next block while holding down left click
```
dries-c, line 684: CONTINUE_DESTROY_BLOCK is also called when changing to another face while breaking the same block which would result in a new SurvivalBlockBreakHandler being constructed for something that's already being broken

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -681,6 +686,7 @@ private function handlePlayerActionFromData(int $action, BlockPosition $blockPos
 
 		switch($action){
```
dries-c, line 682: As far as I researched CRACK_BREAK & ACTION_BREAK_BLOCK are no longer used and can thus be removed

dries-c (changes requested):

src/network/mcpe/handler/InGamePacketHandler.php
dktapps: Ok, good call I wasn't aware. Will verify

dktapps: It looks like it works normally, the only issue is that the face that the particles are generated from doesn't change

src/network/mcpe/handler/InGamePacketHandler.php
dktapps: My understanding is that CRACK_BLOCK is broken currently, I'd prefer to keep it in case they fix it in the future

src/network/mcpe/handler/InGamePacketHandler.php
dktapps: Verified the behaviour of this matches vanilla (bug for bug too).

dries-c (approved):

Issue #2427, opened 2018-09-04
Windows timezone detection does not account for daylight savings


dktapps:
### Issue description
When in timezone BST, the +1 offset to the hour is not accounted for, resulting in timestamps 1 hour into the past during summer time.

### Steps to reproduce the issue

1. Run a server on Windows during UK daylight savings period (25/03 - 28/10 for 2018)
2. Notice the timestamps on the log are 1 hour in the past, despite being detected from the system (which has the correct time).

### OS and versions
* PocketMine-MP: 034472bfe754b014ffc245d395f236213c7d3f2d
* PHP: 7.2.9
* Server OS: Windows 10

dktapps:
This is further complicated by the problem of what to do when the server is running during a daylight savings shift, since the change won't be detected at runtime.

PR #6721, opened 2025-06-02
Fixed timezone error caused by WMIC deprecation in new Windows

Sunch233:
### Related issues & PRs
Due to [Microsoft has decided to deprecate WMIC in new versions of Windows](https://techcommunity.microsoft.com/blog/windows-itpro-blog/wmi-command-line-wmic-utility-deprecation-next-steps/4039242) ,  PocketMine running in Windows 11 and Windows Server 2025 will have an error when getting the time zone

image

This PR avoids this issue by getting the time zone offset with the registry

### Behavioural changes
Windows 11 and Windows Server 2025 will not report an error at startup

## Backwards compatibility
Yes


src/utils/Timezone.php
```diff
@@ -105,40 +108,40 @@ public static function init() : void{
...
-				}
-
-				$offset = $matches[2];
-
-				if($offset === ""){
-					return "UTC";
+				exec("reg query " . escapeshellarg($keyPath), $output);
+
+				foreach($output as $line){
+					if(preg_match('/ActiveTimeBias\s+REG_DWORD\s+0x+([0-9a-fA-F]+)/', $line, $matches) > 0){
```
dktapps:
```suggestion
					if(preg_match('/ActiveTimeBias\s+REG_DWORD\s+0x([0-9a-fA-F]+)/', $line, $matches) > 0){
```

Just by eyeballing it, the extra plus doesn't seem right to me. Though it depends on the output. A sample for future readers would be nice.

src/utils/Timezone.php
```diff
@@ -105,40 +108,40 @@ public static function init() : void{
...
+						$offsetMinutes = hexdec(trim($matches[1]));
+						if ($offsetMinutes > 2147483647) {
+							$offsetMinutes -= 4294967296; //signed int
+						}
```
dktapps: `Binary::signInt()` can be used instead of this check

dktapps:
Is there a reason to favour registry queries instead of other ways MS has recommended in that thread? I don't have a problem with doing it this way, but I don't want it to blow up in my face if they decide to change this too.

Sunch233:
> Is there a reason to favour registry queries instead of other ways MS has recommended in that thread? I don't have a problem with doing it this way, but I don't want it to blow up in my face if they decide to change this too.

It is also possible to use the tzutil command, but this command does not support Windows 7

dktapps:
I'm not sure if PM will actually run on W7 anyway tbh. I don't think our binaries support it.

Sunch233:
> I'm not sure if PM will actually run on W7 anyway tbh. I don't think our binaries support it.

Windows server 2008 and 2012 also do not support this command, I used to run pmmp5 on these systems

dktapps (approved):

dktapps:
It looks like this includes daylight savings time, which differs from old `WMIC` behaviour where it would only include the base time. ~~I'm not sure if we want this or not. There is an open issue about it (see #2427), but it would probably be confusing for `server.log` between restarts because it doesn't include any timezone.~~ Edit: Never mind, it appears we were already including DST by default on other platforms anyway, so this just brings Windows in line with the rest...

Perhaps we should just record server.log in UTC time and keep the timezone-adjusted time for the console only ...?

Issue #4593, opened 2021-11-24
PluginDescription is still loaded from disk twice


dktapps:
### Issue description
When loading a plugin, the PluginDescription is still loaded twice, even after recent changes:

1. when initial triage and dependency scanning happens
2. when `PluginLoader->loadPlugin()` is called, `getPluginDescription()` is called a second time because `loadPlugin()` doesn't have any other way to get the description.

This has been a problem since the recent introduction of `src-namespace-prefix`.

Fixing this problem would require one of:

a) PluginLoader to cache previously loaded descriptions (not ideal)
b) loadPlugin() to accept an additional `PluginDescription` parameter (problematic due to BC break).

### OS and versions

* PocketMine-MP: 4.0.0-BETA13+dev (git hash: 5c7125f190ad00c3eddbb65b3bc4dd29d6d56ab2-dirty)

PR #6723, opened 2025-06-02
Allow optional preloaded PluginDescription in loadPlugin method

AzaleeX:
The issue reported in the `PluginDescription` is that it gets loaded twice, which causes a problem. **Dylan** suspects that this is due to the `src-namespace-prefix parameter`. **The solution Dylan suggested** — which I also believe is the best is **method B** 'b) loadPlugin() to accept an additional PluginDescription parameter (problematic due to BC break).' . Although it introduces a **`BC break`**, it is cleaner and more efficient than the first solution, which involves **loading it into a cache**.

### Related issues & PRs
https://github.com/pmmp/PocketMine-MP/issues/4593

### API changes
loadPlugin() Now requests the `PluginDescription` parameter

Log File Temp
```txt
2025-06-02T17:05:39+00:00 : getPluginDescription called for D:\Projet\PocketMine-MP\plugins\EasyEdit.phar
```

### DIFF
--- a/src/plugin/PharPluginLoader.php
+++ b/src/plugin/PharPluginLoader.php
@@ -24,8 +24,12 @@ declare(strict_types=1);
 namespace pocketmine\plugin;
 
 use pocketmine\thread\ThreadSafeClassLoader;
+use function date;
+use function file_put_contents;
+use function getcwd;
 use function is_file;
 use function str_ends_with;
+use const FILE_APPEND;
 
 /**
  * Handles different types of plugins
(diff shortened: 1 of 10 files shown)


AzaleeX:
The run cannot be correct, because DevTools would also have to be modified so that FolderPluginLoader would change the loadPlugin function by adding PluginDescription to the parameters.

src/plugin/PharPluginLoader.php
```diff
@@ -53,6 +59,8 @@ public function loadPlugin(string $file) : void{
 	 * Gets the PluginDescription from the file
 	 */
 	public function getPluginDescription(string $file) : ?PluginDescription{
+		// TEMPORARY LOG TO TRACK HOW MANY TIMES getPluginDescription IS CALLED
+		file_put_contents(getcwd() . '/plugin_description_log.txt', date('c') . " : getPluginDescription called for $file\n", FILE_APPEND);
```
dktapps: Remove this please

dktapps (changes requested):
This is still BC breaking. Implementors of PharPluginLoader will receive deprecation warnings due to the extra parameter.

It doesn't make sense to try and pretend this is backwards compatible. These APIs are cold anyway and there are already API changes around this stuff in PM6 anyway.

PR #6732, opened 2025-06-13
fix: prevent item name color leak in /give output

### LINKED ISSUE #5342: Colours from item custom names leak into /give output
### Issue description

Since the formatting codes from `Item->getName()` are not contained when embedding them into the translations, this happens:

image

In this image, only the `Legendary` part should be grey. The rest of the message should be white.

This is likely a problem in other areas too, and it's rather inconvenient to fix in versions prior to the introduction of `TextFormat::addBase()`, so this probably won't be fixed until after 4.11.

1. /give yourself an item whose custom name contains formatting codes
2. Notice that the remainder of the audit message after the item name is the same format and colour as the item name itself

### OS and versions

* PocketMine-MP: 4.9.1 / 5.0.0-ALPHA4

ItzxDwi:
### Related issues & PRs
* Fixes #5342 

### Behavioural changes
Item name’s formatting is now isolated, preventing it from affecting the rest of the /give output

## Backwards compatibility
None

### DIFF
--- a/src/command/defaults/GiveCommand.php
+++ b/src/command/defaults/GiveCommand.php
@@ -101,7 +101,7 @@ class GiveCommand extends VanillaCommand{
 		$player->getInventory()->addItem($item);
 
 		Command::broadcastCommandMessage($sender, KnownTranslationFactory::commands_give_success(
-			$item->getName() . " (" . $args[1] . ")",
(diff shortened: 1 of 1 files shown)


dktapps:
This won't work correctly for operator audit broadcasts

PR #6740, opened 2025-06-22
feat: implements a way to enable vibrant visuals

### LINKED ISSUE #6739: API to control Vibrant Visuals
### Problem description

Currently, within the `ResourcePacksPacketHandler` code, the `ResourcePacksInfoPacket` is sent with `forceDisableVibrantVisuals` set to true.

https://github.com/pmmp/PocketMine-MP/blob/9c71f4fc1ce6ed1beeff9f55b8b209fa361983cb/src/network/mcpe/handler/ResourcePacksPacketHandler.php#L124

This hardcoded value means that we have no direct way to enable or disable the Vibrant Visuals feature for players.

### Proposed solution

We propose adding a mechanism for plugin developers to modify the value of `forceDisableVibrantVisuals` before the `ResourcePacksInfoPacket` is sent to the client. Here are a few potential approaches:

- Introduce a new method within `PlayerResourcePackOfferEvent`, `setForceDisableVibrantVisuals(bool $forceDisable)` or `setVibrantVisualsEnabled(bool $enabled)`, then pass it to the `ResourcePacksPacketHandler`.
- Make `forceDisableVibrantVisuals` public in [`ResourcePacksInfoPacket`](https://github.com/pmmp/BedrockProtocol/blob/6f60c9075dc07f9a1a5716d48d2ece4108c0e28a/src/ResourcePacksInfoPacket.php#L32)

Currently, the only known workaround involves using the `DataPacketSendEvent` in conjunction with Reflection or a closure hack to modify the `forceDisableVibrantVisuals` value within the packet. Plugin developers can't rely on this method as it's fragile and has no official backing.

ItzxDwi:
### Related issues & PRs
* Resolves #6739 

### API changes
Added:
```PlayerResourcePackOfferEvent->setForceDisableVibrantVisuals(bool $forceDisable) :  void{}```
```PlayerResourcePackOfferEvent->forceDisableVibrantVisuals() : bool{}```
```ResourcePacksPacketHandler__construct()``` now accepts new parameter ```$forceDisableVibrantVisuals```

### Behavioural changes
You can now enable vibrant visuals 🎉

## Backwards compatibility
None

## Tests
```php
		$this->getServer()->getPluginManager()->registerEvent(PlayerResourcePackOfferEvent::class, function(PlayerResourcePackOfferEvent $event): void{
                       $event->setForceDisableVibrantVisuals(false);
	}
}

### DIFF
--- a/src/event/player/PlayerResourcePackOfferEvent.php
+++ b/src/event/player/PlayerResourcePackOfferEvent.php
@@ -46,7 +46,8 @@ class PlayerResourcePackOfferEvent extends Event{
 		private readonly PlayerInfo $playerInfo,
 		private array $resourcePacks,
 		private array $encryptionKeys,
-		private bool $mustAccept
+		private bool $mustAccept,
+		private bool $forceDisableVibrantVisuals = true
 	){}
 
 	public function getPlayerInfo() : PlayerInfo{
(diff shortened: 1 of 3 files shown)


dktapps:
I don't like this much. The flag will probably be gone soon when they changed how it works again, and then we end up with dead APIs.

ItzxDwi:
> I don't like this much. The flag will probably be gone soon when they changed how it works again, and then we end up with dead APIs.

I think we should add setter in the ```ResourcePacksInfoPacket```

PR #6742, opened 2025-06-23
Fixed held block placement after respawn anchor explosion

ipad54:
Vanilla doesn't place the player held block when exploding respawn anchor.

## Tests
https://github.com/user-attachments/assets/fa8562e7-b404-44b7-969c-efb1e1dd1a0c


ItzxDwi:
Source?

ipad54:
> Source?

It is not document on the wiki, but this behavior can be seen in MCJE decompiled code or manually tested in Bedrock vanilla.

dktapps (approved):

PR #6746, opened 2025-06-24
Add the slowfall effect

### LINKED ISSUE #6694: Slow falling effect dont seems to work
### Plugin information

I haven't tested without plugins

### Problem description

So im not sure if it just a my side problem, but slow falling potion when i drink it, i dont receive the effect and when i do /effect {username} slow_falling or slowfalling and such, it seems to not be found.

### Expected behaviour

The expected behaviour is that i receive the effect.

### PocketMine-MP version

5.28.0

### PHP version

8.3.4

### Server OS

win (windows)

_No response_

Superomarking:
Implements the Slow falling effect

### API changes
* Added `VanillaEffects:SLOW_FALLING()`

### Behavioural changes
When having the slowfall effect, the (living) entity's gravity will decrease making him fall slowly.

## Tests
https://github.com/user-attachments/assets/7934e87d-2f9a-4189-bdb0-14c27ed78af9

### DIFF
--- a/src/entity/Living.php
+++ b/src/entity/Living.php
@@ -33,6 +33,7 @@ use pocketmine\entity\animation\HurtAnimation;
 use pocketmine\entity\animation\RespawnAnimation;
 use pocketmine\entity\effect\EffectInstance;
 use pocketmine\entity\effect\EffectManager;
+use pocketmine\entity\effect\SlowFallEffect;
 use pocketmine\entity\effect\VanillaEffects;
 use pocketmine\event\entity\EntityDamageByChildEntityEvent;
 use pocketmine\event\entity\EntityDamageByEntityEvent;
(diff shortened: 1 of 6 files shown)


src/entity/effect/SlowFallEffect.php
```diff
@@ -0,0 +1,41 @@
...
+
+namespace pocketmine\entity\effect;
+
+use pocketmine\entity\Entity;
+use pocketmine\entity\Living;
+
+class SlowFallEffect extends Effect{
+	public function getGravity() : float{
+		return 0.01;
+	}
```
dktapps: Not sure what the point is of making this a function. It's not overridable anyway since the effect is statically referenced from VanillaEffects.

dktapps:
Try spawning a zombie at height and start the player falling at the same time as the zombie (e.g. by breaking a block underneath you both). Zombie normally has the same gravity as a player, so if this is implemented correctly, they should match. I'm not sure changing the gravity is the right way to do this.

Superomarking:
The entity has the same gravity as the player.

src/entity/Living.php
```diff
@@ -139,6 +139,21 @@ protected function getInitialDragMultiplier() : float{ return 0.02; }
...
+		return parent::getGravity();
+	}
+
+	protected function applyDragBeforeGravity() : bool
+	{
+		if($this->effectManager->has(VanillaEffects::SLOW_FALLING())){
+			return false;
+		}
+		return parent::applyDragBeforeGravity();
+	}
```
dktapps: This is probably not necessary. Only dropped items and projectiles override it, and those can't have effects anyway.

dktapps (changes requested):
Testing with zombies made it pretty clear that the gravity override does nothing, which figures because `getGravity()` is not used within PM code (it's an API for plugin use).

The correct approach would be to use the effect manager's add/remove hooks to check for slow falling, and set the `gravity` field accordingly, either to `0.1` or to `$this->getInitialGravity()` depending on the context.

dktapps:
There's a GitHub bug that causes commits to appear on your branch that are already on `minor-next`, you can switch the branch base to a different branch and back again to clear it

Superomarking:
tweaking

Superomarking:
just a small issue 

https://github.com/user-attachments/assets/8a5ecf8c-db35-4667-bf90-e2b0e8442d1c

dktapps:
Right, I think that's because the altered gravity is only supposed to apply if the entity's Y velocity is already negative. In that case it'll probably have to be implemented in `tryChangeMovement` instead so it can do custom checks.

PR #6755, opened 2025-07-14
fix register name

Clarence2810:
### **Fix: correct block/item references for compatibility**
* Renamed `stained_clay` to `stained_hardened_clay` to match ItemTypeNames::STAINED_HARDENED_CLAY and supports different types of terracotta color in BlockTypeNames.
* Replaced `clay_block` with `clay`, since clay_block and clay refer to different things. clay_block doesn't exist, and clay is the correct block ID.
* Changed `grass` to `grass_block` because minecraft:grass does not exist in vanilla minecraft, only minecraft:grass_block is valid and compatible with BlockTypeNames::GRASS_BLOCK.
* Removed `$result->register("clay", fn() => Items::CLAY());` since "clay" should refer to the block, not the item (Items::CLAY() returns clay balls).

### DIFF
--- a/src/item/StringToItemParser.php
+++ b/src/item/StringToItemParser.php
@@ -70,7 +70,7 @@ final class StringToItemParser extends StringToTParser{
 			$register("concrete", fn() => Blocks::CONCRETE()->setColor($color));
 			$register("concrete_powder", fn() => Blocks::CONCRETE_POWDER()->setColor($color));
 			$register("glazed_terracotta", fn() => Blocks::GLAZED_TERRACOTTA()->setColor($color));
-			$register("stained_clay", fn() => Blocks::STAINED_CLAY()->setColor($color));
+			$register("stained_hardened_clay", fn() => Blocks::STAINED_CLAY()->setColor($color));
 			$register("stained_glass", fn() => Blocks::STAINED_GLASS()->setColor($color));
 			$register("stained_glass_pane", fn() => Blocks::STAINED_GLASS_PANE()->setColor($color));
 			$register("stained_hardened_glass", fn() => Blocks::STAINED_HARDENED_GLASS()->setColor($color));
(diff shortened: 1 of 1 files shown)


dktapps (changes requested):
Existing aliases must not be changed. It's fine to add new ones, but existing ones may be used by user data and config files, so they can't be removed.

This should also target `minor-next`.

Clarence2810:
But isn't that confusing? Because when you try to get the string name identifier from `BlockTypeNames` or `ItemTypeNames` (which are generated automatically), and then use `StringToItemParser`, it won’t return what you expect.

alvin0319:
Changing the existing name will cause BC break which will eventually break plugins using old name.

dktapps:
> But isn't that confusing? Because when you try to get the string name identifier from `BlockTypeNames` or `ItemTypeNames` (which are generated automatically), and then use `StringToItemParser`, it won’t return what you expect.

You're not supposed to use those in StringToItemParser anyway.

ryuk170:
Que hagan pm6 ala verga

dktapps:
This is one thing we can't break because user data depends on it. It's not like plugins where API changes can be adapted to.

ShockedPlot7560 (changes requested):
You should keep the existing alias as @dktapps said before. Your changes can be related to STEAK where it still exist : `steak` and `cooked_beef`.

If your ambition is to follow `ItemTypeNames`, there's more than your articles to do, that's only part of the problem.

If this PR only adds aliases, it may have a chance, otherwise it's not possible to merge the delete of aliases.

Clarence2810:
So, there's no possible way for `ItemTypeNames` and `BlockTypeNames` to be connected with `StringToItemParser`? It's confusing for other developers to know what to expect when using `TypeNames` as parameters with `StringToItemParser`. I know @dktapps mentioned that these aren't supposed to be used directly, but let's say other developers aren’t aware of that — they might expect it to work the way it used to, especially since StringToItemParser is documented as: **"Handles parsing items from strings. This is used to interpret names from the /give command (and others)"**. TypeNames also work with the /give command in vanilla, which adds to the confusion.

dktapps:
> So, there's no possible way for ItemTypeNames and BlockTypeNames to be connected with StringToItemParser?

No, these were never supposed to be connected.
The purpose of `StringToItemParser` is to support mapping IDs similar to how it used to be done in PM3 with `constant("ItemIds::$WHATEVER")`. There's no direct relationship between vanilla's string IDs (which are an inconsistent mess anyway).

I understand your concerns about deviating from the vanilla `/give` command, however I'm not really clear what can be done to fix this. As I said, this system also supports processing items stored in databases and config files, so we can't change existing aliases without breaking such data.

dktapps:
I'd be fine with adding the aliases for grass_block and stained_clay alongside the existing aliases, but I think the clay alias is best left alone for now until we can figure out a solution for this.

Clarence2810:
How should we handle `hard_{color}_stained_glass`, `hard_{color}_stained_glass_pane`, since registerDynamicBlocks puts the color in the prefix?

dktapps:
Why do you want to mess with those?

dktapps:
On a side note (not related to the changes, but the general issue of colliding aliases), there is a new `bush` block in 1.21.70, but we already have a mapping for `bush => Blocks::DEAD_BUSH()`. This is also a bit of a problem that I'm not sure how to deal with.
Ideally we'd do what the user expects for commands, but loading legacy data should be able to retrieve the original item.

PR #6759, opened 2025-07-30
BlockObjectToStateSerializer: Avoid unnecessary Writer and Closure

### LINKED ISSUE #6675: BlockObjectToStateSerializer: Allow serializers to return an ID string directly instead of a BlockStateWriter instance
### Link to crashdump

https://crash.pmmp.io/view/12462321

Most common blocks don't need to store any properties. In these cases we can avoid redundant object allocations by allowing the serializer handlers to just return IDs directly.

This crashdump was likely caused by a mistake, but the idea is a good one.

ShockedPlot7560:
In the case of a simple mapping for the block serializer, we don't even need to go through `BlockStateWriter` to get the `BlockStateData`.

This PR, as the issue suggests, allows the serializer to return a string directly instead of a Writer in order to limit the allocation of unnecessary objects.

### Related issues & PRs
* fixes #6675 

### API changes
- Serializers in `BlockObjectToStateSerializer` can now return a string safely.

### DIFF
--- a/src/data/bedrock/block/convert/BlockObjectToStateSerializer.php
+++ b/src/data/bedrock/block/convert/BlockObjectToStateSerializer.php
@@ -189,6 +189,7 @@ use pocketmine\data\bedrock\block\convert\BlockStateWriter as Writer;
 use pocketmine\math\Axis;
 use pocketmine\math\Facing;
 use function get_class;
+use function is_string;
 
 final class BlockObjectToStateSerializer implements BlockStateSerializer{
 	/**
@@ -196,7 +197,7 @@ final class BlockObjectToStateSerializer implements BlockStateSerializer{
 	 * describe the bottom type of a type hierarchy only containing Block.
 	 *
 	 * @var (\Closure|BlockStateData)[]
-	 * @phpstan-var array<int, \Closure(never) : (Writer|BlockStateData)|BlockStateData>
+	 * @phpstan-var array<int, \Closure(never) : (Writer|BlockStateData|string)|BlockStateData>
 	 */
 	private array $serializers = [];
 
@@ -233,12 +234,13 @@ final class BlockObjectToStateSerializer implements BlockStateSerializer{
 	/**
 	 * @phpstan-template TBlockType of Block
 	 * @phpstan-param TBlockType $block
-	 * @phpstan-param \Closure(TBlockType) : (Writer|BlockStateData)|Writer|BlockStateData $serializer
+	 * @phpstan-param \Closure(TBlockType) : (Writer|BlockStateData|string)|Writer|BlockStateData $serializer
 	 */
 	public function map(Block $block, \Closure|Writer|BlockStateData $serializer) : void{
 		if(isset($this->serializers[$block->getTypeId()])){
 			throw new \InvalidArgumentException("Block type ID " . $block->getTypeId() . " already has a serializer registered");
 		}
+		//writer accepted for convenience only
 		$this->serializers[$block->getTypeId()] = $serializer instanceof Writer ? $serializer->getBlockStateData() : $serializer;
 	}
 
@@ -272,7 +274,7 @@ final class BlockObjectToStateSerializer implements BlockStateSerializer{
 			throw new BlockStateSerializeException("No serializer registered for " . get_class($blockState) . " with type ID $typeId");
 		}
 
-		if($locatedSerializer instanceof BlockStateData){
+		if($locatedSerializer instanceof BlockStateData){ //static data, not dependent on state
 			return $locatedSerializer;
 		}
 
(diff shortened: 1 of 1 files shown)


dktapps:
I did look at doing this myself, but since we already cache state IDs in this case anyway, the benefit is fairly minimal. It does allow us to have less useless closures and such, but the performance benefit is close to 0.

Edit: I see this goes further than I originally envisioned. I hadn't thought about getting rid of `Writer` for switched IDs with no states.

src/data/bedrock/block/convert/BlockObjectToStateSerializer.php
```diff
@@ -243,7 +243,7 @@ public function map(Block $block, \Closure $serializer) : void{
 	}
 
 	public function mapSimple(Block $block, string $id) : void{
-		$this->map($block, fn() => Writer::create($id));
+		$this->map($block, fn() => $id);
```
dktapps: This doesn't make much sense, it would be better if we could ditch the closure entirely

ShockedPlot7560: Yeah, that's was my second intention, it will be done in this PR instead of a follow up

src/data/bedrock/block/convert/BlockObjectToStateSerializer.php
```diff
@@ -277,14 +278,20 @@ public function serializeBlock(Block $blockState) : BlockStateData{
 		 * the type ID of the block (which never makes sense, even in a world where overriding block types is a thing).
 		 * In the future we'll need some way to guarantee that type IDs are never reused (perhaps spl_object_id()?)
 		 *
-		 * @var \Closure $serializer
-		 * @phpstan-var \Closure(TBlockType) : Writer $serializer
+		 * @var \Closure|string $serializer
+		 * @phpstan-var \Closure(TBlockType) : (Writer|string)|string $serializer
 		 */
 		$serializer = $locatedSerializer;
+		if(is_callable($serializer)){
```
dktapps: `instanceof \Closure` would be better here, a string ID might be incorrectly considered callable

dktapps: This can also be reduced to a ternary

dktapps: Actually this whole thing can be simplified

dktapps:
Are there any situations where we return a `Writer` without taking any info from the `Block`?

ShockedPlot7560:
> Are there any situations where we return a `Writer` without taking any info from the `Block`?

I've gone through everything and normally there shouldn't be any more, subject to error but low risk.

dktapps:
The code you've written would allow for `Writer|string` or `Closure(Block) : (Writer|string)` if we wanted it to. So I'm just wondering if there's any cases where we write properties that aren't derived from the block (e.g. to polyfill and make Minecraft happy).

dktapps:
https://github.com/pmmp/PocketMine-MP/blob/stable/src/data/bedrock/block/convert/BlockObjectToStateSerializer.php#L594
https://github.com/pmmp/PocketMine-MP/blob/a83c62a4a2e5eea17fdfc3eb2bfd55f31534a9e3/src/data/bedrock/block/convert/BlockObjectToStateSerializer.php#L1282-L1283
https://github.com/pmmp/PocketMine-MP/blob/a83c62a4a2e5eea17fdfc3eb2bfd55f31534a9e3/src/data/bedrock/block/convert/BlockObjectToStateSerializer.php#L1629-L1630
https://github.com/pmmp/PocketMine-MP/blob/a83c62a4a2e5eea17fdfc3eb2bfd55f31534a9e3/src/data/bedrock/block/convert/BlockObjectToStateSerializer.php#L1706
https://github.com/pmmp/PocketMine-MP/blob/a83c62a4a2e5eea17fdfc3eb2bfd55f31534a9e3/src/data/bedrock/block/convert/BlockObjectToStateSerializer.php#L1714

All of these cases could be mapped to a `Writer` instance directly (or `BlockStateData`) since they don't use the block input for context. If you ctrl+f for `fn() =>` there's probably more cases.

ShockedPlot7560:
I see our point of view, it is necessary to define the serialiser as `Closure|string|Writer` to cover all cases. Good catch

dktapps:
I've been thinking about this, and I'm not sure that going to `string` actually makes sense. Since we turn it into `BlockStateData` before returning it anyway, and there are cases where we just want to return a static `BlockStateData` anyway.

Perhaps it would make more sense to just have `BlockStateData|\Closure(Block) : (Writer|BlockStateData)`.

ShockedPlot7560:
I'm not sure at the moment about using BlockStateData::current and getBlockStateData directly in this class. Looks weird for me.

I'm still dubious about the use of BlockStateData rather string in the sense that the changes there are aimed more at reducing the memory footprint than overall performance. By replacing Writer with BlockStateData, there's no real gain, so do we really need this change?

dktapps:
I've tidied this up. There are now the following cases:

- Closure(Block) -> Writer (regular case)
- Closure(Block) -> BlockStateData (block with flattened IDs, no state data)
- Writer (polyfills useless properties, not dependent on input)
- BlockStateData (ID only, not dependent on input)

Though I want to rethink the way flattened IDs are handled at some point

dktapps (approved):

PR #6769, opened 2025-08-18
Unify block serializers

dktapps:
This PR makes a start on unifying common block serializers.
This means that we'll be able to implement most blocks with a model description in a single file, instead of needing to write a separate serializer and deserializer.

This has several advantages:
- Easier to implement new blocks (one less file to modify)
- Easier to adjust serialization of existing blocks
- Guaranteed consistency between serializers and deserializers
- Exposes more metadata for programmatic analysis, instead of having everything baked inside opaque `Closure`s

Caveats:
- Flattened blocks (PM blocks `1:N` MC IDs) are a chore to deal with, especially in the case of copper and coral blocks where multiple properties are baked into the ID
- Split blocks (PM blocks `N:1` MC IDs) are also an unsolved problem
- This is probably slightly worse for performance than the old way

However, for all cases where a `1:1` association is present, this cuts the code required in half.

> [!CAUTION]
> Since this is building on top of existing systems, it's possible to migrate things piecemeal to the new system without having lots of downtime. However I'd prefer to get as much of the existing blocks migrated as possible before merging this, in case there are design flaws I need to adjust for.

### Related issues & PRs
Work towards #6311

### API changes
- Added `BlockSerializerDeserializerRegistrar`
- Various classes for dealing with different property types, e.g. `BoolProperty`, `IntProperty`, `EnumProperty` etc

### Behavioural changes
Functionally this should work the same as before.
Performance impacts have not yet been measured.

## Backwards compatibility
No BC breaks. Since this builds on top of the preexisting split APIs, there are no backwards compatibility concerns I'm currently tracking.
Some legacy methods were deprecated though.

## Follow-up
- Get rid of deprecated `BlockState(De)SerializerHelper` methods (It's possible we may be able to get rid of these helpers entirely with the new approach
- Hopefully make all serializers symmetric???

## Tests
PHPUnit symmetric serializer tests are green, so I'm happy


dktapps:
All blocks have now been unified, except for blocks which have 1 ID but have several PM block counterparts. This has typically been used in cases where the different property values caused such a significant change in the block's behaviour that it didn't make sense to consider them as the same anymore. I haven't figured out how to unify these split blocks yet, and the technical challenge this presents may not be worth solving given that only a handful of blocks need this anyway.

Flattened blocks turned out to be not that difficult to deal with. I ended up with an API that uses property descriptors much like real states, and compiles a matrix of IDs from those descriptions. This produced a dramatic reduction in the amount of code repetition needed for the likes of copper, coral and various coloured blocks, while functionally requiring no additional features and performance being as good as a regular block's handlers.

This PR is now mostly done, although I'll probably do some cleaning up and try and deal with some of the `property` namespace copy paste before merge. I'll also pull out separable changes to `ChiseledBookshelf` and other block APIs that were added.

PR #6771, opened 2025-08-22
Fix sneaking hitbox height

Dasciam:
BDS sends 1.49, so 53/64 (0.828125) gonna be good for it.  Reference: https://github.com/df-mc/dragonfly/pull/1000

### API changes
- `Living` now has  `getSneakOffset()`  to define hitbox height offset while sneaking.

## Tests
Tested

https://github.com/user-attachments/assets/6f59cdbc-2a78-407c-84f6-c79540bf13f9

### DIFF
--- a/src/entity/Living.php
+++ b/src/entity/Living.php
@@ -296,7 +296,7 @@ abstract class Living extends Entity{
 			$width = $size->getWidth();
 			$this->setSize((new EntitySizeInfo($width, $width, $width * 0.9))->scale($this->getScale()));
 		}elseif($this->isSneaking()){
-			$this->setSize((new EntitySizeInfo($size->getHeight() - $this->getSneakOffset(), $size->getWidth(), 3 / 4 * $size->getEyeHeight()))->scale($this->getScale()));
+			$this->setSize((new EntitySizeInfo($size->getHeight() - $this->getSneakOffset(), $size->getWidth(), $size->getEyeHeight() - $this->getSneakOffset()))->scale($this->getScale()));
 		}else{
 			$this->setSize($size->scale($this->getScale()));
 		}
--- a/src/entity/Living.php
+++ b/src/entity/Living.php
@@ -243,7 +243,7 @@ abstract class Living extends Entity{
 	}
 
 	public function getSneakOffset() : float{
-		return 0.0
+		return 0.0;
 	}
 
 	public function isSneaking() : bool{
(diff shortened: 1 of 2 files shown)


dktapps:
This logic seems sketchy to me, not sure who implemented it like that. In particular the eye height being scaled doesn't seem right since it'll mean the eyes are closer to the top when sneaking than when standing.

dktapps:
Also if, say, an enderman were to sneak, it'd probably change the height by a different amount than a regular biped, since an enderman is more leggy than other mobs.

Perhaps a better way to deal with this would be to have a sneak height offset (defaulting to 1.8 - 1.49) and then allow entities to override it as needed.

Dasciam:
I think the Human is one and only entity that can sneak.

dktapps:
Custom entities could also sneak, so we don't want to be overly restrictive.

Dasciam:
> Custom entities could also sneak, so we don't want to be overly restrictive.

Anyway  I made so that every Living can have its own sneak offset so I think it can be merged

It is just defaulted to zero

src/entity/Living.php
```diff
@@ -292,7 +296,7 @@ private function recalculateSize() : void{
 			$width = $size->getWidth();
 			$this->setSize((new EntitySizeInfo($width, $width, $width * 0.9))->scale($this->getScale()));
 		}elseif($this->isSneaking()){
-			$this->setSize((new EntitySizeInfo(3 / 4 * $size->getHeight(), $size->getWidth(), 3 / 4 * $size->getEyeHeight()))->scale($this->getScale()));
+			$this->setSize((new EntitySizeInfo($size->getHeight() - $this->getSneakOffset(), $size->getWidth(), 3 / 4 * $size->getEyeHeight()))->scale($this->getScale()));
```
kostamax27: you messed something up

```
1.8 - 1.49 = 0.31
```

Dasciam: I realised it before you reviewed it but anyway thanks

dktapps: Sneak offset should also apply to eye height

dktapps (changes requested):

src/entity/Living.php
Dasciam: alright

dktapps (dismissed):

dktapps (approved):

ShockedPlot7560 (approved):

PR #6773, opened 2025-08-22
refactor: deprecate PMMP Config and migrate to Stellaris framework

CelestifyX:
- Refactor PMMP Config: deprecated the old Config class and migrated functionality to the Stellaris framework.

## Changes
- `Config` class is now marked as `@deprecated`.
- New configuration handling should use `Stellaris\ConfigManager`.

## Tests
image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,7 +50,7 @@
       "pocketmine/snooze": "^0.5.0",
       "ramsey/uuid": "~4.9.0",
       "symfony/filesystem": "~6.4.0",
-      "celestifyx/stellaris": "1.0.3"
+      "celestifyx/stellaris": "1.0.4"
    },
    "require-dev": {
       "phpstan/phpstan": "2.1.17",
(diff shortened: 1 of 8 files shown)


dktapps:
Uhh, this is awkward... #3766

What exactly do we gain from this change?

CelestifyX:
More extensions for every taste ("ini", "json", "php", "toml", "xml", "yaml", "properties", "dotenv", "serialized", "hjson", "json5", "txt"), a full set of methods for working with configs

dktapps:
Anyone who wants those formats is rolling their own stuff anyway. And as I said, we intend to get rid of Config. It just gets in the way of using stuff like validators or object class model mapping.

Rolling a library that is used by no one else, to supersede a feature we intend to get rid of anyway, into a large project with thousands of users with no prior discussion is not the play. Sorry, but I just don't think this change makes sense.

PR #6774, opened 2025-08-25
Support editing the back side of signs

dktapps:
Working, but could be 10% cleaner if I was able to make BC breaks.

Preexisting `text` properties of sign-related objects refers to the front side; new `backText` properties have been added where appropriate.

## API changes
- `SignChangeEvent` now fires when the back side of a sign is edited (check `isFrontSide()`)
- `BaseSign->updateText()` now accepts an optional `bool $frontSide = true` parameter
- `Player->openSignEditor()` now accepts an optional `bool $frontSide = true` parameter
- Added `BaseSign->getBackText()` and `BaseSign->setBackText()`
- `BaseSign` inheritors should implement `getFacingDegrees()` and may need to implement `getHitboxCenter()` for sign editor to open the correct side when right-clicking on the block

## Issues
- Duplication of code for `getFacingDegrees()` in different types of sign classes
- `getFacingDegrees()` should be abstract, but limited by requirement for backwards compatibility
- Various usages of read/write text now have to do `$frontSide ? $text : $backText` or similar, which is prone to bugs

### DIFF
--- a/src/block/BaseSign.php
+++ b/src/block/BaseSign.php
@@ -135,11 +135,11 @@ abstract class BaseSign extends Transparent implements WoodMaterial{
 		}
 	}
 
-	private function doSignChange(SignText $newText, Player $player, Item $item, bool $frontSide) : bool{
-		$ev = new SignChangeEvent($this, $player, $newText, $frontSide);
+	private function doSignChange(SignText $newText, Player $player, Item $item, bool $frontFace) : bool{
+		$ev = new SignChangeEvent($this, $player, $newText, $frontFace);
 		$ev->call();
 		if(!$ev->isCancelled()){
-			$frontSide ? $this->setText($ev->getNewText()) : $this->setBackText($ev->getNewText());
+			$this->setFaceText($frontFace, $ev->getNewText());
 			$this->position->getWorld()->setBlock($this->position, $this);
 			$item->pop();
 			return true;
@@ -148,9 +148,9 @@ abstract class BaseSign extends Transparent implements WoodMaterial{
 		return false;
 	}
 
-	private function changeSignGlowingState(bool $glowing, Player $player, Item $item, bool $frontSide) : bool{
-		$text = $frontSide ? $this->text : $this->backText;
-		if($text->isGlowing() !== $glowing && $this->doSignChange(new SignText($text->getLines(), $text->getBaseColor(), $glowing), $player, $item, $frontSide)){
+	private function changeSignGlowingState(bool $glowing, Player $player, Item $item, bool $frontFace) : bool{
+		$text = $this->getFaceText($frontFace);
+		if($text->isGlowing() !== $glowing && $this->doSignChange(new SignText($text->getLines(), $text->getBaseColor(), $glowing), $player, $item, $frontFace)){
 			$this->position->getWorld()->addSound($this->position, new InkSacUseSound());
 			return true;
 		}
@@ -177,7 +177,7 @@ abstract class BaseSign extends Transparent implements WoodMaterial{
 			return true;
 		}
 
-		$clickedFront = $this->interactsFront($this->getHitboxCenter(), $player->getPosition(), $this->getFacingDegrees());
+		$frontFace = $this->interactsFront($this->getHitboxCenter(), $player->getPosition(), $this->getFacingDegrees());
 
 		$dyeColor = $item instanceof Dye ? $item->getColor() : match($item->getTypeId()){
 			ItemTypeIds::BONE_MEAL => DyeColor::WHITE,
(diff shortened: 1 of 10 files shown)


src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -755,6 +755,43 @@ public function handlePlayerHotbar(PlayerHotbarPacket $packet) : bool{
 		return true; //this packet is useless
 	}
 
+	/**
+	 * @throws PacketHandlingException
+	 */
+	private function updateSignText(CompoundTag $nbt, string $tagName, bool $isFront, BaseSign $block, Vector3 $pos) : bool{
+		$frontTextTag = $nbt->getTag($tagName);
```
ipad54: This variable name is now a bit confusing since it’s not only the front tag.

src/block/CeilingEdgesHangingSign.php
```diff
@@ -65,4 +66,14 @@ private function canBeSupportedAt(Block $block) : bool{
 			$supportBlock->getSupportType(Facing::DOWN) === SupportType::FULL ||
 			(($supportBlock instanceof WallHangingSign || $supportBlock instanceof CeilingEdgesHangingSign) && Facing::axis($supportBlock->getFacing()) === Facing::axis($this->facing));
 	}
+
+	protected function getFacingDegrees() : float{
+		return match ($this->facing) {
```
ipad54:
```suggestion
		return match($this->facing) {
```

src/block/WallHangingSign.php
```diff
@@ -78,4 +79,14 @@ private function canBeSupportedAt(Block $block, int $face) : bool{
 			($block instanceof WallHangingSign && Facing::axis(Facing::rotateY($block->getFacing(), clockwise: true)) === Facing::axis($face)) ||
 			$block->getSupportType(Facing::opposite($face)) === SupportType::FULL;
 	}
+
+	protected function getFacingDegrees() : float{
+		return match ($this->facing) {
```
ipad54:
```suggestion
		return match($this->facing) {
```

ipad54 (changes requested):

src/network/mcpe/handler/InGamePacketHandler.php
dktapps: Yeah, it's a refactor leftover

src/block/WallHangingSign.php
dktapps: Honestly it might be time to do a formatting overhaul. I can't get any tool to format it without the spaces, so it has to be manually formatted each time I use a `match`. I'm losing the patience for it.

src/event/block/SignChangeEvent.php
```diff
@@ -71,4 +75,8 @@ public function getNewText() : SignText{
 	public function setNewText(SignText $text) : void{
 		$this->text = $text;
 	}
+
+	public function isFrontSide() : bool{ return $this->frontSide; }
+
+	public function setFrontSide(bool $frontSide) : void{ $this->frontSide = $frontSide; }
```
dktapps: Jury's out on whether this makes sense. I'm leaning towards "no", but I'm not sure. Thoughts?

dktapps:
Also open question whether we should allow different players to edit opposite sides of the sign at the same time. Vanilla doesn't support this as far as I can tell, but it would be fairly trivial to support it in PM if we wanted to.

src/event/block/SignChangeEvent.php
ipad54: I can’t come up with any use case, but I don’t mind having this. We have an ability to change the text, then why not have an ability to change its side? Maybe someone will find it useful.

ipad54:
> Also open question whether we should allow different players to edit opposite sides of the sign at the same time. Vanilla doesn't support this as far as I can tell, but it would be fairly trivial to support it in PM if we wanted to.

I guess it could be a pretty cool behavior, but essentially we follow the vanilla with such stuff, right? Plugins can implement this in their way, especially with the new API in `SignChangeEvent`

ipad54 (dismissed):

dktapps:
It might be possible with SignChangeEvent but it'd be clunky. It'd require clearing the editor runtime ID from the sign after opening the sign UI, and also uncancelling the SignChangeEvent. Feels like it'd be easier to just have a separate editor ID for each side. (Frankly I'm not sure why we put the editor ID in the NBT anyway ...)

src/event/block/SignChangeEvent.php
dktapps: My concern is that the "old text" won't be an accurate reflection if the side is overridden.

ipad54: I see the problem. We can either add getOldSide() or remove the setter since it doesn’t have much use cases. I’m leaning towards removing it now because getOldSide() will be very confusing with isFrontSide().

ipad54:
Yeah, that makes sense. By the way, does writing editor RID into the spawn NBT do something like blocking editor window opening on client-side or it’s useless and can be removed?

dktapps:
> Yeah, that makes sense. By the way, does writing editor RID into the spawn NBT do something like blocking editor window opening on client-side or it’s useless and can be removed?

The server controls opening the UI, so I'm pretty sure it's useless. I'm not sure why I had it written in the NBT anyway tbh.

src/event/block/SignChangeEvent.php
dktapps: Agreed

dktapps:
Actually, scratch that - I'd forgotten that signs are normally restricted to editing only by the player who originally placed it anyway, so front vs back side wouldn't make any difference.

One other thing on my mind is how to deal with the new APIs. `getText()` is left to return the front text, but it's not obvious. Should we have `getFrontText()` et al? And what about `getText(bool $frontSide = true)`? I'm undecided what's the best way to approach it. I will note that there's a lot of `$front ? setText() : setBackText()` ternaries.

ipad54:
> I'd forgotten that signs are normally restricted to editing only by the player who originally placed it
>

afaik anyone can edit a sign now if it’s not being currently edited

ipad54:
About the new API. I think we should add `getFrontText()`, `setFrontText()` and deprecate `getText()`, `setText()` for now. I’m afraid that someone could override `getText()`, so changing it to `getText(bool $frontSide = true)` will cause BC Break. In PM6 I guess we can add get/set text methods with `bool $frontSide` parameters and replace ternary operators with them.

dktapps:
Really, in PM we restrict it only to the player who placed it. I presumed that's why the NBT has an entity ID in it.
(Actually now I think about it, that would make signs behave as if they were waxed after a restart, since we don't actually save that. Maybe that needs to be looked into.)

In any case it's a non-essential feature. I'm more interested in getting the API on this done right.

dktapps:
Ok I think I'm done

src/player/Player.php
```diff
@@ -2840,11 +2840,11 @@ protected function removePermanentInventories() : void{
 	 * Opens the player's sign editor GUI for the sign at the given position.
 	 * TODO: add support for editing the rear side of the sign (not currently supported due to technical limitations)
```
ipad54: This TODO can be removed

src/block/BaseSign.php
```diff
@@ -234,29 +287,38 @@ public function setEditorEntityRuntimeId(?int $editorEntityRuntimeId) : self{
...
 			return TextFormat::clean($line, false);
-		}, $text->getLines()), $this->text->getBaseColor(), $this->text->isGlowing()));
+		}, $text->getLines()), $oldText->getBaseColor(), $oldText->isGlowing()), $frontFace);
 		if($this->waxed || $this->editorEntityRuntimeId !== $author->getId()){
 			$ev->cancel();
 		}
 		$ev->call();
 		if(!$ev->isCancelled()){
-			$this->setText($ev->getNewText());
+			$this->setFaceText($frontFace, $text);
```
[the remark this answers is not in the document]
dktapps: Good bot

ipad54 (dismissed):

PR #6775, opened 2025-08-25
Fixed hanging signs placement criteria

ipad54:
#### Wall:
 - Wall hanging signs require full support type (they can't be placed on leaves, etc).
 - Wall hanging signs can provide support to each other, but only if have matching axis.
 - Wall hanging signs can be still placed when clicked axis is Y. In that case they try to find support based on player's horizontal facing.
 
 #### Ceiling:
  - Sneaking will always result in placing center variant.
  - Edge variant can be placed on blocks with full support or on a wall/edge hanging sign with the same axis.
  - Center variant can be placed on blocks with center support or on a wall/edge/center hanging signs with different facings.

### API changes
Added `?Player` parameter to `getPlacementBlock()`. Not a BC Break since it's not on a stable yet.

## Follow-up
There is still one bug remains. Vanilla doesn't open the sign editor menu when placing hanging sign on another hanging sign. Not sure how to fix that without huge code duplication.

## Tests
https://youtu.be/njGMGLCZhSI

### DIFF
--- a/src/block/CeilingCenterHangingSign.php
+++ b/src/block/CeilingCenterHangingSign.php
@@ -25,6 +25,7 @@ namespace pocketmine\block;
 
 use pocketmine\block\utils\SignLikeRotation;
 use pocketmine\block\utils\SignLikeRotationTrait;
+use pocketmine\block\utils\StaticSupportTrait;
 use pocketmine\item\Item;
 use pocketmine\math\Facing;
 use pocketmine\math\Vector3;
@@ -33,6 +34,7 @@ use pocketmine\world\BlockTransaction;
 
 final class CeilingCenterHangingSign extends BaseSign implements SignLikeRotation{
 	use SignLikeRotationTrait;
+	use StaticSupportTrait;
 
 	protected function getSupportingFace() : int{
 		return Facing::UP;
@@ -44,19 +46,18 @@ final class CeilingCenterHangingSign extends BaseSign implements SignLikeRotatio
 			return false;
 		}
 
-		$supportBlock = $blockReplace->getSide(Facing::UP);
-		if(
-			!$supportBlock->getSupportType(Facing::UP)->hasCenterSupport() &&
-			!$supportBlock instanceof WallHangingSign &&
-			!$supportBlock instanceof CeilingEdgesHangingSign &&
-			!$supportBlock instanceof CeilingCenterHangingSign
-		){
-			return false;
-		}
-
 		if($player !== null){
 			$this->rotation = self::getRotationFromYaw($player->getLocation()->getYaw());
 		}
 		return parent::place($tx, $item, $blockReplace, $blockClicked, $face, $clickVector, $player);
+	}
+
+	private function canBeSupportedAt(Block $block) : bool{
+		$supportBlock = $block->getSide(Facing::UP);
+		return
+			$supportBlock->getSupportType(Facing::DOWN)->hasCenterSupport() ||
+			$supportBlock instanceof WallHangingSign ||
+			$supportBlock instanceof CeilingEdgesHangingSign ||
+			$supportBlock instanceof CeilingCenterHangingSign;
 	}
 }
--- a/src/block/CeilingCenterHangingSign.php
+++ b/src/block/CeilingCenterHangingSign.php
@@ -44,6 +44,15 @@ final class CeilingCenterHangingSign extends BaseSign implements SignLikeRotatio
 			return false;
 		}
 
+		$supportBlock = $blockReplace->getSide(Facing::UP);
+		if(
+			!$supportBlock->getSupportType(Facing::UP)->hasCenterSupport() &&
+			!$supportBlock instanceof CeilingEdgesHangingSign &&
+			!$supportBlock instanceof CeilingCenterHangingSign
+		){
+			return false;
+		}
+
 		if($player !== null){
 			$this->rotation = self::getRotationFromYaw($player->getLocation()->getYaw());
 		}
--- a/src/block/CeilingCenterHangingSign.php
+++ b/src/block/CeilingCenterHangingSign.php
@@ -56,8 +56,6 @@ final class CeilingCenterHangingSign extends BaseSign implements SignLikeRotatio
 		$supportBlock = $block->getSide(Facing::UP);
 		return
 			$supportBlock->getSupportType(Facing::DOWN)->hasCenterSupport() ||
-			$supportBlock instanceof WallHangingSign ||
-			$supportBlock instanceof CeilingEdgesHangingSign ||
-			$supportBlock instanceof CeilingCenterHangingSign;
+			$supportBlock->hasTypeTag(BlockTypeTags::HANGING_SIGN);
 	}
 }
--- a/src/block/CeilingCenterHangingSign.php
+++ b/src/block/CeilingCenterHangingSign.php
@@ -47,6 +47,7 @@ final class CeilingCenterHangingSign extends BaseSign implements SignLikeRotatio
 		$supportBlock = $blockReplace->getSide(Facing::UP);
 		if(
 			!$supportBlock->getSupportType(Facing::UP)->hasCenterSupport() &&
+			!$supportBlock instanceof WallHangingSign &&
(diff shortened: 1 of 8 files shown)


src/block/CeilingCenterHangingSign.php
```diff
@@ -44,6 +44,16 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
 			return false;
 		}
 
+		$supportBlock = $blockReplace->getSide(Facing::UP);
+		if(
+			!$supportBlock->getSupportType(Facing::UP)->hasCenterSupport() &&
```
dktapps: This doesn't look right. It should be Facing::DOWN

src/block/CeilingEdgesHangingSign.php
```diff
@@ -46,6 +47,15 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
 			$this->facing = Facing::opposite($player->getHorizontalFacing());
 		}
 
+		$supportBlock = $blockReplace->getSide(Facing::UP);
+		if(
+			$supportBlock->getSupportType(Facing::UP) !== SupportType::FULL &&
```
dktapps: Same again here

src/block/WallHangingSign.php
```diff
@@ -50,16 +51,38 @@ protected function recalculateCollisionBoxes() : array{
...
-		$this->facing = Facing::rotateY($face, clockwise: true);
+		foreach([
+			$attachFace,
+			Facing::opposite($attachFace)
+		] as $side){
+			if($this->canBeSupportedAt($blockReplace->getSide($side), $side)){
+				$direction = $side;
+				break;
+			}
+		}
```
dktapps: I feel like a simple if with 2 canBeSupportedAt calls would be less code...

dktapps: This code is now really confusing. If the player clicked a horizontal face, this extra logic shouldn't apply

src/item/HangingSign.php
```diff
@@ -40,13 +43,21 @@ public function __construct(
...
 		//we don't verify valid placement conditions here, only decide which block to return
-		$result = $face === Facing::DOWN ?
-			$blockReplace->getSide(Facing::UP)->getSupportType(Facing::DOWN) === SupportType::CENTER ?
+		if($face === Facing::DOWN){
+			if($player !== null && $player->isSneaking()){
+				return clone $this->centerPointCeilingVariant;
+			}
+
+			$support = $blockReplace->getSide(Facing::UP);
+			$result = (($support instanceof CeilingEdgesHangingSign || $support instanceof WallHangingSign) && ($player === null || Facing::axis($player->getHorizontalFacing()) !== Facing::axis($support->getFacing()))) ||
```
dktapps: I don't think this condition is quite right

In this code, center sign is placed if:

- Support is a ceiling edges sign or wall hanging sign, and
- Player is facing perpendicular to the sign

Or, support type is CENTER

However, a ceiling center sign can also be hung from another ceiling center sign, which this code doesn't account for (it'll select a ceiling edges sign).

The code is quite complex and hard to reason about, it could use some explanatory comments

src/block/CeilingEdgesHangingSign.php
```diff
@@ -46,6 +47,15 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
...
+			(!$supportBlock instanceof WallHangingSign || Facing::axis($supportBlock->getFacing()) !== Facing::axis($this->facing)) &&
+			!$supportBlock instanceof CeilingEdgesHangingSign
+		){
+			return false;
+		}
```
dktapps, line 51: This should probably be pulled out into proper support logic so the sign self-destructs when these conditions aren't met

dktapps (changes requested):

dktapps:
> There is still one bug remains. Vanilla doesn't open the sign editor menu when placing hanging sign on another hanging sign. Not sure how to fix that without huge code duplication.

It does in my tests. Not even sneaking disables it.

ipad54:
> > There is still one bug remains. Vanilla doesn't open the sign editor menu when placing hanging sign on another hanging sign. Not sure how to fix that without huge code duplication.
> 
> It does in my tests. Not even sneaking disables it.

It lets you place a new hanging sign and only after that opens its menu, while PM just opens the editor of the clicked one.

dktapps:
Oh, I see the problem. That's pretty annoying. I don't think any other block behaves like that.
So basically if a sign could be placed, it shouldn't open the editor...

src/block/CeilingCenterHangingSign.php
```diff
@@ -44,19 +46,18 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
...
 		return parent::place($tx, $item, $blockReplace, $blockClicked, $face, $clickVector, $player);
 	}
+
+	private function canBeSupportedAt(Block $block) : bool{
+		$supportBlock = $block->getSide(Facing::UP);
+		return
+			$supportBlock->getSupportType(Facing::DOWN)->hasCenterSupport() ||
+			$supportBlock instanceof WallHangingSign ||
+			$supportBlock instanceof CeilingEdgesHangingSign ||
+			$supportBlock instanceof CeilingCenterHangingSign;
```
dktapps: Probably should use tags for this

src/block/CeilingEdgesHangingSign.php
```diff
@@ -48,4 +51,12 @@ public function place(BlockTransaction $tx, Item $item, Block $blockReplace, Blo
 
 		return parent::place($tx, $item, $blockReplace, $blockClicked, $face, $clickVector, $player);
 	}
+
+	private function canBeSupportedAt(Block $block) : bool{
+		$supportBlock = $block->getSide(Facing::UP);
+		return
+			$supportBlock->getSupportType(Facing::DOWN) === SupportType::FULL ||
+			($supportBlock instanceof WallHangingSign && Facing::axis($supportBlock->getFacing()) === Facing::axis($this->facing)) ||
+			$supportBlock instanceof CeilingEdgesHangingSign;
```
dktapps: This should check the axis for ceiling edges too

dktapps: Basically the same logic as for selecting the sign type in HangingSign getPlacementBlock()

src/item/HangingSign.php
```diff
@@ -40,13 +44,26 @@ public function __construct(
...
+			//we select the center variant when support is edge/wall sign with perpendicular player facing,
+			//support is a center sign itself, or support provides center support.
+			//otherwise use the edge variant.
+			$support = $blockReplace->getSide(Facing::UP);
+			$result =
+				(($support instanceof CeilingEdgesHangingSign || $support instanceof WallHangingSign) && ($player === null || Facing::axis($player->getHorizontalFacing()) !== Facing::axis($support->getFacing()))) ||
+				$support instanceof CeilingCenterHangingSign ||
+				$support->getSupportType(Facing::DOWN) === SupportType::CENTER ?
+					$this->centerPointCeilingVariant :
+					$this->edgePointCeilingVariant;
```
dktapps, line 63: I'm thinking this could logically be reduced to:
`$this->edgePointCeilingVariant->canBeSupportedAt($support) ? $this->edgePointCeilingVariant : $this->centerPointCeilingVariant`
Since ceiling center signs have a wider range of acceptable support cases than ceiling edge signs.

If we could publicise that logic or otherwise extract it, we could do that. What do you think?
Also raises a more general question around `StaticSupportTrait->canBeSupportedAt()` on whether it should be public or not. I'm undecided.

src/block/CeilingEdgesHangingSign.php
```diff
@@ -33,6 +35,7 @@
 
 final class CeilingEdgesHangingSign extends BaseSign implements HorizontalFacing{
 	use HorizontalFacingTrait;
+	use StaticSupportTrait;
```
dktapps: I don't think we can use `StaticSupportTrait` for this. Since the support conditions depend on the current state of the block.

`StaticSupportTrait` logic is supposed to be independent of the block's state.

dktapps (dismissed):

dktapps (approved):

PR #6780, opened 2025-08-27
infinite effects

fadijalal07:
- added a way to add infinite effects

bcz its gud

### API changes

yes u can add bool $infinite to effect instance

## Backwards compatibility

maybe idk

## Follow-up

change the command usage + the message it sends

## Tests

image

### DIFF
--- a/src/command/defaults/EffectCommand.php
+++ b/src/command/defaults/EffectCommand.php
@@ -75,12 +75,19 @@ class EffectCommand extends VanillaCommand{
 		$amplification = 0;
 
 		if(count($args) >= 3){
-			if(($d = $this->getBoundedInt($sender, $args[2], 0, (int) (Limits::INT32_MAX / 20))) === null){
-				return false;
+			if(strtolower($args[2]) === "infinite"){
+				$duration = -1;
+				$infinite = true;
+			}else{
+				if(($d = $this->getBoundedInt($sender, $args[2], 0, (int) (Limits::INT32_MAX / 20))) === null){
+					return false;
+				}
+				$duration = $d * 20;
+				$infinite = false;
 			}
-			$duration = $d * 20; //ticks
 		}else{
 			$duration = null;
+			$infinite = false;
 		}
 
 		if(count($args) >= 4){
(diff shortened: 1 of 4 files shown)


dktapps:
This is BC breaking for the same reason as #6531

fadijalal07:
> This is BC breaking for the same reason as #6531 

what do you suggest. how can we go onwards for it to not be BC break. sir

dktapps:
https://github.com/pmmp/PocketMine-MP/pull/6531#issuecomment-2498379655

fadijalal07:
> https://github.com/pmmp/PocketMine-MP/pull/6531#issuecomment-2498379655

can't i keep my same code, but just remove infinite from constructor and in the isInfinite function i return $this->duration === Limits::INT32_MAX

fadijalal07:
> https://github.com/pmmp/PocketMine-MP/pull/6531#issuecomment-2498379655


so just remove the infinite parameter in construct and in the constructor i do $this->infinite = $this->duration === Limits::INT32_MAX

dktapps:
I don't see why a separate field is needed

DaisukeDaisuke:
Visual infinite effects can be implemented easily with a plugin. There is no need to include this in the PMMP core
In Japanese language mode, the infinity symbol causes garbled text, so it is not user-friendly.

https://github.com/DaisukeDaisuke/pmScriptPlugins/blob/main/plugins/InfiniteEffects.php

dktapps:
Well it's not the first time someone PR'd this feature, so it seems in demand for some reason. I'm fine with it as long as it's not breaking BC.

Beats me why people care more about this than other gameplay features when you can just set a really long duration though...

Wraith0x10:
> Well it's not the first time someone PR'd this feature, so it seems in demand for some reason. I'm fine with it as long as it's not breaking BC.
> 
> Beats me why people care more about this than other gameplay features when you can just set a really long duration though...

Imo an infinite symbol looks better than a really long duration time

ShockedPlot7560:
> In Japanese language mode, the infinity symbol causes garbled text, so it is not user-friendly.

If we add it as a translation string, the problem will be solved.

ShockedPlot7560 (commented):
The new parameter $infinite seems redundant as a duration to -1 will make the same.

SmallkingDev:
An easy solution without breaking BC would be to add helper methods, for example:

```php
public const INFINITE_DURATION = -1;

public function isInfinite(): bool {
    return $this->duration === self::INFINITE_DURATION;
}

// It feels like a workaround to me, but maybe we can use setDuration instead.
public function setInfinite(bool $infinite): void {
    $this->duration = $infinite ? self::INFINITE_DURATION : 0;
}
```

As for the infinity symbol showing differently across locales, that’s not really a problem — users who care about the visuals can simply use a texture pack.

dktapps:
As I said previously we can't use -1 for duration, it's BC breaking.

dktapps:
> > Well it's not the first time someone PR'd this feature, so it seems in demand for some reason. I'm fine with it as long as it's not breaking BC.
> > Beats me why people care more about this than other gameplay features when you can just set a really long duration though...
> 
> Imo an infinite symbol looks better than a really long duration time

Sure, it's just sad that this is the type of thing that actually gets people to engage with the repo, and not more useful changes

dktapps:
Following issues remain with this PR:
- Periodic effects like poison, regeneration etc. will not work
- `EntityEffectRemoveEvent` will not work
- BC break for plugins doing `getDuration() <= 0` to check for expired effects
- `EffectCommand` doesn't appear to adjust output messages

PR #6783, opened 2025-08-29
Release 5.33.0

dktapps:
DO NOT SQUASH MERGE!

Changelog only up for review


ipad54 (commented):
I guess it’s also worth to mention about addition of `getMaxAge()` and `getPlacementTransaction()`.

dktapps:
Good catch, thanks

ipad54 (dismissed):

PR #6786, opened 2025-08-31
First look at strings for block type IDs

### LINKED ISSUE #6062: BlockTypeIds::newId() is not synchronized between threads
## Description
Block type IDs must be the same across all threads, as they are used in the form of state IDs to transfer chunks between threads for sending over the network.

This can become problematic if one particular plugin fails to initialize its blocks (or if the setup happens in an unusual order), as the type IDs given will then be different.

The current recommendation is for plugins to copy the type ID from the main thread rather than using ::newId() within any other threads. However, this creates additional problems: plugins could then configure new blocks whose IDs conflict with those of existing custom blocks, as the system won't be aware of the fact that these dynamic type IDs were given by another thread. (In short, if a plugin creates a new block type within an AsyncTask, it might be given a type ID that's already in use).

dktapps:
This solves the following issues:
- Conflicts on BlockTypeIds every time someone makes a PR to add a new block
- Type IDs not consistent on different threads

We still have an auto assigned type number internally, which is used to compute state IDs for runtime chunk storage. However, this is not directly exposed to plugins, and is fully dynamic.

FastChunkSerializer, instead of directly passing state IDs to the next thread, now deconstructs state IDs into their type number and state data componets, and writes a string type name + state data separately. This allows the blocks to be consistently decoded on the destination thread, particularly when custom blocks are involved, as custom blocks are expected to claim unique string type IDs

Plugins implementing custom blocks **may** use their Minecraft save ID as their type ID, however this is not required. A type ID must, however, be globally unique and must not conflict with any other blocks.

I haven't yet measured the performance impact to FastChunkSerializer from this change, however, I do expect some minor performance degradation, since individual calls on BinaryStream are slower than a bulk pack() / unpack() call.

### Related issues & PRs
- Fixes #6062 

### API changes
- `Block->getTypeId()` now returns a `string`
- All `BlockTypeIds` constants are now converted to `string`
- `BlockIdentifier` now accepts `string` instead of `int`
- Several new static methods in `BlockTypeIds` to facilitate internal functionality

### Behavioural changes
Functionally everything should be basically the same.

## Backwards compatibility
BC breaks to APIs
`Block->getStateId()` may now change without warning between versions. This was true previously too (since the definition of the state data could change), but the type numbers may also move around without warning now, which will affect the resulting state ID. The function _is_ marked as `@internal`, but that never seems to stop anyone...

## Follow-up
Consider using incrementing IDs in chunk runtime storage, and using type ID + state data as a hash key to look up state ID, rather than using type number + state data as state ID as we do currently. This would allow getting rid of type numbers completely, make chunk block ID validation easier, and remove current restrictions from state data bits (although expanding state data bits would probably require rethinking `RuntimeBlockStateRegistry`).

## Tests
Playtested locally and PHPUnit tested.


dktapps:
This is stalled pending improvements to FastChunkSerializer and/or the structure of blockstate IDs stored in subchunks. The process of converting a state ID into a string type ID + unmasked state data feels like a hack, and will almost definitely be slow.

dktapps:
Decided not to proceed with this. It will almost certainly be much less BC breaking and overall faster to just plug in a `ThreadSafeArray` to `BlockTypeIds` that will be shared on all threads, and just have `BlockTypeIds::newId()` accept a string ID to lookup the numeric ID.

Issue #6649, opened 2025-03-11
Separate item network NBT from persistent NBT


dktapps:
### Problem description

Currently we send all data on an item over the network. This is fine for small items, but for larger complex ones like shulker boxes, it's both unnecessary and harmful (see #4664, #4665).

I was hopeful that the ItemStackRequest system would solve this eventually by removing all places where the client could send item NBT to the server, but this has turned out not to be the case.

### Proposed solution

A network-specific variant of `Item->serializeCompoundTag()` that only includes data needed to make an item display correctly. This would exclude custom plugin NBT and potentially strip unnecessary vanilla data like inventory contents, which can't be viewed client-side anyway.

This has a precedent in tiles, where we already differentiate between network and persistent NBT.

dktapps:
This also affects performance because of `InventoryManager`'s need to compare NBT for prediction verification.

Issue #6149, opened 2023-11-09
World I/O performance degradation in PM5


dktapps:
## Description
Due to the increased amount of work needed to load chunks stored in the modern Bedrock format, PM5 servers have often seen noticeable performance losses compared to PM4 in factions and similar server types, where the chunks are frequently loaded and unloaded.

I believe this performance drop is due to the large amount of NBT that needs to be processed during chunk loading, but this has yet to be profiled and confirmed.

dktapps:
Profiling indicates that while NBT is a large part of this problem, there were also performance issues in `BlockStateUpgrader` (which have now been addressed). Further improvements are still being worked on.

PR #6790, opened 2025-09-02
Strip unnecessary NBT from network items

### LINKED ISSUE #4665: You can install a block with any NBT length by clicking on the interactive block
### Issue description
If you get the same item as in #4664, then you can place it by clicking on the interactive blocks. If you spam with clicks, the server will crash from lack of RAM 

- Expected result: Chest is not placed 
- Actual result: Server crashes 

1. Turn on creative mode 
2. Place the chest
3. Copy it by pressing the mouse wheel and the control key 
4. Put the received item in the chest
5. Repeat steps several times
6. Spam by installing this chest on interactive blocks 
7. Server crashed 

### OS and versions

* PocketMine-MP: 4.0.3 795ebd182461f96beeb47b067f7fe33e76bc9059
* PHP: 8.0.12
* Using JIT: no
* Server OS: win

### Plugins
none

[Sat_Dec_18-20.33.36-CAT_2021.log](https://github.com/pmmp/PocketMine-MP/files/7739756/Sat_Dec_18-20.33.36-CAT_2021.log)

dktapps:
Item containers like shulker boxes, or chests with block entity data obtained with ctrl+middle click, will often have extremely large NBT payloads. This problem gets exponentially worse in cases where it's possible to nest inventories, as in #4665.

We can't easily address this at the core level, because tiles are not able to exist within items (due to position requirement) so we don't have a good way to avoid this useless NBT in the first place. However, we can strip it before the item is sent to the client, which dramatically reduces the network costs of such items, as well as removing any reason the client could have to send such enormous items to the server.

A fully loaded shulker box with written books now takes only 5-6 KB on the wire instead of ~1 MB, which is a substantial improvement.

Dealing with this in a less hacky way is currently blocked on #6147.

### Related issues & PRs
- Fixes #4665 
- Fixes, kinda, #6649 , although this needs to be redone in a better way once #6147 is dealt with

### API changes
None

### Behavioural changes
Significant network performance improvement when dealing with items containing inventories or block entity data (containing inventories)
Pains were taken to ensure that the lore of shulker boxes doesn't break with this change, however it is possible I may have missed something. Nonetheless, I think this change is worth it.

## Backwards compatibility
Backwards compatible

## Follow-up
Redo this in PM6 once #6149 is done

## Tests
Playtested with shulkerboxes and nested chests

### DIFF
--- a/src/network/mcpe/convert/TypeConverter.php
+++ b/src/network/mcpe/convert/TypeConverter.php
@@ -23,6 +23,7 @@ declare(strict_types=1);
 
 namespace pocketmine\network\mcpe\convert;
 
+use pocketmine\block\tile\Container;
 use pocketmine\block\VanillaBlocks;
 use pocketmine\crafting\ExactRecipeIngredient;
 use pocketmine\crafting\MetaWildcardRecipeIngredient;
@@ -31,10 +32,16 @@ use pocketmine\crafting\TagWildcardRecipeIngredient;
 use pocketmine\data\bedrock\BedrockDataFiles;
 use pocketmine\data\bedrock\item\BlockItemIdMap;
 use pocketmine\data\bedrock\item\ItemTypeNames;
+use pocketmine\data\SavedDataLoadingException;
 use pocketmine\item\Item;
 use pocketmine\item\VanillaItems;
+use pocketmine\nbt\LittleEndianNbtSerializer;
+use pocketmine\nbt\NBT;
 use pocketmine\nbt\NbtException;
 use pocketmine\nbt\tag\CompoundTag;
+use pocketmine\nbt\tag\ListTag;
+use pocketmine\nbt\tag\Tag;
+use pocketmine\nbt\TreeRoot;
 use pocketmine\network\mcpe\protocol\serializer\ItemTypeDictionary;
 use pocketmine\network\mcpe\protocol\serializer\PacketSerializer;
 use pocketmine\network\mcpe\protocol\types\GameMode as ProtocolGameMode;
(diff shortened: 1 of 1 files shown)


src/network/mcpe/convert/TypeConverter.php
```diff
@@ -197,6 +206,90 @@ public function netRecipeIngredientToCore(ProtocolRecipeIngredient $ingredient)
...
+	 * NBT but different server-side NBT.
+	 */
+	protected function hashNBT(Tag $tag) : string{
+		$encoded = (new LittleEndianNbtSerializer())->write(new TreeRoot($tag));
+		return hash('sha256', $encoded, binary: true);
+	}
+
+	/**
+	 * TODO: HACK!
+	 * Strips non-viewable data from the NBT of an item.
```
dries-c: it strips it of a copy of the item, might be nice to include that in the doc

src/network/mcpe/convert/TypeConverter.php
```diff
@@ -197,6 +206,90 @@ public function netRecipeIngredientToCore(ProtocolRecipeIngredient $ingredient)
...
 
+	/**
+	 * Strips unnecessary block actor NBT from items that have it.
+	 * This tag can potentially be extremely large, and is not read by the client anyway.
+	 */
+	protected function stripBlockEntityNBT(CompoundTag $tag) : bool{
+		if(
+			($blockEntityTag = $tag->getTag(Item::TAG_BLOCK_ENTITY_TAG)) !== null &&
+			$blockEntityTag instanceof CompoundTag &&
+			$blockEntityTag->count() > 0
```
dries-c: Does it make a difference if it's an empty CompoundTag?

dktapps: Yeah, we don't want to put a hash of the full NBT in the item if it wasn't modified

dries-c (approved):

PR #6799, opened 2025-09-10
First pass ext-encoding support (high-level network I/O and read-only data)

dktapps:
> [!WARNING]
> Please note that this has **not been thoroughly tested** yet and is not safe for production use. It may crash your game or the server may be unstable.

ext-encoding provides significant performance improvements to data encoding and decoding, which will mostly affect Bedrock protocol packet processing. Since this makes up a significant chunk of a server's runtime, noticeable performance gains are expected.

This first phase focuses primarily on network I/O and on decoding data only. This allows us to iron out any issues in the wild without any lasting damage being done.

The second phase will integrate ext-encoding in persistent world storage once data has been gathered on stability from the first phase.

This includes breaking changes to the following libraries:

- BedrockProtocol
- RakLib

However, these are both considered part of the network system, and are not covered by the API version guarantee.
Thus, this should be able to be integrated with a minor release.

## Future scope
- ext-encoding to be used for data storage I/O (NBT, world data storage)
- `BinaryUtils`, `BinaryStream`, `Limits` and `BinaryDataException` should now be considered deprecated, and will be removed in a major release
- We may be able to get further performance improvements by allowing the NBT library to reuse write buffers
- We may be able to get further performance improvements by avoiding copies when reading from batch packets (sub buffers with a specified start and end offset)
- To be continued...


dktapps:
Scope of this PR has ben reduced to non-permanent data only, like network I/O and read-only data. This avoids putting user data at risk, and if there are any significant bugs in the extension, no permanent damage will be done.

PR #6826, opened 2025-10-06
Implement exposure calculation for explosions

Wraith0x10:
Right now in PM, explosion damage is always calculated as if players have full exposure, which causes odd situations like taking damage through walls. This PR adds proper exposure checks using raycasts to make explosions behave more accurately and realistically.

A lot of the logic for this was referenced from [Dragonfly](https://github.com/df-mc/dragonfly/blob/master/server/block/explosion.go#L183). 

### Behavioural changes
Explosion damage now accurately matches vanilla behavior by using a new `getExposure()` method in the `Explosion` class to calculate real exposure.

## Follow-up
One thing I wasn’t completely sure about while making this PR was whether the `lerp()` method belongs inside `Explosion`. It might make more sense to move it to `Math` instead.

## Tests
Here’s a video demonstrating the behavior before and after the changes.

https://github.com/user-attachments/assets/5498dbdc-ea00-4a9f-a7b4-40482f0d4158

### DIFF
--- a/src/world/Explosion.php
+++ b/src/world/Explosion.php
@@ -280,8 +280,8 @@ class Explosion{
 		$diff = (new Vector3($bb->getXLength(), $bb->getYLength(), $bb->getZLength()))->multiply(2)->add(1, 1, 1);
 		$step = new Vector3(1.0 / $diff->x, 1.0 / $diff->y, 1.0 / $diff->z);
 
-		$xOffset = (1.0 - floor($diff->x) / $diff->x) / 2.0;
-		$zOffset = (1.0 - floor($diff->z) / $diff->z) / 2.0;
+		$xOffset = (1.0 - (floor($diff->x) / $diff->x)) / 2.0;
+		$zOffset = (1.0 - (floor($diff->z) / $diff->z)) / 2.0;
 
 		$checks = 0.0;
 		$misses = 0.0;
--- a/src/world/Explosion.php
+++ b/src/world/Explosion.php
@@ -277,7 +277,7 @@ class Explosion{
 	public function getExposure(World $world, Vector3 $origin, Entity $entity) : float{
 		$bb = $entity->getBoundingBox();
 
-		$diff = new Vector3($bb->getXLength(), $bb->getYLength(), $bb->getZLength())->multiply(2)->add(1, 1, 1);
+		$diff = (new Vector3($bb->getXLength(), $bb->getYLength(), $bb->getZLength()))->multiply(2)->add(1, 1, 1);
 		$step = new Vector3(1.0 / $diff->x, 1.0 / $diff->y, 1.0 / $diff->z);
 
 		$xOffset = (1.0 - floor($diff->x) / $diff->x) / 2.0;
--- a/src/world/Explosion.php
+++ b/src/world/Explosion.php
@@ -38,6 +38,7 @@ use pocketmine\item\VanillaItems;
 use pocketmine\math\AxisAlignedBB;
 use pocketmine\math\Facing;
 use pocketmine\math\Vector3;
+use pocketmine\math\VoxelRayTrace;
 use pocketmine\utils\AssumptionFailedError;
 use pocketmine\utils\Utils;
 use pocketmine\world\format\SubChunk;
@@ -218,13 +219,14 @@ class Explosion{
 
 			if($distance <= 1){
 				$motion = $entityPos->subtractVector($this->source)->normalize();
+				$exposure = $this->getExposure($this->world, $this->source, $entity);
 
-				$impact = (1 - $distance) * ($exposure = 1);
+				$impact = (1 - $distance) * $exposure;
 
 				$damage = (int) ((($impact * $impact + $impact) / 2) * 8 * $explosionSize + 1);
 
 				if($this->what instanceof Entity){
-					$ev = new EntityDamageByEntityEvent($this->what, $entity, EntityDamageEvent::CAUSE_ENTITY_EXPLOSION, $damage);
+					$ev = new EntityDamageByEntityEvent($this->what, $entity, EntityDamageEvent::CAUSE_ENTITY_EXPLOSION, $damage, [], $exposure);
 				}elseif($this->what instanceof Block){
 					$ev = new EntityDamageByBlockEvent($this->what, $entity, EntityDamageEvent::CAUSE_BLOCK_EXPLOSION, $damage);
 				}else{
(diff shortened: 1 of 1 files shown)


src/world/Explosion.php
```diff
@@ -269,6 +271,63 @@ public function explodeB() : bool{
...
+	 */
+	public function getExposure(World $world, Vector3 $origin, Entity $entity) : float{
+		$bb = $entity->getBoundingBox();
+
+		$boxMin = new Vector3($bb->minX, $bb->minY, $bb->minZ);
+		$boxMax = new Vector3($bb->maxX, $bb->maxY, $bb->maxZ);
+
+		$diff = $boxMax->subtractVector($boxMin)->multiply(2)->add(1, 1, 1);
+
+		if($diff->x <= 0.0 || $diff->y <= 0.0 || $diff->z <= 0.0){
```
dktapps: Don't think this can ever be less than zero. The AABB constructor doesn't allow it.

src/world/Explosion.php
```diff
@@ -269,6 +271,63 @@ public function explodeB() : bool{
...
+			return 0.0;
+		}
+
+		$step = new Vector3(1.0 / $diff->x, 1.0 / $diff->y, 1.0 / $diff->z);
+
+		if($step->x < 0.0 || $step->y < 0.0 || $step->z < 0.0){
+			return 0.0;
+		}
```
dktapps: Same again here, the components of $diff should always be positive

src/world/Explosion.php
```diff
@@ -269,6 +271,63 @@ public function explodeB() : bool{
...
 	}
 
+	/**
+	 * Returns the explosion exposure of an entity, used to calculate explosion impact.
+	 */
+	public function getExposure(World $world, Vector3 $origin, Entity $entity) : float{
+		$bb = $entity->getBoundingBox();
+
+		$boxMin = new Vector3($bb->minX, $bb->minY, $bb->minZ);
+		$boxMax = new Vector3($bb->maxX, $bb->maxY, $bb->maxZ);
```
dktapps: These vectors could be replaced by `new Vector3($bb->getXLength(), $bb->getYLength(), $bb->getZLength())`

dktapps (commented):
Am I understanding correct that this attempts to cast rays at the entity per square of AABB face? So larger entities will get hit by more rays

src/world/Explosion.php
Wraith0x10: Done

src/world/Explosion.php
Wraith0x10: Yep, removed those checks

Wraith0x10:
> Am I understanding correct that this attempts to cast rays at the entity per square of AABB face? So larger entities will get hit by more rays

Indeed, you are correct

src/world/Explosion.php
```diff
@@ -269,6 +271,51 @@ public function explodeB() : bool{
 		return true;
 	}
 
+	/**
+	 * Returns the explosion exposure of an entity, used to calculate explosion impact.
+	 */
+	public function getExposure(World $world, Vector3 $origin, Entity $entity) : float{
+		$bb = $entity->getBoundingBox();
+
+		$diff = new Vector3($bb->getXLength(), $bb->getYLength(), $bb->getZLength())->multiply(2)->add(1, 1, 1);
```
dktapps:
```suggestion
		$diff = (new Vector3($bb->getXLength(), $bb->getYLength(), $bb->getZLength()))->multiply(2)->add(1, 1, 1);
```
We still support 8.1 on this branch

Wraith0x10: Ah alright, resolved

src/world/Explosion.php
```diff
@@ -269,6 +271,51 @@ public function explodeB() : bool{
 		return true;
 	}
 
+	/**
+	 * Returns the explosion exposure of an entity, used to calculate explosion impact.
+	 */
+	public function getExposure(World $world, Vector3 $origin, Entity $entity) : float{
```
dktapps: Better keep this private for now

src/world/Explosion.php
```diff
@@ -269,6 +271,51 @@ public function explodeB() : bool{
...
+		$bb = $entity->getBoundingBox();
+
+		$diff = (new Vector3($bb->getXLength(), $bb->getYLength(), $bb->getZLength()))->multiply(2)->add(1, 1, 1);
+		$step = new Vector3(1.0 / $diff->x, 1.0 / $diff->y, 1.0 / $diff->z);
+
+		$xOffset = (1.0 - floor($diff->x) / $diff->x) / 2.0;
+		$zOffset = (1.0 - floor($diff->z) / $diff->z) / 2.0;
```
dktapps: These should have extra braces to show the operation order intent, otherwise this might be subject to operator precedence issues. I don't know whether you wanted to do `1 - (floor($diff->z) / $diff->z)` or `(1 - floor($diff->z)) / $diff->z`.

src/world/Explosion.php
```diff
@@ -269,6 +271,51 @@ public function explodeB() : bool{
...
+					foreach(VoxelRayTrace::betweenPoints($origin, $point) as $pos){
+						$block = $world->getBlock($pos);
+						if($block->calculateIntercept($origin, $point) !== null){
+							$collided = true;
+							break;
+						}
+					}
+
+					if(!$collided){
+						$misses++;
```
dktapps: Since it's the number of rays striking the entity that we care about, `misses` is a bit misleading for this. Its actual meaning is that it missed all blocks (meaning that the ray will hit the entity).

src/world/Explosion.php
```diff
@@ -269,6 +271,51 @@ public function explodeB() : bool{
...
+
+					foreach(VoxelRayTrace::betweenPoints($origin, $point) as $pos){
+						$block = $world->getBlock($pos);
+						if($block->calculateIntercept($origin, $point) !== null){
+							$collided = true;
+							break;
+						}
+					}
+
+					if(!$collided){
```
dktapps: `intercepted` would be a better name for this

src/world/Explosion.php
```diff
@@ -282,4 +329,8 @@ public function setFireChance(float $fireChance) : void{
 		}
 		$this->fireChance = $fireChance;
 	}
-}
+
+	private function lerp(float $t, float $a, float $b) : float{
```
dktapps: This can be `static`. Also, `t` could perhaps be renamed to `scale` for easier reading.

src/world/Explosion.php
dktapps, line 234: Probably exposure should be shown in these other events too

src/world/Explosion.php
dktapps: The `world` parameter also isn't necessary since the class already has a `world` property

src/world/Explosion.php
Wraith0x10: What should we name it?

src/world/Explosion.php
Wraith0x10: The reason im passing exposure to EntityDamageByEntityEvent is because it accepts knockback

src/world/Explosion.php
dktapps: Well, `hits` would be the most obvious

src/world/Explosion.php
```diff
@@ -269,6 +271,51 @@ public function explodeB() : bool{
...
+	 * Returns the explosion exposure of an entity, used to calculate explosion impact.
+	 */
+	private function getExposure(Vector3 $origin, Entity $entity) : float{
+		$bb = $entity->getBoundingBox();
+
+		$diff = (new Vector3($bb->getXLength(), $bb->getYLength(), $bb->getZLength()))->multiply(2)->add(1, 1, 1);
+		$step = new Vector3(1.0 / $diff->x, 1.0 / $diff->y, 1.0 / $diff->z);
+
+		$xOffset = (1.0 - (floor($diff->x) / $diff->x)) / 2.0;
+		$zOffset = (1.0 - (floor($diff->z) / $diff->z)) / 2.0;
```
dktapps, line 284: Also not super clear what the purpose of these offsets are. For a player, `diff->x` and `diff->z` would be `2.2` (`(0.6 * 2) + 1`).

- `floor(2.2) / 2.2` = `0.909090909`
- `1 - 0.909090909` = `0.0909090909`
- `0.0909090909 / 2` = `0.0454545454`

This doesn't seem big enough to make any significant difference, considering how it's applied? Are you sure this is implemented correctly?

src/world/Explosion.php
dktapps: Right, I see. Though I'd think the plain exposure would give too large a knockback under fully exposed conditions, no? Default is 0.4-0.5, if the player was fully exposed they'd get a knockback of 1.0.

src/world/Explosion.php
Wraith0x10: Dragonfly can be blamed on that one, if this is an issue, it exists there too
https://github.com/df-mc/dragonfly/blob/master/server/block/explosion.go#L183

src/world/Explosion.php
Wraith0x10: Good catch, let me test that out.

src/world/Explosion.php
dktapps: Yeah this is definitely not right, I'm reading the JE code now

Wraith0x10: Are we sure JE matches BE with this logic?

Wraith0x10: If so, what do you recommend we change

src/world/Explosion.php
Wraith0x10: Okay yeah you were right. 
If we change that to `0.4 * $exposure` we start to get values looking like this:

string(24) "Exposure: 0 Knockback: 0"
string(29) "Exposure: 0.4 Knockback: 0.16"
string(26) "Exposure: 1 Knockback: 0.4"
string(29) "Exposure: 0.4 Knockback: 0.16"

From my tests in vanilla it feels exactly the same.

src/world/Explosion.php
dktapps: Never mind, it appears that it is the same as JE, it's just because the operations were done differently

dktapps (approved):

PR #6835, opened 2025-10-09
Fix StringToItemParser::override() not fully overriding aliases

### LINKED ISSUE #6815: `StringToItemParser::override()` doesn't fully override aliases
### Plugin information

Bug happens without plugins

### Problem description

`StringToTParser::override()` will not trigger this logic:
https://github.com/pmmp/PocketMine-MP/blob/d6e05c90fea2ab3afb1e1e118a4839248a83b70c/src/item/StringToItemParser.php#L1596-L1597

This means that `StringToItemParser->lookupAliases()` will not return the correct results when an alias is overridden by a plugin, as overridden aliases will still be considered as bound to the old item.

### Expected behaviour

`StringToItemParser->lookupAliases($customItem)` should return all aliases it's bound, not just the ones it overrode.

### PocketMine-MP version

5.34.0

### PHP version

irrelevant

### Server OS

irrelevant

irrelevant

RenzMc:
Fixes #6815

This PR fixes an issue where `StringToItemParser::override()` did not fully override aliases,
causing `lookupAliases()` to return incorrect results.

### DIFF
--- a/src/item/StringToItemParser.php
+++ b/src/item/StringToItemParser.php
@@ -1602,15 +1602,10 @@ final class StringToItemParser extends StringToTParser{
 		if($oldItem !== null){
 			$oldStateId = $oldItem->getStateId();
 			unset($this->reverseMap[$oldStateId][$alias]);
-			if($this->reverseMap[$oldStateId] === []){
-				unset($this->reverseMap[$oldStateId]);
-			}
 		}
 		parent::override($alias, $callback);
-		$newItem = $this->parse($alias);
-		if($newItem !== null){
-			$this->reverseMap[$newItem->getStateId()][$alias] = true;
-		}
+		$newItem = $callback($alias);
+		$this->reverseMap[$newItem->getStateId()][$alias] = true;
 	}
 
 	/** @phpstan-param \Closure(string $input) : Block $callback */
--- a/src/item/StringToItemParser.php
+++ b/src/item/StringToItemParser.php
@@ -38,6 +38,7 @@ use pocketmine\item\VanillaItems as Items;
 use pocketmine\utils\SingletonTrait;
 use pocketmine\utils\StringToTParser;
 use function array_keys;
+use function count;
 use function strtolower;
 
 /**
--- a/src/item/StringToItemParser.php
+++ b/src/item/StringToItemParser.php
@@ -1597,6 +1597,16 @@ final class StringToItemParser extends StringToTParser{
 		$this->reverseMap[$item->getStateId()][$alias] = true;
 	}
 
+	public function override(string $alias, \Closure $callback) : void{
+		// Remove the alias from the old item's reverse map
+		foreach($this->reverseMap as $stateId => $aliases){
+			unset($this->reverseMap[$stateId][$alias]);
+		}
+		parent::override($alias, $callback);
+		$item = $callback($alias);
+		$this->reverseMap[$item->getStateId()][$alias] = true;
+	}
+
 	/** @phpstan-param \Closure(string $input) : Block $callback */
 	public function registerBlock(string $alias, \Closure $callback) : void{
 		$this->register($alias, fn(string $input) => $callback($input)->asItem());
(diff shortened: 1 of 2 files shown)


src/item/StringToItemParser.php
```diff
@@ -1597,6 +1597,16 @@ public function register(string $alias, \Closure $callback) : void{
 		$this->reverseMap[$item->getStateId()][$alias] = true;
 	}
 
+	public function override(string $alias, \Closure $callback) : void{
+		// Remove the alias from the old item's reverse map
+		foreach($this->reverseMap as $stateId => $aliases){
+			unset($this->reverseMap[$stateId][$alias]);
+		}
```
dktapps, line 1609: This could invoke the old callback to get the previously bound state ID instead of iterating the entire map
Also this should account for the possibility there may be no aliases left

src/item/StringToItemParser.php
```diff
@@ -1597,6 +1597,16 @@ public function register(string $alias, \Closure $callback) : void{
 		$this->reverseMap[$item->getStateId()][$alias] = true;
 	}
 
+	public function override(string $alias, \Closure $callback) : void{
+		// Remove the alias from the old item's reverse map
```
dktapps: Remove the redundant ChatGPT comment please

dktapps (changes requested):
Please do not use ChatGPT to send PRs

dktapps:
Please also add unit tests to verify the fix

RenzMc:
> Please also add unit tests to verify the fix

hi dktapps,

i have applied all requested changes. the redundant chatgpt comment is removed, override method is fixed, and unit tests are added to verify the behavior. please review when you have time.

best regards
- Renz

src/item/StringToItemParser.php
```diff
@@ -1603,13 +1602,15 @@ public function override(string $alias, \Closure $callback) : void{
 		if($oldItem !== null){
 			$oldStateId = $oldItem->getStateId();
 			unset($this->reverseMap[$oldStateId][$alias]);
-			if(count($this->reverseMap[$oldStateId]) === 0){
+			if($this->reverseMap[$oldStateId] === []){
```
dktapps: This change doesn't seem necessary

src/item/StringToItemParser.php
```diff
@@ -1603,13 +1602,15 @@ public function override(string $alias, \Closure $callback) : void{
...
 				unset($this->reverseMap[$oldStateId]);
 			}
 		}
 		parent::override($alias, $callback);
-		$item = $callback($alias);
-		$this->reverseMap[$item->getStateId()][$alias] = true;
+		$newItem = $this->parse($alias);
+		if($newItem !== null){
+			$this->reverseMap[$newItem->getStateId()][$alias] = true;
+		}
```
dktapps: Same here

RenzMc: Hi again Dktaps, I’ve applied all requested changes. The reverseMap update is kept because parent::override() doesn’t re-register aliases, and removing it breaks reverse lookups. Ready for review

src/item/StringToItemParser.php
```diff
@@ -1597,6 +1597,17 @@ public function register(string $alias, \Closure $callback) : void{
 		$this->reverseMap[$item->getStateId()][$alias] = true;
 	}
 
+	public function override(string $alias, \Closure $callback) : void{
+		$oldItem = $this->parse($alias);
+		if($oldItem !== null){
+			$oldStateId = $oldItem->getStateId();
+			unset($this->reverseMap[$oldStateId][$alias]);
```
dktapps, line 1605: Again, this needs to account for there being no aliases left. It was correct before, I don't know why you changed it...

RenzMc: Hi @dktapps thanks for the review. I’ve restored the cleanup logic you requested, after unsetting the alias I now check if there are no aliases left using `count()` and remove the empty entry from `$this->reverseMap`All tests and CI checks pass

dktapps (approved):
For what it's worth, I don't recommend using AI, Copilot Agents etc. It's a lot easier to get things done if you learn how to do it yourself, rather than trying to get an AI to not mess it up.

But we did eventually get to an end result, so I think this is OK to merge.

ipad54 (approved):

Issue #6519, opened 2024-11-17
Command target selectors


dktapps:
## Description
There are 3 tiers of this task, ranging from easy to hard:

- [ ] Implement basic (without arguments) `@p` (nearest player), `@r` (random player) and `@s` (self). These should be fairly easy as they each only support 1 target.
- [ ] Implement basic (without arguments) `@a` (all players) and `@e` (all entities). Many commands would have to be refactored to support multiple targets for these.
- [ ] Implement selector arguments. There are a ton of them I won't document here, but this would be a whole exercise of parsing in itself.

## Open questions
I'm not sure what the default behaviour should be where multi-world is concerned. Should selectors target all worlds? Only the default world unless `world=somethingElse` is provided? Please discuss.

## Justification
This is a needed feature that's gone ignored for a long time.

Issue #6506, opened 2024-11-15
Scrap `plugin.yml` commands


dktapps:
## Description
It's 2024 and most people just declare them in code anyway.

More seriously, command handlers are like event handlers, and the system we create them by should reflect that. It doesn't make much sense to separate the base info from the implementation the way we do right now, and no one likes being bound to `onCommand` (see #5960).

In addition, it's currently impractical to do stuff like using localized descriptions, and `plugin.yml` commands also aren't suited to a future where the command parser actually processes arguments into proper types instead of just passing `string[] $args` to everything.

The only real downside to scrapping these commands is that it won't be possible to tell what commands a plugin has without running the server and doing `/help`. However, since many people already don't use `plugin.yml` for commands, this is kinda a moot point.

We can reduce the friction of creating commands procedurally by implementing changes like #6063.

In the future, we could also implement a system using @doccomments and a magic interface like `Listener` for registering command handlers if there's a demand for it.

## Justification
The whole thing is misconceived. I wrote about this at length in #5960.

This has been discussed for years and years, and I'm frankly amazed that a formal change proposal has never been submitted before.

ShockedPlot7560:
My first thought on this subject is: should we completely remove `onCommand`? As the `ClosureCommand` currently exists, it can be ‘easily’ migrated. If that's the case, the `PluginOwned` interface, the functions derived from it and the PluginOwnedCommand are no longer needed and are destined to disappear with

dktapps:
Yeah, getting rid of `onCommand` is the main motivation for this. `ClosureCommand` provides an easy duct tape fix for plugins that break from this change.

dktapps:
I'm not sure what to do with `PluginOwned` though, I think it may still be of use to have commands signal which plugin they belong to? but in the absence of any requirement to pass a plugin, I'm not sure what to do with it...

ShockedPlot7560:
About `PluginOwned`, as currently, the main use in the core was to register yaml commands, are any developers using it on their side? Because a `CommandExecutor` is needed to work, it will become like `ClosureCommand`: you need a `Closure` instead of a `CommandExecutor`
My opinion would be to remove it too and the `Server::getPluginCommand` etc. will also disappear.
Because linking a command to plugin is optionnal (for now), to be have a right output of the getPluginCommand, we need to have a consistent way to link a command to a plugin, like listeners do. This is probably another part of the command rework

dktapps:
[...]

PR #6837, opened 2025-10-12
Pathfinder for new commands system

### LINKED ISSUE #6517: Client-side command hints
image

For this to work, PocketMine would need to automatically support command enumerations. I understand this might break backward compatibility, but since this feature is already part of vanilla Minecraft, it logically should be implemented in PocketMine-MP. This would allow for suggestions to appear when typing a command and could also make it interesting to enable the creation of custom command enumerations.

This means that within the Command class, there would be a preparation of the required options. One could consider adding a class or concept such as `CommandParamOption `or even `CommandParamType`.

### LINKED ISSUE #6518: Command overloading, & automatic command argument parsing & validation system

dktapps:
This PR implements a command overloading system. Partially inspired by @alvin0319's prior (abandoned) work on #3871, but with deeper integration.

## Relevant issues
- Closes #6518
- Closes #6517 (well, not entirely, but basic support is here)

## Highlights
- No more nasty switch-cases based on the number of args like we famously had in the `/tp` command
- No more checking `if(isset($args[$offset]))` - your callback simply won't be called if the number of arguments is wrong or the arguments can't be parsed the way you asked for
- Different operations of commands can be implemented in different functions, instead of putting them all into one giant if/elseif/else or a switch
- Very basic client side command hints are now supported
- Command usage messages are now built from overload parameter infos, instead of needing to be manually written
- Commands may have multiple usage messages (one per overload)
- `/me`, `/say`, `/tell` now handle quoted inputs correctly - previously `/say hello "some user"` would broadcast `hello some user`, but now broadcasts exactly what you typed, i.e. `hello "some user"`

## How it works
- Commands register one or more callbacks, along with parameter description information.
  - The callback can accept native types of the appropriate args. For example, if you want an int, you can accept `int $param`
  - The number of required args is inferred from the callback given by reflection. For example, if you have an optional `string` parameter, you can do `?string $param = null` and check for `null` to see if the parameter was specified or not.
  - Parameter descriptions are required along with the callable, to tell the system about stuff like:
    - Translatable name of the parameter, used in usage messages and client side suggestions
    - Accepted values, min/max etc
    - How to convert the raw string into the type you want (e.g. for parsing an arg into an `Item` object; see `MappedParameter`, and how it's used in commands like `/enchant`, `/effect`
    - Literals - these are provided as raw strings instead of `Parameter` objects - these aren't given to the callback, but must be typed exactly in the position specified (e.g. a subcommand name); see how these are used for `/effect <target> clear`, `/time set <timestamp>` etc
  - While not easily statically analysed, the command callback will be verified upon registration using reflection against the provided parameter infos. 
    - Your command callback must accept at least the types indicated by your parameter infos.
    - You can also accept wider (contravariant) types in your callback's parameters (e.g. you can have `IntRangeParameter("xp")` whose value is accepted by `int $xp`, `?int $xp`, `int|string $xp`, but not `string` $xp)
    - As mentioned above, literals are not passed to the callback.
- When a command string is received, `Command->executeOverloaded()` is called
- Each overload is tried one by one.
  - The first one that succeeds in parsing the whole command string with no left over tokens is used to run the command. (This is possibly not optimal, but it's the simplest way to do this for now.)
  - If no overloads can parse all the args successfully, the usages of all overloads are displayed to the user and no overload is called

## Gotchas
- No conflict resolution is done. If two overloads could both accept the user's input, the first one registered will be called currently. This may be changed in the future.
- Literal matching is done by bruteforce search. This might be slow for subcommands. This is an active area of research.

## Follow-up
- Implement support for composite parameters like `x y z` (which are actually 3x `RelativeFloat` parameters)
- Implement support for fixed enums
- Implement support for dynamic (soft) enums (this will be non-trivial, because there'll need to be a way to sync updates to clients)
- Add translation keys for various parameter names to `pocketmine/locale-data`
- Implement ambiguous overload detection (when a command line could be accepted by multiple overloads, we currently just use the first one)
- Explore auto-inferring `Parameter` infos from a provided callback instead of having them explicitly specified
  - This would be fine when all the info needed to parse the parameter is contained in the type, but wouldn't work for stuff like int/float bounds, dynamic enums, or mapped parameters
  - Attributes like `#[Bounds(0, 100)] int $param` could be explored, but this would only work for simple stuff known ahead of time
- Implement support for command selectors (#6519)
- Get rid of `plugin.yml` commands, which are hella dated, awful, and can't take advantage of this cool new system (#6506)

## API changes
- `Command->__construct()` now requires a non-empty list in `array $overloads`
- The following methods are added:
  - `Command->executeOverloaded()` - called by `SimpleCommandMap`
  - `Command->getOverloads()`
  - `Command->getUsages()` - returns usages for all overloads the given `CommandSender` has permissions for
  - `Command::fetchPermittedPlayerTarget()` - similar to `VanillaCommand->fetchPermittedPlayerTarget()`, but now accessible to all commands instead of just `VanillaCommand` inheritors
  - `CommandStringHelper::parseQuoteAwareSingle()` - parses a `"quoted string \"like this\""` into `quoted string "like this"` - similar to `parseQuoteAware()` but only parses 1 arg instead of the entire command
- The following methods are removed:
  - `Command->getPermissions()` - permissions are now defined per-overload instead of for the whole command
  - `Command->setPermissions()`
  - `Command->testPermission()`
  - `Command->testPermissionSilent()`
  - `Command->getUsage()` - command usages are now derived from overloads, and there may be more than one per command
  - `Command->setUsage()`
- The following classes are removed:
  - `VanillaCommand` - no longer served any purpose since it was all about manual arg parsing
- The following classes are added:
  - `LegacyCommand` - provides a similar API to the old-style `Command` for a quick & dirty upgrade path for old code if you don't want to switch to the new system immediately (this is also still used by `ClosureCommand`, `FormattedCommandAlias` and `PluginCommand` for the time being)
    - `LegacyCommand` will use your old-style hardcoded usage message if provided, but you should consider switching to the new system. This is only provided to ease upgrading and won't stick around forever.
  - `Parameter` and various subclasses in `pocketmine\command\overload` package
  - `CommandOverload` - accepts parameter info and callbacks to be given to `Command->__construct()`
  - `ParameterParseException` - thrown when a `Parameter` can't parse a command token
- `Utils::validateCallableSignature()` now accepts `Closure|Prototype` for both arguments


alvin0319:
Is it safe to use non-mb_* prefixed methods while parsing? Haven't looked code deeply but I saw a lot of substr and strlen used in code - wouldn't this affect to non-english commands?

dktapps:
It should be fine. We only care about byte lengths, and splittable whitespace chars will never be more than 1 byte.

dktapps:
@inxomnyaa raised concerns about performance when processing subcommands. I'm not sure to what extent this is a problem, but I have seen some plugins create much more complicated command signatures than vanilla. The way that overloading works, "subcommands" can be implemented as overloads using literals in this system, but it currently has to try each one, one at a time, to find a matching overload. This could become a problem if many overloads are present. This is sometimes the case with complex plugin commands, where nested commands blow up to create a large number of overload permutations.

Possibly, it might be advantageous to detect which overloads have the same first N parameters, so that if other overloads have the same leading arguments, we can eliminate multiple overloads at once in some cases without having to reparse the entire command over and over again.

It might also be worth exploring whether we can avoid bruteforce lookups when some or all overloads have a literal as their first parameter(s). This would be a common case for subcommands.

src/command/Command.php
```diff
@@ -26,34 +26,40 @@
...
-use function explode;
+use function array_unique;
+use function count;
 use function implode;
 use function str_replace;
 use function strtolower;
 use function trim;
-use const PHP_INT_MAX;
 
 abstract class Command{
```
SOF3: Should `Command` be non-abstract now? There are no abstract functions, and there doesn't seem to be a good reason why we need to force users to create a separate class for each command.

SOF3:
Optimizing overload selection shouldn't be a blocker for this PR. Plugins that are particularly concerned about huge number of overloads could always fall back to RawParameter to leverage their own optimized parsing algorithm.

src/command/Command.php
dktapps: Good catch. Maybe this also makes `ClosureCommand` obsolete too.

dktapps:
> Optimizing overload selection shouldn't be a blocker for this PR. Plugins that are particularly concerned about huge number of overloads could always fall back to RawParameter to leverage their own optimized parsing algorithm.

Yeah, but then they'd lose client-side command hints, which we all know is the thing plugin devs _really_ care about, for some reason...

WenoxGB:
any plan to add `PlayerArgument` ?

Adds a `tick()` method to populate `static::$VALUES` with online player names,
quoted if they contain spaces. This improves compatibility with command input
parsing and tab-completion for string-based arguments.

PlayerArgument resolves actual Player instances, not raw name strings. This
approach allows dynamic, formatted suggestions without requiring resolution,
making it suitable for custom argument types that expect player names as strings.

Implementation:
- Iterates over `Server::getInstance()->getOnlinePlayers()`
- Quotes names with spaces: "Mr Hacker"
- Leaves simple names untouched: Steve, Alex
- Uses `array_combine()` to build a key-value map for suggestion use

dktapps:
I don't know about the tick stuff, that doesn't seem necessary to me, but yes there are plans to expand the system.

I'm currently unsure how to deal with target args because of the `.self` vs `.other` permissions for various commands.

Enums, soft (dynamic) enums, and composite types like Vector3 are also not supported yet.

This PR implements the minimum of what was needed to get PM's builtin commands moved over to the new system.

src/command/defaults/BanCommand.php
```diff
@@ -25,42 +25,38 @@
...
 	}
 
-	public function execute(CommandSender $sender, string $commandLabel, array $args){
-		if(count($args) === 0){
-			throw new InvalidCommandSyntaxException();
-		}
-
-		$name = array_shift($args);
-		$reason = implode(" ", $args);
+	private static function execute(CommandSender $sender, string $playerName, string $reason) : void{
```
john-bv: I know this is legacy BC compat, but is there a way we can implement a player argument here now? It shouldn't take much more effort.

src/command/defaults/ClearCommand.php
```diff
@@ -37,42 +41,36 @@
...
 
 	public function __construct(string $namespace, string $name){
 		parent::__construct(
 			$namespace,
 			$name,
-			KnownTranslationFactory::pocketmine_command_clear_description(),
-			KnownTranslationFactory::pocketmine_command_clear_usage()
+			[new CommandOverload([
+				new StringParameter("playerName", "player name"),
+				new MappedParameter("targetItem", "item name", static function(string $v) : Item{
```
john-bv, line 61: How did i not know about this lol

src/command/defaults/ClearCommand.php
```diff
@@ -37,42 +41,36 @@
...
+	private const OVERLOAD_PERMS = [self::SELF_PERM, self::OTHER_PERM];
 
 	public function __construct(string $namespace, string $name){
 		parent::__construct(
 			$namespace,
 			$name,
-			KnownTranslationFactory::pocketmine_command_clear_description(),
-			KnownTranslationFactory::pocketmine_command_clear_usage()
+			[new CommandOverload([
+				new StringParameter("playerName", "player name"),
```
john-bv: Same as above, we could use that player argument in cases like here.

src/command/overload/ExecutorOverload.php
```diff
@@ -0,0 +1,252 @@
...
+		return $this->requiredInputCount;
+	}
+
+	/**
+	 * @return string[]
+	 * @phpstan-return list<string>
+	 */
+	public function getPermissions() : array{ return $this->permissions; }
+
+	public function senderHasAnyPermissions(CommandSender $sender) : bool{
```
john-bv, line 158: I feel like this is verbose, maybe just `senderHasPermission` would suffice?

src/command/overload/Parameter.php
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	public function getCodeType() : BaseType{
+		return $this->codeType;
+	}
+
+	/**
+	 * Returns whether this command will consume all remaining inputs.
+	 */
+	public function consumesAllRemainingInputs() : bool{
```
john-bv, line 55: This feels like a underlying implementation detail being exposed to the caller. Is there a way we can build this into the parameters without exposing the details? If you want this to be accessible, maybe renaming this to `isLast` might make sense, but in my head I immediately think of something like a new overload named `FinalParameter`, that way during usage it's clear to the caller what they're doing and whats functionally going to happen rather than having two conflicting parameters that consume all inputs. Would be curious to hear your thoughts.

src/network/mcpe/NetworkSession.php
```diff
@@ -1113,18 +1122,57 @@ public function syncAvailableCommands() : void{
 			//use filtered aliases for command name discovery - this allows /help to still be shown as /pocketmine:help
 			//on the client without conflicting with the client's built-in /help command
 			$lname = strtolower($firstNetworkAlias);
-			$aliasObj = new CommandEnum(ucfirst($firstNetworkAlias) . "Aliases", $aliases);
+			$aliasObj = count($aliases) > 1 ? new CommandEnum(ucfirst($firstNetworkAlias) . "Aliases", $aliases) : null;
+
+			$overloads = [];
+			foreach($command->getOverloads() as $overload){
+				//TODO: we should only send permissible overloads here
```
john-bv: Is there a reason this was left as todo? Does it cause problems with the client  or something?

src/network/mcpe/NetworkSession.php
```diff
@@ -1113,18 +1122,57 @@ public function syncAvailableCommands() : void{
...
+							$parameters[] = CommandParameter::standard($translated, $simpleArgType, optional: $k >= $required);
+						}
+					}
+				}
+				$overloads[] = new CommandOverload(chaining: false, parameters: $parameters);
+			}
 
 			$description = $command->getDescription();
 			$data = new CommandData(
 				$lname, //TODO: commands containing uppercase letters in the name crash 1.9.0 client
```
john-bv, line 1206: can't we just `strtolower` here? I don't think any vanilla commands allow caps anyway unless i've been out of it that long lol

src/command/Command.php
```diff
@@ -189,4 +194,31 @@ public static function broadcastCommandMessage(CommandSender $source, Translatab
 			}
 		}
 	}
+
+	protected static function fetchPermittedPlayerTarget(
```
john-bv: Are we really only concerned about `Player` on `Target`? It might make sense for a plugin dev to target entities as well.

john-bv (commented):
Just wanted to leave some of my thoughts on here :)

src/command/overload/ExecutorOverload.php
dktapps: Overloads can have multiple permissions, I felt it should be clear whether it needs one or all of them

src/network/mcpe/NetworkSession.php
dktapps: Never got around to it

src/network/mcpe/NetworkSession.php
dktapps: that's what it's doing

src/command/Command.php
dktapps: It's migrated old code, none of the existing commands support entity targets. The way targeting works needs to be reworked anyway but that's beyond the scope of this PR

src/command/overload/Parameter.php
dktapps: I don't really see what difference it makes. It's here so that `CommandOverload` can verify there are no parameters after any that consume all inputs. I don't think a `FinalParameter` really changes anything other than the way in which it gets checked (`$param->consumesAllRemainingInputs()` vs `$param instanceof FinalParameter`).

john-bv: I feel like abstracting this out at the call site is the most optimal approach for readability. I agree functionally it makes no difference, but for the sake of readability, it's a lot easier to assume what `FinalParameter` is doing when you call it than it is to guess if a command has `consumesAllRemainingInputs(): true`, definitely a matter of preference though.

SOF3: Don't abuse inheritance please. An explicit check method always has better clarity than an `instanceof` check.

src/network/mcpe/NetworkSession.php
```diff
@@ -1113,18 +1122,57 @@ public function syncAvailableCommands() : void{
...
-			$aliasObj = new CommandEnum(ucfirst($firstNetworkAlias) . "Aliases", $aliases);
+			$aliasObj = count($aliases) > 1 ? new CommandEnum(ucfirst($firstNetworkAlias) . "Aliases", $aliases) : null;
+
+			$overloads = [];
+			foreach($command->getOverloads() as $overload){
+				//TODO: we should only send permissible overloads here
+				$parameters = [];
+				$required = $overload->getRequiredParameterCount();
+				foreach($overload->getParameters() as $k => $parameter){
+					if(is_string($parameter)){
```
inxomnyaa, line 1168: Not sure why this is an extra branch, couldnt this be merged into the match below as StringParameter?

src/network/mcpe/NetworkSession.php
```diff
@@ -1113,18 +1122,57 @@ public function syncAvailableCommands() : void{
...
+							default => CommandParameterTypes::ID //string
+						};
+						$suffix = $parameter->getSuffix();
+						$name = $parameter->getPrintableName();
+						$translated = $name instanceof Translatable ? $language->translate($name) : $name;
+						if($suffix !== ""){
+							//umm... client only allows suffixes on integer params???
+							//TODO: as of 1.21.111, the client crashes if we try to provide actual suffixes for /xp.
+							//The cause for this seems to be undefined behaviour on the client linked to the order and
+							//number of certain hardcoded enums like GameMode, IntParam etc which we don't provided. I
```
inxomnyaa: nitpick, *provide

inxomnyaa (commented):
Over all, complex, but i can't say i hate it. Good job

src/network/mcpe/NetworkSession.php
dktapps: No, this needs to be sent as an enum so that the client can render it like `/effect <player> clear` rather than `/effect <player> <string>`

dktapps:
I'm considering possible options to introduce this API in PM5 to let people adopt it early, but it's not too clear how it would work. Current thinking is to introduce `OverloadedCommand` which implements the new features of this PR over `Command`, and then scrap the legacy API in PM6.

However, a PM6 scrapping would then mean renaming `OverloadedCommand` to `Command` and/or merging their functionality, which would then be an API break for early adopters anyway. Preferably I'd like a way of doing it that allows early adopters to get to PM6 with no effort. That way we're not punishing early adopters.

Or maybe I'm thinking too hard about this, since changing `OverloadedCommand` references to `Command` would be much less work if your code is already upgraded as opposed to having to do this all at once at PM6...

remminiscent:
I agree, this approach will let us get used to the new command system early in PM5 and move to PM6 without a major difference.

dktapps:
Demo of how LuckPerms command tree might be implemented using the API in this PR as of e762b12: https://gist.github.com/dktapps/57a0a2ba161592d0e18577c4e93f6264

remminiscent:
in my opinion, a Target parameter is necessary for the initial release of this

dktapps:
Yeah, target and enum are on my todo list. I've been more focused on initial system architecture so far though.

dktapps:
Overview of changes since last reviews:

- Overload registration is now substantially changed.
  - Each command now has only one root overload, which may be either an `ExecutorOverload` (which accepts a callback) or a `BranchingOverload` (delegates to sub-overloads).
  - Simple commands with one callback can just pass a `new ExecutorOverload` directly to `Command::__construct()`.
  - Overloaded commands (commands with multiple callbacks) must now use `BranchingOverloadBuilder` to create an overload tree.
    - A branching overload must have at least two child overloads. These children can be added to `BranchingOverloadBuilder` easily using `->executor()` or `->branch()`. (Constructing overloads directly is not allowed for registering nested overloads, because they need to know about all of their parent parameters to check the provided callback.)
    - Branching overloads can have nested branches, allowing nested subcommands and similar patterns.
    - If the first unique argument of a nested overload is a literal (i.e. a subcommand name), the branching overload will put it in a map, allowing fast lookups without bruteforce searches. This eliminates potential performance bottlenecks for common subcommand design patterns.
    - A branching overload may declare common leading parameters, which all of its child overloads will accept before their own unique parameters.
      - For example, `/effect` uses a branching overload which accepts `player` as a common arg, and then has two executor overloads, one of which accepts `clear`, and the other accepts args to apply an effect.
      - This avoids unnecessary leading parameter info duplication.
      - This is also better for performance, as leading arguments can be parsed once for many overloads which belong to the same branch. Previously they would be reparsed multiple times.
- Usage message generation on invalid command syntax now shows only the most likely usages, rather than showing all of them. For example, typing `/time set` without the time parameter will now only show `/time set <time name>` and `/time set <timestamp>`, and won't show the usage for `/time add` et al.
- Implemented `RelativeXYZ` parameters (3x relative floats).
- Stripped out legacy `commands.*.usage` and `pocketmine.command.*.usage` translation keys.

dktapps:
Remaining items of interest:

- [ ] Targeting parameters. These are a headache because of their interactions with permissions.
  - All PM commands currently declare different permissions for targeting `self` vs `other` targets. Since `Parameter`s currently have no context, they don't have a way to check whether a target is permitted or not.
  - Technically we can just do a `MappedParameter<Player>` and leave the permission checking up to the handler, but this feels like an easy way to screw up, and the necessary permissions wouldn't be visible in overload metadata.
- [ ] Enums. I'm not sure exactly how I want to implement these.
  - The easiest thing would be to just pass the string directly to the handler, but I don't like that much.
  - Could also run it through a `MappedParameter`
  - Really it'd be nice if we could just do something like `new EnumParameter(GameMode::class)` and have all the aliases and mapping stuff automatically handled. However we don't want to force people to declare a new `enum Whatever` for every command parameter.
  - Not all command enums will correspond to code enums also. Consider stuff like item names, which will be a fixed list, but are registered dynamically in `StringToItemParser` rather than being declared as an enum.
- [ ] Usage messages going only to the next branch, rather than spamming the console with multiple overloads for a subcommand we might not want to use.
  - In the LuckPerms demo above, typing `lp` will generate 5 usage messages starting with `lp verbose`. Ideally we'd just show something like `lp verbose ...` instead, and then show the more specific messages only when the user actually types `lp verbose`. This would avoid unnecessary console spam.
  - I tried to implement this already, but it was a headache. Will revisit it.
- [ ] Possibly, we might want the ability to do optional literals before required parameters. This was an item of interest for `cmdalias`, where in practice the `global` token could be an optional literal. However I haven't figured out what the rules for this would need to be to keep the system sane and predictable.
- [ ] Translation keys for parameter display names. Vanilla doesn't do this, but PM has had translated usage messages for donkey's years, so it'd be kinda lame if we didn't still support that after introducing this system.
- [x] Builtin commands to just construct `Command` directly, rather than extending it, since extending isn't needed anymore
- [ ] A way to potentially implement this system in PM5 without BC breaks.
  - Currently, the only things that `Command` actually directly tracks are namespace, name, root overload, custom permission message, and root description.
    - Namespace, name and root overload could be tracked directly by `CommandMap`
    - I'm thinking about allowing `ExecutorOverload` to have individual descriptions. This would be useful for subcommands to have different usage messages.
    - There's also an argument to be made for moving custom permission messages to overloads, or possibly getting rid of them.
  - Essentially, it's looking like `Command` might become completely obsolete, which would allow us to implement this system in PM5 in theory, while leaving legacy systems untouched.
- [ ] A way to get both the mapped value and original input for `MappedParameter`. This is needed for stuff like `/give`, where we want both the `Item` instance (to give to the player) and the alias used to give it (for the audit messages).
- [ ] Handlers really shouldn't be throwing `InvalidCommandSyntaxException`. Explore what's still doing this and why.

alvin0319:
For enums, perhaps we could just pass closures that returns a list of string into the last parameter of MappedParameter?
For enum classes, add helpers for converting enums into list of enums for stuff like GameMode and etc.

pseudo code be like
```php
BranchingOverloadBuilder::make(commonParameters: [
	new StringParameter("target", "target")
])
	->executor(
		["clear"],
		//TODO: our permission system isn't granular enough for this right now - the permission required
		//differs not by the usage, but by the target selected
		self::OVERLOAD_PERMS, self::removeEffect(...),
	)
	->executor([
		new MappedParameter("effect", "effect name", static fn(string $v) : Effect =>
			StringToEffectParser::getInstance()->parse($v) ??
				throw new ParameterParseException("Invalid effect name")
		, static fn() => StringToEffectParser::getInstance()->getKnownAliases()),
		new IntRangeParameter("duration", "duration", 0, (int) (Limits::INT32_MAX / 20)),
		new IntRangeParameter("amplifier", "amplifier", 0, 255),
		new BoolParameter("bubbles", "bubbles")

		//TODO: our permission system isn't granular enough for this right now - the permission required
		//differs not by the usage, but by the target selected
	], self::OVERLOAD_PERMS, self::modifyEffect(...))
->build()
```

in MappedParameter

```php
/**
 * @tempate T of \UnitEnum
 * @phpstan-param class-string<T>
 */
public static function fromEnum(
	string $codeName,
	Translatable|string $printableName,
	string $enumClass
) : MappedParameter{
	$cases = $enumClass::cases();
	return new MappedParameter($codeName, $printableName, static function(string $v) use ($cases){
		foreach($cases as $case){
			if(strtolower($case->name) === strtolower($v)){
				return $case;
			}
		}
		throw new ParameterParseException("Invalid enum name");
	}, static fn() => array_map(static fn(\UnitEnum $enum) => strtolower($enum->name), $cases));
}

**Comment by dktapps** [comment]
Would prefer to bind a mapping for it. But also ideally I'd like to implement it in such a way that cases can't suddenly become invalid later on.

There's also the matter of enum constraints to consider, although what we'd use those for I'm not very clear yet.

**Comment by NTT1906** [comment]
I haven't read the code that much so I'm sorry if I'm missing something. 3 years ago, I written a virion framework [Ovommand](https://github.com/GalaxyGamesMC/Ovommand/tree/dev) that tried to do the same and recently, I came across this repo [Blockception/BC-Minecraft-Bedrock-Command](https://github.com/Blockception/BC-Minecraft-Bedrock-Command) that may not help you much :L

I'd like to address something along my old experients, tho... these are what I came up limited by virion scope so it may not suit this well:
### 1. Position
- Position parameter has expandable size parameter in Bedrock. The parameter has maximum of three spans but they can be written as one or two spans.
eg: `123 456 789`, `~~~`, `~ ~ ~`, `~ ~-1.2~` are all valid inputs.
My solution to this was to try to read until you found three valid param; otherwise, return an invalid syntax. Another solution may be to split this at [the start of the parsing](https://github.com/pmmp/PocketMine-MP/blob/23904f05d7625befe478d31531abd2d6bebe4017/src/command/utils/CommandStringHelper.php#L98) so each notation will be its own param.
- The position parameter can be **local**. This type of param will begin with caret notation '^' instead of tilde notation, and it cannot be combined with relative position. Previously, I wrote [CoordinateResult](https://github.com/GalaxyGamesMC/Ovommand/blob/dev/src/galaxygames/ovommand/parameter/result/CoordinateResult.php) to retain the position value and any notation with it. Then, call CoordinateResult::asPosition() to get the exact position depending on the provided entity input.
### 2. Enums
The way I implemented it was having [enum](https://github.com/GalaxyGamesMC/Ovommand/blob/dev/src/galaxygames/ovommand/enum/BaseEnum.php) being a dictionary where upon parsing will return the value from the key if there is any.
> Really it'd be nice if we could just do something like new EnumParameter(GameMode::class) and have all the aliases and mapping stuff automatically handled.

I just create a native enum [DefaultEnums](https://github.com/GalaxyGamesMC/Ovommand/blob/dev/src/galaxygames/ovommand/enum/DefaultEnums.php) and map out all of those...
```php
enum DefaultEnums : string{
	case BOOLEAN = "Boolean";
	case VANILLA_GAMEMODE = "GameMode";
	case PM_GAMEMODE = "PMGameMode";
	case ONLINE_PLAYERS = "OnlinePlayers";

	public function encode() : IDynamicEnum | IStaticEnum{
		return match ($this) {
			self::BOOLEAN => new HardEnum($this->value, ["true" => true, "false" => false], isProtected: true, isDefault: true),
			self::PM_GAMEMODE => new HardEnum($this->value,
				["survival" => GameMode::SURVIVAL, "creative" => GameMode::CREATIVE, "adventure" => GameMode::ADVENTURE, "spectator" => GameMode::SPECTATOR],
				visibleAliases: ["survival" => "s", "creative" => "c", "adventure" => "a", "spectator" => "v"],
				hiddenAliases: ["survival" => "0", "creative" => "1", "adventure" => "2", "spectator" => "6"],
				isProtected: true, isDefault: true
			),
			self::VANILLA_GAMEMODE => new HardEnum($this->value,
				["survival" => GameMode::SURVIVAL, "creative" => GameMode::CREATIVE, "adventure" => GameMode::ADVENTURE, "spectator" => GameMode::SPECTATOR],
				["survival" => "s", "creative" => "c", "adventure" => "a"],
				isProtected: true, isDefault: true
			),
			self::ONLINE_PLAYERS => new SoftEnum("OnlinePlayers", isProtected: true, isDefault: true)
		};
	}
}
```
where it maps some handy enums such as DefaultEnums::BOOLEAN returns `true/false` based on input string or DefaultEnums::GAMEMODE for gamemode. For those soft enum suchs as online players, we will [update](https://github.com/GalaxyGamesMC/Ovommand/blob/cb91eebbfe27518c98eeca7d3af3ed536ef4ee89/src/galaxygames/ovommand/OvommandHook.php#L67) it on player join and quit event.

So in my command setup, I can register the enum parameter with this [code](https://github.com/idumpster/OvoTest/blob/1e342737a4d5526bfc4f1bc85fce376d5348d75a/src/arie/test/TestSubCommand.php#L38C3-L41C5)
```php
$this->registerParameters(
	new PositionParameter("a"),
	new EnumParameter("aa", DefaultEnums::PM_GAMEMODE),
);
```

> However we don't want to force people to declare a new enum Whatever for every command parameter.

They can save those newly created classes and reuse them :> jk. As two enum parameters can have different names, I think that new may be needed. I might misunderstand your words here; you might mean the same thing as returning a new HardEnum when the DefaultEnum gets encoded.
> Would prefer to bind a mapping for it. But also ideally I'd like to implement it in such a way that cases can't suddenly become invalid later on.

Were those cases where they become invalid due to the item being dynamically removed? Anyway, we can use soft enums instead of hard enums for those... but we need to [update](https://github.com/GalaxyGamesMC/Ovommand/blob/cb91eebbfe27518c98eeca7d3af3ed536ef4ee89/src/galaxygames/ovommand/enum/SoftEnum.php#L72) the new changes via UpdateSoftEnumPacket every time the list gets changed.

>  Possibly, we might want the ability to do optional literals before required parameters. This was an item of interest for cmdalias, where in practice the global token could be an optional literal. However I haven't figured out what the rules for this would need to be to keep the system sane and predictable.

Can you explain this more? I may not understand this… Why not just create two overloads, where one has the optional literal and the other doesn't?

Since literals are string enums? I also noticed these kind of enums are kind of buggy? There are cases where the hints break in when you try the string enum with other datatypes.

dktapps:
> Why not just create two overloads, where one has the optional literal and the other doesn't?

That's what I did, but it's double the number of usages



For the enums thing, as I said, I don't want to force people (or be forced myself) to define a new enum class every time they want to add a new set of options to a command. It should be possible to use an enum if desired, but not required.

> Anyway, we can use soft enums instead of hard enums for those

Alvin's suggestion for binding StringToItemParser to an enum would have cases magically vanish from existence if the parser mappings were modified at runtime, and the command system wouldn't have any way to know about it. We don't want to use soft enums for everything because it'll complicate things for users who just want a fixed set of options. So the enums need to capture their valid value set and ensure that the mapper accepts every value specified. Not exactly sure how that's going to work currently.

> Since literals are string enums? I also noticed these kind of enums are kind of buggy? There are cases where the hints break in when you try the string enum with other datatypes.

Literals are not enums in this system, not on the server side anyway. They may be sent as enums over the network, but that's kinda irrelevant.

My priority here is building a system for plugin devs to use. The client side command suggestions are a happy bonus, but if they get in the way I'm not gonna cry about breaking them. Already had this headache with --flags in the style that LuckPerms uses. Plugin use cases are far more complex than anything vanilla does.

dktapps:
To be clear, the holdup isn't a lack of solutions, but that decisions haven't been made in affected areas

Coordinate params were already implemented, those were quite trivial. I didn't initially implement them because they weren't necessary and I needed to refactor some code to avoid duplication. But that's dealt with now.

Though, I wasn't aware that the inputs can be combined with no whitespace (which seems kinda sus to me, that's probably a bug in Bugrock), but if I wanted to support that here it'd just be a case of deleting some zero-whitespace checks in `RelativeXYZParameter::parse()`.
I haven't done `^` yet either because it hasn't been needed to support PM commands yet and I'm unfamiliar with how it works. That can come later on. The goal of this PR is to implement the framework for the new system and make sure it's possible to do everything we want. The smaller stuff can come later on.

kostamax27:
Are there any plans to improve `CommandEvent` or to implement a new event?

dktapps:
Not currently

PR #6838, opened 2025-10-12
Implement Cactus Flower

Azvyl:
This PR adds the [Cactus Flower](https://minecraft.wiki/w/Cactus_Flower) block (`minecraft:cactus_flower`) to PocketMine-MP

1.21.70 - Spring to Life

### API changes
- Added `BlockTypeIds::CACTUS_FLOWER`
- Added `CactusFlower` class
- Added `VanillaBlocks::CACTUS_FLOWER()`
- Added `cactus_flower` to `StringToItemParser`

### Behavioural changes
The top cactus will be used for growing

## Tests
https://youtu.be/rRXb1og43FA


dktapps:
Looks interesting, might take me a bit to review this one though because I've never heard of this block

Wraith0x10:
> Looks interesting, might take me a bit to review this one though because I've never heard of this block

Yep, same here. From a quick bit of research, it looks like this PR does align with what’s described on the Minecraft Wiki.

src/block/Cactus.php
```diff
@@ -78,26 +79,54 @@ public function ticksRandomly() : bool{
...
 				}
-				$this->age = 0;
-				$world->setBlock($this->position, $this, update: false);
-			}else{
-				++$this->age;
-				$world->setBlock($this->position, $this, update: false);
 			}
 		}
+
+		if($this->age > self::MAX_AGE){
```
dktapps: It'd be preferable if we make sure this doesn't exceed the bounds. We might use integer range bounds on this in the future.

src/block/Cactus.php
```diff
@@ -78,26 +79,54 @@ public function ticksRandomly() : bool{
...
+
+		$world = $this->position->getWorld();
+
+		if(!$world->isInWorld($up->position->x, $up->position->y, $up->position->z)){
+			return;
+		}
+
+		$height = 1;
+		$current = $this;
+		while($height < 3 && ($down = $current->getSide(Facing::DOWN))->hasSameTypeId($this)){
```
dktapps: The 3 should be turned into a constant since it now appears in multiple places

src/block/Cactus.php
```diff
@@ -78,26 +79,54 @@ public function ticksRandomly() : bool{
...
+			foreach(Facing::HORIZONTAL as $side){
+				if($up->getSide($side)->isSolid()){
+					$canGrowFlower = false;
+					break;
+				}
+			}
+
+			if($canGrowFlower){
+				$chance = $height >= 3 ? 0.25 : 0.10;
+				if((mt_rand(1, 100) / 100) <= $chance){
```
dktapps: If you use `25` and `10` for the chance you can avoid the division here

src/block/Cactus.php
```diff
@@ -78,26 +79,54 @@ public function ticksRandomly() : bool{
...
+		$height = 1;
+		$current = $this;
+		while($height < 3 && ($down = $current->getSide(Facing::DOWN))->hasSameTypeId($this)){
+			$current = $down;
+			$height++;
+		}
+
+		$this->age++;
+
+		if($this->age === 9){
```
dktapps, line 99: Why is this restricted to age 9?

Azvyl: > A cactus also has a chance to grow a cactus flower at any height. This happens after a cactus block has received exactly 9 random ticks (i.e. earlier than growing another cactus block)...

The number 9 is used according to wiki so that the cactus has the first chance to grow flowers, before it tries to grow itself taller

src/block/Cactus.php
Azvyl: Should we use `>=` for this?

Azvyl:
I chose `MAX_HEIGHT` as the const name for readability, but I can change it to `MAX_GROWTH_HEIGHT` for clarity if needed

src/block/Cactus.php
dktapps: I didn't see any mention of age 9 in the wiki?

src/block/Cactus.php
dktapps: I meant that `$this->age` should never contain a value bigger than 15. Ideally it should be incremented only after we've checked it's not already max age.

dktapps:
> I chose `MAX_HEIGHT` as the const name for readability, but I can change it to `MAX_GROWTH_HEIGHT` for clarity if needed

Seems fine to me, I just wanted to make sure we don't repeat the same information in a bunch of places

src/block/Cactus.php
Azvyl: It's in the [Cactus Wiki](https://minecraft.wiki/w/Cactus#:~:text=A%20cactus%20also%20has%20a%20chance%20to%20grow%20a%20cactus%20flower%20at%20any%20height.%20This%20happens%20after%20a%20cactus%20block%20has%20received%20exactly%209%20random%20ticks%20(i.e.%20earlier%20than%20growing%20another%20cactus%20block))

Or maybe I misunderstood?

dktapps: Oh, my bad, I didn't see that

src/block/Cactus.php
```diff
@@ -78,26 +80,54 @@ public function ticksRandomly() : bool{
...
+
+		$world = $this->position->getWorld();
+
+		if(!$world->isInWorld($up->position->x, $up->position->y, $up->position->z)){
+			return;
+		}
+
+		$height = 1;
+		$current = $this;
+		while($height < self::MAX_HEIGHT && ($down = $current->getSide(Facing::DOWN))->hasSameTypeId($this)){
```
dktapps: This can use `$this->getSide(Facing::DOWN, $height)` to avoid leaking the `$current` variable into the outer scope (and generally simplify the code)

Azvyl: Reminds me of something, shouldn't we reduce the number of `getWorld()` calls since they say it can reduce performance?
maybe I'll change this to `$world->getBlockAt()`

dktapps: Don't worry about that. Reduce `getWorld()` calls where you've called it directly, but `getSide()` exists for a reason, so you might as well use it.

dktapps: You can do it that way if you want (the old code also used `getBlockAt()`), but I'm not particularly concerned if your version uses `getSide()`.
`$this->getSide(Facing::DOWN, $height)` is less boilerplaty than `$world->getBlockAt($this->position->x, $this->position->y - $height, $this->position->z)`. I leave it up to you.

dktapps (dismissed):
LGTM

ShockedPlot7560 (dismissed):
Looks good to me

ipad54 (commented):
According to the block properties table and the wiki, Cactus Flower is a flammable block: https://github.com/pmmp/BedrockData/blob/master/block_properties_table.json#L1346-L1347

ipad54 (approved):

dktapps:
Strange that wasn't picked up by the tests

Azvyl:
> Strange that wasn't picked up by the tests

It seems the test only checks hardness and blast resistance

dktapps (approved):

dktapps:
Thanks!

Issue #3722, opened 2020-07-23
Water Logging


Dapro718:
### Issue description

- Expected result: When you place the blocks that are supposed to get water logged in water, they get water logged.
- Actual result: Blocks do not get water logged when placed in water. Blocks do get water logged when you place a water bucket on them, but are not water logged in the eyes of other players & are in water logged when the server resets.

### Steps to reproduce the issue
1. ...
2. ...

### OS and versions

* PocketMine-MP: 3.14.2
* PHP: 7.3
* Server OS: Windows
* Game version: PE/Win10 (delete as appropriate)

### Plugins

- If you remove all plugins, does the issue still occur?
- If the issue is **not** reproducible without plugins:
  - Have you asked for help on our forums before creating an issue?
  - Can you provide sample, *minimal* reproducing code for the issue? If so, paste it in the bottom section

fuyutsuki:
It seems that detailed steps and information are needed to reproduce

dktapps:
It's easily reproducible. It's just not implemented, and implementing it will take a lot of work.

fuyutsuki:
Ah, the water disappears after placed it on blocks such as stairs. I was misunderstanding, sorry.

PR #6840, opened 2025-10-13
Implemented waterlogging

ipad54:
## Introduction
This PR implements support for waterlogging. For now I've marked this PR as "incomplete" only because I added waterlogging support for basic blocks (stairs, slabs, doors, levers, fences) to make sure everything matches vanilla and to not add a lot of noise to the diff. When the API is fully discussed and we make sure everything works fine, I'll make other remaining blocks to be waterloggable.

### Related issues & PRs
#3722

### API changes
Added `CoveredByWater` interface, as well as `CoveredByWaterTrait`. Blocks which implement this interface are automatically considered waterloggable.

As Bedrock allows flowing water to waterlog some blocks too, I've decided to keep Water block inside of Waterloggable one, so the API is similar to the plant blocks in decorated pot.

To change block's waterlogging state there is a `CoveredByWater->setWaterCover(?Water $waterCover) : self`.
To get the water cover there is a `CoveredByWater->getWaterCover() : ?Water`.

Example usages:
- `$block->setWaterCover(VanillaBlocks::WATER());` 
- `$block->setWaterCover(VanillaBlocks::WATER()->setDecay(5));`
- `$block->setWaterCover(null)`;

To check that the block is waterlogged you can do `$block->getWaterCover() !== null`.

### Internal API changes
On the Internal level there is a new `displaced block` abbreviation, which refers to a block, placed on the second layer.

The following internal public methods have been added, which should NOT be used by plugins:
- `World->delayDisplacedBlockUpdate(Vector3 $pos, int $delay) : void` Similar to `scheduleDelayedBlockUpdate()`, but for displaced blocks. Unfortunately, I had to add this to avoid collisions with blocks which use a regular schedule delayed updates apart from being waterloggable (e.x campfire, brewing stand for cooking). The name is probably quite odd, but I want to make sure no one accidentally uses this internal method instead of the regular one by IDE autocomplete.
- `Block->onDisplacedScheduledUpdate() : void` Similar to `onScheduledUpdate()`, but called for displaced blocks. 
- `Block->getDisplacedBlock() : Block` Returns AIR by default, overriden by waterloggable blocks to return their water state. Called in the core to retrieve a displaced block to send an update block packet, write that block to the second layer, run a separate schedule delayed update, update its position.
- `Block->setDisplacedBlock(Block $block) : void` called during `getBlockAt()` to retrieve a block from the second layer.
## Backwards compatibility
Plugins which interact with the second layer via internal methods ([like this one](https://github.com/GameParrot/DoubleSlabs)) will stop working because blocks, which are not expected to be in the second layer, will be overwritten by air.
## Follow-up
Implement snowlogging, it can be done very quickly after these changes.

## Tests
A small video can be found here: https://www.youtube.com/watch?v=_HQG7-bTJXg
From what can I see, it seems like it's fully matches vanilla behaviour, but I'm not 100% sure.

Note: in the end you might see that a lava can't collide with a non-waterlogged lever. I have fixed that bug after video recording.


src/block/Block.php
```diff
@@ -623,6 +652,10 @@ final public function getPosition() : Position{
 	final public function position(World $world, int $x, int $y, int $z) : void{
 		$this->position = new Position($x, $y, $z, $world);
 		$this->collisionBoxes = null;
+
+		if(($block = $this->getDisplacedBlock())->getTypeId() !== BlockTypeIds::AIR){
+			$block->position($world, $x, $y, $z);
```
dktapps, line 649: This seems a bit sketch. In the default case, this will allocate a new air block, position it, and then immediately throw it away. That doesn't seem good for performance.

src/block/Farmland.php
```diff
@@ -162,6 +163,8 @@ protected function canHydrate() : bool{
 		$startY = $this->position->getFloorY();
 		$startZ = $this->position->getFloorZ() - (int) (self::WATER_SEARCH_HORIZONTAL_LENGTH / 2);
 
+		$isWater = static fn(Block $block) : bool => $block instanceof Water || $block instanceof CoveredByWater && $block->getWaterCover() !== null;
```
dktapps: This logic seems to get repeated a lot. Is there a way to avoid it?

src/block/Flowable.php
```diff
@@ -41,7 +42,9 @@ public function isSolid() : bool{
 	}
 
 	public function canBePlacedAt(Block $blockReplace, Vector3 $clickVector, int $face, bool $isClickedBlock) : bool{
-		return (!$this->canBeFlowedInto() || !$blockReplace instanceof Liquid) &&
+		return
+			(($this instanceof CoveredByWater && $this->canBeCovered() && $blockReplace->getTypeId() === BlockTypeIds::WATER) ||
```
dktapps: `$this instanceof` is usually a code smell

dktapps (commented):
This was mostly discussed on Discord, but I'll leave some comments here too.

I haven't reviewed this fully yet so this is just a few nits.

src/block/Farmland.php
ipad54: Perhaps I could make a `WaterCoverHelper` class with `isWater(Block $block) : bool` and `getWater(Block $block) : ?Water`. What do you think?

dktapps: It'd be better than repeating it everywhere

src/block/Block.php
ipad54: Should probably be fine now, as `getDisplacedBlock()` now returns `null` in most cases.

src/block/Block.php
```diff
@@ -414,15 +414,15 @@ public function writeStateToWorld() : void{
...
+	public function getDisplacedBlock() : ?Block{
+		return null;
 	}
 
 	/**
 	 * @internal
 	 * Called when block is loaded to update its cover.
 	 */
-	public function setDisplacedBlock(Block $block) : void{
+	public function setDisplacedBlock(?Block $block) : void{
```
dktapps: Conditionally useless methods aren't great also, didn't notice this earlier

dktapps: We'd be able to constrain the accepted classes where applicable if this wasn't here also

src/block/utils/BlockDisplacer.php
```diff
@@ -0,0 +1,40 @@
...
+ * Used in things like waterlogging & snowlogging.
+ */
+interface BlockDisplacer{
+
+	/**
+	 * @internal
+	 * Called when a block is loaded to update its contained block.
+	 * Do NOT use this in plugins, use your block provided API instead.
+	 */
+	public function setDisplacedBlock(?Block $block) : void;
```
dktapps: I can't think of a reason why we'd want a generalised version of this, unless it was also checking the acceptance conditions. Since it should anyway require checking the type of displacer to know whether the displaced block can coexist with it.

src/item/Bucket.php
```diff
@@ -39,11 +40,17 @@ public function getMaxStackSize() : int{
 
 	public function onInteractBlock(Player $player, Block $blockReplace, Block $blockClicked, int $face, Vector3 $clickVector, array &$returnedItems) : ItemUseResult{
 		//TODO: move this to generic placement logic
-		if($blockClicked instanceof Liquid && $blockClicked->isSource()){
+
+		[$resultBlock, $liquid] = match(true){
+			$blockClicked instanceof CoveredByWater && ($cover = $blockClicked->getWaterCover()) !== null && $cover->isSource() => [(clone $blockClicked)->setWaterCover(null), $blockClicked->getWaterCover()],
+			$blockClicked instanceof Liquid && $blockClicked->isSource() => [VanillaBlocks::AIR(), $blockClicked],
+			default => [null, null]
```
dktapps: I don't feel like this match makes the code any easier to read

dktapps:
Sorry for not getting to this sooner, it's just quite a large PR and I'm trying to get the older backlogs cleared

src/world/World.php
```diff
@@ -1126,6 +1152,12 @@ public function createBlockUpdatePackets(array $blocks) : array{
 				UpdateBlockPacket::FLAG_NETWORK,
 				UpdateBlockPacket::DATA_LAYER_NORMAL
 			);
+			$packets[] = UpdateBlockPacket::create(
+				$blockPosition,
+				$blockTranslator->internalIdToNetworkId($fullBlock->getDisplacedBlock()?->getStateId() ?? Block::EMPTY_STATE_ID),
+				UpdateBlockPacket::FLAG_NETWORK,
+				UpdateBlockPacket::DATA_LAYER_LIQUID
+			);
```
dktapps: In most cases the second layer won't be present and this will be doubling bandwidth consumption for block updates without benefit

src/world/World.php
```diff
@@ -963,6 +978,17 @@ protected function actuallyDoTick(int $currentTick) : void{
...
+		while($this->scheduledDisplacedBlockUpdateQueue->count() > 0 && $this->scheduledDisplacedBlockUpdateQueue->current()["priority"] <= $currentTick){
+			/** @var Vector3 $vec */
+			$vec = $this->scheduledDisplacedBlockUpdateQueue->extract()["data"];
+			unset($this->scheduledDisplacedBlockUpdateQueueIndex[World::blockHash($vec->x, $vec->y, $vec->z)]);
+			if(!$this->isInLoadedTerrain($vec)){
+				continue;
+			}
+			$block = $this->getBlock($vec)->getDisplacedBlock();
+			$block?->onDisplacedScheduledUpdate();
+		}
```
dktapps, line 990: Not a fan of having a separate queue & mechanism for this. Any possibility we could just use the normal system and have both blocks updated together? Not sure if that might cause some unintended side effects though.

src/world/World.php
```diff
@@ -1977,14 +2027,20 @@ public function getBlockAt(int $x, int $y, int $z, bool $cached = true, bool $ad
 			$chunk = $this->chunks[$chunkHash] ?? null;
 			if($chunk !== null){
 				$block = $this->blockStateRegistry->fromStateId($chunk->getBlockStateId($x & Chunk::COORD_MASK, $y, $z & Chunk::COORD_MASK));
+				$displacedBlock = $this->blockStateRegistry->fromStateId($chunk->getDisplacedBlockStateId($x & Chunk::COORD_MASK, $y, $z & Chunk::COORD_MASK));
```
dktapps: This is going to hurt performance in the majority case, which will be air / second layer not allocated

src/world/Explosion.php
```diff
@@ -142,6 +143,9 @@ public function explodeA() : bool{
 								if($blastForce > 0){
 									if(!isset($this->affectedBlocks[World::blockHash($vBlockX, $vBlockY, $vBlockZ)])){
 										$_block = $this->world->getBlockAt($vBlockX, $vBlockY, $vBlockZ, true, false);
+										if(WaterCoverHelper::isWater($_block)){
+											continue;
+										}
```
dktapps: This seems kinda suspicious. Probably we should be taking `max()` of the blast resistances of the blocks on each layer.

src/world/format/SubChunk.php
```diff
@@ -83,6 +83,29 @@ public function setBlockStateId(int $x, int $y, int $z, int $block) : void{
...
+
+	/**
+	 * Sets the given internal block state ID to the second layer.
+	 */
+	public function setDisplacedBlockStateId(int $x, int $y, int $z, int $block) : void{
+		if(count($this->blockLayers) < 2){
+			if(count($this->blockLayers) === 0){
+				$this->blockLayers[] = new PalettedBlockArray($this->emptyBlockId);
+			}
+			$this->blockLayers[] = new PalettedBlockArray($this->emptyBlockId);
```
dktapps, line 108: If the `$block` is equal to `emptyBlockId` we can return here without doing anything (which will be the majority case)

src/world/World.php
ipad54: > Any possibility we could just use the normal system and have both blocks updated together? Not sure if that might cause some unintended side effects though.

Not really sure about alternatives. For example, a brewing stand is a waterloggable block, and if it currently brewing potions, it can cause water to flow faster

dktapps: Hm, true

kostamax27:
vanilla:

https://github.com/user-attachments/assets/299f45e9-108e-4753-b5e9-4c704f66f6b5

this rp:


https://github.com/user-attachments/assets/3eebed1b-efc7-4724-a1be-60199c94c99d

dktapps:
Once again Bedrock makes no sense...

src/item/LiquidBucket.php
```diff
@@ -55,18 +57,35 @@ public function getFuelResidue() : Item{
...
+		$coveredBlock = match(true){
+			$blockClicked instanceof CoveredByWater && $blockClicked->canBeCovered() => $blockClicked,
+			$blockReplace instanceof CoveredByWater && $blockReplace->canBeCovered() => $blockReplace,
+			default => null
+		};
+		if($coveredBlock !== null && ($this->liquid instanceof Water || $coveredBlock->canBeCoveredByFlowing())){
+			$targetBlock = $coveredBlock;
+			$resultBlock = $this->liquid instanceof Water ?
+				(clone $targetBlock)->setWaterCover((clone $this->liquid)->setStill(false)) :
+				clone $this->liquid;
```
dktapps: Isn't this going to cause partial-loggable blocks to get replaced by lava? This doesn't feel right, if we have any partial-loggable blocks that have canBeReplaced() = false in the future, this is going to cause unexpected behaviour

src/block/utils/CoveredByWater.php
```diff
@@ -0,0 +1,51 @@
...
+	public function liquidCollide(Block $cause, Block $result) : bool;
+
+	/**
+	 * Returns whether block can be covered by water at the current moment.
+	 * For example, double slabs can't be waterlogged, while single can.
+	 */
+	public function canBeCovered() : bool;
+
+	/** Returns whether the block can be covered not only by source water, but also by the flowing. */
+	public function canBeCoveredByFlowing() : bool;
```
dktapps: Maybe a BlockTypeTag would be better for this

dktapps: Also the naming of this feels awkward

src/block/utils/CoveredByWater.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\block\utils;
+
+use pocketmine\block\Block;
+use pocketmine\block\Water;
+
+interface CoveredByWater{
```
dktapps: Why do we call this `CoveredByWater` rather than `Waterloggable`?

src/item/LiquidBucket.php
dktapps: e.g. if a plugin made signs partial-loggable, they'd get replaced here by lava

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -499,13 +500,27 @@ private function handleUseItemTransaction(UseItemTransactionData $data) : bool{
...
 				$blockPos = $data->getBlockPosition();
 				$vBlockPos = new Vector3($blockPos->getX(), $blockPos->getY(), $blockPos->getZ());
-				$this->player->interactBlock($vBlockPos, $data->getFace(), $clickPos);
-				if($data->getClientInteractPrediction() === PredictedResult::SUCCESS){
-					//always sync this in case plugins caused a different result than the client expected
-					//we *could* try to enhance detection of plugin-altered behaviour, but this would require propagating
-					//more information up the stack. For now I think this is good enough.
-					//if only the client would tell us what blocks it thinks changed...
-					$this->syncBlocksNearby($vBlockPos, $data->getFace());
+				if($this->player->canInteract($vBlockPos->add(0.5, 0.5, 0.5), 15)){
```
dktapps: I actually meant to just do a distance check like the one we do in `syncBlocksNearby`. I don't remember what this function does, but it might be more complex than we need. Also, hardcoding a max distance of 15 might get in the way if plugins customise reach distance.

src/player/Player.php
```diff
@@ -1926,15 +1926,15 @@ public function breakBlock(Vector3 $pos) : bool{
 	 *
 	 * @return bool if it did something
 	 */
-	public function interactBlock(Vector3 $pos, int $face, Vector3 $clickOffset) : bool{
+	public function interactBlock(Vector3 $pos, int $face, Vector3 $clickOffset, bool $interactContainedBlock = false) : bool{
```
dktapps: This terminology isn't consistent with the `displaced` terminology used elsewhere, we should pick one and use it everywhere

src/world/World.php
```diff
@@ -1977,14 +2029,21 @@ public function getBlockAt(int $x, int $y, int $z, bool $cached = true, bool $ad
 			$chunk = $this->chunks[$chunkHash] ?? null;
 			if($chunk !== null){
 				$block = $this->blockStateRegistry->fromStateId($chunk->getBlockStateId($x & Chunk::COORD_MASK, $y, $z & Chunk::COORD_MASK));
+				$displacedBlock = $chunk->getDisplacedBlockStateId($x & Chunk::COORD_MASK, $y, $z & Chunk::COORD_MASK);
```
dktapps: This should be `displacedBlockStateId` to avoid type confusion

src/world/World.php
```diff
@@ -1977,14 +2029,21 @@ public function getBlockAt(int $x, int $y, int $z, bool $cached = true, bool $ad
...
+				$displacedBlock = Block::EMPTY_STATE_ID;
 			}
 		}else{
 			$block = VanillaBlocks::AIR();
+			$displacedBlock = Block::EMPTY_STATE_ID;
 		}
 
+		if($block instanceof CoveredByWater && $displacedBlock !== Block::EMPTY_STATE_ID){
+			$displacedBlock = $this->blockStateRegistry->fromStateId($displacedBlock);
+			$block->setWaterCover($displacedBlock instanceof Water ? $displacedBlock : null);
```
dktapps: Hmm. Maybe we need some more general way to deal with this, will need to think about this more

src/world/World.php
```diff
@@ -2239,7 +2298,7 @@ private function destroyBlockInternal(Block $target, Item $item, ?Player $player
 	 * @param bool        $playSound      Whether to play a block-place sound if the block was placed successfully.
 	 * @param Item[]      &$returnedItems Items to be added to the target's inventory (or dropped if the inventory is full)
 	 */
-	public function useItemOn(Vector3 $vector, Item &$item, int $face, ?Vector3 $clickVector = null, ?Player $player = null, bool $playSound = false, array &$returnedItems = []) : bool{
+	public function useItemOn(Vector3 $vector, Item &$item, int $face, ?Vector3 $clickVector = null, ?Player $player = null, bool $playSound = false, array &$returnedItems = [], bool $interactContainedBlock = false) : bool{
```
dktapps: This terminology is not consistent as mentioned previously

src/block/utils/CoveredByWater.php
ipad54: I named it this way initially for consistency (we don’t have many interfaces ending on `able`). I also feel that this name better tells the block’s ability to be covered by water rather than `Waterloggable`. I may be wrong though and maybe it’s better to change the interface name?

dktapps: I think most people are going to expect to see it called "waterlogging".

Covered by water feels a bit weird to me. Since the likes of e.g. stairs aren't really covered by water, but their cells are partially occupied by it.

I'm open to persuasion on this. I just fel like this name felt clunky

ipad54: Alright, I’ll rename this to Waterloggable. I can’t come up with a good reason why this should stay as CoveredByWater, especially when most people, as you mentioned, are expected to see it matching the mechanic name itself.

ipad54:
> vanilla:
> 
>  vanilla.mp4 
> this rp:
> 
>  pr.mp4

Fixed, thank you!

src/item/LiquidBucket.php
ipad54: I've changed the logic, but it's still remains incorrect. Vanilla for some reason allows some non-source waterloggable blocks to be replaced by lava here. Yeah, Bedrock again makes no sense, and I don't know if we want to entirely fix it (we would probably have to add something like `Block->canBeReplacedBy(Block $block) : bool`).

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -498,13 +499,27 @@ private function handleUseItemTransaction(UseItemTransactionData $data) : bool{
...
-				$this->player->interactBlock($vBlockPos, $data->getFace(), $clickPos);
-				if($data->getClientInteractPrediction() === PredictedResult::SUCCESS){
-					//always sync this in case plugins caused a different result than the client expected
-					//we *could* try to enhance detection of plugin-altered behaviour, but this would require propagating
-					//more information up the stack. For now I think this is good enough.
-					//if only the client would tell us what blocks it thinks changed...
-					$this->syncBlocksNearby($vBlockPos, $data->getFace());
+				if($vBlockPos->distanceSquared($this->player->getLocation()) < 10000){
+					$block = $this->player->getWorld()->getBlock($vBlockPos);
+					$blockTranslator = TypeConverter::getInstance()->getBlockTranslator();
```
Azvyl: We should be able to use `NetworkSession.getTypeConverter()` here

src/world/World.php
```diff
@@ -1977,14 +2029,21 @@ public function getBlockAt(int $x, int $y, int $z, bool $cached = true, bool $ad
...
 			}
 		}else{
 			$block = VanillaBlocks::AIR();
+			$displacedBlockStateId = Block::EMPTY_STATE_ID;
 		}
 
+		if($block instanceof Waterloggable && $displacedBlockStateId !== Block::EMPTY_STATE_ID){
+			$displacedBlockStateId = $this->blockStateRegistry->fromStateId($displacedBlockStateId);
+			$block->setContainedWater($displacedBlockStateId instanceof Water ? $displacedBlockStateId : null);
+		}
```
dktapps: I'm thinking we could probably use `readStateFromWorld()` to do this, instead of baking the logic into `getBlock` directly

src/world/World.php
```diff
@@ -2272,6 +2331,8 @@ public function useItemOn(Vector3 $vector, Item &$item, int $face, ?Vector3 $cli
 			if($player->isSneakPressed()){
 				$ev->setUseItem(false);
 				$ev->setUseBlock($item->isNull()); //opening doors is still possible when sneaking if using an empty hand
+			}else{
+				$ev->setUseBlock(!$interactDisplacedBlock);
```
dktapps: This feels dodgy. Shouldn't this be calling the `onInteract()` of the displaced block based on this flag? As opposed to disabling the interaction completely

src/world/World.php
```diff
@@ -1126,6 +1152,14 @@ public function createBlockUpdatePackets(array $blocks) : array{
 				UpdateBlockPacket::FLAG_NETWORK,
 				UpdateBlockPacket::DATA_LAYER_NORMAL
 			);
+			if($fullBlock->getTypeId() === BlockTypeIds::AIR || $fullBlock instanceof Waterloggable){
```
dktapps, line 1154: Why is this checking for air?

ipad54: This is needed to sync the second layer if a waterlogged block was destroyed by plugins, explosions, etc.

dktapps: Hmm, I feel like this isn't reliable. The block could've been directly replaced by something else

ipad54: I can’t come up with any other reliable ways to save the bandwidth, as we always need to keep the client in sync with the current blocks anyway.

kostamax27:
1. you forgot `Trapdoor`
2. `shift + right-click` with `Water Bucket` doesn't work

ipad54:
> 1. you forgot `Trapdoor`
> 2. `shift + right-click` with `Water Bucket` doesn't work

I just haven’t made trapdoor to be waterloggable yet. I added waterlogging functionality just to few basic blocks to not add a bunch of noise to the diff and make reviewing waterlogging logic easier, this is why it’s marked as incomplete. When the final API is fully discussed, I’ll make all the remaining blocks waterloggable.

PR #6843, opened 2025-10-17
Implement natural blocks introduced in the Nether Update

Azvyl:
This PR adds several previously unimplemented natural blocks from the Nether Update:
- Crimson Fungus [`minecraft:crimson_fungus`](https://minecraft.wiki/w/Crimson_Fungus)
- Warped Fungus [`minecraft:warped_fungus`](https://minecraft.wiki/w/Warped_Fungus)
- Nether Sprouts [`minecraft:nether_sprouts`](https://minecraft.wiki/w/Nether_Sprouts)
- Crimson Nylium [`minecraft:crimson_nylium`](https://minecraft.wiki/w/Crimson_Nylium)
- Warped Nylium [`minecraft:warped_nylium`](https://minecraft.wiki/w/Warped_Nylium)

This PR does not implement:
- Bone Meal particles when using it on Nylium, Netherrack, and nether fungus

1.16.0 - Nether Update

### API changes
- Add new type ids to `BlockTypeIds`
- Add `BlockTypeTags::NYLIUM`, `BlockTypeTags::HUGE_FUNGUS_REPLACEABLE`
- Add `NetherFungus`, `NetherSprouts`, `Nylium`, `NetherTree` classes
- Register the new block to `StringToItemParser`
- Add new `TreeType`: `CRIMSON`, `WARPED`
- Add the new `TreeType` to `TreeFactory`

### Behavioural changes
- NetherRoots and mushrooms get Nylium support
- Using Bone Meal on Netherrack placed next to Nylium will turn the Netherrack into Nylium

## Tests
https://youtu.be/OAmaJGrlAYM


src/world/generator/object/NetherTree.php
```diff
@@ -0,0 +1,196 @@
...
+			if($random->nextBoundedInt(7) === 0){
+				$i *= 2;
+			}
+
+			$maxAge = NetherVines::MAX_AGE;
+			$startAge = 23;
+
+			for($v = 0; $v < $i; ++$v){
+				$vy = $currentY - $v;
+				if($vy < 0) break;
```
dktapps: Conditions should always use braces

src/world/generator/object/TreeFactory.php
```diff
@@ -37,6 +38,8 @@ public static function get(Random $random, ?TreeType $type = null) : ?Tree{
 			TreeType::JUNGLE => new JungleTree(),
 			TreeType::ACACIA => new AcaciaTree(),
 			TreeType::BIRCH => new BirchTree($random->nextBoundedInt(39) === 0),
+			TreeType::CRIMSON_FUNGUS_PLANTED => new NetherTree(VanillaBlocks::CRIMSON_STEM(), VanillaBlocks::NETHER_WART_BLOCK(), $random, true, true),
+			TreeType::WARPED_FUNGUS_PLANTED => new NetherTree(VanillaBlocks::WARPED_STEM(), VanillaBlocks::WARPED_WART_BLOCK(), $random, false, true),
```
dktapps: Please use named parameters for the bools here

src/world/generator/object/NetherTree.php
```diff
@@ -0,0 +1,196 @@
...
+		if($random->nextBoundedInt(12) === 0){
+			$i *= 2;
+		}
+		$this->huge = !$this->planted && $random->nextFloat() < 0.06;
+		parent::__construct($stemBlock, $hatBlock, $i);
+	}
+
+	public function canPlaceObject(ChunkManager $world, int $x, int $y, int $z, Random $random) : bool{
+		if($this->planted){
+			return true;
```
dktapps: I don't like this. Why is the placement logic different when placed by a player vs the generator?

src/block/Nylium.php
```diff
@@ -0,0 +1,102 @@
...
+		if($above instanceof Opaque){
+			BlockEventHelper::spread($this, VanillaBlocks::NETHERRACK(), $this);
+			return;
+		}
+
+		$random = new Random();
+		for($i = 0; $i < 4; ++$i){
+			$pos = $this->position->add($random->nextRange(-1, 1), $random->nextRange(-2, 0), $random->nextRange(-1, 1));
+			$block = $world->getBlock($pos);
+			if($block instanceof Netherrack && !$world->getBlock($pos->up()) instanceof Opaque){
```
dktapps: `instanceof` is generally frowned upon for blocks

src/block/Nylium.php
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	public function ticksRandomly() : bool{
+		return true;
+	}
+
+	public function onRandomTick() : void{
+		$world = $this->position->getWorld();
+		$above = $this->getSide(Facing::UP);
+		if($above instanceof Opaque){
```
dktapps: Same here

src/block/Nylium.php
```diff
@@ -0,0 +1,102 @@
...
+		$tx = new BlockTransaction($world);
+
+		$vegetation = $this->getTypeId() === BlockTypeIds::CRIMSON_NYLIUM ? [
+			VanillaBlocks::CRIMSON_FUNGUS(),
+			VanillaBlocks::CRIMSON_ROOTS()
+		] : [
+			VanillaBlocks::WARPED_FUNGUS(),
+			VanillaBlocks::WARPED_ROOTS(),
+			VanillaBlocks::NETHER_SPROUTS()
+		];
```
dktapps: Not sure how I feel about this. I know it's a pattern in older code, but I don't love hardcoding self type ID checks. It makes the class have limited use for plugin custom blocks.

src/block/NetherFungus.php
```diff
@@ -0,0 +1,86 @@
...
+						return true;
+					}
+				}
+			}
+		}
+		return parent::onInteract($item, $face, $clickVector, $player, $returnedItems);
+	}
+
+	private function grow(?Player $player) : bool{
+		$treeType = $this->getTypeId() === BlockTypeIds::CRIMSON_FUNGUS ? TreeType::CRIMSON_FUNGUS_PLANTED : TreeType::WARPED_FUNGUS_PLANTED;
```
dktapps: See above comments about hardcoding self type IDs

src/block/NetherFungus.php
```diff
@@ -0,0 +1,86 @@
...
+			$supportBlock->hasTypeTag(BlockTypeTags::NYLIUM) ||
+			$supportBlock->getTypeId() === BlockTypeIds::SOUL_SOIL;
+	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if($item instanceof Fertilizer){
+			$down = $this->getSide(Facing::DOWN);
+			if(
+				($this->getTypeId() === BlockTypeIds::CRIMSON_FUNGUS && $down->getTypeId() === BlockTypeIds::CRIMSON_NYLIUM) ||
+				($this->getTypeId() === BlockTypeIds::WARPED_FUNGUS && $down->getTypeId() === BlockTypeIds::WARPED_NYLIUM)
```
dktapps: Same here

dktapps (changes requested):

src/world/generator/object/NetherTree.php
Azvyl: > Huge fungi grow even with blocks above them, and do not replace any solid blocks when grown; however, they do not grow if blocked by the world height limit, or if situated outside of the world.

According to [the wiki](https://minecraft.wiki/w/Huge_fungus#cite_ref-1:~:text=Huge%20fungi%20grow%20even%20with%20blocks%20above%20them%2C%5B1%5D%20and%20do%20not%20replace%20any%20solid%20blocks%20when%20grown%3B%20however%2C%20they%20do%20not%20grow%20if%20blocked%20by%20the%20world%20height%20limit%2C%20or%20if%20situated%20outside%20of%20the%20world.), it turns out it will still grow in both conditions, should I change the function to return `true` or should I just remove the call from `getBlockTransaction()`?

Local test

image

seed: 397832894068853595

src/block/Nylium.php
Azvyl: We can compare type id for Netherrack, but for Opaque.. we use `isSolid()`?

src/block/Nylium.php
Azvyl: > custom blocks

Are we talking about custom Nylium? this class is `final`, but maybe to work around that we can make this class (and NetherFungus) `abstract`, then add a new class containing Nylium: `getVegetation()`; NetherFungus: `getTreeType()`, `getNyliumTypeId()` for each variant

src/block/Nylium.php
dktapps: `!isTransparent()` would be the usual way. Though you may want to question whether that makes sense

src/block/Nylium.php
dktapps: Or parameterising the constructor

Azvyl: > Or parameterising the constructor

That would be better because i wanted to avoid adding new classes

dktapps: Agreed. I prefer only to add classes if new logic is needed. Data should generally be injectable.

src/world/generator/object/NetherTree.php
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	public function getBlockTransaction(ChunkManager $world, int $x, int $y, int $z, Random $random) : ?BlockTransaction{
+		if(!$this->canPlaceObject($world, $x, $y, $z, $random)){
+			return null;
+		}
+
+		$transaction = new BlockTransaction($world);
+		$this->placeStem($x, $y, $z, $random, $this->generateTrunkHeight($random), $transaction);
+		$this->placeHat($x, $y, $z, $random, $transaction);
```
dktapps: Is there a reason to rename these functions? It looks to me like this logic is identical to the base tree class. Only the stem and hat functions have different names. If we're going to pretend these are trees, we might as well be consistent about it.

src/world/generator/object/NetherTree.php
dktapps: I'm not sure. It's possible we could rely on `BlockTransaction` validators, though I'm not sure if there's a reason we can't call this code when the player grows the fungus.

src/block/NetherFungus.php
```diff
@@ -0,0 +1,95 @@
...
+			$supportBlock->hasTypeTag(BlockTypeTags::NYLIUM) ||
+			$supportBlock->getTypeId() === BlockTypeIds::SOUL_SOIL;
+	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if($item instanceof Fertilizer){
+			$down = $this->getSide(Facing::DOWN);
+			if($down->getTypeId() === $this->nyliumTypeId){
+				if(mt_rand(1, 100) <= 40){
+					if($this->grow($player)){
```
dktapps: These conditions could be &&'d together

src/block/Nylium.php
```diff
@@ -0,0 +1,100 @@
...
+		if(!$above->isTransparent()){
+			BlockEventHelper::spread($this, VanillaBlocks::NETHERRACK(), $this);
+			return;
+		}
+
+		$random = new Random();
+		for($i = 0; $i < 4; ++$i){
+			$pos = $this->position->add($random->nextRange(-1, 1), $random->nextRange(-2, 0), $random->nextRange(-1, 1));
+			$block = $world->getBlock($pos);
+			if($block->getTypeId() === BlockTypeIds::NETHERRACK && $world->getBlock($pos->up())->isTransparent()){
```
dktapps, line 70: Beware this will replace stuff like glass, which you probably don't want

src/world/generator/object/NetherTree.php
```diff
@@ -0,0 +1,198 @@
...
+					$blockZ = $z + $k;
+
+					if($this->canOverride($transaction->fetchBlockAt($blockX, $blockY, $blockZ))){
+						if($isCorner){
+							if($random->nextFloat() < 0.1){
+								$transaction->addBlockAt($blockX, $blockY, $blockZ, $this->trunkBlock);
+							}
+						}else{
+							$transaction->addBlockAt($blockX, $blockY, $blockZ, $this->trunkBlock);
+						}
```
dktapps: This could be simplified to `if(!$isCorner || $random->nextFloat() < 0.1)`, not needing an `else` branch (and could also be merged into the outer condition)

src/world/generator/object/NetherTree.php
```diff
@@ -0,0 +1,198 @@
...
+					$transaction->addBlockAt($x, $vy, $z, VanillaBlocks::WEEPING_VINES()->setAge($vineAge));
+				}else{
+					break;
+				}
+			}
+		}
+	}
+
+	protected function canOverride(Block $block) : bool{
+		return $block->canBeReplaced() || $block instanceof NetherFungus || $block instanceof NetherVines;
```
dktapps: Don't like these `instanceof` checks. Consider using tags instead

src/block/Nylium.php
Azvyl: It's a check if the netherrack is not covered by a solid opaque block, so I think it will be fine

src/world/generator/object/NetherTree.php
Azvyl: I was previously only thinking about naming the helper functions (`placeHatBlock()`, `placeHatDropBlock()`)

src/block/Nylium.php
dktapps: Oh, it's replacing the netherrack itself. My bad

src/world/generator/object/NetherTree.php
Azvyl: `REPLACEABLE_BY_HUGE_FUNGUS`, `HUGE_FUNGUS_REPLACEABLE`, `FUNGUS_REPLACEABLE`, `FUNGUS_GROW_THROUGH`

I'm not sure about the tag naming

Azvyl: image

image

It turns out that the nether vine is not replaced in bedrock, but is destroyed in java

I didn't implement destroying the block

src/world/generator/object/TreeType.php
```diff
@@ -45,6 +47,8 @@ enum TreeType{
 	case JUNGLE;
 	case ACACIA;
 	case DARK_OAK;
+	case CRIMSON_FUNGUS_PLANTED;
+	case WARPED_FUNGUS_PLANTED;
```
Azvyl: Do I need to change this?
to `PLANTED_CRIMSON` and `PLANTED_WARPED` perhaps

src/world/generator/object/NetherTree.php
```diff
@@ -0,0 +1,191 @@
...
+		if(!$this->canPlaceObject($world, $x, $y, $z, $random)){
+			return null;
+		}
+
+		$transaction = new BlockTransaction($world);
+		$this->placeTrunk($x, $y, $z, $random, $this->generateTrunkHeight($random), $transaction);
+		$this->placeCanopy($x, $y, $z, $random, $transaction);
+
+		return $transaction;
+	}
```
dktapps: I don't think this function is needed anymore if inheritance is used correctly. The logic should be identical to the parent

src/world/generator/object/NetherTree.php
```diff
@@ -0,0 +1,191 @@
...
+			for($k = -$i; $k <= $i; ++$k){
+				$isCorner = $this->huge && abs($j) === $i && abs($k) === $i;
+
+				for($l = 0; $l < $trunkHeight; ++$l){
+					$blockX = $x + $j;
+					$blockY = $y + $l;
+					$blockZ = $z + $k;
+
+					if($this->canOverride($transaction->fetchBlockAt($blockX, $blockY, $blockZ))){
+						if(!$isCorner || $random->nextFloat() < 0.1){
```
dktapps: This can again be merged into the outer condition if you wrap it in (). Also this check is probably faster than the canOverride check, so you could put it before that

dktapps (commented):
I'm unclear what the intended behaviour is for the override checks. Currently, the fungus will just grow around and through obstructions, which doesn't seem like desired behaviour. More likely it should fail if it encounters a non-overridable block?

dktapps:
You could accomplish that by adding a validator to the `BlockTransaction` instead of having lots of `canOverride()` checks

Azvyl:
> You could accomplish that by adding a validator to the `BlockTransaction` instead of having lots of `canOverride()` checks

That would make it not grow at all, which is not vanilla behavior. Nether trees will still grow by replacing replaceable blocks (tested with tall grass, fire, glow lichen, resin clump, vine) and replacing nether fungus (regardless of the type of nether tree grown), non-replaceable blocks will remain unchanged

dktapps:
> > You could accomplish that by adding a validator to the `BlockTransaction` instead of having lots of `canOverride()` checks
> 
> That would make it not grow at all, which is not vanilla behavior. Nether trees will still grow by replacing replaceable blocks (tested with tall grass, fire, glow lichen, resin clump, vine) and replacing nether fungus (regardless of the type of nether tree grown), non-replaceable blocks will remain unchanged

Yeah, but then they'll grow through ceilings and get chopped in half by them. That doesn't make sense to me.

src/world/generator/object/TreeType.php
```diff
@@ -34,9 +34,9 @@
  * @method static TreeType DARK_OAK()
  * @method static TreeType JUNGLE()
  * @method static TreeType OAK()
+ * @method static TreeType PLANTED_CRIMSON()
+ * @method static TreeType PLANTED_WARPED()
  * @method static TreeType SPRUCE()
- * @method static TreeType CRIMSON_FUNGUS_PLANTED()
- * @method static TreeType WARPED_FUNGUS_PLANTED()
```
dktapps: You don't need to update this doc comment. It's only here for legacy purposes. New usages of new cases should use the native enum syntax.

dktapps:
Apparently Bugrock is stupider than I thought, once again:

Screenshot_20251018_220025_Minecraft

At this point I'm no longer interested in implementing Bugrock behaviour bug for bug and trying to guess whether what they did was intentional or not, on a general level.

This smacks of them forgetting to check for overlapping blocks. I would prefer implementing the Java growth behaviour as described in the wiki so we don't have to constantly guess whether it's correct or not. Eventually Bugrock will no doubt alter the behaviour to match Java after they get dozens of user bug reports over a few years, as with every other parity issue.

The chance of growing a huge fungus is fine to keep, since they can hardly do that by accident...

Azvyl:
> I would prefer implementing the Java growth behaviour as described in the wiki so we don't have to constantly guess whether it's correct or not.

Sorry
I don't think I understand what the problem is in the picture, [this](https://minecraft.wiki/w/Huge_fungus#:~:text=Huge%20fungi%20grow%20even%20with%20blocks%20above%20them%2C%5B1%5D%20and%20do%20not%20replace%20any%20solid%20blocks%20when%20grown%3B%20however%2C%20they%20do%20not%20grow%20if%20blocked%20by%20the%20world%20height%20limit%2C%20or%20if%20situated%20outside%20of%20the%20world.) also happens in Java Edition

image

> I would prefer implementing the Java growth behaviour as described in the wiki

Which part is missing from this implementation?

dktapps:
> > I would prefer implementing the Java growth behaviour as described in the wiki so we don't have to constantly guess whether it's correct or not.
> 
> Sorry I don't think I understand what the problem is in the picture, [this](https://minecraft.wiki/w/Huge_fungus#:~:text=Huge%20fungi%20grow%20even%20with%20blocks%20above%20them%2C%5B1%5D%20and%20do%20not%20replace%20any%20solid%20blocks%20when%20grown%3B%20however%2C%20they%20do%20not%20grow%20if%20blocked%20by%20the%20world%20height%20limit%2C%20or%20if%20situated%20outside%20of%20the%20world.) also happens in Java Edition
> image
> 
> > I would prefer implementing the Java growth behaviour as described in the wiki
> 
> Which part is missing from this implementation?

Oh, I see. I didn't expect Java to have this level of absurdity too...
In that case it may be done. I'll re-review it tomorrow.

src/block/Nylium.php
```diff
@@ -0,0 +1,100 @@
...
+		$tx = new BlockTransaction($world);
+
+		for($x = -2; $x <= 2; ++$x){
+			for($z = -2; $z <= 2; ++$z){
+				if($random->nextBoundedInt(3) === 0){
+					$pos = $this->position->add($x, 1, $z);
+					$replace = $world->getBlock($pos);
+					$place = $this->vegetation[$random->nextBoundedInt(count($this->vegetation))];
+					if($replace->getTypeId() === BlockTypeIds::AIR && $place->canBePlacedAt($replace, Vector3::zero(), Facing::DOWN, true)){
+						$tx->addBlock($pos, $place);
```
dktapps: I'm not sure why this uses a BlockTransaction. It doesn't look like it gets used for anything but setting blocks.

src/world/generator/object/NetherTree.php
```diff
@@ -0,0 +1,177 @@
...
+class NetherTree extends Tree{
+
+	private bool $huge;
+
+	public function __construct(Block $stemBlock, Block $hatBlock, Random $random, private readonly bool $hasVines, private readonly bool $planted = false){
+		$i = $random->nextBoundedInt(9) + 4;
+		if($random->nextBoundedInt(12) === 0){
+			$i *= 2;
+		}
+		$this->huge = !$this->planted && $random->nextFloat() < 0.06;
```
dktapps: I think `huge` should be a constructor parameter instead, and have the caller decide whether to use a huge or small one. That would be more plugin-friendly.

dktapps: Pass the height as a constructor parameter too

dktapps (changes requested):

src/block/Nylium.php
Azvyl: I was thinking about adding a StructureGrowEvent, what do you think?
I'm confused because in Grass there doesn't seem to be an event when using bonemeal.

dktapps: It's an open issue with bone meal in general. For now, it's not used. If we want to add events later, whoever does that can change it to use a transaction.

Azvyl:
Doing a little testing locally, it turns out that the huge variant can still be grown when using bonemeal

Screenshot_2025-10-20-00-20-53-97_5c8300b655012b1930f2e0a7b81bf6a9

dktapps:
Yep, seems fine to me

Only thing left seems to be the CS check (it's complaining about an unused import)

dktapps (dismissed):

src/block/NetherFungus.php
```diff
@@ -0,0 +1,91 @@
...
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if($item instanceof Fertilizer){
+			$down = $this->getSide(Facing::DOWN);
+			if($down->getTypeId() === $this->nyliumTypeId && mt_rand(1, 100) <= 40 && $this->grow($player)){
+				$item->pop();
+				return true;
+			}
+		}
+		return parent::onInteract($item, $face, $clickVector, $player, $returnedItems);
```
ipad54: I think this should return false instead of parent call for consistency with other onInteract() overrides

src/block/NetherFungus.php
```diff
@@ -0,0 +1,91 @@
...
+			$supportBlock->hasTypeTag(BlockTypeTags::DIRT) ||
+			$supportBlock->hasTypeTag(BlockTypeTags::MUD) ||
+			$supportBlock->hasTypeTag(BlockTypeTags::NYLIUM) ||
+			$supportBlock->getTypeId() === BlockTypeIds::SOUL_SOIL;
+	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if($item instanceof Fertilizer){
+			$down = $this->getSide(Facing::DOWN);
+			if($down->getTypeId() === $this->nyliumTypeId && mt_rand(1, 100) <= 40 && $this->grow($player)){
```
ipad54: I don’t think this is correct. In vanilla, bone meal is still subtracted on fail. Also, the chance should not be applied for creative players, vanilla lets them grow the fungus instantly.

src/block/Nylium.php
```diff
@@ -0,0 +1,97 @@
...
+use pocketmine\block\utils\BlockEventHelper;
+use pocketmine\item\Fertilizer;
+use pocketmine\item\Item;
+use pocketmine\math\Facing;
+use pocketmine\math\Vector3;
+use pocketmine\player\Player;
+use pocketmine\utils\Random;
+use function count;
+
+class Nylium extends Block{
```
ipad54: Shouldn’t this extend Opaque?

src/block/Nylium.php
```diff
@@ -0,0 +1,97 @@
...
+		}
+	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if($item instanceof Fertilizer && $this->getSide(Facing::UP)->getTypeId() === BlockTypeIds::AIR){
+			$item->pop();
+			$this->growVegetation(new Random());
+			return true;
+		}
+		return parent::onInteract($item, $face, $clickVector, $player, $returnedItems);
```
ipad54: Same about parent call

src/block/Nylium.php
```diff
@@ -0,0 +1,97 @@
...
+		$world = $this->position->getWorld();
+
+		for($x = -2; $x <= 2; ++$x){
+			for($z = -2; $z <= 2; ++$z){
+				if($random->nextBoundedInt(3) === 0){
+					$pos = $this->position->add($x, 1, $z);
+					$replace = $world->getBlock($pos);
+					$place = $this->vegetation[$random->nextBoundedInt(count($this->vegetation))];
+					if($replace->getTypeId() === BlockTypeIds::AIR && $place->canBePlacedAt($replace, Vector3::zero(), Facing::DOWN, true)){
+						$world->setBlock($pos, $place);
```
ipad54, line 95: This should check that the position is in the world

src/block/Nylium.php
```diff
@@ -0,0 +1,97 @@
...
+	private function growVegetation(Random $random) : void{
+		$world = $this->position->getWorld();
+
+		for($x = -2; $x <= 2; ++$x){
+			for($z = -2; $z <= 2; ++$z){
+				if($random->nextBoundedInt(3) === 0){
+					$pos = $this->position->add($x, 1, $z);
+					$replace = $world->getBlock($pos);
+					$place = $this->vegetation[$random->nextBoundedInt(count($this->vegetation))];
+					if($replace->getTypeId() === BlockTypeIds::AIR && $place->canBePlacedAt($replace, Vector3::zero(), Facing::DOWN, true)){
```
ipad54: I think canBePlacedAt() check is useless since vegetation can always be supported by nylium

ipad54 (changes requested):

ipad54:
Nylium should be affected by silk touch enchantment

ipad54:
Nether roots, red/brown mushrooms can also be supported by nylium

src/block/Nylium.php
Azvyl: `canBePlacedAt()` is needed because it can grow to blocks other than Nylium

Azvyl:
> This PR does not implement:
> - Nylium support for mushroom and NetherRoots
> - Using Bone Meal on Netherrack placed next to Nylium should convert the Netherrack into Nylium

Do i need to implement that Netherrack part too?

src/block/Nylium.php
ipad54: I see

ipad54:
> > This PR does not implement:
> > 
> > * Nylium support for mushroom and NetherRoots
> > * Using Bone Meal on Netherrack placed next to Nylium should convert the Netherrack into Nylium
> 
> Do i need to implement that Netherrack part too?

Ah, sorry, I didn’t read the follow up part. You can probably implement all of this in separate PRs if you wish

src/block/NetherFungus.php
Azvyl: I think sapling has this problem too

dktapps: Bone meal is generally not consistent with vanilla at the moment. But that's a topic for another time.

src/world/generator/object/TreeType.php
```diff
@@ -45,6 +45,8 @@ enum TreeType{
 	case JUNGLE;
 	case ACACIA;
 	case DARK_OAK;
+	case CRIMSON;
+	case WARPED;
 	//TODO: cherry blossom, mangrove, azalea
 	//TODO: do crimson and warped "trees" belong here? I'm not sure if they're actually trees or just fungi
```
dktapps: This TODO can be removed now

dktapps:
Will re-review this on Monday

src/world/generator/object/TreeType.php
Azvyl: To be sure, do i need to delete all three lines or just the last 2 lines?

Azvyl:
> > > This PR does not implement:
> > > 
> > > * Nylium support for mushroom and NetherRoots
> > > * Using Bone Meal on Netherrack placed next to Nylium should convert the Netherrack into Nylium
> > 
> > 
> > Do i need to implement that Netherrack part too?
> 
> Ah, sorry, I didn’t read the follow up part. You can probably implement all of this in separate PRs if you wish

I would really like to combine both into this PR, but usually these are made in separate PRs.
What do you think?

src/world/generator/object/TreeType.php
dktapps: Just this line only

dktapps:
Those should be easy to do, right? I don't see a reason not to do them here

src/block/Nylium.php
```diff
@@ -0,0 +1,101 @@
...
+use pocketmine\player\Player;
+use pocketmine\utils\Random;
+use function count;
+
+class Nylium extends Opaque{
+
+	/**
+	 * @param Block[] $vegetation An array of Block instances that can be grown on this Nylium block using Bone Meal.
+	 */
+	public function __construct(BlockIdentifier $idInfo, string $name, BlockTypeInfo $typeInfo, private readonly array $vegetation){
```
ShockedPlot7560, line 40: Unless I am mistaken, I do not see any discussion on this subject. If we keep this, a developer will not be able to add a block to bone meal without overriding the block—which we want to avoid. This may also be the case for grass. Should we modify it here or later to include all other cases?

ShockedPlot7560 (dismissed):
I didn't tested and reviewed the NetherTree structure yet. Will do later if needed

src/block/Nylium.php
Azvyl: https://github.com/pmmp/PocketMine-MP/pull/6843#discussion_r2441880618

Sorry I don't quite understand what you mean, `$vegetation` was added to avoid hardcoding self type IDs (`$this->getTypeId() === BlockTypeIds::CRIMSON_NYLIUM`)

ShockedPlot7560: It avoid to hardcode IDs in the block himself but if a plugin developper want to add their vegetations to nylium, a block override will be needed.

Azvyl: From what i understand, we didn't create it to help modify Nylium, but to add custom (native) blocks

DavyCraft648:
~I didn't implement Nylium support to mushrooms because their placement logic was incorrect, they shouldn't be able to be placed directly under the sky.~

src/block/Nylium.php
Azvyl: > a block override will be needed.

Are we talking about overriding the block or the class?

dktapps: We do need to rethink how bone meal works generally, but it should come as a more generalised API involving something like a `BoneMealEvent`. But this is outside scope for this PR.

dktapps (dismissed):
I think this is fine. Barring issues around bone meal handling, but that needs to be more generally reworked anyway and is a topic for another time.

src/block/Nylium.php
ShockedPlot7560: Yep, thats not only for this PR, it was a general suggestion not for only this PR scope.

src/block/Netherrack.php
```diff
@@ -23,9 +23,66 @@
...
+		if($item instanceof Fertilizer){
+			if($this->tryTransform()){
+				$item->pop();
+				return true;
+			}
+		}
+		return false;
+	}
+
+	public function tryTransform() : bool{
```
ipad54: This should be private

ipad54 (approved):

dktapps (approved):

dktapps:
FWIW, in the future it'd be better to split up the PR where possible. This PR contains quite a few additions that probably didn't need to be in a single PR. It didn't create problems this time, but might in other PRs.

Issue #6158, opened 2023-11-13
Timings auto paste


dktapps:
## Description
A setting to auto-paste timings on shutdown or periodically would be useful.

## Justification
I can't count the number of times I accidentally stopped my server without pasting timings, or forgot they were running.

Matze997:
Yes, pleaseeee
I've added an extra code just for that because I always forget to paste the stupid timing

remminiscent:
sounds simple enough, but before i take a go at this, is there a reason on why the important functions, such as creating the report file or uploading the report are within the command file itself? wouldnt it be better if they were in `TimingsHandler` class?

dktapps:
None that I'm aware of, it's just old ad-hoc code

remminiscent:
> None that I'm aware of, it's just old ad-hoc code

well then, ill most likely move it out of the command file. so it can also be used inside the `shutdown` function without code repettiton

dktapps:
It was decided not to upload the timings by default, but rather to write them to a file automatically instead. Users can then upload them to timings.pmmp.io manually if they desire.

Issue #5644, opened 2023-03-19
Allow pasting timings using an API method


dktapps:
## Description
Currently it's quite inconvenient to automatically paste timings, since only a command invocation can currently do it.

It might be useful for plugins to be able to automatically paste reports using the API, without having to use dispatchCommand() to do so.

## Justification
Automatic performance profiling plugins might find this useful.

PR #6848, opened 2025-10-20
Write timings to file on shutdown & add API for writing timings file

remminiscent:
if you have Timings enabled, as well as the option set to true in `pocketmine.yml`, when the server CORRECTLY shutsdown, itll provide 1 last timings report.

### Related issues & PRs
* Fixes https://github.com/pmmp/PocketMine-MP/issues/6158
* Fixes https://github.com/pmmp/PocketMine-MP/issues/5644

### API changes
maybe `TimingsHandler::uploadReport` & `TimingsHandler::createReportFile`

## Follow-up
possibly make it periodically send timings, given interval will also be in `pocketmine.yml`

## Tests
https://streamable.com/38giks


resources/pocketmine.yml
```diff
@@ -151,6 +151,9 @@ auto-updater:
 timings:
   #Choose the host to use for viewing your timings results.
   host: timings.pmmp.io
+  #Enable or disable automatic timings report uploading.
+  #If enabled, a report will be sent when the server shuts down. (if timings were on)
+  auto-paste: false
```
dktapps: `shutdown-auto-paste` would be better, in case we want to add other auto paste triggers in the future

src/timings/TimingsHandler.php
```diff
@@ -332,4 +343,98 @@ public function reset() : void{
...
 		$this->timingDepth = 0;
 	}
+
+	/**
+	 * Creates a timings report file locally in the server data folder.
+	 *
+	 * @param string[] $lines
+	 * @phpstan-param list<string> $lines
+	 */
+	public static function createReportFile(array $lines, CommandSender $sender) : void{
```
dktapps: Probably better to put this in a utility class instead of crowding `TimingsHandler`
Also, it'd be better if this didn't require a `CommandSender`. Let the caller decide how to deal with it.

src/timings/TimingsHandler.php
```diff
@@ -332,4 +343,98 @@ public function reset() : void{
...
+		Command::broadcastCommandMessage($sender, KnownTranslationFactory::pocketmine_command_timings_timingsWrite($timingsFile));
+	}
+
+	/**
+	 * Uploads a timings report to the configured host and provides the link to the user.
+	 *
+	 * @param string[] $lines
+	 * @phpstan-param list<string> $lines
+	 */
+	public static function uploadReport(array $lines, CommandSender $sender) : void{
```
dktapps: Same for this

dktapps (commented):
The `createReportFile()` and `uploadReport()` functions should handle collecting the timings as well as generating the files, so that the API isn't inconvenient for plugins. You'll want to return `Promise` to allow the caller to do stuff before & after.

src/timings/TimingsHandler.php
```diff
@@ -332,4 +343,98 @@ public function reset() : void{
...
+		$data = [
+			"browser" => $agent,
+			"data" => implode("\n", $lines),
+			"private" => "true"
+		];
+
+		$host = $server->getConfigGroup()->getPropertyString(
+			YmlServerProperties::TIMINGS_HOST,
+			"timings.pmmp.io"
+		);
```
dktapps: better to make this a parameter

src/timings/TimingsHandler.php
```diff
@@ -332,4 +343,98 @@ public function reset() : void{
...
+
+		if(!file_exists($timingsFolder)){
+			mkdir($timingsFolder, 0777);
+		}
+
+		$index = 0;
+		$timingsFile = Path::join($timingsFolder, "timings.txt");
+		while(file_exists($timingsFile)){
+			$timingsFile = Path::join($timingsFolder, "timings" . (++$index) . ".txt");
+		}
```
dktapps: This should be parameterisable too, although I'm not too sure how that'd work

src/timings/TimingsHandler.php
remminiscent: how would we provide `Server` to the functions if we remove CommandSender?

removing CommandSender will also mean having to remove the messages that are sent to them, what do you think?

dktapps: We shouldn't be requiring a `Server` in here either IMO.

An API method shouldn't be sending messages to the console. That should be done by the caller.

remminiscent: i understand, that were my intentions too for not sending messages in the functions, but how will we e.g get the pm ver, how will we get the scheduler for the task, etc without a `Server`

dktapps: Uh, not sure. To be honest I'm not sure any of these functions should be static really.

remminiscent: sure, but they still need a way to be called. some cool benchmarking stuff can be done with them.

remminiscent: perhaps put them in a class that has access to `Server`?

src/Server.php
```diff
@@ -1500,6 +1500,14 @@ public function shutdown() : void{
 		if($this->isRunning){
 			$this->isRunning = false;
 			$this->signalHandler->unregister();
+
+			$autoPaste = $this->getConfigGroup()->getPropertyBool(Yml::TIMINGS_SHUTDOWN_AUTO_PASTE, false);
+			$host = $this->getConfigGroup()->getPropertyString(Yml::TIMINGS_HOST, "timings.pmmp.io");
+			if($autoPaste && TimingsHandler::isEnabled() && (($sender = $this->consoleSender) instanceof ConsoleCommandSender)){
```
ShockedPlot7560: Why checking for the instance of the consoleSender ?

src/timings/TimingsHandler.php
```diff
@@ -332,4 +342,105 @@ public function reset() : void{
...
+
+	/**
+	 * Creates a timings report file locally in the server data folder.
+	 * Collects timings data and returns a promise that resolves with the timings lines.
+	 *
+	 * @param string $fileName A given name to the timings file
+	 *
+	 * @phpstan-return Promise<list<string>>
+	 */
+	public static function createReportFile(CommandSender $sender, string $fileName = "timings") : Promise{
```
ShockedPlot7560: If we give the ability to devs to create report files, it will be better to accept a different folder. For me, it will be better to have the filePath and not only fileName shadowed by the server datapath. 
For people wanted to put their own report file in a different folder or any linux stream or anything else, it will be a big headcache

src/timings/TimingsHandler.php
```diff
@@ -332,4 +342,105 @@ public function reset() : void{
...
+			function(array $lines) use ($sender, $fileName) : void{
+				$server = $sender->getServer();
+				$timingsFolder = Path::join($server->getDataPath(), "timings");
+				if(!file_exists($timingsFolder)){
+					mkdir($timingsFolder, 0777);
+				}
+				$index = 0;
+				$timingsFile = Path::join($timingsFolder, "{$fileName}.txt");
+				while(file_exists($timingsFile)){
+					$timingsFile = Path::join($timingsFolder, "{$fileName}" . (++$index) . ".txt");
```
ShockedPlot7560: This may represent a blocking loop if the folder isn't emptied many times. I know its like this today but wouldn't be better to suffix by the date or smth like this ?

src/timings/TimingsHandler.php
```diff
@@ -332,4 +342,105 @@ public function reset() : void{
...
+				fclose($handle);
+				Command::broadcastCommandMessage($sender, KnownTranslationFactory::pocketmine_command_timings_timingsWrite($timingsFile));
+			},
+			fn() => throw new \AssertionError("This promise is not expected to be rejected")
+		);
+		return $timingsPromise;
+	}
+
+	/**
+	 * Uploads a timings report to the configured host and provides the link to the user.
```
ShockedPlot7560:
```suggestion
	 * Uploads a timings report to the given host and provides the link to the user.
```

src/Server.php
```diff
@@ -1500,6 +1500,14 @@ public function shutdown() : void{
 		if($this->isRunning){
 			$this->isRunning = false;
 			$this->signalHandler->unregister();
+
+			$autoPaste = $this->getConfigGroup()->getPropertyBool(Yml::TIMINGS_SHUTDOWN_AUTO_PASTE, false);
+			$host = $this->getConfigGroup()->getPropertyString(Yml::TIMINGS_HOST, "timings.pmmp.io");
+			if($autoPaste && TimingsHandler::isEnabled() && (($sender = $this->consoleSender) instanceof ConsoleCommandSender)){				
+				TimingsHandler::uploadReport($sender, $host);
```
ShockedPlot7560: Wouldn't it be better for the pmmp server to create a default file report and offer the option to choose whether we want to download or create a file?

ShockedPlot7560 (changes requested):
In more for me

src/timings/TimingsHandler.php
dktapps: I'd say it's outside scope, but yeah we should probably change that

src/timings/TimingsHandler.php
dktapps: I'm not sure what to do about it. We do already have other `TimingsHandler` util functions that use `Server::getInstance()`, but we don't really want to do that here if it can be avoided. But requiring a `CommandSender` is also pretty annoying too.

src/Server.php
dktapps: Are you suggesting we just always dump the timings to a file on shutdown? That's ... a decent idea now I think about it

src/Server.php
remminiscent: i shouldve probably made it a null check, will fix.

src/timings/TimingsHandler.php
remminiscent: great idea. will fix

src/timings/TimingsHandler.php
remminiscent: {fileName}-{date}.txt sounds good. will do

src/Server.php
remminiscent: so if the option is enabled in `pocketmine.yml`, even if the timings arent on, on shutdown we always provide a timings report but as a file? doesnt sound bad

dktapps:
Per internal discussion, I think it would be better to:

- Always create a timings file locally on shutdown (if timings was enabled, ofc)
- Get rid of the auto paste - users will probably leave it enabled by mistake and accidentally upload timings reports they weren't aware of
- Change the file naming of timings files to be more like crashdumps

I think if we generate a report file, the user can then do what they want with the file (including uploading it manually), without running the risk of spamming the timings server or accidentally uploading stuff they didn't intend to.

remminiscent:
did what you requested, but i removed the ability to give custom names to the report file from external api usage, is that fine with you? or bring it back.

could be optional param set to null, if custom string is provided, itll use that. obviously with the index check for it too

src/timings/TimingsHandler.php
```diff
@@ -332,4 +342,113 @@ public function reset() : void{
...
+				// do we need this?
+				// while(file_exists($timingsFile)){
+				// 	$timingsFile = Path::join($timingsFolder, "{$date}-(" . (++$index) . ").txt");
+				// }
+				$handle = ErrorToExceptionHandler::trapAndRemoveFalse(fn() => fopen($timingsFile, "a+b"));
+				foreach($lines as $line){
+					fwrite($handle, $line . PHP_EOL);
+				}
+				fclose($handle);
+				Command::broadcastCommandMessage($sender, KnownTranslationFactory::pocketmine_command_timings_timingsWrite($timingsFile));
```
dktapps: This should be done by the caller, and only in the command. For the auto shutdown paste, I think a log message is sufficient.

src/timings/TimingsHandler.php
```diff
@@ -332,4 +342,113 @@ public function reset() : void{
...
+		$timingsPromise = self::requestPrintTimings();
+		$timingsPromise->onCompletion(
+			function(array $lines) use ($sender, $filePath) : void{
+				$server = $sender->getServer();
+
+				if($filePath === null){
+					$timingsFolder = Path::join($server->getDataPath(), "timings");
+				}else{
+					$timingsFolder = $filePath;
+				}
```
dktapps, line 366: I think this should just be a mandatory parameter

src/timings/TimingsHandler.php
```diff
@@ -332,4 +342,113 @@ public function reset() : void{
...
+				if(!file_exists($timingsFolder)){
+					mkdir($timingsFolder, 0777, true);
+				}
+				$date = date('Y-m-d_H.i.s_T');
+				$index = 0;
+				$timingsFile = Path::join($timingsFolder, "{$date}.txt");
+				// do we need this?
+				// while(file_exists($timingsFile)){
+				// 	$timingsFile = Path::join($timingsFolder, "{$date}-(" . (++$index) . ").txt");
+				// }
```
dktapps: This can be deleted

src/timings/TimingsHandler.php
```diff
@@ -332,4 +342,113 @@ public function reset() : void{
...
+				}else{
+					$timingsFolder = $filePath;
+				}
+
+				if(!file_exists($timingsFolder)){
+					mkdir($timingsFolder, 0777, true);
+				}
+				$date = date('Y-m-d_H.i.s_T');
+				$index = 0;
+				$timingsFile = Path::join($timingsFolder, "{$date}.txt");
```
dktapps: Probably better to keep some kind of prefix on this so it can be visually distinguished from a crashdump

src/timings/TimingsHandler.php
dktapps: To be clear, I'm saying that we should return a `Promise<timingsFile>`, rather than doing logging here, and let the caller decide what to do with that information.

dktapps:
> could be optional param set to null, if custom string is provided, itll use that. obviously with the index check for it too

I'd say allow it to be specified, but if it already exists, just bail out.

remminiscent:
> > could be optional param set to null, if custom string is provided, itll use that. obviously with the index check for it too
> 
> I'd say allow it to be specified, but if it already exists, just bail out.

wdym "just bail out"? also for the file path, i dont think we make it mandatory.

src/timings/TimingsHandler.php
```diff
@@ -332,4 +359,118 @@ public function reset() : void{
...
+				if($fileName === null){
+					$timingsName = "timings_{$date}.txt";
+				}else{
+					$timingsName = "{$fileName}.txt";
+				}
+				if(!file_exists($timingsFolder)){
+					mkdir($timingsFolder, 0777, true);
+				}
+				$timingsFile = Path::join($timingsFolder, $timingsName);
+				$handle = ErrorToExceptionHandler::trapAndRemoveFalse(fn() => fopen($timingsFile, "a+b"));
```
dktapps: This will throw if the file permission was denied, or if it already exists and isn't a folder
It should explicitly check that instead

dktapps:
> also for the file path, i dont think we make it mandatory.

I'm saying we should make it mandatory, so that the function doesn't have an inconvenient `CommandSender` dependency

src/timings/TimingsHandler.php
remminiscent: sorry, i didnt quiet understand what you mean.

remminiscent:
im really going to need to setup phpstorm if im gonna keep making PRs...

src/timings/TimingsHandler.php
```diff
@@ -361,37 +361,31 @@ public function reset() : void{
...
-				}else{
-					$timingsFolder = $filePath;
-				}
-				if($fileName === null){
-					$timingsName = "timings_{$date}.txt";
-				}else{
-					$timingsName = "{$fileName}.txt";
-				}
+
+				$timingsFolder = $filePath ?? Path::join($filePath, "timings");
```
dktapps: This will never be null

src/command/defaults/TimingsCommand.php
```diff
@@ -85,7 +85,7 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
 			if($paste){
 				TimingsHandler::uploadReport($sender, $host);
 			}else{
-				TimingsHandler::createReportFile($sender)->onCompletion(
+				TimingsHandler::createReportFile($sender->getServer()->getDataPath())->onCompletion(
```
dktapps: This needs to specify the full folder

src/Server.php
```diff
@@ -1502,7 +1502,7 @@ public function shutdown() : void{
 			$this->signalHandler->unregister();
 
 			if(TimingsHandler::isEnabled() && (($sender = $this->consoleSender) !== null)){
-				TimingsHandler::createReportFile($sender)->onCompletion(
+				TimingsHandler::createReportFile($this->getDataPath())->onCompletion(
```
dktapps: Same here

src/timings/TimingsHandler.php
dktapps: You're not dealing with the error cases properly. We shouldn't just overwrite the file if it already exists (which also might not work if it's not a file)

remminiscent:
I’ll finish this up once I’m actually awake. I’m making unnecessary mistakes right now because I’m too tired.

dktapps:
No rush

remminiscent:
for the "which also might not work if it's not a file", what do u want me to do? throw an exception, or reject the promise?

dktapps:
> for the "which also might not work if it's not a file", what do u want me to do? throw an exception, or reject the promise?

Reject the promise, we don't want to crash a random async task, but rather the caller who gave the invalid file name

src/timings/TimingsHandler.php
```diff
@@ -332,4 +359,120 @@ public function reset() : void{
...
+					mkdir($filePath, 0777, true);
+				}
+				$index = 0;
+				$timingsFile = Path::join($filePath, $timingsName . ".txt");
+				while(file_exists($timingsFile)){
+					if(!is_file($timingsFile)){
+						$resolver->reject();
+						return;
+					}
+					$timingsFile = Path::join($filePath, $timingsName . (++$index) . ".txt");
```
dktapps: I don't know why we're still doing this. What I meant was, if the provided file name already exists, just reject and return. Don't bother with the `(++$index)` stuff.

src/Server.php
```diff
@@ -1500,6 +1500,16 @@ public function shutdown() : void{
 		if($this->isRunning){
 			$this->isRunning = false;
 			$this->signalHandler->unregister();
+
+			if(TimingsHandler::isEnabled() && (($sender = $this->consoleSender) !== null)){
+				TimingsHandler::createReportFile($this->getDataPath() . DIRECTORY_SEPARATOR . "timings")->onCompletion(
+					function(string $timingsFile) use ($sender) : void{
+						Command::broadcastCommandMessage($sender, KnownTranslationFactory::pocketmine_command_timings_timingsWrite($timingsFile));
```
dktapps: A log message would be better, and wouldn't require a command sender that may not be present anyway. It's not the console sender doing this action.

remminiscent:
done 👍

src/Server.php
```diff
@@ -1500,6 +1500,16 @@ public function shutdown() : void{
...
 			$this->isRunning = false;
 			$this->signalHandler->unregister();
+
+			if(TimingsHandler::isEnabled()){
+				TimingsHandler::createReportFile($this->getDataPath() . DIRECTORY_SEPARATOR . "timings")->onCompletion(
+					function(string $timingsFile) : void{
+						$this->logger->info($this->language->translate(KnownTranslationFactory::pocketmine_command_timings_timingsWrite($timingsFile)));
+						TimingsHandler::setEnabled(false);
+					},
+					fn() => throw new \AssertionError("This promise is not expected to be rejected")
```
dktapps: This doesn't seem like a good idea. The promise could be rejected if the file already exists.

src/command/defaults/TimingsCommand.php
```diff
@@ -103,97 +80,23 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
...
-			);
+
+			if($paste){
+				TimingsHandler::uploadReport($sender, $host);
+			}else{
+				TimingsHandler::createReportFile($sender->getServer()->getDataPath() . DIRECTORY_SEPARATOR . "timings")->onCompletion(
+					function(string $timingsFile) use ($sender) : void{
+						Command::broadcastCommandMessage($sender, KnownTranslationFactory::pocketmine_command_timings_timingsWrite($timingsFile));
+					},
+					fn() => throw new \AssertionError("This promise is not expected to be rejected")
```
dktapps: As above

src/timings/TimingsHandler.php
```diff
@@ -332,4 +360,120 @@ public function reset() : void{
...
+				}
+				$handle = ErrorToExceptionHandler::trapAndRemoveFalse(fn() => fopen($timingsFile, "a+b"));
+				foreach($lines as $line){
+					fwrite($handle, $line . PHP_EOL);
+				}
+				fclose($handle);
+
+				$resolver->resolve($timingsFile);
+			},
+			fn() => throw new \AssertionError("This promise is not expected to be rejected")
```
dktapps: This should use `AssumptionFailedError`

src/timings/TimingsHandler.php
```diff
@@ -332,4 +360,120 @@ public function reset() : void{
...
+							}
+							Command::broadcastCommandMessage($sender, KnownTranslationFactory::pocketmine_command_timings_timingsRead($url));
+						}else{
+							$sender->getServer()->getLogger()->debug("Invalid response from timings server (" . $result->getCode() . "): " . $result->getBody());
+							Command::broadcastCommandMessage($sender, KnownTranslationFactory::pocketmine_command_timings_pasteError());
+						}
+					}
+				));
+			},
+			fn() => throw new \AssertionError("This promise is not expected to be rejected")
```
dktapps: As above

dktapps (changes requested):
The file report side looks mostly fine.

I still don't like the way the paste/upload stuff is handled though. It's inconvenient for plugins to need a `CommandSender`, and they also don't have any way to collect the timings URL with the current design. We also probably don't want to broadcast command audit messages when a plugin uses the upload API, for the same reason we don't want it broadcasting audit messages when generating a report on shutdown - it's not really the sender doing the action.

Since we don't need the upload API for implementing shutdown timings dumping, I'd be OK with reverting that part of the change and only doing the file report side for now. I think the upload stuff might require some thinking about how to deal with upload errors. The `Promise` API doesn't have any way to pass error information to the reject callback currently.

remminiscent:
> The file report side looks mostly fine.
> 
> I still don't like the way the paste/upload stuff is handled though. It's inconvenient for plugins to need a `CommandSender`, and they also don't have any way to collect the timings URL with the current design. We also probably don't want to broadcast command audit messages when a plugin uses the upload API, for the same reason we don't want it broadcasting audit messages when generating a report on shutdown - it's not really the sender doing the action.
> 
> Since we don't need the upload API for implementing shutdown timings dumping, I'd be OK with reverting that part of the change and only doing the file report side for now. I think the upload stuff might require some thinking about how to deal with upload errors. The `Promise` API doesn't have any way to pass error information to the reject callback currently.

i agree with the `CommandSender` param, ill see if i can figure a workaround, till then i did your requested changes

src/Server.php
```diff
@@ -1500,6 +1500,16 @@ public function shutdown() : void{
...
 			$this->isRunning = false;
 			$this->signalHandler->unregister();
+
+			if(TimingsHandler::isEnabled()){
+				TimingsHandler::createReportFile($this->getDataPath() . DIRECTORY_SEPARATOR . "timings")->onCompletion(
+					function(string $timingsFile) : void{
+						$this->logger->info($this->language->translate(KnownTranslationFactory::pocketmine_command_timings_timingsWrite($timingsFile)));
+						TimingsHandler::setEnabled(false);
+					},
+					fn() => throw new AssumptionFailedError("Failed to create timings report file")
```
dktapps: Uhh, this is not what I meant. `AssumptionFailedError` is only supposed to be thrown in cases which are thought to be impossible. Probably a log message needs to be generated here instead.

src/command/defaults/TimingsCommand.php
```diff
@@ -103,12 +95,24 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
...
+				$timingsPromise->onCompletion(
+					fn(array $lines) => $this->uploadReport($lines, $sender),
+					fn() => throw new AssumptionFailedError("This promise is not expected to be rejected")
+				);
+			}else{
+				TimingsHandler::createReportFile($sender->getServer()->getDataPath() . DIRECTORY_SEPARATOR . "timings")->onCompletion(
+					function(string $timingsFile) use ($sender) : void{
+						Command::broadcastCommandMessage($sender, KnownTranslationFactory::pocketmine_command_timings_timingsWrite($timingsFile));
+					},
+					fn() => throw new AssumptionFailedError("Failed to create timings report file")
```
dktapps: As above

src/timings/TimingsHandler.php
```diff
@@ -332,4 +342,53 @@ public function reset() : void{
...
+					$timingsName = "timings_{$date}";
+				}else{
+					$timingsName = "{$fileName}";
+				}
+				if(!file_exists($filePath)){
+					mkdir($filePath, 0777, true);
+				}
+				$timingsFile = Path::join($filePath, $timingsName . ".txt");
+				if(file_exists($timingsFile)){
+					if(!is_file($timingsFile)){
```
dktapps: Not sure what the point of this branch is. It's rejecting the promise either way.

src/Server.php
remminiscent: want me to send a logger alert message then?

dktapps:
Sorry for giving you so much of a runaround with this. Obviously a lot of these issues aren't really your fault because you're just moving around existing code. But in doing so it's highlighting new problems.

src/Server.php
dktapps: I'm undecided what to do with it tbh. The old code didn't handle these errors at all and it's raised some questions for me.

dktapps: But logging an error should be OK for now.

remminiscent:
> Sorry for giving you so much of a runaround with this. Obviously a lot of these issues aren't really your fault because you're just moving around existing code. But in doing so it's highlighting new problems.

it's absolutely fine, i do not mind at all. I guess our only issues are that in the upload function, we need to figure a way to not require a `CommandSender`, and that `Promise` does not provide a way to send an error "message" for it's rejection. Perhaps someone else can come up with a better way later. but for now ill finish up the shutdown timings report.

remminiscent:
for future reference, and maybe future PRs by me, why do you dislike `Server::getInstance()`?

dktapps:
> for future reference, and maybe future PRs by me, why do you dislike `Server::getInstance()`?

Because the dependencies aren't outwardly obvious and it's a pain for unit testing. Singletons in general are problematic in that regard.

src/timings/TimingsHandler.php
dktapps: Bump

src/Server.php
```diff
@@ -1500,6 +1500,16 @@ public function shutdown() : void{
 		if($this->isRunning){
 			$this->isRunning = false;
 			$this->signalHandler->unregister();
+
+			if(TimingsHandler::isEnabled()){
+				TimingsHandler::createReportFile($this->getDataPath() . DIRECTORY_SEPARATOR . "timings")->onCompletion(
```
dktapps: This can use `Path::join(dataPath, "timings")`

src/command/defaults/TimingsCommand.php
```diff
@@ -103,12 +96,21 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
...
-			);
+			if($paste){
+				$timingsPromise = TimingsHandler::requestPrintTimings();
+				Command::broadcastCommandMessage($sender, KnownTranslationFactory::pocketmine_command_timings_collect());
+				$timingsPromise->onCompletion(
+					fn(array $lines) => $this->uploadReport($lines, $sender),
+					fn() => throw new AssumptionFailedError("This promise is not expected to be rejected")
+				);
+			}else{
+				TimingsHandler::createReportFile($sender->getServer()->getDataPath() . DIRECTORY_SEPARATOR . "timings")->onCompletion(
```
dktapps: Same here

src/command/defaults/TimingsCommand.php
```diff
@@ -118,31 +120,7 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
...
-			fwrite($fileTimings, $line . PHP_EOL);
-		}
-		fclose($fileTimings);
-
-		Command::broadcastCommandMessage($sender, KnownTranslationFactory::pocketmine_command_timings_timingsWrite($timings));
-	}
-
-	/**
-	 * @param string[] $lines
+	 *
```
dktapps: Unrelated formatting change

dktapps (commented):
Almost done I think

src/command/defaults/TimingsCommand.php
```diff
@@ -38,25 +37,20 @@
 use pocketmine\utils\InternetRequestResult;
 use pocketmine\YmlServerProperties;
 use Symfony\Component\Filesystem\Path;
+
```
dktapps: Remove this line please

remminiscent: what else is left?

src/command/defaults/TimingsCommand.php
dktapps: One more here

dktapps:
Just one last formatting fix and this is good to go I think

src/command/defaults/TimingsCommand.php
remminiscent: huh? what u want me to do?

dktapps: Delete the green line, you added it

dktapps (approved):

remminiscent:
whats left for this? @dktapps @ShockedPlot7560

ShockedPlot7560 (approved):
Looks good to me as we focus on creating file report only

PR #6849, opened 2025-10-20
Optimize Living#lookAt()

armorshard1:
This PR proposes to change Living#lookAt() to avoid needlessly repeating the same operations

### DIFF
--- a/src/entity/Living.php
+++ b/src/entity/Living.php
@@ -930,12 +930,12 @@ abstract class Living extends Entity{
 	 * their heads to turn.
 	 */
 	public function lookAt(Vector3 $target) : void{
-		$horizontal = sqrt(($target->x - $this->location->x) ** 2 + ($target->z - $this->location->z) ** 2);
-		$vertical = $target->y - ($this->location->y + $this->getEyeHeight());
-		$pitch = -atan2($vertical, $horizontal) / M_PI * 180; //negative is up, positive is down
-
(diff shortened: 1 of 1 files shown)


src/entity/Living.php
```diff
@@ -930,13 +930,13 @@ public function getTargetBlock(int $maxDistance, array $transparent = []) : ?Blo
...
 	 */
 	public function lookAt(Vector3 $target) : void{
-		$horizontal = sqrt(($target->x - $this->location->x) ** 2 + ($target->z - $this->location->z) ** 2);
-		$vertical = $target->y - ($this->location->y + $this->getEyeHeight());
-		$pitch = -atan2($vertical, $horizontal) / M_PI * 180; //negative is up, positive is down
-
 		$xDist = $target->x - $this->location->x;
 		$zDist = $target->z - $this->location->z;
 
+		$horizontal = sqrt($xDist * $xDist + $zDist * $zDist);
```
dktapps: Why the change of operations?

armorshard1: Right, I didn't realise that **2 is used everywhere else. I'll change it back for consistency

ShockedPlot7560 (dismissed):
LGTM

dktapps:
For future reference, you don't need to force push PRs. In fact we generally prefer you don't do that, it's easier to review changes

dktapps (dismissed):

dktapps (approved):

PR #6851, opened 2025-10-20
Add deepslate variant of Infested Block

Azvyl:
Add deepslate variant of [Infested Block](https://minecraft.wiki/w/Infested_Block)
This PR does not implement silverfish spawning

1.17.0 - Caves & Cliffs

## Tests
https://youtu.be/klIeBHD1LYU


dktapps:
Not sure this is a good idea. As I've said elsewhere, we don't generally implement things without implementing them fully. This has caused a lot of headaches in the past. I know the existing infested stones were added in this condition, but I'm not convinced it's a good idea to worsen the problem.

Azvyl:
So this will also apply to [Creaking Heart](https://minecraft.wiki/w/Creaking_Heart) block (spawn [Creaking](https://minecraft.wiki/w/Creaking#Spawning)) and [brushable blocks](https://minecraft.wiki/w/Suspicious_Block) (requires [loot table](https://minecraft.wiki/w/Loot_table) depending on the structure where it is generated)
Unfortunately it doesn't seem to be possible at this time to implement silverfish without behavior

dktapps:
Brushable blocks could have their loot tables provided by plugins if we had events for it. Similar to what's been done for respawn anchor despite having no nether, it's fully functional and plugins can alter event behaviour to make it usable rather than exploding.

Possibly we could add events for infested stone that are cancelled by default (although `BlockBreakEvent` could also be used for this purpose anyway, I suppose). I haven't given this much thought.

Azvyl:
So does this pullrequest fulfill it?
The implementation of silverfish spawning can be done by a plugin

tests/phpstan/configs/actual-problems.neon
```diff
@@ -588,6 +588,12 @@ parameters:
 			count: 1
 			path: ../../../src/data/bedrock/block/convert/VanillaBlockMappings.php
 
+		-
+			message: '#^Parameter \#1 \$properties of method pocketmine\\data\\bedrock\\block\\convert\\Model\<pocketmine\\block\\InfestedStone\>\:\:properties\(\) expects list\<pocketmine\\data\\bedrock\\block\\convert\\property\\Property\<contravariant pocketmine\\block\\InfestedStone\>\>, array\{pocketmine\\data\\bedrock\\block\\convert\\property\\ValueFromStringProperty\<pocketmine\\block\\utils\\PillarRotation, int\>\} given\.$#'
```
dktapps: This looks suspicious

dktapps:
> So does this pullrequest fulfill it? The implementation of silverfish spawning can be done by a plugin

Fine, I suppose so.
I suppose we could add a specialised event for it with a `setEntity` API, so that plugins can avoid spawning multiple of them at once. But I don't have a good idea how that would work, so maybe it's better left for another time.

tests/phpstan/configs/actual-problems.neon
Azvyl: Not sure, but isn't the Infested Deepslate anonymous class already implementing PillarRotation?

dktapps: Oh, I see the problem. It's because the generated `VanillaBlocks` docblock return type doesn't account for interfaces implemented by anonymous classes...

dktapps: Should be fixed by https://github.com/pmmp/PocketMine-MP/commit/15bc98c9fbaa303ee609aa197b630d90f673b281

dktapps: Though, I think it'd be better to create a named class for this. PhpStorm for one doesn't appear to understand `A&B` in `@method`. Plus, it's easier to navigate to implementation if there's a named class.

src/block/InfestedRotatedPillarBlock.php
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\block;
+
+use pocketmine\block\utils\PillarRotation;
+use pocketmine\block\utils\PillarRotationTrait;
+
+class InfestedRotatedPillarBlock extends InfestedStone implements PillarRotation{
```
dktapps: `InfestedPillar` would be a simpler name

dktapps (approved):

PR #6857, opened 2025-10-22
Implement Powder Snow

Azvyl:
This pull request adds support for powder snow as a [block](https://minecraft.wiki/w/Powder_Snow) and [item](https://minecraft.wiki/w/Powder_Snow_Bucket), including its interactions with entities and buckets.

### Powder Snow Block and Item Implementation

* Added new `PowderSnow` block
* Added powder snow bucket as a new item. Includes bucket empty/fill powder snow sounds
* Introduced freezing mechanics for entities: entities accumulate freeze ticks when inside powder snow, which slow movement and eventually cause damage. Entities are protected from freezing if wearing leather armor
* Synced freeze progress ratio to clients for visual effects and added new damage cause for freezing
* Added new sound classes for powder snow bucket interactions

This PR does not implement:
- Damage should not occur if an entity falls onto a carpet placed on powder snow (#2041)
- Powder Snow Cauldron
- Entity attribute modifiers protocol (#1551)

1.17.0 - Caves & Cliffs

### Related issues & PRs
#5964
#5163

### API changes
- Added `PowderSnow` class
- Added `SolidBucket` class
- Added `Living::getFreezeProgressTicks()`
- Added `Living::setFreezeProgressTicks(int $freezeProgressTicks)`
- Added `Living::getFreezeProgressRatio()`
- Added `Living::getFreezeThresholdTicks()`
- Added `Living::getFreezeProgressState()`
- Added `Living::setFreezeProgressState(?bool $state)`
- Added `Living::canFreeze()`
- Added `EntityDamageEvent::CAUSE_FREEZE`

## Tests
https://www.youtube.com/watch?v=QVkKMMUlHjE

Plugin:
https://gist.github.com/DavyCraft648/ca56d2587da2b08a983cf3e648e5c0c8


src/entity/Entity.php
```diff
@@ -760,6 +769,98 @@ protected function dealFireDamage() : void{
...
+			if($this->freezeProgressTicks >= $threshold && (($this->freezeProgressTicks % 40 === 0) || $tickDiff > 40)){
+				$this->applyFreezeDamage();
+			}
+
+			$this->onFreezeAttributeModifierChanged(-0.05 * $this->getFreezeProgressRatio());
+			return true;
+		}
+
+		if($this->freezeProgressTicks > 0){
+			$this->setFreezeProgressTicks(max(0, min($this->freezeProgressTicks, $threshold) - 2 * $tickDiff));
```
Azvyl: Maybe we can add `isFreezingProgressLocked` to prevent the reduction and allow plugins to make the freeze effect persistent?

src/entity/Entity.php
```diff
@@ -760,6 +769,98 @@ protected function dealFireDamage() : void{
...
+			return true;
+		}
+
+		return false;
+	}
+
+	protected function applyFreezeDamage() : void{
+		$ev = new EntityDamageEvent($this, EntityDamageEvent::CAUSE_FREEZE, 1);
+		$this->attack($ev);
+	}
```
dktapps: I don't think this stuff should be in `Entity`

src/item/SolidBucket.php
```diff
@@ -0,0 +1,64 @@
...
+			$returnedItems[] = $ev->getItem();
+			return ItemUseResult::SUCCESS;
+		}
+
+		return ItemUseResult::FAIL;
+	}
+
+	public function getBlock(?int $clickedFace = null) : Block{
+		return $this->block;
+	}
```
dktapps: This probably shouldn't be here. It's not a placeable item

dktapps:
I don't think #2041 should affect this. Carpets are a thin layer at the bottom of their cell, so you can detect when an entity enters that cell with `onEntityInside()`. #2041 only affects full-blocks which need to be collidable.

src/entity/Entity.php
Azvyl: Do we empty `applyFreezeDamage()` in `Entity` and then override it in `Living`?

src/item/SolidBucket.php
Azvyl: Should we just delete it? or rename it to `getContainedBlock()`?

dktapps: Just get rid of it, I don't see a need for it

src/entity/Entity.php
dktapps: I don't think it should be handled by `Entity` at all.

Azvyl:
I think this seems to be something missing in pocketmine, Slime block should also not deal damage if there is carpet on top of it. Jumping on a Carpet can also damage Farmland and on Beds it also gives a bouncing effect if they are under it.
Do we need to create an onEntityLand function on the carpet that calls onEntityLand on the block below it?

dktapps:
> I think this seems to be something missing in pocketmine, Slime block should also not deal damage if there is carpet on top of it. Jumping on a Carpet can also damage Farmland and on Beds it also gives a bouncing effect if they are under it. Do we need to create an onEntityLand function on the carpet that calls onEntityLand on the block below it?

Hmm, not sure. Maybe leave that out of this PR then.

dktapps:
I'm not very clear how powder snow is supposed to work. Are there any non-mobs affected by it? It seems weird to me that general entities should be affected.

Azvyl:
Freezing a burning entity should extinguish its fire, such as an arrow fired with the flame enchantment. However, we handle it in the PowderSnow block here.
Should we just move the freeze codes to Living?

dktapps:
Yeah, I think so. Unless there are any non-living entities that can have the freeze effects.

Azvyl:
I've attached a plugin to the test section for plugin developers who want to try it out, in case they want to create a troll plugin or custom enchant. Any suggestions would be helpful.

src/entity/Living.php
```diff
@@ -438,6 +444,127 @@ public function setOnFire(int $seconds) : void{
...
+	 */
+	protected function onFreezeAttributeModifierChanged(float $addValue) : void{
+		$base = $this->moveSpeedAttr->getDefaultValue();
+		$oldAdd = $this->freezeMovementAdd;
+		$denom = max(1e-6, $base + $oldAdd);
+		$multiplier = $this->getMovementSpeed() / $denom;
+		$target = ($base + $addValue) * $multiplier;
+
+		$this->setMovementSpeed(max(0.0, $target));
+		$this->freezeMovementAdd = $addValue;
```
dktapps: This function is giving me a headache. What's it trying to do?

src/entity/Living.php
```diff
@@ -438,6 +444,127 @@ public function setOnFire(int $seconds) : void{
...
+	 * @param bool $locked Whether to lock freeze ticking
+	 */
+	public function lockFreezeTicks(bool $locked = true) : void{
+		$this->freezeLocked = $locked;
+	}
+
+	/**
+	 * Whether this entity can be frozen (i.e. accumulate freeze progress from environments such as powder snow).
+	 */
+	public function isFreezable() : bool{
```
dktapps: `canFreeze` would probably make more sense

src/entity/Living.php
```diff
@@ -438,6 +444,127 @@ public function setOnFire(int $seconds) : void{
...
+	 * Returns the number of ticks required for this entity to be considered fully frozen.
+	 */
+	public function getFreezeThresholdTicks() : int{
+		return 140;
+	}
+
+	/**
+	 * Returns whether the entity is currently flagged as accumulating freeze progress (transient).
+	 */
+	public function isAccumulatingFreeze() : bool{
```
dktapps: Maybe `isFreezing` would be sufficient?

src/entity/Living.php
```diff
@@ -438,6 +444,127 @@ public function setOnFire(int $seconds) : void{
...
+		return $this->freezeLocked;
+	}
+
+	/**
+	 * Locks or unlocks freeze ticking for this entity. When locked, the entity will not accumulate or lose freeze
+	 * progress.
+	 *
+	 * @param bool $locked Whether to lock freeze ticking
+	 */
+	public function lockFreezeTicks(bool $locked = true) : void{
```
dktapps: What's the purpose of this?

src/block/PowderSnow.php
```diff
@@ -0,0 +1,83 @@
...
+	public function hasEntityCollision() : bool{
+		return true;
+	}
+
+	public function getDropsForCompatibleTool(Item $item) : array{
+		return [];
+	}
+
+	public function getPickedItem(bool $addUserData = false) : Item{
+		return VanillaItems::POWDER_SNOW_BUCKET();
```
dktapps, line 42: Are you sure this is correct? Normally picking doesn't give you items that you don't already have (aside from the block itself, of course). I'd expect this to fail rather than giving a bucket.

src/block/PowderSnow.php
```diff
@@ -0,0 +1,83 @@
...
+		if($player !== null && $item->getTypeId() === ItemTypeIds::BUCKET){
+			if(!$player->isCreative()){
+				$item->pop();
+				$returnedItems[] = VanillaItems::POWDER_SNOW_BUCKET();
+			}
+			$this->getPosition()->getWorld()->setBlock($this->getPosition(), VanillaBlocks::AIR());
+			$this->getPosition()->getWorld()->addSound($this->getPosition(), new BucketFillPowderSnowSound());
+
+			return true;
+		}
```
dktapps: This is not consistent with how buckets are normally implemented (and also won't call `PlayerBucketFillEvent`)

src/block/PowderSnow.php
```diff
@@ -0,0 +1,83 @@
...
+		if($entity->isOnFire()){
+			$entity->extinguish(EntityExtinguishEvent::CAUSE_POWDER_SNOW);
+			BlockEventHelper::melt($this, VanillaBlocks::AIR());
+		}
+		return true;
+	}
+
+	public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player $player = null, array &$returnedItems = []) : bool{
+		if($player !== null && $item->getTypeId() === ItemTypeIds::BUCKET){
+			if(!$player->isCreative()){
```
dktapps: Generally we don't expect this to be needed. `$returnedItems` is expected to handle this.

dktapps (changes requested):

src/entity/Living.php
Azvyl: `isFreezing` is closer in meaning to `freezeProgressTicks > 0`, so it's more to avoid ambiguity

src/entity/Living.php
Azvyl: > Sets if the entity currently has its freeze ticks locked, preventing default vanilla freeze tick modification.

This prevents increasing/decreasing `freezeProgressTicks` in `updateFreezeState()` (can be used by plugins to display a persistent frosty vignette).

src/block/PowderSnow.php
Azvyl: Powder Snow normally doesn't have an item form, in vanilla you'll get a Powder Snow Bucket by blockpicking.

src/block/PowderSnow.php
Azvyl: Should we move this into the `Bucket` class? Then what if we create a `BucketableBlock` interface for Powder Snow (and Liquid)?

src/entity/Living.php
Azvyl: The concept here is that freeze is an `ADDITION` modifier which is applied to the attribute before any multiplicative modifiers are calculated. First I remove the old additive value from the base value, apply the freeze value to that base, and then reapply the multiplicative scaling to the modified total.

src/block/PowderSnow.php
dktapps: Not sure why an interface would be needed, but yes, for the sake of consistency the logic should be in `Bucket`.

There's some questions about why buckets are implemented this way, but a resolution hasn't yet been reached, so for now it should be implemented the same way as the others to maintain behavioural consistency.

src/entity/Living.php
dktapps: I'm not convinced that will have the desired effect. You don't know if the current movement speed might've been modified by something else. For example sprinting

dktapps: Perhaps we need to think about implementing attribute modifiers properly rather than doing this hack. But that's probably a job for a separate PR.

src/block/PowderSnow.php
Azvyl: Strangely after I moved to Bucket, `$returnedItems` seems to still apply to creative players

src/entity/Living.php
Azvyl: So shall we ignore (remove) the implementation of the _decreasing FOV_ for now?

src/block/PowderSnow.php
dktapps: Hmm, it's probably affected by #3898. If it happens by existing code, don't worry about it.

src/data/bedrock/block/convert/VanillaBlockMappings.php
```diff
@@ -466,6 +466,7 @@ private static function registerSimpleIdOnlyMappings(BlockSerializerDeserializer
 		$reg->mapSimple(Blocks::NETHER_SPROUTS(), Ids::NETHER_SPROUTS);
 		$reg->mapSimple(Blocks::CRIMSON_NYLIUM(), Ids::CRIMSON_NYLIUM);
 		$reg->mapSimple(Blocks::WARPED_NYLIUM(), Ids::WARPED_NYLIUM);
+		$reg->mapSimple(Blocks::POWDER_SNOW(), Ids::POWDER_SNOW);
```
dktapps: This should be placed in alphabetical order with other blocks

src/entity/Living.php
```diff
@@ -438,6 +444,127 @@ public function setOnFire(int $seconds) : void{
...
+	 * Returns the number of ticks this entity has accumulated toward being frozen.
+	 */
+	public function getFreezeProgressTicks() : int{
+		return $this->freezeProgressTicks;
+	}
+
+	/**
+	 * Set the entity's freeze progress in ticks.
+	 *
+	 * @throws \InvalidArgumentException if $freezeProgressTicks is negative
```
dktapps: We don't usually document these

src/entity/Living.php
dktapps: As a verb, "freezing" means to approach a state of being frozen. Going the other way would be considered melting. So I'm not sure I agree on that.

I'm not married to "is freezing", but I think "accumulating freeze" sounds very awkward and unintuitive.

src/entity/Living.php
```diff
@@ -438,6 +444,127 @@ public function setOnFire(int $seconds) : void{
...
+	/**
+	 * Returns whether the entity is currently flagged as accumulating freeze progress (transient).
+	 */
+	public function isAccumulatingFreeze() : bool{
+		return $this->isAccumulatingFreeze;
+	}
+
+	/**
+	 * Sets the transient freezing flag. This flag indicates that the entity should accumulate freeze progress during
+	 * the next run of `entityBaseTick()`.
```
dktapps: This sounds like a lot of words to say `Sets whether the entity is accumulating freeze progress` or something to that effect. (Although I don't much like "accumulating freeze" as I mentioned above.)

src/entity/Living.php
```diff
@@ -438,6 +444,127 @@ public function setOnFire(int $seconds) : void{
...
+	 *
+	 * @param bool $accumulating Whether to start accumulating freeze progress
```
dktapps: This seems redundant

src/entity/Living.php
```diff
@@ -438,6 +444,127 @@ public function setOnFire(int $seconds) : void{
...
+	 */
+	public function isFreezeTickingLocked() : bool{
+		return $this->freezeLocked;
+	}
+
+	/**
+	 * Locks or unlocks freeze ticking for this entity. When locked, the entity will not accumulate or lose freeze
+	 * progress.
+	 *
+	 * @param bool $locked Whether to lock freeze ticking
```
dktapps: This seems redundant

src/entity/Living.php
dktapps: I feel like this could be wrapped into the freezing flag as `?bool`, where `null` would indicate that the freeze state isn't progressing in either direction.

src/entity/Living.php
dktapps: Also, when this flag is false, the freeze amount is decreasing if positive. The name of this function currently suggests nothing is happening when the result is false.

src/entity/Living.php
dktapps: Yes, I think so. I don't like cooking hacks like this into the code.

src/event/entity/EntityDamageEvent.php
```diff
@@ -65,6 +65,7 @@ class EntityDamageEvent extends EntityEvent implements Cancellable{
 	public const CAUSE_CUSTOM = 14;
 	public const CAUSE_STARVATION = 15;
 	public const CAUSE_FALLING_BLOCK = 16;
+	public const CAUSE_FREEZE = 17;
```
dktapps: `CAUSE_FREEZING` would be more correct

dktapps (commented):
From what I can gather, the implementation seems mostly OK. My main issues are with the API design.

src/entity/Living.php
```diff
@@ -438,6 +443,84 @@ public function setOnFire(int $seconds) : void{
...
+	}
+
+	protected function updateFreezeState(int $tickDiff) : bool{
+		$threshold = $this->getFreezeThresholdTicks();
+		if($this->freezeProgressState === true){
+			$this->setFreezeProgressTicks($this->freezeProgressTicks + $tickDiff);
+			if($this->freezeProgressTicks >= $threshold && (($this->freezeProgressTicks % 40 === 0) || $tickDiff > 40)){
+				$this->applyFreezeDamage();
+			}
+			//apply movement modifier
```
dktapps: Is this supposed to be a TODO?

src/block/PowderSnow.php
```diff
@@ -0,0 +1,70 @@
...
+use pocketmine\item\Item;
+use pocketmine\item\VanillaItems;
+use pocketmine\world\sound\BucketFillPowderSnowSound;
+use pocketmine\world\sound\Sound;
+
+class PowderSnow extends Flowable{
+
+	public function hasEntityCollision() : bool{
+		return true;
+	}
```
dktapps: This should be placed near the affected functions

dktapps (commented):
This is looking mostly done now I think

src/entity/Living.php
Azvyl: Yes, but to be implemented outside this PR

dktapps: Ok, please prefix it with TODO: so it shows up in searches

src/block/PowderSnow.php
Azvyl: Ah
now I'm confused should we use `getPickedItem()` or `asItem(`) for this

dktapps: I think this is fine. Like I said, I think it's a bit cursed for this to spawn a bucket out of nowhere anyway, but we need to follow vanilla in this instance...

We don't override asItem() for liquids e.g.

dktapps:
This looks good now. Will test soon and feedback

dktapps:
Looks like we have a simulation issue with the collision boxes. Since the collision AABB is different if the entity is wearing leather boots or not (they should be able to stand on the powder snow). For non player entities I don't think this will work currently.

I'm not too sure how that should be addressed though, since we don't really have the systems to deal with that...

Azvyl:
Hmm... Some mobs like foxes can also stand on powder snow (unless it does a jumping attack on a chicken in the powder snow). This might be a problem if we want to implement Scaffolding block. I think the collision is just as complex because it only prevents falling, but allows for passing from the side.
Should we enforce the collision as a full block for now?

dktapps:
No, that would prevent regular mobs from falling inside it.

This needs to be fixed properly with a more general solution. I have a rough idea how to deal with it, but collision detection is very performance-sensitive code, so it needs to be handled carefully.

Issue #3897, opened 2020-11-09
EffectInstance is mutated by its target entity


dktapps:
### Issue description
Plugin devs have been surprised by the fact that it's not possible to keep an `EffectInstance` around and reapply it to several mobs. This is a reasonable ask since it's possible to want to instantiate an effect with a predefined duration, strength etc. without having to make a new `EffectInstance` each time.

Of course, it _is_ possible to simply clone the `EffectInstance` before using it, but this takes us right back to square one of the `Effect` type/instance merged model which is the whole reason why `EffectInstance` was introduced to begin with.

This issue proposes an `ActiveEffectInstance` type which will be used internally by `EffectManager` to track duration changes for each entity. It will contain the following methods moved from `EffectInstance`:
- `getDuration()`
- `setDuration()`
- `decreaseDuration()`
- `hasExpired()`

I'm not a big fan of the name, so I have also considered `EffectDurationTracker`, but I'm not very happy with that either.

### Caveats
This adds yet another layer of abstraction to the `Effect` API, although one that won't affect developers very much. I've been avoiding this change for a long time for this exact reason.

It comes to my attention that `Effect` and `EffectInstance` might need renaming to make the intents of the different parts of this system more clear. Perhaps something like this:
- `Effect` -> `EffectType`
- `EffectInstance` -> `EffectParameters` (?)

I'm open to better naming suggestions.

ConnorHopper:
no

dktapps:
Probably better to just add a map<EffectInstance, duration> to the `EffectManager`, and then add a method like `EffectManager::getRemainingDuration(Effect $effect) : int`.

However, this would still be a behavioural BC break, because `$effectManager->get($effect)->getDuration()` would no longer reflect the _remaining_ duration but instead the _initial_ duration (and `decreaseDuration()` and `setDuration()` would have no effect).

While we're breaking BC, might as well make `EffectInstance` fully immutable too, to avoid all the kinds of headaches that come from doing this while forgetting to clone objects: https://github.com/pmmp/PocketMine-MP/blob/45917d495c14f3ce37bbf0848262ad8a470030f3/src/entity/projectile/SplashPotion.php#L117-L121

PR #6866, opened 2025-10-25
infinite effects

### LINKED ISSUE #6862: Implement infinite effects (not as easy as it seems)
This has now been attempted and abandoned by third-party contributors twice.

Since this isn't as easy as it looks, I thought it best to write up some considerations if someone wants to take another shot at it:

- We can't use `-1` or any negative number for infinite duration internally, because this will be a BC break for plugins doing `getDuration() <= 0` to check for expired effects. We'd probably have to use `PHP_INT_MAX` for this instead, and then translate it to `-1` on network or for disk NBT storage.
- Periodic effects like poison, regeneration etc. still need a clock. Currently the duration ticker is used for this. We'd need some other mechanism to tick these effects if the duration wasn't updated.
- `EffectCommand` needs different output messages for infinite effects.

_Originally posted by @dktapps in https://github.com/pmmp/PocketMine-MP/issues/6780#issuecomment-3339260528_

remminiscent:
Implements Infinite effects.
### Related issues & PRs
* Fixes https://github.com/pmmp/PocketMine-MP/issues/6862
### API changes
`EffectInstance->isInfinite()`
## Backwards compatibility
no break.... hopefully
## Tests
https://streamable.com/ffaecx


remminiscent:
pocketmine commands are so outrageously bad i dont even know how to add the infinite argument, im going to need your help to implement the command part @dktapps

remminiscent:
how can i add the new message?

remminiscent:
ahhh, checking with tickable effects, they dont work... apologies.

what did you mean by "Clock" @dktapps

dktapps:
> ahhh, checking with tickable effects, they dont work... apologies.
> 
> what did you mean by "Clock" @dktapps

I mean that the tickable effects need some way to know when to tick, currently it uses the duration, but if the duration is not updated that's obviously not gonna work.

Possibly we could just init them with `PHP_INT_MAX` and then treat anything with duration > `Limits::INT32_MAX` as infinite. Since it would take longer than the current age of the universe for the ticker to count down from `PHP_INT_MAX` to `Limits::INT32_MAX` anyway, that would allow infinite effects to still tick

src/command/defaults/EffectCommand.php
```diff
@@ -113,7 +117,14 @@ public function execute(CommandSender $sender, string $commandLabel, array $args
 		}else{
 			$instance = new EffectInstance($effect, $duration, $amplification, $visible);
 			$effectManager->add($instance);
-			self::broadcastCommandMessage($sender, KnownTranslationFactory::commands_effect_success($effect->getName(), (string) $instance->getAmplifier(), $player->getDisplayName(), (string) ($instance->getDuration() / 20)));
+
+			$durationDisplay = $duration === Limits::INT32_MAX ? "infinite" : (string) ($instance->getDuration() / 20);
```
dktapps: This will look weird for non-English users. I'm fairly sure there should be a dedicated string for this in vanilla if you check their language files.

remminiscent: there is, however im unknowledgeable regarding language files and pocketmine translation stuff.

dktapps: Just add a string to `resources/translations/eng.ini` following the PM format, and then run `composer update-codegen` to regenerate the `KnownTranslationFactory`

dktapps: https://github.com/pmmp/PocketMine-MP/blob/stable/resources/translations/README.md

remminiscent:
ok ill take a go at what you said. 

> > ahhh, checking with tickable effects, they dont work... apologies.
> > what did you mean by "Clock" @dktapps
> 
> I mean that the tickable effects need some way to know when to tick, currently it uses the duration, but if the duration is not updated that's obviously not gonna work.
> 
> Possibly we could just init them with `PHP_INT_MAX` and then treat anything with duration > `Limits::INT32_MAX` as infinite. Since it would take longer than the current age of the universe for the ticker to count down from `PHP_INT_MAX` to `Limits::INT32_MAX` anyway, that would allow infinite effects to still tick

wouldnt an exception be sent?

`		
if($duration < 0 || $duration > Limits::INT32_MAX){
			throw new \InvalidArgumentException("Effect duration must be in range 0 - " . Limits::INT32_MAX . ", got $duration");
		}
`
		(idk how to reference a piece of code.)

dktapps:
> wouldnt an exception be sent?
> 
> ` if($duration < 0 || $duration > Limits::INT32_MAX){ throw new \InvalidArgumentException("Effect duration must be in range 0 - " . Limits::INT32_MAX . ", got $duration"); } ` (idk how to reference a piece of code.)

We can change that code if need be. I'm not too sure what implications there might be for plugins saving the duration into persistent storage though.

remminiscent:
im not gonna lie to you bro. i cant figure it out... sounds simple but i cant seem to know what the goal is

dktapps:
My thinking is that if we set the duration to a sufficiently large number, we can let it count down (so still allowing the ticking effects to work) while making sure its duration is so large that it'll never run out during a normal server runtime.
Then we can just check if `duration > Limits::INT32_MAX` to see if the effect is infinite.

I'm not sure how well this would work in practice though. It raises some questions about what should happen when doing `setDuration(INT32_MAX < number < PHP_INT_MAX)`. Should doing `setDuration(PHP_INT_MAX - 10)` be allowed?
- It would turn into `PHP_INT_MAX` after saving and reloading, since we can't save a 64-bit value for the duration. This might be confusing in itself.
- But if we don't allow it, then `getDuration()` would be able to return values that `setDuration()` doesn't accept (it would have to, so that code like [this](https://github.com/pmmp/PocketMine-MP/blob/55c110e5c700c713120b6063b24d64da9128ed0b/src/entity/effect/PoisonEffect.php#L42) could work correctly). This would also be confusing for plugin developers.

I haven't given this much thought, so I'm not sure what other possible options there are. But if the duration isn't allowed to count down, then anything that uses the poison-style approach to tick effects will break. So we're kinda constrained with our options in 5.x.

In 6.0 we could do this any way we like, but I doubt anyone would want to wait for 6.0 to get this seemingly trivial feature.

remminiscent:
i mean, for now we can do the php int max, and int32 max approach, where if the duration is PHP_INT_MAX, itll still count down, then for save and loading nbt storage, if the duration is above int32 max, we just save -1, and when loading, we load php_int_max.


then we can specifiy in the doc block that ONLY PHP_INT_MAX will provide an infinite effect, anything below it will provide a normal effect, but if someone does PHP_INT_MAX - 10, itll get converted to INT32_MAX - 10?

dktapps:
Don't love that. `$newEffect->setDuration($infiniteEffect->getDuration())` would then become non-infinite.

remminiscent:
so whats the plan now? change target branch to major next and make it an actual good implementation of infinite effects, or figure a workaround for minor next? @dktapps

dktapps:
Workaround would be preferred as I said before

remminiscent:
i cant think of any right now, if someone else has an idea, please give it

dktapps:
I worked this over some locally. I've pushed my solution in https://github.com/pmmp/PocketMine-MP/pull/6866/commits/30ac82bd31a35ce12c2c3e804cc1c69e246bb218. Let me know what you think.

It's not ideal, but for the scope of PM5 I don't see a better option. I'd had plans to rework EffectInstance in PM6 anyway

remminiscent:
yep, looks great for PM5.👍

though a rework for the effect system would be great too, but we'll leave that for PM6

dktapps (dismissed):
For PM5 I think this will do. I plan on reworking effect duration handling in PM6 anyway (see #3897) so any hacks won't survive too long.

src/entity/effect/EffectInstance.php
```diff
@@ -76,11 +77,22 @@ public function setDuration(int $duration) : EffectInstance{
...
 	/**
 	 * Decreases the duration by the given number of ticks, without dropping below zero.
+	 * Note: Infinite effects will still have their durations updated, but they'll wrap around to Limits::INT32_MAX
+	 * if they drop below zero. This is because periodic effects rely on the duration ticker to know when to tick.
 	 *
 	 * @return $this
 	 */
 	public function decreaseDuration(int $ticks) : EffectInstance{
-		$this->duration = max(0, $this->duration - $ticks);
+		//Leave the flag set during subtraction - this prevents the
```
Dasciam: It seems this comment is not finished

dktapps: Actually I meant to delete it, must've missed a line

dktapps (approved):
I think this is as good as it gets in PM5. In PM6 it's planned to do some breaks to effect durations anyway, so the hacks will be cleaned up then.

In the meantime plugins can start preparing for that by making use of #6881

ShockedPlot7560 (approved):
LGTM for pm5

Issue #6024, opened 2023-08-28
Player gets buried in a block when sneaking + jumping.


ghost:
### Issue description
When player is sneaking, gets buried in a block on the head.
[>>Movie<<](https://twitter.com/nnse276/status/1696184482829885695?s=20)

### Steps to reproduce the issue
1. Stand on below the block.
2. Jump & Sneak

### OS and versions
* PocketMine-MP: Server version: 5.4.3 (git hash: e852a4382116705d7558fd808f1b388666b0c0cb)
* PHP: 8.1.22
* Server OS:
* Game version: All

### Plugins
Plugins (2): DEVirion v1.3.0, DevTools v1.17.0+dev

dktapps:
Reproduced, but I don't see how this is going to be fixed.

The problem is that you're releasing the jump key when your AABB is still 1.5 blocks high, but the top of your AABB is touching the block above. When you stop sneaking, your AABB returns to 1.8 blocks, causing an intersection by 0.3 blocks for as long as it takes to fall back to the ground.

I don't see an obvious solution for this problem.

PR #6544, opened 2024-11-29
Use PlayerAuthInputPacket::SNEAKING flag to test for sneaking

Muqsit:
`PlayerAuthInputPacket::SNEAKING` flag covers the case where a player is sneaking during flight, something which `START_SNEAKING` & `STOP_SNEAKING` currently do not.

### Related issues & PRs

* Fixes #5792 - sneaking during flight would cause ghost block to appear when placing block against a chest
* Fixes #5903 - players could not place blocks when within the 1.5 block space (i.e., under a slab)

### Behavioural changes

The server now correctly recognizes players who are sneaking while flying. Players are once again able to place blocks against interactable blocks (like chests and item frames) while in this state (without triggering block interaction). When under a slab (1.5 block space), players can place blocks and must explicitly sneak (on PC, by pressing shift key) to perform the usual sneak functionality.

## Tests

```php
// Debug chat message to test whether PlayerToggleSneakEvent is triggered as expected
// (i.e., no duplicate event calls, sneak event is initiated and terminated at proper times)
$plugin->getServer()->getPluginManager()->registerEvent(PlayerToggleSneakEvent::class, function(PlayerToggleSneakEvent $event) : void{
	$event->getPlayer()->sendMessage(gmdate("Y-m-d H:i:s.v") . ": sneaking=" . ($event->isSneaking() ? "yes" : "no"));
}, EventPriority::MONITOR, $plugin);
```
https://github.com/user-attachments/assets/dd4ec2fb-6785-4dbd-8f86-ad215943a406

https://github.com/user-attachments/assets/309a87ea-0503-44f2-b668-d7e70827fb28

Issue #6548, opened 2024-12-01
Separation of block interact disabling vs actual sneak/crouch


dktapps:
## Problem description
There are currently two functions of player sneaking:

1) to make the player appear to crouch & reduce their hitbox size
2) to disable block interactions when right-clicking on stuff

These two things are no longer coupled to the same thing. There are now 4 states for this:

| State | Crouching | Disable block interaction | Note |
|:------|:----------|:--------------------------|:-----|
| Walking | `false` | `false` | | 
| Pressing shift while flying | `false` | `true` | #5792 |
| Being in a 1.5 block space but not pressing shift | `true` | `false` | #5903 |
| Pressing shift while walking | `true` | `true` | |

## Proposed solution
There needs to be a separation of crouching from the player's desire to disable block interactions.

Unfortunately, this is difficult to do cleanly, as `isSneaking` is a `Living` function, meaning that it should refer to crouching, but the most common usages of sneaking in plugins will probably be to disable block interactions.

It's also difficult to choose a name for such a property. `Player->disableBlockInteractions()` is the best I can come up with, but I don't like it much.

Relates to #6544

dktapps:
Fixed by #6867

PR #6867, opened 2025-10-26
Separate sneak keydown flag from crouching flag

### LINKED ISSUE #6066: Crawling or swimming animation show up while jumping and sneaking at the same time, at a height of 2 blocks
### Issue description
Player animation showing crawling or swimming while attempting to jump while sneaking in a small space two blocks high
- Expected result: No animation
- Actual result: Animation show up for a split of a second then gone.

1. Build up an upside down L
2. Sneak under the roof of that L, stay close to the wall
3. Jumping while sneaking
image

https://github.com/pmmp/PocketMine-MP/assets/54394881/1a7edff4-8201-4617-a4f3-ffb44a5e02d5

https://github.com/pmmp/PocketMine-MP/assets/54394881/60f0737a-f894-4b73-861d-212d6e610736

### OS and versions

* PocketMine-MP:
* PHP: 8.1
* Server OS: Win10
* Game version: Android/iOS/Win10/Xbox/PS4/Switch (delete as appropriate)

### Plugins

Plugins (2): DevTools v1.17.1, Test v0.0.1
Test does not change any things related to movements and entities

### LINKED ISSUE #6733: Phase Glitch
### Plugin information

Bug happens without plugins

### Problem description

- Steps to reproduce the issue: i do not completely know the actual way to reproduce it. but place a 1x1x2 square, with trapdoor or slab on top, and u can phase through. by spamming w key and space and shift. i also checked vanilla mc. vanilla mc doesnt even let u crouch while in the air like that

- Error backtraces: no error
- Crashdumps: no crashes
Plugin code that triggers the issue: none
List of installed plugins (use /plugins): none

Image
https://streamable.com/pbi5fr

i tried w slabs, couldnt phase

### Expected behaviour

not allow u to phase through the blocks

### PocketMine-MP version

5.28.2

### PHP version

8.2.27

### Server OS

Linux

Windows 10, possibly others too

Dasciam:
### Related issues & PRs
fixes #6066
fixes #6733

### API changes

- `Player` now has `toggleSneakBindPressed` method to toggle sneak bind state
- `Player` now has `isSneakBindPressed` method to get toggle sneak bind state

### Behavioural changes

To determine whether player is actually sneaking for item/block interactions, BDS uses `PlayerAuthInputFlags::SNEAKING` , for player sneaking state only `PlayerAuthInputFlags::START_SNEAKING` and `PlayerAuthInputFlags::STOP_SNEAKING`.

This PR reverts #6544 changes due to incorrect flag usage

## Backwards compatibility

I don't know whether it can entail some BC break but it changes the `use item on` logic in `World` class

## Tests

First test: press crouch bind and release when you're forced to crouch
Be sure that you're not experiencing #6066 and #6733

https://github.com/user-attachments/assets/969bb4d1-ab5e-47a6-942b-c888d9171e88

Second test: press crouch bind while flying and place some interactable blocks
Be sure that you're experiencing #6544

https://github.com/user-attachments/assets/47b7dc71-0e01-407d-9fae-178d14e80a03

Third last test: carry some item in hand, press crouch bind and release when you're forced to crouch
- be sure that you can interact with door in that state
- press crouch bind
- you shouldn't be able to open and close the door
- select free slot
- you can interact with the door again (open and close)

https://github.com/user-attachments/assets/0f83189a-e2d6-46fb-acc4-68952a8178cd

### DIFF
--- a/src/event/player/PlayerToggleSneakEvent.php
+++ b/src/event/player/PlayerToggleSneakEvent.php
@@ -47,7 +47,8 @@ class PlayerToggleSneakEvent extends PlayerEvent implements Cancellable{
 	 * isSneakPressed is different from isSneaking because is responsible for interaction limitation and don't change visual player state.
 	 * isSneaking is responsible for visual player state (hitbox height, sneaking).
 	 *
-	 * Returns true when the player is actually holding the sneak key.
+	 * Returns whether the player is pressing the sneak key.
+	 * The player may still be sneaking even if this is false due to gameplay mechanics (e.g. releasing sneak while in a 1.5 block high space).
 	 */
 	public function isSneakPressed() : bool{
 		return $this->isSneakPressed;
--- a/src/event/player/PlayerToggleSneakEvent.php
+++ b/src/event/player/PlayerToggleSneakEvent.php
@@ -43,10 +43,6 @@ class PlayerToggleSneakEvent extends PlayerEvent implements Cancellable{
 	}
 
 	/**
-	 * Since 1.20.10, sneaking has multiple states that affects interacting with items/blocks.
-	 * isSneakPressed is different from isSneaking because is responsible for interaction limitation and don't change visual player state.
-	 * isSneaking is responsible for visual player state (hitbox height, sneaking).
-	 *
 	 * Returns whether the player is pressing the sneak key.
 	 * The player may still be sneaking even if this is false due to gameplay mechanics (e.g. releasing sneak while in a 1.5 block high space).
 	 */
--- a/src/event/player/PlayerToggleSneakEvent.php
+++ b/src/event/player/PlayerToggleSneakEvent.php
@@ -32,12 +32,17 @@ class PlayerToggleSneakEvent extends PlayerEvent implements Cancellable{
 
 	public function __construct(
 		Player $player,
-		protected bool $isSneaking
+		protected bool $isSneaking,
+		protected bool $isSneakBindPressed,
 	){
 		$this->player = $player;
 	}
 
 	public function isSneaking() : bool{
 		return $this->isSneaking;
+	}
+
+	public function isSneakBindPressed() : bool{
+		return $this->isSneakBindPressed;
 	}
 }
--- a/src/event/player/PlayerToggleSneakEvent.php
+++ b/src/event/player/PlayerToggleSneakEvent.php
@@ -33,7 +33,7 @@ class PlayerToggleSneakEvent extends PlayerEvent implements Cancellable{
 	public function __construct(
 		Player $player,
 		protected bool $isSneaking,
-		protected bool $isSneakBindPressed,
+		protected bool $isSneakBindPressed
 	){
 		$this->player = $player;
 	}
--- a/src/event/player/PlayerToggleSneakEvent.php
+++ b/src/event/player/PlayerToggleSneakEvent.php
@@ -33,7 +33,7 @@ class PlayerToggleSneakEvent extends PlayerEvent implements Cancellable{
 	public function __construct(
 		Player $player,
 		protected bool $isSneaking,
-		protected bool $isSneakBindPressed
+		protected bool $isSneakPressed
 	){
 		$this->player = $player;
 	}
@@ -42,7 +42,7 @@ class PlayerToggleSneakEvent extends PlayerEvent implements Cancellable{
 		return $this->isSneaking;
 	}
 
-	public function isSneakBindPressed() : bool{
-		return $this->isSneakBindPressed;
+	public function isSneakPressed() : bool{
+		return $this->isSneakPressed;
 	}
 }
--- a/src/event/player/PlayerToggleSneakEvent.php
+++ b/src/event/player/PlayerToggleSneakEvent.php
@@ -42,6 +42,13 @@ class PlayerToggleSneakEvent extends PlayerEvent implements Cancellable{
 		return $this->isSneaking;
 	}
 
+	/**
+	 * Since 1.20.10, sneaking has multiple states that affects interacting with items/blocks.
+	 * isSneakPressed is different from isSneaking because is responsible for interaction limitation and don't change visual player state.
+	 * isSneaking is responsible for visual player state (hitbox height, sneaking).
+	 *
+	 * Returns true when the player is actually holding the sneak key.
+	 */
 	public function isSneakPressed() : bool{
 		return $this->isSneakPressed;
 	}
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -1285,6 +1285,13 @@ class Player extends Human implements CommandSender, ChunkListener, IPlayer, Nev
 		$this->sneakPressed = $sneakPressed;
 	}
 
+	/**
+	 * Since 1.20.10, sneaking has multiple states that affects interacting with items/blocks.
+	 * isSneakPressed is different from isSneaking because is responsible for interaction limitation and don't change visual player state.
+	 * isSneaking is responsible for visual player state (hitbox height, sneaking).
+	 *
+	 * Returns true when the player is actually holding the sneak key.
+	 */
 	public function isSneakPressed() : bool{
 		return $this->sneakPressed;
 	}
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -294,7 +294,7 @@ class Player extends Human implements CommandSender, ChunkListener, IPlayer, Nev
 
 	protected float $flightSpeedMultiplier = self::DEFAULT_FLIGHT_SPEED_MULTIPLIER;
 
-	/** @phpstan-var positive-int|null  */
+	/** @phpstan-var positive-int|null */
 	protected ?int $lineHeight = null;
 	protected string $locale = "en_US";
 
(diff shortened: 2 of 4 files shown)


dktapps:
Probably worth reading https://github.com/pmmp/PocketMine-MP/issues/6548. The reason this wasn't already changed is because of the behavioural BC break of `isSneaking()`, since it wouldn't refer to pressing shift/disabling block interaction anymore, but rather if you were in a forced crouch state.

There are some other sneak-related issues that this might resolve.

dktapps:
Also, there's probably other places where `isSneaking` would need to be changed

Dasciam:
Perhaps this **protocol crap** should be removed from the player class?

I also doubt that `isSneaking` should be changed anywhere else in my opinion

I will definitely look for issues related to the sneaking state

dktapps:
I think having it in the API is fine. The other option is that we'd have to pass the flag everywhere that needs it, and there's probably other things that do need it - e.g. events

The only thing I'm bothered about is that isSneaking changes meaning from "is pressing shift || is crouching" to just "is crouching". I don't like making behaviour changes like this.

That being said, there's a case to be made that plugins using isSneaking the old way were broken anyway from a player perspective, I suppose, so maybe people can won't mind

Dasciam:
Can `PlayerToggleSneakEvent` have two values: `sneaking` and `sneakBindPressed`?

Also, maybe these changes could be in `major-next` instead of `minor-next`

Dasciam:
So, `Player::toggleSneak` will accept next values: `sneaking` and `sneakBindPressed`

dktapps:
It could go in major-next, but major-next won't be released for a while yet. Probably this is fine for minor-next, but I'll probably have to get some opinions from the team.

Not sure about `PlayerToggleSneakEvent` since it might confuse existing plugins. e.g. it could be called without `isSneaking()` changing at all if the player presses shift+space while flying in creative. Or it would be called without `isSneakBindPressed()` changing when popping out from underneath a 1.5 block space. But even just leaving it bound to `isSneaking()` alone already probably causes some disruption, since you can't really "toggle" `isSneaking()` that way anymore. Pretty confusing

Dasciam:
Got you, a very hard thing to be honest

probably this PR also closes #6024

dktapps:
Ok, I've had a think about this.

So, up to this point `isSneaking` and `PlayerToggleSneakEvent` have always reflected whether you were crouching, irrespective of the reason why. I think we should try to preserve this behaviour.
Yes, some plugins have probably used this to check whether the player was attempting to disable block-use, but perhaps it's not our responsibility to fix that. Such plugins would behave basically the same anyway - the only time they'd behave different would be when a player presses shift+space at the same time while flying (disabling use-block, but not crouching).

Possibly what we could do for `PlayerToggleSneakEvent` is to call it with `cancelled=true`, if the use-block input changes but the crouching state doesn't change. There are 2 cases I can think of that we'd need to beware of:

1) `pressSneak = false, crouching = unchanged` - if the player is under a slab (use-block changes, but doesn't stop crouching)
2) `pressSneak = true, crouching = unchanged` - if the player presses space followed by shift in creative (use-block changes, but doesn't start crouching)

dktapps:
> Got you, a very hard thing to be honest
> 
> probably this PR also closes #6024

Yeah, for sure.
There's probably other sneak related issues that this probably fixes too.

Dasciam:
Alright

Dasciam:
I think it would be fine :smile_cat:

src/player/Player.php
```diff
@@ -2075,11 +2080,17 @@ public function toggleSprint(bool $sprint) : bool{
 		return true;
 	}
 
-	public function toggleSneak(bool $sneak) : bool{
-		if($sneak === $this->sneaking){
+	public function toggleSneak(bool $sneak, bool $sneakBindPressed) : bool{
```
dktapps: This is a BC break

src/player/Player.php
```diff
@@ -2075,11 +2080,17 @@ public function toggleSprint(bool $sprint) : bool{
...
+		if($sneak === $this->sneaking && $sneakBindPressed === $this->sneakBindPressed){
 			return true;
 		}
-		$ev = new PlayerToggleSneakEvent($this, $sneak);
+
+		$ev = new PlayerToggleSneakEvent($this, $sneak, $sneakBindPressed);
+		if($sneak === $this->sneaking) {
+			$ev->cancel();
+		}
+		$this->sneakBindPressed = $sneakBindPressed;
```
dktapps: Hmm. I suppose it's impossible to prevent someone from enabling this if done this way.

dktapps (commented):
I don't love the `sneak bind pressed` terminology. I don't think it's very clear what it means.

`sneak pressed` would probably be fine, or possibly `use block disabled`.

src/player/Player.php
Dasciam: We can't prevent player's intention

src/player/Player.php
Dasciam: Alright

src/player/Player.php
dktapps: Well, we do allow cancelling all the other inputs (sneak, glide, swim etc).

I suppose there are easier ways to control this from an API perspective (e.g. `PlayerInteractEvent->setUseBlock()`). But it looks weird to do this directly before calling the event.

Dasciam: I don't think this should be exposed to the developer, because this is pure player intention to hold sneak button and we don't want to modify this and we shouldn't

also, sneak, swim cancelling is making gameplay more buggy because the game haven't exposed any clear versatility with it

Dasciam: But why not okay got it

Dasciam: by versatility I mean we can cancel sneak but the client will try to sneak again and again till you release sneak button and not accepting new player state. Same with swimming

src/player/Player.php
```diff
@@ -2075,12 +2084,18 @@ public function toggleSprint(bool $sprint) : bool{
...
 		}
-		$ev = new PlayerToggleSneakEvent($this, $sneak);
+
+		$ev = new PlayerToggleSneakEvent($this, $sneak, $sneakPressed);
+		if($sneak === $this->sneaking) {
+			$ev->cancel();
+		}
 		$ev->call();
+
+		$this->setSneakPressed($sneakPressed);
```
dktapps: Better to put this before any event logic. Since it's not affected by the event.

src/player/Player.php
```diff
@@ -1280,6 +1281,14 @@ public function isSpectator() : bool{
 		return $this->gamemode === GameMode::SPECTATOR;
 	}
 
+	public function setSneakPressed(bool $sneakPressed) : void{
+		$this->sneakPressed = $sneakPressed;
+	}
+
+	public function isSneakPressed() : bool{
```
dktapps, line 1292: This should be documented to explain how it's different to `isSneaking()`

src/event/player/PlayerToggleSneakEvent.php
```diff
@@ -32,12 +32,17 @@ class PlayerToggleSneakEvent extends PlayerEvent implements Cancellable{
...
+		protected bool $isSneakPressed
 	){
 		$this->player = $player;
 	}
 
 	public function isSneaking() : bool{
 		return $this->isSneaking;
 	}
+
+	public function isSneakPressed() : bool{
```
dktapps, line 49: This should be documented to explain how it differs from isSneaking()

src/event/player/PlayerToggleSneakEvent.php
```diff
@@ -32,12 +32,24 @@ class PlayerToggleSneakEvent extends PlayerEvent implements Cancellable{
...
 	public function isSneaking() : bool{
 		return $this->isSneaking;
 	}
+
+	/**
+	 * Since 1.20.10, sneaking has multiple states that affects interacting with items/blocks.
+	 * isSneakPressed is different from isSneaking because is responsible for interaction limitation and don't change visual player state.
+	 * isSneaking is responsible for visual player state (hitbox height, sneaking).
+	 *
+	 * Returns true when the player is actually holding the sneak key.
```
dktapps: It'd be simpler to say something like

> Returns whether the player is pressing the sneak key.
> The player may still be sneaking even if this is false due to gameplay mechanics (e.g. releasing sneak while in a 1.5 block high space).

src/player/Player.php
```diff
@@ -1280,6 +1281,21 @@ public function isSpectator() : bool{
...
+	}
+
+	/**
+	 * Since 1.20.10, sneaking has multiple states that affects interacting with items/blocks.
+	 * isSneakPressed is different from isSneaking because is responsible for interaction limitation and don't change visual player state.
+	 * isSneaking is responsible for visual player state (hitbox height, sneaking).
+	 *
+	 * Returns true when the player is actually holding the sneak key.
+	 */
+	public function isSneakPressed() : bool{
```
dktapps, line 1292: As above

dktapps (dismissed):

src/player/Player.php
```diff
@@ -2075,12 +2088,18 @@ public function toggleSprint(bool $sprint) : bool{
...
-		if($sneak === $this->sneaking){
+	public function toggleSneak(bool $sneak, bool $sneakPressed = true) : bool{
+		if($sneak === $this->sneaking && $sneakPressed === $this->sneakPressed){
 			return true;
 		}
-		$ev = new PlayerToggleSneakEvent($this, $sneak);
+		$this->setSneakPressed($sneakPressed);
+
+		$ev = new PlayerToggleSneakEvent($this, $sneak, $sneakPressed);
+		if($sneak === $this->sneaking) {
```
```suggestion
		if($sneak === $this->sneaking){
```

dktapps (commented):
Looks basically fine now

dktapps (approved):

ipad54 (approved):

PR #6477, opened 2024-10-27
Weather System Implementation 

AzliRynz:
## Introduction
Adding `Weather System`

### API changes
Added `WeatherCommand`
Added permission `pocketmine.command.weather`
Added file in src/world/weather
* WeatherManager
* WeatherType

Added method in src/Server.php
* toggleWeather

Added method in src/world/World.php `getWeatherManager`
Requires translations:

## Tests

I tested this PR by doing the following (tick all that apply):
- [ ] Writing PHPUnit tests (commit these in the `tests/phpunit` folder)
- [ ] Playtesting using a Minecraft client (provide screenshots or a video)
- [ ] Writing a test plugin (provide the code and sample output)
- [ ] Other (provide details)

PR #6870, opened 2025-10-28
feat: implement weather system

GroXzaLYs:
Adds a basic dynamic weather system with automatic transitions between clear weather, rain, and thunderstorms.

Rain, thunder, and lightning are handled server-side and synchronized for all players.

---

### Related issues & PRs
* #6477 

---

### Changes
- Added `Lightning` weather entity
- Added weather-related methods to `World`
- Added `/weather` command

---

### Compatibility

No breaking changes.  
Worlds without weather support are unaffected.

---

### Testing

Manual testing:
- `/weather rain` and `/weather thunder` work correctly
- Thunder sounds and lightning visuals are synced
- No crashes or performance issues observed

https://github.com/user-attachments/assets/176103e9-9572-41e7-bc9c-a5817e5e68e2


resources/translations/eng.ini
```diff
@@ -515,6 +525,7 @@ pocketmine.permission.command.transferserver=Allows the user to transfer self to
 pocketmine.permission.command.unban.ip=Allows the user to unban IP addresses
 pocketmine.permission.command.unban.player=Allows the user to unban players
 pocketmine.permission.command.version=Allows the user to view the version of the server
+pocketmine.permission.command.weather=Hehels
```
dktapps: This doesn't look right

resources/translations/eng.ini
```diff
@@ -98,6 +98,16 @@ commands.unban.usage=/pardon <name>
 commands.unbanip.invalid=You have entered an invalid IP address
 commands.unbanip.success=Unbanned IP address {%0}
 commands.unbanip.usage=/pardon-ip <address>
+pocketmine.permission.command.weather=Allows the player to change the weather
+commands.weather.clear=Changing to clear weather
+commands.weather.description=Sets the weather.
```
dktapps: Command descriptions don't normally have a . at the end

src/command/defaults/WeatherCommand.php
```diff
@@ -0,0 +1,99 @@
...
+use function count;
+use function max;
+use function strtolower;
+
+class WeatherCommand extends VanillaCommand{
+
+	public function __construct(){
+		parent::__construct(
+			"weather",
+			new Translatable("commands.weather.description"),
```
dktapps: This should use `KnownTranslationFactory`

src/command/defaults/WeatherCommand.php
```diff
@@ -0,0 +1,99 @@
...
+			"/weather <clear|rain|thunder> [duration]"
+		);
+		$this->setPermission(DefaultPermissionNames::COMMAND_WEATHER);
+	}
+
+	public function execute(CommandSender $sender, string $label, array $args) : void{
+		if(!$sender instanceof Player){
+			$sender->sendMessage(new Translatable("commands.generic.ingame"));
```
dktapps: As above

src/command/defaults/WeatherCommand.php
```diff
@@ -0,0 +1,99 @@
...
+			return;
+		}
+
+		if(!$this->testPermission($sender)){
```
dktapps: This is not required since 5.0

src/command/defaults/WeatherCommand.php
```diff
@@ -0,0 +1,99 @@
...
+		$world = $sender->getWorld();
+		$weatherManager = $world->getWeatherManager();
+
+		if(count($args) < 1){
+			/** @var WeatherType $current */
+			$current = $weatherManager->getWeather();
+			$stateName = match($current){
+				WeatherType::CLEAR => new Translatable("commands.weather.query.clear"),
+				WeatherType::RAIN => new Translatable("commands.weather.query.rain"),
+				WeatherType::THUNDER => new Translatable("commands.weather.query.thunder"),
```
dktapps: These should use `KnownTranslationFactory`

src/command/defaults/WeatherCommand.php
```diff
@@ -0,0 +1,99 @@
...
+
+		if(!$this->testPermission($sender)){
+			return;
+		}
+
+		$world = $sender->getWorld();
+		$weatherManager = $world->getWeatherManager();
+
+		if(count($args) < 1){
+			/** @var WeatherType $current */
```
dktapps: This doc comment should not be necessary if `getWeather()` has the proper types

src/command/defaults/WeatherCommand.php
```diff
@@ -0,0 +1,99 @@
...
+			$current = $weatherManager->getWeather();
+			$stateName = match($current){
+				WeatherType::CLEAR => new Translatable("commands.weather.query.clear"),
+				WeatherType::RAIN => new Translatable("commands.weather.query.rain"),
+				WeatherType::THUNDER => new Translatable("commands.weather.query.thunder"),
+			};
+
+			$sender->sendMessage(new Translatable("commands.weather.query", [$stateName]));
```
dktapps: This should use `KnownTranslationFactory`

src/command/defaults/WeatherCommand.php
```diff
@@ -0,0 +1,99 @@
...
+
+		$type = match(strtolower($args[0])){
+			"clear" => WeatherType::CLEAR,
+			"rain" => WeatherType::RAIN,
+			"thunder" => WeatherType::THUNDER,
+			default => null
+		};
+
+		if($type === null){
+			$sender->sendMessage("Invalid weather type. Use: clear, rain, or thunder.");
```
dktapps: This should throw `InvalidCommandSyntaxException` instead of sending a hardcoded message

src/command/defaults/WeatherCommand.php
```diff
@@ -0,0 +1,99 @@
...
+			return;
+		}
+
+		$duration = isset($args[1]) ? max(100, (int) $args[1]) : 6000;
+		$weatherManager->setWeather($type, $duration);
+
+		switch($type){
+			case WeatherType::CLEAR:
+				$sender->sendMessage(new Translatable("commands.weather.clear"));
```
dktapps: This should use `KnownTranslationFactory`

src/command/defaults/WeatherCommand.php
```diff
@@ -0,0 +1,99 @@
...
+				break;
+			case WeatherType::RAIN:
+				$sender->sendMessage(new Translatable("commands.weather.rain"));
```
dktapps: As above

src/command/defaults/WeatherCommand.php
```diff
@@ -0,0 +1,99 @@
...
+				break;
+			case WeatherType::THUNDER:
+				$sender->sendMessage(new Translatable("commands.weather.thunder"));
```
dktapps: As above

src/world/WeatherManager.php
```diff
@@ -0,0 +1,196 @@
...
+					break;
+
+				default:
+					$rainEvent = LevelEventPacket::create(LevelEvent::STOP_RAIN, 0, $pos);
+					$thunderEvent = LevelEventPacket::create(LevelEvent::STOP_THUNDER, 0, $pos);
+					break;
+			}
+
+			$player->getNetworkSession()->sendDataPacket($rainEvent);
+			$player->getNetworkSession()->sendDataPacket($thunderEvent);
```
dktapps: This can use `NetworkBroadcastUtils::broadcastPackets()` instead of sending the packets one by one

src/world/WeatherManager.php
```diff
@@ -0,0 +1,196 @@
...
+			$this->world->setThunderLevel(1.0);
+		}else{
+			$this->world->setRainLevel(0.0);
+			$this->world->setThunderLevel(0.0);
+		}
+	}
+
+	private function sendWeatherPackets() : void{
+		foreach($this->world->getPlayers() as $player){
+			$pos = $player->getPosition();
```
dktapps: This position shouldn't be needed. Weather is a positionless event, so you should be able to pass `null`.

src/world/WeatherManager.php
```diff
@@ -0,0 +1,196 @@
...
+		if(count($players) === 0){
+			return;
+		}
+
+		$eid = Entity::nextRuntimeId();
+
+		$add = AddActorPacket::create(
+			$eid,
+			$eid,
+			"minecraft:lightning_bolt",
```
dktapps: This should use `EntityIds` constants

src/world/WeatherManager.php
```diff
@@ -0,0 +1,196 @@
...
+			0,
+			0.0,
+			0.0,
+			[],
+			[],
+			new PropertySyncData([], []),
+			[]
+		);
+
+		$sound = PlaySoundPacket::create("ambient.weather.thunder", (float) $pos->x, (float) $pos->y, (float) $pos->z, 1.0, 1.0);
```
dktapps: This should probably be using `LevelSoundEventPacket`. Also, a proper `Sound` class should be created for this.

src/world/WeatherManager.php
```diff
@@ -0,0 +1,196 @@
...
+
+		foreach($players as $p){
+			$p->getNetworkSession()->sendDataPacket(LevelEventPacket::create(LevelEvent::START_THUNDER, 250000, $p->getPosition()));
```
dktapps: This should probably be positionless as mentioned above

src/world/weather/Lightning.php
```diff
@@ -0,0 +1,84 @@
...
+use pocketmine\math\AxisAlignedBB;
+use pocketmine\math\Vector3;
+use pocketmine\network\mcpe\NetworkBroadcastUtils;
+use pocketmine\network\mcpe\protocol\AddActorPacket;
+use pocketmine\network\mcpe\protocol\PlaySoundPacket;
+use pocketmine\network\mcpe\protocol\types\entity\PropertySyncData;
+use pocketmine\world\World;
+use function count;
+
+final class Lightning {
```
dktapps: This should be a proper entity

src/world/weather/Lightning.php
```diff
@@ -0,0 +1,84 @@
...
+		if($event->isCancelled()){
+			return;
+		}
+
+		$players = $this->world->getPlayers();
+		if(count($players) === 0) return;
+
+		$eid = Entity::nextRuntimeId();
+		$add = AddActorPacket::create($eid, $eid, "minecraft:lightning_bolt", $this->pos, null, 0, 0, 0.0, 0.0, [], [], new PropertySyncData([], []), []);
+		$sound = PlaySoundPacket::create("ambient.weather.thunder", (float) $this->pos->x, (float) $this->pos->y, (float) $this->pos->z, 1.0, 1.0);
```
dktapps: Same as above wrt `LevelSoundEventPacket`

src/world/weather/Lightning.php
```diff
@@ -0,0 +1,84 @@
...
+		$this->pos = $pos;
+
+		$event = new LightningStrikeEvent($world, $pos);
+		$event->call();
+		if($event->isCancelled()){
+			return;
+		}
+
+		$players = $this->world->getPlayers();
+		if(count($players) === 0) return;
```
dktapps: This style of `if` is frowned upon, always use braces

src/world/World.php
```diff
@@ -1752,6 +1764,22 @@ public function getSkyLightReduction() : int{
 		return $this->skyLightReduction;
 	}
 
+  public function getRainLevel() : float{
+	return $this->rainLevel;
+  }
+
+  public function setRainLevel(float $level) : void{
+	$this->rainLevel = max(0.0, min(1.0, $level));
```
dktapps: This should throw if the values are out of bounds instead of clamping

src/world/World.php
```diff
@@ -1752,6 +1764,22 @@ public function getSkyLightReduction() : int{
...
+  }
+
+  public function getThunderLevel() : float{
+	return $this->thunderLevel;
+  }
+
+  public function setThunderLevel(float $level) : void{
+	$this->thunderLevel = max(0.0, min(1.0, $level));
```
dktapps: As above

src/command/defaults/WeatherCommand.php
```diff
@@ -0,0 +1,99 @@
...
+				break;
+		}
```
dktapps, line 107: This switch should be converted to a `match`

src/world/WeatherType.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\world;
+
+enum WeatherType : int{
+	case CLEAR = 0;
+	case RAIN = 1;
+	case THUNDER = 2;
```
dktapps: Not sure why this is a backed enum. You aren't using the values anywhere.

src/world/weather/Lightning.php
```diff
@@ -0,0 +1,84 @@
...
+		  $pos->x - 3,
+		  $pos->y - 3,
+		  $pos->z - 3,
+		  $pos->x + 3,
+		  $pos->y + 3,
+		  $pos->z + 3
+		);
+
+		foreach($this->world->getNearbyEntities($aabb) as $entity){
+			if($entity instanceof Entity){
```
dktapps: This is always true

src/world/WeatherManager.php
```diff
@@ -0,0 +1,196 @@
...
+use pocketmine\network\mcpe\protocol\PlaySoundPacket;
+use pocketmine\network\mcpe\protocol\types\entity\PropertySyncData;
+use pocketmine\network\mcpe\protocol\types\LevelEvent;
+use pocketmine\world\weather\Lightning;
+use function array_rand;
+use function count;
+use function max;
+use function mt_rand;
+
+final class WeatherManager{
```
dktapps: I'm not really sure what the point of having this in a separate class is, since it's coupled to `World` anyway

dktapps (changes requested):

src/entity/LightningBolt.php
```diff
@@ -0,0 +1,85 @@
...
+					0.0,
+					0.0,
+					0.0,
+					[],
+					[],
+					new PropertySyncData([], []),
+					[]
+				),
+			]
+		);
```
dktapps: `spawnToAll` should not be overridden. I don't see why you need to create this packet manually anyway.

src/entity/LightningBolt.php
```diff
@@ -0,0 +1,85 @@
...
+
+		$this->getWorld()->addSound($this->location, new ThunderSound());
```
dktapps: This could be sent with a `sendSpawnPacket` override, or more likely `onFirstUpdate()`

src/entity/LightningBolt.php
```diff
@@ -0,0 +1,85 @@
...
+
+	public function getName() : string{
+		return "Lightning Bolt";
+	}
+
+	protected function initEntity(CompoundTag $nbt) : void{
+	  $this->creationTick = $this->getWorld()->getServer()->getTick();
+		parent::initEntity($nbt);
+		$this->setNameTagVisible(false);
+		$this->setSilent();
```
dktapps: Why is this needed?

src/entity/LightningBolt.php
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	public function onUpdate(int $currentTick) : bool{
```
dktapps: This is not supposed to be overridden (use `entityBaseTick`)

src/entity/LightningBolt.php
```diff
@@ -0,0 +1,85 @@
...
+		if($currentTick - $this->creationTick > 20){
```
dktapps: This can use `ticksLived` if you don't override `onUpdate()`

src/world/World.php
```diff
@@ -442,7 +454,7 @@ public static function getBlockXYZ(int $hash, ?int &$x, ?int &$y, ?int &$z) : vo
 		$extraZ = ((($baseY >> self::BLOCKHASH_Z_SHIFT) & self::BLOCKHASH_XZ_EXTRA_MASK) << self::MORTON3D_BIT_SIZE);
 
 		$x = (($baseX & self::BLOCKHASH_XZ_MASK) | $extraX) << self::BLOCKHASH_XZ_SIGN_SHIFT >> self::BLOCKHASH_XZ_SIGN_SHIFT;
-		$y = ($baseY & self::BLOCKHASH_Y_MASK) - self::BLOCKHASH_Y_OFFSET;
+		$y = ($baseY & self::BLOCKHASH_Y_MASK) - self::BLOCKHASH_Y_OFF;
```
dktapps: Why is this changed?

src/world/World.php
```diff
@@ -394,13 +399,20 @@ public static function chunkHash(int $x, int $z) : int{
...
 	private const BLOCKHASH_X_SHIFT = self::BLOCKHASH_Y_BITS;
 	private const BLOCKHASH_Z_SHIFT = self::BLOCKHASH_X_SHIFT + self::BLOCKHASH_XZ_EXTRA_BITS;
-	private WeatherManager $weatherManager;
+
+	private WeatherType $weather = WeatherType::CLEAR;
+	private int $weatherDuration = 0;
+	private int $weatherTick = 0;
+
+	/** @var LightningBolt[] */
+	private array $lightnings = [];
```
dktapps: Not sure what this is needed for

src/world/World.php
```diff
@@ -593,10 +604,120 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
 		}
 	}
 
-  public function getWeatherManager() : WeatherManager{
-	return $this->weatherManager;
+  /**
+   * @phpstan-return LightningBolt[]
+   */
+  public function getLightnings() : array{
```
dktapps: I don't think this is necessary. People can just filter the world's entities if they need this sort of functionality.

src/world/World.php
```diff
@@ -593,10 +604,120 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
...
+
+		$this->weather = $type;
+		$this->weatherDuration = $duration;
+		$this->weatherTick = 0;
+
+		$this->applyWeather();
+		$this->broadcastWeatherPackets();
+	}
+
+	public function tickWeather() : void{
```
dktapps: This should be private

src/world/World.php
```diff
@@ -593,10 +604,120 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
...
+					LevelEventPacket::create(LevelEvent::START_RAIN, 80000, $pos),
+					LevelEventPacket::create(LevelEvent::STOP_THUNDER, 0, $pos),
+				],
+				WeatherType::THUNDER => [
+					LevelEventPacket::create(LevelEvent::START_RAIN, 160000, $pos),
+					LevelEventPacket::create(LevelEvent::START_THUNDER, 200000, $pos),
+				],
+			};
+
+			NetworkBroadcastUtils::broadcastPackets([$player], [$rainEvent, $thunderEvent]);
```
dktapps: Uh, that's not what I meant. `broadcastPackets()` is useful only when you're sending packets to multiple players at the same time. But you'd need to null the positions to do that (which should be fine for weather events as I mentioned before).

src/world/World.php
```diff
@@ -593,10 +604,120 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
...
+	}
+
+	private function broadcastWeatherPackets() : void{
+		$players = $this->getPlayers();
+		if(count($players) === 0){
+			return;
+		}
+
+		foreach($players as $player){
+			$pos = $player->getPosition();
```
dktapps: As I said before these LevelEvents are positionless, this shouldn't be needed

src/world/World.php
```diff
@@ -593,10 +604,120 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
...
+					LevelEventPacket::create(LevelEvent::STOP_RAIN, 0, $pos),
+					LevelEventPacket::create(LevelEvent::STOP_THUNDER, 0, $pos),
+				],
+				WeatherType::RAIN => [
+					LevelEventPacket::create(LevelEvent::START_RAIN, 80000, $pos),
+					LevelEventPacket::create(LevelEvent::STOP_THUNDER, 0, $pos),
+				],
+				WeatherType::THUNDER => [
+					LevelEventPacket::create(LevelEvent::START_RAIN, 160000, $pos),
+					LevelEventPacket::create(LevelEvent::START_THUNDER, 200000, $pos),
```
dktapps: What do these magic numbers mean?

src/world/World.php
```diff
@@ -593,10 +604,120 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
...
+
+			NetworkBroadcastUtils::broadcastPackets([$player], [$rainEvent, $thunderEvent]);
+		}
+	}
+
+	public function spawnLightning(Vector3 $pos) : void{
+		$bolt = new LightningBolt(Location::fromObject($pos, $this));
+		$bolt->spawnToAll();
+		$this->lightnings[spl_object_id($bolt)] = $bolt;
+	}
```
dktapps, line 689: This shouldn't be needed, plugins that want to create lightning can do so as they would any other entity

src/entity/EntityFactory.php
```diff
@@ -195,6 +195,10 @@ public function __construct(){
 			'ThrownTrident' //as above
 		]);
 
+		$this->register(LightningBolt::class, function(World $world, CompoundTag $nbt) : LightningBolt{
+		  return new LightningBolt(Helper::parseLocation($nbt, $world), $nbt);
+		}, ['Lightning Bolt', 'minecraft:lightning_bolt']);
```
dktapps: Save IDs are not supposed to contain spaces. Check the wiki for the correct one.

The first should match Java 1.10, the second should match Bedrock.

dktapps: Though it may not be necessary to register this for a save ID at all (not sure if lightning is saved in the world in vanilla).

src/command/defaults/WeatherCommand.php
```diff
@@ -77,23 +77,16 @@ public function execute(CommandSender $sender, string $label, array $args) : voi
 		};
 
 		if($type === null){
-			$sender->sendMessage("Invalid weather type. Use: clear, rain, or thunder.");
-			return;
+		  throw new InvalidCommandSyntaxException();
```
dktapps: This could be thrown directly from the default case of the match

dktapps (changes requested):

resources/translations/eng.ini
```diff
@@ -98,6 +98,14 @@ commands.unban.usage=/pardon <name>
 commands.unbanip.invalid=You have entered an invalid IP address
 commands.unbanip.success=Unbanned IP address {%0}
 commands.unbanip.usage=/pardon-ip <address>
+commands.weather.clear=Changing to clear weather
+commands.weather.disabled=Weather Cycle isn't enabled in this world.
+commands.weather.query=Weather state is: %s
```
dktapps: This is not the correct parameter format for PM, you can't just copy paste directly from vanilla language files

dktapps (changes requested):
Also I don't know what editor you're using but it's made an enormous mess of your indentation. Indents should use tabs.

dktapps:
Overall this is looking better since the last time I reviewed, but there's still some work left to do. Thanks for sticking with it 🙂

src/entity/EntityFactory.php
Dasciam: Lightnings are saved in the world, but disposes fast. I don't think we would want lightning saving in the world only because vanilla does

dktapps: Yeah, but if it's not registered, the server will generate errors when it sees a lightning bolt in a save from vanilla

Dasciam: It's a very rare case

src/world/World.php
Dasciam: I think they are from BDS and they may mean weather intensity

I think it would be good if weather API have some sort of `Weather strength`

dktapps: Yeah, that makes sense

src/world/World.php
```diff
@@ -588,6 +604,120 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
...
+	private function broadcastWeatherPackets() : void{
+		$players = $this->getPlayers();
+		if(count($players) === 0){
+			return;
+		}
+
+		foreach($players as $player){
+			$pos = $player->getPosition();
+
+			[$rainEvent, $thunderEvent] = match($this->weather){
```
Dasciam: It seems `rainLevel` and `thunderLevel` are unused right here.

It's not clear why we're inventing `WeatherType` if there still rain level and thunder level values

There can be two ways to resolve:
-  `WeatherType` system without rain and thunder level tuning
- Ability to tune rain and thunder level without that `WeatherType` (aka intensity)

src/world/World.php
GroXzaLYs: I forgot about that😅

GroXzaLYs:
I don't know why `tests/travis.sh` is empty.

src/command/defaults/WeatherCommand.php
```diff
@@ -0,0 +1,86 @@
...
+		}
+
+		if(count($args) < 1){
+			$current = $world->getWeather();
+
+		$stateName = match($current){
+			WeatherType::CLEAR => KnownTranslationFactory::commands_weather_query_clear(),
+			WeatherType::RAIN => KnownTranslationFactory::commands_weather_query_rain(),
+			WeatherType::THUNDER => KnownTranslationFactory::commands_weather_query_thunder(),
+		};
```
dktapps: This block is not indented correctly

src/command/defaults/WeatherCommand.php
```diff
@@ -0,0 +1,86 @@
...
+
+	public function execute(CommandSender $sender, string $label, array $args) : void{
+		if($sender instanceof Player){
+		  $world = $sender->getWorld();
+		}else{
+			$world = $sender->getServer()->getWorldManager()->getDefaultWorld();
+		}
+
+		if($world === null){
+			throw new \RuntimeException("Failed to retrieve world instance. Default world is not loaded.");
```
dktapps: Throw `AssumptionFailedError` here

src/command/defaults/WeatherCommand.php
```diff
@@ -0,0 +1,86 @@
...
+			"weather",
+			KnownTranslationFactory::pocketmine_command_weather_description(),
+			KnownTranslationFactory::pocketmine_command_weather_usage(),
+		);
+		$this->setPermission(DefaultPermissionNames::COMMAND_WEATHER);
+	}
+
+	public function execute(CommandSender $sender, string $label, array $args) : void{
+		if($sender instanceof Player){
+		  $world = $sender->getWorld();
```
dktapps: Indentation is not correct here

src/entity/LightningBolt.php
```diff
@@ -0,0 +1,98 @@
...
+			0.0,
+			0.0,
+			0.0,
+			0.0,
+			[],
+			[],
+			new PropertySyncData([], []),
+			[]
+		));
+	}
```
dktapps: Still not clear why this needs to be overridden

src/world/WeatherIntensity.php
```diff
@@ -0,0 +1,44 @@
...
+
+	// Rain
+	public const RAIN_LIGHT = 50000;
+	public const RAIN_NORMAL = 100000;
+	public const RAIN_HEAVY = 160000;
+
+	// Thunder
+	public const THUNDER_LIGHT = 100000;
+	public const THUNDER_NORMAL = 160000;
+	public const THUNDER_HEAVY = 200000;
```
dktapps: It's still not clear what these values actually mean. Do they represent ticks, are they a scale, or what? Are they used for anything apart from packets?

src/world/World.php
```diff
@@ -588,6 +600,110 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
...
+		  })(),
+		  WeatherType::RAIN => (function(){
+			$this->setRainLevel(1.0);
+			$this->setThunderLevel(0.0);
+		})(),
+		WeatherType::THUNDER => (function(){
+			$this->setRainLevel(1.0);
+			$this->setThunderLevel(1.0);
+		})(),
+	};
```
dktapps: This function's indentation is a mess
Also, it'd make more sense to just use a switch instead of inline closures

src/world/World.php
dktapps: This is still not correct. The point of getting rid of the packet positions was so that `broadcastPackets` could be used to send the packets to all players at once.

src/world/World.php
dktapps: Probably it's needed because of `/weather` needing to be able to tell what the current weather is. Though maybe that could be implemented differently or use thresholds.

src/world/World.php
dktapps: Uh, please don't resolve it without addressing it, thanks

src/world/World.php
```diff
@@ -1752,6 +1871,28 @@ public function getSkyLightReduction() : int{
...
+	public function getThunderLevel() : float{
+		return $this->thunderLevel;
+	}
+
+	public function setThunderLevel(float $level) : void {
+		if($level < 0.0 || $level > 1.0){
+			throw new \InvalidArgumentException("Thunder level must be between 0.0 and 1.0, $level given");
+		}
+		$this->thunderLevel = $level;
+	}
```
dktapps: Why are these functions in a different place to the rest of the weather-related functions?

src/event/weather/LightningStrikeEvent.php
```diff
@@ -0,0 +1,47 @@
...
+
+namespace pocketmine\event\weather;
+
+use pocketmine\event\Cancellable;
+use pocketmine\event\CancellableTrait;
+use pocketmine\event\Event;
+use pocketmine\math\Vector3;
+use pocketmine\world\World;
+
+class LightningStrikeEvent extends Event implements Cancellable {
```
dktapps: This seems not to be used

dktapps (changes requested):

src/world/World.php
```diff
@@ -398,6 +406,10 @@ public static function chunkHash(int $x, int $z) : int{
 	private const BLOCKHASH_X_SHIFT = self::BLOCKHASH_Y_BITS;
 	private const BLOCKHASH_Z_SHIFT = self::BLOCKHASH_X_SHIFT + self::BLOCKHASH_XZ_EXTRA_BITS;
 
+	private WeatherType $weather = WeatherType::CLEAR;
+	private int $weatherDuration = 0;
+	private int $weatherTick = 0;
```
dktapps: Why are these in a different place than the other weather-related fields?

src/world/WeatherIntensity.php
```diff
@@ -0,0 +1,52 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\world;
+
+/**
+ * Weather intensity values used only for client-side rendering.
+ *
+ * These numbers correspond to the intensity parameter of
+ * LevelEventPacket::START_RAIN and START_THUNDER.
```
dktapps: If these are specific to the protocol, they shouldn't be exposed to the API. At the minimum, this class should be marked as @internal. Although it likely belongs in BedrockProtocol.

src/world/World.php
```diff
@@ -588,6 +599,135 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
...
+			})(),
+			WeatherType::RAIN => (function(){
+				$this->setRainLevel(1.0);
+				$this->setThunderLevel(0.0);
+			})(),
+			WeatherType::THUNDER => (function(){
+				$this->setRainLevel(1.0);
+				$this->setThunderLevel(1.0);
+			})(),
+		};
```
dktapps: As I said before, a switch would be better for this, instead of baking closures inline

src/world/World.php
```diff
@@ -588,6 +599,135 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
...
+	}
+
+	private function broadcastWeatherPackets() : void{
+		$players = $this->getPlayers();
+		if(count($players) === 0){
+			return;
+		}
+
+		[$rainEvent, $thunderEvent] = match($this->weather){
```
dktapps: This array destructuring is redundant. You're creating a new array to pass them to broadcastPackets anyway.

src/world/World.php
```diff
@@ -588,6 +599,135 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
...
+		NetworkBroadcastUtils::broadcastPackets(
+			$players,
+			[$rainEvent, $thunderEvent]
+		);
+	}
+
+	public function spawnLightning(Vector3 $pos) : void{
+		$bolt = new LightningBolt(Location::fromObject($pos, $this));
+		$bolt->spawnToAll();
+	}
```
dktapps, line 667: Once again, this should be removed, there's no reason this entity can't be created like any other

dktapps (changes requested):

src/world/World.php
```diff
@@ -588,6 +599,135 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
...
+			],
+			WeatherType::RAIN => [
+				LevelEventPacket::create(LevelEvent::START_RAIN, WeatherIntensity::RAIN_NORMAL, null),
+				LevelEventPacket::create(LevelEvent::STOP_THUNDER, WeatherIntensity::NONE, null),
+			],
+			WeatherType::THUNDER => [
+				LevelEventPacket::create(LevelEvent::START_RAIN, WeatherIntensity::RAIN_HEAVY, null),
+				LevelEventPacket::create(LevelEvent::START_THUNDER, WeatherIntensity::THUNDER_HEAVY, null),
+			],
+		};
```
dktapps: It also seems sus to me to hardcode the network weather intensities when they're dynamic on the server side.

dktapps: I think this was also mentioned here https://github.com/pmmp/PocketMine-MP/pull/6870#discussion_r2479832178

src/command/defaults/WeatherCommand.php
```diff
@@ -0,0 +1,85 @@
...
+			default => throw new InvalidCommandSyntaxException(),
+		};
+
+		$duration = isset($args[1]) ? max(100, (int) $args[1]) : 6000;
+		$world->setWeather($type, $duration);
+		$sender->sendMessage(match($type){
+			WeatherType::CLEAR => KnownTranslationFactory::commands_weather_clear(),
+			WeatherType::RAIN => KnownTranslationFactory::commands_weather_rain(),
+			WeatherType::THUNDER => KnownTranslationFactory::commands_weather_thunder(),
+		});
```
dktapps: This should use `Command::broadcastCommandMessage()` for audit logging

dktapps (commented):
This looks mostly OK now, code-wise.

I think my main sticking points remaining are:
- There doesn't appear to be any way to disable the system, short of just setting the weather to clear for a very large duration. People aren't going to want it to rain in their lobbies.
- I don't like that we're constrained to clear/rainy/thunder.
  - setRainLevel and setThunderLevel will not update the WeatherType, so that means plugins can make the `/weather` command return wrong results anyway.
  - Instead of generally constraining it by the type, it would be better to have the `/weather` command dynamically decide the weather type based on the rain and thunder amount (e.g. rain=0 + thunder=0 = clear, rain>0 + thunder=0 = rainy, thunder > 0 = thunder (whether or not it's raining)).

GroXzaLYs:
@dktapps Would it be better if we add an on off argument in WeatherCommand?

dktapps:
That might be a good idea. But I'm more thinking about an `enable-weather` setting for `server.properties`.

Normally weather cycle would be controlled by game rules, but we don't have a general system for those currently. But we should still save the appropriate NBT in `level.dat`.

Actually, I just noticed this doesn't save anything related to weather in level.dat. You'll need to fix that.

Dasciam:
My idea is, to save `doweathercycle` key in `level.dat`, if the state of whether weather cycle is changed, by default we apply setting from `server.properties` or `pocketmine.yml`(best case), but not saving it

dktapps:
I don't want to put this in pocketmine.yml. pocketmine.yml is supposed to be the place for "don't touch, might break stuff" settings. Something this simple should be easily accessible.

src/entity/EntityFactory.php
```diff
@@ -195,6 +195,10 @@ public function __construct(){
 			'ThrownTrident' //as above
 		]);
 
+		$this->register(LightningBolt::class, function(World $world, CompoundTag $nbt) : LightningBolt{
+		  return new LightningBolt(Helper::parseLocation($nbt, $world), $nbt);
+		}, ['lightning_bolt', 'minecraft:lightning_bolt']);
```
Dasciam, line 200: nvm

src/command/defaults/WeatherCommand.php
```diff
@@ -0,0 +1,85 @@
...
+		);
+		$this->setPermission(DefaultPermissionNames::COMMAND_WEATHER);
+	}
+
+	public function execute(CommandSender $sender, string $label, array $args) : void{
+		if($sender instanceof Player){
+			$world = $sender->getWorld();
+		}else{
+			$world = $sender->getServer()->getWorldManager()->getDefaultWorld();
+		}
```
ShockedPlot7560, line 53: It could be better to have the ability to give world name or use the current/default as default. Even though, in my opinion, the name should be mandatory if it is not a player.

src/command/defaults/WeatherCommand.php
```diff
@@ -0,0 +1,85 @@
...
+			$sender->sendMessage(KnownTranslationFactory::commands_weather_query($stateName));
+			return;
+		}
+
+		$type = match(strtolower($args[0])){
+			"clear" => WeatherType::CLEAR,
+			"rain" => WeatherType::RAIN,
+			"thunder" => WeatherType::THUNDER,
+			default => throw new InvalidCommandSyntaxException(),
+		};
```
ShockedPlot7560, line 91: As it's an enum, why not using `WeatherType::from` or `tryFrom` ?

src/command/defaults/WeatherCommand.php
```diff
@@ -0,0 +1,85 @@
...
+
+		$duration = isset($args[1]) ? max(100, (int) $args[1]) : 6000;
```
ShockedPlot7560: `$args[1]` is set, you checked the length before

src/event/weather/WeatherChangeEvent.php
```diff
@@ -0,0 +1,52 @@
...
+
+	public function getWorld() : World{
+		return $this->world;
+	}
+
+	public function getOldWeather() : WeatherType{
+		return $this->oldWeather;
+	}
+
+	public function getNewWeather() : WeatherType{
```
ShockedPlot7560: We should be able to change the newWeather

src/world/sound/ThunderSound.php
```diff
@@ -0,0 +1,41 @@
...
+
+final class ThunderSound implements Sound{
+
+	public function encode(Vector3 $pos) : array{
+		return [
+			LevelSoundEventPacket::nonActorSound(
+				LevelSoundEvent::THUNDER,
+				$pos,
+				false
+			)
```
ShockedPlot7560: To align it with the rest, it might be necessary to put it on the same line.

src/world/World.php
```diff
@@ -588,6 +599,134 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
...
+	}
+
+	private function broadcastWeatherPackets() : void{
+		$players = $this->getPlayers();
+		if(count($players) === 0){
+			return;
+		}
+
+		$rainInt = (int) ($this->rainLevel * 65535);
+		$thunderInt = (int) ($this->thunderLevel * 65535);
```
ShockedPlot7560: Will be clearer by using floor or round instead of casting to an int

ShockedPlot7560 (changes requested):
What I saw when I did a quick review

GroXzaLYs:
Okay, maybe I’ll continue this later, since I have other things to do.

src/command/defaults/WeatherCommand.php
dktapps: These commands are typically meant to match vanilla, there's no option to specify world for other commands

src/command/defaults/WeatherCommand.php
GroXzaLYs: ```from()``` / ```tryFrom()``` can’t really be used here since the enum has no string values and the command input is lowercase. A manual mapping like this is still the simplest option.

src/world/World.php
GroXzaLYs: Since the packet requires an int, the final value needs to stay integer. The fields are floats, so casting is fine, but using round() before casting can make the conversion more explicit.

GroXzaLYs:
I don't know how to save to level.dat

src/world/World.php
```diff
@@ -588,6 +603,140 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
 		}
 	}
 
+	public function getWeather() : WeatherType{
+		return $this->weather;
+	}
```
dktapps: As I said previously, "weather type" should be confined to commands only. It doesn't make sense anywhere else considering the dynamic rain & thunder strengths.

src/world/format/io/data/BedrockWorldData.php
```diff
@@ -250,4 +252,12 @@ public function getLightningLevel() : float{
...
 		$this->compoundTag->setFloat(self::TAG_LIGHTNING_LEVEL, $level);
 	}
+
+	public function isWeatherEnabled() : bool{
+		return $this->compoundTag->getByte(self::TAG_WEATHER_ENABLED, 1) === 1;
+	}
+
+	public function setWeatherEnabled(bool $enabled) : void{
+	  $this->compoundTag->setByte(self::TAG_WEATHER_ENABLED, $enabled ? 1 : 0);
+	}
```
dktapps: - These are not used anywhere
- The tag name likely doesn't match vanilla (check the wiki)
- It should be added to the `WorldData` interface
- Other world data types also need to implement something for this

src/world/format/io/data/BedrockWorldData.php
```diff
@@ -250,4 +252,12 @@ public function getLightningLevel() : float{
 	public function setLightningLevel(float $level) : void{
 		$this->compoundTag->setFloat(self::TAG_LIGHTNING_LEVEL, $level);
 	}
+
+	public function isWeatherEnabled() : bool{
+		return $this->compoundTag->getByte(self::TAG_WEATHER_ENABLED, 1) === 1;
+	}
+
+	public function setWeatherEnabled(bool $enabled) : void{
+	  $this->compoundTag->setByte(self::TAG_WEATHER_ENABLED, $enabled ? 1 : 0);
```
dktapps: Indentation is also not correct here

dktapps (commented):
On closer inspection there's a few features related to rain that need to be implemented:

- Fire should be extinguished when raining
- Rain should extinguish burning entities
- Farmland should hydrate when raining

Basically most stuff in this section needs to be implemented: https://minecraft.wiki/w/Rain#Effects

src/command/defaults/WeatherCommand.php
```diff
@@ -0,0 +1,116 @@
...
+
+			default:
+				throw new InvalidCommandSyntaxException("Invalid weather type.");
+		}
+
+		(function() use ($world, $duration){
+			$world->weatherDuration = $duration;
+			$world->weatherTick = 0;
+			$world->broadcastWeatherPackets();
+		})->bindTo(null, get_class($world))();
```
dktapps: ????

tests/phpstan/configs/actual-problems.neon
```diff
@@ -1415,3 +1415,21 @@ parameters:
...
+			message: '#^Access to private property pocketmine\\world\\World\:\:\$weatherTick\.$#'
+			identifier: property.private
+			count: 1
+			path: src/command/defaults/WeatherCommand.php
+
+		-
+			message: '#^Call to private method broadcastWeatherPackets\(\) of class pocketmine\\world\\World\.$#'
+			identifier: method.private
+			count: 1
+			path: src/command/defaults/WeatherCommand.php
```
dktapps: If you have to add errors to this file it usually means you're doing something wrong

src/block/Fire.php
```diff
@@ -75,8 +75,18 @@ public function onRandomTick() : void{
...
+		$world = $this->position->getWorld();
+			if($world->isRainingAt($this->position)) {
+				$world->setBlock($this->position, VanillaBlocks::AIR());
+				return;
+			}
+
 		if(!$down->burnsForever()){
-			//TODO: check rain
+			$world = $this->position->getWorld();
+			if($world->isRainingAt($this->position)){
```
dktapps: You already checked this above

src/block/Fire.php
```diff
@@ -187,8 +197,9 @@ private function spreadFire() : void{
 					if($block->getTypeId() !== BlockTypeIds::AIR){
 						continue;
 					}
-
-					//TODO: fire can't spread if it's raining in any horizontally adjacent block, or the current one
+					if($world->isRainingAt($this->position) || $world->isRainingAt($block->position)){
```
dktapps: The fire's own position was already checked above

src/world/World.php
```diff
@@ -588,6 +602,193 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
 		}
 	}
 
+	/**
+	 * @return "clear"|"rain"|"thunder"
```
dktapps: When I said get rid of the enum, I didn't mean replace it with bare strings...
I meant to get/set the rain & thunder level directly, and use the rain and thunder level to calculate what the weather status is in /weather. The weather status isn't needed anywhere except /weather.

src/world/World.php
```diff
@@ -588,6 +602,193 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
...
+
+		$x = (int) $pos->x;
+		$y = (int) $pos->y;
+		$z = (int) $pos->z;
+
+		if(!$this->isChunkLoaded($x >> 4, $z >> 4)){
+			return false;
+		}
+
+		for($i = $y + 1, $maxY = $this->getMaxY(); $i <= $maxY; $i++){
```
dktapps: This is going to be laggy AF.
Use `World->getHighestBlockAt()` and count down until you hit a block that stops rain or you reach the affected Y coordinate.

src/world/World.php
```diff
@@ -588,6 +602,193 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
...
+	}
+
+	public function isRainingAt(Vector3 $pos) : bool{
+		if($this->getRainLevel() <= 0){
+			return false;
+		}
+
+		$x = (int) $pos->x;
+		$y = (int) $pos->y;
+		$z = (int) $pos->z;
```
dktapps: `(int)` cast should not be used on coordinates. Use `getFloorZ()` etc

src/world/World.php
```diff
@@ -588,6 +602,193 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
...
+
+		if(!$this->isChunkLoaded($x >> 4, $z >> 4)){
```
dktapps: This should use `Chunk::COORD_BIT_SIZE`

dktapps (changes requested):

src/world/World.php
```diff
@@ -588,6 +602,187 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
...
+		$type = strtolower($type);
+
+		$currentRain = $this->getRainLevel();
+		$currentThunder = $this->getThunderLevel();
+
+		$old = "clear";
+		if($currentRain > 0.0 && $currentThunder > 0.0){
+			$old = "thunder";
+		}elseif($currentRain > 0.0){
+			$old = "rain";
```
dktapps: I said to get rid of these strings and use the rain and thunder strengths directly

GroXzaLYs:
I forgot about this...

GroXzaLYs:
I think I forgot to add ```NetworkBroadcastUtils::broadcastPackets()```

GroXzaLYs:
It's been a week

src/entity/LightningBolt.php
```diff
@@ -0,0 +1,79 @@
...
+
+	public function onFirstUpdate(int $currentTick) : void{
+		parent::onFirstUpdate($currentTick);
+		$this->getWorld()->addSound($this->location, new ThunderSound());
+	}
+
+	public function entityBaseTick(int $tickDiff = 1) : bool{
+		$hasUpdate = parent::entityBaseTick($tickDiff);
+
+		foreach($this->getWorld()->getNearbyEntities($this->getBoundingBox()->expandedCopy(3, 3, 3), $this) as $entity){
```
Dasciam: I don't think that logic should be in entity base tick. Move this to `onFirstUpdate`. otherwise, every entity in range of lightning bounding box expanded by 3 in all directions, will constantly get damage by lightning, what is not vanilla

src/block/Farmland.php
```diff
@@ -162,6 +162,11 @@ protected function canHydrate() : bool{
 		$startY = $this->position->getFloorY();
 		$startZ = $this->position->getFloorZ() - (int) (self::WATER_SEARCH_HORIZONTAL_LENGTH / 2);
 
+		if($world->getRainLevel() > 0 && $world->isRainingAt($this->position)){
```
kostamax27: Why call `getRainLevel` if a similar check is already performed in `isRainingAt`?

src/entity/Entity.php
```diff
@@ -667,6 +667,13 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
 			$hasUpdate = true;
 		}
 
+		if($this->isOnFire()){
+			$world = $this->getWorld();
+			if($world->getRainLevel() > 0 && $world->isRainingAt($this->getPosition())){
```
kostamax27: Same here

src/world/World.php
```diff
@@ -588,6 +602,144 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
...
+			throw new \InvalidArgumentException("Thunder level must be between 0.0 and 1.0, $level given");
+		}
+		$this->thunderLevel = $level;
+	}
+
+	public function isWeatherEnabled() : bool{
+		return $this->weatherEnabled;
+	}
+
+	public function isRainingAt(Vector3 $pos) : bool{
```
kostamax27: The current name doesn’t quite match the existing functionality in World.
You might consider adding `isRaining`, which takes a `Vector3`, and `isRainingAt`, which accepts `x, y, z` coordinates-similar to how `getBlock`, `getBlockAt`, and other related methods are structured.

src/block/Farmland.php
kostamax27: Maybe this check should be moved to the beginning of the function?

src/event/world/WeatherChangeEvent.php
```diff
@@ -0,0 +1,52 @@
...
+declare(strict_types=1);
+
+namespace pocketmine\event\world;
+
+use pocketmine\event\Cancellable;
+use pocketmine\event\CancellableTrait;
+use pocketmine\event\Event;
+use pocketmine\world\World;
+
+class WeatherChangeEvent extends Event implements Cancellable{
```
kostamax27: Instead of `Event`, you should use `WorldEvent`.

src/world/World.php
```diff
@@ -588,6 +602,144 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
 		}
 	}
 
+	public function setWeather(float $rainLevel, float $thunderLevel, int $duration = 6000) : void{
+		if($this->rainLevel === $rainLevel && $this->thunderLevel === $thunderLevel){
+			$this->weatherDuration = $duration;
```
kostamax27: Why is this? It looks suspicious.

src/world/World.php
```diff
@@ -588,6 +602,144 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
...
+		$this->rainLevel = $ev->getNewRainLevel();
+		$this->thunderLevel = $ev->getNewThunderLevel();
+		$this->weatherDuration = $duration;
+		$this->weatherTick = 0;
+
+		$this->broadcastWeatherPackets();
+	}
+
+	private function tickWeather() : void{
+		if(!$this->isWeatherEnabled()){
```
kostamax27: Maybe it's better to use `weatherEnabled` directly here?

src/entity/Entity.php
```diff
@@ -667,6 +667,13 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
 			$hasUpdate = true;
 		}
 
+		if($this->isOnFire()){
```
kostamax27: Maybe it would be better to move this logic into `doOnFireTick`?

src/entity/Entity.php
```diff
@@ -667,6 +667,13 @@ protected function entityBaseTick(int $tickDiff = 1) : bool{
...
+			$world = $this->getWorld();
+			if($world->isRaining($this->getPosition())){
+				$this->extinguish();
```
kostamax27: Perhaps we should introduce a new cause in `EntityExtinguishEvent`?

src/world/World.php
GroXzaLYs: 😅

src/entity/Entity.php
```diff
@@ -737,6 +737,13 @@ protected function doOnFireTick(int $tickDiff = 1) : bool{
 			return false;
 		}
 
+		if($this->isOnFire()){
```
kostamax27: This seems redundant, since `isOnFire` is already checked earlier in `entityBaseTick`.

src/entity/Entity.php
```diff
@@ -737,6 +737,11 @@ protected function doOnFireTick(int $tickDiff = 1) : bool{
 			return false;
 		}
 
+		$world = $this->getWorld();
+		if($world->isRaining($this->getPosition())){
```
kostamax27: Use `$this->location` directly

src/entity/LightningBolt.php
```diff
@@ -0,0 +1,77 @@
...
+
+	protected function initEntity(CompoundTag $nbt) : void{
+		parent::initEntity($nbt);
+		$this->setNameTagVisible(false);
+	}
+
+	public function onFirstUpdate(int $currentTick) : void{
+		parent::onFirstUpdate($currentTick);
+		$this->getWorld()->addSound($this->location, new ThunderSound());
+		foreach($this->getWorld()->getNearbyEntities($this->getBoundingBox()->expandedCopy(3, 3, 3), $this) as $entity){
```
kostamax27: Use `$this->boundingBox` directly

src/world/World.php
```diff
@@ -588,6 +602,139 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
...
+		if($this->thunderLevel > 0.8 && mt_rand(0, 200) === 0){
+			$players = $this->getPlayers();
+			if(count($players) > 0){
+				$target = $players[array_rand($players)];
+				$spawn = $target->getPosition();
+				$x = (int) $spawn->x + mt_rand(-30, 30);
+				$z = (int) $spawn->z + mt_rand(-30, 30);
+				$y = $this->getHighestBlockAt($x, $z) ?? max(64, (int) $spawn->y);
+
+				$loc = Location::fromObject(new Vector3($x, $y, $z), $this);
```
kostamax27: `$location`

remminiscent:
whats left for this? i can complete it if need be

kostamax27:
https://github.com/pmmp/PocketMine-MP/blob/e9a3bc69d243d9208aea8933cfe81a1fbf667707/src/network/mcpe/NetworkSession.php#L1282
https://github.com/pmmp/PocketMine-MP/blob/e9a3bc69d243d9208aea8933cfe81a1fbf667707/src/network/mcpe/handler/PreSpawnPacketHandler.php#L80-L81

kostamax27:
https://minecraft.fandom.com/wiki/Rain#:~:text=Rain%20fills%20up%20cauldrons%20with%20water%2C%20albeit%20slowly

remminiscent:
It seems like snow isn’t acknowledged here either.

src/world/World.php
```diff
@@ -1435,6 +1583,8 @@ public function save(bool $force = false) : bool{
 		$timings->startTiming();
 
 		$this->provider->getWorldData()->setTime($this->time);
+		$this->provider->getWorldData()->setRainLevel($this->rainLevel);
+		$this->provider->getWorldData()->setLightningLevel($this->thunderLevel);
```
remminiscent: The naming is inconsistent for thunder/lightning.

src/network/mcpe/NetworkSession.php
```diff
@@ -1309,14 +1309,30 @@ public function stopUsingChunk(int $chunkX, int $chunkZ) : void{
...
+
+		$packets[] = $world->getRainLevel() > 0
+			? LevelEventPacket::create(LevelEvent::START_RAIN, $rain, null)
+			: LevelEventPacket::create(LevelEvent::STOP_RAIN, 0, null);
+
+		$packets[] = $world->getThunderLevel() > 0
+			? LevelEventPacket::create(LevelEvent::START_THUNDER, $thunder, null)
+			: LevelEventPacket::create(LevelEvent::STOP_THUNDER, 0, null);
+
+		$this->session->sendDataPacket(...$packets);
```
kostamax27: I don't think it will work that way.

src/network/mcpe/NetworkSession.php
```diff
@@ -1309,14 +1309,30 @@ public function stopUsingChunk(int $chunkX, int $chunkZ) : void{
...
-			//TODO: weather needs to be synced here (when implemented)
-		}
+	if($this->player !== null){
+		$world = $this->player->getWorld();
+
+		$this->syncWorldTime($world->getTime());
+		$this->syncWorldDifficulty($world->getDifficulty());
+		$this->syncWorldSpawnPoint($world->getSpawnLocation());
+
+		// weather sync
```
kostamax27: It might be worth moving this into a separate function.

src/network/mcpe/NetworkSession.php
```diff
@@ -1309,14 +1309,30 @@ public function stopUsingChunk(int $chunkX, int $chunkZ) : void{
...
+		$rain = (int) ($world->getRainLevel() * 65535);
+		$thunder = (int) ($world->getThunderLevel() * 65535);
+
+		$packets = [];
+
+		$packets[] = $world->getRainLevel() > 0
```
kostamax27: It might be enough to check `$rain > 0` instead of `getRainLevel() > 0`.

src/network/mcpe/NetworkSession.php
```diff
@@ -1309,14 +1309,30 @@ public function stopUsingChunk(int $chunkX, int $chunkZ) : void{
...
+			? LevelEventPacket::create(LevelEvent::START_RAIN, $rain, null)
+			: LevelEventPacket::create(LevelEvent::STOP_RAIN, 0, null);
+
+		$packets[] = $world->getThunderLevel() > 0
```
kostamax27: Same here

dktapps:
Ok, I didn't want to come off as a jerk and say this so bluntly, but this has gone on long enough that the notification spam from here is really getting on my nerves. I'm on the verge of muting the PR.

There are very basic errors in each commit that could be easily addressed by using the proper tools locally instead of spamming commits. Spamming commits chasing after CI failures is unproductive and it's also annoying for everyone getting notifications.

- Use an IDE like PhpStorm for writing and refactoring code. Notepads and generic text editors aren't enough without other tools
- Run `vendor/bin/phpstan` locally before you commit, it'll catch the obvious errors
- Run `vendor/bin/phpunit tests/phpunit` to run the unit tests
- [php-cs-fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) can help with some (not all) formatting problems

GroXzaLYs:
Got it, sorry about all the commit spam. I’ll run the checks on my end and won’t push small changes based on CI results anymore. Thanks for bringing this up!

src/world/World.php
```diff
@@ -588,6 +602,139 @@ private function initRandomTickBlocksFromConfig(ServerConfigGroup $cfg) : void{
...
+		if($roll <= 70){
+			$this->setWeather(0.0, 0.0, mt_rand(6000, 12000));
+		}elseif($roll <= 90){
+			$this->setWeather(1.0, 0.0, mt_rand(4000, 9000));
+		}else{
+			$this->setWeather(1.0, 1.0, mt_rand(2000, 6000));
+		}
+	}
+
+	private function broadcastWeatherPackets() : void{
```
kostamax27: Perhaps it should be done similarly to how it's implemented in `World->sendTime`.

src/world/World.php
```diff
@@ -945,6 +1077,12 @@ protected function actuallyDoTick(int $currentTick) : void{
 			$this->sendTimeTicker = 0;
 		}
 
+		if(++$this->sendWeatherTicker === 200){
+			$this->sendWeather();
+			$this->sendWeatherTicker = 0;
+		}
```
kostamax27: This behavior is incorrect and deviates from the vanilla implementation.

Weather should not be sent this way. Simply remove this part of the code, as well as `private int $sendWeatherTicker = 0;`.

src/entity/LightningBolt.php
```diff
@@ -0,0 +1,77 @@
...
+	protected function getInitialSizeInfo() : EntitySizeInfo{
+		return new EntitySizeInfo(1.8, 0.6);
+	}
+
+	protected function getInitialGravity() : float{
+		return 0.0;
+	}
+
+	protected function getInitialDragMultiplier() : float{
+		return 0.02;
```
kostamax27, line 46: Is this correct?

src/entity/LightningBolt.php
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	public function getName() : string{
+		return "Lightning Bolt";
+	}
+
+	protected function initEntity(CompoundTag $nbt) : void{
+		parent::initEntity($nbt);
+		$this->setNameTagVisible(false);
```
kostamax27, line 55: What is this for?

kostamax27:
I no longer know how this can be expanded...

src/entity/LightningBolt.php
GroXzaLYs: I don't think so, because I just guessed it.

PR #6876, opened 2025-10-31
Fix critical animation particles not displaying

Dasciam:
Reference: https://github.com/df-mc/dragonfly/pull/1127

Since 1.21.120, just sending AnimatePacket with AnimatePacket::ACTION_CRITICAL_HIT is not enough, must fill the `data` field as well

Magic value is dumped from BDS

### DIFF
--- a/src/entity/animation/CriticalHitAnimation.php
+++ b/src/entity/animation/CriticalHitAnimation.php
@@ -28,11 +28,11 @@ use pocketmine\network\mcpe\protocol\AnimatePacket;
 
 final class CriticalHitAnimation implements Animation{
 
-	public function __construct(private Living $entity){}
+	public function __construct(private Living $entity, private int $particleCount = 55){}
 
 	public function encode() : array{
 		return [
-			AnimatePacket::create($this->entity->getId(), AnimatePacket::ACTION_CRITICAL_HIT, 55)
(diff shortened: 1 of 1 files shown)


Wraith0x10:
LGTM

dktapps:
Any idea what the value means?

Dasciam:
I’ll open IDA to learn more about it and #6877 as well

dktapps:
Thanks. Let me know if you learn anything

dktapps:
It appears this value affects the number of particles spawned

Dasciam:
> It appears this value affects the number of particles spawned

Yeah I explored the code in IDA and can say this is true

src/entity/animation/CriticalHitAnimation.php
```diff
@@ -28,11 +28,11 @@
 
 final class CriticalHitAnimation implements Animation{
 
-	public function __construct(private Living $entity){}
+	public function __construct(private Living $entity, private int $particleCount = 55){}
```
ShockedPlot7560, line 31: 55 particles for a critical hit seems a lot, are you sure we are talking in "particle" and not another unit ?

dktapps: Yep, I tested it personally

dktapps (approved):

ShockedPlot7560 (approved):

PR #6877, opened 2025-10-31
Display magic hit particles when the entity is damaged by a weapon with damage bonus enchantments

Dasciam:
Recently Mojang made an update that made `AnimatePacket::ACTION_MAGICAL_CRITICAL_HIT` no longer useless. At latest version magical critical hits are displayed

### API changes
- Added `MagicHitAnimation` class

## Tests
Some later because I can't join from 1.21.120 (latest)


src/entity/animation/MagicHitAnimation.php
```diff
@@ -0,0 +1,37 @@
...
+
+use pocketmine\entity\Living;
+use pocketmine\network\mcpe\protocol\AnimatePacket;
+
+class MagicHitAnimation implements Animation{
+	public function __construct(private Living $entity){}
+
+	public function encode() : array{
+		return [
+			AnimatePacket::create($this->entity->getId(), AnimatePacket::ACTION_MAGICAL_CRITICAL_HIT, 15)
```
dktapps: Any idea what this magic number means?

Dasciam: I barely have idea what these numbers means, some guy from community dumped them

Dasciam:
I think it's good to merge

src/entity/animation/MagicHitAnimation.php
```diff
@@ -0,0 +1,37 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\entity\animation;
+
+use pocketmine\entity\Living;
+use pocketmine\network\mcpe\protocol\AnimatePacket;
+
+class MagicHitAnimation implements Animation{
+	public function __construct(private Living $entity, private int $particleCount = 15){}
```
ShockedPlot7560, line 30: From where this particleCount come from ? Do you get any from a wiki or tested in game ? 
That's the only blocking point to merge for me

dktapps: It's from BDS MITM traces afaik

dktapps (approved):

src/entity/animation/MagicHitAnimation.php
Dasciam: > It's from BDS MITM traces afaik

I've reversed the places where `AnimatePacket` is creating and can confirm they're as valid as from MITM dump

ShockedPlot7560: good for me, then

ShockedPlot7560 (approved):

PR #6116, opened 2023-10-23
Add moss

AzaleeX:
## Introduction
Addition of blocks / foam mats with breaking time with a hoe add in the BlockBreakInfo

## Hoe add in the BlockBreakInfo

https://github.com/AzaleeX/PocketMine-MP/blob/04eb5ddabc8e3debd8a5945e8a84d797fe56849b/src/block/BlockBreakInfo.php#L72-#L74

## Add Moss Block
https://github.com/AzaleeX/PocketMine-MP/blob/04eb5ddabc8e3debd8a5945e8a84d797fe56849b/src/block/Moss.php#L1-L36

## Add Carpet Moss

https://github.com/AzaleeX/PocketMine-MP/blob/04eb5ddabc8e3debd8a5945e8a84d797fe56849b/src/block/MossCarpet.php#L1-L75

## Tests Screen

image

PR #6885, opened 2025-11-02
Implement Moss blocks

Azvyl:
This PR implements moss blocks and moss carpets, including their pale variants. This also update relevant block logic to recognize moss blocks as valid support.
- [Moss Block](https://minecraft.wiki/w/Moss_Block)
- [Moss Carpet](https://minecraft.wiki/w/Moss_Carpet)
- [Pale Moss Block](https://minecraft.wiki/w/Pale_Moss_Block)
- [Pale Moss Carpet](https://minecraft.wiki/w/Pale_Moss_Carpet)

This PR does not implement Azalea and Flowering Azalea

1.17.0 - Caves and Cliffs, 1.21.50 - The Garden Awakens

### Related issues & PRs
#6116

### API changes
* Added new block classes: `MossBlock`, `MossCarpet`, `PaleMossBlock`, `PaleMossCarpet`, and `PaleMossCarpetVine`
* Registered `MOSS_BLOCK`, `MOSS_CARPET`, `PALE_MOSS_BLOCK`, `PALE_MOSS_CARPET`, and `PALE_MOSS_CARPET_VINE` in `VanillaBlocks`
* Added new block type IDs for moss blocks and carpets in `BlockTypeIds`
* Introduced the `MOSS_REPLACEABLE` tag in `BlockTypeTags` and applied it to relevant existing blocks to allow moss spreading and placement
* Updated support logic in various blocks to allow placement on moss blocks and recognize moss blocks as valid supports.

### Behavioural changes
* These blocks now recognize moss block as valid support:
  * Bamboo & Bamboo Sapling
  * Big Dripleaf
  * Deadbush
  * DoublePlant
  * Flowers
  * Nether Fungus
  * Nether Roots
  * Nether Sprouts
  * Pink Petals
  * Saplings
  * Small Dripleaf
  * Sugarcane
  * Sweet Berry Bush
  * Tall Grass

## Follow-up
Implement Azalea and Flowering Azalea

## Tests
https://youtu.be/mBI6AJ52Lx8


src/block/MossBlock.php
```diff
@@ -0,0 +1,105 @@
...
+		$originX = $this->position->x;
+		$originY = $this->position->y;
+		$originZ = $this->position->z;
+
+		for($dx = -$maxX; $dx <= $maxX; ++$dx){
+			for($dz = -$maxZ; $dz <= $maxZ; ++$dz){
+				$absdx = $dx < 0 ? -$dx : $dx;
+				$absdz = $dz < 0 ? -$dz : $dz;
+
+				if(($absdx === $maxX && $absdz === $maxZ) || (($absdx === $maxX || $absdz === $maxZ) && mt_rand(1, 100) > 75)){
```
dktapps: The left and right side of this || seem to be identical, not sure what you intended here but this is confusing

src/block/Moss.php
```diff
@@ -0,0 +1,105 @@
...
+					}
+				}else{
+					for($y = $startY; $y <= $startY + 4; ++$y){
+						$b = $world->getBlockAt($x, $y, $z);
+						if($b->getTypeId() !== BlockTypeIds::AIR && $world->getBlockAt($x, $y + 1, $z)->getTypeId() === BlockTypeIds::AIR){
+							$foundBlock = $b;
+							break;
+						}
+					}
+				}
```
dktapps, line 79: You could deduplicate this code by making limit and diff variables instead of hardcoding the bounds and ++/--

src/block/MossBlock.php
```diff
@@ -0,0 +1,105 @@
...
+						$b = $world->getBlockAt($x, $y, $z);
+						if($b->getTypeId() !== BlockTypeIds::AIR && $world->getBlockAt($x, $y + 1, $z)->getTypeId() === BlockTypeIds::AIR){
+							$foundBlock = $b;
+							break;
+						}
+					}
+				}
+
+				if($foundBlock !== null && $foundBlock->hasTypeTag(BlockTypeTags::MOSS_REPLACEABLE) &&
+					($foundBlock === $this || BlockEventHelper::spread($foundBlock, (clone $this), $this)) &&
```
dktapps: Blocks should never be compared by identity, there's no guarantee that `getBlock()` will return the same object twice

src/block/PaleMossCarpet.php
```diff
@@ -0,0 +1,204 @@
...
+class MossyCarpet extends Flowable{
+	use StaticSupportTrait;
+
+	protected bool $top = false;
+
+	/** @var WallConnectionType[] */
+	protected array $sides = [];
+
+	protected function describeBlockOnlyState(RuntimeDataDescriber $w) : void{
+		$w->wallConnections($this->sides);
```
dktapps: huh???

src/block/PaleMossCarpet.php
```diff
@@ -0,0 +1,204 @@
...
+		$w->bool($this->top);
```
dktapps: huh???????

src/block/PaleMossCarpet.php
```diff
@@ -0,0 +1,204 @@
...
+		foreach(Facing::HORIZONTAL as $f){
+			$side = null;
+			if($above->getAdjacentSupportType($f)->hasEdgeSupport() && $base->getSideConnection($f) !== null){
+				$side = WallConnectionType::SHORT;
+			}
+			$new->setSideConnection($f, $side);
+		}
+
+		return $new->hasFaces() ? $new : null;
+	}
```
dktapps, line 90: I have absolutely no idea what's going on in this class. But Moss Carpet is supposed to be a simple stateless block much the same as a regular carpet. Please explain.

src/block/PaleMossCarpet.php
```diff
@@ -0,0 +1,204 @@
...
+	public function isTop() : bool{ return $this->top; }
+
+	public function setTop(bool $top) : self{
+		$this->top = $top;
+		return $this;
+	}
+
+	public function getDrops(Item $item) : array{
+		return $this->isTop() ? [] : parent::getDrops($item);
+	}
```
dktapps: This is not correct

src/block/MossCarpet.php
```diff
@@ -0,0 +1,40 @@
...
+	use StaticSupportTrait;
+
+	protected function recalculateCollisionBoxes() : array{
+		return [AxisAlignedBB::one()->trim(Facing::UP, 15 / 16)];
+	}
+
+	private function canBeSupportedAt(Block $block) : bool{
+		return $block->getSide(Facing::DOWN)->getTypeId() !== BlockTypeIds::AIR;
+	}
+}
```
dktapps, line 40: uhhh????

dktapps (changes requested):
I don't know WTF is going on with Mossy Carpet

src/block/MossCarpet.php
Azvyl: The existing carpet class is more specific to WoolCarpet/ColoredCarpet (ColoredTrait), besides that Moss Carpet is not flammable

src/block/PaleMossCarpet.php
Azvyl: We can only get the Pale Moss Carpet item through the base part, whether using tools or silktouch.

dktapps: Ok, apparently I don't know how pale mossy carpets work. WTF

src/block/PaleMossCarpet.php
Azvyl: Unfortunately no, it has a similar state to wall. it is used to display decorative vines if it is surrounded by blocks

src/block/MossyCarpet.php
```diff
@@ -0,0 +1,204 @@
...
+use pocketmine\data\runtime\RuntimeDataDescriber;
+use pocketmine\item\Fertilizer;
+use pocketmine\item\Item;
+use pocketmine\math\AxisAlignedBB;
+use pocketmine\math\Facing;
+use pocketmine\math\Vector3;
+use pocketmine\player\Player;
+use pocketmine\world\BlockTransaction;
+
+class MossyCarpet extends Flowable{
```
dktapps: Better rename this to `PaleMossCarpet`, since it has significantly different behaviour than regular moss carpets

src/block/NormalCarpet.php
```diff
@@ -0,0 +1,40 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\block;
+
+use pocketmine\block\utils\StaticSupportTrait;
+use pocketmine\math\AxisAlignedBB;
+use pocketmine\math\Facing;
+
+class NormalCarpet extends Flowable{
```
dktapps: Probably this should be called `MossCarpet`

src/block/PaleMossCarpet.php
```diff
@@ -0,0 +1,204 @@
...
+
+	/** @var WallConnectionType[] */
+	protected array $sides = [];
+
+	protected function describeBlockOnlyState(RuntimeDataDescriber $w) : void{
+		$w->wallConnections($this->sides);
+		$w->bool($this->top);
+	}
+
+	public function getSideConnection(int $face) : ?WallConnectionType{
```
dktapps: This is very misleading and had me super confused.

- This should be documented as to what it actually does (sets the height of vegetation growing from this side of the block, none/short/tall)
- It probably shouldn't be called "connection", this is more akin to vines than walls. Maybe something like `getVinesHeight()` would be less confusing
- It probably shouldn't use the `WallConnectionType` enum

src/block/MossBlock.php
Azvyl: I want the moss block spread to have a 25% chance on the outer edges, but 0% in the corners

src/block/PaleMossCarpet.php
```diff
@@ -0,0 +1,204 @@
...
+	public function setSideConnection(int $face, ?WallConnectionType $value) : self{
+		if($value === null){
+			unset($this->sides[$face]);
+			return $this;
+		}
+		$this->sides[$face] = $value;
+		return $this;
+	}
+
+	public function isTop() : bool{ return $this->top; }
```
dktapps: This doesn't make sense to me. Since it actually decides whether or not the block is a carpet. Given the design of the states, it's also probably possible to stack multiple "top" blocks to get vines taller than 2 blocks from one carpet. Whether or not this _should_ be possible is another question.

We probably should split this into Pale Moss Carpet and Pale Moss (Carpet) Vines.

src/block/MossBlock.php
dktapps: Well, this is probably not going to do what you want. Currently it has a 25% chance of skipping to the next iteration if you're not on a corner (meaning a 75% chance of proceeding).

Should be sufficient to do `if((abs($dx) === $maxX && abs($dz) === $maxZ) || mt_rand(1, 100) > 25){ continue; }`

dktapps:
Sorry for all the WTFs btw. Since I'm not familiar with pale moss carpets I was tripping out wondering if you'd vibecoded the class for it 😂

src/block/MossBlock.php
Azvyl: Ah, my bad. 75% is the correct chance
Wiki:
> Each corner has a 0% to be selected, each edge has a 75% chance, and the rest of the blocks are always selected.

dktapps: Oh, I misread the condition. I didn't see you had a || on the right side. Very confusing

dktapps: Maybe worth breaking this condition up into two lines and adding some comments for people like me who can't read

src/block/MossBlock.php
Azvyl: The `$foundBlock === $this` code is there to avoid calling an event and spreading the moss block to itself (the spreader), but still have the opportunity to grow vegetation on it. Shall we just delete it?

src/block/PaleMossCarpet.php
Azvyl: Simply, I want to use `getVinesHeight(int $face): int`, but it seems to be misleading, we don't use `getBlock()` to check how high the vine is (we use state). So do we need to create a new enum in `block\utils`? The content will be the same as `WallConnectionType`

dktapps: Yes, create a new one. Also, maybe NONE/HALF/FULL would be more appropriate names for the cases.

The nullability of WallConnectionType created some annoyances that required it to be shimmed in the new serializer system, so better use a proper case for that.

src/block/MossBlock.php
dktapps: Looks to me like you can check if dx and dz are 0, and avoid the whole Y scanning entirely

src/block/VanillaBlocks.php
```diff
@@ -963,6 +970,42 @@ public function getBreakTime(Item $item) : float{
...
+			}
+		}) extends MossBlock{
+			public function getFlameEncouragement() : int{
+				return 15;
+			}
+
+			public function getFlammability() : int{
+				return 100;
+			}
+		});
```
dktapps: Considering the amount of customisation these classes need, I don't think it really makes sense to have an injectable callback. Might as well just make proper classes. I don't really like baking code inline like this.

Azvyl: What about the line order? Does it need to be moved?

dktapps: Uh, they probably should be ordered alphabetically. But this function is one big mess anyway so I wouldn't worry too much about it.

src/block/PaleMossCarpetVine.php
```diff
@@ -0,0 +1,45 @@
...
+ *
+ */
+
+declare(strict_types=1);
+
+namespace pocketmine\block;
+
+use pocketmine\item\Item;
+
+class PaleMossCarpetVine extends PaleMossCarpet{
```
dktapps: This seems suspicious. When is a carpet not a carpet?

Azvyl: Sorry I'm not sure I understand the question. If it's the top part, it doesn't have the carpet texture

dktapps: `instanceof PaleMossCarpet` will return true for the vines, which are not really carpets

Azvyl: So we should separate these two classes? These two blocks have similarities so I thought we could make `isCarpetPart()`

src/block/NetherFungus.php
```diff
@@ -53,7 +53,6 @@ public function __construct(
 	}
 
 	private function canBeSupportedAt(Block $block) : bool{
-		//TODO: moss
```
ShockedPlot7560, line 56: As I see, nothing was changed to include moss in the support block of fungus ?

src/block/NetherRoots.php
```diff
@@ -30,7 +30,6 @@ final class NetherRoots extends Flowable{
 	use StaticSupportTrait;
 
 	private function canBeSupportedAt(Block $block) : bool{
-		//TODO: moss
```
ShockedPlot7560, line 33: same here

src/block/NetherSprouts.php
```diff
@@ -30,7 +30,6 @@ class NetherSprouts extends Flowable{
 	use StaticSupportTrait;
 
 	private function canBeSupportedAt(Block $block) : bool{
-		//TODO: moss
```
ShockedPlot7560, line 33: Same here

src/block/DeadBush.php
```diff
@@ -62,8 +62,9 @@ private function canBeSupportedAt(Block $block) : bool{
 				BlockTypeIds::DIRT,
 				BlockTypeIds::GRASS,
 				BlockTypeIds::HARDENED_CLAY,
-				BlockTypeIds::STAINED_CLAY => true,
-				//TODO: moss block
+				BlockTypeIds::STAINED_CLAY,
+				BlockTypeIds::MOSS_BLOCK,
+				BlockTypeIds::PALE_MOSS_BLOCK => true,
```
ShockedPlot7560, line 67: Would a tag for moss blocks be redundant? I'm asking myself.

src/block/PaleMossCarpetVine.php
ShockedPlot7560: The class name looks weird too. Its more a PaleMossVine only ? I'm agree that extending PaleMossCarpet is creepy, would be better to add a trait if you have many duplicate code ?

src/block/PinkPetals.php
```diff
@@ -66,7 +66,6 @@ public function setCount(int $count) : self{
 
 	private function canBeSupportedAt(Block $block) : bool{
 		$supportBlock = $block->getSide(Facing::DOWN);
-		//TODO: Moss block
```
ShockedPlot7560, line 69: Same for the todo

ShockedPlot7560 (changes requested):
I wonder if I missed something with the TODO removing ? Does minecraft changed the behavior ?

Azvyl:
I added the [DIRT tag](https://wiki.bedrock.dev/blocks/block-tags) to the [moss block](https://github.com/pmmp/PocketMine-MP/pull/6885/files#diff-f8dd11649e3d8efba46130b591eca72321895c2693f810c023d50e1c7149b149R1082). That's what the game does to check for support for most blocks

src/block/PaleMossCarpetVine.php
dktapps: > So we should separate these two classes? These two blocks have similarities so I thought we could make `isCarpetPart()`

You could have a common base class if you want, but the `instanceof PaleMossCarpet` returning true feels like a problem to me

src/block/DeadBush.php
Azvyl: I don't think it's needed at this time

src/block/PaleMossCarpetVine.php
Azvyl: I'm not sure about the naming of the base class, `BasePaleMossCarpet` seems a bit odd in the "base" part. What do you think? Or maybe I'm overthinking it
It's in line with the naming of abstract classes from other blocks though

dktapps: `PaleMossVegetation`? `PaleMossPlant`? idk

src/block/PaleMossVine.php
```diff
@@ -0,0 +1,140 @@
...
+			return;
+		}
+		if($updated){
+			$world->setBlock($this->position, $this);
+		}
+	}
+
+	protected function canBeSupportedAt(Block $block) : bool{
+		$below = $block->getSide(Facing::DOWN);
+		return $below instanceof PaleMossVine && $below->isCarpetPart();
```
dktapps: Type ID would make more sense for this

src/block/PaleMossVine.php
```diff
@@ -0,0 +1,140 @@
...
+			if($lastSide !== $side){
+				$this->setVineGrowth($f, $side);
+				$changed++;
+			}
+		}
+
+		return $changed > 0;
+	}
+
+	public function isCarpetPart() : bool{ return false; }
```
dktapps: Don't love this, especially not with it being public

src/block/PaleMossCarpet.php
```diff
@@ -0,0 +1,102 @@
...
+				$side = PaleMossVineGrowth::HALF;
+			}
+			$new->setVineGrowth($f, $side);
+		}
+
+		return $new->hasFaces() ? $new : null;
+	}
+
+	public function asItem() : Item{
+		return Block::asItem();
```
dktapps: This looks suspicious

src/block/MossBlock.php
```diff
@@ -0,0 +1,119 @@
...
+use pocketmine\item\Fertilizer;
+use pocketmine\item\Item;
+use pocketmine\math\Facing;
+use pocketmine\math\Vector3;
+use pocketmine\player\Player;
+use pocketmine\world\Position;
+use function abs;
+use function mt_rand;
+
+class MossBlock extends Opaque{
```
dktapps: Conventionally this would probably just be called `Moss`

src/block/MossBlock.php
```diff
@@ -0,0 +1,119 @@
...
+		if($rand <= 5208){
+			$world->setBlock($pos, VanillaBlocks::TALL_GRASS());
+		}elseif($rand <= 5208 + 2604){
+			$world->setBlock($pos, VanillaBlocks::MOSS_CARPET());
+		}elseif($rand <= 5208 + 2604 + 1042){
+			if($world->isInWorld($pos->x, $pos->y + 1, $pos->z)){
+				$world->setBlock($pos, VanillaBlocks::DOUBLE_TALLGRASS());
+				$world->setBlock($pos->up(), VanillaBlocks::DOUBLE_TALLGRASS()->setTop(true));
+			}
+		}elseif($rand <= 5208 + 2604 + 1042 + 729){
```
dktapps: What's the deal with these weird looking conditions?

src/block/PaleMossCarpet.php
```diff
@@ -0,0 +1,102 @@
...
+use pocketmine\block\utils\StaticSupportTrait;
+use pocketmine\item\Fertilizer;
+use pocketmine\item\Item;
+use pocketmine\math\AxisAlignedBB;
+use pocketmine\math\Facing;
+use pocketmine\math\Vector3;
+use pocketmine\player\Player;
+use pocketmine\world\BlockTransaction;
+
+class PaleMossCarpet extends PaleMossVine{
```
dktapps: Not a fan of this either.

My thinking was to have a base class and then the carpet and vine would individually extend it.
That way you wouldn't need a dodgy hack to avoid the infinite recursion for asItem()

src/data/bedrock/block/convert/VanillaBlockMappings.php
```diff
@@ -1593,6 +1598,37 @@ private static function registerSplitMappings(BlockSerializerDeserializerRegistr
...
+			$paleMossVineSideProperty(StateNames::PALE_MOSS_CARPET_SIDE_SOUTH, Facing::SOUTH),
+			$paleMossVineSideProperty(StateNames::PALE_MOSS_CARPET_SIDE_WEST, Facing::WEST)
+		]);
+		$paleMossCarpetVineModel = Model::create(Blocks::PALE_MOSS_VINE(), Ids::PALE_MOSS_CARPET)->properties([
+			new DummyProperty(StateNames::UPPER_BLOCK_BIT, true),
+			$paleMossVineSideProperty(StateNames::PALE_MOSS_CARPET_SIDE_NORTH, Facing::NORTH),
+			$paleMossVineSideProperty(StateNames::PALE_MOSS_CARPET_SIDE_EAST, Facing::EAST),
+			$paleMossVineSideProperty(StateNames::PALE_MOSS_CARPET_SIDE_SOUTH, Facing::SOUTH),
+			$paleMossVineSideProperty(StateNames::PALE_MOSS_CARPET_SIDE_WEST, Facing::WEST)
+		]);
```
dktapps, line 1642: This repetition could be avoided

src/block/MossBlock.php
Azvyl: Should we just add them up straight away? (5208, 7812, 8854, 9583)

src/block/MossBlock.php
Azvyl: I assume this also applies to the `PaleMossBlock` class

PR #6900, opened 2025-11-15
Fix #4830: Correct slime block physics for items and sneaking players

### LINKED ISSUE #4830: ItemEntity doesn't bounce on slime block

siyFred:
Closes #4830. This PR moves part of `onHitGround` logic (which only applied to `Living`) to `Entity` so `ItemEntity` can use it.

The `Living::onHitGround` method was updated to call the parent method first, ensuring fall damage is only calculated after a block (like `Slime`) has a chance to reset the fall distance.

I also fixed a logic bug in `Slime` where sneaking players would not bounce (correct) but would still take fall damage (incorrect).

I have tested the following in-game scenarios:
* `ItemEntity` (dropped items) now correctly bounce on slime blocks.
* Players correctly bounce on slime blocks and take no fall damage.
* Players holding "sneak" land on slime blocks without bouncing and without taking fall damage.
* Players still take normal fall damage when landing on any other block.
* Players who bounce off a slime block and then land on a normal block correctly take fall damage (confirming `fallDistance` resets and re-accumulates properly after the bounce).
* Arrows, Snowballs, and Eggs correctly break or stick on Slime blocks (they do not bounce).

### DIFF
--- a/src/block/Slime.php
+++ b/src/block/Slime.php
@@ -33,8 +33,8 @@ final class Slime extends Transparent{
 	}
 
 	public function onEntityLand(Entity $entity) : ?float{
-    		$entity->resetFallDistance();
-    		if($entity instanceof Living && $entity->isSneaking()){
+			$entity->resetFallDistance();
+			if($entity instanceof Living && $entity->isSneaking()){
 			return null;
 		}
 		return -$entity->getMotion()->y;
--- a/src/block/Slime.php
+++ b/src/block/Slime.php
@@ -33,8 +33,8 @@ final class Slime extends Transparent{
 	}
 
 	public function onEntityLand(Entity $entity) : ?float{
-    $entity->resetFallDistance();
-    if($entity instanceof Living && $entity->isSneaking()){
+    		$entity->resetFallDistance();
+    		if($entity instanceof Living && $entity->isSneaking()){
 			return null;
 		}
 		return -$entity->getMotion()->y;
--- a/src/block/Slime.php
+++ b/src/block/Slime.php
@@ -33,10 +33,10 @@ final class Slime extends Transparent{
 	}
 
 	public function onEntityLand(Entity $entity) : ?float{
-		if($entity instanceof Living && $entity->isSneaking()){
+    $entity->resetFallDistance();
+    if($entity instanceof Living && $entity->isSneaking()){
 			return null;
 		}
-		$entity->resetFallDistance();
 		return -$entity->getMotion()->y;
 	}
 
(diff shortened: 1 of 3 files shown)


src/block/Slime.php
```diff
@@ -33,10 +33,10 @@ public function getFrictionFactor() : float{
 	}
 
 	public function onEntityLand(Entity $entity) : ?float{
-		if($entity instanceof Living && $entity->isSneaking()){
+			$entity->resetFallDistance();
+			if($entity instanceof Living && $entity->isSneaking()){
```
dktapps, line 37: This indentation is not correct

src/entity/Entity.php
```diff
@@ -1111,7 +1113,18 @@ protected function updateFallState(float $distanceThisTick, bool $onGround) : ?f
...
+			if(count($fallBlock->getCollisionBoxes()) === 0){
+				$fallBlockPos = $fallBlockPos->down();
+				$fallBlock = $this->getWorld()->getBlock($fallBlockPos);
+			}
+			$newVerticalVelocity = $fallBlock->onEntityLand($this);
+
+			if($fallBlock->getTypeId() !== BlockTypeIds::AIR){
+				$this->broadcastSound(new EntityLandSound($this, $fallBlock));
+			}
+			return $newVerticalVelocity;
```
dktapps, line 1127: Same again here

src/entity/Entity.php
```diff
@@ -1111,7 +1113,18 @@ protected function updateFallState(float $distanceThisTick, bool $onGround) : ?f
...
+			$fallBlock = $this->getWorld()->getBlock($fallBlockPos);
+			if(count($fallBlock->getCollisionBoxes()) === 0){
+				$fallBlockPos = $fallBlockPos->down();
+				$fallBlock = $this->getWorld()->getBlock($fallBlockPos);
+			}
+			$newVerticalVelocity = $fallBlock->onEntityLand($this);
+
+			if($fallBlock->getTypeId() !== BlockTypeIds::AIR){
+				$this->broadcastSound(new EntityLandSound($this, $fallBlock));
+			}
```
dktapps, line 1126: This is going to make dropped items, XP orbs, arrows etc have step sounds

dktapps (changes requested):

PR #6901, opened 2025-11-15
Feature #6869: Add VoidGenerator

### LINKED ISSUE #6869: Implement void generator
### Problem description

example: level-type=void

### Proposed solution

A space where everything is empty

_No response_

siyFred:
Closes #6869.

This PR adds a new `VoidGenerator` to the `GeneratorManager`, implementing completely empty worlds.

I registered the alias as `"void"` in `GeneratorManager`.

You can try it by deleting all the worlds in `worlds/` and setting `level-type=void` in `server.properties`.

### DIFF
--- a/src/world/generator/VoidGenerator.php
+++ b/src/world/generator/VoidGenerator.php
@@ -1,0 +1,47 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
+ * |  _ \ ___   ___| | _____| |_|  \/  (_)_ __   ___      |  \/  |  _ \
+ * | |_) / _ \ / __| |/ / _ \ __| |\/| | | '_ \ / _ \_____| |\/| | |_) |
+ * |  __/ (_) | (__|   <  __/ |_| |  | | | | | |  __/_____| |  | |  __/
+ * |_|   \___/ \___|_|\_\___|\__|_|  |_|_|_| |_|\___|     |_|  |_|_|
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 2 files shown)


dktapps:
This is not correct, a void world isn't supposed to be completely empty

src/world/generator/VoidGenerator.php
```diff
@@ -0,0 +1,47 @@
...
+	public function __construct(int $seed, string $preset){
+		parent::__construct($seed, $preset);
+		$this->generateBaseChunk();
+	}
+
+  protected function generateBaseChunk() : void{
+	$this->chunk = new Chunk([], false);
+	}
+
+	public function generateChunk(ChunkManager $world, int $chunkX, int $chunkZ) : void{
```
Dasciam, line 41: There should be generation of a stone plate from (-16 -64 -16) to (16 -64 16) with cobblestone block on a center (0 64 0)

Image

Image

Image

Dasciam (commented):
Good, you need to resolve some indentation issues too

src/world/generator/VoidGenerator.php
```diff
@@ -0,0 +1,47 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\world\generator;
+
+use pocketmine\world\ChunkManager;
+use pocketmine\world\format\Chunk;
+
+class VoidGenerator extends Generator{
+	private Chunk $chunk;
```
Dasciam, line 30: Honestly don't get why we would want this for

TrixNEW: It is actually pretty useful, especially if you are trying to make new worlds and load a schematic.

dktapps: It's a private field not accessible to plugins. In what way is it useful ...?

PR #6904, opened 2025-11-16
First look at encoding chunks on the main thread

### LINKED ISSUE #6902: Possibility of encoding chunks on the main thread?
### Problem description

Currently, we serialize chunks with FastChunkSerializer to send them to an AsyncTask for encoding, which has the added overhead of extra serialization and async tasks. With ext-encoding's performance gains, would it be better to encode them on the main thread?

### Proposed solution

See how the server performs when chunk encoding is done fully on the main thread vs the current system.

_No response_

GameParrot:
This PR moves chunk encoding from AsyncTasks to the main thread. Chunk compression is still done on background threads with CompressBatchTask.

### Related issues & PRs
Closes #6902 

### Behavioural changes
Chunk encoding is on the main thread now, which might slightly increase load but should reduce overall CPU and memory usage.

## Backwards compatibility
Plugins that override ChunkCache methods will need to be updated.

## Follow-up
Continue testing performance with this change.

## Tests
From my performance testing, Player Send Chunk Prepare uses around 3-5% load when flying around with fly speed set to 6x the default on 32 chunk render distance with chunk sending per tick set to 400 (sending chunks way faster than they would be sent under normal conditions).


dktapps:
Gonna need A/B testing for this

dktapps:
Local testing shows that this slightly worsens performance for simple terrain. Player Send Chunk Prepare timings show about 55 us before these changes, and 66 us after.

There are a few factors you're not considering. First of all is the performance cost of complex palettes, as I mentioned in the original issue. A quick performance test shows the following:

```
Unique block states: 2
string(34) "10,237 ns per serialization (fast)"
string(37) "11,038 ns per serialization (network)"
Unique block states: 4
string(34) "10,157 ns per serialization (fast)"
string(37) "10,352 ns per serialization (network)"
Unique block states: 8
string(34) "10,224 ns per serialization (fast)"
string(37) "10,990 ns per serialization (network)"
Unique block states: 16
string(34) "10,336 ns per serialization (fast)"
string(37) "11,256 ns per serialization (network)"
Unique block states: 32
string(34) "10,428 ns per serialization (fast)"
string(37) "12,188 ns per serialization (network)"
Unique block states: 64
string(34) "10,643 ns per serialization (fast)"
string(37) "14,208 ns per serialization (network)"
Unique block states: 128
string(34) "11,134 ns per serialization (fast)"
string(37) "17,883 ns per serialization (network)"
Unique block states: 256
string(34) "11,759 ns per serialization (fast)"
string(37) "25,256 ns per serialization (network)"
Unique block states: 512
string(34) "13,710 ns per serialization (fast)"
string(37) "42,825 ns per serialization (network)"
Unique block states: 1024
string(34) "17,388 ns per serialization (fast)"
string(37) "72,203 ns per serialization (network)"
Unique block states: 2048
string(34) "24,393 ns per serialization (fast)"
string(38) "131,519 ns per serialization (network)"
```

You can see that the performance of the network serializer gets significantly worse as the size of the palettes grow. This is because the network serializer has to do a lot more work than the fast serializer (mapping IDs, individual varint encoding).

Second, you aren't actually reducing the overhead at all by avoiding `ChunkRequestTask`, since you're still handing off the packet to `CompressBatchTask` anyway. So we still incur an async task performance penalty, and there's no obvious memory savings either.

All in all, I'm not convinced by this change.

dktapps:
Closing this as performance testing suggests it doesn't really make sense. Not to say that the performance couldn't be improved, but this doesn't seem to be the way to do it.

PR #6481, opened 2024-11-03
feat: add Crimson Nylium and Warped Nylium blocks

AzaleeX:
## Introduction
`Crimson Nylium` and `Warped Nylium` blocks added
image

## Tests

I tested this PR by doing the following (tick all that apply):
- [x] Writing PHPUnit tests (commit these in the `tests/phpunit` folder)
- [x] Playtesting using a Minecraft client (provide screenshots or a video)
- [x] Writing a test plugin (provide the code and sample output)
- [x] Other (provide details)

PR #6905, opened 2025-11-20
Fully generated registry accessor classes

dktapps:
The functions generated by these are significantly faster than their RegistryTrait callStatic counterparts (about 40%). This new system also avoids modifying handwritten code, and prevents the introduction of members that aren't supposed to be there via reflection.

Unfortunately, reflection can probably still modify the registry fields, but this new approach should make it impossible to do dodgy stuff like overriding registry members with wrong types, which should eliminate a lot of crashes caused by plugins doing stuff they shouldn't.

Registries now work like so:

- A source class is declared, which contains a function returning an `array<string, ?>` of members to register. The source class must be included in the build, since it will be linked against by the generated class. (This avoids a huge amount of complexity that originally discouraged me from attempting this change.)
- Source class declares the `#[RegistrySource(...)]` attribute
- `build/generate-registry-interface.php` generates a class with a hardwired function generated for every member, giving fast access
- The `generated` directory contains the generated classes

Advantages:
- Significantly faster than `RegistryTrait::__callStatic()`
- Reflection can't make registry members have wrong types and blow stuff up anymore
- Reflection can't add new members that aren't part of the definition
- Gencode now fully separated from manually written code
- Doxygen should be able to understand these functions instead of generating [this giant mess](https://apidoc.pmmp.io/df/dec/classpocketmine_1_1block_1_1_vanilla_blocks.html)
- Optimisation-friendly - these generated functions can be inlined by OPcache and JIT-compiled for an even bigger performance gain

### API changes
Outwardly the API should be the same as before

### Behavioural changes
registries now work quite differently internally.

## Backwards compatibility
This will break plugins that were modifying `VanillaBlocks` etc using reflection.
However, this has never been supported behaviour anyway, so I'm not going to shed any tears for people whose plugins break because of this change.

A change was also required for Nylium as introduced in PR #6481, since `VanillaBlocksInputs` cannot use `VanillaBlocks` at registration time. This previously worked with `VanillaBlocks`, but it probably shouldn't have.

In the future, we could probably rewire `VanillaBlocksInputs::setup()` to use a `Generator` to allow this to work, but that's out of scope for now. Since Nylium was introduced on minor-next, we're OK to BC-break it before it makes it to a stable release.

## Follow-up
All generated code to be moved out of `src` into `generated`
Possibly not keeping gencode in the repo at all? although that would kinda suck for plugin devs searching the repo for API info

## Tests
Unit tests should be enough, I think?


Azvyl:
Ah, the Nylium part was actually my fault.
I should have seen why short grass also uses Closure to pass its tall variant.

dktapps:
It's fine, nobody else noticed it either :)

There are two approaches usually used for this sort of thing:
1) closure to delay resolution
2) keep the target blocks in local vars to allow them to be directly captured

I've never been a big fan of option 2 though, since it makes the code hard to follow and prevents blocks from being declared in alphabetical order.

dktapps:
Ok, turns out it's not just the VanillaBlocks/VanillaBlocksInputs dependency that's a problem...
VanillaItems and VanillaBlocks depend on each other, and VanillaItems also depends on VanillaArmorMaterials.

Since the script doesn't try to resolve dependencies of registries, it doesn't know which order to generate code in. So it will never be possible to just do `rm -rf generated` and regenerate the gencode from scratch. I don't see a good way to solve that.

This is a larger issue than the old `@method` approach, because newly added members would still automagically work although they wouldn't pass static analysis. This means it'll also be a pain in the ass to do stuff like adding new types of signs.

Previously, this was doable like so:
- Add the block, and reference the static method that will automagically work via VanillaItems::__callStatic
- Add the item, and reference the VanillaBlocks static method that will also automagically work
- Regenerate both docblocks

However, with full codegen, it'll now be like so:
- Add a sign block, but stub its item to air or another dummy, since it can't reference an item that isn't registered yet
- Generate VanillaBlocks
- Add the item that references newly generated stuff from VanillaBlocks
- Generate VanillaItems
- Update VanillaBlocks registration to reference the newly generated VanillaItems entry

This change is still worth doing, but it's not as clear of a W as I thought.

dktapps:
Latest commits should resolve the aforementioned issues with registry interdependence. `registerDelayed` can now be used to provide a callback that will only be invoked when the registry is setup, while providing type info only to the code generator. This allows the `generated` folder to be deleted and regenerated from scratch using the script, resolves circular dependencies, and also allows registry members to depend on other members in the same registry (e.g. nylium depending on fungus and other nether blocks), as long as their delayed registrations happen in the appropriate order.

I'm dicing with the idea of making this mode the default, but that's a decision for another time.

PR #6922, opened 2025-11-30
Implemented Composter

Its123Miguel321:
This PR implements the [Composter](https://minecraft.wiki/w/Composter)

### Related issues & PRs

Related to [#4742](https://github.com/pmmp/PocketMine-MP/pull/4742)

#### Block
* Added Block Classes: 
  - `Composter`
  - `Fern`
  - `LargeFern`
  - `NetherWart`
  - `PitcherPlant`
  - `ShroomLight`
  - `TorchFlower`
  - `WarpedWart`
* Updated: `Block` Class:
  - Added `getCompostabilityChance()`
  - Added `isCompostable()`
* Updated Blocks:
  - `BaseBigDripleaf`
  - `BaseCake`
  - `Cactus`
  - `CactusFlower`
  - `DoubleTallGrass`
  - `DriedKelp`
  - `Flower`
  - `GlowLichen`
  - `Grass`
  - `HangingRoots`
  - `HayBale`
  - `Leaves`
  - `MangroveRoots`
  - `Melon`
  - `MushroomStem`
  - `NetherRoots`
  - `NetherVines`
  - `NetherWartPlant`
  - `PinkPetals`
  - `Pumpkin`
  - `RedMushroom`
  - `RedMushroomBlock`
  - `Sapling`
  - `SeaPickle`
  - `SmallDripleaf`
  - `SporeBlossom`
  - `Stem`
  - `SugarCane`
  - `TallGrass`
  - `Vine`
  - `WaterLily`
  - `WitherRose`

#### Event
* Added Events:
  - `ComposterEmptyEvent`
  - `ComposterFillEvent`
  - `ComposterReadyEvent`

#### Item
* Added Item: `Wheat`
* Updated `Item` & `ItemBlock` Classes:
  - Added `getCompostabilityChance()`
  - Added `isCompostable()`
* Updated Items:
  - `Apple`
  - `BakedPotato`
  - `Beetroot`
  - `BeetrootSeeds`
  - `Bread`
  - `Carrot`
  - `CocoaBeans`
  - `Cookie`
  - `DriedKelp`
  - `GlowBerries`
  - `Melon`
  - `MelonSeeds`
  - `PitcherPod`
  - `Potato`
  - `PumpkinPie`
  - `PumpkinSeeds`
  - `SweetBerries`
  - `TorchflowerSeeds`
  - `WheatSeeds`

#### World
* Added Particle: `CropGrowthEmitterParticle`
* Added Sounds:
  - `ComposterEmptySound`
  - `ComposterFillSound`
  - `ComposterFillSuccessSound`
  - `ComposterReadySound`

## Follow-up
Once more items/blocks are added could check if they are compostable

## Tests
Composter functionality [Video](https://youtu.be/qOqxAlEeYM4)


src/block/Composter.php
```diff
@@ -0,0 +1,139 @@
...
+
+	protected function describeBlockOnlyState(RuntimeDataDescriber $w) : void{
+		$w->boundedIntAuto(0, self::MAX_COMPOST_LAYERS, $this->layers);
+	}
+	
+	protected function recalculateCollisionBoxes() : array{
+		foreach(Facing::HORIZONTAL as $f){
+			$result[] = AxisAlignedBB::one()->trim($f, 14 / 16);
+		}
+		return $result;
```
dktapps, line 63: Shouldn't this also have an AABB for the bottom of the composter, so entities can't fall through it?

src/block/Composter.php
```diff
@@ -0,0 +1,139 @@
...
+			);
+			$event->call();
+
+			if(!$event->isCancelled()){
+				$this->position->getWorld()->addSound($this->position, new ComposterFillSound());
+
+				if($event->getResult()){
+					$this->setCompostLayers($event->getNewFillLayer());
+					$this->position->getWorld()->setBlock($this->position, $this);
+					$this->position->getWorld()->addSound($this->position, new ComposterFillSuccessSound());
```
dktapps: This will generate the "fill" and "fill success" sounds together. Is that what you intended?

src/block/Composter.php
dktapps: This also needs to consider the number of layers

src/block/Composter.php
```diff
@@ -0,0 +1,139 @@
...
+		}
+
+		$this->layers = min(self::MAX_COMPOST_LAYERS, ++$this->layers);
+		$this->position->getWorld()->setBlock($this->position, $this);
+		$this->position->getWorld()->addSound($this->position, new ComposterReadySound());
+	}
+
+	public function getDropsForIncompatibleTool(Item $item) : array{
+		return [$this->asItem()];
+	}
```
dktapps: This seems unnecessary, the breakinfo harvest level is 0 so all tools are compatible anyway

src/block/Composter.php
```diff
@@ -0,0 +1,139 @@
...
+		}
+
+		if(!$item->isNull() && $item->isCompostable()){
+			$event = new ComposterFillEvent(
+				$this, 
+				$item, 
+				$this->layers,
+				$this->layers + 1, 
+				mt_rand(1, 100) <= $item->getCompostabilityChance() && $this->layers < self::MAX_COMPOST_LAYERS - 1
+			);
```
dktapps, line 102: Instead of having a `result` field, I think it'd be better to set the `new fill layers` property to be the same as the old if the chance of adding a layer wasn't met, which would be less confusing for plugin devs using `setNewFillLayers` (although I'm doubtful that's necessary either)

src/world/particle/CropGrowthEmitterParticle.php
```diff
@@ -0,0 +1,43 @@
...
+use pocketmine\math\Vector3;
+use pocketmine\network\mcpe\protocol\SpawnParticleEffectPacket;
+use pocketmine\network\mcpe\protocol\types\DimensionIds;
+
+class CropGrowthEmitterParticle implements Particle{
+
+	public const CROP_GROWTH_EMITTER_PARTICLE_ID = "minecraft:crop_growth_emitter";
+
+	public function encode(Vector3 $pos) : array{
+		return [SpawnParticleEffectPacket::create(
```
dktapps: There's probably a `LevelEventPacket` version of this which will take less bandwidth. I don't like using this packet because it's unnecessarily wasteful.

src/world/particle/CropGrowthEmitterParticle.php
```diff
@@ -0,0 +1,43 @@
...
+
+namespace pocketmine\world\particle;
+
+use pocketmine\math\Vector3;
+use pocketmine\network\mcpe\protocol\SpawnParticleEffectPacket;
+use pocketmine\network\mcpe\protocol\types\DimensionIds;
+
+class CropGrowthEmitterParticle implements Particle{
+
+	public const CROP_GROWTH_EMITTER_PARTICLE_ID = "minecraft:crop_growth_emitter";
```
dktapps: This is protocol info and should not be publicly exposed

src/world/particle/CropGrowthEmitterParticle.php
```diff
@@ -0,0 +1,43 @@
...
+
+declare(strict_types=1);
+
+namespace pocketmine\world\particle;
+
+use pocketmine\math\Vector3;
+use pocketmine\network\mcpe\protocol\SpawnParticleEffectPacket;
+use pocketmine\network\mcpe\protocol\types\DimensionIds;
+
+class CropGrowthEmitterParticle implements Particle{
```
dktapps: `CropGrowthParticle` would be less clunky

dktapps (changes requested):
This looks mostly OK, good work

However, some comments below

src/block/Composter.php
```diff
@@ -0,0 +1,139 @@
...
+	}
+	
+	protected function recalculateCollisionBoxes() : array{
+		foreach(Facing::HORIZONTAL as $f){
+			$result[] = AxisAlignedBB::one()->trim($f, 14 / 16);
+		}
+		return $result;
+	}
+
+	public function getCompostLayers() : int{ return $this->layers; }
```
kostamax27: Maybe the method names are a bit redundant. Since this is already inside the `Composter` class, the `Compost` prefix might not be necessary. `getLayers()` and `setLayers()` could be clearer and shorter.

src/block/Composter.php
Its123Miguel321: Sorry, but could you explain this a little bit more? I am a little bit confused lol.

dktapps: `$ev->setNewFillLevel($amount)` won't have the expected result if `$ev->getResult()` is `false`.
I'm saying, get rid of `$ev->getResult()` and just let people check the new fill level directly.

Then, `$ev->getOldFillLevel()` will be the same as `$ev->getNewFillLevel()` if no layer is to be added.

src/block/Composter.php
```diff
@@ -95,26 +95,29 @@ public function onInteract(Item $item, int $face, Vector3 $clickVector, ?Player
 				$this, 
 				$item, 
 				$this->layers,
-				$this->layers + 1, 
-				mt_rand(1, 100) <= $item->getCompostabilityChance() && $this->layers < self::MAX_COMPOST_LAYERS - 1
+				$this->layers
```
dktapps: Uh, this isn't what I meant. This just makes `getNewFillLayers` useless.

I meant that it should pass
`mt_rand(1, 100) <= $item->getCompostabilityChance() ? $this->layers + 1 : $this->layers`

dktapps:
Please make the build pass.
The CS issues are mostly cases of trailing whitespace or missing blank lines at the end of the file. `php-cs-fixer` can fix those for you.

I also don't like the idea of adding functions inside which the compostability gets hardcoded. I've been trying to move away from that generally, since it's inflexible and requires tons of useless classes. It'd be better if we had some way to inject this information into item constructors where it's appropriate. For Blocks, it could go in `BlockTypeInfo`. For Items I suppose it'd have to just go in the `Item` class directly...

PR #6923, opened 2025-11-30
Adjust collision box calculation for `SnowLayer` blocks

kostamax27:
### Behavioural changes
The collision height now correctly matches the number of snow layers.

## Backwards compatibility
Changes are backward-compatible, except in cases where code relied on the previous collision height for the minimal layer.

## Tests
This has been tested in-game.

Vanilla:
image

Before:
image

After:
image

### DIFF
--- a/src/block/SnowLayer.php
+++ b/src/block/SnowLayer.php
@@ -66,8 +66,10 @@ class SnowLayer extends Flowable implements Fallable{
 	}
 
 	protected function recalculateCollisionBoxes() : array{
-		//TODO: this zero-height BB is intended to stay in lockstep with a MCPE bug
-		return [AxisAlignedBB::one()->trim(Facing::UP, $this->layers >= 4 ? 0.5 : 1)];
+		if($this->layers === self::MIN_LAYERS){
+			return [];
+		}
+		return [AxisAlignedBB::one()->trim(Facing::UP, 1 - ($this->layers * 0.125))];
 	}
 
 	public function getSupportType(int $facing) : SupportType{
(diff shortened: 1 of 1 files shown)


kostamax27:
After testing the snow layers a bit more, I noticed one detail: the first layer also affects the height of entities, but it doesn’t seem to affect the player.

image

dktapps:
Fairly sure the old logic is correct for players, I tested it not that long ago

kostamax27:
> Fairly sure the old logic is correct for players, I tested it not that long ago

There is one problem with old snow layers:

https://github.com/user-attachments/assets/e38953a5-3aef-4d7e-a6c9-c411e5d73ebd

dktapps:
Right. I think we should mirror what happens with players and ignore items. With layers of 1, the AABB appears to have a height of 0 if I observed correctly.

`(self::MAX_LAYERS - $this->layers - 1) / 8` would probably be the correct logic, or something similar.

kostamax27:
> (self::MAX_LAYERS - $this->layers - 1) / 8

image

`1 - ($this->layers * 0.125)` *without checks for `self::MIN_LAYERS`*:

image

dktapps:
Oh, maybe that should've been +1

kostamax27:
> Oh, maybe that should've been +1

`(self::MAX_LAYERS - $this->layers + 1) / 8`?

image

dktapps:
That looks like it matches what your player is doing, no?
You can also test it in vanilla to make sure the result is the same

kostamax27:
ready?

dktapps (commented):
Looks good to me, although I've not had time to test it yet

dktapps (approved):

Issue #6915, opened 2025-11-24
Netherite armor should give 10% knockback resistance per piece


dktapps:
### Problem description

As per title

### Proposed solution

Implement this

PR #6927, opened 2025-12-03
Implement kb resistance for netherite armor

b1zeyofficial:
Related to https://github.com/pmmp/PocketMine-MP/issues/6915
This PR adds knockback resistance to netherite armor for each piece by 10%

## Tests
https://youtu.be/9K547LoWOq4


b1zeyofficial:
I didnt like that i was using ItemTypeIds to get netherite armor, maybe we can add an ArmorTier class or something like that? Any suggestions on this matter?

dktapps:
There is Armor->getMaterial() and VanillaArnorMaterials. But I'm not convinced this is the right way to implement it anyway

Wraith0x10:
> I didnt like that i was using ItemTypeIds to get netherite armor, maybe we can add an ArmorTier class or something like that? Any suggestions on this matter?

This is what I did when I implemented the kb reduction to my server
```php
    $reduction = 0.0;
    $armorInv = $player->getArmorInventory();

    foreach($armorInv->getContents() as $armor){
        /* @var Armor $armor */
        if($armor->getMaterial() === VanillaArmorMaterials::NETHERITE()){
            $reduction += 0.10;
        }
    }
```
    
Hope this helps!

src/entity/Living.php
```diff
@@ -617,6 +617,24 @@ public function knockBack(float $x, float $z, float $force = self::DEFAULT_KNOCK
...
+
+			$netheritePieces = 0;
+			foreach($armorInv->getContents() as $armor){
+				if($armor instanceof Armor){
+					if($armor->getMaterial() === VanillaArmorMaterials::NETHERITE()){
+						$netheritePieces++;
+					}
+				}
+			}
+			if($netheritePieces > 0){
```
Wraith0x10: This check isn’t needed. If the count is 0, it just multiplies by 1 anyway, so nothing happens.

src/entity/Living.php
```diff
@@ -617,6 +617,22 @@ public function knockBack(float $x, float $z, float $force = self::DEFAULT_KNOCK
 		if($f <= 0){
 			return;
 		}
+
+		//TODO: perhaps this also applies to regular entities?
```
dktapps: Most likely yes

src/entity/Living.php
```diff
@@ -617,6 +617,22 @@ public function knockBack(float $x, float $z, float $force = self::DEFAULT_KNOCK
...
+		if($this instanceof Player){
```
dktapps: Don't like this

src/entity/Living.php
```diff
@@ -617,6 +617,22 @@ public function knockBack(float $x, float $z, float $force = self::DEFAULT_KNOCK
...
+			$armorInv = $this->getArmorInventory();
+
+			$netheritePieces = 0;
+			foreach($armorInv->getContents() as $armor){
+				if($armor instanceof Armor){
+					if($armor->getMaterial() === VanillaArmorMaterials::NETHERITE()){
```
dktapps: This could be written in a single condition instead of two nested ifs

dktapps (commented):
I'm not sure this is the right way to implement it, but functionally it looks ok

kostamax27:
This implementation is far too limited :(

Wraith0x10:
> This implementation is far too limited :(

I agree. 
Perhaps we should add a `knockbackResistance` field to `ArmorTypeInfo` rather than hardcoding to just netherite in `Living->knockback`

b1zeyofficial:
closed to https://github.com/pmmp/PocketMine-MP/pull/6930

PR #6928, opened 2025-12-05
Implemented Crossbow

### LINKED ISSUE #6864: Implement Crossbows
### Problem description

so we can enjoy shooting crossbows

### Proposed solution

I will be making a PR to hopefully add this feature, but if i cant figure it out, a issue is opened for someone else to maybe add it.

i have currently fully implemented crossbows, with their enchantments (not piercing yet). however i have one issue, for crossbows, they are `Chargeable` and i didnt seem to find a interface that deals with chargeable items. though items like Bows, or Tridents, but those arent really chargeable, theyre more of like pull and automatically shoot. however crossbows are different, and need a ticking mechanism to check when to stop the charging process to either "Load" it, or bail out if not held for long enough.

so what im saying is, the crossbow will implement `Releasable` & `Chargeable` (new interface). and in the Player or Human class, perhaps on entity base tick, if the `Player` is using an item `isUsingItem` returns true, and if the item is an `instanceof Chargeable`, itll call some sort of `tickCharge` or `continueCharge` method which will be in the chargeable interface class.

and in those functions, we can have code to check, wether `getItemUseDuration` exceeds the needed amount of ticks to load a crossbow, if so, stop charging and give a charged crossbow

im about to explore what said above ^ (if @dktapps agrees or gives a better solution?)

alvin0319:
This PR implements crossbow item with two additional vanilla enchantments for crossbow.

### Related issues & PRs
Fixes #6864

This PR includes logics from [dragonfly](github.com/df-mc/dragonfly/blob/6f0cd108589982e42044c09b188e79f2449bbc61/server/item/crossbow.go).

### API changes
* Introduced `Chargeable` interface: Implemented by items that can be charged. (e.g., crossbow)
* Added `VanillaEnchantments::QUICK_CHARGE()` and `VanillaEnchantments::MULTISHOT()`.
* Added `EntityShootCrossbowEvent`.
* Added `CrossbowLoadSound` and `CrossbowShootSound`.

### Behavioural changes
* `Player::releaseHeldItem()` now handles `Chargeable` items when player is using an item.

## Backwards compatibility
No breaking changes.

## Follow-up
* Implement support for loading and shooting Firework Rockets.

## Tests
Manual playtest have been done, attaching video below.

  https://github.com/user-attachments/assets/f4ec9a71-61e2-4fcb-9e19-a16ff7fb507c


remminiscent:
exact system i had in mind. great work.

kostamax27:
https://minecraft.fandom.com/wiki/Enchanting#Summary_of_enchantments

| Enchantment        | Primary / Secondary |
|--------------------|---------------------|
| Quick Charge       | Primary             |
| Multishot          | Primary             |
| Piercing           | Primary             |
| Unbreaking         | Secondary           |
| Mending            | Secondary           |
| Curse of Vanishing | Secondary           |

kostamax27:
Oops. `Unbreaking` in `Primary`...

alvin0319:
I reverted that since ItemEnchantmentTagRegistry covers `CROSSBOW` in `WEAPON`

src/item/VanillaItems.php
```diff
@@ -496,6 +497,7 @@ protected static function setup() : void{
 		self::register("coral_fan", fn(IID $id) => new CoralFan($id));
 		self::register("crimson_sign", fn(IID $id) => new ItemBlockWallOrFloor($id, Blocks::CRIMSON_SIGN(), Blocks::CRIMSON_WALL_SIGN()));
 		self::register("crimson_hanging_sign", fn(IID $id) => new HangingSign($id, "Crimson Hanging Sign", Blocks::CRIMSON_CEILING_CENTER_HANGING_SIGN(), Blocks::CRIMSON_CEILING_EDGES_HANGING_SIGN(), Blocks::CRIMSON_WALL_HANGING_SIGN()));
+		self::register("crossbow", fn(IID $id) => new Crossbow($id, "Crossbow"));
```
kostamax27: `self::register("crossbow", fn(IID $id) => new Crossbow($id, "Crossbow", [Tags::CROSSBOW]));`

src/item/Crossbow.php
```diff
@@ -0,0 +1,207 @@
...
+
+					$entity = new ArrowEntity(Location::fromObject(
+						$player->getEyePos(),
+						$player->getWorld(),
+						($arrowYaw > 180 ? 360 : 0) - $arrowYaw,
+						-$arrowPitch
+					), $player, true);
+					$entity->setOwningEntity($player);
+
+					if($i !== 1 || $player->isCreative(true)){
```
dktapps: `hasFiniteResources()` should be used here

src/item/Crossbow.php
```diff
@@ -0,0 +1,207 @@
...
+	private ?Item $chargedItem = null;
+
+	public function continueCharge(Player $player, int $useDuration) : bool{
+		$quickCharge = $this->getEnchantmentLevel(VanillaEnchantments::QUICK_CHARGE());
+		if($useDuration <= 1){
+			$player->getWorld()->addSound($player->getPosition(), new CrossbowLoadSound(CrossbowLoadSound::LOADING_START, $quickCharge > 0));
+			return false;
+		}
+		$chargeDuration = $this->getChargeDuration();
+		$multiplier = 25.0 / $chargeDuration;
```
dktapps, line 57: This could result in a DivisionByZeroError

src/player/Player.php
```diff
@@ -1711,7 +1719,9 @@ public function useHeldItem() : bool{
 		$this->resetItemCooldown($oldItem);
 		$this->returnItemsFromAction($oldItem, $item, $returnedItems);
 
-		$this->setUsingItem($item instanceof Releasable && $item->canStartUsingItem($this));
+		$shouldStartUsing = ($item instanceof Releasable && $item->canStartUsingItem($this)) ||
+			($item instanceof Chargeable && $result !== ItemUseResult::SUCCESS && $item->canStartChargingItem($this));
```
dktapps: I don't like this. We could just as easily implement Releasable on Crossbow and achieve the exact same result

src/network/mcpe/handler/InGamePacketHandler.php
```diff
@@ -510,7 +511,9 @@ private function handleUseItemTransaction(UseItemTransactionData $data) : bool{
 				return true;
 			case UseItemTransactionData::ACTION_CLICK_AIR:
 				if($this->player->isUsingItem()){
-					if(!$this->player->consumeHeldItem()){
+					if($this->player->getInventory()->getItemInHand() instanceof Chargeable){
+						$this->player->releaseHeldItem();
```
dktapps: This seems sus. Normally this would go through ReleaseItemTransaction for similar items

src/item/Crossbow.php
```diff
@@ -0,0 +1,207 @@
...
+			return ItemUseResult::FAIL;
+		}
+		$inventory?->removeItem($arrow);
+		$this->setCharged($arrow);
+		return ItemUseResult::SUCCESS;
+	}
+
+	public function getChargeDuration() : int{
+		$quickChargeLevel = $this->getEnchantmentLevel(VanillaEnchantments::QUICK_CHARGE());
+		return 25 - (5 * $quickChargeLevel);
```
dktapps: This could return 0 if the enchantment level is 5, which would crash in other parts of the code
It could also return a negative result

src/player/Player.php
```diff
@@ -1530,6 +1531,13 @@ public function onUpdate(int $currentTick) : bool{
 				$this->fireTicks = 1;
 			}
 
+			$chargeable = $this->getInventory()->getItemInHand();
+			if($chargeable instanceof Chargeable && $this->isUsingItem()){
+				if($chargeable->continueCharge($this, $this->getItemUseDuration())){
```
dktapps: I feel like `continueCharge` is weirdly specific. We probably need this for general `Releasable` anyway if we want to make eating sounds server-sided. So it seems sus to make a special case for this

src/network/mcpe/NetworkSession.php
```diff
@@ -1350,6 +1352,16 @@ public function onItemCooldownChanged(Item $item, int $ticks) : void{
...
 
+	public function onChargeItemComplete() : void{
+		if($this->player !== null){
+			$this->sendDataPacket(ActorEventPacket::create(
+				$this->player->getId(),
+				ActorEvent::CHARGED_ITEM,
+				0
+			));
+		}
+	}
```
dktapps, line 1366: Is this actually necessary? What does it do?

src/item/Crossbow.php
```diff
@@ -0,0 +1,207 @@
...
+	public function setCharged(?Item $item) : void{
+		$this->chargedItem = $item !== null ? clone $item : null;
+	}
+
+	public function isCharged() : bool{
+		return $this->chargedItem !== null;
+	}
+
+	public function onClickAir(Player $player, Vector3 $directionVector, array &$returnedItems) : ItemUseResult{
+		if($this->chargedItem !== null){
```
dktapps: This condition could be inverted to avoid unnecessary indentation

src/item/Crossbow.php
```diff
@@ -0,0 +1,207 @@
...
+		$arrow = VanillaItems::ARROW();
+		$inventory = match(true){
+			$player->getOffHandInventory()->contains($arrow) => $player->getOffHandInventory(),
+			$player->getInventory()->contains($arrow) => $player->getInventory(),
+			default => null
+		};
+		if($player->hasFiniteResources() && $inventory === null){
+			return ItemUseResult::FAIL;
+		}
+		$inventory?->removeItem($arrow);
```
kostamax27: `$player->hasFiniteResources()`

kostamax27:
https://github.com/user-attachments/assets/f370b7d7-9fbc-4042-aede-339a2e14f713

src/network/mcpe/NetworkSession.php
alvin0319: The client doesn't seem to notify the server whether crossbow was charged or not (they don't send ReleaseItem unless this packet was sent), so we need to notify client we have charged the crossbow.

src/item/Crossbow.php
alvin0319: ~~Impossible since chargeDuration never be 0 (getChargeDuration() never returns 0 unless some other weird factor applies)~~ nvm didn't see comments below

src/network/mcpe/handler/InGamePacketHandler.php
alvin0319: Maybe Mojang inconsistency...

It appears that canceling crossbow loading does fire ReleaseItemTransaction but receiving notify from client fires UseItemTransaction instead...

alvin0319:
> sound.mp4

I'm not sure what it means? is it some kind of sound desync?

src/player/Player.php
alvin0319:
```php
public function continueUsing(Player $player, int $useDuration) : bool{
    return false;
}
```
Maybe add stuff like this in Item so that we can generalize them? I think we cannot add things like this in Releaseable since it would require breaking changes

kostamax27:
> > sound.mp4
> 
> I'm not sure what it means? is it some kind of sound desync?

A similar issue exists on dragonfly - but it’s even worse there. The server keeps spamming the crossbow sound, and it’s not just me who hears it; nearby players hear it as well.

alvin0319:
> > > sound.mp4
> > 
> > 
> > I'm not sure what it means? is it some kind of sound desync?
> 
> A similar issue exists on dragonfly - but it’s even worse there. The server keeps spamming the crossbow sound, and it’s not just me who hears it; nearby players hear it as well.

Can you try with latest commit, it should *fix* that temporary (although I don't like this fix, need to find proper solution)

kostamax27:
In the video, you can see my inputs in the bottom-left corner, but I'll outline the steps just in case:
1. Equip the crossbow.
2. Charge the crossbow (hold right mouse button) and click the left mouse button.

https://github.com/user-attachments/assets/e53db10d-9c7f-4649-a975-e0fa48423390

alvin0319:
> In the video, you can see my inputs in the bottom-left corner, but I'll outline the steps just in case:
> 
> 1. Equip the crossbow.
> 2. Charge the crossbow (hold right mouse button) and click the left mouse button.
> 
>  sound.mp4

I can't reproduce it in current environment where I have no PC client to test... and sound should be fixed with latest commit. Can you verify you are using latest commit using `/version` command?

kostamax27:
> > In the video, you can see my inputs in the bottom-left corner, but I'll outline the steps just in case:
> > 
> > 1. Equip the crossbow.
> > 2. Charge the crossbow (hold right mouse button) and click the left mouse button.
> > 
> > sound.mp4
> 
> I can't reproduce it in current environment where I have no PC client to test... and sound should be fixed with latest commit. Can you verify you are using latest commit using `/version` command?

```log
Command output | This server is running PocketMine-MP
Command output | Server version: 5.37.4+dev (git hash: 31ec74ea561df94ed3dfa1a5261a724d0fe51417)
Command output | Compatible Minecraft version: 1.21.124 (protocol version: 860)
Command output | PHP version: 8.4.12
Command output | PHP JIT: enabled (CRTO: 1255)
Command output | Operating system: linux
```

kostamax27:
> I have no PC client to test.


https://github.com/user-attachments/assets/5a84e4cd-d8fc-40bc-8ce9-99fe8e8e81ec

src/network/mcpe/handler/InGamePacketHandler.php
dktapps: > receiving notify from client fires UseItemTransaction instead...

No idea what you mean by this

src/player/Player.php
dktapps: Yeah, it would be BC breaking. Possibly a second interface `Releasable2 extends Releasable` would solve that issue. But considering we have `onReleaseUsing()` in `Item` anyway, perhaps your solution is fine.

src/item/Crossbow.php
```diff
@@ -0,0 +1,209 @@
...
+		$this->setCharged($arrow);
```
dktapps: Something seems fishy about this. There's no release action involved with fully charging a crossbow. I think this should be called in `continueUsing()` when the use duration >= the charge duration.

I wouldn't expect to see release used unless the player stopped charging the crossbow.

src/player/Player.php
```diff
@@ -1711,7 +1719,10 @@ public function useHeldItem() : bool{
 		$this->resetItemCooldown($oldItem);
 		$this->returnItemsFromAction($oldItem, $item, $returnedItems);
 
-		$this->setUsingItem($item instanceof Releasable && $item->canStartUsingItem($this));
+		$shouldStartUsing = $item instanceof Chargeable
+			? ($result !== ItemUseResult::SUCCESS && $item->canStartUsingItem($this))
```
dktapps: Why is this needed?

dktapps (commented):
I don't think the network actions for crossbow loading are handled correctly.
My gut instinct is that:

- left-click-air should begin loading the crossbow
- continue use should finish loading when the use duration >= the required load duration
- release should only be used if the player aborts loading the crossbow

src/player/Player.php
alvin0319: This prevents auto-reloading crossbow after shooting them, not sure why it works like that.

src/item/Crossbow.php
alvin0319: I have tested your solution, but making item charged and setting item back in `continueUsing()` didn't seem work - crossbow keeps loading fail, with no debug logs for mismatched or some kind of them in console.

src/network/mcpe/handler/InGamePacketHandler.php
alvin0319: Client sends UseItemTransaction to let server know crossbow charge was finished on client side.

alvin0319:
Here's BDS's MITM packet [log.txt](https://github.com/user-attachments/files/24103635/log.txt) for those who are interested in, I'll be digging into this more in coming weekend.

src/player/Player.php
dktapps: To be fair, I suppose we should only be activating the releasable action if there was no result from the right-click-air action generally. So maybe just apply this condition to all releasables.

Also it should check for NONE, since we probably don't want to start using the item if the use result was FAIL (which usually means cancelled by a plugin)

src/network/mcpe/NetworkSession.php
```diff
@@ -1352,13 +1353,17 @@ public function onItemCooldownChanged(Item $item, int $ticks) : void{
...
 		if($this->player !== null){
 			$this->sendDataPacket(ActorEventPacket::create(
 				$this->player->getId(),
 				ActorEvent::CHARGED_ITEM,
 				0
 			));
+			$this->sendDataPacket(CompletedUsingItemPacket::create(
+				$this->typeConverter->coreItemStackToNet($item)->getId(),
+				CompletedUsingItemPacket::ACTION_UNKNOWN // BDS sends this as UNKNOWN, why?
+			));
```
dktapps: I doubt this is necessary, it's only for telemetry

kostamax27:
First, I want to point out that `CompletedUsingItemPacket` truly has no effect on the behavior.
I managed to achieve the "vanilla" crossbow behavior by doing the following:

1. Remove: `$player->setUsingItem(false);`
from `Crossbow::onReleaseUsing`.
2. In `Crossbow::continueUsing`, when `$useDuration >= $chargeDuration`, replace `return true;`
with:
`return $player->releaseHeldItem();`

I understand this isn’t the correct or intended approach, but hopefully it gives you an idea in the right direction.

https://github.com/user-attachments/assets/a4e2339c-f25a-4f82-a1bb-1f2738531d6a


https://github.com/user-attachments/assets/c61770b1-b118-43e3-86b8-bab61a7b51f5

src/item/Crossbow.php
dktapps: It's because you need to set the changed item back into the inventory. onReleaseUsing normally handles that automatically.

That being said, server-side duration charging doesn't seem like a good idea on second thought. Since the player might have a lag spike and stop charging during the lag spike, which would then have unintended consequences.

I still don't think onReleaseUsing() is right for this though. Since release implies that the player let go of the use-item button, but what's actually happening here is that the use duration exceeds a threshold. Same applies to food, potions and milk buckets.

alvin0319:
I have pushed commits that address your comments, it seems this approach does better job than the original though.

src/item/Crossbow.php
```diff
@@ -62,6 +62,23 @@ public function continueUsing(Player $player, int $useDuration) : bool{
...
+				$inventory = match(true){
+					$player->getOffHandInventory()->contains($arrow) => $player->getOffHandInventory(),
+					$player->getInventory()->contains($arrow) => $player->getInventory(),
+					default => null
+				};
+				if($player->hasFiniteResources() && $inventory === null){
+					// No ammo available, charging cannot complete
+					return false;
+				}
+				$inventory?->removeItem($arrow);
```
kostamax27: Additionally, `Player->hasFiniteResources` must be added.
If this is not done, arrows will disappear for players in `Creative mode`.
See the `Bow` implementation:
https://github.com/pmmp/PocketMine-MP/blob/5d77a46d364dab48b40999054615ccb32945644e/src/item/Bow.php#L117-L122

src/player/Player.php
```diff
@@ -1544,6 +1544,15 @@ public function onUpdate(int $currentTick) : bool{
 				$this->fireTicks = 1;
 			}
 
+			$item = $this->getInventory()->getItemInHand();
+			if($item instanceof Releasable && $this->isUsingItem()){
+				$oldItem = clone $item;
+				if($item->continueUsing($this, $this->getItemUseDuration())){
+					$this->getNetworkSession()->onChargeItemComplete();
+					$this->returnItemsFromAction($oldItem, $item, []);
```
dktapps: We should probably do this irrespective of the result of continueUsing(). For example, a custom item might have the crossbow string snap and break the item during usage, which wouldn't count as charging it, but would still require the inventory to be updated.

src/item/Crossbow.php
dktapps: Infinity also doesn't seem to be checked here, although I don't know if it applies to crossbows

kostamax27: > Infinity also doesn't seem to be checked here, although I don't know if it applies to crossbows

Infinity cannot be applied to a crossbow, but the `Player->hasFiniteResources` check should remain.

dktapps:
> I have pushed commits that address your comments, it seems this approach does better job than the original though.

Not too sure about that. The issue is that the player may release the crossbow before the time is up, aborting the load, but the server might not get the action until after the crossbow is already loaded server-side. That then leads to the crossbow being unexpectedly loaded when the client's connection catches up. I mentioned this previously and also in #6958.

Issue #6716, opened 2025-05-29
Respawn Anchor explodes underwater


ItzxDwi:
### Plugin information

I haven't tested without plugins

### Related issues & PRs
* Related to #6646 

### Vanilla Behaviour:
https://youtu.be/0H3BrTl65f4

### PMMP:
https://youtu.be/MoY3Z8XVDVs

### Steps to reproduce the issue
1. Give yourself respawn anchor & glowstone
2. Place respawn anchor underwater
3. Charge respawn anchor with glowstone
4. Respawn anchor explodes underwater

### Expected behaviour

Respawn Anchor should **NOT** explode underwater!

### PocketMine-MP version

5.28.3+dev (latest minor-next)

### PHP version

8.2.27

### Server OS

Android

### Game version (if applicable)

Android

dktapps:
Source?

ItzxDwi:
> Source?

### Vanilla Behaviour:
https://youtu.be/0H3BrTl65f4

### PMMP:
https://youtu.be/MoY3Z8XVDVs

dktapps:
That's weird. Are you sure that isn't a bug in vanilla?

ItzxDwi:
> That's weird. Are you sure that isn't a bug in vanilla?

I tested it again in vanilla Bedrock, and the Respawn Anchor does **not** explode underwater. This seems to be a bug in PR #6646 , not vanilla bug.

dktapps:
Better question would be what happens in Java. This behaviour is not mentioned in the wiki.

ItzxDwi:
> Better question would be what happens in Java. This behaviour is not mentioned in the wiki.

I can't test it in Java since I don't have a PC, but in Bedrock, they don't explode underwater. So it seems like this is a bug related to PR #6646.

b1zeyofficial:
I confirm, this is not observed in vanilla, I checked it only in bedrock edition
It looks like we need to implement something like worksUnderwater, no?
Like in TNT

b1zeyofficial:
It would be worth noting that in vanilla minecraft bedrock edition (1.21.82) the block itself explodes but blocks underwater do not explode...
Correct me ItzxDwi

ItzxDwi:
> It would be worth noting that in vanilla minecraft bedrock edition (1.21.82) the block itself explodes but blocks underwater do not explode... Correct me ItzxDwi

[...]

PR #6929, opened 2025-12-07
Fixed respawn anchor explode blocks underwater

b1zeyofficial:
As per title

### Related issues & PRs
 - Fixes https://github.com/pmmp/PocketMine-MP/issues/6716

### Behavioural changes
i think the servers behavior shouldnt change, but need to check

## Tests
https://www.youtube.com/watch?v=FK-LLQQMVEo


src/block/RespawnAnchor.php
```diff
@@ -110,14 +111,28 @@ private function explode(?Player $player) : void{
...
+			return true;
+		}
+
+		foreach($block->getHorizontalSides() as $sideBlock){
+			if($sideBlock->getTypeId() === BlockTypeIds::WATER){
+				return true;
+			}
+		}
+		return false;
+	}
```
dktapps, line 153: Why isn't this checking the down face?

b1zeyofficial: I seem to have forgotten that the Java code checks for water under each horizontal block, since the existing code is similar to vanilla behavior.

P.S:
Apparently we should only check for horizontal blocks and not the Respawn Anchor itself, im confused
```java
	private static boolean isWaterThatWouldFlow(BlockPos blockPos, Level level) {
		FluidState fluidState = level.getFluidState(blockPos);
		if (!fluidState.is(FluidTags.WATER)) {
			return false;
		} else if (fluidState.isSource()) {
			return true;
		} else {
			float f = fluidState.getAmount();
			if (f < 2.0F) {
				return false;
			} else {
				FluidState fluidState2 = level.getFluidState(blockPos.below());
				return !fluidState2.is(FluidTags.WATER);
			}
		}
	}
	
	private void explode(BlockState blockState, Level level, BlockPos blockPos) {
		level.removeBlock(blockPos, false);
		boolean bl = Direction.Plane.HORIZONTAL.stream().map(blockPos::relative).anyMatch(blockPosx -> isWaterThatWouldFlow(blockPosx, level));
```

b1zeyofficial:
It doesnt really mimic vanilla behavior, for example, when water flows and a respawn anchor is placed in place, it doesnt explode blocks like it does in vanilla. Can we measure the fluidity of water using the Liquid->getDecay method?

dktapps:
Yes, but decay is opposite to the fluid height. 0 is source, 7 is the shallowest.

b1zeyofficial:
everything remained the same, only the style changed and a couple of changes, it seems like I check that decay >= 1 but it still explodes after only 2 streams, maybe I miscalculated somewhere? 
I also removed `instanceof Liquid` because we dont want lava to be able to somehow muffle the anchor's explosion, do we?

PR #6930, opened 2025-12-07
Implement Knockback Resistance for Armor

### LINKED ISSUE #6915: Netherite armor should give 10% knockback resistance per piece

Wraith0x10:
This PR introduces a new `knockbackResistance` field to `ArmorTypeInfo` allowing different armor types to apply different levels of knockback reduction.

### Related issues & PRs
- Fixes [#6915](https://github.com/pmmp/PocketMine-MP/issues/6915)

### API changes
- Added `Living->getKnockbackReduction`
- Added `Armor->getKnockbackResistance`, `ArmorTypeInfo->getKnockbackResistance`, `Item->getKnockbackResistance`
- Added a new `knockbackResistance` field to `ArmorTypeInfo`

### Behavioural changes
- Netherite armor now reduces knockback by 10% per piece.

## Tests
Tested in game and everything works as expected. My recording software is currently not working.


b1zeyofficial:
This PR will be more flexible than the last one, im closing my PR

src/item/VanillaItems.php
```diff
@@ -709,29 +709,29 @@ private static function registerArmorItems() : void{
...
 		self::register("golden_leggings", fn(IID $id) => new Armor($id, "Golden Leggings", new ArmorTypeInfo(3, 106, ArmorInventory::SLOT_LEGS, material: ArmorMaterials::GOLD()), [EnchantmentTags::LEGGINGS]));
 		self::register("iron_leggings", fn(IID $id) => new Armor($id, "Iron Leggings", new ArmorTypeInfo(5, 226, ArmorInventory::SLOT_LEGS, material: ArmorMaterials::IRON()), [EnchantmentTags::LEGGINGS]));
 		self::register("leather_pants", fn(IID $id) => new Armor($id, "Leather Pants", new ArmorTypeInfo(2, 76, ArmorInventory::SLOT_LEGS, material: ArmorMaterials::LEATHER()), [EnchantmentTags::LEGGINGS]));
-		self::register("netherite_leggings", fn(IID $id) => new Armor($id, "Netherite Leggings", new ArmorTypeInfo(6, 556, ArmorInventory::SLOT_LEGS, 3, true, material: ArmorMaterials::NETHERITE()), [EnchantmentTags::LEGGINGS]));
+		self::register("netherite_leggings", fn(IID $id) => new Armor($id, "Netherite Leggings", new ArmorTypeInfo(6, 556, ArmorInventory::SLOT_LEGS, 3, true, material: ArmorMaterials::NETHERITE(), knockbackResistance: 0.1), [EnchantmentTags::LEGGINGS]));
```
dktapps: This could be a property of the armor material, rather than having to write it for every piece

src/entity/Living.php
```diff
@@ -614,6 +630,7 @@ public function knockBack(float $x, float $z, float $force = self::DEFAULT_KNOCK
 		if($f <= 0){
 			return;
 		}
+		$force *= 1 - 0.1 * $this->getKnockbackReduction();
```
dktapps: This seems weird. If the entity is wearing 4 pieces of netherite armor, getKnockbackReduction() will return 0.4. But this then multiplies it by 0.1, so the multiplier will be `1 - 0.04`, which is basically nothing.

Not sure that's what it should be doing?

src/item/VanillaItems.php
Wraith0x10: Surely this would be better for users to be able to apply different knockback reduction to each individual armor piece rather than the material. Especially when creating custom armor via a plugin. Lets say if someone wanted their boots to reduce 0.2 but their chestplate 0.5

src/entity/Living.php
Wraith0x10: This was my bad, forgot to update this line

src/item/VanillaItems.php
dktapps: Hm, fine

kostamax27:
What about allowing this logic to be disabled or configured? For example, PvP servers are unlikely to need such a system. And how can developers manage the resistance value without having to register or replace armor?

Wraith0x10:
Ready?

dktapps:
> What about allowing this logic to be disabled or configured? For example, PvP servers are unlikely to need such a system. And how can developers manage the resistance value without having to register or replace armor?

Good question.
On a semi related note, I wonder if the knockback alteration from this is supposed to show up in the knockback_resistance attribute. Maybe someone can pull up a MITM trace or a proxy and find out

Wraith0x10:
> What about allowing this logic to be disabled or configured? For example, PvP servers are unlikely to need such a system. And how can developers manage the resistance value without having to register or replace armor?

Do you have a proposal for this? @kostamax27

Wraith0x10:
> Good question. On a semi related note, I wonder if the knockback alteration from this is supposed to show up in the knockback_resistance attribute. Maybe someone can pull up a MITM trace or a proxy and find out

I don't think it does, but I can check It out soon.

kostamax27:
> > What about allowing this logic to be disabled or configured? For example, PvP servers are unlikely to need such a system. And how can developers manage the resistance value without having to register or replace armor?
> 
> Do you have a proposal for this? @kostamax27

It might be an event, but I don’t think it’s a good decision.

Wraith0x10:
> It might be an event, but I don’t think it’s a good decision.

It could be handled as an event, but I don’t think that’s a good idea.

I’m not a fan of introducing an event for this, but I’m also not seeing a clean alternative yet. One option could be adding an optional bool param to `setKnockback()` for enabling/disabling knockback reduction.

dktapps:
I don't think any extra APIs are needed as long as the knockback attribute is able to affect it.

Side note: if someone is really bored and wants to try implementing attribute modifiers, that's been on the TODO list for ages and would help with this

PR #6932, opened 2025-12-07
Implement EntityTotemConsumeEvent

Wraith0x10:
This PR introduces a new event, `EntityResurrectEvent`, which is triggered when a player is saved by a Totem of Undying.

### API Changes
- Added `EntityResurrectEvent`.

## Tests
This test showcases how to prevent the Totem from activating when a player is attacked by another entity. Everything works as expected.

```php
public function onEntityResurrect(EntityResurrectEvent $event): void {
    $cause = $event->getCause();
    var_dump("Cause: $cause");
    
    if ($cause === EntityDamageEvent::CAUSE_ENTITY_ATTACK) {
        $event->cancel();
    }
}
```


src/entity/Human.php
```diff
@@ -385,6 +386,12 @@ public function applyDamageModifiers(EntityDamageEvent $source) : void{
 
 			$compensation = $this->getHealth() - $source->getFinalDamage() - 1;
 			if($compensation <= -1){
+				$ev = new EntityResurrectEvent($this, $type);
```
dktapps: I don't feel like this achieves the objective we discussed. As I said the totem action can be prevented by setting the totem modifier to zero in EntityDamageEvent. The real issue was that there wasn't a way to prevent the totem being consumed when it actually does activate.

Wraith0x10: Alright, I've moved the event call into the totem logic

dktapps:
Perhaps EntityResurrectEvent isn't the right name for this.
If it's fired from post-damage effects, the totem has already prevented the entity's death. Cancelling the event at that point would prevent the totem from being consumed and from adding effects, but wouldn't prevent the entity from being saved by the totem, since the modifier for that is setup in pre-damage effects.

Wraith0x10:
> Perhaps EntityResurrectEvent isn't the right name for this. If it's fired from post-damage effects, the totem has already prevented the entity's death. Cancelling the event at that point would prevent the totem from being consumed and from adding effects, but wouldn't prevent the entity from being saved by the totem, since the modifier for that is setup in pre-damage effects.

Do you have any ideas? The only name I can think of is EntityTotemConsumeEvent

dktapps:
Yeah, that's the only thing I could think of too

Wraith0x10:
> Yeah, that's the only thing I could think of too

Gotcha, renamed it now, should be ready.

Wraith0x10:
Ready?

dktapps:
The code looks fine, I'm just hung up on the naming. I feel like it's easy to confuse this with PlayerItemConsumeEvent, particularly if we end up adding EntityItemConsumeEvent (we probably will, since it'd be needed for mobs eating & drinking). But I don't have a better idea what to call it.

Wraith0x10:
> The code looks fine, I'm just hung up on the naming. I feel like it's easy to confuse this with PlayerItemConsumeEvent, particularly if we end up adding EntityItemConsumeEvent (we probably will, since it'd be needed for mobs eating & drinking). But I don't have a better idea what to call it.

I don't have any other ideas of what to call it as of right now. EntityTotemConsumeEvent makes the most sense to me.

Issue #6653, opened 2025-03-15
Min use duration for Releasable


dktapps:
### Problem description

All releasable items need some minimum use time before release (or consume in the case of food). We currently don't have a general way to handle this. Bows currently do it themselves, and food items don't check it at all.

### Proposed solution

Add `Releasable::getMinUseDuration()` which would return the minimum use time in ticks. Then, use this to verify against `Player::getItemUseDuration()` in `Player::releaseHeldItem()` and `Player::consumeHeldItem()`.

Unfortunately this is BC breaking, unless we add a second `Releasable` interface.

PR #6957, opened 2025-12-23
Implement `Releaseable::getMinUseDuration()`

remminiscent:
Implemented a new function `getMinUseDuration` in the `Releaseable` interface which will return the minimum needed ticks to use the item for an action to occur.

Also, regarding the Bow, should it be 5 ticks, or 0?
### Related issues & PRs
* Fixes https://github.com/pmmp/PocketMine-MP/issues/6653

### API changes
* Added `Releaseable::getMinUseDuration`

## Backwards compatibility
Nope, It's BC Breaking. good for major-next branch

## Tests
Works 👍
https://streamable.com/2l8zru

### DIFF
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -1738,6 +1738,10 @@ class Player extends Human implements CommandSender, ChunkListener, IPlayer, Nev
 	public function consumeHeldItem() : bool{
 		$slot = $this->inventory->getItemInHand();
 		if($slot instanceof ConsumableItem){
+			if($this->getItemUseDuration() < $slot->getMinUseDuration()){
+				return false;
+			}
+
 			$oldItem = clone $slot;
 
 			$ev = new PlayerItemConsumeEvent($this, $slot);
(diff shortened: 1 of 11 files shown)


src/player/Player.php
```diff
@@ -1738,6 +1738,10 @@ public function useHeldItem() : bool{
 	public function consumeHeldItem() : bool{
 		$slot = $this->inventory->getItemInHand();
 		if($slot instanceof ConsumableItem){
+			if($this->getItemUseDuration() < $slot->getMinUseDuration()){
```
kostamax27, line 1741: It might be worth moving this to `PlayerItemConsumeEvent`, similar to how `hasItemCooldown` is handled below.

dktapps:
I'm not sure how I want to handle this tbh. Interfaces like this have been a constant pain in the ass for BC breaks.

Considering we already have `onReleaseUsing()` in `Item` anyway, and #6928 is adding `continueUsing()` to `Item` directly for similar reasons, we should probably make sure the approaches for this are consistent.

remminiscent:
> I'm not sure how I want to handle this tbh. Interfaces like this have been a constant pain in the ass for BC breaks.
> 
> 
> 
> Considering we already have `onReleaseUsing()` in `Item` anyway, and #6928 is adding `continueUsing()` to `Item` directly for similar reasons, we should probably make sure the approaches for this are consistent.

so do i /close this PR? ☹️

or, in my opinion, we have a parent `ChargeableItem` class for food, bows, crossbows, trident, etc. no need for bunch of interfaces like `Chargeable` & `Releaseable`

dktapps:
Bow would have to extend both Durable and Chargeable, that's not gonna work

remminiscent:
> Bow would have to extend both Durable and Chargeable, that's not gonna work

You're right. well, i'll move onto making other PR's until this gets a plan

PR #6960, opened 2025-12-24
Implemented Void Generator

remminiscent:
Adds a "void" type generator.
### Related issues & PRs
* Fixes https://github.com/pmmp/PocketMine-MP/issues/6869

## Changes
void gen 👍 

## Tests
make default level type to "void" and make pm generate a world, join, and see
image

### DIFF
--- a/src/world/generator/VoidGenerator.php
+++ b/src/world/generator/VoidGenerator.php
@@ -44,14 +44,16 @@ class VoidGenerator extends Generator{
 			$chunk->setSubChunk($y, new SubChunk(Block::EMPTY_STATE_ID, [], clone $biomeArray));
 		}
 
-		if($chunkX >= 15 && $chunkX <= 16 && $chunkZ >= 15 && $chunkZ <= 16){
-			for($x = 0; $x < Chunk::EDGE_LENGTH; $x++){
-				for($z = 0; $z < Chunk::EDGE_LENGTH; $z++){
-					$worldX = ($chunkX * Chunk::EDGE_LENGTH) + $x;
-					$worldZ = ($chunkZ * Chunk::EDGE_LENGTH) + $z;
-					if($worldX >= 248 && $worldX <= 263 && $worldZ >= 248 && $worldZ <= 263){
-						$chunk->setBlockStateId($x, 69, $z, VanillaBlocks::STONE()->getStateId()); // 16x16 stone platform surrounding 256 69 256
-					}
+		$stoneState = VanillaBlocks::STONE()->getStateId();
+		$cobbleState = VanillaBlocks::COBBLESTONE()->getStateId();
+
+		for($x = 0; $x < Chunk::EDGE_LENGTH; $x++){
+			for($z = 0; $z < Chunk::EDGE_LENGTH; $z++){
+				$worldX = ($chunkX * Chunk::EDGE_LENGTH) + $x;
+				$worldZ = ($chunkZ * Chunk::EDGE_LENGTH) + $z;
+
+				if($worldX >= -16 && $worldX <= 16 && $worldZ >= -16 && $worldZ <= 16){
+					$chunk->setBlockStateId($x, -64, $z, ($worldX === 0 && $worldZ === 0) ? $cobbleState : $stoneState);
 				}
 			}
 		}
--- a/src/world/generator/VoidGenerator.php
+++ b/src/world/generator/VoidGenerator.php
@@ -23,26 +23,27 @@ declare(strict_types=1);
 
 namespace pocketmine\world\generator;
 
+use pocketmine\block\Block;
 use pocketmine\block\VanillaBlocks;
 use pocketmine\data\bedrock\BiomeIds;
 use pocketmine\world\ChunkManager;
 use pocketmine\world\format\Chunk;
+use pocketmine\world\format\PalettedBlockArray;
+use pocketmine\world\format\SubChunk;
 
 class VoidGenerator extends Generator{
-	private Chunk $chunk;
-
 	public function __construct(int $seed, string $preset){
 		parent::__construct($seed, $preset);
-		$this->generateBaseChunk();
-	}
-
-	protected function generateBaseChunk() : void{
-		$this->chunk = new Chunk([], false);
-		$this->chunk->setBiomeId(0, 69, 15, BiomeIds::PLAINS);
 	}
 
 	public function generateChunk(ChunkManager $world, int $chunkX, int $chunkZ) : void{
-		$chunk = clone $this->chunk;
+		$chunk = new Chunk([], false);
+		
+		$biomeArray = new PalettedBlockArray(BiomeIds::PLAINS);
+		foreach($chunk->getSubChunks() as $y => $subChunk){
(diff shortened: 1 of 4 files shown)


src/world/generator/VoidGenerator.php
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	protected function generateBaseChunk() : void{
+		$this->chunk = new Chunk([], false);
+	}
+
+	public function generateChunk(ChunkManager $world, int $chunkX, int $chunkZ) : void{
+		$chunk = clone $this->chunk;
+		if($chunkX >= 15 && $chunkX <= 16 && $chunkZ >= 15 && $chunkZ <= 16){
+			for($x = 0; $x < 16; $x++){
```
dktapps:
```suggestion
			for($x = 0; $x < Chunk::EDGE_LENGTH; $x++){
```

src/world/generator/VoidGenerator.php
```diff
@@ -0,0 +1,60 @@
...
+				for($z = 0; $z < 16; $z++){
```
dktapps:
```suggestion
				for($z = 0; $z < Chunk::EDGE_LENGTH; $z++){
```

src/world/generator/VoidGenerator.php
```diff
@@ -0,0 +1,60 @@
...
+					$worldX = ($chunkX * 16) + $x;
```
dktapps:
```suggestion
					$worldX = ($chunkX * Chunk::EDGE_LENGTH) + $x;
```

src/world/generator/VoidGenerator.php
```diff
@@ -0,0 +1,60 @@
...
+					$worldZ = ($chunkZ * 16) + $z;
```
dktapps:
```suggestion
					$worldZ = ($chunkZ * Chunk::EDGE_LENGTH) + $z;
```

src/world/generator/VoidGenerator.php
```diff
@@ -0,0 +1,60 @@
...
+						$chunk->setBlockStateId($x, 69, $z, VanillaBlocks::STONE()->getStateId()); // 16x16 stone platform surrounding 256 69 256
+					}
+				}
+			}
+		}
+		$world->setChunk($chunkX, $chunkZ, $chunk);
+	}
+
+	public function populateChunk(ChunkManager $world, int $chunkX, int $chunkZ) : void{
+	}
```
dktapps, line 65: Put a //NOOP comment inside this function if it's supposed to be empty

src/world/generator/VoidGenerator.php
```diff
@@ -0,0 +1,60 @@
...
+class VoidGenerator extends Generator{
+	private Chunk $chunk;
+
+	public function __construct(int $seed, string $preset){
+		parent::__construct($seed, $preset);
+		$this->generateBaseChunk();
+	}
+
+	protected function generateBaseChunk() : void{
+		$this->chunk = new Chunk([], false);
```
dktapps: You should set biomes here

dktapps (changes requested):

remminiscent:
256 256 spawns you in between 4 chunks, does that mean i need to generate 4 base chunks?

dktapps:
I wouldn't do a base chunk at all. Just create a new one and do the appropriate work depending on the coordinate range available. No need to store anything since this doesn't need to copy paste terrain

src/world/generator/VoidGenerator.php
```diff
@@ -0,0 +1,63 @@
...
+		$this->chunk->setBiomeId(0, 69, 15, BiomeIds::PLAINS);
```
dktapps: This is only gonna set the biome for a single block. You need to fill all the subchunks with the biome you want 

You'll want to do `new PalettedBlockArray(BiomeIds::PLAINS)` and then fill an array of Subchunk instances providing clones of it. You can look at `Normal::generateChunk()` to see how this is done elsewhere

(yes, it's clunky, this code was never properly updated for a world with 3D biomes)

dktapps: (I was gonna suggest you take a look at Flat but it doesn't look like Flat sets the biomes at all. I wonder when that got broken...)

remminiscent: dont worry, will fix that next PR

remminiscent:
off-topic, but did MrCakeSlayer forward you a message about smth regarding a phpstorm OSS partnership? if not, you can apply pocketmine to be a OSS partner of JetBrains, which will grant contributors a free license of PHPStorm (if you havent noticed, i need it badly for PRs)

dktapps:
> off-topic, but did MrCakeSlayer forward you a message about smth regarding a phpstorm OSS partnership? if not, you can apply pocketmine to be a OSS partner of JetBrains, which will grant contributors a free license of PHPStorm (if you havent noticed, i need it badly for PRs)

I'm aware of the program, but never applied for it since I've pretty much always been making a bit of money working on servers and I wasn't sure how that would interact with the partnership. Easy enough for me to just expense it anyway.

In any case, I could be wrong, but I'm fairly sure you'd have to be a member of the PMMP organisation to benefit from that even if we were partnered.

They do also have free student licenses if that would work for you.

If you hit me up on Discord we can discuss this more privately

remminiscent:
alright. when you got time, checkout the new commit on this PR and lmk if it looks good to you

src/world/generator/VoidGenerator.php
```diff
@@ -0,0 +1,64 @@
...
+			$chunk->setSubChunk($y, new SubChunk(Block::EMPTY_STATE_ID, [], clone $biomeArray));
+		}
+
+		if($chunkX >= 15 && $chunkX <= 16 && $chunkZ >= 15 && $chunkZ <= 16){
+			for($x = 0; $x < Chunk::EDGE_LENGTH; $x++){
+				for($z = 0; $z < Chunk::EDGE_LENGTH; $z++){
+					$worldX = ($chunkX * Chunk::EDGE_LENGTH) + $x;
+					$worldZ = ($chunkZ * Chunk::EDGE_LENGTH) + $z;
+					if($worldX >= 248 && $worldX <= 263 && $worldZ >= 248 && $worldZ <= 263){
+						$chunk->setBlockStateId($x, 69, $z, VanillaBlocks::STONE()->getStateId()); // 16x16 stone platform surrounding 256 69 256
```
dktapps: Don't call `VanillaBlocks` in a loop, it'll allocate and discard a new Block object each time. Get the state ID into a variable outside the loops and then use the variable here.

dktapps: This Y coordinate also doesn't seem right, in Bedrock vanilla it spawns right at y=-64

dktapps (changes requested):
Comparing this to Bedrock, the behaviour doesn't match. In Bedrock, the platform is centered on x=0,z=0, and generates at y=-64. The platform also extends from -16 to +16 inclusive (so 33 x 33 blocks) and the center block is cobblestone.

remminiscent:
lgtm 😆

dktapps:
Code seems fine, I'll test it tomorrow

dktapps (approved):

PR #6974, opened 2025-12-26
fixed sneak pressed

leolee3914:
### Related issues & PRs
Related to #6867 

https://github.com/pmmp/PocketMine-MP/pull/6867/files#diff-f405c0ec75b0721cb66a250d6938b9267fd68285ffa443dfe8e0e156afac04e0R229

if both START_SNEAKING and STOP_SNEAKING are always false, any changes of SNEAKING is never handled and Player->isSneakPressed() is never changed
this happened when a player is holding shift to enter a 1.5 block space and then release shift

### API changes

~~`Player->toggleSneak(bool $sneak, bool $sneakPressed = true) : bool` -> `Player->toggleSneak(?bool $sneak, bool $sneakPressed = true) : bool`~~

## Tests

i just tested the 4 states on #6548 with my own branch and haven't tested for PlayerToggleSneakEvent

### DIFF
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -2106,7 +2106,7 @@ class Player extends Human implements CommandSender, ChunkListener, IPlayer, Nev
 		}
 		$this->setSneakPressed($sneakPressed);
 
-		$ev = new PlayerToggleSneakEvent($this, $sneak ?? $this->sneaking, $sneakPressed);
+		$ev = new PlayerToggleSneakEvent($this, $sneak, $sneakPressed);
 		if(!$sneakChanged){
 			$ev->cancel();
 		}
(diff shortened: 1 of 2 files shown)


src/player/Player.php
```diff
@@ -2092,22 +2092,25 @@ public function toggleSprint(bool $sprint) : bool{
 		return true;
 	}
 
-	public function toggleSneak(bool $sneak, bool $sneakPressed = true) : bool{
-		if($sneak === $this->sneaking && $sneakPressed === $this->sneakPressed){
+	public function toggleSneak(?bool $sneak, bool $sneakPressed = true) : bool{
```
dktapps: I don't think we should be changing this signature, especially not on the stable branch. You could do `?? $this->player->isSneaking()` in the packet handler instead, that should be sufficient and shouldn't require any changes to `Player`

dktapps (changes requested):
I see the issue, but I think the fix can be simplified

dktapps (dismissed):
Will need to test this myself to verify, but this looks good to me

dktapps:
Thanks!

PR #6984, opened 2026-01-02
Implemented Wind charge

remminiscent:
Implemented Wind Charges. credits to @cjmustard. i used a workaround instead of adding a new method to the Block class
### Related issues & PRs
* Related to https://github.com/pmmp/PocketMine-MP/pull/6427

## Follow-up
how to not hardcode iron checks in door/trapdoor

## Tests
see previous pr for the clips

### DIFF
--- a/src/entity/projectile/WindCharge.php
+++ b/src/entity/projectile/WindCharge.php
@@ -35,7 +35,6 @@ use pocketmine\math\Vector3;
 use pocketmine\network\mcpe\protocol\types\entity\EntityIds;
 use pocketmine\world\particle\WindExplosionParticle;
 use pocketmine\world\sound\WindChargeBurstSound;
-use function ceil;
 use function floor;
 use function round;
 
@@ -92,7 +91,7 @@ class WindCharge extends Throwable{
 
 			$entityPos = $entity->getPosition();
 			$distance = $entityPos->distance($source) / 2.5;
-			$motion = $entityPos->subtractVector($source)->normalize();
+			$direction = $entityPos->subtractVector($source)->normalize();
 
 			$exposure = 1;
 			if ($entity->isUnderwater()){
@@ -110,17 +109,22 @@ class WindCharge extends Throwable{
 				return;
 			}
 
-			$entity->setMotion($entity->getMotion()->add(0, $impact * 0.4, 0)->addVector($motion->multiply($impact * $exposure)));
+			$entity->setMotion($entity->getMotion()->add(0, $impact * 0.4, 0)->addVector($direction->multiply($impact * $exposure)));
 		}
 	}
 
 	public function onHitBlock(Block $blockHit, RayTraceResult $hitResult) : void{
 		$bound = $this->getBound($this->getPosition(), self::RADIUS - 1);
-		for($x = $bound->minX; $x <= $bound->maxX; $x++) {
-			for($y = $bound->minY; $y <= $bound->maxY; $y++) {
-				for($z = $bound->minZ; $z <= $bound->maxZ; $z++) {
-					$block = $this->getWorld()->getBlockAt((int) floor($x), (int) floor($y), (int) floor($z));
-
+		$minX = (int) floor($bound->minX);
+		$minY = (int) floor($bound->minY);
+		$minZ = (int) floor($bound->minZ);
+		$maxX = (int) floor($bound->maxX);
+		$maxY = (int) floor($bound->maxY);
+		$maxZ = (int) floor($bound->maxZ);
+		for($x = $minX; $x <= $maxX; $x++) {
+			for($y = $minY; $y <= $maxY; $y++) {
+				for($z = $minZ; $z <= $maxZ; $z++) {
+					$block = $this->getWorld()->getBlockAt($x, $y, $z);
 					$block->onProjectileHit($this, $hitResult);
 				}
 			}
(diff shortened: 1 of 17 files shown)


src/entity/projectile/WindCharge.php
```diff
@@ -0,0 +1,142 @@
...
+
+	private function getBound(Vector3 $source, float $radius) : AxisAlignedBB {
+		return new AxisAlignedBB(
+			(int) floor($source->x - $radius - 1),
+			(int) floor($source->y - $radius - 1),
+			(int) floor($source->z - $radius - 1),
+			(int) ceil($source->x + $radius + 1),
+			(int) ceil($source->y + $radius + 1),
+			(int) ceil($source->z + $radius + 1)
+		);
```
dktapps: You can use `AxisAlignedBB::one()->expand()` for this
Also, the int casts are not necessary, AxisAlignedBB will convert them back to floats anyway

src/entity/projectile/WindCharge.php
```diff
@@ -0,0 +1,142 @@
...
+
+	public function attack(EntityDamageEvent $source) : void{
+		if(!$source instanceof EntityDamageByEntityEvent || ($entity = $source->getDamager()) === null){
+				return;
+		}
+
+		$this->setOwningEntity($entity);
+
+		$this->setMotion($entity->getDirectionVector()->multiply(1.5));
+	}
```
dktapps, line 59: This is confusing. How is an entity going to attack a projectile?

src/entity/projectile/WindCharge.php
```diff
@@ -0,0 +1,142 @@
...
+			if ($entity->isUnderwater()){
+				$exposure = 0.5;
+			}
+
+			$impact = (1.3 - $distance) * $exposure;
+			if ($impact <= 0) {
+				continue;
+			}
+
+			if (round($entityPos->getX(), 1) === round($source->getX(), 1) && round($entityPos->getZ(), 1) === round($source->getZ(), 1)) {
```
dktapps, line 106: Not sure what your intent was here, but this logic will only activate if the coordinates are the same. Usually this sort of check works by distance, so this doesn't make sense to me.

I may be wrong as I know basically nothing about wind charges, but this seems weird to me

src/entity/projectile/WindCharge.php
```diff
@@ -0,0 +1,142 @@
...
+				$entity->setMotion($entity->getMotion()->add(0, 0.75 * $exposure, 0));
```
dktapps, line 107: `$entity->addMotion()` can probably be used here

src/entity/projectile/WindCharge.php
```diff
@@ -0,0 +1,142 @@
...
+
+				return;
+			}
+
+			$entity->setMotion($entity->getMotion()->add(0, $impact * 0.4, 0)->addVector($motion->multiply($impact * $exposure)));
```
dktapps: These variable names are confusing. `$motion` isn't the velocity, it's the direction vector of the line between `$entity` and `$source`.

src/entity/projectile/WindCharge.php
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	public function onHitBlock(Block $blockHit, RayTraceResult $hitResult) : void{
+		$bound = $this->getBound($this->getPosition(), self::RADIUS - 1);
+		for($x = $bound->minX; $x <= $bound->maxX; $x++) {
+			for($y = $bound->minY; $y <= $bound->maxY; $y++) {
+				for($z = $bound->minZ; $z <= $bound->maxZ; $z++) {
+					$block = $this->getWorld()->getBlockAt((int) floor($x), (int) floor($y), (int) floor($z));
+
+					$block->onProjectileHit($this, $hitResult);
```
dktapps, line 128: This feels a bit suspicious. My issue with the original PR wasn't the addition of a new hook, but that the name was too vague. The projectile isn't hitting any of these blocks, so it feels a bit misleading to call onProjectileHit here

Also this will get fired for blocks with no collision boxes, which wouldn't be expected

src/entity/projectile/WindCharge.php
```diff
@@ -0,0 +1,142 @@
...
+			$entity->setMotion($entity->getMotion()->add(0, $impact * 0.4, 0)->addVector($motion->multiply($impact * $exposure)));
+		}
+	}
+
+	public function onHitBlock(Block $blockHit, RayTraceResult $hitResult) : void{
+		$bound = $this->getBound($this->getPosition(), self::RADIUS - 1);
+		for($x = $bound->minX; $x <= $bound->maxX; $x++) {
+			for($y = $bound->minY; $y <= $bound->maxY; $y++) {
+				for($z = $bound->minZ; $z <= $bound->maxZ; $z++) {
+					$block = $this->getWorld()->getBlockAt((int) floor($x), (int) floor($y), (int) floor($z));
```
dktapps: The conventional approach to this would be to int cast and floor the bounds instead, similar to World->getCollisionBlocks()

dktapps (commented):
The action for bells seems to be missing

remminiscent:
after research i SHOULD'VE done BEFORE opening the PR. it seems like logic for wind charges has changed quite a bit from 1.21.0... will fix & take your comments into consideration. 😕

remminiscent:
> The action for bells seems to be missing

Bells already have the logic for `onProjectileHit` 😄

src/entity/projectile/WindCharge.php
remminiscent: it acts similar to fire balls, hitting the entity will redirect it. similar to how players hit a ghast's fire ball to shoot it back at the ghast

cjmustard:
sorry for the poor code quality in my original pr, not quite sure what I was doing lol. best of luck with this one!

remminiscent:
> sorry for the poor code quality in my original pr, not quite sure what I was doing lol. best of luck with this one!

I originally thought dktapps liked the wind charge logic you made, which is why I went with it, but I guess I misunderstood.

Issue #6977, opened 2025-12-26
Mob heads Tile rotation are read/saved in legacy format


IvanCraft623:
### Plugin information

I haven't tested without plugins

### Problem description

Since 1.10 mob heads tile uses a "Rotation"(Float) tag that represents yaw instead of legacy "Rot"(Byte) tag

### Expected behaviour

Use modern format

### PocketMine-MP version

5.36.0

### PHP version

8.2

### Server OS

Any

remminiscent:
✅

IvanCraft623:
Closed by 892243d509ae39902e36ce4ac0e1f2e2bcbf4e12

PR #6995, opened 2026-01-06
Fix legacy NBT data

remminiscent:
Fix Mob heads Tile rotation are in legacy format
### Related issues & PRs
* Fixes https://github.com/pmmp/PocketMine-MP/issues/6977

## Backwards compatibility
I only made it use conversion from Rot -> Yaw and save. but setRotation still takes 0-15 due to BC

### DIFF
--- a/src/block/tile/MobHead.php
+++ b/src/block/tile/MobHead.php
@@ -58,7 +58,7 @@ class MobHead extends Spawnable{
 		if($nbt->getTag(self::TAG_ROTATION) instanceof FloatTag){
 			$yaw = $nbt->getFloat(self::TAG_ROTATION, 0.0);
 			$this->rotation = ((int) floor(($yaw * 16 / 360) + 0.5)) & 0xf;
-		} else {
+		}else{
 			$rotation = $nbt->getByte(self::TAG_ROT, 0);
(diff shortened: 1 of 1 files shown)


src/block/tile/MobHead.php
```diff
@@ -51,15 +54,21 @@ public function readSaveData(CompoundTag $nbt) : void{
...
 			$this->mobHeadType = $mobHeadType;
 		}
-		$rotation = $nbt->getByte(self::TAG_ROT, 0);
-		if($rotation >= 0 && $rotation <= 15){
-			$this->rotation = $rotation;
+
+		if($nbt->getTag(self::TAG_ROTATION) instanceof FloatTag){
+			$yaw = $nbt->getFloat(self::TAG_ROTATION, 0.0);
+			$this->rotation = ((int) floor(($yaw * 16 / 360) + 0.5)) & 0xf;
+		} else {
```
dktapps:
```suggestion
		}else{
```

dktapps (commented):
Mostly ok, just CS needs fixing

remminiscent:
thanks for doing it.

dktapps:
Sorry, I got distracted while I was waiting for the build to go green

PR #6998, opened 2026-01-12
Raise the soft limit of characters per page in books

### LINKED ISSUE #6996: Book pages are limited to 256 characters
### Plugin information

Bug happens without plugins

### Problem description

0. debug: 2
1. Open a "Book and Quill"
2. Write Cyrillic text in it (exactly 256 symbols)
```
Безусловно, перспективное планирование предопределяет высокую востребованность поставленных обществом задач. Банальные, но неопровержимые выводы, а также представители современных социальных резервов, вне зависимости от их уровня, должны быть заблокированы
```
3. Make sure that you cannot enter more characters due to the client-side limit
4. Close the book
5. Open the book again
6. Try to enter more characters on the same page and note that you can enter characters all the way to the end of the page
7. Close the book
8. Open the book again and see that the text has not been saved and there is an error in the log
```
Cancelled book edit due to page text exceeded soft limit of 256 chars
```
https://github.com/pmmp/PocketMine-MP/blob/400eb2dddf91a9c112aa09f3b498ffc8c85e98ed/src/network/mcpe/handler/InGamePacketHandler.php#L930

### Expected behaviour

In a single-player game, the server does not validate the number of characters and allows you to save more than 256 using this method.

Although this is a game bug, it causes confusion among players when they try to edit a page, and it doesn't save after closing the book. This can sometimes lead to the loss of a large amount of text.

I suggest slightly increasing the text limit per page to prevent players from encountering this issue.

This issue affects not only Cyrillic but also other Unicode characters.

In the case of Cyrillic, it is possible to enter up to 407 characters of text in the input field.

Image

### PocketMine-MP version

5.39.0

### PHP version

8.2

### Server OS

Linux

1.21.130

Gaprix:
Fixes #6996

### DIFF
--- a/src/item/WritableBookPage.php
+++ b/src/item/WritableBookPage.php
@@ -30,6 +30,10 @@ use function strlen;
 
 class WritableBookPage{
 	public const PAGE_LENGTH_HARD_LIMIT_BYTES = Limits::INT16_MAX;
+	//TODO: The client-side per-page character limit is inconsistent for non-ASCII text,
+	//allowing input beyond 256 chars. Use a slightly higher bounded soft limit to
+	//prevent rejected edits while still mitigating book-bomb attacks
+	public const PAGE_LENGTH_SOFT_LIMIT_CHARS = 512;
 	public const PHOTO_NAME_LENGTH_HARD_LIMIT_BYTES = Limits::INT16_MAX;
 
 	private string $text;
--- a/src/item/WritableBookPage.php
+++ b/src/item/WritableBookPage.php
@@ -30,10 +30,6 @@ use function strlen;
 
 class WritableBookPage{
 	public const PAGE_LENGTH_HARD_LIMIT_BYTES = Limits::INT16_MAX;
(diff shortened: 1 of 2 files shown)


src/item/WritableBookPage.php
```diff
@@ -30,6 +30,10 @@
 
 class WritableBookPage{
 	public const PAGE_LENGTH_HARD_LIMIT_BYTES = Limits::INT16_MAX;
+	//TODO: The client-side per-page character limit is inconsistent for non-ASCII text,
+	//allowing input beyond 256 chars. Use a slightly higher bounded soft limit to
+	//prevent rejected edits while still mitigating book-bomb attacks
+	public const PAGE_LENGTH_SOFT_LIMIT_CHARS = 512;
```
dktapps: I think it's better to put this in the InGamePacketHandler, since it's a protocol-specific oddity

dktapps (approved):

PR #7003, opened 2026-01-27
fix query token comparison

NopeNotDark:
Fix query token validation by using signed integers consistently

## Changes
Changed `$token = BE::readUnsignedInt($stream);` to `$token = BE::readSignedInt($stream);` as `$this->token` can be negative 

## Tests
Used a known working Python library ([mcquery](https://pypi.org/project/mcquery/)) to query a PocketMine-MP server.
Before the patch, mcquery was unable to query the server when it had a negative token (approximately 60% of the time)
After the patch, rebooted the server multiple times to regenerate tokens and mcquery successfully retrieved query information regardless of whether the token was positive or negative.

### DIFF
--- a/src/network/query/QueryHandler.php
+++ b/src/network/query/QueryHandler.php
@@ -104,7 +104,7 @@ class QueryHandler implements RawPacketHandler{
 
 					return true;
 				case self::STATISTICS: //Stat
-					$token = BE::readUnsignedInt($stream);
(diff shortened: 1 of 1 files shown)


octoberssss:
looks good to me

dktapps (approved):
Good catch, looks like this was incorrectly converted by #6799

PR #7006, opened 2026-01-30
Add an API to remove a player's item cooldown

ItsMax123:
Currently, plugins cannot remove a player's server-side item cooldown. The only publicly available API is [`Player->resetItemCooldown(Item $item, ?int $ticks = null)`](https://github.com/pmmp/PocketMine-MP/blob/4bda8738834730bd0be194d0a488229aa89d6d0e/src/player/Player.php#L774), which intentionally only supports cooldown values greater than 0.

### API changes
- The following API methods have been added:
  - `public Player->removeItemCooldown(Item $item) : void` - removes the player's cooldown time for the given item

## Tests
This testing plugin removes the cooldown on ender pearls and goat horns after 10 and 40 ticks respectively.
Video: https://github.com/user-attachments/assets/0f716c52-1369-486b-8062-902f97504a24
```php
<?php

declare(strict_types=1);

namespace Max\Test;

use pocketmine\event\player\PlayerItemUseEvent;
use pocketmine\item\EnderPearl;
use pocketmine\item\GoatHorn;
use pocketmine\scheduler\ClosureTask;

class Test extends PluginBase implements Listener {
    public function onEnable(): void {
    }

    public function onItemUse(PlayerItemUseEvent $event): void {
        $item = $event->getItem();
        if ($item instanceof EnderPearl) {
            $delay = 10;
        } elseif ($item instanceof GoatHorn) {
            $delay = 40;
        } else {
            return;
        }
        $this->getScheduler()->scheduleDelayedTask(new ClosureTask(function() use ($player, $item) {
            $player->removeItemCooldown($item);
        }), $delay);
    }
}
```

### DIFF
--- a/src/player/Player.php
+++ b/src/player/Player.php
@@ -779,6 +779,16 @@ class Player extends Human implements CommandSender, ChunkListener, IPlayer, Nev
 		}
 	}
 
+	/**
+	 * Removes the player's cooldown time for the given item.
+	 */
+	public function removeItemCooldown(Item $item) : void{
+		if($this->hasItemCooldown($item)){
+			unset($this->usedItemsCooldown[$item->getCooldownTag() ?? $item->getStateId()]);
+			$this->getNetworkSession()->onItemCooldownChanged($item, 0);
+		}
+	}
+
 	protected function checkItemCooldowns() : void{
 		$serverTick = $this->server->getTick();
 		foreach($this->usedItemsCooldown as $itemId => $cooldownUntil){


ItsMax123:
Also, thoughts on adding a `PlayerItemCooldownStartEvent`? That way plugins can override an item's cooldown without having to do hacky workarounds.
I can add it to this PR or create a new one if you'd like.

dktapps:
Code looks OK but I don't know if this will actually work. Needs playtesting

ItsMax123:
Ok, I added tests to the original post.

kaxyum:
> Also, thoughts on adding a `PlayerItemCooldownStartEvent`? That way plugins can override an item's cooldown without having to do hacky workarounds. I can add it to this PR or create a new one if you'd like.

Any news about that ? I think it is a good idea

PR #7007, opened 2026-01-31
Absorption Hearts: fixed client health desync

TrixNEW:
Video Of DeSync:
https://github.com/user-attachments/assets/6fc65bfe-6d78-4186-8164-77e31da16d55

## Summary
Fix health/absorption desync on damage by routing attack damage through `damageWithAbsorption`

When a Living entity took damage, absorption was reduced via `MODIFIER_ABSORPTION` on the damage event and then health was subtracted using the already-reduced `getFinalDamage()`. This caused health and absorption to desync since they weren't being updated together.

This introduces a protected `applyDamage(float)` hook in Entity, extracted from the end of `attack()`. Living overrides it to route damage through `damageWithAbsorption`, which consumes absorption first and applies any remainder to health in a single pass.

`damageWithAbsorption` is also useful beyond combat. Any code path that reduces health while absorption is active (e.g. certain enchantment effects) can now use it directly to avoid the same class of desync bug.

`MODIFIER_ABSORPTION` has been removed from `applyDamageModifiers` and the corresponding absorption update in `applyPostDamageEffects`, as `damageWithAbsorption` now owns that logic entirely.

### With Fix:
https://github.com/user-attachments/assets/3f3bad04-8379-4de6-88d1-43d010785b49
https://github.com/user-attachments/assets/34068956-03e7-4571-8ada-7402a811aecf

### Without Fix:
https://github.com/user-attachments/assets/3f565170-c628-45bb-b2ed-53fe192d7db6
https://github.com/user-attachments/assets/2c54034e-af27-49d6-8541-f24764c1b529


TrixNEW:
LGTM (yo dylan can i get an unban on pocketmine cord plz i repent)

dktapps:
No

IvanCraft623:
This fix is too hacky and doesn't actually fix the real issue

TrixNEW:
How is it too hacky? Its legit checking if it should remove absorption hearts before normal hearts.

TrixNEW:
We're calling this a hack like the entirely of pocketmine arent filled with "hacky" code 😆

e991f665b7e62df5a54fdef19053a4e75117b89:
This is far too hacky - as a PocketMine-MP core contributor, I think we should reject this PR, and "TrixNEW" should be permanently banned from the Discord.

This could be handled in a much more elegant way, have you tried checking if it should remove absorption hearts before normal hearts?

LGT**N**M

dktapps:
Implementing absorption as a damage modifier may have been a mistake, but removing the modifier would be a BC break and therefore cannot be done on PM5.

The solutions I can see in PM5:
- Set the absorption damage modifier last, after every other pre-event modifier has been added - this may not fully fix the issue if plugins modify the modifiers
- Set the absorption modifier to the amount of absorption hearts, ignoring final damage - however this would result in the final damage being negative in some cases, so I'm not sure about that

PR #7012, opened 2026-02-16
Implemented native spectator mode

xRookieFight:
### API Changes
- Added GameMode::NATIVE_SPECTATOR
### Behavioural changes
- Implemented native spectator mode

NOTE: It needs a new release for bedrock protocol in packagist because of PR [#](https://github.com/pmmp/BedrockProtocol/pull/340)[340](https://github.com/pmmp/BedrockProtocol/pull/340). I couldn't update it so we can't merge it without the release. Also thats why it doesn't pass PHPStan CI i guess

### DIFF
--- a/generated/lang/KnownTranslationFactory.php
+++ b/generated/lang/KnownTranslationFactory.php
@@ -982,6 +982,10 @@ final class KnownTranslationFactory{
 		return new Translatable(KnownTranslationKeys::GAMEMODE_CREATIVE, []);
 	}
 
+	public static function gameMode_native_spectator() : Translatable{
+		return new Translatable(KnownTranslationKeys::GAMEMODE_NATIVE_SPECTATOR, []);
+	}
+
 	public static function gameMode_spectator() : Translatable{
 		return new Translatable(KnownTranslationKeys::GAMEMODE_SPECTATOR, []);
 	}
--- a/generated/lang/KnownTranslationKeys.php
+++ b/generated/lang/KnownTranslationKeys.php
@@ -215,6 +215,7 @@ final class KnownTranslationKeys{
 	public const GAMEMODE_ADVENTURE = "gameMode.adventure";
 	public const GAMEMODE_CHANGED = "gameMode.changed";
 	public const GAMEMODE_CREATIVE = "gameMode.creative";
+	public const GAMEMODE_NATIVE_SPECTATOR = "gameMode.nativeSpectator";
 	public const GAMEMODE_SPECTATOR = "gameMode.spectator";
 	public const GAMEMODE_SURVIVAL = "gameMode.survival";
 	public const GAMEMODE_INFO = "gamemode_info";
--- a/generated/lang/KnownTranslationParameterInfo.php
+++ b/generated/lang/KnownTranslationParameterInfo.php
@@ -218,6 +218,7 @@ final class KnownTranslationParameterInfo{
 		Keys::GAMEMODE_ADVENTURE => [],
 		Keys::GAMEMODE_CHANGED => ["0"],
 		Keys::GAMEMODE_CREATIVE => [],
+		Keys::GAMEMODE_NATIVE_SPECTATOR => [],
 		Keys::GAMEMODE_SPECTATOR => [],
 		Keys::GAMEMODE_SURVIVAL => [],
 		Keys::GAMEMODE_INFO => [],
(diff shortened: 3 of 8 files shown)


kostamax27:
This `spectator mode` may lead to certain issues. Specifically, the client stops sending interaction with blocks and items, which can create challenges for developers who rely on this behavior.

xRookieFight:
> This `spectator mode` may lead to certain issues. Specifically, the client stops sending interaction with blocks and items, which can create challenges for developers who rely on this behavior.

Is there any solution for that? I didn't found any solution for that actually

Wraith0x10:
We could introduce a new `GameMode::NATIVE_SPECTATOR`

xRookieFight:
> We could introduce a new `GameMode::NATIVE_SPECTATOR`

That would be better, i'll do that. Current implementation is breaking interactions and i can't see any solution for that

xRookieFight:
> We could introduce a new `GameMode::NATIVE_SPECTATOR`

Done but its pretty out of PR context now. Its also not same as vanilla but best solution we currently have found is this i guess.

dktapps:
I don't really see the point in this. The whole reason we never used vanilla spectator to begin with is precisely because of the unwanted client side filtering of player actions. Is there any tangible benefit to this over the current approach?

xRookieFight:
> I don't really see the point in this. The whole reason we never used vanilla spectator to begin with is precisely because of the unwanted client side filtering of player actions. Is there any tangible benefit to this over the current approach?

We can put it as different gamemode, i don't see any solution for the player actions

dktapps:
Sure, but that isn't answering my question. What does this extra gamemode offer that PM's existing spectator mode doesn't?

Wraith0x10:
> Sure, but that isn't answering my question. What does this extra gamemode offer that PM's existing spectator mode doesn't?

Seeing other spectators as a transparent head

xRookieFight:
> Sure, but that isn't answering my question. What does this extra gamemode offer that PM's existing spectator mode doesn't?

Making vanilla spectator gamemode improves gameplay in minigames servers etc.(current one is kinda creative-based so it looks bad in these servers) Also we aren't breaking the API so it shouldn't make a problem

dktapps:
> Seeing other spectators as a transparent head

Ok, valid.
Well, it'd be worth investigating if there's a way to achieve that with the existing spectator setup.
Either that, or finding a way to enable player-block interactions in vanilla spectator.

> Making vanilla spectator gamemode improves gameplay in minigames servers etc.(current one is kinda creative-based so it looks bad in these servers)

This doesn't really answer my question either. In what way does it look bad?

> Also we aren't breaking the API so it shouldn't make a problem

Half-assing this creates cruft that'll have to be cleaned up later on. So there needs to be a meaningful benefit to justify it that can't be achieved any other way.

dktapps:
The block interaction behaviour would likely be affected by this code: https://github.com/pmmp/PocketMine-MP/blob/dec84fe67dc3319e33f5066e5d03b16f00f2ff4c/src/network/mcpe/NetworkSession.php#L1109

Specifically LAYER_SPECTATOR may be of interest

xRookieFight:
> This doesn't really answer my question either. In what way does it look bad?

The overlay looks like creative. If we can make interactions work with this we should merge the default spectator gamemode to this because i also didn't like seperating gamemode idea.

kostamax27:
> The block interaction behaviour would likely be affected by this code:
> 
> https://github.com/pmmp/PocketMine-MP/blob/dec84fe67dc3319e33f5066e5d03b16f00f2ff4c/src/network/mcpe/NetworkSession.php#L1109
> 
> Specifically LAYER_SPECTATOR may be of interest

There is another issue: in this game mode, the HUD is hidden. How is a player supposed to know which slot contains the required item if the inventory isn’t visible? Even if they randomly select the correct slot, they still won’t be able to see which item they picked up because the hand is not visible.

I don’t think this game mode is worth it, especially considering how many problems it creates. It might be better to implement this functionality through a plugin instead. Or we could turn it into its own game mode and name it something like `NATIVE_SPECTATOR` or `NEW_SPECTATOR`.

xRookieFight:
> There is another issue: in this game mode, the HUD is hidden. How is a player supposed to know which slot contains the required item if the inventory isn’t visible? Even if they randomly select the correct slot, they still won’t be able to see which item they picked up because the hand is not visible.
> 
> I don’t think this game mode is worth it, especially considering how many problems it creates. It might be better to implement this functionality through a plugin instead. Or we could turn it into its own game mode and name it something like `NATIVE_SPECTATOR` or `NEW_SPECTATOR`.

I thought that was vanilla behaviour

kostamax27:
> I thought that was vanilla behaviour

Perhaps, but is that a sufficient reason to break compatibility if the previous version still works correctly?

xRookieFight:
I don't think invisible HUD would break any compatibility to be honest

PR #7032, opened 2026-04-01
vertical fly speed

remminiscent:


kostamax27:
It might be better to allow developers to adjust the vertical flight speed separately :thinking:

remminiscent:
> It might be better to allow developers to adjust the vertical flight speed separately 🤔


I did consider that, but I think `getFlightSpeedMultiplier()` is meant to be a general multiplier for flight speed as a whole, adding a separate vertical one would feel a bit inconsistent with that idea.



Renaming the function would also cause a BC break, so I'll wait for dylan to give his thoughts, if he ever even sees this 😆

dktapps:
I'm on the fence with it. It kinda sucks to couple them if we have the ability to change them independently, but on the other hand, it's kinda weird to have `getVerticalFlightSpeedMultiplier`

remminiscent:
> I'm on the fence with it. It kinda sucks to couple them if we have the ability to change them independently, but on the other hand, it's kinda weird to have `getVerticalFlightSpeedMultiplier`

yes, very weird. why not add an optional parameter to `setFlightSpeedMultiplier()` like `bool $syncVertical = true`?

but then there would be no way for explicitly vertical speed

PR #7040, opened 2026-05-09
Bedrock 1.26.20

dktapps:


kostamax27:
Hello, I found a couple of bugs related to the new version. I don’t think this PR is ready yet.

